PR #416, opened 2017-01-11
Support per-object undo logs and rollback

no2chem:
This patch adds support for VersionLockedObjects to rollback
by adding a per-object undo log. The system tracks
active transactions and provides an interface to query
what the oldest transaction in the system is.

It uses this data to bound the size of the undo log.

PR #445, opened 2017-02-26
Undo log for objects

no2chem:
This PR reopens #416, which adds per-object undo log support as well as rollback support.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/TransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/TransactionalContext.java
@@ -5,6 +5,9 @@ import lombok.extern.slf4j.Slf4j;
 import java.time.Duration;
 import java.util.Deque;
 import java.util.LinkedList;
+import java.util.NavigableSet;
+import java.util.NoSuchElementException;
+import java.util.concurrent.ConcurrentSkipListSet;
 
 /** A class which allows access to transactional contexts, which manage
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/object/transactions/TransactionalContext.java
```diff
@@ -86,6 +110,7 @@ public static AbstractTransactionalContext newContext(AbstractTransactionalConte
      */
     public static AbstractTransactionalContext removeContext() {
         AbstractTransactionalContext r = getTransactionStack().pollFirst();
+        contextSet.remove(r);
```
kspirov: I would add something like:

     public static AbstractTransactionalContext removeContext() {
Deque<AbstractTransactionalContext> stack = getTransactionStack();
if (stack.isEmpty()) {
    threadTransactionStack.set(null);
}

Otherwise if the current thread is polled and it is returned back to the pool, you are doing a small memory leak.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a1122ff**.

**Throughput**: _Average_: 38.4615384615, _Standard Deviation_: 8.11565803404
**Latency**: _Average_: 256.208, _Standard Deviation_: 117.712755197

[Pull Request #445 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/0uzTgRS1nAme6seVboqyn7obBHH5WWzZ)

PR #447, opened 2017-02-27
Add option to disable hole-filling

no2chem:
This patch adds a flag to the runtime to disable hole filling.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
@@ -193,6 +193,21 @@ public class BackpointerStreamView extends AbstractQueuedStreamView {
                     }
                 }
 
+                // If hole filling is disabled, we will retry forever.
+                if (runtime.isHoleFillingDisabled()) {
+                    int i = 0; // For exponential backoff.
+                    while (currentEntry.getType() == DataType.EMPTY) {
+                        currentEntry =
+                                runtime.getAddressSpaceView().read(currentRead);
+                        try {
+                            Thread.sleep(1 << i);
+                            i++;
+                        } catch (InterruptedException ie) {
+                            throw new RuntimeException(ie);
+                        }
+                    }
+                }
+
                 // If we STILL don't have the data, now we need to do a hole
                 // fill.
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -193,6 +193,21 @@ protected boolean fillReadQueue(final long maxGlobal,
...
                 }
 
+                // If hole filling is disabled, we will retry forever.
+                if (runtime.isHoleFillingDisabled()) {
+                    int i = 0; // For exponential backoff.
+                    while (currentEntry.getType() == DataType.EMPTY) {
+                        currentEntry =
+                                runtime.getAddressSpaceView().read(currentRead);
+                        try {
+                            Thread.sleep(1 << i);
```
slfritchie: I don't understand the reason for adding isHoleFillingDisabled(), sorry.  But if the client is a bit sensitive to latency, then perhaps add a maximum limit on any single sleep time here?

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -193,6 +193,21 @@ protected boolean fillReadQueue(final long maxGlobal,
...
+                            i++;
+                        } catch (InterruptedException ie) {
+                            throw new RuntimeException(ie);
```
slfritchie: Is it worth escalating the problem here?  Or better to ignore this particular sleep interruption and go around again through the loop?

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
no2chem: Someone wanted a client which did not hole fill. Seems like a reasonable request.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
no2chem: If an interrupted exception occurs, some entity (perhaps via a signal) is requesting the exception be propagated so that graceful termination can occur.

PR #448, opened 2017-02-27
Flush log unit servers

zalokhan:
### Flushes the write queue of the log unit server.

This call would block until the server has successfully flushed its current state to disk.
Every time we need to seal the epoch, we need to flush the queues of all log unit servers.


zalokhan:
Is there any use case where the client would want to flush the writes manually?

dahliamalkhi:
I guess it *might* be useful for certain clients. They can insist to speed up their turnaround, rather than wait for a batch to fill up. 

This is in no way urgent. We can keep this in mind.

I am going to merge this branch for now.

PR #453, opened 2017-03-04
Performance improvement in transactional context

rogermichoud:
This addresses chattiness issue during transactional context. Upon access of an object, the version would be validated against the global snapshot in order to make a decision of syncing it or not. If the version was not at the global snapshot, it would sync. The problem is the object would almost never be at the global snapshot, because the global snapshot is the tail of the entire log, not the tail for the object. 

On each access, we would sync and ask the tail of the object to the sequencer. Even without context switch.

This changes also introduces a cache, that will keep the version of the objects at the global snapshot. It gets populated upon first access to each object during the transaction. It means that, after a context switch, if both threads had the same version of the object at their global snapshot (each global snapshot might be different), we would not have to resync the object. We would only unroll for optimistic changes.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -192,6 +192,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
      * this function...
      * @param underlyingObject  The object to update.
      * @param timestamp         The timestamp to update the object to.
+     * @return The version number of the object after sync.
      */
     public long syncObjectUnsafe(VersionLockedObject<T> underlyingObject,
                                       long timestamp) {
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -193,7 +193,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
      * @param underlyingObject  The object to update.
      * @param timestamp         The timestamp to update the object to.
      */
-    public void syncObjectUnsafe(VersionLockedObject<T> underlyingObject,
+    public long syncObjectUnsafe(VersionLockedObject<T> underlyingObject,
                                       long timestamp) {
         underlyingObject.getStreamViewUnsafe().remainingUpTo(timestamp).stream()
             // Turn this into a flat stream of SMR entries
@@ -227,6 +227,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
                         }
                     });
             });
+        return underlyingObject.getVersionUnsafe();
     }
 
     @Override
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
@@ -10,14 +10,7 @@ import org.corfudb.runtime.object.ICorfuSMRAccess;
 import org.corfudb.runtime.object.ICorfuSMRProxyInternal;
 import org.corfudb.runtime.object.VersionLockedObject;
 
-import java.util.Collections;
-import java.util.ConcurrentModificationException;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -55,6 +48,13 @@ public class OptimisticTransactionalContext extends AbstractTransactionalContext
     @Getter
     private final Set<ICorfuSMRProxyInternal> modifiedProxies =
             new HashSet<>();
+
+    /** Keep the version (at global snapshot time)
+     *  of each accessed object.
+     */
+    private final Map<ICorfuSMRProxyInternal, Long> streamSnapshots =
+            new HashMap<>();
+
 
     OptimisticTransactionalContext(TransactionBuilder builder) {
         super(builder);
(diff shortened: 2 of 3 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **c998708**.

**Throughput**: _Average_: 35.7142857143 [executions/sec], _Standard Deviation_: 14.686867659
**Latency**: _Average_: 276.484 [ms], _Standard Deviation_: 138.483976488

[Pull Request #453 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/TfOBlbtLQYS7UUQ8DigXO8QnK2bWewqJ)

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -193,7 +193,7 @@ else if (underlyingObject.isOptimisticallyUndoableUnsafe()){
      * @param underlyingObject  The object to update.
      * @param timestamp         The timestamp to update the object to.
      */
```
no2chem: Could you update the javadoc to reflect the new return?

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -56,6 +49,13 @@
     private final Set<ICorfuSMRProxyInternal> modifiedProxies =
             new HashSet<>();
 
+    /** Keep the version (at global snapshot time)
```
no2chem: I don't know if this does that you want it to. The proxy is just a thin wrapper, the version is in the VersionLockedObject itself. Keeping versions would be quite expensive for large objects like maps.

rogermichoud: What I am trying to achieve is to keep the version number of the underlying object at the global snapshot time. Upon access (after a context switch), I compare if the underlying object is at the same version number than where we started from in this context (the version of the object at the global snapshot).

Since we always compare the global snapshot against the version of the object, they will almost never be the same (global vs stream tail). The first time we access the object in the context, we sync the object (readto to the global tail) and we store the stream tail in the cache.

Next time we context switch, if this version number is the same in both context, we don't sync the object, we only perform optimistic rollbacks if any.

The confusion is in my language, I should have said version number, not version. Will change it.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -97,12 +97,18 @@
...
-        // next, if the version is older than what we need
-        // sync.
-        if (object.getVersionUnsafe() < getSnapshotTimestamp()) {
-            proxy.syncObjectUnsafe(proxy.getUnderlyingObject(), getSnapshotTimestamp());
+        // Sync only if we never read the object in this transaction
+        // or if the version is older than what we need
+        OptimisticTransactionalContext oCtxt = (OptimisticTransactionalContext)
+                object.getModifyingContextUnsafe();
+        long snapshotOfThisStream =  streamSnapshots.computeIfAbsent(proxy,
+                k -> k.syncObjectUnsafe(k.getUnderlyingObject(), getSnapshotTimestamp()));
```
dahliamalkhi: There is a bit of redundant work in one case. If this is the first time this object is accessed in a transaction, but the underlyingObject version is right, we invoke syncObjectUnsafe unnecessarily.

Why not change the logic to:

if (oCtxt == null || oCtxt != this &&
   proxy.getUnderlyingObject().getVersionUnsafe() < getSnapshotTimestamp() ) {
     long snapshotOfThisStream = ...; 
     if (..getVersionUnsafe() < snapshotOfThisStream) proxy.syncObjectUnsafe(..)
}

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
dahliamalkhi: @no2chem , the proxy is used here only as the identifier of a stream, it is the key to the hashmap.

no2chem: I see, I got version and version number confused. I probably should have read through the code more carefully. Unfortunately, I'm not sure it improves performance at all... not really sure why.

rogermichoud: I also didn't see a huge improvement in the multithreaded case. It improves significantly in memory runs and disk singlethreaded. (the runs that are not shown currently in the performance graphs.) We might plateau in the multithreaded disk runs because of another bottleneck. I will try to investigate further what is happening.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
rogermichoud: This would happen only if the global tail address at snapshot time corresponds to the version number of the object, which means the latest global entry was from the object we are accessing.

We should get that case right as well. Will do the changes.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **0f1f1b5**.

**Throughput**: _Average_: 43.4782608696 [executions/sec], _Standard Deviation_: 20.7361678679
**Latency**: _Average_: 217.333 [ms], _Standard Deviation_: 139.496996781

[Pull Request #453 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Lc09QLgof45l0IvwxfqxkPdcOTVoXUZR)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **73ae6e2**.

 [Pull Request #453 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/1aEstlUKW2Wv9Hfl4WlX5YylxOsO414R)

dahliamalkhi:
@rogermichoud  : should this PR be closed? It is probably obsoleted by #499 , and in any case, would require significant re-factoring.

PR #456, opened 2017-03-06
minor fix to optimistic rollback flag

dahliamalkhi:
without this, tests like FGMap::loadsFollowedByGetsConcurrent fail intermittently.


dahliamalkhi:
tests still failing. need to look further into it.

PR #457, opened 2017-03-06
fix stream bug

no2chem:
Fix a bug which would occur during hole fill where unsuccessful
hole fill attempts would fail to read the correctly written
entry. 

This patch forces the client to re-read on the overwrite exception
of a hole fill, correctly reading failed hole-fills.

In addition, it also cleans up AbstractQueuedStreamView::getNextEntries
so that it is more readable.


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d7903d7**.

**Throughput**: _Average_: 55.5555555556 [executions/sec], _Standard Deviation_: 36.9116395708
**Latency**: _Average_: 169.682 [ms], _Standard Deviation_: 131.501934875

[Pull Request #457 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ZU9csqvaZpoBU6O7XxMHWOrWM0Hy52a4)

PR #449, opened 2017-03-02
Quorum replication and some correspondign code refactoring (continued)

kspirov:
This thread continues https://github.com/CorfuDB/CorfuDB/pull/438, the other pull request will be closed.
The code is rebased top of the current master.

PR #461, opened 2017-03-06
Overwrite / trim file support on the replica side.

kspirov:
Next independent/incremental submission for the quorum replication, based on https://github.com/CorfuDB/CorfuDB/pull/449

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -4,7 +4,11 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.DataInputStream;
+import java.io.BufferedInputStream;
 import java.io.OutputStream;
+import java.io.DataOutputStream;
+import java.io.FileOutputStream;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.FileChannel;
@@ -99,10 +103,9 @@ public class StreamLogFiles implements StreamLog {
             Collection<File> files = FileUtils.listFiles(dir, extension, true);
 
             for (File file : files) {
-                try {
-                    FileInputStream fIn = new FileInputStream(file);
-                    FileChannel fc = fIn.getChannel();
+                try (FileInputStream fIn = new FileInputStream(file)){
 
+                    FileChannel fc = fIn.getChannel();
 
                     ByteBuffer metadataBuf = ByteBuffer.allocate(METADATA_SIZE);
                     fc.read(metadataBuf);
@@ -163,7 +166,7 @@ public class StreamLogFiles implements StreamLog {
     @Override
     public void trim(LogAddress logAddress) {
         SegmentHandle handle = getSegmentHandleForAddress(logAddress);
-        if (!handle.getKnownAddresses().contains(logAddress.getAddress()) ||
+        if (!handle.getKnownAddresses().containsKey(logAddress.getAddress()) ||
                 handle.getPendingTrims().contains(logAddress.getAddress())) {
             return;
         }
@@ -388,6 +391,37 @@ public class StreamLogFiles implements StreamLog {
         return set;
     }
 
+    private boolean isOverwritten(SegmentHandle fh, long filePosition) {
+        return fh.overwrittenFilePositions.contains(filePosition);
+    }
+
+
+    private void readOverwrittenPositions(SegmentHandle fh) throws IOException {
+        synchronized (fh.lock) {
+            File f = new File(fh.getOverwriteFileName());
+            if (f.exists()) {
+                try (DataInputStream is = new DataInputStream(new BufferedInputStream(new FileInputStream(fh.getOverwriteFileName())))) {
+                    while (is.available() > 4) {
+                        fh.overwrittenFilePositions.add(is.readLong());
+                    }
+                }
+            }
+        }
+    }
+
+    private void addOverwrittenPosition(SegmentHandle fh, long position) throws IOException {
+        synchronized (fh.lock) {
+            File f = new File(fh.getOverwriteFileName());
+            if (!f.exists()) {
+                f.getParentFile().mkdirs();
+            }
+            try (DataOutputStream os = new DataOutputStream(new FileOutputStream(f.getAbsolutePath(), f.exists()))) {
+                os.writeLong(position);
+            }
+            fh.overwrittenFilePositions.add(position);
+        }
+    }
+
     /**
      * Find a log entry in a file.
      *
@@ -420,13 +454,16 @@ public class StreamLogFiles implements StreamLog {
 
         Metadata headerMetadata = Metadata.parseFrom(headerMetadataBuf.array());
 
-        fc.position(fc.position() + headerMetadata.getLength());
-        ByteBuffer o = ByteBuffer.allocate((int) logFileSize - (int) fc.position());
+        long initialPosition = fc.position() + headerMetadata.getLength();
+        fc.position(initialPosition);
+
+        ByteBuffer o = ByteBuffer.allocate((int)(logFileSize - fc.position()));
         fc.read(o);
         fc.close();
         o.flip();
 
         while (o.hasRemaining()) {
+            long filePosition = initialPosition+o.position();
 
             short magic = o.getShort();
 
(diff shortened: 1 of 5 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5946821**.

**Throughput**: _Average_: 45.4545454545 [executions/sec], _Standard Deviation_: 19.7362777939
**Latency**: _Average_: 213.088 [ms], _Standard Deviation_: 132.872616652

[Pull Request #461 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/SOmQMNqmuSb8FmIaR5fZwx2xDJiRLYhv)

no2chem:
@Maithem how does this interact with your PR?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5946821**.

 [Pull Request #461 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/FeTgsxIgIZIJBS5ysW4eacMpDGw5LkPv)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **2d88a02**.

 [Pull Request #461 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/XhmZ8bV4s21kdG6wcXK9oQakE2AFedkY)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -275,13 +306,15 @@ private LogData readRecord(FileHandle fh, long address)
                     }
                 }
 
-                if (address == -1) {
+                if (isOverwritten(fh, filePosition)) {
+                    log.debug("Trim at {} hit, reading (size={}), position {}.", address, metadata.getLength(), filePosition);
+                } else if (address == -1) {
                     //Todo(Maithem) : maybe we can move this to getChannelForAddress
-                    fh.knownAddresses.add(entry.getGlobalAddress());
+                    fh.knownAddresses.put(entry.getGlobalAddress(), filePosition);
```
dahliamalkhi: @Maithem the only thing I would worry about here is atomicity, since the steamLog and the overwriteAddresses log are updates separately.

Do you see any potential issue here?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
```diff
@@ -78,10 +79,30 @@ default Long getRank() {
      * @param rank The rank of this append.
      */
     default void setRank(long rank) {
+        if (rank>=0 && getRank()<0) {
+            throw new IllegalStateException("Cannot set rank when the overwrite is permitted");
```
dahliamalkhi: explain..

kspirov: The idea was if you have after quorum replication of 3 replicas values 1 1 and 2, to send such write request to the replica that has 1.

Good to discuss this as there is another approach here - just when we are up to throw and overwrite exception to read the old rank and compare.
Extra read operation would not add extra cost, as overwrite exceptioon is rare, also there is no atomicity issue as you have a lock for the segment.

I could change it to this approach, but probably we will need to change Rank as long to Rang as Comparable or Rank as long+uuid. Do you think it is OK to have rank = long+uuid globally, not for quorum replication only. If yes, this solves all problems.

kspirov: Actually in append(...) you don't hold the lock, you hold it on writeRecord(...) that is called from append, but this is solvable.

I believe this was the reason I decided not to read the old rank here - I didn't want to change your locks for such rare situation. The disadvantage was that the responsibility to force overwrite was moving to the client.

I don't believe if the write lock is extended to append from write, something bad will happen.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
kspirov: Yes, please double check:

The only that could look like an issue IMHO is:

  readOverwrittenPositions(fh);
   readRecord(fh, -1);

Here you could ask - what if between these 2 new record is overwritten? readRecord() will return the old value. But as we are in a different thread and we don't break the sequential consistency, I don't think it is an issue.

Otherwise you see I hold the same write lock as for the segment.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
dahliamalkhi: I am not sure this rank is being used for anything other than quorum replication; I think it was prepared in advance for that purpose. no?

kspirov: This rank field was existing before me, but currently is only saved and transfered - never red or actually used.

dahliamalkhi: OK, now I think I get it.

Indeed, I believe the rank field's only purpose is to support quorum replication. Therefore, it should not have such a conflict, where some module set it to -1 and another module tries to set a positive rank.

A couple of more comments: 

I understand why you refer to a quorum operation as "overwrite", but it might be confusing, since there is no real overwrite. Rather, I think of your quorum replication implementation as transforming the stream-log from a write-once-address-space logging unit, into a ranked-address-space logging-unit (and a Paxos "acceptor"). I would suggest to comment and use names that reflect this.

 Given this, I believe that the test above should be:
     if (rank < getrank()) throw exception ..
Unless I am again missing something?

Last, to your question about rank: Again, unless I am missing something, rank is only used for quorum replication.

kspirov: Thanks for the detailed comments, suggestions appreciated&accepted, of course.
Next steps:
1) I just resolved the conflicts on the top of your "compact" change and I will update the request
2) I will plug into your compact and unit test, that compact for overwritten entries also works
3) I will change to your approach with Rank + one very carefull change in the locking scope (your existing lock in segment handler), to preserve the atomicity when decide whether to throw an OverwriteException based on rank.
I will keep the changes in this review if I can.

kspirov: Step 1 of 3 done

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **0ade9d8**.

 [Pull Request #461 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/kaqjRZkKKw5l0CYgxQfDdbPGfPJ2yZMW)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -388,6 +391,37 @@ private LogData getLogData(LogEntry entry) {
...
+        return fh.overwrittenFilePositions.contains(filePosition);
+    }
+
+
+    private void readOverwrittenPositions(SegmentHandle fh) throws IOException {
+        synchronized (fh.lock) {
+            File f = new File(fh.getOverwriteFileName());
+            if (f.exists()) {
+                try (DataInputStream is = new DataInputStream(new BufferedInputStream(new FileInputStream(fh.getOverwriteFileName())))) {
+                    while (is.available() > 4) {
```
Maithem: The segment lock is used because we can't read partial files (i.e. can't read while entries are partially written, we need a delimiter), so if a write is happening a lock would prevent reading and vice versa. Since you're code doesn't need that requirement I don't see why you need to use a lock here.

Also note, that there is only a single writer thread that is calling append, so you also dont need a lock in addOverwrittenPosition.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -388,6 +391,37 @@ private LogData getLogData(LogEntry entry) {
...
+    }
+
+    private void addOverwrittenPosition(SegmentHandle fh, long position) throws IOException {
+        synchronized (fh.lock) {
+            File f = new File(fh.getOverwriteFileName());
+            if (!f.exists()) {
+                f.getParentFile().mkdirs();
+            }
+            try (DataOutputStream os = new DataOutputStream(new FileOutputStream(f.getAbsolutePath(), f.exists()))) {
+                os.writeLong(position);
```
Maithem: You also need to write checksums too.

dahliamalkhi:
@Maithem thanks for the review, very helpful!

kspirov:
@Maithem. I like this, it is clear. Let me explain why I didn't propose this solution immediately - I was concerned I will need to remove return getLogData(entry); from readRecord and to leave the loop to continue until the end. This means performance regression! BUT with the cache, I see you will anyway have getSegment() with readRecord(-1) before calling readRecord(position). This means the performance regression is not likely to happen.

Good idea, thanks.

Maithem:
@kspirov You're welcome. I've implemented what we have discussed (pull request #484)
Now you can add some type of ref counting to AddressMetaData from #484 to detect overwrites.

kspirov:
I will add as separate pull requests adding of rank and compact(), closing this pull requestr as the Maithem implemented what we concluded is necessary, and his code does more correct refactoring.

PR #466, opened 2017-03-07
when resetting the object, start from Address.NEVER_READ

no2chem:
During object reset, the version was being erroneously reset to 0
which caused the object to skip the first update. This patch
fixes that bug by setting the address to NEVER_READ instead.


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **0831f89**.

**Throughput**: _Average_: 38.4615384615 [executions/sec], _Standard Deviation_: 17.6553124915
**Latency**: _Average_: 249.166 [ms], _Standard Deviation_: 151.184213607

[Pull Request #466 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ONb31N22Fd8lpxF5wAsq9G135YAXSaOB)

PR #467, opened 2017-03-07
added undo tests

dahliamalkhi:
these undo tests surfaced a problem which has been dealt with in #466 . Adding these tests to the repo now.


dahliamalkhi:
oops, wrong branch.

PR #469, opened 2017-03-07
adding undo tests and some fixes and cleanups to it

dahliamalkhi:
This branch contains several tests that validate undo-correctness. 

One test -- UndoTest::canUndoAfterNoUndo -- is about performance, not correctness. It verifies that VersionLockedObject::optimisitcallyUndoable is properly reset (currently, it is not). 

This PR contains a fix to setting the optimistic flag properly, as well as some cleanups around the logic of optimistic undo/redo.

### DIFF
--- a/test/src/test/resources/logback-test.xml
+++ b/test/src/test/resources/logback-test.xml
@@ -8,7 +8,7 @@
 
     <logger name="org.corfudb.runtime.object" level="DEBUG"/>
 
-    <root level="ERROR">
+    <root level="TRACE">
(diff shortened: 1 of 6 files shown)


test/src/test/resources/logback-test.xml
```diff
@@ -6,7 +6,9 @@
         </encoder>
     </appender>
 
-    <root level="TRACE">
```
no2chem: take out this leftover debug config

PR #471, opened 2017-03-07
Compact

Maithem:
Rebased against master.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -259,7 +259,7 @@ public class LogUnitServer extends AbstractServer {
 
     @ServerHandler(type = CorfuMsgType.TRIM)
     private void trim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        batchWriter.write(new LogAddress(msg.getPayload().getPrefix(), msg.getPayload().getStream()), null);
+        batchWriter.trim(new LogAddress(msg.getPayload().getPrefix(), msg.getPayload().getStream()));
         //TODO(Maithem): should we return an error if the write fails
         r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
     }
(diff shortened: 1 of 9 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **f63704e**.

**Throughput [CorfuDiskNode:10 thread(s)]**: _Average_: 45.4545454545 [executions/sec], _Standard Deviation_: 18.8311330049
**Latency [CorfuDiskNode:10 thread(s)]**: _Average_: 210.77 [ms], _Standard Deviation_: 128.462839374

 [Pull Request #471 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/M5YgKJawJy6opq4mo6LQ2Nu97dMnwMvl)

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -256,8 +245,8 @@ private void fillHole(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ct
 
     @ServerHandler(type = CorfuMsgType.TRIM)
     private void trim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        trimMap.compute(msg.getPayload().getStream(), (key, prev) ->
-                prev == null ? msg.getPayload().getPrefix() : Math.max(prev, msg.getPayload().getPrefix()));
+        batchWriter.trim(new LogAddress(msg.getPayload().getPrefix(), msg.getPayload().getStream()));
+        //TODO(Maithem): should we return an error if the write fails
```
no2chem: no reason, it's just a hint

kspirov:
@no2chem @Maithem  Thanks for the review, I understand the purpose of this code and it looks good for me. Let me quickly explain the difference between this and the conflicting https://github.com/CorfuDB/CorfuDB/pull/461 . The two reviews have different purposes - here you want to trim log entries at given position, so they are compacted later, in my case I want to overwrite entry on a given position (in very exceptional situation when there is quorum, but some of the other servers have different value).

My approach and the approach of Dalia are a bit different - the code adds to the trim list by log entry position, my code adds to similar trim file by absolute position on the file. So for example the entry 10 here is trimmed on 10, but in my implementation it can have 2 trims at 200 and 300, and it can have valid value on 400. But my review doesn't do real trim or compact, it leaves this for later.

I believe these 2 implementations are a bit orthogonal and it is OK to stay independent. I can just rename "trimmed" with "replaced". Any thoughts?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **f63704e**.

 [Pull Request #471 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/WuxImaOZGNqeP6j9JPsFTtkcOGJ6KEBl)

Maithem:
@kspirov Can you please elaborate more. If i understand correctly, in you are referring to trimming the the metadata file (i.e. *.overwrite), is that correct? If that's the case they do seem orthogonal issues.

PR #472, opened 2017-03-08
Fix the version after undo operation in rollbackUnsafe

rogermichoud:
Previously, once we were undoing a record, we would leave the object at the version of the
record we undid.
This would trigger a resetObjectUnsafe, because once we undo the latest record, we will check that
the current version of the object is <= version we wanted to unroll to. This would not be the case
and would iterate again through the while loop, generating a NoRollBackException (undoLog is empty).

In this commit, we use the backpointermap to get the previous version of the object and update the
version to its predecessor.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -220,7 +220,11 @@ public class VersionLockedObject<T> {
                     .doUndo(object, undoRecord.getUndoRecord(),
                                     undoRecord.getSMRArguments());
 
-            this.version = undoRecord.getEntry().getGlobalAddress();
+            // After undoing the record, we should get back to the previous version
+            // We use the backpointerMap to get the previous entry version
+            // If none found, throw NoRollbackException.
+            this.version = undoRecord.getEntry().getBackpointerMap().values().stream().
+                    findFirst().orElseThrow(NoRollbackException::new);
 
 
             // check if we rolled back to the requested version
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -223,8 +223,8 @@ public class VersionLockedObject<T> {
             // After undoing the record, we should get back to the previous version
             // We use the backpointerMap to get the previous entry version
             // If none found, throw NoRollbackException.
-            this.version = undoRecord.getEntry().getBackpointerMap().values().stream().
-                    findFirst().orElseThrow(NoRollbackException::new);
+            this.version = Optional.ofNullable(undoRecord.getEntry().getBackpointer(sv.getID()))
+                    .orElseThrow(NoRollbackException::new);
 
 
             // check if we rolled back to the requested version
--- a/test/src/test/java/org/corfudb/runtime/object/transactions/UndoTest.java
+++ b/test/src/test/java/org/corfudb/runtime/object/transactions/UndoTest.java
@@ -198,4 +198,62 @@ public class UndoTest extends AbstractObjectTest {
         });
 
     }
+
+    /**
+     * Check that rollbackUnsafe can rollback on context switch
+     *
+     * <p>
+     * Verify that if we have two threads with different snapshot we can
+     * Undo the log upon context switch.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void CanUndoDifferentVersion()
+        throws Exception {
+        Map<Integer, String> testMap = getRuntime()
+                .getObjectsView()
+                .build()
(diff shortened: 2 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -220,7 +220,11 @@ public void rollbackUnsafe(long version) {
                     .doUndo(object, undoRecord.getUndoRecord(),
                                     undoRecord.getSMRArguments());
 
-            this.version = undoRecord.getEntry().getGlobalAddress();
+            // After undoing the record, we should get back to the previous version
+            // We use the backpointerMap to get the previous entry version
+            // If none found, throw NoRollbackException.
+            this.version = undoRecord.getEntry().getBackpointerMap().values().stream().
```
no2chem: I think you want

getBackpointerMap().get(sv.getID())

Which will produce the correct backpointer

no2chem: or alternatively

undoRecord.getEntry().getBackpointer(sv.getID());

test/src/test/java/org/corfudb/runtime/object/transactions/UndoTest.java
```diff
@@ -198,4 +198,62 @@ public void canUndoAfterNoUndo()
...
+        final int mapSize = 10 * PARAMETERS.NUM_ITERATIONS_LARGE;
+
+        // populate the map with many elements
+        for (int i = 0; i < mapSize; i++)
+            testMap.put(i, normalValue);
+
+        // t0 is starting a transaction and reading the map
+        // t0 snapshot is at version x (object is at version x)
+        t(0, () -> getRuntime().getObjectsView().TXBegin());
+        t(0, () -> testMap.size());
```
dahliamalkhi: this seems like a vacuous statement?

test/src/test/java/org/corfudb/runtime/object/transactions/UndoTest.java
```diff
@@ -198,4 +198,62 @@ public void canUndoAfterNoUndo()
...
+
+        // External non-transactional put in the map
+        // After puts, Underlying object is at version x+2
+        testMap.put(mapSize + 1, "k");
+        testMap.put(mapSize + 2, "l");
+
+        // t1 starts a transaction and read the map
+        // t1 snapshot is at x+2 (object is at version x+2)
+        t(1, () -> getRuntime().getObjectsView().TXBegin());
+        t(1, () -> testMap.size());
```
dahliamalkhi: same?

test/src/test/java/org/corfudb/runtime/object/transactions/UndoTest.java
```diff
@@ -198,4 +198,62 @@ public void canUndoAfterNoUndo()
...
+        t(0, () -> testMap.size());
+        endTime = System.currentTimeMillis();
+
+        if (!testStatus.equals("")) {
+            testStatus += ";";
+        }
+        testStatus += "undo rebuild time=" +
+                String.format("%.0f", (float)(endTime-startTime))
+                + "ms";
+    }
```
dahliamalkhi: here (and also in the previous test, the one I added), we should probably assertThat about the short time, rather than just print it.

test/src/test/java/org/corfudb/runtime/object/transactions/UndoTest.java
rogermichoud: This will force a read to the map. Transactional snapshot is lazily initialized upon the first access and size() is a accessor. If we don't read it, we won't have different snapshot for each transaction.

test/src/test/java/org/corfudb/runtime/object/transactions/UndoTest.java
rogermichoud: Asserting on time seems like we would need to rely on a lot of assumption about the execution. This might end up in flaky test.

The best way would be to have a side effect if we have to reset the map.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
rogermichoud: Will update it. With my solution, I am relying on the fact that backpointermap has only one stream. Can one entry has multiple stream backpointers ?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **bd8758f**.

 [Pull Request #472 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/r4hNu0JR1RPRQdTOFQxtfgUk71U5HHBa)

dahliamalkhi:
This fix is now incorporated into #499 . I am closing this PR.

PR #473, opened 2017-03-08
Utility class - Multi read write lock

kspirov:
Next independent/incremental submission for the quorum replication, based on  https://github.com/CorfuDB/CorfuDB/pull/449 

This is a small request, while waiting decision how to proceed with the overwrite on the server.

This utility class is self explanatory, it allows multiple R/W locks identified by a long value.

One example when we will need it:  in quorum replication we need to read the result from phase 1 from the disk and then to take decision on writing based on the data,  we need this operation to be done only by a single thread, per log position. We don't want a global lock, this is very ineffective.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/MultiReadWriteLock.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/MultiReadWriteLock.java
@@ -7,6 +7,7 @@ import lombok.AllArgsConstructor;
 import lombok.Data;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
@@ -18,7 +19,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 public class MultiReadWriteLock {
 
     // all used locks
-    private HashMap<Long, ReentrantReadWriteLock> locks = new HashMap<>();
+    private ConcurrentHashMap<Long, LockReference> locks = new ConcurrentHashMap<>();
 
     // lock references per thread
     private final ThreadLocal<LinkedList<LockMetadata>> threadLockReferences = new ThreadLocal<>();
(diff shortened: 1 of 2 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **2e1d11f**.

 [Pull Request #473 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/99bPjilI6Zrq7dyen99FoSrzjHUuIZAT)

kspirov:
@dahliamalkhi  

- As I suggested in my review - just where the 2 phase message request is handled - you lock on the position, you read the rank, you take decision and eventually write, then unlock. In such case the things looks safe and straigforward. Actually here we need only write lock. 

- From the direction your comments I figure out there could be eventually another use  - for example in StreamLogFiles, where you do synchronized(lock),  with the lock which an object in FileHandle.  If you keep instead the segment position in the file handle instead of the lock object, you could really take advantage on read/write locks with this class, by having additionally assertion on the lock acquisition order.

1) About lock acquisition order - it can be done easy and I don't see any harm. I will suggest you something.

2) If really these locks can be used in StreamLogFiles too, the log should be on the segments, yes. If we are using them on individual 2 phase messages, I don't see a reason.

kspirov:
@dahliamalkhi  Implemented your idea for  lock acquisition order. Keeping of reference in the thread scope also allowed me to add some other checks - wrong order of removing of locks, entering with write lock in read lock. Also (just for perfection) made the close operation idempotent.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **353062e**.

 [Pull Request #473 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/2G3Ma0JZ9LuLgkQvUz3Uf4RLjwIh7YMh)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **de94010**.

 [Pull Request #473 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/QmiYPCi5iY1lloyNmj504xwnddd8Fyxt)

PR #474, opened 2017-03-08
Implement richer stream interface

no2chem:
This patch adds several interfaces to IStreamView:
seek() - moves the stream pointer
find() - given a global address, get the global address for next/prev entries
previous() - return the previous entry
current() - return the previously returned entry

It also patches versionLockedObject to use these interfaces and
eliminates the undoLog and associated tracking mechanisms.

PR #475, opened 2017-03-10
Fix undo log rollback

no2chem:
This patch extracts a performance fix from #474 which
fixes a performance regression.

The regression was caused by the elimination of the undoable
flag, which was necessary because some undo records could
be null (for example, in a map the previous value of the map
could have been not set). This would cause many 
unneeded rollbacks based on the content of the 
undo record.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
@@ -40,17 +40,26 @@ public class SMREntry extends LogEntry implements ISMRConsumable {
      *
      */
     @Getter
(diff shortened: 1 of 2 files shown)


no2chem:
@dahliamalkhi @medhavidhawan @rogermichoud If we could fast track this review because it's fixing a serious performance regression, I think that'd be best

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **9dab976**.

 [Pull Request #475 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/sjVXneRLP35nD88zGTDk7Hyq2hyw5IZ9)

PR #476, opened 2017-03-11
add logging on CorfuCompileProxy::logUpdate

dahliamalkhi:


runtime/src/main/java/org/corfudb/runtime/collections/SMRMap.java
```diff
@@ -21,6 +22,7 @@
 /**
  * Created by mwei on 1/7/16.
  */
+@Slf4j
```
no2chem: This seems to be added but unused

test/src/test/resources/logback-test.xml
```diff
@@ -6,7 +6,7 @@
         </encoder>
     </appender>
 
-    <logger name="org.corfudb.runtime.object" level="DEBUG"/>
+    <logger name="org.corfudb.runtime.object" level="TRACE"/>
```
no2chem: Yeah, let's not commit this into the repo

PR #477, opened 2017-03-11
Sync interface - update objects without an access

no2chem:
This patch adds a sync interface to objects view, which 
enables objects to be sync'd without calling an
accessor. Note that sync can also be called from
within a transaction, which will actually force a
sync (undoing optimistic updates if necessary).


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **414b965**.

 [Pull Request #477 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/x0dFGVUJv18TRKvCE9zwzsaVgXn7fb2U)

PR #478, opened 2017-03-11
patch tx commit

no2chem:
Previously, tx commit skipped an entry in the stream view,
which may not have been correct in some cases. This patch
selects the tryCommitAllProxies change from #474 to fix
transaction commit

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
@@ -325,7 +325,7 @@ public class OptimisticTransactionalContext extends AbstractTransactionalContext
         updateAllProxies(x -> {
             // If some other client updated this object, sync
             // it forward to grab those updates
-            if (getSnapshotTimestamp() !=
+            if (getSnapshotTimestamp() <
                     committedEntry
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -301,13 +304,64 @@ public long commitTransaction() throws TransactionAbortedException {
...
+                .getAddressSpaceView().read(commitAddress);
+
+        if (committedEntry.getType() == DataType.EMPTY) {
+            return;
+        }
+
+        updateAllProxies(x -> {
+            // If some other client updated this object, sync
+            // it forward to grab those updates
+            if (getSnapshotTimestamp() !=
```
dahliamalkhi: individual streams backpointers might not be equal to the global snapshot. Did you mean 
     getSnapshotTS() < backpointer
?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **27d1aab**.

 [Pull Request #478 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/4NAheoMksN9fGCbn0gZYuEwUNo4zt5Jx)

PR #480, opened 2017-03-11
dont save upcalls if they are not needed

no2chem:
logUpdate currently saves upcalls whether they are needed or not,
which is unnecessary in the case of mutators or mutatorAccessors
which return void.

This causes a very small amount of memory to leak, since the
result of an upcall that is not needed is probably null.

This patch fixes logUpdate by adding a keepUpcallResult parameter,
which controls whether or not an upcall is added to the pendingUpcalls
list or not.


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **8844d5a**.

 [Pull Request #480 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/vztEwpsMu1Hq1ss9l4c7dj5ga2xWvM9C)

PR #484, opened 2017-03-14
Avoid Scanning When Reading A Record

Maithem:
When a segment is opened its address space is read into memory
along with the file offsets that correspond to log entries.
As a consequence, record reads do not linearly scan the log
file anymore, instead the record is directly read from
the log file using its offset.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/AddressMetaData.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/AddressMetaData.java
@@ -1,0 +1,16 @@
+package org.corfudb.infrastructure.log;
+
+/**
+ * Created by maithem on 3/14/17.
+ */
+public class AddressMetaData {
+    public final int checksum;
+    public final int length;
+    public final long offset;
+
+    public AddressMetaData(int checksum, int length, long offset) {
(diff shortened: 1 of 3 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **0a7d038**.

 [Pull Request #484 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/dVIy69woF6a4YuddsBcozjb3R0wIYAAg)

kspirov:
This change is more correct then what I started to do. Especially the part with AddressMetadata is very good, also it good you now split the read method. I don't see any danger from this push, folks - can you merge it when you are ready, because I want to add my changes on the top of this.

@Maithem - I believe the compact() story is still not resolved here, are you OK if I take care for it?

Maithem:
@kspirov Currently, append doesn't allow duplicates (I left it out intentionally), so compact should have the same behavior as before. Once you add your overwrite logic the compact/trim methods should be modified appropriately. 

Please go ahead, I'll review your new pull request whenever you're ready.

kspirov:
Will do a pull request on the readOffset branch a bit later tonight.

PR #486, opened 2017-03-14
options to disable undo flag

no2chem:
This patch adds an option to disable undo and optimistic undo


no2chem:
Unless there's only a few trace statements, let's submit them in a different PR...

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -31,6 +32,18 @@
 @Accessors(chain = true)
 public class CorfuRuntime {
 
+    @Data
+    public static class CorfuRuntimeParameters {
```
medhavidhawan: add the holefilling into this as well. but do not hold back because of that.

PR #489, opened 2017-03-15
Ranked address space support.

kspirov:
* Rank is a separate class, with basic part and UUID.
* PROPOSAL DataType and write operations based on rank only when a rank is given (for quorum).
* Using the new read/write locks in StreamLogFiles (deadlock protection and multiple readers)
* Single address from the global space for the compact operation
* Bugfixes relatred to closing of  streams and channels (more to come)

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -14,22 +14,11 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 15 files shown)


kspirov:
This pull request is based on https://github.com/CorfuDB/CorfuDB/pull/484

kspirov:
I see travis fails, hold down the review, I will move to Linux VM.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -167,16 +160,13 @@ public void trim(LogAddress logAddress) {
...
             return;
         }
-
         TrimEntry entry = TrimEntry.newBuilder()
                 .setChecksum(getChecksum(logAddress.getAddress()))
                 .setAddress(logAddress.getAddress())
                 .build();
 
         // TODO(Maithem) possibly move this to SegmentHandle. Do we need to close and flush?
-        OutputStream outputStream = Channels.newOutputStream(handle.getPendingTrimChannel());
```
kspirov: @Maithem I believe this change is correct, but it causes your StreamLogFiles.trim() test to fail on the place where you check the half segment is still not trimmed. Let me see does your test work due to side effect from non-closed stream?

kspirov: Ok, ok, closing this will close the parent channel. But I am still not certain this stream should stay open.  Let me fix it on my side

kspirov:
@Maithem  OK, we are back in the game. Now you can review...

kspirov:
As @Maithem merged his change, closing this review and opening another rebased on the top of master https://github.com/CorfuDB/CorfuDB/pull/489

PR #490, opened 2017-03-15
Ranked address space support.

kspirov:
* Rank is a separate class, with basic part and UUID.
* New DataType PROPOSAL and write operations based on rank and data type only when a rank is given (for quorum).
* Using the new read/write locks in StreamLogFiles (deadlock protection and multiple readers)
* Single address from the global space for the compact operation
* Bugfixes relatred to closing of  streams and channels (more to come)

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -39,13 +39,13 @@ import org.corfudb.protocols.wireprotocol.TrimRequest;
 import org.corfudb.protocols.wireprotocol.WriteMode;
 import org.corfudb.protocols.wireprotocol.WriteRequest;
 import org.corfudb.runtime.exceptions.DataCorruptionException;
+import org.corfudb.runtime.exceptions.DataOutrankedException;
 import org.corfudb.runtime.exceptions.OverwriteException;
+import org.corfudb.runtime.exceptions.ValueAdoptedException;
 import org.corfudb.util.MetricsUtils;
 import org.corfudb.util.Utils;
-import org.corfudb.util.retry.IRetry;
 import org.corfudb.util.retry.IntervalAndSentinelRetry;
 
-import static org.corfudb.infrastructure.ServerContext.SMALL_INTERVAL;
 
 /**
  * Created by mwei on 12/10/15.
@@ -124,7 +124,7 @@ public class LogUnitServer extends AbstractServer {
             String logdir = opts.get("--log-path") + File.separator + "log";
             File dir = new File(logdir);
             if (!dir.exists()) {
-                dir.mkdir();
+                dir.mkdirs();
             }
             streamLog = new StreamLogFiles(logdir, (Boolean) opts.get("--no-verify"));
         }
@@ -167,10 +167,15 @@ public class LogUnitServer extends AbstractServer {
                 r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg());
             }
         } catch (OverwriteException ex) {
-            if (msg.getPayload().getWriteMode() != WriteMode.REPLEX_STREAM)
+            if (msg.getPayload().getWriteMode() != WriteMode.REPLEX_STREAM) {
                 r.sendResponse(ctx, msg, CorfuMsgType.ERROR_OVERWRITE.msg());
-            else
+            } else {
                 r.sendResponse(ctx, msg, CorfuMsgType.ERROR_REPLEX_OVERWRITE.msg());
+            }
+        } catch (DataOutrankedException e) {
+            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_OUTRANKED.msg());
+        } catch (ValueAdoptedException e) {
+            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_VALUE_ADOPTED.payloadMsg(e.getReadResponse()));
         }
     }
 
@@ -248,12 +253,18 @@ public class LogUnitServer extends AbstractServer {
     @ServerHandler(type = CorfuMsgType.FILL_HOLE, opTimer = metricsPrefix + "fill-hole")
     private void fillHole(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx, IServerRouter r,
                           boolean isMetricsEnabled) {
+        LogAddress l = new LogAddress(msg.getPayload().getPrefix(), msg.getPayload().getStream());
         try {
-            dataCache.put(new LogAddress(msg.getPayload().getPrefix(), msg.getPayload().getStream()), LogData.HOLE);
+            dataCache.put(l, LogData.HOLE);
             r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg());
 
         } catch (OverwriteException e) {
             r.sendResponse(ctx, msg, CorfuMsgType.ERROR_OVERWRITE.msg());
+        } catch (DataOutrankedException e) {
+            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_OUTRANKED.msg());
+        } catch (ValueAdoptedException e) {
+
+            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_VALUE_ADOPTED.payloadMsg(e.getReadResponse()));
         }
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -20,14 +20,13 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.hash.Hasher;
@@ -44,6 +43,7 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.commons.io.FileUtils;
+import org.corfudb.format.Types;
 import org.corfudb.format.Types.TrimEntry;
 import org.corfudb.format.Types.DataType;
 import org.corfudb.format.Types.LogEntry;
@@ -55,6 +55,8 @@ import org.corfudb.runtime.exceptions.DataCorruptionException;
 import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.exceptions.TrimmedException;
 
+import javax.annotation.Nullable;
+
 
 /**
  * This class implements the StreamLog by persisting the stream log as records in multiple files.
@@ -65,7 +67,7 @@ import org.corfudb.runtime.exceptions.TrimmedException;
  */
 
 @Slf4j
-public class StreamLogFiles implements StreamLog {
+public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpace {
 
     static public final short RECORD_DELIMITER = 0x4C45;
     static public int VERSION = 1;
@@ -81,6 +83,7 @@ public class StreamLogFiles implements StreamLog {
     public final String logDir;
     private Map<String, SegmentHandle> writeChannels;
     private Set<FileChannel> channelsToSync;
+    private MultiReadWriteLock segmentLocks = new MultiReadWriteLock();
 
     public StreamLogFiles(String logDir, boolean noVerify) {
         this.logDir = logDir;
@@ -130,14 +133,14 @@ public class StreamLogFiles implements StreamLog {
                         throw new RuntimeException(msg);
                     }
 
-                    if (noVerify == false && header.getVerifyChecksum() == false) {
+                    if (!noVerify && !header.getVerifyChecksum()) {
                         String msg = String.format("Log file {} not generated with checksums, can't verify!",
                                 file.getAbsoluteFile());
                         throw new RuntimeException(msg);
                     }
 
                 } catch (IOException e) {
-                    throw new RuntimeException(e.getMessage(), e.getCause());
+                    throw new RuntimeException(e.getMessage(), e);
                 }
             }
         }
@@ -167,7 +170,6 @@ public class StreamLogFiles implements StreamLog {
                 handle.getPendingTrims().contains(logAddress.getAddress())) {
             return;
         }
-
         TrimEntry entry = TrimEntry.newBuilder()
                 .setChecksum(getChecksum(logAddress.getAddress()))
                 .setAddress(logAddress.getAddress())
@@ -175,7 +177,6 @@ public class StreamLogFiles implements StreamLog {
 
         // TODO(Maithem) possibly move this to SegmentHandle. Do we need to close and flush?
         OutputStream outputStream = Channels.newOutputStream(handle.getPendingTrimChannel());
-
         try {
             entry.writeDelimitedTo(outputStream);
             outputStream.flush();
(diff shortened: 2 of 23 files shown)


kspirov:
I think this covers almost everything we need on the server side.

3 TODO for the server and we are done:

React the review comments (if any)

To figure out why the tests that open the StreamLogFiles don't work on my Windows notebook - this is usually a sign there is some unclosed read input stream. I noticed this long time ago, but as I, didn't pay too much attention as I am not using Windows often. But if there is a forgotten stream, we must profile and find it.

For compact() to discuss what will be the criteria the log file is for compacting. Should we read all entries (slow operation) or we should mark externaly the file has duplicate global addresses. Everything depends on how often compact() will be called. Currently the entries with the same global address will be compacted only if the compact is started because of trims() which I am not sure is correct.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **262d852**.

 [Pull Request #490 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/T0TNwxZv2al1VW8Y5GHUFaaYjL5TkIpo)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -725,7 +742,23 @@ public void append(LogAddress logAddress, LogData entry) {
             SegmentHandle fh = getSegmentHandleForAddress(logAddress);
             if (fh.getKnownAddresses().containsKey(logAddress.address) ||
                     fh.getTrimmedAddresses().contains(logAddress.address)) {
-                throw new OverwriteException();
+                if (entry.getRank()==null) {
+                    throw new OverwriteException();
+                } else {
+                    try (MultiReadWriteLock.AutoCloseableLock ignored = segmentLocks.acquireWriteLock(fh.getSegment())) {
```
Maithem: Since append is serially called, is this lock necessary? also not write locks are acquired in write record too.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogWithRankedAddressSpace.java
```diff
@@ -0,0 +1,37 @@
...
+        LogData oldEntry = read(logAddress);
+        if (oldEntry.getType()==DataType.EMPTY) {
+            return Optional.of(Boolean.TRUE);
+        }
+        if (newEntry.getType()==DataType.PROPOSAL && oldEntry.getType() != DataType.PROPOSAL) {
+            // the new data is a proposal, the other data is not, so the old value should be accepted
+            return Optional.of(Boolean.FALSE);
+        }
+        int compare = newEntry.getRank().compareTo(oldEntry.getRank());
+        return compare==0? Optional.empty(): compare<0? Optional.of(Boolean.FALSE): Optional.of(Boolean.TRUE);
```
Maithem: Why bother returning empty? if compare == 0, then just return false (i.e. can't append) since the value already exists.

So if you are trying to write the same entry you'll throw overwrite exception and when you try to write with a lower rank  you get a rank exception.

This code is coalescing two different cases, lower rank writes are being interpreted as overwrites and same entry overwrites are being ignored.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
```diff
@@ -1,19 +1,17 @@
 package org.corfudb.protocols.wireprotocol;
 
+import com.esotericsoftware.kryo.NotNull;
 import com.google.common.reflect.TypeToken;
-import lombok.Data;
-import lombok.Getter;
-import lombok.RequiredArgsConstructor;
+import lombok.*;
```
Maithem: nit: fix expand star imports

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -725,7 +742,23 @@ public void append(LogAddress logAddress, LogData entry) {
...
+                        Optional<Boolean> canAppend = isAppendPermitted(logAddress, entry);
+                        if (canAppend.isPresent()) {
+                            if (canAppend.get()) {
+                                AddressMetaData addressMetaData = writeRecord(fh, logAddress.address, entry);
+                                fh.getKnownAddresses().put(logAddress.address, addressMetaData);
+                            } else {
+                                throw new OverwriteException();
+                            }
+                        } else {
+                            log.debug("Duplicate requests for {} ", logAddress);
```
Maithem: Do you care about the case where you have the same ranks, but the data in the entry differs ? would checking that be a good way to find out if some bad happened?

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -209,5 +208,93 @@ public void testVerifyWithNoVerifyLog() throws Exception {
...
+        final long ADDRESS_1 = 100L;
+        //write at 0
+        ByteBuf b = ByteBufAllocator.DEFAULT.buffer();
+        Serializers.CORFU.serialize("0".getBytes(), b);
+        WriteRequest m = WriteRequest.builder()
+                .writeMode(WriteMode.NORMAL)
+                .data(new LogData(DataType.DATA, b))
+                .build();
+        m.setGlobalAddress(ADDRESS_0);
+        // m.setStreams(Collections.singleton(CorfuRuntime.getStreamID("a")));
```
Maithem: remove comment

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -725,7 +742,23 @@ public void append(LogAddress logAddress, LogData entry) {
...
-                throw new OverwriteException();
+                if (entry.getRank()==null) {
+                    throw new OverwriteException();
+                } else {
+                    try (MultiReadWriteLock.AutoCloseableLock ignored = segmentLocks.acquireWriteLock(fh.getSegment())) {
+                        Optional<Boolean> canAppend = isAppendPermitted(logAddress, entry);
+                        if (canAppend.isPresent()) {
+                            if (canAppend.get()) {
+                                AddressMetaData addressMetaData = writeRecord(fh, logAddress.address, entry);
+                                fh.getKnownAddresses().put(logAddress.address, addressMetaData);
```
Maithem: After the put, you can then trim the address of the old entry.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogWithRankedAddressSpace.java
kspirov: Just a habit to write idempotent code. But in the quorum case we don't need it, nobody will retry the message, OK.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
kspirov: See below.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
kspirov: agree

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -14,22 +14,11 @@
...
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
kspirov: ... and setup that Intellij, OK :)

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -124,7 +124,7 @@ public LogUnitServer(ServerContext serverContext) {
             String logdir = opts.get("--log-path") + File.separator + "log";
             File dir = new File(logdir);
             if (!dir.exists()) {
-                dir.mkdir();
+                dir.mkdirs();
```
kspirov: intentionally :)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
kspirov: @Maithem  you cannot, as the log address doesn't change!  But my code makes sure the trim() will still work, just we need to see how to change criteria when to be activated.

The trick is that now the code which prepares the entries to reconstruct after trimming is linked hash map by exactly this log address, not array list. So the duplicate will disappear naturally, once trim is activated.

kspirov:
Done. Going to the bed again, tomorrow I will reply your email.  :)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **e24af34**.

 [Pull Request #490 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/tZ26rhoa36Y43xMa8I5wmIa5RK0sH0f0)

kspirov:
@Maithem Ops, I need to limit the commit message line to 72 chars for my Ubuntu env too.

kspirov:
I think we are OK with this part, now.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **b344770**.

 [Pull Request #490 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/uNDS39H1YcF5TpM6itxJPg9o37JW2VsJ)

kspirov:
Returning back to "in progress", some changes in SteamLogWithRankedAddressSpace will be necessary

kspirov:
The server should be able to return back the value during phase1. Working on this, ETA today

kspirov:
I made the changes discussed with Dalia, now the client can adopt values. I believe this is the last change that concerns the server side.

In the last commit I have also some code in the LogUnitClient, but it is minimal and its purpose is just to test whether we are able to send and receive the new packets end to end.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **85f1cba**.

 [Pull Request #490 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/P7kQIJ1upr97vnn6jIjlp1DcET5VceGk)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogWithRankedAddressSpace.java
```diff
@@ -0,0 +1,49 @@
...
+        LogData oldEntry = read(logAddress);
+        if (oldEntry.getType()==DataType.EMPTY) {
+             return;
+        }
+        int compare = newEntry.getRank().compareTo(oldEntry.getRank());
+        if (compare<0) {
+            throw new DataOutrankedException();
+        }
+        if (compare>0) {
+            if (newEntry.getType()==DataType.PROPOSAL && oldEntry.getType() != DataType.PROPOSAL) {
```
kspirov: Currently PROPOSAL means just persisted entry after phase 1 - with unspecified value yet, but with reserved rank. I can change this if it confuses the reader.

So on phase 2 we would send DATA. I remember you discussed there should be assertion, that phase 2 should not be sent without phase 1, so we could change the naming. For example:

PHASE1_PROPOSAL and PHASE2_PROPOSAL?

Maithem:
Can you create a new pull request for your latest commit? Everything (clean up, stream fix, R/W lock) up till "Merge branch 'master' into server-quorum-stuff" seems to be good to be checked in. This pull request is getting large and its hard to review. Ignore, this if you're not adding more code to this pull request.

kspirov:
@Maithem  - if Dahlia is OK, of course you can merge the previous changes, it won't do harm. I can extract the last as a separate request, NP.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogWithRankedAddressSpace.java
dahliamalkhi: got it.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **e1f0415**.

 [Pull Request #490 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ITy3ZMMMT7mWx0wfMpzZg1cUqWcmwpDY)

kspirov:
@dahliamalkhi , @Maithem  - sorry, but I finally rebased everything on the top of the current master. I believe we won't need any general changes not only on replica side, but in the communication messages too. 

@Maithem - it is almost the same to the version you reviewed, but we have a new exception, that can return the data to the client, and the method in StreamLogFilesWithRankedAddressSpace is changed again - it directly throws the exception. As a side effect, the behavior when you receive the same rank is again idempotent (you remember you wanted a different behavior), but I don't think it will be a big harm. Otherwise this method would need to return 3-rd type of exception, only for this case.

@dahliamalkhi  - I changed the data type from PROPOSAL to RANK_ONLY, so it is more understandable and fixed the typo you showed.

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -209,5 +208,90 @@ public void testVerifyWithNoVerifyLog() throws Exception {
...
+                .data(new LogData(DataType.DATA, b))
+                .build();
+        m.setGlobalAddress(ADDRESS_0);
+        m.setStreams(Collections.EMPTY_SET);
+        m.setBackpointerMap(Collections.emptyMap());
+
+
+        WriteRequest m2 = WriteRequest.builder()
+                .writeMode(WriteMode.NORMAL)
+                .data(new LogData(DataType.DATA, b))
```
kspirov: Just reminder - we decided, that this code is correct, but I renamed the data type from PROPOSAL to RANK_ONLY for better readability.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogWithRankedAddressSpace.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Check whether the data can be appended to a given log address
+     * @param logAddress
+     * @param newEntry
+     * @throws DataOutrankedException if the log entry cannot be assigned to this log address as there is a data with higher rank
+     * @throws ValueAdoptedException if the new message is a proposal during the two phase recovery write and there is an existing
+     * data at this log address already.
+     */
+    default void assertAppendPermitted(LogAddress logAddress, LogData newEntry) throws DataOutrankedException, ValueAdoptedException {
```
dahliamalkhi: I believe that this method must be invoked with a lock on the logAddress' segment. If so, then to make sure no one invokes it without a lock in the future, please (a) indicate this in the comment, and (b) use a naming convention such as "unsafe" suffix or the like.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -192,6 +211,29 @@ private static Object handleReadDataCorruption(CorfuPayloadMsg<ReadResponse> msg
...
     /**
+     * Asynchronously write a proposal to a logging unit with ranked address space. This proposal data keeps
+     * the rank to the log unit during the two phase recovery write in the quorum replication.
+     *
+     * @param address        The address to write to.
+     * @param streams        The streams, if any, that this write belongs to.
+     * @param rank           The rank of this write]
+     *
+     */
+    CompletableFuture<Boolean> writeProposal(long address, Set<UUID> streams, IMetadata.DataRank rank) {
```
dahliamalkhi: I hate to bug you again about terminology, but I think that 'proposal' may be a confusing term. It is sometimes used for the second phase of quorum-consensus.

could we change to 'writePrepare' ?

kspirov: In my last client version I changed the signature of this method too and it is now  public CompletableFuture<Boolean> writeEmptyData(long address, DataType type, Set<UUID> streams, IMetadata.DataRank rank)  - we also supply DataType. I will port this and update in the reveiw now (I am in the airport)

kspirov:
@dahliamalkhi  Done, please wait Travis, I might have forgotten something

PR #493, opened 2017-03-17
Generate Code for IConflictObjectsExtractor

Maithem:
Augmented the annotation processor to generate methods that can
extract conflict parameters from an SMREntry.

The idea behind this pull request is to generate conflict objects getter methods for smr objects. For every method signature that contains a a conflict parameter annotation we will generate a getter method that can extract the conflict objects from the arguments.

For example, we the SMRMap has a put(K, V) method where the K is a conflict object, the annotion processor will generate the following method:

  public Object[] get_GETSMRCONFLICTOBJECTS(Object key) {
    Object[] conflictField_CORFUSMR = new Object[]{key};
    return conflictField_CORFUSMR;
  }

and 

  @Override
  public Object[] getConflictObject(String smrMethod, Object[] smrArguments) throws Exception {
    Method m = this.getClass().getMethod(ReflectionUtils.getMethodNameOnlyFromString(smrMethod) + "_GETSMRCONFLICTOBJECTS",
                                            ReflectionUtils.getArgumentTypesFromString(smrMethod));
    Object ret = m.invoke(this, smrArguments);
    return (Object[])ret;
  }

So, when an SMREntry is unpacked, one can obtain the conflict objects for an SMRObject by calling
getConflictObject. This functionality is needed for garbage collection on SMR objects. For instance,
one can write code that iterates through a stream and unpack the smr objects one by one while building a
conflict objects map. This conflict objects map can be used to determine weather a stream entry is stale or not and can be trimmed as a result.

### DIFF
--- a/annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
+++ b/annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
@@ -2,12 +2,34 @@ package org.corfudb.annotations;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.squareup.javapoet.*;
-import org.corfudb.runtime.object.*;
+import com.squareup.javapoet.AnnotationSpec;
+import com.squareup.javapoet.ClassName;
+import com.squareup.javapoet.FieldSpec;
+import com.squareup.javapoet.JavaFile;
+import com.squareup.javapoet.MethodSpec;
+import com.squareup.javapoet.ParameterSpec;
+import com.squareup.javapoet.ParameterizedTypeName;
+import com.squareup.javapoet.TypeName;
+import com.squareup.javapoet.TypeSpec;
+import com.squareup.javapoet.TypeVariableName;
+import org.corfudb.runtime.object.ICorfuSMR;
+import org.corfudb.runtime.object.ICorfuSMRProxy;
+import org.corfudb.runtime.object.ICorfuSMRUpcallTarget;
+import org.corfudb.runtime.object.IUndoFunction;
+import org.corfudb.runtime.object.IUndoRecordFunction;
 
-import javax.annotation.processing.*;
+import javax.annotation.processing.AbstractProcessor;
+import javax.annotation.processing.Filer;
+import javax.annotation.processing.Messager;
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.annotation.processing.RoundEnvironment;
+import javax.annotation.processing.SupportedAnnotationTypes;
 import javax.lang.model.SourceVersion;
-import javax.lang.model.element.*;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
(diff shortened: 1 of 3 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d95c99f**.

 [Pull Request #493 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/trZJ6YYRruxEVI0KpLpasUIC2YasNpb6)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d25220b**.

 [Pull Request #493 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/22rukMSuYUQGISE607SGYrgdQFmXaJZC)

annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
```diff
@@ -562,6 +593,47 @@ private void addUpcallMap(TypeSpec.Builder typeSpecBuilder, TypeName originalNam
                 .build());
     }
 
+    private void addConflictExtractorMap(TypeSpec.Builder typeSpecBuilder, TypeName originalName,
+                                         Set<SMRMethodInfo> methodSet) {
+        // Generate the upcall string and associated map.
+        String conflictExtractorString = methodSet.stream()
+                .filter(x -> x.hasConflictAnnotations)
+                .map(x -> "\n.put(\"" + getSMRFunctionName(x.method) + CONFLICT_METHODS_SUFFIX + "\", " +
```
no2chem: So, looking over this again, in #508 I introduce a ConflictFn, which if not specified should be any parameter annotated with @ConflictParameter. Shouldn't you just generate the lambda directly in the map, rather than generating an additional function? So - if it's a conflictFn, add a call to the function, and if not, just return the correct fields. The function should be of type IConflictFunction, with a signature of Object[] getConflictObjects(Object[] params).

annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
```diff
@@ -562,6 +593,47 @@ private void addUpcallMap(TypeSpec.Builder typeSpecBuilder, TypeName originalNam
...
+                                        "(" + x.method.getParameters().get(i).asType().toString() + ")" +
+                                                " args[" + i + "]")
+                                .collect(Collectors.joining(", "))
+                        + ");" + (x.method.getReturnType().getKind().equals(TypeKind.VOID)
+                        ? "return null;" : "") + "})")
+                .collect(Collectors.joining());
+
+        FieldSpec conflictExtractorMap = FieldSpec.builder(ParameterizedTypeName.get(ClassName.get(Map.class),
+                ClassName.get(String.class),
+                ParameterizedTypeName.get(ClassName.get(ICorfuSMRUpcallTarget.class),
```
no2chem: Also this should be IConflictFunction.class, right?

Maithem: You are right. Didn't know about that interface.

annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
Maithem: Let me see if I understood you correctly. You're saying instead of generating ConflictFn for all signatures that have a conflict parameter,  generate and use a single generic method than can extract the conflict parameters given (method signature, Object[]) for signatures that have conflictParameterFunction.equals(""), is this correct?

no2chem:
This PR is closed for the time being until the conflict extractor is required

PR #502, opened 2017-03-19
added a few more details to object logging messages

dahliamalkhi:
I added logging messages on all corfu-object operations.

On every mutator call, you get 
    - the method-name, 
    - its arguments, 
    - the conflict-parameters, 
    - if it is invoked within a transactional context, a TX-mark and the TX-id.

On every accessor call, you cannot get the method-name (because it is a lambda built on-the-fly), but you get as a hint the conflict-parameters. so you get
    - the conflict-parameters, 
    - if it is invoked within a transactional context, a TX-mark and the TX-id.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/TransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/TransactionalContext.java
@@ -1,6 +1,7 @@
 package org.corfudb.runtime.object.transactions;
 
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.util.Utils;
 
 import java.time.Duration;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/object/transactions/TransactionalContext.java
```diff
@@ -84,6 +85,7 @@ public static boolean isInTransaction() {
      * @return          The context which was added to the transaction stack.
      */
     public static AbstractTransactionalContext newContext(AbstractTransactionalContext context) {
+        log.trace("TX begin[{}]", Utils.toReadableID(context.transactionID));
```
no2chem: should just be log.trace("TX...", context);, as we should rely on the toString override for the context

PR #506, opened 2017-03-20
added a test of map 'set' operations

dahliamalkhi:
Reminder: All three of the methods of SMRMap ,

	SMRMap::keyset
	SMRMap::valueSet
	SMRMap::entitySet

currently return -immutable copies- of the underlying map. As in Java, taking
the key/value/entry-set over a map which is concurrently being manipulated yield
undefined results. Any manipulation to the map –after- the set was snapshot does
not affect it. And operations like add/remove are not supported (it is
immutable).

This PR also fixes MultipleOverlappingConcurrent, to address note by @slfritchie

### DIFF
--- a/test/src/test/java/org/corfudb/runtime/collections/SMRMapEntrySetTest.java
+++ b/test/src/test/java/org/corfudb/runtime/collections/SMRMapEntrySetTest.java
@@ -1,0 +1,162 @@
+package org.corfudb.runtime.collections;
+
+import org.corfudb.runtime.collections.SMRMap;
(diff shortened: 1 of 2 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **94fbc9c**.

 [Pull Request #506 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/XPV83SNC6fuxyHqeMO10b2V5mpr7XXWB)

PR #510, opened 2017-03-22
Rename TXConflictScenarios & TXsFromTwoRuntimes with 'Test' suffix

slfritchie:
These two classes contain @Test annotations that aren't being run by typical unit test executions, including TravisCI's runs.  Renaming with a `Test` suffix should get them run regularly again.


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **836015b**.

 [Pull Request #510 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/02NL9RBKpZtR01z6FROJoilQSQndQzbI)

slfritchie:
Heh, it's Maven that's doing that.  It doesn't run tests without `Test` suffix when using the `test` or `install` targets, but will run tests in any (?) class if you add the `-Dtest=classname` flag.

PR #512, opened 2017-03-23
revalidate lock on exception

no2chem:
Currently, access() in versionLockedObject performs an optimistic read.
Because the read is optimistic, sometimes the object could be concurrently
modified, causing an exception to be thrown. 

Previously we did not revalidate the lock on an exception, causing the client
to see the thrown exception. This patch modifies the VersionLockedObject to
revalidate the lock when an exception is thrown. If the read was valid,
the exception is re-thrown, otherwise, we upgrade to a write lock since
the exception was thrown due to an invalid view of then object.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -146,9 +146,21 @@ public class VersionLockedObject<T> {
         if (ts != 0) {
             if (directAccessCheckFunction.apply(this)) {
                 log.trace("Access [{}] Direct (optimistic-read) access at {}", this, getVersionUnsafe());
-                R ret = accessFunction.apply(object);
-                if (lock.validate(ts)) {
-                    return ret;
+                try {
+                    R ret = accessFunction.apply(object);
+                    if (lock.validate(ts)) {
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -146,9 +146,21 @@ public VersionLockedObject(Supplier<T> newObjectFn,
         if (ts != 0) {
             if (directAccessCheckFunction.apply(this)) {
                 log.trace("Access [{}] Direct (optimistic-read) access at {}", this, getVersionUnsafe());
-                R ret = accessFunction.apply(object);
-                if (lock.validate(ts)) {
-                    return ret;
+                try {
```
medhavidhawan: I am not sure if we have seen a lot of performance benefits with optimistic locking here. Have we measured this against just simply taking a read lock ? The reason I ask is the subtle bugs and the code readability

no2chem: Well, I could add a flag to switch between optimistic and full blown read locks and we could benchmark. When there's no contention, the optimistic read is definitely much faster than the full read lock, so it's definitely very workload dependent.

PR #513, opened 2017-03-24
Fix for consecutive nested transactions

rogermichoud:
This fix ensures that upon consecutive nested transactions, latest transactions can see optimistic updates from previous ones.

This commit introduces various fixes:

1. When accessing/modifying an object, we need to verify that its WriteSetSMRStream correctly reflects the current transactions
   that modified this object. WriteSetSMRStream contains a list of transaction context that have modified this stream.

   Upon access/modification of an object, we would check if the object's WriteSetSMRStream was the correct one by
   comparing if the root transaction context of the WriteSetSMRStream was the same than the root transaction
   context of the thread.

   In the case of consecutive nested transactions, when entering subsequent nested tx, the root tx would still
   be the same, but the WriteSetSMRStream of the object would reflect the previous nested tx. By checking if the
   root tx is the same, we would miss this state change.

   In our fix, we test that the thread's current transaction, is the current tx in the WriteSetSMRStream of the
   underlying object.

2. In order to simplify the sharing of write set between nested transactions, we made the nested tx inherit the
   write set of its parent. This allows the child tx to operate on the current WriteSetSMRStream without further
   optimistic syncing.

   When merging the nested tx in the parent one, we add new write set entries (corresponding to new streams) into
   the parent write set. Since both write set share reference of already present entries, these get updated automatically.

3. When setting a new WriteSetSMRStream for an object, we were not setting its current context to the current tx context
   of the thread. By default, it was being set to the root tx context.

   We ensured that the correct tx context was set.

We also added a test to validate correct execution of consecutive nested transaction

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
@@ -276,23 +276,14 @@ public abstract class AbstractTransactionalContext implements
 
     void mergeWriteSetInto(Map<UUID, AbstractMap.SimpleEntry<Set<Integer>, List<SMREntry>>> otherWSet) {
         otherWSet.entrySet().forEach(e-> {
-            // create an entry for this streamID
-            writeSet.computeIfAbsent(e.getKey(),                    // the streamID
-                    u -> new AbstractMap.SimpleEntry<>(new HashSet<>(), new ArrayList<>() ));
 
-            // copy all the conflict-params set for this streamID
-            writeSet.get(e.getKey())                 // the entry pair for this streamID
-                    .getKey()                        // the left componentof the pair is a conflict-param set
-                    .addAll(e.getValue()             // the pair from the otherWSet
-                            .getKey());              // the left component of the pair
-
-            // copy all the WriteSetEntry list for this streamID
-            writeSet.get(e.getKey())                 // the entry pair for this streamID
-                    .getValue()                      // the right componentof the pair is a WriteSetEntry list
-                    .addAll(e.getValue()             // the pair from the otherWSet
-                            .getValue());            // the right component of the pair
-
-
+            // Since nested transactions inherit the write set of its parent, in order to merge the
+            // write set of the child transaction, we only need to add new entries for non-existing
+            // streams.
+            // For entries that were inherited, they share the same reference in the write set map. Any
+            // modification of it will also appear in the parent write set.
+            writeSet.computeIfAbsent(e.getKey(),
+                    k -> e.getValue());
         });
     }
 
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
@@ -55,6 +55,12 @@ public class OptimisticTransactionalContext extends AbstractTransactionalContext
 
     OptimisticTransactionalContext(TransactionBuilder builder) {
         super(builder);
+
+        // When starting a nested transaction, the child transaction inherits the write set of its parent.
+        // This will ensure that any access to the objects will reflect parent optimistic updates.
+        if (getParentContext() != null) {
+            writeSet.putAll(getParentContext().writeSet);
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **8700863**.

 [Pull Request #513 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/awqXLywgzw46dEwjU2jn60OVxOgog0k1)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5576a2a**.

 [Pull Request #513 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/yGTyDkN7GnJwtNlPs8HOQvJ4VTw4zv1M)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5576a2a**.

*** 9.0% transaction FAILURE rate for 5 threads, CorfuMemoryNode mode
*** 11.0% transaction FAILURE rate for 10 threads, CorfuMemoryNode mode
*** 6.0% transaction FAILURE rate for 5 threads, CorfuDiskNode mode
*** 11.0% transaction FAILURE rate for 10 threads, CorfuDiskNode mode

 [Pull Request #513 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Vfp1DwR1zKzDWefhS0u52yWsLeh3sceZ)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **cd94253**.

*** 0.0% transaction FAILURE rate for 5 threads, CorfuDiskNode mode
*** 0.07% transaction FAILURE rate for 10 threads, CorfuDiskNode mode

 [Pull Request #513 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/t25JrO5P4IPrMPN1Fn3bYFd7g4pCWVvP)

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -276,23 +276,14 @@ void addToWriteSet(ICorfuSMRProxy proxy, SMREntry updateEntry, Object[] conflict
...
-                    .addAll(e.getValue()             // the pair from the otherWSet
-                            .getValue());            // the right component of the pair
-
 
+            // Since nested transactions inherit the write set of its parent, in order to merge the
+            // write set of the child transaction, we only need to add new entries for non-existing
+            // streams.
+            // For entries that were inherited, they share the same reference in the write set map. Any
+            // modification of it will also appear in the parent write set.
+            writeSet.computeIfAbsent(e.getKey(),
```
no2chem: Hm, I wonder if this is a correctness issue? What if the nested transaction aborts? Does it's write set get removed?

rogermichoud: I don't think that a nested transaction could be aborted. The decision to abort a transaction comes from the sequencer, and the commit goes there only if we merged all the nested transactions (only from the root transactional context).
Potential caveat is if the application developer put a catch (specific exception that is not a TransactionAbortedException) outside a nested transaction and decide to keep going with it's transaction even if the nested one failed. 
Not sure if we need to tackle this case. The other way would be to do deep copies of the write set on nested transaction. Might hit our performance pretty bad if we have a lot of nested transaction (which is our case).

Maithem:
@no2chem @rogermichoud 
Since we don;t have transaction propagation modes why don't we just use the current transaction when tx.begin is invoked , so if you have tx1.begin and tx2.begin on the same thread, then tx2.begin would use the same transactional context as tx1. That why you dont have to worry about merging and un-merging transactional contexts

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **cd94253**.

*** 0.03% transaction FAILURE rate for 5 threads, CorfuDiskNode mode
*** 0.04% transaction FAILURE rate for 10 threads, CorfuDiskNode mode

 [Pull Request #513 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/DIpkt73tK1LWeWaEAzusFHnvmUXgGRsS)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **cd94253**.

*** 0.01% transaction FAILURE rate for 5 threads, CorfuDiskNode mode
*** 0.15% transaction FAILURE rate for 10 threads, CorfuDiskNode mode

 [Pull Request #513 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/A7HQ0YW4GxiPB8xJ8n7A86WKsicbvceB)

rogermichoud:
We could just have all nested transaction having a reference to the read set and the write set. Like this, as you point, no merging happens upon commit. Played with a fix like this and it yields similar results.
Only concern is if we want to leave the API for merging transactions in case we want to implement different propagation modes in the future.

no2chem:
Well, you need to be able to abort a nested transaction, so that's the reason for having multiple contexts.



On Tue, Mar 28, 2017 at 5:59pm, Maithem < notifications@github.com [notifications@github.com] > wrote:
@no2chem [https://github.com/no2chem] @rogermichoud [https://github.com/rogermichoud]
Since we don;t have transaction propagation modes why don't we just use the current transaction when tx.begin is invoked , so if you have tx1.begin and tx2.begin on the same thread, then tx2.begin would use the same transactional context as tx1. That why you dont have to worry about merging and un-merging transactional contexts

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
no2chem: I don't know, a nested transaction can abort if it calls TXAbort() at any time. Only the outer most transaction should abort in this case, I think.

Issue #509, opened 2017-03-21
Incorrect Code Generation For Overloaded SMR Methods


Maithem:
When overloading methods the annotation processor will generate incorrect code. For example, in the case of an smr map, if you overload put and have something like v1 put(k1, v1, v2)
then the generated class would have an upcall map that will omit one of the signatures which will cause a runtime exception. The upcall map would look something like:

public final Map<String, ICorfuSMRUpcallTarget<SMRMap<K, V>>> upcallMap_CORFUSMR = new ImmutableMap.Builder<String, ICorfuSMRUpcallTarget<SMRMap<K, V>>>()
  .put("remove", (obj, args) -> { return obj.remove((java.lang.Object) args[0]);})
  .put("putAll", (obj, args) -> { obj.putAll((java.util.Map<? extends K,? extends V>) args[0]);return null;})
  .put("clear", (obj, args) -> { obj.clear();return null;})
  .put("put", (obj, args) -> { return obj.put((K) args[0], (V) args[1]);})
.put("put", (obj, args) -> { return obj.put((K) args[0], (V) args[1], (V) args[2]);}).build();

A possible fix would be to have the whole signature as the key (instead of the name).
What do you think @no2chem ?

no2chem:
Yeah, this is a bug. When name is not set in the annotation, the full signature should be used. We used to call `ReflectionUtils.getShortMethodName(method.toString())` in the old codebase, maybe we can do that again?

Maithem:
Yes, that will implicitly capture the number of parameters which is what is needed. 

You also bring up an important point, the annotation method naming can break java's typing system. Not javac, but our annotation processor. For instance, one method name can map to two methods with different return types. Can we just remove the method annotation naming? it seems error-prone.

no2chem:
No, I think we need to keep the method annotation naming. Otherwise method names could get quite long. Maybe they should automatically end with the number of parameters automatically though (put becomes put2), but I'm not sure that's a good idea. Maybe it'd be sufficient for the processor to check for duplicate up all targets and error. The reason we need the name is for methods like blindPut, which take a different name, and backwards compatibility.

PR #514, opened 2017-03-24
Properly Resolve Overloaded Methods

Maithem:
SMRObjects can overload methods via annotations. For example,
methods that are annotated will be persisted with the annotation
name instead of the actual method name. In certain cases, this
allows some types to escape the java typing system and thus
method collisions that could have been detected during compile
time would surface at runtime.

Since annotation overloading can only affect the method name,
only two cases need to be verified during processing. In order
to detect if there is a collision, we need to check if two
methods have the same signature, but different return type and
two methods with the same signature and same return type.


Maithem:
This partially addresses #509

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **29c303e**.

 [Pull Request #514 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/EWxOKTNyhoCvKsFr8E1aBxDAAaQcX0Lm)

PR #523, opened 2017-04-04
Proper synchronization of optimistic updates in nested transactions.

rogermichoud:
The issue was that we were using the same validation function for 2 different conditions.

When we test if the optimisticStream of the underlying object is the correct one for our current transaction, we need to check that current context of the optimisticStream is the same as the current context of the thread.

When we test if the thread owns the optimisticStream of the underlying object, we need to check if the root context of the optimisticStream is the same as the root context of the current thread. 

Usually, on context switch (or on starting nested transaction), we will have the optimisticStream at it's root context. (We unrolled optimistic updates pertaining to the thread we are coming from). This is expected. The current thread, though, is in the nested transaction context. In that case, we need to use the function that test for the root condition.

Then, we proceed on synchronizing the object by applying optimistic updates of nested transactions starting from the root. At the end of the process, the underlying object is in the state of the top of nested transactions stack.

This fix also revert changes of mergeWriteSetInto and remove write set inheritance in nested transactions.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -324,7 +324,7 @@ public class VersionLockedObject<T> {
      *          False otherwise.
      */
     public boolean optimisticallyOwnedByThreadUnsafe() {
-        return optimisticStream == null ? false : optimisticStream.isStreamForThisTransaction();
+        return optimisticStream == null ? false : optimisticStream.isStreamForThisThread();
     }
 
     /** Set the optimistic stream for this thread, rolling back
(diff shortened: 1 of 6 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **50cda3e**.

*** 0.04% transaction FAILURE rate for 5 threads, CorfuDiskNode mode
*** 0.13% transaction FAILURE rate for 10 threads, CorfuDiskNode mode

 [Pull Request #523 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/zHyFhhu7SWfia7b30ZsTgqPD1vL0caSy)

Issue #522, opened 2017-03-31
BackpointerStreamView fillReadQueue hole fill leaves empty entry


kjames88:
With two threads operating on a single map, BackpointerStreamView::fillReadQueue encounters a null/EMPTY element for an address.  It attempts to hole-fill and gets an OverwriteException.  currentEntry is updated to the local state, which is null/EMPTY, and processing continues.  Because the entry is not present, containsStream fails and nothing is added to readQueue.  When the affected thread later tries to sync to the skipped address, AbstractQueuedStreamView::getNextEntries returns nothing due to (context.readQueue.first() > maxGlobal).

Use of cache for read in fillReadQueue allows the client to be unaware of changes on the server and hole filling logic allows the stream view to advance without filling gaps in the history.  This seems to be a correctness problem.

*BackpointerStreamView::fillReadQueue*
`currentEntry = runtime.getAddressSpaceView().read(currentRead);`

*AddressSpaceView::read()*
`ILogData data = readCache.get(address);
            if (data == null) {
                data = new InMemoryLogData(DataType.EMPTY);
                data.setGlobalAddress(address);
}`

The failure was observed with a non-transaction 2 thread test writing to a single map.  In this case, CorfuCompileProxy::getUpcallResultInner throws RuntimeException because the entry, while received into the cache, was never propagated into the stream view, and therefore never triggered setUpcallResult.

medhavidhawan:
/**
     * Reset all in-memory caches.
     */
    public void resetCaches() {
        readCache = Caffeine.<Long, ILogData>newBuilder()
                .<Long, ILogData>weigher((k, v) -> v.getSizeEstimate())
                .maximumWeight(runtime.getMaxCacheSize())
                .recordStats()
                .build(new CacheLoader<Long, ILogData>() {
                    @Override
                    public ILogData load(Long aLong) throws Exception {
                        return cacheFetch(aLong);
                    }

                    @Override
                    public Map<Long, ILogData>
                    loadAll(Iterable<? extends Long> keys) throws Exception {
                        return cacheFetch((Iterable<Long>) keys);
                    }
                });
    }

If you look at the readCache initialization code, it does a cacheFetch() inside the load method which is used if the data is not in the cache.  So the get will either load the data into the cache and then get it or it will directly get it from the cache if it is present.

Maithem:
Addressed by #522

Issue #439, opened 2017-02-09
Hole Filling Bug


Maithem:
In addition to writing holes, the hole filling mechanism should also propagate values that have been 
partially written in the chain. For example, If a client writes to the head then it fails, other clients would
try to fill in the hole for that address, but when they try to write to the head an overwrite exception
is thrown. Instead, the clients should take the value for that address from the head and propagate it to the tail.

dahliamalkhi:
I think this can be closed already? @Maithem

slfritchie:
Mid-June issues review: ping @Maithem, is @kspirov's new & refactored code in #527 and subsequent runtime PRs sufficient?

no2chem:
Closed by #524

PR #524, opened 2017-04-06
Reads block for data to commit.

medhavidhawan:
Temporary fix for #522.
In BackpointerStreamView.fillReadQueue reads block for writer or hole
fill to commit. This prevents the BackpointerStreamView to create EMPTY
entries for reads from uncommitted entries.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
@@ -1,7 +1,10 @@
 package org.corfudb.runtime.view.stream;
 
 import lombok.extern.slf4j.Slf4j;
-import org.corfudb.protocols.wireprotocol.*;
+import org.corfudb.protocols.wireprotocol.DataType;
+import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.protocols.wireprotocol.LogData;
+import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.view.Address;
@@ -254,10 +257,12 @@ public class BackpointerStreamView extends AbstractQueuedStreamView {
                         }
                     }
                 }
-
-                // If we STILL don't have the data, now we need to do a hole
-                // fill.
-                if (currentEntry.getType() == DataType.EMPTY) {
+                //FIXME To be replaced by cleaner interface. This is just a patch till the real fix comes in.
+                // If we STILL don't have the data, it means the writer is either dead or too slow.
+                // We will now try to make sure we have committed data for the currentEntry. We will first
+                // try a HOLE FILL, failing which we will re-read the entry to see if the writer completed.
+                // We will try this in a loop till we get committed data for the currentEntry.
+                while (currentEntry.getType() == DataType.EMPTY) {
                     try {
                         runtime.getAddressSpaceView().fillHole(currentRead);
                         // If we reached here, our hole fill was successful.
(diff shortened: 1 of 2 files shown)


test/src/test/java/org/corfudb/runtime/collections/SMRMultiLogunitTest.java
```diff
@@ -0,0 +1,216 @@
+package org.corfudb.runtime.collections;
```
Maithem: Clean Up test code: variable names, dont print to system.out, remove commented code,

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -254,10 +257,12 @@ protected boolean fillReadQueue(final long maxGlobal,
...
-
-                // If we STILL don't have the data, now we need to do a hole
-                // fill.
-                if (currentEntry.getType() == DataType.EMPTY) {
+                //FIXME To be replaced by cleaner interface. This is just a patch till the real fix comes in.
+                // If we STILL don't have the data, it means the writer is either dead or too slow.
+                // We will now try to make sure we have committed data for the currentEntry. We will first
+                // try a HOLE FILL, failing which we will re-read the entry to see if the writer completed.
+                // We will try this in a loop till we get committed data for the currentEntry.
+                while (currentEntry.getType() == DataType.EMPTY) {
```
Maithem: Is this patch really needed? Why not just fix the underlying problem (bug i reported #439). Also, this fix only works for a single node. Please consider a chain where the head is not the tail. For example, client1 writes to the head, but doesn't complete its write to the tail(it dies), client2 reads the tail and finds null. Consequently, client2 tries to fill a whole, but can't since it will encounter an overwrite exception when it tries to write to the head. Now client2 will wait forever for another client to propagate the write, but it should try to propagate that value itself as part of the fill hole mechanism. 

The hole filling mechanism, should not only write holes, but should propagate partial writes.

medhavidhawan: This bug is a short term fix where the reads block till they read committed data. This is a requirement from the read interface. The larger interface changes will fix the problem you have mentioned in #439. 
The interface changes proposed are
1. Reads will surface only committed data.
2. Hole filling is an abstraction that sits somewhere between address space and replication, so it should be in the abstract replication layer. I think this is where we should fix the bug in #439.



It does fix one problem small minor problem in a chain:
1. If the actual write completes, read will finally read it.
2. If the hole fill fails for some reason, it will keep trying. Assuming that hole fill is the data that will be eventually put into the address space.

test/src/test/java/org/corfudb/runtime/collections/SMRMultiLogunitTest.java
medhavidhawan: Yes will do that. These are borrowed tests will get help from the author. Did not want to checkin code without tests.

test/src/test/java/org/corfudb/runtime/collections/SMRMultiLogunitTest.java
```diff
@@ -0,0 +1,216 @@
...
+                .build()
+                .setStreamName("test")
+                .setTypeToken(new TypeToken<SMRMap<String, String>>() {})
+                .open();
+        mutableInt threadsComplete = new mutableInt(0);
+        addTestStep((step) -> {
+            for (int i=0; i < numKeys; i++) {
+                String key = "key" + String.valueOf(i);
+                String val = "value" + String.valueOf(step) + "_" + String.valueOf(i);
+                //System.out.println("Thread " + String.valueOf(Thread.currentThread()) + " key " + key);
```
no2chem: Yeah, remove these comments...

test/src/test/java/org/corfudb/runtime/collections/SMRMultiLogunitTest.java
medhavidhawan: Ok cleaned up the tests to the best of my understanding.

PR #527, opened 2017-04-07
chain replication implementation

no2chem:
This PR implements a basic chain replication protcol on top of the new 
replication protocol abstraction. 

The implementation is not yet fully functional, since the log unit
client needs to be modified to not perform serialization or
handle metadata.

Also, bulk operators for read and peek are added (but not implemented
in chain replication, only default implementations are added).

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
@@ -7,7 +7,9 @@ import org.corfudb.protocols.logprotocol.StreamData;
 import org.corfudb.protocols.logprotocol.StreamedLogData;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.exceptions.OverwriteException;
+import org.corfudb.runtime.exceptions.RecoveryException;
 import org.corfudb.runtime.view.Layout;
+import org.corfudb.util.AutoClosableByteBuf;
 import org.corfudb.util.CFUtils;
 import org.corfudb.util.serializer.Serializers;
 
@@ -34,12 +36,11 @@ public class ChainReplicationProtocol extends AbstractReplicationProtocol {
                       @Nonnull Map<UUID, StreamData> entryMap) throws OverwriteException {
         int numUnits = layout.getSegmentLength(globalAddress);
 
-        ByteBuf b = Unpooled.buffer();
-        try {
+        try (AutoClosableByteBuf b = new AutoClosableByteBuf()) {
             // To reduce the overhead of serialization, we serialize only the
             // first time we write, saving when we go down the chain.
             StreamedLogData sld = new StreamedLogData(entryMap);
-            sld.getSerializedForm(Serializers.CORFU::serialize, b);
+            sld.getSerializedForm(Serializers.CORFU::serialize, b.getBuf());
 
             for (int i = 0; i < numUnits; i++) {
                 log.trace("Write[{}]: chain {}/{}", layout, i + 1, numUnits);
@@ -59,8 +60,6 @@ public class ChainReplicationProtocol extends AbstractReplicationProtocol {
                     throw oe;
                 }
             }
-        } finally {
-            b.release();
         }
     }
 
@@ -111,7 +110,7 @@ public class ChainReplicationProtocol extends AbstractReplicationProtocol {
         // it should have read to determine whether the
         // write was successful or not.
         if (ld == null) {
-            throw new RuntimeException();
+            throw new RecoveryException("Failed to read data during recovery at chain head.");
         }
         // now we go down the chain and write, ignoring any overwrite exception we get.
         for (int i = 1; i < numUnits; i++) {
--- a/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
@@ -1,7 +1,6 @@
 package org.corfudb.runtime.view.replication;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.Unpooled;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.logprotocol.StreamData;
 import org.corfudb.protocols.logprotocol.StreamedLogData;
@@ -14,6 +13,7 @@ import org.corfudb.util.CFUtils;
 import org.corfudb.util.serializer.Serializers;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.util.Map;
 import java.util.UUID;
 
@@ -42,16 +42,15 @@ public class ChainReplicationProtocol extends AbstractReplicationProtocol {
             StreamedLogData sld = new StreamedLogData(entryMap);
             sld.getSerializedForm(Serializers.CORFU::serialize, b.getBuf());
 
-            for (int i = 0; i < numUnits; i++) {
-                log.trace("Write[{}]: chain {}/{}", layout, i + 1, numUnits);
-                // In chain replication, we write synchronously to every unit
-                // in the chain.
+            log.trace("Write[{}]: chain head {}/{}", 1, numUnits);
+            // In chain replication, we start at the chain head.
                 try {
                     CFUtils.getUninterruptibly(
-                            layout.getLogUnitClient(globalAddress, i)
+                            layout.getLogUnitClient(globalAddress, 0)
                                     // TODO: Currently, this call won't work until the log unit client is refactored.
                                     .write(globalAddress, null, null, b, null)
                             , OverwriteException.class);
+                    propagate(globalAddress, b.getBuf());
                 } catch (OverwriteException oe) {
                     // Some other wrote here (usually due to hole fill)
                     // We need to invoke the recovery protocol, in case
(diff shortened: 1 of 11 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5e817d9**.

*** 0.0% transaction FAILURE rate for 5 threads, CorfuDiskNode mode
*** 0.01% transaction FAILURE rate for 10 threads, CorfuDiskNode mode

 [Pull Request #527 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/anTNPgIoDrSpDNGhc40myYWXiqH6pUc7)

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -0,0 +1,155 @@
...
+                    CFUtils.getUninterruptibly(
+                            layout.getLogUnitClient(globalAddress, i)
+                                    // TODO: Currently, this call won't work until the log unit client is refactored.
+                                    .write(globalAddress, null, null, b, null)
+                            , OverwriteException.class);
+                } catch (OverwriteException oe) {
+                    // Some other wrote here (usually due to hole fill)
+                    // We need to invoke the recovery protocol, in case
+                    // the write wasn't driven to completion.
+                    recover(globalAddress);
```
medhavidhawan: would be good to release the buffer b before you start recovery

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -0,0 +1,155 @@
...
+                        layout.getLogUnitClient(globalAddress, i)
+                                // TODO: Currently, this call won't work until the log unit client is refactored.
+                                .write(globalAddress, null, null, ld, null)
+                        , OverwriteException.class);
+                // We successfully recovered a write to this member of the chain
+                log.debug("Recover[{}]: recovered write at chain {}/{}", layout, i + 1, numUnits);
+            } catch (OverwriteException oe) {
+                // This member already had this data (in some cases, the write might have
+                // been committed to all members, so this is normal).
+                log.debug("Recover[{}]: overwritten at chain {}/{}", layout, i + 1, numUnits);
```
medhavidhawan: Would like to assert here that the committed data is the same as we were wanting to commit. How can we do that ?

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -0,0 +1,155 @@
...
+        // If nothing was at the head, this is a bug and we
+        // should fail with a runtime exception, as there
+        // was nothing to recover - if the head was removed
+        // due to a reconfiguration, a network exception
+        // would have been thrown and the client should have
+        // retried it's operation (in this case of a write,
+        // it should have read to determine whether the
+        // write was successful or not.
+        if (ld == null) {
+            throw new RuntimeException();
```
no2chem: Thanks, good catch.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
no2chem: I think the best way would be to do this at the logging unit. When it reports overwrite exception, it can report if the data was the same or not (maybe using the checksum it's already generating anyway.)

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
no2chem: Is it that critical? That means I have to do a bit of refactoring since currently the buffer is released by the try-with-resources block.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5e817d9**.

 [Pull Request #527 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/eVSzeAOGWhyraU5isBpq3xAcXV3Emc30)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5e817d9**.

 [Pull Request #527 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Dp4Qiq1X2W304g0d1VDu5yuavx1so9tD)

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
medhavidhawan: not critical, but good to have. I have not thought through what happens when we see a lot of these happening at the same time.

dahliamalkhi: while at it: if possible, actually use try-with-resource, rather than finally

no2chem: The memory overhead would just be a single 4K log entry for the duration of the recovery...

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
medhavidhawan: Yes I think I was wanting that. Sorry for not being clear. The assert can use the checksum. You need to boost the overwriteexception with this useful data.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
medhavidhawan: Not sure if I would measure it like that. This recovery going on for one write. What if a lot of recoveries are going on at the same time. A scenario where client(s) sputtered with a bunch of writes and all of them are competing with hole fills now.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -0,0 +1,155 @@
...
+    @Override
+    protected void holeFill(long globalAddress) {
+        int numUnits = layout.getSegmentLength(globalAddress);
+        for (int i = 0; i < numUnits; i++) {
+            log.trace("fillHole[{}]: chain {}/{}", globalAddress, i + 1, numUnits);
+            // In chain replication, we write synchronously to every unit in
+            // the chain.
+            try {
+                CFUtils.getUninterruptibly(layout.getLogUnitClient(globalAddress, i)
+                        .fillHole(globalAddress), OverwriteException.class);
```
medhavidhawan: As long as your fillHole on the log unit server side returns a success even if there was a hole already committed on the address, it will be ok. We will have less exception handling to do.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
no2chem: The key here is that clients is plural here. Unless the client is reading from a million streams that all happen to have holes in them at different global addresses at the same time, there really should only be 1 or very few recoveries happening per client.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
no2chem: Not ok. This is precisely the problem in the code right now. Overwrite occurs at the head,  but no write has propogated to the tail. The client must recover if it gets an overwrite exception from any log unit during a write to adopt the other value.

no2chem:
@dahliamalkhi I don't see how copyKnown reduces load. It results in the same number of read/write operations, it seems?

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
no2chem: ok, updated to use try-with-resources

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
medhavidhawan: No no I am not saying that. It has to do that , but if the whole fill already happened on a log unit, you can return success so that the client can go try the next logunit. Not sure if I made myself clear

no2chem:
@dahliamalkhi Ok, added

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -0,0 +1,185 @@
...
+        this.layout = layout;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void write(long globalAddress,
+                      @Nonnull Map<UUID, StreamData> entryMap) throws OverwriteException {
+        int numUnits = layout.getSegmentLength(globalAddress);
+
+        try (AutoClosableByteBuf b = new AutoClosableByteBuf()) {
```
Maithem: Why do you need a AutoClosableByteBuf? why not just use a Unpooled.buffer?

no2chem: To be sure to release the buffer using try-with-resources. Internally autoclosable bytebuf uses a unpooled allocator but it can use a pooled one in the future.

Maithem: Isee. I thought that we wanted to move to jvm managed buffers so that we can avoid explicitly managing them.

PR #532, opened 2017-04-08
first round of changes to expose streams to the log unit

no2chem:
This PR introduces the first round of changes which will
expose the contents of each stream to the logging units.

The log unit interface has been simplified to take a
single parameter, which is the log data itself.

Managining serialization during a write to multiple
logging units has also been enhanced.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/StreamedLogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/StreamedLogData.java
@@ -1,48 +1,142 @@
 package org.corfudb.protocols.logprotocol;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
 import lombok.Getter;
 import org.corfudb.protocols.wireprotocol.DataType;
+import org.corfudb.protocols.wireprotocol.ICorfuPayload;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
 
 import javax.annotation.Nonnull;
-import java.util.EnumMap;
-import java.util.Map;
-import java.util.UUID;
-import java.util.function.BiConsumer;
+import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * Temporary class for transition to log data.
  *
  * Created by mwei on 4/6/17.
  */
-public class StreamedLogData implements ILogData {
+public class StreamedLogData implements ILogData, ICorfuPayload<StreamedLogData>{
 
+    /** A map of stream data. */
     @Getter
     public final @Nonnull
-    Map<UUID, StreamData> entryMap;
+    Map<UUID, StreamData> streamDataMap;
 
-    public StreamedLogData(Map<UUID, StreamData> entryMap) {
-        this.entryMap = entryMap;
+    /** The metadata for this log entry. */
+    @Getter
+    final EnumMap<LogUnitMetadataType, Object> metadataMap;
+
+    /** The serialized form, cached during writing only
+     * and auto-released by the handle.
+     */
+    protected ByteBuf serializedForm;
+
+    /** {@inheritDoc} */
+    @Override
+    public Set<UUID> getStreams() {
+        return streamDataMap.keySet();
     }
 
-    public void getSerializedForm(BiConsumer<Object, ByteBuf> serializer, ByteBuf b) {
-        // go through each stream data and serialize it.
+    /** {@inheritDoc} */
+    @Override
+    public boolean containsStream(UUID stream) {
+        return streamDataMap.containsKey(stream);
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public Map<UUID, Long> getBackpointerMap() {
+        return streamDataMap.entrySet().stream()
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue().backpointer));
+    }
+
+    /** Serialize to the given buffer, using the cached form
+     * if possible.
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/protocols/logprotocol/StreamedLogData.java
```diff
@@ -1,48 +1,142 @@
...
+    @Override
+    public boolean containsStream(UUID stream) {
+        return streamDataMap.containsKey(stream);
     }
 
+    /** {@inheritDoc} */
     @Override
-    public Object getPayload(CorfuRuntime t) {
-        return null;
+    public Map<UUID, Long> getBackpointerMap() {
```
dahliamalkhi: I can't imagine why we would need this method in the future; but perhaps we need it in the interim?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WriteRequest.java
```diff
@@ -38,6 +38,11 @@ public WriteRequest(WriteMode writeMode, DataType dataType,  Map<UUID, Long> str
         this.data = new LogData(dataType, buf);
     }
 
+    public WriteRequest(ILogData data) {
+        writeMode = WriteMode.NORMAL;
+        this.data = data;
+        streamAddresses = null;
```
dahliamalkhi: can we get rid of streamAddresses?

with quorum replication, we are not supporting Replex anyway.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **6bee0cd**.

 [Pull Request #532 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/oqdFxm4FT5qW9s7eIIWy9kQWjSV8WgmS)

runtime/src/main/java/org/corfudb/protocols/logprotocol/StreamedLogData.java
no2chem: We do need it in the interim.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WriteRequest.java
no2chem: Would be nice, but it's pretty entrenched in the code base. I thought the goal was small PRs. :)

runtime/src/main/java/org/corfudb/protocols/logprotocol/StreamedLogData.java
```diff
@@ -1,48 +1,142 @@
...
     @Getter
     public final @Nonnull
-    Map<UUID, StreamData> entryMap;
+    Map<UUID, StreamData> streamDataMap;
 
-    public StreamedLogData(Map<UUID, StreamData> entryMap) {
-        this.entryMap = entryMap;
+    /** The metadata for this log entry. */
+    @Getter
+    final EnumMap<LogUnitMetadataType, Object> metadataMap;
```
medhavidhawan: Do we not want to specify the metadata type ?

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -84,7 +86,7 @@ public ILogData peek(long globalAddress) {
      * @param data          The data to propagate, or NULL,
      *                      if it is to be a hole.
      */
-    protected void propagate(long globalAddress, @Nullable ByteBuf data) {
+    protected void propagate(long globalAddress, @Nullable ILogData data) {
         int numUnits = layout.getSegmentLength(globalAddress);
 
         for (int i = 1; i < numUnits; i++) {
```
medhavidhawan: propagate starts from i = 1.
and recover also starts from i = 1 even if the failure could have been at the second log unit. This seems to be inefficient.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -234,7 +234,15 @@ private static Object handleReadDataCorruption(CorfuPayloadMsg<ReadResponse> msg
...
 
-
+    /**
+     * Asynchronously write to the logging unit.
+     * @param payload   The log data to write to the logging unit.
+     * @return          A CompletableFuture which will complete with the WriteResult once the
+     *                  write completes.
+     */
+    public CompletableFuture<Boolean> write(ILogData payload) {
+        return router.sendMessageAndGetCompletable(CorfuMsgType.WRITE.payloadMsg(new WriteRequest(payload)));
```
medhavidhawan: I did check the StreamLog code and it does not take into account that if an address already has hole fill data and the  new write request is for a hole fill, returning success would be a good optimization but it throws OverwriteException.
Something to consider as hole fills might have a tendency to come from many clients concurrently.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
dahliamalkhi: good point. it should be easy to fix -- let proparte take a start-index parameter.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
dahliamalkhi: yes.

this is not blocking us currently, so we can go ahead and merge this PR, but definitely keep this enhancement under consideration for the future.

PR #539, opened 2017-04-13
Cache Segment Handles

Maithem:
This patch introduces a loading cache to cache the segment handles.
One consequence of not having all segment handles in memory is
compaction can only compact segments that are visible from the
cache.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -27,10 +27,17 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.cache.RemovalListener;
+import com.google.common.cache.RemovalNotification;
 import com.google.common.hash.Hasher;
 import com.google.common.hash.Hashing;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.google.protobuf.AbstractMessage;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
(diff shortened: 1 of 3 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -81,19 +90,74 @@
...
+        return CacheBuilder.newBuilder()
+                .maximumWeight(numSegments)
+                .weigher((Long, SegmentHandle) -> {
+                            return 1;
+                        }
+                )
+                .removalListener(new RemovalListener() {
+                    @Override
+                    public void onRemoval(RemovalNotification notification) {
+                        // Would this cause NPE for other objects holding this segment's reference?
```
no2chem: Well would it? If it does, please check for it...

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **1e16110**.

 [Pull Request #539 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/4e08lt0HzoYy0tBNa3n6wxk6vRxoz37r)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: Theoretically this can happen when you have an object that holds a reference to a write channel, but the segment that corresponds to that channel has been evicted from the cache. Writing to that channel will  cause an NPE because on eviction the channel is closed and set to null.

So, this can happen on a trim, append or ready. Since those operations call getSegmentHandleForAddress before accessing the channels, the NPE can happen only if the corresponding segment is evicted by the time those operations try to use that channel. For that segment to be evicted within that small time frame, a large number of read/write requests must occur, which is extremely unlikely given the large size of the cache and our access pattern.

For example, when writing address 0 (assume the cache is empty), for an NPE to happen 10,000 random read/write(each segment resulting in loading a new segment into the cache) request must happen after getSegmentHandleForAddress(addr 0) and before channel.write(addr 0). 

The possibility of this happening can be reduced further at the expense of increasing the size of the cache, which adds a small reasonably memory overhead.

I think the probability of it occurring is extremely small that it can be ignored.

no2chem: Okay, so why don't you just check if notification.getValue is null?, or try-catch the NPE?

Maithem: The comment was actually not about an NPE inside the removal listener, it was about other objects holding references to the channel inside the segment handler after it has been evicted from the cache. For example fh.logChannel.write(recordBuf); in writeRecord will throw an NPE if fh (segment handle) has been evicted within the critical time described above,

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a00916a**.

 [Pull Request #539 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/D0DJ6jEkxxzwmKsinL9tCTVuYJUuXY46)

Maithem:
@no2chem can we merge this?

Issue #548, opened 2017-04-19
FileChannel leak when we fetch a LogData from file on disk


rogermichoud:
In the readRecord function (that is called when we fetch the entry from the file in the cache on the logUnitServer) we never close the FileChannel.
 ```
    private LogData  readRecord(SegmentHandle sh, long address)
            throws IOException {

        FileChannel fc = getChannel(sh.fileName, true);
        AddressMetaData metaData = sh.getKnownAddresses().get(address);
        if (metaData == null) {
            return null;
        }

        fc.position(metaData.offset);

        try {
            ByteBuffer entryBuf = ByteBuffer.allocate(metaData.length);
            fc.read(entryBuf);
            return getLogData(LogEntry.parseFrom(entryBuf.array()));
        } catch (InvalidProtocolBufferException e) {
            throw new DataCorruptionException();
        }
    }
```
This problem got exposed on longevity testing. After a while, performance dropped completely and I had ~ 300'000 File Descriptor open on the same log file.

PR #549, opened 2017-04-19
Clean up FileChannel leak

rogermichoud:
In the readRecord function (that is called when we fetch the entry from the file in the cache on the logUnitServer) we never close the FileChannel.

This problem got exposed on longevity testing. After a while, performance dropped completely and I had ~ 300'000 File Descriptor open on the same log file.

#548

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -501,7 +501,9 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
                 throw new DataCorruptionException();
             }
         } finally {
-            fc.close();
+            if (fc != null) {
+                fc.close();
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -483,21 +483,25 @@ private void readAddressSpace(SegmentHandle sh) throws IOException {
...
-            throw new DataCorruptionException();
+            try {
+                ByteBuffer entryBuf = ByteBuffer.allocate(metaData.length);
+                fc.read(entryBuf);
+                return getLogData(LogEntry.parseFrom(entryBuf.array()));
+            } catch (InvalidProtocolBufferException e) {
+                throw new DataCorruptionException();
+            }
+        } finally {
+            fc.close();
```
kspirov: if (fc != null) ?

kspirov:
Congratulations you found this. Since I first played with StreamLogFiles was almost certain there is such issue,  Windows unit tests related to StreamLogFiles fail which is a clear symptom for unclosed read stream.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
no2chem: yes, good catch!

no2chem:
closes #548

rogermichoud:
Made the change. Thanks @kspirov for the review.

Issue #553, opened 2017-04-25
Corfu Server default heap size is too small


rogermichoud:
Default value of the jvm heap for corfu server is 1 Gb. Max size of LogUnit cache is 1Gb. What could go wrong :).

There are 2 approaches:

1. Bumping the default jvm heap size to (e.g. 2gb or 1.5gb)
2. Reducing the cache size to 500Mb
3. Make the cache size based on the heap max size (0.5 * total heap, for example to be safe).

On longevity test with 1gb cache size and 2gb max heap, we are cruising at 1.5 gb of heap allocated and 1.3 gb of old gen.

Which one you think make more sense ? I like number 3, like this we are safe whatever the user do.

PR #554, opened 2017-04-26
Change default size of CorfuServer jvm heap to 2gb

rogermichoud:
Default LogUnit cache is 1gb. If we run everything on default we are doomed.

#553


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5058f03**.

 [Pull Request #554 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/PQoDeADD2ooLMxrm5Vjjpped5E0NP7Pa)

scripts/corfu_server.sh
```diff
@@ -37,7 +37,7 @@ fi
 
 
 # default heap for corfudb
-CORFUDB_HEAP="${CORFUDB_HEAP:-1000}"
+CORFUDB_HEAP="${CORFUDB_HEAP:-2000}"
```
Maithem: Can you just read the vm option from the logunit server and compute a ratio for the cache. That is a small enough change.

rogermichoud: Sure. As mentioned on the issue, this was my favorite approach as well !

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a50b4db**.

 [Pull Request #554 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/APMA0VyaX8IOCxfNbawvpWKlsrcQ1ld9)

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -106,7 +108,11 @@
     public LogUnitServer(ServerContext serverContext) {
         this.opts = serverContext.getServerConfig();
         this.serverContext = serverContext;
-        maxCacheSize = Utils.parseLong(opts.get("--max-cache"));
+        long maxCacheSizeFromUser = Utils.parseLong(opts.get("--max-cache"));
+
+        /* Ensure that maxCacheSize is at most CACHE_HEAP_RATIO * maxHeapSize */
```
Maithem: "at most" seems to be arbitrary. If a user specifies a cache size that should be used, otherwise use a default size which is the the cache heap ratio.

rogermichoud: It prevents the user to specify a cache that is too big for the heap. If we let the user do this, it will end up in OutOfMemory exception. The cache heap ratio is the biggest ratio that deem safe to use.

Maithem: Right, but that's no different from supplying other wrong configs. I don't think we should mask bad config parameters. Also, if I was tweaking the server to figure out the best cache ratio, it doesn't make sense to recompile every time.

rogermichoud: Agree on the compiling, the cache ratio should be handed as a parameter. On the other hand, if we can prevent the user to do something wrong or something that doesn't make sense I think we should do it.
Let's replace max-cache parameters with cache ratio. If the ration is more than 1, this is stupid, we should not let anybody do this. What do you think?

Maithem: I think that's reasonable.

PR #556, opened 2017-04-27
Check Metadata on Cache Load

Maithem:
Check that the entries have correct metadata when they are loaded
into the cache.


slfritchie:
Hi, Maithem.  The "5L" constant on line 196 is angering the style checker.

Maithem:
@slfritchie thanks for the note :), ill fix it and rebase

PR #558, opened 2017-04-27
Sequencer amnesia fix

dahliamalkhi:
This PR is another intermediate-step towards fixing the sequencer failover amnesia bug.

I removed the checkpointing/recovery of sequencer state, and left it as dead-code in comments with "TODO" and "FIXME" annotations. This is so if we need to make use of parts of it we can do it quickly. It should be removed once  we have confidence in this fix.


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -242,7 +242,7 @@ public void syncObjectUnsafe(long timestamp) {
         // and it belongs to this thread use that
         if (optimisticallyOwnedByThreadUnsafe()) {
             // If there are no updates, ensure we are at the right snapshot
-            if (optimisticStream.pos() == Address.NEVER_READ) {
+            if (optimisticStream.pos() == Address.NON_ADDRESS) {
```
no2chem: Can we move to NON_ADDRESS change to another PR, I don't agree with this because the constant is being used in a lot of places where it means precisely that: that the stream/object/etc has never been read.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **19e16d3**.

 [Pull Request #558 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/6lQDRZmYky8Uk8LKG1T62GExZvnTQr08)

no2chem:
The latest push still uses NON_ADDRESS in places where it should be NEVER_READ (a stream which has never returned a read result should return NEVER_READ - NON_ADDRESS is not clear at all!). I don't want to put add a semantic argument in this PR, plus all the NON_ADDRESS additions definitely conflict with an upcoming PR which implements spliterator.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **88f85a0**.

 [Pull Request #558 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/BigTEjP65rBqaspTPujKiN1V5h94DY0T)

runtime/src/main/java/org/corfudb/runtime/view/Address.java
```diff
@@ -8,11 +8,34 @@
...
+    /**
+     * @param addr
+     * @return true is addr is a legitimate address; false for all flag non-address constants
+     */
+    public static boolean isAddress(long addr) { return addr >= 0; }
+
+    /**
+     * @return a constant which can be used as the base for address iterations
+     */
+    public static long minAddress() { return 0L; }
```
no2chem: Shouldn't this be just a static field like MAX?

runtime/src/main/java/org/corfudb/runtime/object/transactions/WriteSetSMRStream.java
```diff
@@ -160,14 +161,14 @@ void mergeTransaction() {
     public List<SMREntry> previous() {
         writePos--;
 
-        if (writePos <= Address.NEVER_READ) {
-            writePos = Address.NEVER_READ;
+        if (writePos <= Address.maxNonAddress()) {
```
medhavidhawan: should this check be a utility of the Address class ? Why do  we need to expose this ?

runtime/src/main/java/org/corfudb/runtime/view/stream/ReplexStreamView.java
```diff
@@ -159,7 +159,7 @@ public ILogData previous() {
         doPendingSeek(getCurrentContext(), Address.MAX);
 
         // If never read or only one entry, return null
-        if (getCurrentContext().streamPointer == Address.NEVER_READ ||
+        if (getCurrentContext().streamPointer <= Address.maxNonAddress() ||
```
medhavidhawan: I think we should pull these comparisons into the Address class. The operators need not be exposed.

test/src/test/java/org/corfudb/integration/ServerRestartTest.java
```diff
@@ -132,6 +132,37 @@ public void testSequencerRecoveryFailoverServerAndClient() throws Exception {
     }
 
     /**
+     * Test recovery with a failed over server and a new client.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testDoubleFailoverRecovery() throws Exception {
```
medhavidhawan: Great test. I think zeeshan can add more variations to this. Including writing and reading more data.

zalokhan: Isn't this test exactly same as the one above this? testSequencerFailoverServerAndClient.
Secondly, contrary to the name there is only one failover taking place in the test.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **7e0e849**.

 [Pull Request #558 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/UvTdXCeG6YRfKuRGOiTDdMrFJ8Olvq6l)

PR #568, opened 2017-04-29
Stateless Sequencer amnesia fix

zalokhan:
No longer persisting state in the sequencer.
Added randomized test for recovering sequencer state.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -292,11 +292,7 @@ public class SequencerServer extends AbstractServer {
     @ServerHandler(type=CorfuMsgType.RESET_SEQUENCER, opTimer=metricsPrefix + "reset")
     public synchronized void resetServer(CorfuPayloadMsg<Long> msg, ChannelHandlerContext ctx, IServerRouter r,
                                          boolean isMetricsEnabled) {
-        // TODO: remove this:
-        long restoredGlobalTail = restoreGlobalLogTail();
-        long initialToken = restoredGlobalTail > msg.getPayload() ? restoredGlobalTail : msg.getPayload();
-        // TODO: and use this instead:
-        // long initialToken = msg.getPayload();
+         long initialToken = msg.getPayload();
 
         //
         // if the sequencer is reset, then we can't know when was
@@ -317,11 +313,6 @@ public class SequencerServer extends AbstractServer {
             globalLogStart.set(initialToken);
             maxConflictWildcard = initialToken-1;
             conflictToGlobalTailCache.invalidateAll();
-            // FIXME: Do not preserve Sequencer State.
-            //streamTailMap.clear();
-            //streamTailToGlobalTailMap.clear();
-            //streamTailMap.putAll(restoreStreamTailMap());
-            //streamTailToGlobalTailMap.putAll(restoreStreamTailToGlobalTailMap());
         }
 
         log.info("Sequencer reset with token = {}", initialToken);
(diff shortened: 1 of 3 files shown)


test/src/test/java/org/corfudb/integration/AbstractIntegrationTest.java
```diff
@@ -0,0 +1,147 @@
...
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.concurrent.Executors;
+
+/**
+ * Created by zlokhandwala on 4/28/17.
+ */
+public class AbstractIntegrationTest extends AbstractCorfuTest {
+
+    private static final String TEST_HOST = "localhost";
```
Maithem: What if you need multiple servers? having static ports doesnt seem like a good idea. We can address that later though.

test/src/test/java/org/corfudb/integration/ServerRestartTest.java
```diff
@@ -1,160 +1,73 @@
...
 
     /**
      * Test recovery with a failed over server and a new client.
      *
      * @throws Exception
      */
     @Test
-    public void testSequencerRecoveryFailoverServerAndClient() throws Exception {
-        setUp();
+    public void testRecoveryFailoverServerAndClient() throws Exception {
```
Maithem: It seems like all these tests are just a subset of testRandomizedRecovery. Can you clobber them?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -509,41 +493,4 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
     }
 
     private static final int globalTokenBatchSize = 100;
-
```
dahliamalkhi: remove previous line as well

test/src/test/java/org/corfudb/integration/AbstractIntegrationTest.java
```diff
@@ -0,0 +1,147 @@
+package org.corfudb.integration;
```
dahliamalkhi: cool, this is a userful utility class for integration tests.

Maithem:
Please make sure that you address the comments in another patch.

test/src/test/java/org/corfudb/integration/ServerRestartTest.java
```diff
@@ -163,87 +76,174 @@ public void testDoubleFailoverRecovery() throws Exception {
...
         map.put("a", expectedValue);
 
-        map = createMap(createRuntime());
+        shutdownCorfuServer(corfuServerProcess);
 
-        assertThat(map.get("a")).isEqualTo(expectedValue);
+        assert !corfuServerProcess.isAlive();
+
+        corfuServerProcess = runCorfuServer();
+
```
dahliamalkhi: add comment here: do we expect some delay until reconfig or reconnect or both ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
zalokhan: Ok Done.

zalokhan:
@Maithem Yes, I will address the comments in the next patch

test/src/test/java/org/corfudb/integration/ServerRestartTest.java
zalokhan: Yes, we do expect slight delay. I'll add a comment.

test/src/test/java/org/corfudb/integration/ServerRestartTest.java
```diff
@@ -1,160 +1,73 @@
...
         final int expectedValue = 5;
 
         Integer previous = map.get("a");
         assertThat(previous).isNull();
         map.put("a", expectedValue);
 
+        // Shut down and assert node is down.
         shutdownCorfuServer(corfuServerProcess);
-
         assert !corfuServerProcess.isAlive();
```
zalokhan: I'll make the change.

test/src/test/java/org/corfudb/integration/ServerRestartTest.java
zalokhan: I'll reduce redundancy.

medhavidhawan:
If you are going to do a second pass, please add facility to dump the sequence of actions when there is a failure. In other words please keep an array of actions (as defined by a test) as a trace of the test and dump it for analysis if there is a failure. You could also dump the seed of the Random number generator. I think we can build on this abstraction and can have a lot of smoke tests for end to end functionality. It can also be a good facilitator in fuzz testing. Please ping me for your second iteration, we can discuss some ideas.

Issue #566, opened 2017-04-28
New runtime streaming LogData to remainingUpTo creates container of all updates


kjames88:
When a corfu stream is accessed on a new client, all logged updates need to be read from log unit.  The streaming implementation in corfu collects all the entries through a series of data structures into a flat list in StreamViewSMRAdapter::remainingUpTo().  Profiling shows 30k reachable LogData objects using 430MB in an L2 test with cache size of 100MB.

slfritchie:
Mid-June issues review: ping @kjames88, I don't know if you've looked at this case in the `master` recently.

no2chem:
Performance issue, let's complete by v1.0-M3.

PR #569, opened 2017-04-29
Removal of deprecated classes, cleanup of address view and implementation of spliterator

no2chem:
This pull request removes support for Replex, which will not be available in the 1.0 release.
It also removes support for runtime instrumentation (after this PR, the depedency on
ByteBuddy can be removed).

The annotationProcessor has been updated to support inner static classes (anonymous classes
were never supported under ByteBuddy). 

Deprecated APIs in AddressSpaceView have been removed (such as holeFill), and deprecated
methods in ObjectView and LogUnitClient have also been removed.

A new spliterator implementation reduces memory pressure during traversal of a large
stream for the first time.


no2chem:
Partially resolves #566

Maithem:
@no2chem do you need the split iterator implementation to remove the deprecated classes and replex?
would it be hard to separate the two?

runtime/src/main/java/org/corfudb/runtime/view/AbstractReplicationView.java
```diff
@@ -1,129 +0,0 @@
-package org.corfudb.runtime.view;
```
dahliamalkhi: I understand we are focusing this branch on quorum replication only. But why not retain the abstractions and interfaces that allow flexible replication schemes in the future?

runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
```diff
@@ -1,14 +1,14 @@
 package org.corfudb.runtime.view;
 
 import com.google.common.reflect.TypeToken;
+import com.sun.xml.internal.bind.v2.model.runtime.RuntimeBuiltinLeafInfo;
```
dahliamalkhi: mistakenly inserted?

runtime/src/main/java/org/corfudb/runtime/view/AbstractReplicationView.java
no2chem: But we do, this code was marked deprecated, and the flexibility is retained in protocols

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -37,7 +37,7 @@ public Object getPayload(CorfuRuntime runtime) {
                     if (data == null) {
                         this.payload.set(null);
                     } else {
-                        ByteBuf copyBuf = Unpooled.copiedBuffer(data);
+                        ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
```
Maithem: I don't think so. The memory leaks were coming from pooled memory.

no2chem: not a memory leak, but an unnecessary copy, I believe

no2chem:
@Maithem will address protobufs in a separate PR

Maithem:
@no2chem dont worry, I can do it. Im working on another cleanup patch.

PR #573, opened 2017-05-01
Remove Javafx Pair

Maithem:
Implemented a generic pair to replace javafx's pair.


Maithem:
@no2chem You mean CompactedEntry and not CompactedEntrySet?
Essentially, change the pair to CompactedEntry and remove the generics. 
Would that be fine?

PR #508, opened 2017-03-21
Add coalesce support, cleanup map interface

no2chem:
This patch adds conflict and undo support for the putAll operation,
as well as support for coalescing operations during transactions.

Coalesced operations have a much smaller write set. In the map implementation,
redundant operations (such as operations before a clear, or remove operations
before a put operation on the same key) are removed, and all put operations are
combined into a single putAll operation. This should reduce both the
serialization and playback overhead of large transactions.

PR #577, opened 2017-05-02
Implement conflict paramters for bulk operations

### LINKED ISSUE #574: putAll() false sharing issue
```java
  @Override
  @Mutator(
      name = "putAll"
  )
  public void putAll(Map<? extends K, ? extends V> m) {
    proxy_CORFUSMR.logUpdate("putAll",false,null,m);
  }
```

no2chem:
This PR is a cherry pick of #508, which implements conflict paramter
resolution for bulk operations such as putAll, which is calculated
using a custom conflict parameter function.

This PR closes #574.

### DIFF
--- a/annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
+++ b/annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
@@ -119,19 +119,19 @@ public class ObjectAnnotationProcessor extends AbstractProcessor {
     /** Class to hold data about SMR Methods. */
     class SMRMethodInfo {
         /** The method itself. */
-        ExecutableElement method;
+        final ExecutableElement method;
         /** The interface, if present, which provided the element. */
-        TypeElement interfaceOverride;
+        final TypeElement interfaceOverride;
         /** If the element should be excluded from instrumentation. */
-        boolean doNotAdd = false;
+        final boolean doNotAdd;
         /** If the element has conflictParameter annotations. */
         final boolean hasConflictAnnotations;
+        /** If the element has a function to calculate conflict params. */
+        final String conflictFunction;
 
         public SMRMethodInfo(ExecutableElement method,
                              TypeElement interfaceOverride) {
-            this.method = method;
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/ISMRMap.java
```diff
@@ -1,158 +1,87 @@
 package org.corfudb.runtime.collections;
 
-import com.google.common.collect.ImmutableSet;
-import org.corfudb.annotations.*;
+import com.google.common.collect.ImmutableMap;
+import org.corfudb.annotations.*;;
 
 import java.util.*;
-import java.util.concurrent.ConcurrentMap;
```
Maithem: expand imports.

annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
```diff
@@ -382,8 +382,28 @@ else if (
...
+                    }
+                    if (mutator != null && !mutator.conflictParameterFunction().equals("")) {
+                        conflictFunction = mutator.conflictParameterFunction();
+                    }
+                    if (mutatorAccessor != null && !mutatorAccessor.conflictParameterFunction().equals("")) {
+                        conflictFunction = mutatorAccessor.conflictParameterFunction();
+                    }
+
+                    if (!conflictFunction.equals("")) {
+                        hasConflictData = true;
```
Maithem: Why don't you remove hasConflictAnnotations from SMRMethodInfo and replace it with hasConflictData and check if there exists either a conflict annotation or a conflict function. Wouldn't that be cleaner? 

Also, your if-else statement gives precedence to conflict functions over conflict annotations, if both exist. Is this by design ? maybe restrict that, so that you cant have both?

no2chem: Good idea, I'll go add that in

no2chem:
@Maithem , I've addressed all your review comments, let me know if it looks good to you.

Maithem:
I would squish the "fix import" commit.

PR #582, opened 2017-05-02
Code cleanups

dahliamalkhi:
This branch contains stylistic code cleanups. 

no functionality change -intended-.


Maithem:
@dahliamalkhi  I think the commit history can be improved. Please squash commits, remove your local merge commit and add git commit messages.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiSMREntry.java
```diff
@@ -64,6 +69,8 @@ public void setEntry(ILogData entry) {
 
     @Override
     public List<SMREntry> getSMRUpdates(UUID id) {
+        // FIXME MultiSMREntry should not implement ISMRConsumable; we have MultiObjectSMREntry for that
```
no2chem: Any consumable SMR entry should implement ISMRConsumable, I don't see how MultiSMREntry should be any different!

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -66,7 +66,7 @@
      * A view of streams in the Corfu server instance.
      */
     @Getter(lazy = true)
-    private final MultiStreamView streamsView = new MultiStreamView(this);
+    private final MultiStreamView multiStreamsView = new MultiStreamView(this);
```
no2chem: this is getting out of hand. I have an idea, we can keep the classes named the name, but the fields shorter, sane names:

runtime.layout()
runtime.streams()
runtime.objects()
runtime.addressSpace()

-etc-

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -79,6 +79,7 @@
     public final TransactionBuilder builder;
 
     /**
+     * TODO remove this!
```
no2chem: Why remove it? It seems like it may be useful to keep track of the running time of a transaction - and either way shouldn't have much overhead.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
dahliamalkhi: agreed, though 'layout' is already taken.

dahliamalkhi: for now, though, let's keep all names. don't want to break applications.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
dahliamalkhi: we have metering for that. this is completely isolated mechanism, mostly unknown and hiding. No reason to have it at this level.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiSMREntry.java
dahliamalkhi: ok.

dahliamalkhi:
@Maithem  done.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
no2chem: Lots of reasons to have it at this level, for example, metering won't give you per transaction level information, which you may very well want to introspect on (like self aborting if the transaction is too old...)

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
no2chem: Ok, then you're going to undo this change, right? If you're going to break applications, you might as well produce the intended change...

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -66,7 +66,7 @@
      * A view of streams in the Corfu server instance.
      */
     @Getter(lazy = true)
-    private final MultiStreamView multiStreamsView = new MultiStreamView(this);
+    private final StreamsView streams = new StreamsView(this);
```
no2chem: I think if you're going to do this, you should get rid of the get prefix on the autogenerated getter (no need to make it lazy, that was a remenant if the past...)

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
slfritchie: The completion time is currently used for a log.trace() message, I think.  If making it more official'y via a new metering statistic is useful, I'm fine with that.  I'll let y'all armwrestle about wisdom of other uses.

dahliamalkhi: is there anything that can't be done at application level (before TXbegin and post TXend)?

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -344,3 +274,74 @@ public String toString() {
...
+
+    public Set<Integer> getConflictSet(UUID streamID) {
+        return getReadSetConflicts().computeIfAbsent(streamID, u -> {
+            return new HashSet<>();
+        } );
+    }
+
+}
+
+@Getter @groovy.util.logging.Slf4j
```
no2chem: This appears to be the wrong slf4j annotation (you want Lombok not groovy)

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -17,27 +17,37 @@
  * Created by mwei on 9/20/16.
  */
 @ToString
-@NoArgsConstructor
 @Slf4j
-public class MultiObjectSMREntry extends LogEntry implements IDivisibleEntry, ISMRConsumable {
+public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
```
dahliamalkhi: ok, I get penalized for my good deeds...

 I'll add some comments.

PR #590, opened 2017-05-04
Broken Commit

Maithem:
This commit breaks the compilation process. This is a sanity check
to make sure that travis doesn't pass when compilation fails.


slfritchie:
Heh, as hoped for, Travis-CI thinks it's broken.

The weird thing about #589's problem is that both Zeeshan and I have confirmed that a fresh Maven cache and fresh checkout of master (or other commits with the bug) **will compile** on an Ubuntu 14 VM.  But the same procedure will fail under OS X.

Maithem:
@slfritchie make sure that the maven and jdk match.

Maithem:
It seems travis is failing on compilation errors.

Issue #591, opened 2017-05-05
IllegalArgumentException No matching method found: write for class - error


nikileshsa:
I just built Corfu from source and I am trying to run Corfu following the README. I am getting the following error when I try to execute the command below. Am I missing something here?
```
ssh:~/CorfuDB/bin$ echo "hello world" | ./corfu_stream write -c localhost:9000 -i test
WARNING: cat already refers to: #'clojure.core/cat in namespace: net.cgrand.parsley.fold, being replaced by: #'net.cgrand.parsley.fold/cat
IllegalArgumentException No matching method found: write for class org.corfudb.runtime.view.stream.BackpointerStreamView  clojure.lang.Reflector.invokeMatchingMethod (Reflector.java:53)
```

PR #592, opened 2017-05-05
Fix clojure CLI

Maithem:
The clojure scripts have went out of sync after the recent clean
up and refactoring. This patch syncs the corfu_stream.


Maithem:
This addresses https://github.com/CorfuDB/CorfuDB/issues/591

zalokhan:
I can tests to make sure the cmdlets stay in sync with the code.

Maithem:
@zalokhan done.

PR #597, opened 2017-05-09
avoid write-lock on access()

dahliamalkhi:
This is a small fix, to avoid taking a write lock on objects unnecessarily on access()


runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -83,9 +83,13 @@
         // to the correct version, reflecting any optimistic
         // updates.
         return proxy.getUnderlyingObject().access(
-                o -> (getWriteSetEntryList(proxy.getStreamID()) == null && // No updates
+                o -> (getWriteSetEntryList(proxy.getStreamID()).size() == 0 && // No updates
```
no2chem: I think you still need to check if the write set is null. So:
```
(getWriteSetEntryList(proxy.getStreamID()) == null ||
                getWriteSetEntryList(proxy.getStreamID()).size() == 0) &&
```

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -83,9 +83,13 @@
...
                         o.getVersionUnsafe() == getSnapshotTimestamp() && // And at the correct timestamp
-                        !o.isOptimisticallyModifiedUnsafe()),
+                                (o.getOptimisticStreamUnsafe() == null || o.getOptimisticStreamUnsafe()
+                                        .isStreamCurrentContextThreadCurrentContext() )
+                ),
+//                                o.optimisticallyOwnedByThreadUnsafe() ),
```
no2chem: probably can take these comments out?

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
dahliamalkhi: actually, getWriteSetEntryList returns empty collection, not null, on empty write set. This is the main reason for introducing this fix, we actually -always- take a write lock on every access.

I don't mind addding this check, to prevent exception occurring if getWriteSetEntryList changes in the future.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
dahliamalkhi: will do.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
no2chem: Well, if you mark getWriteSetEntryList's (and MultiObjectSMREntry::getSMRUpdates) return as @NotNull, then there is no need for the null check.

PR #598, opened 2017-05-09
Remove The useCompiledClass Flag

Maithem:
We no longer have runtime instrumentation, so this flag is not
needed.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
@@ -43,9 +43,6 @@ public class ObjectBuilder<T> implements IObjectBuilder<T> {
     @Setter(AccessLevel.NONE)
     Object[] arguments = new Object[0];
 
-    @Setter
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
```diff
@@ -83,31 +80,26 @@ public T open() {
...
+                                return CorfuCompileWrapperBuilder.getWrapper(type, runtime, streamID,
+                                        arguments, serializer);
+                            } catch (Exception ex) {
+                                throw new RuntimeException(ex);
+                            }
+                        }
+                );
             }
+        } catch (Exception ex) {
+            log.error("Couldn't use compiled class for {}, using runtime instrumentation.", type);
```
no2chem: I would use the message from before:
"Runtime instrumentation no longer supported and no compiled class found"

Maithem: Nice catch.

PR #601, opened 2017-05-10
Introduce SMRMap state in-stream checkpoint writer

slfritchie:
I've eliminated (I hope) the double-serialization, added a wrapper class that writes all checkpoint records for a given stream & SMRMap, updated unit tests, rebased to master (and compensated for the Spliterator refactoring), and fixed all other failing unit tests (which had been disturbed by serialization changes).

API for checkpoint dumping is currently via Java or Clojure.  For Clojure, use the following to set a runtime that connects to `localhost:9000` and uses an SMRMap on top of a stream called `my-stream`.  First, run `bin/shell`, then enter at the prompt:

    (get-runtime "localhost:9000")
    (.connect *r)
    (def my-stream "my-stream")
    (def my-stream-id (uuid-from-string my-stream))
    (def m (.. *r getObjectsView build (setType org.corfudb.runtime.collections.SMRMap) (setStreamName my-stream) open))

If the map is empty, here is a snippet to write 7 keys.
    
    (doseq [x [1 2 3 4 5 6 7]] (.put m (.concat "key" (Integer/toString x)) (+ 42 x)))

Then to checkpoint the map (by writing summary state into the same `my-stream` stream):

    (org.corfudb.runtime.CheckpointWriter/appendCheckpoint *r my-stream-id "Your name here" m)

Squash & rebase of slfritchie/checkpoint-in-stream2 at
commit 1e49965e490cf4dcfc62522a74987432433eb5f8

### DIFF
--- a/format/proto/types.proto
+++ b/format/proto/types.proto
@@ -8,6 +8,11 @@ enum DataType {
     PROPOSAL = 4;
 }
 
+enum CheckpointEntryType {
+    START = 0;
+    CONTINUATION = 1;
+    END = 2;
+}
 
 message DataRank {
     required int64 rank = 1;
@@ -23,6 +28,9 @@ message LogEntry {
     map<string, int64> logical_addresses = 7;
     map<string, int64> backpointers = 8;
     optional DataRank rank = 9;
+    optional CheckpointEntryType checkpointEntryType = 10;
+    optional int64 checkpointID_most_significant = 11;
+    optional int64 checkpointID_least_significant = 12;
 }
 
 message LogHeader {
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -50,6 +50,7 @@ import org.corfudb.format.Types.LogEntry;
 import org.corfudb.format.Types.LogHeader;
 import org.corfudb.format.Types.Metadata;
 import org.corfudb.infrastructure.ServerContext;
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.wireprotocol.IMetadata;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.runtime.exceptions.DataCorruptionException;
@@ -424,6 +425,7 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
         LogData logData = new LogData(org.corfudb.protocols.wireprotocol.
                 DataType.typeMap.get((byte) entry.getDataType().getNumber()), data);
 
+        // Checkpoint-related metadata are set by LogData constructor.
         logData.setBackpointerMap(getUUIDLongMap(entry.getBackpointersMap()));
         logData.setGlobalAddress(entry.getGlobalAddress());
         logData.setRank(createDataRank(entry));
@@ -713,6 +715,15 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
         Optional<Types.DataRank> rank = createProtobufsDataRank(entry);
         if (rank.isPresent()) {
             logEntryBuilder.setRank(rank.get());
+        }
+        if (entry.hasCheckpointMetadata()) {
+            logEntryBuilder.setCheckpointEntryType(
+                    Types.CheckpointEntryType.forNumber(
+                            entry.getCheckpointType().ordinal()));
+            logEntryBuilder.setCheckpointIDMostSignificant(
+                    entry.getCheckpointID().getMostSignificantBits());
+            logEntryBuilder.setCheckpointIDLeastSignificant(
+                    entry.getCheckpointID().getLeastSignificantBits());
         }
 
         return logEntryBuilder.build();
--- a/logReader/src/main/java/org/corfudb/logReader/logReader.java
+++ b/logReader/src/main/java/org/corfudb/logReader/logReader.java
@@ -17,6 +17,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.util.Map;
+import java.util.UUID;
 
 public class logReader {
     private static int metadataSize;
@@ -218,7 +219,7 @@ public class logReader {
                 System.out.println("DataType: TRIMMED");
                 break;
             default:
-                System.out.println("UNKNOWN DataType");
+                System.out.printf("UNKNOWN DataType %s\n", dt);
                 break;
         }
         if (showBinary) {
@@ -229,6 +230,16 @@ public class logReader {
                 dr.hasRank() ? dr.getRank() : 0L,
                 dr.hasUuidMostSignificant() ? dr.getUuidMostSignificant() : 0L,
                 dr.hasUuidLeastSignificant() ? dr.getUuidLeastSignificant() : 0L);
+        if (entry.hasCheckpointEntryType()) {
+            System.out.format("Checkpoint type: %s, ID %s\n",
+                    entry.getCheckpointEntryType(),
+                    new UUID(entry.hasCheckpointIDMostSignificant()
+                             ? entry.getCheckpointIDMostSignificant()
+                             : 0L,
+                             entry.hasCheckpointIDLeastSignificant()
+                             ? entry.getCheckpointIDLeastSignificant()
+                             : 0L));
+        }
     }
 
     // Read and conditionally replace a record
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
@@ -1,0 +1,194 @@
+package org.corfudb.protocols.logprotocol;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.UnpooledByteBufAllocator;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+import org.corfudb.runtime.CorfuRuntime;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Object & serialization methods for in-stream checkpoint
+ * summarization of SMR object state.
+ */
+@ToString(callSuper = true)
+@NoArgsConstructor
+public class CheckpointEntry extends LogEntry {
+
+    @RequiredArgsConstructor
+    public enum CheckpointEntryType {
+        START(0),           // Mandatory: 1st record in checkpoint
+        CONTINUATION(1),    // Optional: 2nd through (n-1)th record
+        END(2);             // Mandatory: final record checkpoint
+
+        public final int type;
+
+        public byte asByte() {
+            return (byte) type;
+        }
+
+        public static final Map<Byte, CheckpointEntryType> typeMap =
+                Arrays.stream(CheckpointEntryType.values())
+                        .collect(Collectors.toMap(CheckpointEntryType::asByte, Function.identity()));
+    };
+
+    @RequiredArgsConstructor
+    public enum CheckpointDictKey {
+        START_TIME(0),
+        END_TIME(1),
+        START_LOG_ADDRESS(2),
+        ENTRY_COUNT(3),
+        BYTE_COUNT(4);
+
+        public final int type;
+
+        public byte asByte() {
+            return (byte) type;
+        }
+
+        public static final Map<Byte, CheckpointDictKey> typeMap =
+                Arrays.stream(CheckpointDictKey.values())
+                        .collect(Collectors.toMap(CheckpointDictKey::asByte, Function.identity()));
+    }
+
+    /** Type of entry
+     */
+    @Getter
+    CheckpointEntryType cpType;
+
+    /**
+     * Unique identifier for this checkpoint.  All entries
+     * for the same checkpoint state must use the same ID.
+     */
+    @Getter
+    UUID checkpointID;
+
+    /** Author/cause/trigger of this checkpoint
+     */
+    @Getter
+    String checkpointAuthorID;
+
+    /** Map of checkpoint metadata, see key constants above
+     */
+    @Getter
+    Map<CheckpointDictKey, String> dict;
+
+    /** Optional: SMREntry objects that contain SMR
+     *  object state of the stream that we're checkpointing.
+     *  May be present in any CheckpointEntryType, but typically
+     *  used by CONTINUATION entries.
+     */
+    @Getter
+    @Setter
+    MultiSMREntry smrEntries;
+
+    /** Byte count of smrEntries in serialized form, zero
+     *  if smrEntries.size() is zero or if value is unknown.
+     */
+    @Getter
+    int smrEntriesBytes = 0;
+
+    public CheckpointEntry(CheckpointEntryType type, String authorID, UUID checkpointID,
+                           Map<CheckpointDictKey,String> dict, MultiSMREntry smrEntries) {
+        super(LogEntryType.SMR);
+        this.cpType = type;
+        this.checkpointID = checkpointID;
+        this.checkpointAuthorID = authorID;
+        this.dict = dict;
+        this.smrEntries = smrEntries;
+    }
+
+    /**
+     * This function provides the remaining buffer. Child entries
+     * should initialize their contents based on the buffer.
+     *
+     * @param b The remaining buffer.
+     * @param rt The CorfuRuntime used by the SMR object.
+     * @return A CheckpointEntry.
+     */
+    @Override
+    void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
+        super.deserializeBuffer(b, rt);
+        cpType = CheckpointEntryType.typeMap.get(b.readByte());
+        checkpointID = new UUID(b.readLong(), b.readLong());
+        checkpointAuthorID = deserializeString(b);
+        dict = new HashMap<>();
+        short mapEntries = b.readShort();
+        for (short i = 0; i < mapEntries; i++) {
+            CheckpointDictKey k = CheckpointDictKey.typeMap.get(b.readByte());
+            String v = deserializeString(b);
+            dict.put(k, v);
+        }
+        smrEntries = null;
+        byte hasSMREntries = b.readByte();
+        if (hasSMREntries > 0) {
+            smrEntries = (MultiSMREntry) MultiSMREntry.deserialize(b, runtime);
+        }
+        smrEntriesBytes = b.readInt();
+    }
+
+    /**
+     * Serialize the given LogEntry into a given byte buffer.
+     *
+     * NOTE: This method has a side-effect of updating the
+     *       this.smrEntriesBytes field.
+     *
+     * @param b The buffer to serialize into.
+     */
+    @Override
+    public void serialize(ByteBuf b) {
+        super.serialize(b);
+        b.writeByte(cpType.asByte());
+        b.writeLong(checkpointID.getMostSignificantBits());
+        b.writeLong(checkpointID.getLeastSignificantBits());
+        serializeString(checkpointAuthorID, b);
+        b.writeShort(dict == null ? 0 : dict.size());
+        if (dict != null) {
+            dict.entrySet().stream()
+                    .forEach(x -> {
+                        b.writeByte(x.getKey().asByte());
+                        serializeString(x.getValue(), b);
+                    });
+        }
+        if (smrEntries != null) {
+            b.writeByte(1);
+            int byteStart = b.readableBytes();
+            smrEntries.serialize(b);
+            smrEntriesBytes = b.readableBytes() - byteStart;
+        } else {
+            b.writeShort(0);
+            smrEntriesBytes = 0;
+        }
+        b.writeInt(smrEntriesBytes);
+    }
+
+    /** Helper function to deserialize a String.
+     *
+     * @param b
+     * @return A String.
+     */
+    private String deserializeString(ByteBuf b) {
+        short len = b.readShort();
+        byte bytes[] = new byte[len];
+        b.readBytes(bytes, 0, len);
+        return new String(bytes);
+    }
+
+    /** Helper function to serialize a String.
+     *
+     * @param b
+     */
+    private void serializeString(String s, ByteBuf b) {
+        b.writeShort(s.length());
+        b.writeBytes(s.getBytes());
+    }
+}
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -3,12 +3,16 @@ package org.corfudb.protocols.wireprotocol;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import lombok.Getter;
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.LogEntry;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.util.serializer.Serializers;
 
 import java.util.EnumMap;
 import java.util.concurrent.atomic.AtomicReference;
+
+import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINT_ID;
+import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINT_TYPE;
 
 /**
  * Created by mwei on 8/15/16.
@@ -112,19 +116,25 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
         this.metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
     }
 
-    public LogData(final Object object) {
-        this.type = DataType.DATA;
-        this.data = null;
-        this.payload.set(object);
-        if (object instanceof LogEntry) {
-            ((LogEntry) object).setEntry(this);
+    public LogData(DataType type, final Object object) {
+        if (object instanceof ByteBuf) {
+            this.type = type;
+            this.data = byteArrayFromBuf((ByteBuf) object);
+            this.metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
+        } else {
+            this.type = type;
+            this.data = null;
+            this.payload.set(object);
+            if (object instanceof LogEntry) {
+                ((LogEntry) object).setEntry(this);
+            }
+            this.metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
+            if (object instanceof CheckpointEntry) {
+                CheckpointEntry cp = (CheckpointEntry) object;
+                setCheckpointType(cp.getCpType());
+                setCheckpointID(cp.getCheckpointID());
+            }
         }
-        this.metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
-    }
-    public LogData(final DataType type, final ByteBuf buf) {
-        this.type = type;
-        this.data = byteArrayFromBuf(buf);
-        this.metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
     }
 
     public byte[] byteArrayFromBuf(final ByteBuf buf) {
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -1,0 +1,226 @@
+package org.corfudb.runtime;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterators;
+import lombok.Getter;
+import lombok.Setter;
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
+import org.corfudb.protocols.logprotocol.MultiSMREntry;
+import org.corfudb.protocols.logprotocol.SMREntry;
+import org.corfudb.runtime.collections.SMRMap;
+import org.corfudb.runtime.object.transactions.AbstractTransactionalContext;
+import org.corfudb.runtime.object.transactions.TransactionalContext;
+import org.corfudb.runtime.view.StreamsView;
+import org.corfudb.runtime.view.stream.AbstractQueuedStreamView;
+import org.corfudb.runtime.view.stream.BackpointerStreamView;
+import org.corfudb.util.serializer.Serializers;
+
+import java.time.LocalDateTime;
+import java.util.*;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/** Checkpoint writer for SMRMaps: take a snapshot of the
+ *  object via TXBegin(), then dump the frozen object's
+ *  state into CheckpointEntry records into the object's
+ *  stream.
+ *
+ * TODO: Generalize to all SMR objects.
+ */
+public class CheckpointWriter {
+    /** Metadata to be stored in the CP's 'dict' map.
+     */
+    private UUID streamID;
+    private String author;
+    @Getter
+    private UUID checkpointID;
+    private LocalDateTime startTime;
+    private long startAddress, endAddress;
+    private long numEntries = 0, numBytes = 0;
+    Map<CheckpointEntry.CheckpointDictKey, String> mdKV = new HashMap<>();
+
+    /** Mutator lambda to change map key.  Typically used for
+     *  testing but could also be used for type conversion, etc.
+     */
+    @Getter
+    @Setter
+    Function<Object,Object> keyMutator = (x) -> x;
+
+    /** Mutator lambda to change map value.  Typically used for
+     *  testing but could also be used for type conversion, etc.
+     */
+    @Getter
+    @Setter
+    Function<Object,Object> valueMutator = (x) -> x;
+
+    /** Batch size: number of SMREntry in a single CONTINUATION.
+     */
+    @Getter
+    @Setter
+    private int batchSize = 50;
+
+    /** BiConsumer to run after every CheckpointEntry is appended
+     * to the stream
+     */
+    @Getter
+    @Setter
+    BiConsumer<CheckpointEntry,Long> postAppendFunc = (cp, l) -> {};
+
+    /** Local ref to the object's runtime.
+     */
+    private CorfuRuntime rt;
+
+    /** Local ref to the stream's view.
+     */
+    StreamsView sv;
+
+    /** Local ref to the object that we're dumping.
+     *  TODO: generalize to all SMR objects.
+     */
+    private SMRMap map;
+
+    public CheckpointWriter(CorfuRuntime rt, UUID streamID, String author, SMRMap map) {
+        this.rt = rt;
+        this.streamID = streamID;
+        this.author = author;
+        this.map = map;
+        checkpointID = UUID.randomUUID();
+        sv = rt.getStreamsView();
+    }
+
+    /** Static method for all steps necessary to append checkpoint
+     *  data for an SMRMap into its own stream.  An optimistic
+     *  read-only transaction is used to freeze the contents of
+     *  the map while the checkpoint entries are written to
+     *  the stream.
+     */
+
+    public List<Long> appendCheckpoint() {
+        return appendCheckpoint(cpw -> {});
+    }
+
+    /**
+     * @param setupWriter Lambda to transform the new CheckPointWriter
+     *                    object (e.g., change batch size, etc) prior
+     *                    to use.
+     * @return List of global addresses of all entries for this checkpoint.
+     */
+
+    public List<Long> appendCheckpoint(Consumer<CheckpointWriter> setupWriter) {
+        List<Long> addrs = new ArrayList<>();
+        setupWriter.accept(this);
+
+        rt.getObjectsView().TXBegin();
+        try {
+            addrs.add(startCheckpoint());
+            addrs.addAll(appendObjectState());
+            addrs.add(finishCheckpoint());
+            return addrs;
+        } finally {
+            rt.getObjectsView().TXAbort();
+        }
+    }
+
+    /** Append a checkpoint START record to this object's stream.
+     *
+     *  Corfu client transaction management, if desired, is the
+     *  caller's responsibility.
+     *
+     * @return Global log address of the START record.
+     */
+    public long startCheckpoint() {
+        startTime = LocalDateTime.now();
+        AbstractTransactionalContext context = TransactionalContext.getCurrentContext();
+        long txBeginGlobalAddress = context.getSnapshotTimestamp();
+
+        this.mdKV.put(CheckpointEntry.CheckpointDictKey.START_TIME, startTime.toString());
+        this.mdKV.put(CheckpointEntry.CheckpointDictKey.START_LOG_ADDRESS, Long.toString(txBeginGlobalAddress));
+
+        ImmutableMap<CheckpointEntry.CheckpointDictKey,String> mdKV = ImmutableMap.copyOf(this.mdKV);
+        CheckpointEntry cp = new CheckpointEntry(CheckpointEntry.CheckpointEntryType.START,
+                author, checkpointID, mdKV, null);
+        startAddress = sv.append(Collections.singleton(streamID), cp, null);
+        postAppendFunc.accept(cp, startAddress);
+        return startAddress;
+    }
+
+    /** Append zero or more CONTINUATION records to this
+     *  object's stream.  Each will contain a fraction of
+     *  the state of the object that we're checkpointing
+     *  (up to batchSize items at a time).
+     *
+     *  Corfu client transaction management, if desired, is the
+     *  caller's responsibility.
+     *
+     *  The Iterators class appears to preserve the laziness
+     *  of Stream processing; we don't wish to use more
+     *  memory than strictly necessary to generate the
+     *  checkpoint.  NOTE: It would be even more useful if
+     *  the map had a lazy iterator: the eagerness of
+     *  map.keySet().stream() is not ideal, but at least
+     *  it should be much smaller than the entire map.
+     *
+     *  NOTE: The postAppendFunc lambda is executed in the
+     *  current thread context, i.e., inside of a Corfu
+     *  transaction, and that transaction will be *aborted*
+     *  at the end of this function.  Any Corfu data
+     *  modifying ops will be undone by the TXAbort().
+     *
+     * @return Stream of global log addresses of the
+     * CONTINUATION records written.
+     */
+    public List<Long> appendObjectState() {
+        ImmutableMap<CheckpointEntry.CheckpointDictKey,String> mdKV = ImmutableMap.copyOf(this.mdKV);
+        List<Long> continuationAddresses = new ArrayList<>();
+
+        Iterators.partition(map.keySet().stream()
+                .map(k -> {
+                    return new SMREntry("put",
+                            new Object[]{keyMutator.apply(k), valueMutator.apply(map.get(k))},
+                            Serializers.JSON);
+                }).iterator(), batchSize)
+                .forEachRemaining(entries -> {
+                    MultiSMREntry smrEntries = new MultiSMREntry();
+                    for (int i = 0; i < ((List) entries).size(); i++) {
+                        smrEntries.addTo((SMREntry) ((List) entries).get(i));
+                    }
+                    CheckpointEntry cp = new CheckpointEntry(CheckpointEntry.CheckpointEntryType.CONTINUATION,
+                            author, checkpointID, mdKV, smrEntries);
+                    long pos = sv.append(Collections.singleton(streamID), cp, null);
+
+                    postAppendFunc.accept(cp, pos);
+                    continuationAddresses.add(pos);
+
+                    numEntries++;
+                    // CheckpointEntry::serialize() has a side-effect we use
+                    // for an accurate count of serialized bytes of SRMEntries.
+                    numBytes += cp.getSmrEntriesBytes();
+                });
+        return continuationAddresses;
+    }
+
+    /** Append a checkpoint END record to this object's stream.
+     *
+     *  Corfu client transaction management, if desired, is the
+     *  caller's responsibility.
+     *
+     * @return Global log address of the END record.
+     */
+
+    public long finishCheckpoint() {
+        LocalDateTime endTime = LocalDateTime.now();
+        mdKV.put(CheckpointEntry.CheckpointDictKey.END_TIME, endTime.toString());
+        numEntries++;
+        numBytes++;
+        mdKV.put(CheckpointEntry.CheckpointDictKey.ENTRY_COUNT, Long.toString(numEntries));
+        mdKV.put(CheckpointEntry.CheckpointDictKey.BYTE_COUNT, Long.toString(numBytes));
+
+        CheckpointEntry cp = new CheckpointEntry(CheckpointEntry.CheckpointEntryType.END,
+                author, checkpointID, mdKV, null);
+        endAddress = sv.append(Collections.singleton(streamID), cp, null);
+
+        postAppendFunc.accept(cp, endAddress);
+        return endAddress;
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
@@ -1,5 +1,6 @@
 package org.corfudb.runtime.object;
 
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.ISMRConsumable;
 import org.corfudb.protocols.logprotocol.SMREntry;
 import org.corfudb.protocols.wireprotocol.DataType;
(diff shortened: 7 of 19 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **7e31b90**.

 [Pull Request #601 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/fv0xAgDv0945yzq9rEjvPvjOUmMcdME2)

slfritchie:
One decision that I'd made early on & not rejected during code walkthrough was keeping `ILogData` for CHECKPOINT entries instead of Long log addresses like DATA entries.  After fiddling around with the debugger, I now think that isn't worth the extra memory use.  Commit d20f1d1 changes CHECKPOINT entry tracking to mirror DATA tracking.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d20f1d1**.

 [Pull Request #601 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ki7Vly1ZDZdRdeTDfncUX6vtnfIxBg66)

logReader/src/main/java/org/corfudb/logReader/logReader.java
```diff
@@ -217,8 +217,11 @@ public final void printLogEntry(final LogEntry entry, final boolean showBinary)
             case TRIMMED:
                 System.out.println("DataType: TRIMMED");
                 break;
+            case CHECKPOINT:
+                System.out.println("DataType: CHECKPOINT");
+                break;
             default:
-                System.out.println("UNKNOWN DataType");
+                System.out.printf("UNKNOWN DataType %s\n", dt.toString());
```
kspirov: remove toString() - it is not necessary, and if the value is null, your code will throw NPE.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -98,9 +120,9 @@ public void seek(long globalAddress) {
     @Override
     public Stream<SMREntry> streamUpTo(long maxGlobal) {
         return streamView.streamUpTo(maxGlobal)
-                .filter(m -> m.getType() == DataType.DATA)
-                .filter(m -> m.getPayload(runtime) instanceof ISMRConsumable)
-                .map(logData -> ((ISMRConsumable)logData.getPayload(runtime)).getSMRUpdates(streamView.getID()))
+                .filter(m -> m.getType() == DataType.DATA || m.getType() == DataType.CHECKPOINT)
```
kspirov: Scott, you have multiple ifs like this (dataType==DATA || DataTppe==CHECKPOINT). Better add in org.corfudb.protocols.wireprotocol.LogData similar attribute like "metadataAware",  for example "contentAware" or the oposite flag "empty" and replace all ifs like this with if (data.isContentArare()) or if (!data.isEmpty()).   Also inspect once again all ifs where you have if (dataType==DATA) too see do you need to do the replace only in the 3-4 places from your review, or to more places.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -0,0 +1,421 @@
...
+     *    the keys written by step #1.
+     * 3. Put a 3rd key into "m".
+     * 4. The current runtime should see all original values of
+     *    the three put() keys.
+     * 5. Open a new runtime and a new map "m2" on the same stream.
+     *    Verify via get() that we cannot see the first two keys,
+     *    we see the snapshot's keys 7 & 8, and we can also see
+     *    key 3.
+     *
+     * This is not correct map behavior: keys shouldn't be lost like
```
kspirov: This approach has a point, but I would suggest the following:
Rename this test to smokeTestWrongBehaviour(), and copy/paste it to smokeTestCorrectBehaviour() + add @Ignore to the second test in order to remind us to fix this (test driven approach, you just state the correct contract)

Then in the future, when you fix it, you remove the @Ignore and modify the wrongBehaviour test to live a different side effect (by flag?)

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,232 @@
...
+            String v = deserializeString(b);
+            dict.put(k, v);
+        }
+        smrEntries = null;
+        int items = b.readShort();
+        if (items > 0) {
+            smrEntries = new SMREntry[items];
+            for (int i = 0; i < items; i++) {
+                int len = b.readInt();
+                ByteBuf rBuf = PooledByteBufAllocator.DEFAULT.buffer();
```
Maithem: please use an unpooled allocator

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,232 @@
...
+                           Map<String,String> dict, SMREntry[] smrEntries) {
+        super(LogEntryType.CHECKPOINT);
+        this.cpType = type;
+        this.checkpointID = checkpointID;
+        this.checkpointAuthorID = authorID;
+        this.dict = dict;
+        this.smrEntries = smrEntries;
+    }
+
+    static final Map<Byte, CheckpointEntryType> typeMap =
```
Maithem: move this map into the enum definition and implement valueOf

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -268,7 +301,52 @@ protected boolean fillReadQueue(final long maxGlobal,
...
+                                         ILogData currentEntry, CheckpointEntry cpEntry,
+                                         boolean considerCheckpoint, long currentRead) {
+        if (context.checkpointSuccessID == null &&
+                cpEntry.getCpType().equals(CheckpointEntry.CheckpointEntryType.END)) {
+            log.trace("Checkpoint: address {} found END, id {} author {}",
+                    currentRead, cpEntry.getCheckpointID(), cpEntry.getCheckpointAuthorID());
+            if (considerCheckpoint) {
+                context.checkpointSuccessID = cpEntry.getCheckpointID();
+                context.checkpointSuccessNumEntries = 1L;
+                context.checkpointSuccessEstBytes = (long) currentEntry.getSizeEstimate();
```
Maithem: Note that getSizeEstimate will return 1, once the entry has be deserialized, so you'll be reporting incorrect checkpoint size numbers.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,232 @@
...
+        b.writeShort(s.length());
+        b.writeBytes(s.getBytes());
+    }
+
+    /**
+     * Hex dump readable contents of ByteBuf to stdout.
+     *
+     * @param b ByteBuf with readable bytes available.
+     */
+    public static void dump(ByteBuf b) {
```
Maithem: Do these methods really belong here? maybe move them to utils?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+    @Setter
+    BiConsumer<CheckpointEntry,Long> postAppendFunc = (cp, l) -> {};
+
+    /** Local ref to the object's runtime.
+     */
+    private CorfuRuntime rt;
+
+    /** Local ref to the stream's view.
+     */
+    BackpointerStreamView sv;
```
Maithem: any reason why you are not using the interface instead? also why not just use the runtime's streamsview?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+     * @param author String to identify the author of this checkpoint
+     * @param map SMRMap do dump into this checkpoint
+     * @param setupWriter Lambda to transform the new CheckPointWriter
+     *                    object (e.g., change batch size, etc) prior
+     *                    to use.
+     * @return List of global addresses of all entries for this checkpoint.
+     */
+
+    public static List<Long> appendCheckpoint(CorfuRuntime rt, UUID streamID, String author, SMRMap map,
+                                              Consumer<CheckpointWriter> setupWriter) {
```
Maithem: This is static and it uses all the helper methods while still mutating the class members (i.e. mdKV). Can you either remove the mutable class members or make this a nonstatic method ?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+     * CONTINUATION records written.
+     */
+    public List<Long> appendObjectState() {
+        ImmutableMap<String,String> mdKV = ImmutableMap.copyOf(this.mdKV);
+        List<Long> continuationAddresses = new ArrayList<>();
+
+        try {
+            Iterators.partition(map.keySet().stream()
+                    .map(k -> {
+                        return new SMREntry("put",
```
Maithem: Maybe you want to use putAll instead ?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+                        long pos = sv.append(cp, null, null);
+                        postAppendFunc.accept(cp, pos);
+                        continuationAddresses.add(pos);
+                        numEntries++;
+                        // CheckpointEntry::serialize() has a side-effect we use
+                        // for an accurate count of serialized bytes of SRMEntries.
+                        numBytes += cp.getSmrEntriesBytes();
+                    });
+        } finally {
+            rt.getObjectsView().TXAbort();
```
Maithem: you use both TXAbort and TXEnd is there a reason why you do this?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+     * @param streamID StreamID for this map's underlying stream
+     * @param author String to identify the author of this checkpoint
+     * @param map SMRMap do dump into this checkpoint
+     * @param setupWriter Lambda to transform the new CheckPointWriter
+     *                    object (e.g., change batch size, etc) prior
+     *                    to use.
+     * @return List of global addresses of all entries for this checkpoint.
+     */
+
+    public static List<Long> appendCheckpoint(CorfuRuntime rt, UUID streamID, String author, SMRMap map,
```
Maithem: This should also consume a time stamp to set the proper snapshot. You can't just run this on multiple streams sequentially and then do a prefix trim. That will a result in an inconsistent cut.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
slfritchie: If checkpoints work correctly, they're invisible at nearly every layer of the system ... and that's a problem for testing.  I want to make it very clear to the reader that this test is doing something weird, and it's intentional.  This test case won't evolve over time, it should stay weird.

logReader/src/main/java/org/corfudb/logReader/logReader.java
slfritchie: Addressed by 09b3b82

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
slfritchie: Ah, I see, thanks.  Addressed by 09b3b82

format/proto/types.proto
```diff
@@ -6,6 +6,7 @@ enum DataType {
     HOLE = 2;
     TRIMMED = 3;
     PROPOSAL = 4;
+    CHECKPOINT = 5;
```
slfritchie: Hmm, no. I think I'll move some of the checkpoint KV dictionary items to the top-level metadata instead.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,232 @@
+package org.corfudb.protocols.logprotocol;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.PooledByteBufAllocator;
+import lombok.*;
```
slfritchie: IntelliJ was "helpful".  Will undo.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Object & serialization methods for in-stream checkpoint
+ * summarization of SMR object state.
+ */
+@ToString(callSuper = true)
+@NoArgsConstructor
+public class CheckpointEntry extends LogEntry {
```
slfritchie: Hm, it's CheckpointWriter that's specific to `SMRMap`.  This class doesn't have SMRMap details in it.  ... Hm, `CheckpointWriter` already has a top-level comment that specifically says SMRMaps are supported and has a TODO for general SMR objects support.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,232 @@
...
+    @Getter
+    Map<String, String> dict;
+
+    /** Optional: array of SMREntry objects that contains
+     *  SMR object state of the stream that we're checkpointing.
+     *  May be present in any CheckpointEntryType, but typically
+     *  used by CONTINUATION entries.
+     */
+    @Getter @Setter
+    SMREntry[] smrEntries = null;
```
slfritchie: Will fix.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
slfritchie: Will do.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * This function provides the remaining buffer. Child entries
+     * should initialize their contents based on the buffer.
+     *
+     * @param b The remaining buffer.
+     * @param rt The CorfuRuntime used by the SMR object.
+     * @return A CheckpointEntry.
+     */
+    @Override
+    void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
```
slfritchie: Hrm, I'd prefer not to.  The `SMREntry` items inside a `CheckpointEntry` have tangle problems with entry & runtime references that need fixing up in various places, similar to https://github.com/CorfuDB/CorfuDB/blob/6a10e5c/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java#L45-L46 , and I'd like to avoid needing yet another case of doing that.  I hope that moving key checkpoint metadata to the top-level metadata map will be sufficient; if not, then I'll return here.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
slfritchie: Oops, sorry, will fix.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
slfritchie: Will do.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
slfritchie: Will switch to using cpEntry.getSmrEntriesBytes().

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
slfritchie: Will do, including the cascade through a different `sv.append()` method & args.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
slfritchie: Hmm, my thinking was that you'd put the upper limit on a prefix trim as the min({L || for all streams named S, L = log position of S's checkpoint start}).

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
slfritchie: I'll make it nonstatic, thanks.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
slfritchie: Dahlia asked the same question, and my reply was "we need either Nx1 or 1xN of key-value pairs".  The checkpoint data is only applied once at most, is it really worth optimizing here?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
slfritchie: I will rearrange the transaction handling here.  The general absence of try/catch is intentional ... but I'll address further that in a PR-wide comment.

slfritchie:
I made changes for most of Maithem's comments in a batch at cc01bdde5bcee8a043cdad24e579000843f550f8.  The remaining item is moving checkpoint related critical metadata to the LogData metadataMap.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,209 @@
...
+
+        static final Map<Byte, CheckpointEntryType> typeMap =
+                Arrays.stream(CheckpointEntryType.values())
+                        .collect(Collectors.toMap(CheckpointEntryType::asByte, Function.identity()));
+    };
+
+    /**
+     * Constants for use as keys in our 'dict' map.
+     */
+    public static String START_TIME = "Start time";
```
no2chem: Can you make this an Enum? Seems error prone and wasteful to serialize a bunch of strings...

format/proto/types.proto
no2chem: I don't think Checkpoint should be a data type. It should just be a field in metadata

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -89,7 +90,13 @@ public ILogData load(Long aLong) throws Exception {
      */
     public void write(IToken token, Object data)
         throws OverwriteException {
-        final ILogData ld = new LogData(data);
+        final ILogData ld;
+        if (data instanceof CheckpointEntry) {
+            ld = new LogData(DataType.CHECKPOINT, data);
```
no2chem: Yeah, this is not good, we can't create a new DataType for every Entry type. Make checkpoint a metadata field, and we can check the metadata field.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -105,13 +112,20 @@ public void write(IToken token, Object data)
...
                         .getReplicationProtocol(runtime)
                         .write(l, ld);
-
             return null;
          });
 
-
         // Cache the successful write
         if (!runtime.isCacheDisabled()) {
+            if (ld.getType() == DataType.CHECKPOINT) {
```
no2chem: This could be avoided by handling the entry population in getPayload?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -64,27 +66,39 @@ protected void addToResolvedQueue(QueuedStreamContext context,
     @Override
     protected ILogData getNextEntry(QueuedStreamContext context,
                                     long maxGlobal) {
+        NavigableSet<Long> pollFrom;
```
no2chem: what does pollfrom contain?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **dafd05c**.

 [Pull Request #601 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/jWFjInk4LqOpQyvZ6xcHYR0oFCBEi7Gm)

Maithem:
Is this ready to be reviewed again ?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
slfritchie: See 9 lines down, line 78: "Is there checkpoint data to consume first?"

I added that variable to avoid checking repeatedly if context.readCpQueue is empty.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
slfritchie: Maithem asked about the types in that map also.  I use <String,String> to simplify serialization.  Key length relative to the bulk data contained elsewhere in these records is pretty miniscule.

slfritchie:
@Maithem Yes, I believe so.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **7e2f896**.

 [Pull Request #601 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/jkkJ1c5PjoUF0TC21JnQHmRjCN8q9vPg)

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -19,7 +23,7 @@
     public static final LogData HOLE = new LogData(DataType.HOLE);
 
     @Getter
-    final DataType type;
+    DataType type;
```
no2chem: is there some reason this isn't final anymore?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -88,7 +92,7 @@ public int getSizeEstimate() {
     }
 
     @Getter
-    final EnumMap<LogUnitMetadataType, Object> metadataMap;
+    EnumMap<LogUnitMetadataType, Object> metadataMap;
```
no2chem: why is this not final either?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -113,18 +117,32 @@ public LogData(DataType type) {
...
+        if (object instanceof ByteBuf) {
+            this.type = type;
+            this.data = byteArrayFromBuf((ByteBuf) object);
+            this.metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
+        } else {
+            commonLogData(type, object);
+        }
+    }
+
+    public void commonLogData(DataType type, final Object object) {
```
no2chem: please use constructor overloading and not this extra method

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
no2chem: well, the size isn't the main issue, it's that simple mistakes like messing up on the case or using UTF-16 instead of UTF-8 will break things. Change it to a enum which writes a byte, there is no reason for this to be a string

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -41,11 +44,27 @@ public StreamViewSMRAdapter(CorfuRuntime runtime,
...
+            return ((ISMRConsumable) logData.getPayload(runtime)).getSMRUpdates(streamView.getID());
+        }
+    }
+
     public List<SMREntry> remainingUpTo(long maxGlobal) {
         return streamView.remainingUpTo(maxGlobal).stream()
                 .filter(m -> m.getType() == DataType.DATA)
-                .filter(m -> m.getPayload(runtime) instanceof ISMRConsumable)
-                .map(logData -> ((ISMRConsumable)logData.getPayload(runtime)).getSMRUpdates(streamView.getID()))
+                .filter(m -> m.getPayload(runtime) instanceof ISMRConsumable || m.hasCheckpointMetadata())
```
no2chem: shouldn't this be done at the stream layer rather than the SMRStream layer? (i.e., the stream layer converts checkpoints to SMREntries which get fed into this adapter?)

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -105,13 +107,20 @@ public void write(IToken token, Object data)
...
                         .getReplicationProtocol(runtime)
                         .write(l, ld);
-
             return null;
          });
 
-
         // Cache the successful write
         if (!runtime.isCacheDisabled()) {
+            if (ld.hasCheckpointMetadata()) {
```
no2chem: I don't think this is necessary (and probably the wrong place to put if it is), the logdata should not have been serialized at this point, so any entry in the cache should reflect the correct in memory state already.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
no2chem: maybe add the comment here, then

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
slfritchie: Yes, as a result of splitting the LogData constructor.  That split isn't needed anymore, so this & several other comments about LogData.java are addressed as I roll back the split.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
slfritchie: Will do.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
slfritchie: I'd wanted that dictionary to be usable for use like adding metadata annotation via Clojure.  Enum style means no ad hoc keys.  So, we'll see if ad hoc is needed later.

slfritchie:
@no2chem @Maithem @dahliamalkhi @kspirov I've addressed Michael's 2nd round (I hope) in ba6af28.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
slfritchie: If I don't do it, I get this explosion of NPEs.  The cached entry needs fixing up, here or elsewhere, prior to use after pulling out of the cache.

    smoke1Test                                                  [FAIL - RuntimeException:L101]
    smoke2Test                                                  [FAIL - RuntimeException:L176]
    checkpointWriterInterleavedTest                             [FAIL - RuntimeException:L350]
    checkpointWriterTest                                        [FAIL - RuntimeException:L237]
    Tests run: 4, Failures: 0, Errors: 4, Skipped: 0, Time elapsed: 2.092 sec <<< FAILURE! - in org.corfudb.runtime.checkpoint.CheckpointSmokeTest
    smoke1Test(org.corfudb.runtime.checkpoint.CheckpointSmokeTest)  Time elapsed: 1.06 sec  <<< ERROR!
    java.lang.RuntimeException: java.lang.NullPointerException
    	at org.corfudb.runtime.object.VersionLockedObject.lambda$syncStreamUnsafe$3(VersionLockedObject.java:532)
    	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
    	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1374)
    	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
    	at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:270)
    	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
    	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
    	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
    	at org.corfudb.runtime.view.stream.StreamSpliterator.tryAdvance(StreamSpliterator.java:63)
    	at java.util.Spliterator.forEachRemaining(Spliterator.java:326)
    	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
    	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
    	at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
    	at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
    	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    	at java.util.stream.ReferencePipeline.forEachOrdered(ReferencePipeline.java:423)
    	at org.corfudb.runtime.object.VersionLockedObject.syncStreamUnsafe(VersionLockedObject.java:517)
    	at org.corfudb.runtime.object.VersionLockedObject.syncObjectUnsafe(VersionLockedObject.java:284)
    	at org.corfudb.runtime.object.CorfuCompileProxy.lambda$accessInner$1(CorfuCompileProxy.java:137)
    	at org.corfudb.runtime.object.VersionLockedObject.access(VersionLockedObject.java:184)
    	at org.corfudb.runtime.object.CorfuCompileProxy.accessInner(CorfuCompileProxy.java:136)
    	at org.corfudb.runtime.object.CorfuCompileProxy.access(CorfuCompileProxy.java:118)
    	at org.corfudb.runtime.collections.SMRMap$CORFUSMR.get(SMRMap$CORFUSMR.java:183)
    	at org.corfudb.runtime.checkpoint.CheckpointSmokeTest.smoke1Test(CheckpointSmokeTest.java:101)
    	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    	at java.lang.reflect.Method.invoke(Method.java:498)
    	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
    	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
    	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
    	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
    	at org.corfudb.AbstractCorfuTest$1$1.evaluate(AbstractCorfuTest.java:76)
    	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
    	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
    	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
    	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
    	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
    	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
    	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
    	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
    	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
    	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
    	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:367)
    	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:274)
    	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:238)
    	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:161)
    	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:290)
    	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:242)
    	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:121)
    Caused by: java.lang.NullPointerException
    	at org.corfudb.runtime.object.VersionLockedObject.lambda$syncStreamUnsafe$3(VersionLockedObject.java:523)
    	... 51 more
    
    Results :
    
    Tests in error: 
      CheckpointSmokeTest.checkpointWriterInterleavedTest:350 » Runtime java.lang.Nu...
      CheckpointSmokeTest.checkpointWriterTest:237 » Runtime java.lang.NullPointerEx...
      CheckpointSmokeTest.smoke1Test:101 » Runtime java.lang.NullPointerException
      CheckpointSmokeTest.smoke2Test:176->lambda$smoke2Test$0:151 » Runtime java.lan...
    
    Tests run: 4, Failures: 0, Errors: 4, Skipped: 0

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
slfritchie: Oops, sorry, I missed this question.  I don't have a strong opinion, but the opinion is that the checkpoint data is an SMR thing and doesn't make sense strictly at the stream level.  How do you remove redundant history (i.e. ignore/summarize/aggregate earlier entries) from a stream if you have no structure or schema to the stream?

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -41,11 +44,27 @@ public StreamViewSMRAdapter(CorfuRuntime runtime,
         this.streamView = streamView;
     }
 
+    private List<SMREntry> dataAndCheckpointMapper(ILogData logData) {
+        if (logData.hasCheckpointMetadata()) {
+            // This is a CHECKPOINT record.  Extract the SMREntries, if any.
+            CheckpointEntry cp = (CheckpointEntry) logData.getPayload(runtime);
```
no2chem: I see the error about the NPE. Why not restore the entry here instead of modifyingAddressSpaceView?

so

``cp.getSMREntries().forEach(x -> x.setEntry(cp.getEntry()))```

Maithem:
@slfritchie @dahliamalkhi @no2chem 


As a rule, a good algorithm is a result of repeated effort and rework. - Anany Levitin

Please allow me to suggest a variation of the algorithm in this patch, which is based on an insight from @medhavidhawan. The main idea is to allow overlapping checkpoints, but only allow a single prefix trim.

A global checkpoint is a collection of checkpoints of all the streams in the system. A prefix trim decision can be computed from a global checkpoint, but not every trim decision should be executed, because that can cause data loss. To illustrate such case, consider the following clients (i.e. C1 and C2) creating global checkpoints in parallel:

t:     0      1     2     3     4     5     6
C1:         CS  CC  CC  CC  CC  CE
C2:                        CS  CC  CE

CS: checkpoint start
CC: checkpoint continuation
CE: checkpoint end

In this scenario C2 starts checkpointing after C1, but ends before C1. The computed prefix trim for C2 would be [0, C2:CS). On the other hand, C1 would compute a prefix trim of [0, C1:CS). When both clients attempt to trim, only C2 will succeed, C1 will fail because it tries to trim an address that has already been trimmed by C2. The logging unit will return an AlreadyTrimmedException to C1. In this case, C1 knows that its checkpoint is not valid and thus it won't update the Streams Registry with the address of its checkpoint, but C2 will. As a consequence, The stream registry will always point to an address that contains a valid checkpoint.

Unlike this patch, a client trying to restore a stream would be able to directly read C2's checkpoint (i.e a valid checkpoint) using the stream registry, as opposed to discovering the checkpoint during backward traversal by assuming that the first CE record it finds corresponds to a valid checkpoint (i.e. a client discovers C1:CE before C2:CE, but C1's checkpoint is invalid, so the stream cannot be restored from that particular checkpoint).

In conclusion, you can think of this mechanism as optimistic checkpointing, where multiple checkpoints can be computed concurrently, but only one checkpoint can be "committed". The decision to "commit" a checkpoint is serialized through the logging unit, so explicit mutual exclusion is not required, thereby simplifying the design by avoiding lock leasing mechanisms to ensure non-overlapping checkpoints. 

An additional note on a related topic: Checkpointing should be based on space reclamation thresholds and not time. Essentially, checkpointing is analogous to writing the coalesced state of all the streams back again to the log. If time based checkpointing is used, then there can be cases where the coalesced state doesn't change, but it is rewritten anyways, therefore generating unnecessary load on the system.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
slfritchie: !@#$! GH notifications aren't being delivered to me, sorry about the delay today.  I've moved ref fixup to `StreamViewSMRAdapter::dataAndCheckpointMapper` in commit d1d019e.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d1d019e**.

 [Pull Request #601 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/5BlRZDBwx2femxPq5ZIPJKtJ8PkpZqcx)

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+                            Serializers.JSON);
+                }).iterator(), batchSize)
+                .forEachRemaining(entries -> {
+                    // Convert Object[] to SMREntry[], combined with byte count.
+                    // java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Lorg.corfudb.protocols.logprotocol.SMREntry;
+                    int numEntries = ((List) entries).size();
+                    SMREntry e[] = new SMREntry[numEntries];
+                    for (int i = 0; i < numEntries; i++) {
+                        e[i] = (SMREntry) ((List) entries).get(i);
+                    }
```
dahliamalkhi: I feel like this should turn into putAll and also code could be java-fied down the line; but let's not wait for this, just remember to get back to it.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -27,17 +27,17 @@
  * access() method. Likewise, if a Corfu object's method is a Mutator or Accessor-Mutator, it invokes the
  * proxy's logUpdate() method.
  *
- * Within transactional context, these methods invoke the transacationalContect accessor/mutator helper.
+ * Within transactional context, these methods invoke the transactionalContext accessor/valueMutator helper.
```
dahliamalkhi: actually, mutator, not valueMutator

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -103,60 +116,67 @@ protected ILogData getNextEntry(QueuedStreamContext context,
...
-        if (!fillReadQueue(maxGlobal, context)) {
+        if (readQueueIsEmpty) {
             return Collections.emptyList();
         }
 
-        // If the lowest element is greater than maxGlobal, there's nothing
-        // to return.
-        if (context.readQueue.first() > maxGlobal) {
-            return Collections.emptyList();
+        if (!context.readQueue.isEmpty() && context.readQueue.first() > maxGlobal) {
```
dahliamalkhi: this test seems erroneous: If readQueue.isEmpty() returns true, then we want if to succeed, no?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -727,6 +729,11 @@ LogEntry getLogEntry(long address, LogData entry) {
         if (rank.isPresent()) {
             logEntryBuilder.setRank(rank.get());
         }
+        if (entry.hasCheckpointMetadata()) {
+            logEntryBuilder.setCheckpointEntryType(Types.CheckpointEntryType.forNumber(entry.getCheckpointType().ordinal()));
```
slfritchie: Will do

logReader/src/main/java/org/corfudb/logReader/logReader.java
```diff
@@ -231,7 +232,12 @@ public final void printLogEntry(final LogEntry entry, final boolean showBinary)
                 dr.hasUuidLeastSignificant() ? dr.getUuidLeastSignificant() : 0L);
         System.out.format("Commit: ");
         System.out.println(entry.getCommit());
-
+        if (entry.hasCheckpointEntryType()) {
+            System.out.format("Checkpoint type: %s, ID %s\n",
+                    entry.getCheckpointEntryType(),
+                    new UUID(entry.hasCheckpointIDMostSignificant() ? entry.getCheckpointIDMostSignificant() : 0L,
```
slfritchie: Will do.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,217 @@
...
+     * @param rt The CorfuRuntime used by the SMR object.
+     * @return A CheckpointEntry.
+     */
+    @Override
+    void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
+        super.deserializeBuffer(b, rt);
+        cpType = CheckpointEntryType.typeMap.get(b.readByte());
+        long cpidMSB = b.readLong();
+        long cpidLSB = b.readLong();
+        checkpointID = new UUID(cpidMSB, cpidLSB);
```
slfritchie: Will do.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -0,0 +1,217 @@
...
+        dict = new HashMap<>();
+        short mapEntries = b.readShort();
+        for (short i = 0; i < mapEntries; i++) {
+            CheckpointDictKey k = CheckpointDictKey.typeMap.get(b.readByte());
+            String v = deserializeString(b);
+            dict.put(k, v);
+        }
+        smrEntries = null;
+        int items = b.readShort();
+        if (items > 0) {
```
slfritchie: 1 & 2: because I write C flavored Java at first draft.  I'll look into it.

3: The comment says where & why the error would happen but omits the NullPointerException detail.  Commit d1d019e makes this unnecessary now, I believe.  I'll remove it.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+     */
+    public long startCheckpoint() {
+        startTime = LocalDateTime.now();
+        AbstractTransactionalContext context = TransactionalContext.getCurrentContext();
+        long txBeginGlobalAddress = context.getSnapshotTimestamp();
+
+        this.mdKV.put(CheckpointEntry.CheckpointDictKey.START_TIME, startTime.toString());
+        this.mdKV.put(CheckpointEntry.CheckpointDictKey.START_LOG_ADDRESS, Long.toString(txBeginGlobalAddress));
+
+        ImmutableMap<CheckpointEntry.CheckpointDictKey,String> mdKV = ImmutableMap.copyOf(this.mdKV);
```
slfritchie: 1. Will do.

2. I was seeing racy buggy results far down the road, if the dictionary was mutable.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -0,0 +1,229 @@
...
+     */
+    private SMRMap map;
+
+    public CheckpointWriter(CorfuRuntime rt, UUID streamID, String author, SMRMap map) {
+        this.rt = rt;
+        this.streamID = streamID;
+        this.author = author;
+        this.map = map;
+        checkpointID = UUID.randomUUID();
+        // sv = new BackpointerStreamView(rt, streamID);
```
slfritchie: Yup

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
slfritchie: Hm, no sure how that snuck in, but I'll undo it.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
slfritchie: `readSet` may contain log addresses for checkpoint data.  I'll move the `if (readQueueIsEmpty)` statement a bit earlier to help clarify.

slfritchie:
Monday morning status: approval from Michael, not sure if @dahliamalkhi is finished or not.  Still waiting for opinion from Ragnar & Robert.

slfritchie:
I've done a final squash & rebase onto today's master branch.  The conflicts in IMetadata.java and logReader.java were small.  My local tests passed as well as 1 of 2 Travis-CI test runs https://travis-ci.org/CorfuDB/CorfuDB/builds/234998916?utm_source=github_status&utm_medium=notification.  the other Travis-CI test appears to be hanging in the `testRandomizedRecovery` test. https://travis-ci.org/CorfuDB/CorfuDB/builds/234998866#L2333.  I'm mostly sure that `testRandomizedRecovery` sporadic failures are not related ... but I'll keep a closer eye on Travis-CI runs this week.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -0,0 +1,420 @@
...
+     * this.  But that's the point: the checkpoint *writer* is guilty
+     * of the bad behavior, and the unit test is looking exactly for
+     * bad behavior to confirm that the checkpoint mechanism is doing
+     * something extremely out-of-the-ordinary.
+     *
+     * @throws Exception
+     */
+    @Test
+    @SuppressWarnings("checkstyle:magicnumber")
```
no2chem: I missed this in the review (since apparently this was collapsed as it's too large), but this is NOT OKAY. Please remove it ASAP.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -0,0 +1,420 @@
...
+     *
+     * Again, this is not correct map behavior, same as the first
+     * smoke test.  We don't have code yet to generate "real"
+     * checkpoint data; still PoC stage.
+     *
+     * @throws Exception
+     */
+
+    @Test
+    @SuppressWarnings("checkstyle:magicnumber")
```
no2chem: Same here, do not suppress style checker warnings without justification

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -0,0 +1,420 @@
...
+
+        setRuntime();
+        Map<String, Long> m3 = instantiateMap(streamName);
+        testAssertions.accept(m3);
+    }
+
+    /** Test the CheckpointWriter class, part 1.
+     */
+    @Test
+    @SuppressWarnings("checkstyle:magicnumber")
```
no2chem: Same here, do not suppress style checker warnings without justification

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -0,0 +1,420 @@
...
+     *  we updated the map.
+     *
+     *  The one exception is for snapshot TXN with an address prior
+     *  to the 'startAddress' of the checkpoint; stream history is
+     *  destroyed (logically, not physically) by the CP, so we have
+     *  to use a different position 'history' for our assertion
+     *  check.
+     */
+    @Test
+    @SuppressWarnings("checkstyle:magicnumber")
```
no2chem: Same here, do not suppress style checker warnings without justification

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -0,0 +1,420 @@
...
+
+    private void writeCheckpointRecords(UUID streamId, String checkpointAuthor, UUID checkpointId,
+                                        Object[] objects)
+            throws Exception {
+        Runnable l = () -> {};
+        writeCheckpointRecords(streamId, checkpointAuthor, checkpointId, objects,
+                l, l, true, true, true);
+    }
+
+    @SuppressWarnings("checkstyle:magicnumber")
```
no2chem: Same here, do not suppress style checker warnings without justification

PR #599, opened 2017-05-09
Cleanup of runtime interface

no2chem:
This patch introduces a less verbose interface to the runtime, dropping
the "get" prefix and the "view" suffix. The old interface is retained
but marked for deprecation. In addition, the layout "future" has been
renamed to distinguish it from the layout view.

PR #607, opened 2017-05-16
restore interface

no2chem:
This PR is a resubmit of PR #599, which was causing performance regressions.


no2chem:
@Maithem to debug issues, since I don't have access to the performance testing framework locally.

corfudb-performance:
@no2chem results are not published because the test took too long (it could not finish for all workloads when a new one kicked in), performance drops to ~1 tx/sec and the number of samples is to high to complete quickly at this rate. However, I've generated a graph with the partial results:

[Dashboard PR # 607](https://snapshot.raintank.io/dashboard/snapshot/45Bel5GN2nBH9NA3bItdS9mvWpwfRKGt)

Issue #605, opened 2017-05-16
Layout populates "localhost" instead of IP address causing client connection redirect


kjames88:
Setup:

Client IP A
Server IP B

Client connects to server and retrieves layout with log server as "localhost"
Client attempts to use log server at address A ("localhost") which has no server

Roger confirms the -a switch is required to populate the server's address, otherwise the layout fails.  Observed failure is that server at B does not write to its log file and client hangs, whereas if a second server is started at A then writes accumulate in A's log file despite the client connect address set to B.  If server at B is killed, server at A log file is not written.

Maithem:
What does "never sees log size increase" mean? are you talking about the sequence number or log files size?

Also, -a tells the server which ip to bind to, so I'm not sure if this is really a bug. If you start the server on a single node that has two ips, how would it know which ip to bind to, if you dont specify it ?

Finally, if the server is started with -a, do you still observe the problem ?

kjames88:
Returning "localhost" to a remote host is a bug.  Receiving "localhost" from another host should cause an error.  Whether or not there is more than one NIC, the address selected and used should be valid for the host.

kjames88:
For the -a flag I'm going on Roger's comment.  I haven't verified it yet.

Maithem:
There are three cases:
- Both client and the server are binded to local host (same machine)
- A client is binded to something other than localhost, but the server is running on localhost (same machine)
- Both client and server are run on two different machines

How would you determine a "remote" connection for all three cases mentioned above?

I think this is a degenerate case. Also, you didn't answer the first question. I'm not sure if you are reporting one or more problems.

kjames88:
Localhost is not an IP address.  Don't send localhost to anyone as your IP address.  The degenerate case is having only one instance in a "distributed" system.  It's convenient to use localhost on the command line but it has to be translated into a real IP address.  If that can't be done then the address switch is not optional, and further "localhost" has to be an error case on the user input.

I already reworded the text to try to clarify the first question but it's really beside the point.

slfritchie:
[...]

PR #609, opened 2017-05-16
Bind to 127.0.0.1

Maithem:
When starting a server in single mode without specifying -a option
and a layout file, the server will automatically create a layout
for the localhost. As a consequence, the server will reply with
localhost addresses, this wont work for external clients. This
patch binds to 127.0.0.1 so that external connections cannot
connect to the server and thus it wont be able to reply with
localhost addresses.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
@@ -96,7 +96,7 @@ public class CorfuServer {
                     + " -l <path>, --log-path=<path>                                                           Set the path to the storage file for the log unit.\n"
                     + " -s, --single                                                                           Deploy a single-node configuration.\n"
                     + "                                                                                        The server will be bootstrapped with a simple one-unit layout.\n"
-                    + " -a <address>, --address=<address>                                                      IP address to advertise to external clients [default: localhost].\n"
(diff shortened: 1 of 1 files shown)


Maithem:
Addresses #605

Maithem:
@no2chem with this you are still allowed to run a single server that can be contacted externally. You just need to specify the -a, or a layout file. This problem happens when you don't specify a layout file, or pass -a parameter. So the server creates a default layout file. 

How do you bind to a non-localhost without specifying one explicitly? If you choose any non-local host to bind to, then how does your client connect to this address?

Note this change impacts the case where you start a single node without specifying -a and use the default layout file.

no2chem:
I know, but what I'm saying is that -s should pick a public address by default, not 127.0.0.1 or localhost. And of course this only controls the address we advertise, not the address we bind to (which I think is 0.0.0.0, and probably needs to be fixed).

If we advertise a public address, it shouldn't matter if the client uses localhost or the public address, it should just work, because a client picking up a client using localhost will use the public address, and a remote client will also use the public address. This seems to be more in line with what a user of a distributed system may naively expect.

Maithem:
Ok, but lets say you have two public addresses and you randomly to bind to anyone of them

no2chem:
but you don't, you would throw an error in that case and force the user to specify via -a

PR #617, opened 2017-05-17
Un-Lombok Current Master

Maithem:


kspirov:
Probably I am missing some context. Why you need to do this?

Maithem:
@kspirov We don't, this is a test pull request, please look at the comment in  https://github.com/CorfuDB/CorfuDB/pull/621

PR #621, opened 2017-05-18
Bad Construction

Maithem:


no2chem:
This is just a test PR to run against the perf framework, right?

Maithem:
@no2chem Right. Also, note that since "this" is leaked outside the constructor, some of the memory guarantees  on final fields might not always be honored. In other words, some parts of the runtime need to be re-organized

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **8a49e72**.

 [Pull Request #621 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/HOcT9PvTAvSIZS5x5BojN67P2bpnk1oF)

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -28,11 +34,94 @@
...
+
+    public static MetricRegistry getMetrics() {
+        return CorfuRuntime.metrics;
+    }
+
+    public CorfuRuntimeParameters getParameters() {
+        return this.parameters;
+    }
+
+    public LayoutView getLayoutView() {
```
kspirov: Annotate with Lombok @Getter all fields you return with this and remove these methods.  - CorfuRuntime is big enough

Maithem: Hi @kspirov , I am aware of lambok notation. This was a **test** pull request to see if lombok's annotation were generating code that affects our performance. 

Apparently, lombok's lazy evaluation getters were masking object construction issues that appear to drop down performance to 1tx/sec

PR #630, opened 2017-05-20
support poisoned transactional streams, which have conflicts against all updates

no2chem:
This PR introduces poisoned streams, which partially resolves an issue where
updates (such as clear()) were not causing conflicts even though they
conflicted agaist all operations.


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -458,20 +458,27 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
...
-                                    // and the stream id.
-                                    conflictToGlobalTailCache.put(
-                                            getConflictHashCode(txEntry
-                                                    .getKey(), conflictParam),
-                                            newTail - 1)));
+        if (req.getTxnResolution() != null) {
+            // Until we have per-stream conflict sets, any poisoned stream causes
+            // max wildcard to be updated
+            if (req.getTxnResolution().getPoisonedStreams().size() > 0) {
+                maxConflictWildcard = newTail - 1;
```
dahliamalkhi: This doesn't seem like the right fix, since maxConflictWildcard is taken into consideration only on conflictParameters that we -do not- have information on.

no2chem: Which is exactly what a entry like clear() is? What do you propose otherwise?

dahliamalkhi: but right now, this has no effect . every key which has an entry in the conflictKey map will override the wildcard.

no2chem: that's not correct, otherwise the test wouldn't past (and the wildcard wouldn't work right?) On the contrary, it seems that the conflictWildcard overrides the tailMap and any transaction below the wildcard will be rejected?

from L219 in SequencerServer:
`                    if (v == null && maxConflictWildcard > txSnapshotTimestamp ) {
                        log.warn("ABORT[{}] conflict-key[{}](WILDCARD ts={})", txInfo, conflictParam,
                                maxConflictWildcard);
                        conflictKey.set(conflictParam);
                        response.set(TokenType.TX_ABORT_CONFLICT);
                    }`

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **04a0580**.

 [Pull Request #630 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/wd10jp34V0fRG65BCtGjDDMPPg7sYdI5)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
dahliamalkhi: no.. the test v==null fails if we find a conflict key, hence we don't even look at the wildcard.

the wildcard says: if we don't have specific information about a conflict key, we have a crude hint instead. The hint stores a maximal "unknown" timestamp. The guarantee is:  no update to unknown conflict keys were made past this timestamp. 

Again, this timestamp is ignored when we have more specific information.

no2chem: okay, the drop the v== null, or we should implement per stream wildcards

dahliamalkhi:
should we close this PR and delete this branch?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **3eca16e**.

 [Pull Request #630 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/a3P6RY0vVzIeo4PD0mOGqtfFfb5t5axz)

Issue #620, opened 2017-05-17
Fast key state recovery using conflict keys


kjames88:
Enable runtime to recover individual keys from the final state in the log.  Conflict keys will be used (when available in the sequencer) to locate the last written log entry.  Log entry will be read and the final put to map extracted.  A partial map will be created to contain one or more keys read using this mechanism.

Maithem:
This is not clear. Can you provide an example and an explanation on how this would improve performance.

rogermichoud:
The goal here is for the client (maybe a special type of "thin client" or even a normal client) to be able to access a map entry without having to load the whole log history.

Let's say that client C starts, he wants to access an entry with key K of map M. He doesn't need the whole map, he just needs the value at key K. The sequencer potentially has the address of the latest update to key K. This would allow the client to directly get the latest value of key K. 

This has the potential to really improve startup. The implementation must be pretty careful, We need to be sure that the client try to load the entry before thinking that there is no entry for this key.

The original idea comes from Dahlia/Michael.

Maithem:
Hmm, seems like an optimization for a contrived use case. 

This assumes that the application would have "sparse" key only reads that also exhibits spatial locality. In other words, this seems like a fragile optimization that's not free. Also, the cost will be incurred in one of the system's bottlenecks (i.e. the sequencer).

Maybe gather more data on the access pattern before going ahead and trying to implement this?

Maithem:
Also, how would this improve startup time?

The problem with startup time (at least the issues that we have been seeing) is related to object initialization, not sparse key reads.

kjames88:
The presumed use case would involve a client that only needs access to a subset of a larger map.  In that case, the client could find the final state of a few keys and get started quickly.  If the entire map is needed at startup then this feature is probably not helpful.

no2chem:
This feature would greatly improve startup time, especially if sync could be done asynchronously.

It's not a contrived use case at all, it's not fragile (how?) by any means, and it doesn't add any more load to the sequencer than is already required for a linearized read.

What does spatial locality have to do with anything? This optimization greatly improves performance for any client only wishing to read the latest entry, if the conflict key is known.

rogermichoud:
I second that. The client could get the entries it needs right away while synchronizing in the background. In the current implementation, the client is blocked until the whole SMR entry history is loaded in.

Maithem:
@no2chem 
You make two claims, this scheme
1. Improves startup time (i.e. object initialization)
[...]

PR #631, opened 2017-05-21
direct initial implementation

no2chem:
This PR introduces the first part of the implementation of the issue
introduced in #620, which allows an accessor to be serviced directly
from the read of a single SMREntry instead of playing back the log.
This PR does not yet modify the behavior of the runtime, rather, it
adds several annotations and provides the information necessary
for CorfuCompileProxy to service an accessor directly from a single
SMREntry.

This PR introduces the DirectReadFunction annotation, which marks
a function which can take the arguments of an accessor and the
most recent mutator, and generate a return value for the accessor.
There can be one DirectReadFunction for every mutator+accessor
combination. MutatorAccessors cannot be directly serviced. In addition,
the accessor annotation also has a new field, directReadFunctions, which
specifies an array of named DirectReadFunctions.

The annotationProcessor generates a map of these functions and provides
them as a SMRMethodName -> DirectReadFunction map during the call to
ICorfuSMRProxy::access. If the runtime can determine that the conflictKeys
for the accessor and the mutator match, and that the mutator is the most
recent update to that conflictKey, the proxy can retrieve the directAccessFunction
and obtain the result as if it had played back the object by retrieving the 
directAccessFunction from the map and calling it with the arguments to the
accessor and mutator.

### DIFF
--- a/annotationProcessor/pom.xml
+++ b/annotationProcessor/pom.xml
@@ -11,6 +11,33 @@
 
     <artifactId>annotationProcessor</artifactId>
 
+    <profiles>
+        <profile>
+            <id>windows-profile</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+                <file>
+                    <exists>${JAVA_HOME}/lib/tools.jar</exists>
+                </file>
+            </activation>
+            <properties>
+                <toolsjar>${JAVA_HOME}/lib/tools.jar</toolsjar>
+            </properties>
+        </profile>
+        <profile>
+            <id>mac-profile</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+                <file>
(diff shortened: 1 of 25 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **bea973b**.

*** 4.92% transaction FAILURE rate for NonConflictingTx+Scan workload, 5 threads, Disk mode
*** 9.62% transaction FAILURE rate for NonConflictingTx+Scan workload, 10 threads, Disk mode
 
 [Pull Request #631 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/y2a7pJo5s4gxzI7bW9PBleuZ2JSXjLym)

dahliamalkhi:
@no2chem : shall we table this PR? (close and retain branch)

PR #632, opened 2017-05-23
Prefix Implementation

Maithem:
This patch introduces the ability to trim the global log up to
a certain address (i.e. [0, prefixAddress). A prefix trim
followed by a compact will deleted the underlaying files that
correspond to that prefix.


kspirov:
Good start!

PR #634, opened 2017-05-23
change readCache to final

no2chem:
This PR changes the readCache from being static (global) to a final instance
field, preventing possible odd behaviors from occurring and runtimes sharing
caches (which they shouldn't, as they should operate independently).

In addition, this patch now also contains a fix to the checkpoint entry type, and
since we have sort of a chicken-and-egg problem of unit tests and fixes
 (the unit tests are "fixed" by this PR, but the code is "fixed" by fixes to the 
checkpointing entries), they are all in this PR.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
@@ -99,7 +99,7 @@ public class CheckpointEntry extends LogEntry {
 
     public CheckpointEntry(CheckpointEntryType type, String authorID, UUID checkpointID,
                            Map<CheckpointDictKey,String> dict, MultiSMREntry smrEntries) {
-        super(LogEntryType.SMR);
+        super(LogEntryType.CHECKPOINT);
         this.cpType = type;
         this.checkpointID = checkpointID;
         this.checkpointAuthorID = authorID;
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/LogEntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/LogEntry.java
@@ -113,7 +113,8 @@ public class LogEntry implements ICorfuSerializable {
         SMR(1, SMREntry.class),
         STREAM_COW(4, StreamCOWEntry.class),
         MULTIOBJSMR(7, MultiObjectSMREntry.class),
-        MULTISMR(8, MultiSMREntry.class);
+        MULTISMR(8, MultiSMREntry.class),
+        CHECKPOINT(10, CheckpointEntry.class);
 
         public final int type;
         public final Class<? extends LogEntry> entryType;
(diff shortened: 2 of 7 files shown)


no2chem:
@slfritchie the checkpoint smoke tests seem to fail this test on some nettybytebuf issue.. any idea why this small change might cause this?

no2chem:
Ok, @slfritchie the issue seemed to be that checkpointing entries used the SMR type, when they should have had their own type defined. The checkpointing mechanism won't work in a real setup without that.

Could you review those changes to see if they look okay to you?

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -99,7 +99,7 @@ public byte asByte() {
 
     public CheckpointEntry(CheckpointEntryType type, String authorID, UUID checkpointID,
                            Map<CheckpointDictKey,String> dict, MultiSMREntry smrEntries) {
-        super(LogEntryType.SMR);
+        super(LogEntryType.CHECKPOINT);
```
Maithem: Why do you need this change?

runtime/src/main/java/org/corfudb/protocols/logprotocol/LogEntry.java
```diff
@@ -113,7 +113,8 @@ public boolean isMutation(UUID stream) {
         SMR(1, SMREntry.class),
         STREAM_COW(4, StreamCOWEntry.class),
         MULTIOBJSMR(7, MultiObjectSMREntry.class),
-        MULTISMR(8, MultiSMREntry.class);
+        MULTISMR(8, MultiSMREntry.class),
+        CHECKPOINT(10, CheckpointEntry.class);
```
Maithem: Why is this related to making the caches non-static ?

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
no2chem: follow the PR dicussion

runtime/src/main/java/org/corfudb/protocols/logprotocol/LogEntry.java
no2chem: also follow the PR discussion, due to a failing test

no2chem:
@slfritchie yes

Forcing per-instance read caches seems to give us more test coverage as it prevents our single-JVM unit tests from falsely sharing caches. Once we have assurance that performance is not affected, I think we should move to merge this PR since we'll want it for checkpoints to work.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d88943e**.

 [Pull Request #634 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/d6cG3rERk89DQQxYoI9InMs0PqZrBUKo)

no2chem:
Ok, perf looks good, will merge after Travis passes.

PR #637, opened 2017-05-24
Forward pointers

kjames88:
Creates but does not use forward pointers so far.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -12,15 +12,11 @@ import com.codahale.metrics.MetricRegistry;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.RemovalCause;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-
 import io.netty.channel.ChannelHandlerContext;
-
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.infrastructure.log.InMemoryStreamLog;
 import org.corfudb.infrastructure.log.StreamLog;
 import org.corfudb.infrastructure.log.StreamLogFiles;
@@ -32,6 +28,14 @@ import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.exceptions.ValueAdoptedException;
 import org.corfudb.util.MetricsUtils;
 import org.corfudb.util.Utils;
+
+import java.lang.invoke.MethodHandles;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 
 /**
@@ -86,6 +90,8 @@ public class LogUnitServer extends AbstractServer {
     private final BatchWriter<Long, ILogData> batchWriter;
 
     private static final String metricsPrefix = "corfu.server.logunit.";
+
+    private final HashMap<LogAddress, Long> forwardPointerMap = new HashMap<>();
 
     public LogUnitServer(ServerContext serverContext) {
         this.opts = serverContext.getServerConfig();
@@ -167,6 +173,12 @@ public class LogUnitServer extends AbstractServer {
                 } else if (e.getType() == DataType.HOLE) {
                     rr.put(l, LogData.HOLE);
                 } else {
+                    for (UUID u : e.getBackpointerMap().keySet()) {
+                        LogAddress a = new LogAddress(l, u);
+                        if (forwardPointerMap.containsKey(a)) {
+                            e.setForwardpointer(u, forwardPointerMap.get(a));
+                        }
+                    }
                     rr.put(l, (LogData)e);
                 }
             }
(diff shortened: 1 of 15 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **c689fbf**.

 [Pull Request #637 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/nrodvVaO14QZSOAL1xmUFTTHZUxwWVD0)

no2chem:
This is a great start, but in order to be performant, the forward pointers need to be added during a read rather than via a special call. I would make it an optional parameter to the read call.

kjames88:
Doesn't this have an issue with caching?  I can wait until the backpointers are all recovered and then write out a forward list, but if the cache is large enough for performance, haven't we then already completed all the reads in question?

no2chem:
@kjames88 I mean the second. It should be a significant performance benefit, because there is no extra call to teach the logunit the new forward pointer, by adding a second call you pay the cost of two message roundtrips.

So a flow would be with entries A (1), B (2) C (3):
1) get backpointer from sequencer to C @ 3
2) read C @ 3 which contains a backpointer to B @ 2
3) read B @ 2 with a hint that we came from C @ 3, which will populate the forward pointer B->3, if not present.
4) read A @ 1 with a hint that we came from B @ 2, which will populate the forward pointer for A->2, if not present.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -223,6 +221,16 @@ private void trim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx, I
...
     }
 
+    @ServerHandler(type = CorfuMsgType.POINTER_UPDATE)
+    private void setPointer(CorfuPayloadMsg<PointerUpdate> msg, ChannelHandlerContext ctx, IServerRouter r,
+                            boolean isMetricsEnabled) {
+        PointerUpdate payload = msg.getPayload();
+        log.debug("setPointer {}->{}", payload.getCurrent(), payload.getNext());
+        forwardPointerMap.putIfAbsent(payload.getStream(), new HashMap<>());
+        forwardPointerMap.get(payload.getStream()).put(payload.getCurrent(), payload.getNext());
+        //r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
```
dahliamalkhi: I guess we should remove this comment

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
```diff
@@ -35,6 +35,17 @@
...
+    /** Write forward pointer to the log for a given address
+     *
+     * This function is non-blocking
+     *
+     * @param layout
+     * @param streamId
+     * @param address
+     * @param next
+     */
+    void writeForwardpointer(Layout layout, UUID streamId, long address, long next); /* throws in case of mismatched pointer? */
```
dahliamalkhi: again, should we delete the comment already?

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
```diff
@@ -131,6 +132,10 @@ public void write(Layout layout, ILogData data) throws OverwriteException {
         }
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public void writeForwardpointer(Layout layout, UUID stream, long address, long nextAddress) {
```
dahliamalkhi: any thoughts on this? should we care that only a quorum might have the actual entry?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
kjames88: Or uncomment it.  Should we ack or just do async?

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
kjames88: It raises a question:  do we need to catch an error here or assume the pointers always match?

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
kjames88: Quorum raises the question of how much coordination is needed to get the pointers replicated (are we just making copies or running paxos like the data replication)?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **0466ba5**.

 [Pull Request #637 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/bmC5buHtGoFbr5LtVqYxO918yE0Zf1XJ)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a9b4e71**.

 [Pull Request #637 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/bsBrRoJKc5uMtYYYnPdvWGycs0AJntPK)

dahliamalkhi:
I am going to close this PR, but keep the branch.

closing status summary: there are two serious issues with the current PR. One is implementation: we want to piggyback forward pointer messages on reads, in order to avoid adding an extra round-trip.

The second one is fundamental, hence closing this: Log entries may contain contents for multiple stream. However,  for every log entry, only the first stream-read actually fetches data from the logging-unit. Remaining stream-reads on the same entries will hit the read cache. Hence, only the first stream-read will provide forward-pointers.

PR #641, opened 2017-05-25
Add MultiCheckpointWriter class + new unit test

slfritchie:
Add MultiCheckpointWriter class + new unit test, which has two variants to try to catch regressions related to a 2nd checkpoint immediately following a checkpoint and also without any regular updates in  between.  The patch to AbstractQueuedStreamView.java fixes the bug exposed by the MultiCheckpointWriter2Test test.

The object's serializer is pulled out of the map's proxy object, avoiding manual management of serialization function for checkpoint data.

Here's an example for the Clojure-curious to create two maps, add a small number of keys, then checkpoint them both.  Run via `bin/shell`.  If you need to specify a special serializer class for your map, add it to the `getObjectsView build` list of builder steps.

    (get-runtime "localhost:9000")
    (.connect *r)
    
    (def m1 (.. *r getObjectsView build (setType org.corfudb.runtime.collections.SMRMap) (setStreamName "StreamA") open))
    (doseq [x [1 2 3 4 5 6 7]] (.put m1 (.concat "key" (Integer/toString x)) (+ 42 x)))
    
    (def m2 (.. *r getObjectsView build (setType org.corfudb.runtime.collections.SMRMap) (setStreamName "StreamB") open))
    (doseq [x [0 1 2 3]] (.put m2 (.concat "KEY" (Integer/toString x)) (+ 84 x)))
    
    (def mcw (new org.corfudb.runtime.MultiCheckpointWriter))
    (.add (.getMaps mcw) m1)
    (.add (.getMaps mcw) m2)
    (.appendCheckpoints mcw *r "Scott")


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -77,6 +77,18 @@ protected ILogData getNextEntry(QueuedStreamContext context,
...
+                // readQueue is empty, readCpQueue is not.
+                // This is a case where we have had 2 checkpoints
+                // adjacent to each other, and no non-checkpoint
+                // entries in the stream in between the checkpoints
+                // or during the 2nd checkpoint.  Processing of
+                // checkpoint entries will not advance our context
+                // globalPointer, only regular entries in readQueue.
+                // However, we know that readQueue is *empty*, so
+                // we advance globalPointer here.
+                context.globalPointer = maxGlobal;
```
dahliamalkhi: This patch concerns me. It seems that there are potentially other cases when a checkpoint will not be followed by updates. So we should set the globalPointer to maxGlobal in those cases as well?

Another corner case would be when there are no updates at all, and there is a checkpoint.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
```diff
@@ -0,0 +1,90 @@
...
+                ISerializer serializer =
+                        ((CorfuCompileProxy<Map>) map.getCorfuSMRProxy())
+                                .getSerializer();
+                cpw.setSerializer(serializer);
+                cpw.setPostAppendFunc(postAppendFunc);
+                List<Long> addresses = cpw.appendCheckpoint();
+                checkpointLogAddresses.addAll(addresses);
+            }
+        } finally {
+            rt.getObjectsView().TXAbort();
```
dahliamalkhi: why abort? this TX has an empty writeSet, so we can commit.

while at it: all the TX's used for checkpointing might as well be snapshotTransaction's .

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -417,4 +421,69 @@ private void writeCheckpointRecords(UUID streamId, String checkpointAuthor, UUID
         }
     }
 
+    @Test
+    public void MultiCheckpointWriter2Test() throws Exception {
+        MultiCheckpointWriterTestInner(true);
+        // While creating & testing MultiCheckpointWriter,
+        // I accidentally discovered that a checkpoint that
+        // immediately follows an earlier checkpoint can cause
+        // map corruption.  Let's test again for posterity.
```
dahliamalkhi: for posterity, I actually think that it is better to explain: 
   'test a dense sequence of (two) checkpoints, without any updates in between'

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -417,4 +421,69 @@ private void writeCheckpointRecords(UUID streamId, String checkpointAuthor, UUID
...
+        MultiCheckpointWriterTestInner(false);
+    }
+
+    public void MultiCheckpointWriterTestInner(boolean regressionFlavor) throws Exception {
```
dahliamalkhi: and a more informative name than regressionFlavor is "consecutiveCkpoints"

PR #648, opened 2017-05-26
Trim to address space, stream and object

no2chem:
This PR supports trim in both the stream and object layers.

In the stream layer, attempting to read from a trimmed stream results
in a TrimmedException. The only way to recover is to reset the stream.

In the object layer, a non-transactional access which encounters a
trim will result in a retry. A transactional access will be aborted
due to a trim exception, because we would be unable to generate the
requested object. The only remedy is to retry the transaction.


no2chem:
@slfritchie The two tests in CheckpointTrimTest appear to be failing because the checkpoint mechanism has two issues:

1) The streamview wants to read two addresses before the checkpoint, even though it encounters the checkpoint start entry.
So if we have a successful checkpoint of map at version 3 written to 4,5,6, the streamview will want to read to version 2, which causes an exception in the first test because it was trimmed. The correct behavior is that it should stop reading at 3.

2) In the previous example, the streamview thinks its pos is 2, even though it read up to a checkpoint finish entry @6. This is a problem for transactions, which will keep trying to sync up to 6 and fail.

no2chem:
@Maithem you could address the issues if you'd like as well

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **93619c6**.

 [Pull Request #648 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/pOmMwvmj4MyQqzftrrYLwHtbXT6b2sQU)

dahliamalkhi:
I believe #2 is precisely what is addressed by #641 

I don't understand comment #1: If the -snapshot- timestamp of a checkpoint is 3, but the snapshot begin record is appended at, say, 6, then we -should- apply updates 4, 5, after restoring from the checkpoint. Agreed?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -506,6 +506,12 @@ protected void rollbackStreamUnsafe(ISMRStream stream, long rollbackVersion) {
      * applied until the current tail of the stream. If Address.OPTIMISTIC
      * is given, updates will be applied to the end of the stream, and
      * upcall results will be stored in the resulting entries.
+     *
+     * When the stream is trimmed, this exception is passed up to the caller,
+     * unless the timestamp was Address.MAX, in which the entire object is
+     * reset and re-try the sync, which should pick up any checkpoint that
+     * was inserted.
+     *
```
dahliamalkhi: this sounds like a desirable behavior, indeed, but I don't see this implemented -- ?

slfritchie:
The multi-checkpoint PR #641 has been merged to master.  If I rebase this PR to master and also restore the MultiCheckpointWriter2Test unit test, I see most checkpoint-related tests passing.

The exceptions are the PR's new`testCheckpointTrimDuringPlayback` and `testCheckpointTrim` tests, both of which are failing with `java.lang.StackOverflowError` exceptions, same as Travis-CI at https://travis-ci.org/CorfuDB/CorfuDB/builds/236275128#L2194

no2chem:
@slfritchie The stack overflow error occurs because the streamview reads 2 addresses before the checkpoint start record, which is trimmed, and the implementation searchers forever for a checkpoint that doesn't throw a trim error.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
no2chem: I see. The comment needs to be updated because this happens in Access now and not in sync

dahliamalkhi:
it seems that #641 does not address the problem of handling a trimedException: If I try to sync with an object, meanwhile the log is trimmed, I reset the object and -should- be able to rebuild it from the checkpoint. Currently, this fails.

slfritchie:
Also, both Travis-CI and I are seeing this failure in the new `prefixTrimThrowsException` test:

    prefixTrimThrowsException                                   org.corfudb.runtime.exceptions.QuorumUnreachableException: Couldn't reach quorum, reachable=0, required=1

... very long stack trace omitted, but it's the same as Travis's run here: https://travis-ci.org/CorfuDB/CorfuDB/builds/236275122#L2015

slfritchie:
I've fixed the stack overflow errors by fixing the unit tests.  Michael was using the low-level methods for checkpoint creation, which only return information about where the checkpoint records are written.  The replay of checkpoint state may have to go backward further than where the checkpoint START record appears.  The MultiCheckpointWriter class gives the log position the tests really need.

slfritchie:
Oops, I forgot to say: added by commit 379906a.

slfritchie:
Here is an attempt to fix the last remaining bug.  I'm not sure this is a technique you'd like to use, but ... if a quorum of responses are the same exception, then throw it to the client.

https://gist.github.com/slfritchie/a55ce41b1a4eb4689b0c29c6c3980181

Maithem:
@slfritchie does it pass all the tests?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **4ad75db**.

 [Pull Request #648 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/dGqbTfJvHwVZZO4P5dHNLufULEXaY5iE)

dahliamalkhi:
Regarding the quorumFutures exception fix:

Thanks for looking into it, it clarifies what is the problem. I am not sure this is the best fix. First, if there are multiple exception types, and a quorum of them are the same, we mat want to throw the quorum's exception, even though (uniqueThrowables.size() == 1) is false.

More generally, if there is only one (or few) trimmedException, why not propagate that to the caller even if we don't have a quorum?

Finally, the most general fix would be to set a policy: When we get a mix of exceptions, which one should we propagate back to the caller?

For now, I propose: Look at the exception type. If any trimmedException is there, throw that. Otherwise, throw quorumUnreachable. Later, we can turn this logic into a more general policy regarding exceptions returned from loggingunits.

no2chem:
@dahliamalkhi We probably need to define what happens when there are units in mixed trim/data states for quorum replication. It's possible that throwing a trimmed exception is always safe as well, so perhaps we should look into that.

I submitted a patch which checks for a quorum of exceptions, if a quorum is met that exception is thrown. Otherwise a QuorumUnreachableException is thrown.

If this patch looks okay with you, I'll rebase it and we can merge it in.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **db9d515**.

 [Pull Request #648 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/9TVR1pXjUjfLcvY4j3JKK0GEddQ38z4k)

dahliamalkhi:
The quorumFutures fix looks fine to me. Unfortunately, it seems that the checkpoint trim is still failing with stackOverflow; I just submitted a patch with a test , if ok with you, please merge it into -this- branch

no2chem:
@dahliamalkhi The test from #657 appears to fail if there is more than one checkpointed object.

@slfritchie This appears to happen because with more than one checkpointed object, the backpointer implementation continues to read one entry past the start entry:
7:57:19.093 TRACE [main] o.c.r.v.s.BackpointerStreamView - Checkpoint: halt backpointer fill at address 1 type START id 9b1a5884-be8e-4c78-b5fd-eda4d9ca89b6 author periodicCkpoint
 17:57:19.093 TRACE [main] o.c.r.v.s.BackpointerStreamView - Read_Fill_Queue[e871@-1] Backpointer 2->0
And inevitably the read of entry 0, which was trimmed causes the implementation to retry, eventually resulting in the stack overflow error that you see.

slfritchie:
> @slfritchie <https://github.com/slfritchie> This appears to happen
because with more than one checkpointed object, the backpointer
implementation continues to read one entry past the start entry:

In general, if a checkpoint START record appears at address A, then A does
not mark the start of *all* data required to preserve the stream's SMRMap
state.  The exact minimum global address you need is, the checkpoint start
address, START_TIME, which is stored inside of the metadata for the
checkpoint START record @ A: that gives you the precise global log address
to preserve the stream's state.

I need another set of eyes to examine this, sorry, ENOTQUITESOBER.
However, my eyes are suggesting that the test assertion is incorrect?

https://github.com/CorfuDB/CorfuDB/pull/657/files#diff-4a6b4a7e1b9410ee9953a12ede1d6c4cR228

Map `localm2B` map's entries are always 0L.  Is line 228's asserted value a
cut-and-paste-o/think-o error based on line 227?

-Scott

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **458513e**.

 [Pull Request #648 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/e32cfm3LCwkPDCSN7PiDznixRQzl3FtE)

slfritchie:
Second observation this evening, also clouded by ENOTQUITESOBER.

Back when discussing the checkpoint design, we settled approximately on a distinction between a "new" instantiation of a map vs. a "veteran" instantiation.  I am guessing now, looking at `TRACE` output of Dahlia's new unit test (which was merged to master when it was still *failing*, hmmm...) we have a problem with veteran instantiation?  A veteran encounters a trimmed location, so it's forced to retry.  The sequencer doesn't tell anything new, and the veteran status implies ignore checkpoints, so ... the poor veteran hits the same trimmed address again, lather, rinse, repeat until stack space is exhausted.

The new/veteran status calculation is done at https://github.com/CorfuDB/CorfuDB/blob/25c123861a512df094827c095f7a6c5adc165b60/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java#L180

... which was sufficient before range trim was added, and after range trim was added, then this simple test at line 180 isn't good enough?

One way to fix the stack overflow bug is to switch from veteran -> new instantiation when a trimmed region is hit.  Logically speaking, trim means that log history has been summarized, aggregated, and destroyed.  For a veteran map instantiation, it's perhaps impossible to reconstruct 100% of required state without examining checkpoint data.  So, this theory says we need to fall back to reconstructing what state **is available now** by looking at the last successful checkpoint's data.

I'm not sure what's best globally for making that veteran -> new state switch.  BackpointerStreamView.java line 180 says that adjusting the stream context's `globalPointer` to -1 can make a map instantiation "new".  I've no idea right now if adjusting that one context state variable is sufficient, beautiful, maintainable, etc..  But it's now clear that a bug has been found and some fix is needed, right?

slfritchie:
Michael, I see three different kinds of intermittent failures.  Stack overflow & cancellation exceptions are caused by the same infinite loop on reset.  I haven't looked into that one deeply.

The third is an assertion error, caused when two different threads race when calling resetUnsafe().  See the gist at https://gist.github.com/slfritchie/f704fbb9974e4f7a43ba429c3b855cd9 for details of the race at time 18:52:20.209.

Note also that the patch below is a hack that I used to help locate this bug and ought to be included (in some form) in a final fix.

    diff --git a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
    index e6f3097..a2c5aad 100644
    --- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
    +++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
    @@ -380,6 +380,12 @@ public abstract class AbstractQueuedStreamView extends
                 super.reset();
                 readCpQueue.clear();
                 readQueue.clear();
    +            checkpointSuccessID = null;
    +            checkpointSuccessStartAddr = Address.NEVER_READ;
    +            checkpointSuccessEndAddr = Address.NEVER_READ;
    +            checkpointSuccessNumEntries = 0L;
    +            checkpointSuccessBytes = 0L;
    +            resolvedEstBytes = 0L;
             }
     
             /** {@inheritDoc} */

dahliamalkhi:
yes, @slfritchie , there was a cut-paste error.

however, I believe that the trimmed-exception induced problem is orthogonal.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTest.java
```diff
@@ -194,7 +194,7 @@ public void periodicCkpointTrimTest() throws Exception {
         scheduleConcurrently(1, ignored_task_num -> {
             for (int i = 0; i < mapSize; i++) {
                 m2A.put(String.valueOf(i), (long)i);
-                m2B.put(String.valueOf(i), (long)0);
+                m2B.put(String.valueOf(i), (long)i);
```
dahliamalkhi: oh, that's the 'wrong' fix, I wanted the maps to differ and for m2B to contains zeros. Anyway, at least you made it consistent.

no2chem: Could you fix the test and push/submit a fix?

no2chem: By fix, I mean fix it to do what you meant. I just fixed it since it seemed to be incorrect

no2chem:
@slfritchie I don't think that's a race... I'm not sure what test your trace is from, but it looks like test-1 and test-4 are working on different views of the same object?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **8a70f30**.

 [Pull Request #648 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Xj1vSNsm7TIBjlsHSfjaUlL6cscVWoCN)

slfritchie:
>> @slfritchie <https://github.com/slfritchie> I don't think that's a
race... I'm not sure what test your trace is from, but it looks like test-1
and test-4 are working on different views of the same object?

It's from the `periodicCkpointTrimTest` test, and yes, two views of the
same object.  But every time that a stream reset happens in one while the
other is applying updates to the map, the map is corrupted.

I have hacks that appear to fix the map corruption with that test.  Stack
overflow is not fixed by them.  WIP quality.
https://gist.github.com/slfritchie/bbdc5386f8b1b83a02a52d1516bc82eb

no2chem:
I've fixed the issues with recursive trim exceptions by fixing the logic that checks for checkpoints, but it seems that for some reason the checkpoints aren't adopted occasionally, resulting in the map ending up being null when read. Could you look into this @slfritchie ? Possibly the check in BackpointerStreamView:L253 is wrong - the core problem is that BackpointerStreamView is following backpointers past a valid checkpoint.

The "race" that @slfritchie observed was due to checkpointed state not getting cleaned up on a reset, not anything to do with locks. I think that issue is now resolved.

no2chem:
also @dahliamalkhi I think that I fixed the test behavior you were concerned about

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **59e59ef**.

 [Pull Request #648 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/HrVflixGPsWootRjyyv1JsRzu0XwGvfH)

slfritchie:
I've added a fix to clearing the stream's resolved queue at stream reset() time, commit 916ebbc

I have not yet checked in the `currentMapSize` correction from an hour ago ... but if I use that correction, I still see failures in `periodicCkpointTrimTest`, approximately 1% of the time.  {sigh}  Still looking into it.

slfritchie:
I think I've found & fixed the remaining race conditions.  My last 3 commits:

* 	56efb77 : clear stream resolved queue state correctly (thanks to Michael for pointing out my omissions)
* 	fcfa4b8 : change to BackpointerStreamView when observed checkpoint isn't available
* 1e9f6ab : make debugging CheckpointTest.java easier when TRACE level logging is enabled

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **1e9f6ab**.

 [Pull Request #648 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/1mYlPYEnD76GZuIIk7TbXQ0sqdzfBC79)

slfritchie:
Any additional review requests, @dahliamalkhi & @no2chem & @Maithem?

Maithem:
@slfritchie fyi you can directly invoke travis using this link https://travis-ci.org/CorfuDB/CorfuDB/builds/238166942?utm_source=github_status&utm_medium=notification

Also, there are merge conflicts, can you or @no2chem resolve them and fix the commit history?
Let's aim to check this in today (if we don't see the error that @dahliamalkhi is reporting)

dahliamalkhi:
This PR already has a pretty long discussion history. @slfritchie could you please pull the branch with all the latest commits and merges, squash them, and then close and reopen this as a new PR? 

I then have a few stylistic remarks to make. We can hopefully merge soon.

slfritchie:
Closing in favor of #662.

PR #652, opened 2017-05-26
Add a test for the stream API

vjeko:
Test to see if the stream API works in case of a failover.

### DIFF
--- a/test/src/test/java/org/corfudb/integration/StreamIT.java
+++ b/test/src/test/java/org/corfudb/integration/StreamIT.java
@@ -1,19 +1,22 @@
 package org.corfudb.integration;
 
+import com.google.common.io.Files;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.view.stream.IStreamView;
 import org.junit.Before;
+import org.junit.Test;
 
+import java.io.File;
 import java.util.Random;
 import java.util.UUID;
+import java.util.stream.IntStream;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * A set integration tests that exercise the stream API.
  */
-
 public class StreamIT extends AbstractIT {
     static String corfuSingleNodeHost;
     static int corfuSingleNodePort;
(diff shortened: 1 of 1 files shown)


no2chem:
@zalokhan you seem to be the best fit for reviewing. I'm ok if it's ok with you

vjeko:
Streaming API is broken -- it does not work with a failedover sequencer.

vjeko:
Let me try to explain what I've observed (slightly modified code to aid with debugging):

Once the sequencer goes into a failover mode. The client will as for hasNext() which translates to:

```java
    @Override
    public boolean getHasNext(QueuedStreamContext context) {
        TokenResponse next = runtime.getSequencerView().nextToken(Collections.singleton(context.id), 0);

        return  !context.readQueue.isEmpty() ||
                next.getToken().getTokenValue() > context.globalPointer;
    }
```

The sequencer...

```java
        if (req.getStreams().size() == 1) {
            UUID streamID = req.getStreams().iterator().next();

            if (streamTailToGlobalTailMap.get(streamID) != null)
                maxStreamGlobalTail = streamTailToGlobalTailMap.get(streamID);

            // if we don't have informatin about this stream tail because of fail-over,
                // return the global tail of the log
            else if (isFailoverSequencer)
                maxStreamGlobalTail = globalLogTail.get() - 1L;
        }
```

... will return `globalLogTail.get() - 1L` which is always greater than zero. Therefore, `next.getToken().getTokenValue() > context.globalPointer`  will always return true (`globalPointer == -1`).

vjeko:
FWIW, here is the exception I am getting:
```
simpleStreamNoFailOverTest                                  [PASS]
simpleStreamFailOverTest                                    [FAIL - NullPointerException:L50]
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 50.681 sec <<< FAILURE! - in org.corfudb.integration.StreamIT
simpleStreamFailOverTest(org.corfudb.integration.StreamIT)  Time elapsed: 10.756 sec  <<< ERROR!
java.lang.NullPointerException
	at org.corfudb.integration.StreamIT.simpleStreamRun(StreamIT.java:93)
	at org.corfudb.integration.StreamIT.simpleStreamFailOverTest(StreamIT.java:50)


Results :

Tests in error: 
  StreamIT.simpleStreamFailOverTest:50->simpleStreamRun:93 NullPointer
```

This does not look like the original problem. Let me see what's happening here.

dahliamalkhi:
I am actually trying to reproduce the problem. so far, your test passes on my machine..

vjeko:
You're right. @dahliamalkhi @zalokhan let me look into what's happening. I can definitely trigger the bug on the NSX side -- for some reason this test is not triggering it.

zalokhan:
@dahliamalkhi Its a race, it passes on my mac too but fails on an ubuntu vm I have.
@vjeko The hasNext api call cannot be relied upon. From the documentation,
```
    /** Returns whether or not there are potentially more entries in this
     * stream - this function may return true even if there are no entries
     * remaining, as addresses may have been acquired by other clients
     * but not written yet, or the addresses were hole-filled, or just failed.
     * @return      True, if there are potentially more entries in the stream.
     */
    boolean hasNext();
```
So it would make more sense to use the `streamUpTo()` call which returns the delta of stream items.

test/src/test/java/org/corfudb/integration/StreamIT.java
```diff
@@ -24,14 +27,43 @@ public void loadProperties() throws Exception {
...
                 .setHost(corfuSingleNodeHost)
                 .setPort(corfuSingleNodePort)
                 .runServer();
 
+        simpleStreamRun();
+        assertThat(shutdownCorfuServer(corfuServerProcess)).isTrue();
+    }
+
+    private void simpleStreamRun() {
         CorfuRuntime rt = createDefaultRuntime();
```
zalokhan: Could you also fix this.
use createRuntime(corfuSingleNodeHost + ":" + corfuSingleNodePort).

test/src/test/java/org/corfudb/integration/StreamIT.java
```diff
@@ -42,8 +74,8 @@ public void simpleStreamTest() throws Exception {
         IStreamView s1 = rt.getStreamsView().get(streamId);
 
         // Verify that the stream is empty
-        assertThat(s1.hasNext())
-                .isFalse();
+        final int N_HASNEXT_TESTS = 20;
+        IntStream.range(0, N_HASNEXT_TESTS).forEach(n -> assertThat(s1.hasNext()).isFalse());
```
zalokhan: hasNext seems to be unreliable and needs fixing. We can assert the results based on `streamUpTo()` call.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **32f443a**.

 [Pull Request #652 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/T5KehcTKhMsfov9RoUMcZOAByddiTUnr)

dahliamalkhi:
I am marking this 'invalid' until further investigation. If we decide to pursue this further, please change the label, otherwise, I'll continue ignoring it.

PR #661, opened 2017-06-01
Properly Construct CP Entries

Maithem:
When a checkpoint(CP)  entry is retrieved from secondary storage,
it not properly constructed. The CP type and UUID are not set. As
a consequence, the stream view ignores the those CP entries
resulting in intermittent problems and data loss.


Maithem:
Please dont check in, yet.

Maithem:
Ok, I think we are good to go.

slfritchie:
These patches were included in #662.  Many thanks again, Maithem, for finding & fixing this bug.

PR #667, opened 2017-06-02
Correct LogUnit State

Maithem:
The interplay between trimming, writing and reading can result
in miscalculating boundaries on the address space. This patch
corrects calculations for initializing first unwritten address,
tail address and last segment.

### DIFF
--- a/test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
+++ b/test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
@@ -422,4 +422,19 @@ public class StreamLogFilesTest extends AbstractCorfuTest {
         // Address 0 is not reflected in trimAddress
         assertThat(trimmedExceptions).isEqualTo(trimAddress + 1);
     }
+
+    @Test
+    public void testPrefixTrimAndStartUp() {
+        StreamLog log = new StreamLogFiles(getContext(), false);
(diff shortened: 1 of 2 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **1780f27**.

 [Pull Request #667 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/bLH2E8ALHCDzXbOfJ30SfZijQIfzqkum)

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
```diff
@@ -422,4 +422,19 @@ public void testPrefixTrim() {
         // Address 0 is not reflected in trimAddress
         assertThat(trimmedExceptions).isEqualTo(trimAddress + 1);
     }
+
+    @Test
+    public void testPrefixTrimAndStartUp() {
+        StreamLog log = new StreamLogFiles(getContext(), false);
+        log.prefixTrim(StreamLogFiles.RECORDS_PER_LOG_FILE / 2);
+        log.compact();
```
Maithem: Without the code fix, the first 3 lines would cause the 4th line to throw an exception which will cause the test to fail. It just simulates a logunit restart after a compaction run.

So, if you remove the code fix (not the first 3 lines), the test wouldnt pass.

PR #669, opened 2017-06-03
move checkpoints to other stream

no2chem:
This PR moves checkpoints to another stream, with the UUID generated
from a string <stream id>_cp. This prevents steady state clients
from incurring the cost of reading checkpoints.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -77,18 +77,7 @@ public abstract class AbstractQueuedStreamView extends
         NavigableSet<Long> getFrom;
         if (context.readCpQueue.size() > 0) {
             getFrom = context.readCpQueue;
-            if (context.readQueue.isEmpty()) {
-                // readQueue is empty, readCpQueue is not.
-                // This is a case where we have had 2 checkpoints
-                // adjacent to each other, and no non-checkpoint
-                // entries in the stream in between the checkpoints
-                // or during the 2nd checkpoint.  Processing of
-                // checkpoint entries will not advance our context
-                // globalPointer, only regular entries in readQueue.
-                // However, we know that readQueue is *empty*, so
-                // we advance globalPointer here.
-                context.globalPointer = maxGlobal;
-            }
+            context.globalPointer = context.checkpointSuccessStartAddr;
         } else {
             getFrom = context.readQueue;
         }
@@ -105,13 +94,10 @@ public abstract class AbstractQueuedStreamView extends
         while (getFrom.size() > 0) {
             final long thisRead = getFrom.pollFirst();
             ILogData ld = read(thisRead);
-            if (ld.containsStream(context.id)) {
-                // Only add to resolved if ld is from readQueue
-                if (getFrom == context.readQueue) {
-                    addToResolvedQueue(context, thisRead, ld);
-                }
-                return ld;
+            if (getFrom == context.readQueue) {
+                addToResolvedQueue(context, thisRead, ld);
             }
+            return ld;
         }
 
         // None of the potential reads ended up being part of this
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -5,6 +5,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.DataType;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.view.Address;
 
 import javax.annotation.Nonnull;
@@ -73,6 +74,12 @@ public abstract class AbstractQueuedStreamView extends
             return null;
         }
 
+        // If maxGlobal is before the checkpoint position, throw a
+        // trimmed exception
+        if (maxGlobal < context.checkpointSuccessStartAddr) {
+            throw new TrimmedException();
+        }
+
         // If checkpoint data is available, get from readCpQueue first
         NavigableSet<Long> getFrom;
         if (context.readCpQueue.size() > 0) {
@@ -91,7 +98,7 @@ public abstract class AbstractQueuedStreamView extends
         // Otherwise we remove entries one at a time from the read queue.
         // The entry may not actually be part of the stream, so we might
         // have to perform several reads.
-        while (getFrom.size() > 0) {
+        if (getFrom.size() > 0) {
             final long thisRead = getFrom.pollFirst();
             ILogData ld = read(thisRead);
             if (getFrom == context.readQueue) {
@@ -120,6 +127,12 @@ public abstract class AbstractQueuedStreamView extends
         // log records less than or equal to maxGlobal.
         // Boolean includes both CHECKPOINT & DATA entries.
         boolean readQueueIsEmpty = !fillReadQueue(maxGlobal, context);
+
+        // If maxGlobal is before the checkpoint position, throw a
+        // trimmed exception
+        if (maxGlobal < context.checkpointSuccessStartAddr) {
+            throw new TrimmedException();
+        }
 
         // We always have to fill to the read queue to ensure we read up to
         // max global.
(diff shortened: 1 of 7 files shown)


no2chem:
@slfritchie I couldn't quite tell what checkpointWriterInterleavedTest() was doing, please check if the test is still correct.

slfritchie:
Hm.  Dahlia's description of "[...] going to tag checkpoint entries with a different stream name [...]' puzzled me this afternoon.  "Tag" brought to my mind a MultiSMREntry-style entry in a stream, which didn't make a lot of sense by the time I reached the bottom of her email message.  It's much more clear now how to interpret "tag" by looking at this PR.

One of the very attractive benefit of the checkpoint-in-stream design is that all efforts to trim & GC/reclaim space used by the stream was limited to a single stream.  No need to involve additional streams and/or external parties in any way.  (Avoiding any need for  2PC or other coordination protocols between external parties is an additional benefit.)  Moving checkpoint data outside of its parent/owner stream opens that big can of worms again.  Or threatens to, at least.

I haven't seen any data that measures the overhead that this PR claims to be avoiding.  So I made judgement based granting benefit of the doubt.

One benefit of Corfu's design that I *do* see here is that a CP entry's position in the global log doesn't change.  When the Sequencer says to write a CP entry, it doesn't care (very much) what Corfu stream it's tagged with: one, many, or even (heresy!) zero streams.  With or without this patch, when in equivalent Corfu clusters when checkpointing some stream `S` and perhaps some other unrelated concurrent writes, each checkpoint record will be assigned the same global addresses by the Sequencer.  As far as I can think of on a Friday night, this same global address assignment means that the plan to use `prefixTrim`-style GC would probably not be affected at all.

For any disk space reclamation scheme other than `prefixTrim`-style GC, however, I would need more time to think about it.  If we ever want to add another GC scheme, mix-and-match other schemes with `prefixTrim`-style, or move sometime later from `prefixTrim`-style GC another scheme in an in-production-environment/zero-downtime kind of way, we ought to think a bit about that now?

no2chem:
@slfritchie I don't see how having the checkpoints in a separate stream causes the problem you describe. Due to the global ordering of the log, the checkpoints have to be ordered after the address of any snapshot taken to generate the checkpoint. 

It's also easy to see how the checkpoint in stream design is untenable, I don't think we need to measure anything - if we periodically checkpoint a 1GB map every minute, and every client is in steady state, every client will have to read 1 GB of checkpoint data, which will be discarded because that data is already in memory. 

I honestly continue to believe that checkpointing in the log is not a great idea, but this PR is not the place for that discussion. Could you please check over checkpointWriterInterleavedTest? Thanks.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **7c242e7**.

 [Pull Request #669 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Dparsuj5bKFnlWsfWcFMzTaJQE9Gbc9O)

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -164,21 +262,25 @@ protected boolean fillReadQueue(final long maxGlobal,
...
+                    .getStreamID(context.id.toString() + "_cp");
+            // Find the checkpoint, if present
+            if (followBackpointers(checkpointID, context.readCpQueue,
+                    runtime.getSequencerView()
+                            .nextToken(Collections.singleton(checkpointID), 0)
+                            .getToken().getTokenValue()
+                    , Address.NEVER_READ, d -> resolveCheckpoint(context, d))) {
+                log.trace("Read_Fill_Queue[{}] Using checkpoint with {} entries",
+                        this, context.readCpQueue.size());
+                return true;
```
dahliamalkhi: why return here, and not fill the regular readQueue as well?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -105,13 +94,10 @@ protected ILogData getNextEntry(QueuedStreamContext context,
...
-            if (ld.containsStream(context.id)) {
-                // Only add to resolved if ld is from readQueue
-                if (getFrom == context.readQueue) {
-                    addToResolvedQueue(context, thisRead, ld);
-                }
-                return ld;
+            if (getFrom == context.readQueue) {
+                addToResolvedQueue(context, thisRead, ld);
             }
+            return ld;
```
dahliamalkhi: why is this a while loop? It is only executed once, no?

slfritchie:
@dahliamalkhi Hmm, the final assertion check of the `checkpointWriterInterleavedTest()` was commented out, as I asked about last night.  When the check is uncommented, that test fails.  I'm investigating it now.

slfritchie:
I've modified the failing unit test, `checkpointWriterInterleavedTest()`, to show actual & expected values of a new map instance `m2` after a checkpoint has been written.  (See [patch](https://gist.github.com/slfritchie/96f581a918bd092b2cb58ee098d51fee).)

For new map instantiations using a global snapshot address earlier than the checkpoint, the unit test is expecting to see three keys in the map: keys with the `first` prefix.  With this patch, the unit test is seeing zero keys in the map.

Several times during design review and code review, I've asked the question: if a map's checkpoint is taken at global address `C`, what should a map look like in a snapshot transaction at log address `s` where `s < C`.  The answer has been: the map should be equal to its value at address `C`.

With this PR, the map's behavior has changed, and the new answer to the above question is that the map's contents are lost.  This is not the behavior that @medhavidhawan asked for the last time that I asked the question during the review of #648.

    appendCheckpoint: txBeginGlobalAddress=02
    Checkpoint entries written to [3, 4, 6, 8, 10]
    m2=new map instance, eH=unit test's expected history value at snapshot address
    
    snap addr=00, m2: []
    snap addr=00, eH: [first2=2, first1=1, first0=0]
    snap addr=01, m2: []
    snap addr=01, eH: [first2=2, first1=1, first0=0]
    snap addr=02, m2: [first2=2, first1=1, first0=0]
    snap addr=02, eH: [first2=2, first1=1, first0=0]
    snap addr=03, m2: [first2=2, first1=1, first0=0]
    snap addr=03, eH: [first2=2, first1=1, first0=0]
    snap addr=04, m2: [first2=2, first1=1, first0=0]
    snap addr=04, eH: [first2=2, first1=1, first0=0]
    snap addr=05, m2: [first2=2, first1=1, first0=0, middle0=0]
    snap addr=05, eH: [first2=2, first1=1, first0=0, middle0=0]
    snap addr=06, m2: [first2=2, first1=1, first0=0, middle0=0]
    snap addr=06, eH: [first2=2, first1=1, first0=0, middle0=0]
    snap addr=07, m2: [middle1=1, first2=2, first1=1, first0=0, middle0=0]
    snap addr=07, eH: [middle1=1, first2=2, first1=1, first0=0, middle0=0]
    snap addr=08, m2: [middle1=1, first2=2, first1=1, first0=0, middle0=0]
    snap addr=08, eH: [middle1=1, first2=2, first1=1, first0=0, middle0=0]
    snap addr=09, m2: [middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=09, eH: [middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=10, m2: [middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=10, eH: [middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=11, m2: [last0=0, middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=11, eH: [last0=0, middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=12, m2: [last1=1, last0=0, middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=12, eH: [last1=1, last0=0, middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=13, m2: [last2=2, last1=1, last0=0, middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]
    snap addr=13, eH: [last2=2, last1=1, last0=0, middle1=1, first2=2, middle2=2, first1=1, first0=0, middle0=0]

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
no2chem: yeah, not my code, but I'll fix it

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **24c8683**.

 [Pull Request #669 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/DlGCblK8IW8VYsKKu1zc6SHGq4j01BLj)

PR #673, opened 2017-06-06
add API to force a checkpoint address

no2chem:
This PR adds an API to force the checkpoint writer to take the snapshot
at a specified address by calling setSnapshotAddress() by default,
the latest token is obtained from the sequencer.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -146,14 +146,11 @@ public class CheckpointWriter {
      */
     public long startCheckpoint() {
         startTime = LocalDateTime.now();
-        AbstractTransactionalContext context = TransactionalContext.getCurrentContext();
-        long txBeginGlobalAddress = context.getSnapshotTimestamp();
-
         this.mdKV.put(CheckpointEntry.CheckpointDictKey.START_TIME, startTime.toString());
-        // Need the actual object's version
-        ICorfuSMR<SMRMap> corfuObject = (ICorfuSMR<SMRMap>) this.map;
+
+        AbstractTransactionalContext context = TransactionalContext.getCurrentContext();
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -239,10 +243,15 @@ public long finishCheckpoint() {
         return endAddress;
     }
 
-    public static long startGlobalSnapshotTxn(CorfuRuntime rt) {
-        TokenResponse tokenResponse =
-                rt.getSequencerView().nextToken(Collections.EMPTY_SET, 0);
-        long globalTail = tokenResponse.getToken().getTokenValue();
+    public static long startGlobalSnapshotTxn(CorfuRuntime rt, Long forceAddress) {
```
Maithem: If this method depends on a member field, do we still want this method to be static?

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
```diff
@@ -74,6 +77,7 @@ public long appendCheckpoints(CorfuRuntime rt, String author,
                                 .getSerializer();
                 cpw.setSerializer(serializer);
                 cpw.setPostAppendFunc(postAppendFunc);
+                cpw.setSnapshotAddress(snapshotAddress);
```
Maithem: what happens if the snapshotaddress is -1 ?

slfritchie:
To be of value, the snapshot's address needs to find its way into the START_LOG_ADDRESS metadata.  Commit cd76ad7 is supposed to do that, please double-check that I'm fetching the address correctly.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **cd76ad7**.

 [Pull Request #673 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/1STj66olLcuviuQSpvwfhFoyluF2KaYV)

PR #678, opened 2017-06-07
Implicit nested transaction from transactional methods inherit transactional context parent type.

rogermichoud:

PR #679, opened 2017-06-08
Fix nested transaction default type

Maithem:
Implicit nested transaction from transactional methods inherit
transactional context parent type.


Maithem:
this pr adds tracing to https://github.com/CorfuDB/CorfuDB/pull/678 
#678 reviewed by @dahliamalkhi

PR #683, opened 2017-06-13
UnsupportedOperationException error handling

vjeko:
An AbortedTransaction whose cause is UnsupportedOperationException
should not be retried, but propagated to the consumer.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -330,9 +330,10 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
     @Deprecated // TODO: Add replacement method that conforms to style
     @SuppressWarnings({"checkstyle:membername", "checkstyle:abbreviation"}) // Due to deprecation
     private <R> R TXExecuteInner(Supplier<R> txFunction, boolean isMetricsEnabled) {
-        long sleepTime = 1L;
         final long maxSleepTime = 1000L;
+        long sleepTime = 1L;
         int retries = 1;
+
         while (true) {
             try {
                 rt.getObjectsView().TXBegin();
@@ -352,10 +353,20 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
                     }
                 }
 
+                if (e.getAbortCause() == AbortCause.UNSUPPORTED_OPERATION) {
+                    // No need to deal with TransactionalContext in this case.
+                    throw e;
+                }
+
                 if (retries == 1) {
                     MetricsUtils
                             .incConditionalCounter(isMetricsEnabled, counterTxnRetry1, 1);
                 }
+
+                if (retries > rt.getParameters().getMaxNumOfImplicitTxRetries()) {
+                    throw e; // Give up.
+                }
+
                 MetricsUtils.incConditionalCounter(isMetricsEnabled, counterTxnRetryN, 1);
                 log.debug("Transactional function aborted due to {}, retrying after {} msec",
                         e, sleepTime);
@@ -454,6 +465,9 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
             // as a TransactionAbortedException.
             snapshotTimestamp = -1L;
             abortCause = AbortCause.NETWORK;
+        } else if (e instanceof UnsupportedOperationException) {
+            snapshotTimestamp = context.getSnapshotTimestamp();
+            abortCause = AbortCause.UNSUPPORTED_OPERATION;
         } else {
             snapshotTimestamp = context.getSnapshotTimestamp();
             abortCause = AbortCause.UNDEFINED;
--- a/test/src/test/java/org/corfudb/runtime/object/transactions/SnapshotTransactionContextTest.java
+++ b/test/src/test/java/org/corfudb/runtime/object/transactions/SnapshotTransactionContextTest.java
@@ -2,15 +2,18 @@ package org.corfudb.runtime.object.transactions;
 
 import com.google.common.reflect.TypeToken;
 import org.corfudb.runtime.collections.SMRMap;
+import org.corfudb.runtime.exceptions.AbortCause;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
+import org.junit.Assert;
 import org.junit.Test;
 
 /**
  * Created by mwei on 11/22/16.
  */
 public class SnapshotTransactionContextTest extends AbstractTransactionContextTest {
+
     @Override
     public void TXBegin() { SnapshotTXBegin(); }
-
 
 
     /** Check if we can read a snapshot from the past, without
@@ -87,6 +90,22 @@ public class SnapshotTransactionContextTest extends AbstractTransactionContextTe
         }));
         t(0, this::TXEnd);
         t(0, this::TXEnd);
+    }
 
+
+    /**
+     * Ensure that TransactionAbortedException(AbortCause.UNSUPPORTED_OPERATION) is thrown when
+     * doing a mutation inside of a snapshot read.
+     */
+    @Test
+    public void testUnsupportedOperationException() {
+        try {
+            SnapshotTXBegin();
+            putIfAbsent("k" , "v1");
+            TXEnd();
+            Assert.fail();
+        } catch (TransactionAbortedException e) {
+            Assert.assertTrue(e.getAbortCause() == AbortCause.UNSUPPORTED_OPERATION);
+        }
     }
 }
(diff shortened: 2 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -298,30 +298,43 @@ public UUID getStreamID() {
     }
 
     private <R> R TXExecuteInner(Supplier<R> txFunction, boolean isMetricsEnabled) {
-        long sleepTime = 1L;
         final long maxSleepTime = 1000L;
+        long sleepTime = 1L;
         int retries = 1;
+        int maxNumOfRetries = 20; // Approximately 10 seconds.
```
rogermichoud: Would it make sense to have the user decide on the number of retries ? If it is default, we can set it to 20.

rogermichoud:
Could you add a test ?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **3e39b7f**.

 [Pull Request #683 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/MYltTJcbRDIph0qay3OoKITIAcPexh7q)

vjeko:
@rogermichoud good idea. I'll add those today.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d20610f**.

 [Pull Request #683 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/092qmKPLoLVLQpVjtRZGDc11Yy3uNlm1)

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -298,30 +298,42 @@ public UUID getStreamID() {
...
+                        // No need to deal with TransactionalContext in this case.
+                        throw e;
+                    }
                 }
 
                 if (retries == 1) {
                     MetricsUtils.incConditionalCounter(isMetricsEnabled, counterTxnRetry1, 1);
                 }
+
+                if (retries > CorfuRuntime.getMaxTxRetries()) { // Approximately 10 seconds.
```
Maithem: If the default MaxTxRetries value is 20 and there is a 1ms sleep in-between retries, then how will this take approximately 10 seconds? wouldn't it be ~20 ms?

Associating time with the number of retries here seems misleading, unnecessary comment?

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -298,30 +298,42 @@ public UUID getStreamID() {
...
                 return ret;
             } catch (TransactionAbortedException e) {
-                // If TransactionAbortedException is due to a 'Network Exception' do not keep retrying a nested
-                // transaction indefinitely (this could go on forever). If this is part of an outer transaction abort
-                // and remove from context. Re-throw exception to client.
-                if (e.getAbortCause() == AbortCause.NETWORK) {
-                    if (TransactionalContext.getCurrentContext() != null) {
+                // If TransactionAbortedException is due to a 'Network Exception' or a
+                // 'Unsupported Operation Exception', do not keep retrying a nested transaction
+                // indefinitely (this could go on forever). If this is part of an outer
```
Maithem: Why do we handle retries differently if its nested ?

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -298,30 +298,42 @@ public UUID getStreamID() {
...
                         throw e;
                     }
+
+                    if (e.getAbortCause() == AbortCause.UNSUPPORTED_OPERATION) {
+                        // No need to deal with TransactionalContext in this case.
+                        throw e;
+                    }
                 }
 
                 if (retries == 1) {
```
Maithem: I know you didnt write this code, but this will result in incorrect stats, since your fixing retries maybe also address that too?

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
vjeko: I am not sure how to do it cleanly. It's either we:

1) Pass the value all the way to CompileProxy from the Runtime that requires changing constructor signatures for a couple of classes. (ugly)
2) Make the parameter static (somewhat less ugly but Maithem doesn't like this approach).

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
vjeko: Good question. I don't know. 

It just occured to me that AbortCause.UNSUPPORTED_OPERATION can be taken out of that getCurrentContext() check.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **8cf8587**.

 [Pull Request #683 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/BtgZQo0nFdzm53T8RM7bJuRhEEQhEnJL)

no2chem:
@Maithem @rogermichoud Please update your reviews if the changes are acceptable.

test/src/test/java/org/corfudb/runtime/object/transactions/SnapshotTransactionContextTest.java
```diff
@@ -87,6 +90,22 @@ public void testSnapshotTxNestedExplicitTx() {
...
+
+    /**
+     * Ensure that TransactionAbortedException(AbortCause.UNSUPPORTED_OPERATION) is thrown when
+     * doing a mutation inside of a snapshot read.
+     */
+    @Test
+    public void testUnsupportedOperationException() {
+        try {
+            SnapshotTXBegin();
+            put("k" , "v1");
```
rogermichoud: I thought the problem was the endless retrying in an implicit transaction while being in a Snapshot Transaction.
This is not what is tested here.

You can test something like this:
```
SnapshotTXBegin();
putIfAbsent("k", "v1);
TXEnd();
```

vjeko: The problem was modifications within snapshot reads and that's what being tested.

rogermichoud: I see what you mean but the function that you modified is TXExecuteInner that is called from TXExecute that is an "internal API" called for implicit transaction. This is used for SMRMaps methods annotated @Transactional.

What you tested here is the abortTransaction function that is called when the UNSUPPORTED_OPERATION is thrown during a non-nested transaction. This is part of the test. Before your fix, the transaction would also be aborted but with an UNDEFINED cause.

If you want to test that within a implicit transaction an UnsupportedOperationException will be thrown to the outer transaction without any retry, you have to use a SMRMap method annotated as @Transactional (e.g. putIfAbsent) so you can exercise TXExecuteInner.

Also, you might also want to test the retry logic within TXExecuteInner but this might be a bit harder to test.

vjeko: Sounds good. I'll trigger the logic for implicit transactions by just using `putIfAbsent`.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **00cc0cd**.

 [Pull Request #683 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/6TJfWcWLuABp3JTob3i7DBUwnBHeg2sQ)

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
no2chem: Could you just make this take settings from CorfuRuntime?

vjeko:
@no2chem done.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **1151e75**.

 [Pull Request #683 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/3ncKn6lWrAsTzVVn8uullVZtqqAIvcId)

dahliamalkhi:
seems like we should have merged this already. @no2chem @medhavidhawan @Maithem any objections ? @rogermichoud if approved, please rebase so we can merge.

annym:
@Maithem No, my PR does not make this change obsolete. Still any transaction failing due to UNSUPPORTED or UNDEFINED cause (case of non-existing indexes) will be retried. So I believe this is still valid and we could actually also consider if this should be extended for UNDEFINED abort cause.

vjeko:
Since this PR was created from a fork that does not exist anymore, the only way to make modifications is to create a new PR: https://github.com/CorfuDB/CorfuDB/pull/2150

PR #712, opened 2017-06-14
Terminate Backpointers at head of stream

annym:
- Fix backpointer termination
- Unit Test

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
@@ -1,9 +1,9 @@
 package org.corfudb.runtime.view.stream;
 
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.TokenResponse;
-import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.exceptions.TrimmedException;
@@ -161,6 +161,10 @@ public class BackpointerStreamView extends AbstractQueuedStreamView {
         INCLUDE_STOP    /** Stop, but also include this address. */
     }
 
+    private long backpointerCount = 0L;
+
+    public long getBackpointerCount() {return backpointerCount;}
+
     protected boolean followBackpointers(final UUID streamId,
                                       final NavigableSet<Long> queue,
                                       final long startAddress,
@@ -172,7 +176,7 @@ public class BackpointerStreamView extends AbstractQueuedStreamView {
         long currentAddress = startAddress;
 
         // Loop until we have reached the stop address.
-        while (currentAddress > stopAddress) {
+        while (currentAddress > stopAddress  && Address.isAddress(currentAddress)) {
             // The queue already contains an address from this
             // range, terminate.
             if (queue.contains(currentAddress)) {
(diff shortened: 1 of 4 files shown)


no2chem:
so it looks like this test is failing due to a magic numbers style issue:

[ERROR] /home/travis/build/CorfuDB/CorfuDB/test/src/test/java/org/corfudb/runtime/collections/SMRMapTest.java:776:26: '101' is a magic number. [MagicNumber]

test/src/test/java/org/corfudb/runtime/view/stream/BackpointerStreamViewTest.java
```diff
@@ -113,4 +111,30 @@ public void moreReadQueueTest() {
...
     }
+
+    /**
+     *  test proper backpointer termination at the head of a stream
+     *
+     * */
+    @Test
+    public void headOfStreamBackpointerTermination() {
+
+        final int totalEntries = 101;
```
dahliamalkhi: Do you mean that totalEntries is the same as PARAMETERS.NUM_ITERATIONS_LOW? then use that instead of a constant. Note that PARAMETERS gets different values inside travis.

annym: totalEntries is  PARAMETERS.NUM_ITERATIONS_LOW (number of entries in StreamA) + 1 (number of entries in StreamB)... I can change it to be PARAMETERS.NUM_ITERATIONS_LOW  1

no2chem: Yeah, I think that would be good.... make the change and I'll approve.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -161,6 +161,10 @@ protected boolean fillFromResolved(final long maxGlobal,
         INCLUDE_STOP    /** Stop, but also include this address. */
     }
 
+    private long backpointerCount = 0L;
+
+    public long getBackpointerCount() {return backpointerCount;}
+
     protected boolean followBackpointers(final UUID streamId,
```
medhavidhawan: This is not related to your work but please break this unreadable method into smaller more understandable methods with comments. The code I was looking at earlier is a few days old and this one looks different from that one. Could you please break this method into smaller more cohesive methods that are less bug prone and more digestible. Debugging these things in a production scenario is not easy.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -199,20 +203,29 @@ protected boolean followBackpointers(final UUID streamId,
...
+                // downgrading to a linear scan
                 currentAddress = currentAddress - 1;
             }
         }
 
         return entryAdded;
+
     }
 
     protected BackpointerOp resolveCheckpoint(final QueuedStreamContext context, ILogData data) {
```
medhavidhawan: This giant method needs comments and needs to be broken down as well.

test/src/test/java/org/corfudb/runtime/collections/SMRMapTest.java
```diff
@@ -734,6 +739,45 @@ public void bulkReads()
                 (endTime - startTime) / MILLISECONDS_TO_MICROSECONDS);
     }
 
+    /**
+     *  test proper backpointer termination at the head of the stream\
+     *
+     * */
+    @Test
+    public void headOfStreamBackpointerTermination() {
```
medhavidhawan: This is a great test. Please can you describe the steps and what and how you are trying to test. It is important for tests to be high quality and have quality reviews. Thanks!

test/src/test/java/org/corfudb/runtime/view/stream/BackpointerStreamViewTest.java
annym: Done!

PR #716, opened 2017-06-15
guarding logging from evaluation

zalokhan:
Guarding logging from evaluating statement.
Even if tracing is not enabled, this is evaluated though not logged resulting in CPU utilization.

Note: Consider moving to Log4j to incorporate lambda based logging which avoids need for such guarding and is evaluated only if logging level is satisfied.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
@@ -127,7 +127,9 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter
             } else {
                 if (validateEpoch(m, ctx)) {
                     // Route the message to the handler.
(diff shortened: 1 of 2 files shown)


zalokhan:
As per the profiler this statement evaluation contributes to 15% CPU utilization.
In this statement in the NettyClientRouter is invoked on every message sent. is This is aggravated by the reflection call as one of the parameters in the logging statement.

You can test this in a simple case as follows: Turn the trace level of the test cases to INFO so that TRACE logs should not be invoked.
```
    public String xyz() throws Exception {
        Thread.sleep(5000L);
        return " ";
    }

    @Test
    public void test() throws Exception {
        System.out.println(log.isTraceEnabled());
//        if(log.isTraceEnabled())
            log.trace("Should not take time : {}", xyz());
    }
```
Run this test with and without the guard.
These statements ideally should not be evaluated at all.

PR #720, opened 2017-06-16
Metadata Service API

Maithem:
This patch defines the interface for the metadata service.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/metadata/Checkpoint.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/metadata/Checkpoint.java
@@ -1,0 +1,14 @@
+package org.corfudb.runtime.view.metadata;
+
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * This class defines metadata associated with a checkpoint
+ */
+public class Checkpoint {
+    private UUID checkpointStream;
+    private long startAddress;
+    private long endAddress;
+    List<Long> checkpointRecords;
+}
--- a/runtime/src/main/java/org/corfudb/runtime/view/metadata/IMetadataServiceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/metadata/IMetadataServiceView.java
@@ -1,0 +1,40 @@
+package org.corfudb.runtime.view.metadata;
+
+import java.util.UUID;
+
+/**
+ * This interface defines the operations provided by the metedata service.
+ */
+
+public interface IMetadataServiceView {
+
+    /**
+     * Create a stream
+     * @param stream The stream to create
+     * @throws StreamAlreadyExistsException This exception is thrown if a previous create opertaion
+     * succeeded with the same stream uuid
+     */
+    void createStream(UUID stream) throws StreamAlreadyExistsException;
+
+    /**
+     * Update a stream with a new checkpoint
+     * @param stream The stream's address
+     * @param checkpoint New checkpoint
+     * @throws InvalidCheckpointException This exception is thrown when the new checkpoint is logically older than
+     * the last successful checkpoint update (i.e. Only monotonic updates are allowed)
+     */
+    void updateStreamCheckpoint(UUID stream, Checkpoint checkpoint) throws InvalidCheckpointException;
+
+    /**
+     * Retreive a checkpoint for a stream
+     * @param stream stream's UUID
+     * @return Checkpoint data or null
+     */
+    Checkpoint getStreamCheckpoint(UUID stream);
+
+    /**
+     * Delete a stream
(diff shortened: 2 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/view/metadata/Checkpoint.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.UUID;
+
+/**
+ * This class defines metadata associated with a checkpoint
+ */
+public class Checkpoint {
+    private UUID checkpointStream;
+    private long startAddress;
+    private long endAddress;
+    List<Long> checkpointRecords;
```
no2chem: I don't think we need to store the address of each checkpoint record?

runtime/src/main/java/org/corfudb/runtime/view/metadata/IMetadataServiceView.java
```diff
@@ -0,0 +1,40 @@
...
+
+public interface IMetadataServiceView {
+
+    /**
+     * Create a stream
+     * @param stream The stream to create
+     * @throws StreamAlreadyExistsException This exception is thrown if a previous create opertaion
+     * succeeded with the same stream uuid
+     */
+    void createStream(UUID stream) throws StreamAlreadyExistsException;
```
no2chem: I think this needs the stream's "start address" too, as well as some metadata: stream name and the type being stored

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **4811e00**.

 [Pull Request #720 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/IZfHMV6AMwhSCFeTDUr8zTIvlRf6OI4i)

runtime/src/main/java/org/corfudb/runtime/view/metadata/Checkpoint.java
dahliamalkhi: I don't see a need right away. however, checkpointWriter does keep track of those. So it might not be a bad idea to retain them for any potential future use. Hopefully, with reasonable batch sizes, this list will not be very long.

Maithem: The reason I have it there is to enable bulk read of a checkpoint streams, instead of having to resolve the backpoints from the end address.

runtime/src/main/java/org/corfudb/runtime/view/metadata/IMetadataServiceView.java
Maithem: If you create a stream, but don't write to it, how would you know the start address?

rogermichoud:
Do we want also to have an API for querying streams Metadata ? 

getStreamTail
getStreamHead (if there is not checkpoint yet, we should still know where it is).
getStreams

Maithem:
@rogermichoud No, stream tails will be tricky because the tail can change at a much higher rate than the start address. I'm trying to see whats the best way to pass start stream address as a part of the create (this was pointed out by @no2chem ). As for getStreams, yes I will add that.

runtime/src/main/java/org/corfudb/runtime/view/metadata/Checkpoint.java
Maithem: If we assume this list is 1,000,000 pointers, then it takes ~35mb to represent in memory (using an ArrayList) and only 4mb in its serialized form. Seems like a small overhead?

rogermichoud: For the recovery loading of the maps in one pass I would also like to have all the addresses.  As Maithem mentioned, I don't have to follow backpointers from the end address, which is nice, since I read forward.

runtime/src/main/java/org/corfudb/runtime/view/metadata/IMetadataServiceView.java
no2chem: One way would be to acquire an address first - I think you have to provide a start address, otherwise what would the flow for creating a stream be?

Maithem: What if you acquire a token but dont write to it? Since an operation that spans a token acquisition and metadata storeupdate is not atomic, then I don't feel like having a stream start address makes sense because it might not be accurate. Maybe just pass the current global tail address instead of passing the first stream address? Using the global tail when creating the stream would tell us no entry belonging to this stream would exist before that tail (i.e. hard lower watermark when traversing the backpointers)

Issue #724, opened 2017-06-16
Support range read on the client side


zjohnny226:
In the following methods, readAll sends multiple reads instead of single range read.
AddressSpaceView::cacheFetch(Iterable<Long> addresses)
IReplicationProtocol::readAll(Layout layout, Set<Long> globalAddresses)

PR #725, opened 2017-06-16
change cache readAll() to utilize ranged read instead of single reads

zjohnny226:
Issue #724 

Only implemented ChainReplicationProtocol for now.

Couple of issues/questions:
1. API currently supports ranged read, not random reads
2. Method signatures need to be changed when 1) is resolved. Either user of these methods need to supply list of ranges (user do more work), or readAll() needs to be able to properly compile a list of ranges.
3. Is this the right way to handle hole filling?
4. In cacheFetch(), can these different address have different replication protocols?

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -5,10 +5,11 @@ import com.github.benmanes.caffeine.cache.CacheLoader;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 
+import java.util.Iterator;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -252,18 +253,23 @@ public class AddressSpaceView extends AbstractView {
     }
 
     /**
-     * Fetch an address for insertion into the cache.
+     * Fetch a collection of addresses for insertion into the cache.
      *
-     * @param addresses An address to read from.
-     * @return A result to be cached. If the readresult is empty,
-     *         This entry will be scheduled to self invalidate.
+     * @param addresses collection of addresses to read from.
+     * @return A result to be cached
      */
     private @Nonnull Map<Long, ILogData> cacheFetch(Iterable<Long> addresses) {
-        return StreamSupport.stream(addresses.spliterator(), true)
-                .map(address -> layoutHelper(l ->
-                        l.getReplicationMode(address).getReplicationProtocol(runtime)
-                        .read(l, address)))
-                .collect(Collectors.toMap(ILogData::getGlobalAddress, r -> r));
+        //turn the addresses into Set for now to satisfy signature requirement down the line
+        Set<Long> readAddresses = new TreeSet<>();
+        Iterator<Long> iterator = addresses.iterator();
+        while(iterator.hasNext()){
+            readAddresses.add(iterator.next());
+        }
+
+        //doesn't handle the case where some address have a different replication mode
+        return layoutHelper(l -> l.getReplicationMode(readAddresses.iterator().next())
+                .getReplicationProtocol(runtime)
+                .readAll(l, readAddresses));
     }
 
 
--- a/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
@@ -1,13 +1,18 @@
 package org.corfudb.runtime.view.replication;
 
+import com.google.common.collect.Range;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
 import javax.annotation.Nullable;
-
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.exceptions.RecoveryException;
 import org.corfudb.runtime.view.Layout;
 import org.corfudb.util.CFUtils;
+
 
 
 /**
@@ -64,6 +69,35 @@ public class ChainReplicationProtocol extends AbstractReplicationProtocol {
                                     .read(globalAddress)).getReadSet()
                 .getOrDefault(globalAddress, null);
         return ret == null || ret.isEmpty() ? null : ret;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map<Long, ILogData> readAll(Layout layout, Set<Long> globalAddresses) {
+        Range<Long> range = Range.encloseAll(globalAddresses);
+        long startAddress = range.lowerEndpoint();
+        long endAddress = range.upperEndpoint();
+        int numUnits = layout.getSegmentLength(startAddress);
+        log.trace("ReadAll[{}-{}]: chain {}/{}", startAddress, endAddress, numUnits, numUnits);
+
+        Map<Long, LogData> logResult = CFUtils.getUninterruptibly(layout
+                .getLogUnitClient(startAddress, numUnits - 1)
+                .read(null, range)).getReadSet();
+
+        //in case of a hole, do a normal read and use its hole fill policy
+        Map<Long, ILogData> returnResult = new TreeMap<>();
+        for (Map.Entry<Long, LogData> entry: logResult.entrySet()){
+            ILogData value = entry.getValue();
+            if (value == null || value.isEmpty()){
+                value = read(layout, entry.getKey());
+            }
+
+            returnResult.put(entry.getKey(), value);
+        }
+
+        return returnResult;
     }
 
     /** Propagate a write down the chain, ignoring
--- a/test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
+++ b/test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
@@ -1,9 +1,13 @@
 package org.corfudb.runtime.view;
 
+import com.google.common.collect.ContiguousSet;
+import com.google.common.collect.DiscreteDomain;
+import com.google.common.collect.Range;
 import org.corfudb.infrastructure.LogUnitServerAssertions;
 import org.corfudb.infrastructure.TestLayoutBuilder;
 import org.corfudb.protocols.wireprotocol.*;
 import org.corfudb.runtime.CorfuRuntime;
+import org.junit.Before;
 import org.junit.Test;
 
 import java.util.*;
@@ -15,10 +19,8 @@ import static org.assertj.core.api.Assertions.assertThat;
  */
 public class AddressSpaceViewTest extends AbstractViewTest {
 
-    @Test
-    @SuppressWarnings("unchecked")
-    public void ensureStripingWorks()
-            throws Exception {
+    @Before
+    public void setup(){
         addServer(SERVERS.PORT_0);
         addServer(SERVERS.PORT_1);
         addServer(SERVERS.PORT_2);
@@ -40,7 +42,12 @@ public class AddressSpaceViewTest extends AbstractViewTest {
                 .addToSegment()
                 .addToLayout()
                 .build());
+    }
 
+    @Test
+    @SuppressWarnings("unchecked")
+    public void ensureStripingWorks()
+            throws Exception {
         CorfuRuntime r = getRuntime().connect();
 
         UUID streamA = UUID.nameUUIDFromBytes("stream A".getBytes());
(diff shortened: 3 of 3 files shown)


zjohnny226:
Map<Long, LogData> logResult = CFUtils.getUninterruptibly(layout
                .getLogUnitClient(startAddress, numUnits - 1)
                .read(null, range)).getReadSet();

This is a single RPC.
The hole filling is one by one.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -61,6 +66,42 @@ public ILogData peek(Layout layout, long globalAddress) {
...
+        long startAddress = range.lowerEndpoint();
+        long endAddress = range.upperEndpoint();
+        int numUnits = layout.getSegmentLength(startAddress);
+        log.trace("ReadAll[{}-{}]: chain {}/{}", startAddress, endAddress, numUnits, numUnits);
+
+        Map<Long, LogData> logResult = CFUtils.getUninterruptibly(layout
+                .getLogUnitClient(startAddress, numUnits - 1)
+                .read(null, range)).getReadSet();
+
+        //in case of a hole, do a normal read and use its hole fill policy
```
Maithem: The hole-filling part seems like it would be common code, move that up?

Maithem:
@zjohnny226 right, sorry didn't see that the if statement was checking for null. 

You might want to change the read request type to reflect both modes of reading. You can add an extra flag field to determine whether the range type represents an actual range or disjoint ranges (i.e. random reads).

[0,5] (with flag true) means read {0,1,2,3,4,5}
[0,5] (with flag false) means read {0,5}

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
Maithem: also why not bulk hole fill too?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **2087502**.

 [Pull Request #725 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/vbLkgB8J1RYs7mMQn5Yf5nOatzayZ1mS)

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
rogermichoud: I would need the LogUnitClient.read(null, range) to return a TreeMap instead of a hashMap.

The change (hack ?) I have done on my side is the following in IReplicationProtocol.java:

```
-    default @Nonnull Map<Long, ILogData> readAll(Layout layout, Set<Long> globalAddresses) {
+    default @Nonnull TreeMap<Long, ILogData> readAll(Layout layout, Set<Long> globalAddresses) {
                 .map(a -> new AbstractMap.SimpleImmutableEntry<>(a, read(layout, a)))
-                .collect(Collectors.toMap(r -> r.getKey(), r -> r.getValue()));
+                .collect(Collectors.toMap(r -> r.getKey(), r -> r.getValue(),(v1, v2)-> v1, TreeMap::new));
```

Then I had to propagate the TreeMap everywhere. I am sure there is a more elegant way to do that.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -61,6 +66,42 @@ public ILogData peek(Layout layout, long globalAddress) {
...
+     * @return                      A map of addresses to committed
+     *                              addresses, hole filling if necessary.
+     */
+    @Override
+    public Map<Long, ILogData> readAll(Layout layout, Set<Long> globalAddresses) {
+        Range<Long> range = Range.encloseAll(globalAddresses);
+        long startAddress = range.lowerEndpoint();
+        long endAddress = range.upperEndpoint();
+        int numUnits = layout.getSegmentLength(startAddress);
+        log.trace("ReadAll[{}-{}]: chain {}/{}", startAddress, endAddress, numUnits, numUnits);
```
dahliamalkhi: minor nit: small 'r' in ReadAll

zjohnny226: Capital letter is in accordance to the style guideline. Though I just happened to copy and changed the trace message from read(). Another method in the same class has a lower case...

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
zjohnny226: 1. Bulk hole fill doesn't exist yet. Message only takes one address. I can make that change in a separate code change. Also, to be consistent, I will have to implement a fillAll() for the hole fill policy classes.
2. I don't like using TreeMaps instead of Map in the method signatures. In my implementation, I can certainly make sure that I use a TreeMap to hold the content.

zjohnny226:
Instead of having a flag to tell the server how to treat the range, I was thinking of modifying the requests to take in a list of Ranges.

I was thinking of breaking this down into smaller commits:
1. This, mainly to allow Roger to use bulk read.
2. Random read. Change Read request to use list of ranges.
3. Bulk hole fill. Hole fill message change, hole fill policy change.
4. Method signature changes as needed.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
zjohnny226: I just looked over the tests. Those are more integration tests rather than unit tests. I was expecting a bunch of unit tests with mocks in them.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
rogermichoud: Sounds good to me. As long as I can traverse it in order, I don't rally care.

rogermichoud:
The readAll doesn't work when the cache is enable. Tested with cache disabled, works fine.

Integration testing code:

```
CorfuRuntime runtime =  new CorfuRuntime("localhost:9000").connect();
runtime.getAddressSpaceView().read(LongStream.rangeClosed(0, 10)::iterator);
```

This is the stacktrace I have.

```
java.lang.IllegalStateException: stream has already been operated upon or closed

	at java.util.stream.AbstractPipeline.spliterator(AbstractPipeline.java:343)
	at java.util.stream.LongPipeline.spliterator(LongPipeline.java:178)
	at java.util.stream.LongPipeline.iterator(LongPipeline.java:173)
	at com.github.benmanes.caffeine.cache.LocalLoadingCache.loadInBulk(LocalLoadingCache.java:106)
	at com.github.benmanes.caffeine.cache.LocalLoadingCache.getAll(LocalLoadingCache.java:72)
	at org.corfudb.runtime.view.AddressSpaceView.read(AddressSpaceView.java:162)
```

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -61,6 +66,42 @@ public ILogData peek(Layout layout, long globalAddress) {
...
+        Map<Long, LogData> logResult = CFUtils.getUninterruptibly(layout
+                .getLogUnitClient(startAddress, numUnits - 1)
+                .read(null, range)).getReadSet();
+
+        //in case of a hole, do a normal read and use its hole fill policy
+        Map<Long, ILogData> returnResult = new HashMap<>();
+        for (Map.Entry<Long, LogData> entry: logResult.entrySet()){
+            ILogData value = entry.getValue();
+            if (value == null){
+                value = read(layout, entry.getKey());
```
rogermichoud: I think this happens only when we have hole(s). This is how he uses the hole fill policy associated with the single read. This is not optimal, but the plan is to get a bulk hole fill eventually.

zjohnny226: Fix for hole filling.

rogermichoud:
Never mind, it doesn't work because of the way I create an iterable (with ::iterator).

rogermichoud:
Looks good to me.

Could you squash your commit history please ? (https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines#git-commit-style)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a6904c1**.

 [Pull Request #725 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Z8WtIt54it534lJzD5whzkeVbTHm0aem)

dahliamalkhi:
@rogermichoud @Maithem @no2chem can we merge this already? 

@rogermichoud does it pass all tests already?

We then need to rebase and squash #726 and merge!

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
no2chem: If we need a tree map, we can return a NavigableMap, I think the best practice is to return the interface....

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -246,20 +249,24 @@ public void invalidateClientCache() {
...
+     * @return A result to be cached
      */
     private @Nonnull Map<Long, ILogData> cacheFetch(Iterable<Long> addresses) {
-        return StreamSupport.stream(addresses.spliterator(), true)
-                .map(address -> layoutHelper(l ->
-                        l.getReplicationMode(address).getReplicationProtocol(runtime)
-                        .read(l, address)))
-                .collect(Collectors.toMap(ILogData::getGlobalAddress, r -> r));
-    }
+        //turn the addresses into Set for now to satisfy signature requirement down the line
```
no2chem: What is the signature requirement "down the line"? Perhaps we should just change readAll to take an iterable rather than a set?

dahliamalkhi:
please rebase anda squash all commits (and address Michael's comment). Then, once it compiles, I'll merge.

zjohnny226:
Couple of questions:
1. Can we do a hole fill during a read, with a flag?
2. The current "ranged read" fails if any one address throws an exception. Shouldn't it be Map<address, result or exception>?
3. If we are to do a bulk hole fill, what should be the behaviour when one address fails?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
zjohnny226: I didn't want to make any changes to method signatures at this point, since I wasn't sure if there'd be any changes required when I implement random reads.

Maithem:
@zjohnny226 No you cant. Consider the following example: You read from the tail and the value is null, you tell the server to fill it with a hole in that same request, but now you can have consistency issues. Because If a non-hole value was written earlier in the chain, but not propagated to the tail, then you can end up in the case where the head and the tail have different values for the same addresses.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
Maithem: Unit tests don't have to have mocks. You need unit tests, since this is a bulk read on the address space, maybe you want to write your tests in AddressSpaceViewTest.

zjohnny226: I'm not saying unit tests need to have mocks. If your test space is sufficiently small, then you don't need mocks. For example, AddressSpaceViewTest::ensureStripingReadAllWorks() tests both write and read, which can be broken down if you want to.

Continuing with AddressSpaceViewTest::ensureStripingReadAllWorks() as an example. It already have a bulk read, but if we are using Mockito, you can verify that only 1 call is made to the server instead of 3.

Maithem:
@zjohnny226 I think it might be ok to have bulk reads with single address hole filling for now. @no2chem your thoughts ?

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
Maithem: Ok. You added new logic where you can do a bulk read, but then fill holes one by one. Can you write a test against that logic?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **ce5ecbd**.

 [Pull Request #725 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/96qsMLYXebBwPpkaehcFB2rVUsU5OxLa)

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
zjohnny226: I have no problem adding that test, but I just want to state that there is no good way to test the fact that we are making 1 call instead of 3 calls to the server. This new test would still pass on the old code.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **22d59a9**.

 [Pull Request #725 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/3SB05kS3Ypj1VVbI6S4OpJBj6U6ddvuQ)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **be58322**.

 [Pull Request #725 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/MFJZy9w7Bc0jIJ9wJRVMzQM5qWA5TATX)

Maithem:
Just rebase

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **f65b079**.

 [Pull Request #725 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/P4ig5Xbx5Kad5fiwV8URiu6yGP1X5OoW)

zjohnny226:
Need someone to merge this. Coveralls is telling me two different things.

PR #726, opened 2017-06-17
FastSmrMapsLoader without support of checkpoints/trim

rogermichoud:
First version of the FastSmrMapsLoader that will reconstruct all the SMR
Maps present in the log in one pass. It reads sequentially all entries
from 0 to log tail and dispatch the SMREntries to its map backed by a
Corfu stream. Every time it encounters a new stream, it creates a new Corfu
Object.

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
@@ -1,0 +1,336 @@
+package org.corfudb.recovery;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.collect.ContiguousSet;
+import com.google.common.collect.DiscreteDomain;
+import com.google.common.collect.Range;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import lombok.Data;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.logprotocol.LogEntry;
+import org.corfudb.protocols.logprotocol.MultiObjectSMREntry;
+import org.corfudb.protocols.logprotocol.SMREntry;
+import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.SMRMap;
+import org.corfudb.runtime.object.CorfuCompileProxy;
+import org.corfudb.runtime.object.ICorfuSMR;
+import org.corfudb.runtime.view.ObjectsView;
+
+/** The FastSmrMapsLoader reconstructs the coalesced state of SMRMaps through sequential log read
+ *
+ * This utility reads Log entries sequentially extracting the SMRUpdates from each entry
+ * and build the Maps as we go. In the presence of checkpoints, the checkpoint entries will
+ * be applied before the normal entries starting after the checkpoint start address.
+ *
+ * As current state, it doesn't support checkpoints/trim.
+ *
+ *
+ * Created by rmichoud on 6/14/17.
+ */
+
+@Slf4j
+@Accessors(chain = true)
+public class FastSmrMapsLoader {
+
+    static final long DEFAULT_BATCH_FOR_FAST_LOADER = 5;
+    static final int DEFAULT_TIMEOUT_MINUTES_FAST_LOADING = 30;
+    static final int STATUS_UPDATE_PACE = 10000;
+
+    private CorfuRuntime runtime;
+
+    @Setter
+    @Getter
+    private boolean loadInCache;
+
+    @Setter
+    @Getter
+    private long logHead = -1;
+
+    @Setter
+    @Getter
+    private long logTail = -1;
+
+    @Setter
+    @Getter
+    private long batchReadSize = DEFAULT_BATCH_FOR_FAST_LOADER;
+
+    @Setter
+    @Getter
+    private int timeoutInMinutesForLoading = DEFAULT_TIMEOUT_MINUTES_FAST_LOADING;
+
+
+    @Setter
+    @Getter
+    private boolean recoverSequencerMode;
+
+    private long addressProcessed = -1;
+
+    // In charge of summoning Corfu maps back in this world
+    private ExecutorService necromancer;
+
+    private Map<UUID, StreamMetaData> streamsMetaData;
+
+    @Getter
+    private Map<UUID, Long> streamTails = new HashMap<>();
+
+    public FastSmrMapsLoader(@Nonnull final CorfuRuntime corfuRuntime) {
+        this.runtime = corfuRuntime;
+        loadInCache = !corfuRuntime.cacheDisabled;
+        streamsMetaData = new HashMap<>();
+        necromancer = Executors.newSingleThreadExecutor(new ThreadFactoryBuilder()
+                .setNameFormat("necromancer-%d").build());
+    }
+
+    /** Check if this entry is relevant
+     *
+     * Entries before checkpoint starts are irrelevant since they are
+     * contained in the checkpoint.
+     *
+     * @param streamId identifies the Corfu stream
+     * @param entry entry to potentially apply
+     * @return if we need to apply the entry.
+     */
+    private boolean shouldEntryBeApplied(UUID streamId, SMREntry entry) {
+        if (!streamsMetaData.containsKey(streamId)) {
+            return true;
+        }
+
+        return (entry.getEntry().getGlobalAddress() >=
+                streamsMetaData.get(streamId).getHeadAddress());
+    }
+
+    private ObjectsView.ObjectID getObjectIdFromStreamID(UUID streamId) {
+        return new ObjectsView.ObjectID(streamId, SMRMap.class);
+    }
+
+    /** Update the corfu object and it's underlying stream with the new entry.
+     *
+     * @param streamId identifies the Corfu stream
+     * @param entry entry to apply
+     */
+    private void applySmrEntryToStream(UUID streamId, SMREntry entry) {
+        if (shouldEntryBeApplied(streamId, entry)) {
+            createSmrMapIfNotExist(streamId);
+
+            ObjectsView.ObjectID thisObjectId = new ObjectsView.ObjectID(streamId, SMRMap.class);
+            CorfuCompileProxy cp = ((CorfuCompileProxy) ((ICorfuSMR) runtime.getObjectsView().getObjectCache().get(thisObjectId)).
+                    getCorfuSMRProxy());
+
+            cp.getUnderlyingObject().applyUpdateToStreamUnsafe(entry);
+        }
+    }
+
+    /** Fetch LogData from Corfu server
+     *
+     * @param address address to be fetched
+     * @return LogData at address
+     */
+    private ILogData getLogData(long address) {
+        if (loadInCache) {
+            return runtime.getAddressSpaceView().read(address);
+        } else {
+            return runtime.getAddressSpaceView().fetch(address);
+        }
+    }
+
+    /** Get a range of LogData from the server
+     *
+     * This is using the underlying bulk read implementation for
+     * fetching a range of addresses. This read will return
+     * a map ordered by address.
+     *
+     * It uses a ClosedOpen range : [start, end)
+     * (e.g [0, 5) == (0,1,2,3,4))
+     *
+     * @param start start address for the bulk read
+     * @param end end address for the bulk read
+     * @return logData map ordered by addresses (increasing)
+     */
+    private Map<Long, ILogData> getLogData(long start, long end) {
+        return runtime.getAddressSpaceView()
+                .cacheFetch(ContiguousSet.create(Range.closedOpen(start, end), DiscreteDomain.longs()));
+    }
+
+
+    /** Create a new object SMRMap as recipient of SMRUpdates (if doesn't exist yet)
+     *
+     * @param streamId
+     */
+    private void createSmrMapIfNotExist(UUID streamId) {
+        if (!runtime.getObjectsView().getObjectCache().containsKey(getObjectIdFromStreamID(streamId))) {
+            runtime.getObjectsView().build()
+                    .setStreamID(streamId)
+                    .setType(SMRMap.class)
+                    .open();
+        }
+    }
+
+    /** Deserialize a logData by getting the logEntry
+     *
+     * Getting the underlying logEntry should trigger deserialization only once.
+     * Next access should just returned the logEntry direclty.
+     *
+     * @param logData
+     * @return
+     * @throws Exception
+     */
+    public LogEntry deserializeLogData(ILogData logData) throws Exception{
+        return logData.getLogEntry(runtime);
+    }
+
+    /** Extract log entries from logData and update the Corfu Objects
+     *
+     * @param logData LogData received from Corfu server.
+     */
+    private void processLogData(ILogData logData) {
+
+        LogEntry logEntry;
+        try {
+            logEntry = deserializeLogData(logData);
+        } catch (Exception e) {
+            log.error("Cannot deserialize log entry" + logData.getGlobalAddress(), e);
+            return;
+        }
+
+        if (logEntry.getType() == LogEntry.LogEntryType.SMR) {
+            // Just one stream, always
+            UUID streamId = logData.getStreams().iterator().next();
+            applySmrEntryToStream(streamId, (SMREntry) logEntry);
+        }
+        else if (logEntry.getType() == LogEntry.LogEntryType.MULTIOBJSMR) {
+            MultiObjectSMREntry multiObjectLogEntry = (MultiObjectSMREntry) logEntry;
+            multiObjectLogEntry.getEntryMap().forEach((stream, multiSmrEntry) -> {
+                multiSmrEntry.getSMRUpdates(stream).forEach((smrEntry) -> {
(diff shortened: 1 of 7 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **2f5a601**.

 [Pull Request #726 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/2VBNnj6KYj9JfEMAwOJnVeSiUNu3vJO2)

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,209 @@
...
+
+            switch (currentLogData.getType()) {
+                case DATA:
+                    processLogData(currentLogData);
+                    break;
+                case HOLE:
+                    break;
+                case TRIMMED:
+                    log.warn("Reading Trimmed Entry at address " + i +
+                            " Head of the log is set to " + logHead);
```
dahliamalkhi: according to the new style guidelines, this printout should be something like:
"loadMaps[{}, start={}] address is trimmed", i, logHead 

todo when stream registry is completed: find where the trim offset is, in order NOT to single-step an entire range of trimmed addresses.

rogermichoud: There is an option in our log format to print the method as well. Should we have the developers write this down if we can have it for free? https://logback.qos.ch/manual/layouts.html

The only concern is performance. Using reflection to get the name of the function might be costly.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **148a29a**.

 [Pull Request #726 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ROVqci3tr9X8r3LrcsgjKzOFB0kqE79l)

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -446,6 +481,13 @@ public synchronized CorfuRuntime connect() {
                 throw new RuntimeException(e);
             }
         }
+
+        if (loadSmrMapsAtConnect) {
+            FastSmrMapsLoader fastLoader = new FastSmrMapsLoader(this)
+                    .setBatchReadSize(bulkReadSizeForFastLoader)
+                    .setTimeoutInMinutesForLoading(timeoutInMinutesForFastLoading);
+            fastLoader.loadMaps();
+        }
```
dahliamalkhi: actually, thinking about this more: connect doesn't return the runtime until after loadMaps() completes, so actually there is no risk of activity under loadMaps() is done, no? Am I missing something?

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,319 @@
...
+     * Pretty useful for understanding how the the db is being used
+     * from the application perspective.
+     *
+     * @param logData
+     */
+    private void printLogAnatomy(ILogData logData) {
+        try {
+            LogEntry le = deserializeLogData(logData);
+            if (le.getType() == LogEntry.LogEntryType.SMR) {
+                System.out.println("Number of Streams: 1");
```
dahliamalkhi: need to remove this before merging.

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,319 @@
...
+            necromancer.execute(() -> {
+                range.forEach((address, logData) -> {
+                    switch (logData.getType()) {
+                        case DATA:
+                            processLogData(logData);
+                            break;
+                        case HOLE:
+                            break;
+                        case TRIMMED:
+                            log.warn("loadMaps[{}, start={}] address is trimmed", address, logHead);
```
dahliamalkhi: should we restart in this case?

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,319 @@
...
+                });
+            });
+        }
+
+        necromancer.shutdown();
+
+        try {
+            necromancer.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
+        } catch (InterruptedException e) {
+            log.warn("loadMaps is taking to long to load the maps. Gave up.");
```
dahliamalkhi: to -> too

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
rogermichoud: For now, not. There is currently no way to get the head of the log. Maithem is working on it. If we get the head of the log and we encounter a trim, it might be okay to just go over it. Since we are applying entries sequentially, not applying a TRIM and just go over it is okay.

I will address this (if I have to) in the second pass whith trimm/checkpoint support

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
rogermichoud: I agree with you. I was not sure if I was not missing something on my side. Since connect is synchronized, we should be safe.

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
rogermichoud: This function is just a utility function. It is not called during the normal process. I left it there for printing the Log Anatomy of a log if somebody want to use it in development.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -141,6 +144,38 @@
...
+    @Getter
+    private boolean loadSmrMapsAtConnect = false;
+
+
+    /**
+     * Set the bulk read size for the Fast Laoder
+     */
+    @Setter
+    @Getter
+    private long bulkReadSizeForFastLoader = DEFAULT_BATCH_FOR_FAST_LOADER;
```
no2chem: Can you make all of these fields part of CorfuRuntimeParameters?

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,323 @@
...
+     *
+     * Print how many streams are contained in the Metadata and
+     * how many entries per stream.
+     *
+     * Pretty useful for understanding how the the db is being used
+     * from the application perspective.
+     *
+     * @param logData
+     */
+    private void printLogAnatomy(ILogData logData) {
```
no2chem: Hm, interesting. Can this be moved to Utils? It's not necessarily a part of the map loader, but seems like a useful utility for development.

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,323 @@
...
+     * @return
+     * @throws Exception
+     */
+    private LogEntry deserializeLogData(ILogData logData) throws Exception{
+        LogEntry logEntry;
+        try {
+            logEntry = logData.getLogEntry(runtime);
+        } catch (Exception e) {
+            log.warn("Cannot deserialize log entry" + logData.getGlobalAddress());
+            e.printStackTrace();
```
no2chem: include the stacktrace using log, not with printStackTrace

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,323 @@
...
+    // In charge of summoning Corfu maps back in this world
+    private ExecutorService necromancer;
+
+    private Map<UUID, StreamMetaData> streamsMetaData;
+
+    public FastSmrMapsLoader(@Nonnull final CorfuRuntime corfuRuntime) {
+        this.runtime = corfuRuntime;
+        loadInCache = !corfuRuntime.cacheDisabled;
+        streamsMetaData = new HashMap<>();
+        necromancer = Executors.newSingleThreadExecutor();
```
no2chem: Could you set the name of this executor so it's easier to debug?

runtime/src/main/java/org/corfudb/runtime/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,323 @@
+package org.corfudb.runtime;
```
no2chem: could we make this part of something like package "org.corfudb.runtime.recovery"?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
rogermichoud: Well, if I put them in the CorfuRuntimeParameters, i cannot chain them nicely with the corfuRuntime instantiation. If I can, I am not sure how. I searched for example of the usage of these corfuParameters I didn't really find any.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **732632e**.

 [Pull Request #726 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/sz7o0LO6ZexKbuG23WtmdWNiyv8NbbWR)

dahliamalkhi:
please rebase this branch and squash all commits.  once it compiles, I'll merge

no2chem:
This is failing because of:

[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/travis/build/CorfuDB/CorfuDB/test/src/test/java/org/corfudb/integration/StreamIT.java:[6,26] error: cannot find symbol

rogermichoud:
The test that fails is the canReadHoles. It is because the current implementation of the bulk read doesn't support hole. This will be fixed with Johnny's PR.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
vjeko: I am not sure I understand why this is bundled with `connect()` and not exposed as a separate API. Right now we cannot differentiate between network related issues (TCP timeouts for example) and slow `loadSmrMapsAtConnect()`.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **02e26c4**.

 [Pull Request #726 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/FjfCvO7BJys4suHDRA66Sxt0139NEGpx)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **571179b**.

 [Pull Request #726 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/SthSHyQaT3omNvxDXeeiU8Aj1At5Hu8G)

runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,328 @@
...
+    /** Entry point to load the SMRMaps in memory.
+     *
+     * When this function returns, the maps are fully loaded.
+     *
+     */
+    public void loadMaps() {
+        log.info("loadMaps: Starting to resurrect maps");
+        initializeHeadAndTails();
+
+        long nextRead = 0;
```
Maithem: Can you add a ticker that is increased on every entry that is processed so that you guarantee that the whole address space (head to tail) is processed, I'm worried that there could be off-by-one problems

runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,328 @@
...
+
+                    }
+                });
+            });
+        }
+
+        necromancer.shutdown();
+
+        try {
+            necromancer.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```
Maithem: I don't think waiting forever is a good thing, maybe have this as a parameter?

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -372,4 +377,36 @@ public static void sleepUninterruptibly(long millis) {
...
+     * from the application perspective.
+     *
+     * @param logData
+     */
+    private void printLogAnatomy(CorfuRuntime runtime, ILogData logData) {
+        FastSmrMapsLoader fastLoader = new FastSmrMapsLoader(runtime);
+        try {
+            LogEntry le = fastLoader.deserializeLogData(logData);
+            if (le.getType() == LogEntry.LogEntryType.SMR) {
+                System.out.println("Number of Streams: 1");
```
Maithem: Use logging instead of println?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -141,6 +145,38 @@
...
+    /**
+     * How much time the Fast Loader has to get the maps up to date
+     *
+     * Once the timeout is reached, the Fast Loader give up. Every maps that are
+     * not up to date will be loaded through normal path.
+     *
+     */
+    @Getter
+    @Setter
+    private int timeoutInMinutesForFastLoading = DEFAULT_TIMEOUT_MINUTES_FAST_LOADING;
```
Maithem: This is never used

test/src/test/java/org/corfudb/recovery/FastSmrMapsLoaderTest.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Created by rmichoud on 6/16/17.
+ */
+public class FastSmrMapsLoaderTest extends AbstractViewTest {
+
```
Maithem: I think the tests should test cases where you process more than a multiple of the batch size

runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /** Initialize log head and log tails
+     *
+     * If logHead and logTail has not been initialized by
+     * the user, initialize to default.
+     *
+     */
+    private void initializeHeadAndTails() {
+        if (logHead < 0) {
```
rogermichoud: For the recovery of the sequencer, we cannot rely on the sequencer to have the logHead and the logtail. this needs to be parametrize.

rogermichoud:
Addressed all Maithem comments.

runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
Maithem: I wasn't implying that you should rely on the sequencer. I was talking about the information flow, it seems like you can acquire both the log head and log tail from within the class, why make this configurable externally ?

runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,324 @@
...
+        long nextRead = 0;
+        while (nextRead <= logTail) {
+            final long start = nextRead;
+            final long stopNotIncluded = Math.min(start + batchReadSize, logTail + 1);
+            nextRead = stopNotIncluded;
+            final Map<Long, ILogData> range = getLogData(start, stopNotIncluded);
+
+            necromancer.execute(() -> {
+                range.forEach((address, logData) -> {
+                    addressProcessed++;
```
Maithem: This ticker just captures the number of entries, can you make it ensure that you are iterating over the addresses space monotonically (i.e. start, start +1, start + 2, start +3, etc...).

test/src/test/java/org/corfudb/recovery/FastSmrMapsLoaderTest.java
```diff
@@ -0,0 +1,244 @@
...
+     * @throws Exception
+     */
+    @Test
+    public void canReloadMaps() throws Exception {
+        CorfuRuntime rt1 = getDefaultRuntime();
+        Map<String, String> map1 = createMap("Map1", rt1);
+        Map<String, String> map2 = createMap("Map2", rt1);
+
+        map1.put("k1", "v1");
+        map1.put("k2", "v2");
```
Maithem: Can you use a loop and create a history that is a multiple of the batch size.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **afcc102**.

 [Pull Request #726 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/l8ZYdZiiwExl6qOLfSKDPksaKd4jya8f)

runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
```diff
@@ -0,0 +1,324 @@
+package org.corfudb.recovery;
+
+import javax.annotation.Nonnull;
+import java.util.*;
```
Maithem: Expand imports ?

rogermichoud:
@Maithem Addressed all your comments.

Maithem:
@rogermichoud ok, thanks.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -141,6 +141,18 @@
     private String usernameFile;
     private String passwordFile;
 
+
+    /**
+     * Trigger the loading of all the SmrMaps upon connect.
+     *
+     * If using this utility, you need to be sure that no one
+     * is acessing objects until the tables are loaded
+     * (i.e. when connect return)
```
medhavidhawan: I think this should be more runtime level fencing and not across runtimes. This event is contained within a runtime.

PR #728, opened 2017-06-19
Checkstyle fixups for org.corfudb.security.*

slfritchie:
Checkstyle error fixes: comments, indentation, etc. for org.corfudb.security.*

### DIFF
--- a/runtime/src/main/java/org/corfudb/security/sasl/plaintext/PlainTextCallbackHandler.java
+++ b/runtime/src/main/java/org/corfudb/security/sasl/plaintext/PlainTextCallbackHandler.java
@@ -1,15 +1,15 @@
 package org.corfudb.security.sasl.plaintext;
 
 import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.NameCallback;
 import javax.security.auth.callback.PasswordCallback;
-import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.UnsupportedCallbackException;
 
 /**
  * Created by sneginhal on 01/27/2017
  *
- * Common callback handler for NameCallback and PasswordCallback.
+ * <p>Common callback handler for NameCallback and PasswordCallback.
  * Used by PlainTextLoginModule and PlainTextSaslNettyClient.
  */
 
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/security/sasl/plaintext/PlainTextLoginModule.java
```diff
@@ -3,19 +3,19 @@
...
 import javax.security.auth.callback.UnsupportedCallbackException;
 import javax.security.auth.login.LoginException;
 import javax.security.auth.spi.LoginModule;
-import javax.security.auth.Subject;
 
 /**
  * Created by sneginhal on 01/27/2017
  *
- * Implementation of the plain text LoginMoodule.
+ * <p>Implementation of the plain text LoginMoodule.
```
no2chem: needs a end </p> tag?

runtime/src/main/java/org/corfudb/security/sasl/plaintext/PlainTextCallbackHandler.java
```diff
@@ -1,15 +1,15 @@
...
 import javax.security.auth.callback.NameCallback;
 import javax.security.auth.callback.PasswordCallback;
-import javax.security.auth.callback.CallbackHandler;
 import javax.security.auth.callback.UnsupportedCallbackException;
 
 /**
  * Created by sneginhal on 01/27/2017
  *
- * Common callback handler for NameCallback and PasswordCallback.
+ * <p>Common callback handler for NameCallback and PasswordCallback.
```
no2chem: also needs a end /p tag?

no2chem:
Closes #700

runtime/src/main/java/org/corfudb/security/sasl/plaintext/PlainTextLoginModule.java
slfritchie: @no2chem  According to http://www.oracle.com/technetwork/articles/java/index-137868.html, the doc text is HTML (some closing tags are optional, like p's) and not XHTML (no optional closing tags).

I put the <p> at the start of the next paragraph.  The Oracle doc examples putting it on the "blank" line.  The Google style guide https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs says to use the style I've used here.  It also says nothing about XHTML compliance or HTML optional tags, alas.

Thoughts?

no2chem: Ah interesting, didn't realize that. Since that's the style from the google style guide, let's go with that.

PR #729, opened 2017-06-19
Add shutdown hooks which call shutdown() in each server.

no2chem:
This PR adds a shutdown hook, which calls shutdown() on each
server so they may shut down gracefully.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
@@ -301,6 +301,9 @@ public class CorfuServer {
         });
 
 
+        // Register shutdown handler
+        Runtime.getRuntime().addShutdownHook(new Thread(CorfuServer::cleanShutdown));
+
         try {
             ServerBootstrap b = new ServerBootstrap();
             b.group(bossGroup, workerGroup)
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -344,13 +347,24 @@ public void initChannel(io.netty.channel.socket.SocketChannel ch) throws Excepti
...
             bossGroup.shutdownGracefully();
             workerGroup.shutdownGracefully();
         }
 
     }
 
+    /** Attempt to cleanly shutdown all the servers. */
+    public static void cleanShutdown() {
+        log.info("CleanShutdown: Starting Cleanup.");
+        sequencerServer.shutdown();
```
Maithem: What if one of the sequencers throws an exception during shutdown ?

Would't it be better to loop through the servers calling each shutdown with a try/catch ?

no2chem:
@Maithem how about now, it's now nicely multithreaded with correct thread names as well.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **1fdfd05**.

 [Pull Request #729 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/5Qay6cGK0eQTojAqydM2jmX6zjNC053e)

PR #738, opened 2017-06-19
Checkstyle fixes

kjames88:


no2chem:
Thanks @kjames88 ! It looks like you just need to remove checks.xml from this commit.

kjames88:
removed checks.xml and rebased

no2chem:
Closes #687

PR #739, opened 2017-06-19
Update package Clients in module Runtime to new style guidelines.

zalokhan:
Updating package org.corfudb.runtime.clients with new style guidelines.


no2chem:
Closes #692

PR #740, opened 2017-06-19
Update package Objects in module Runtime to new style guidelines.

zalokhan:
package org.corfudb.runtime.object


no2chem:
Closes #696

PR #751, opened 2017-06-21
Introduce Stream Options

Maithem:
This patch introduces options that can be passed when a stream
is opened. One such option is to ignore addresses when
backpointers are traversed.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/StreamOptions.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/StreamOptions.java
@@ -1,0 +1,35 @@
+package org.corfudb.runtime.view;
+
+/**
+ * Created by maithem on 6/20/17.
+ */
+public class StreamOptions {
+    public static StreamOptions DEFAULT = new StreamOptions(false);
+
+    public final boolean ignoreTrimmed;
+
+    public StreamOptions(boolean ignoreTrimmed) {
+        this.ignoreTrimmed = ignoreTrimmed;
+    }
+
+    public static StreamOptionsBuilder builder() {
+        return new StreamOptionsBuilder();
+    }
+
+    public static class StreamOptionsBuilder {
+        private boolean ignoreTrimmed;
+
+        public StreamOptionsBuilder() {
+
+        }
+
+        public StreamOptionsBuilder ignoreTrimmed(boolean ignore) {
+            this.ignoreTrimmed = ignore;
+            return this;
+        }
+
+        public StreamOptions build() {
+            return new StreamOptions(ignoreTrimmed);
+        }
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
@@ -8,6 +8,7 @@ import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.view.Address;
+import org.corfudb.runtime.view.StreamOptions;
 import org.corfudb.util.Utils;
 
 import javax.annotation.Nonnull;
@@ -26,14 +27,24 @@ import java.util.stream.Collectors;
 @Slf4j
 public class BackpointerStreamView extends AbstractQueuedStreamView {
 
+    final StreamOptions options;
+
     /** Create a new backpointer stream view.
      *
      * @param runtime   The runtime to use for accessing the log.
      * @param streamID  The ID of the stream to view.
      */
     public BackpointerStreamView(final CorfuRuntime runtime,
+                                 final UUID streamID,
+                                 @Nonnull final StreamOptions options) {
+        super(runtime, streamID);
+        this.options = options;
+    }
+
+    public BackpointerStreamView(final CorfuRuntime runtime,
                                  final UUID streamID) {
         super(runtime, streamID);
+        this.options = StreamOptions.DEFAULT;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -45,6 +47,37 @@ public void canReadWriteFromStream()
         assertThat(sv.next())
                 .isEqualTo(null);
     }
+    // test with option false
```
dahliamalkhi: This comment seems wrong

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -45,6 +47,37 @@ public void canReadWriteFromStream()
...
+
+    @Test
+    public void testRemainingUpToWithTrim() {
+        StreamOptions options = new StreamOptions(true);
+        IStreamView txStream = runtime.getStreamsView().get(ObjectsView.TRANSACTION_STREAM_ID, options);
+        final int firstIter = 50;
```
dahliamalkhi: use PARAMETERS for this?

Maithem:
@dahliamalkhi Good catch. Addressed all comments.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **1ff7f7d**.

 [Pull Request #751 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/lSPcbR0zCRNXrvADUXhnkg2B9yQo4HBT)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **9a353cd**.

 [Pull Request #751 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ikVyY5raJFbHLjzJFXNJekh0KqW0BfYC)

runtime/src/main/java/org/corfudb/runtime/view/StreamOptions.java
```diff
@@ -0,0 +1,16 @@
+package org.corfudb.runtime.view;
+
+/**
+ * Created by maithem on 6/20/17.
+ */
+public class StreamOptions {
```
vjeko: From a usability perspective, `new StreamOptions(true)` is not very intuitive, especially with multiple options. Either the fluent pattern or setters would be much nicer.

no2chem: Agree, can we modify this to be fluent?

runtime/src/main/java/org/corfudb/runtime/view/StreamOptions.java
```diff
@@ -0,0 +1,16 @@
+package org.corfudb.runtime.view;
+
+/**
+ * Created by maithem on 6/20/17.
```
no2chem: Also, create a static DEFAULT value, so we can supply StreamOptions.DEFAULT.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -26,14 +27,24 @@
...
+    final StreamOptions options;
+
     /** Create a new backpointer stream view.
      *
      * @param runtime   The runtime to use for accessing the log.
      * @param streamID  The ID of the stream to view.
      */
     public BackpointerStreamView(final CorfuRuntime runtime,
+                                 final UUID streamID,
+                                 final StreamOptions options) {
```
no2chem: and this should be annotated with @NonNull

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -26,14 +27,24 @@
...
+        super(runtime, streamID);
+        this.options = options;
+    }
+
+    public BackpointerStreamView(final CorfuRuntime runtime,
                                  final UUID streamID) {
         super(runtime, streamID);
+        this.options = new StreamOptions();
```
no2chem: So this should use .DEFAULT in the future

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -47,6 +49,36 @@ public void canReadWriteFromStream()
     }
 
     @Test
+    public void testRemainingUpToWithTrim() {
```
no2chem: Needs a javadoc

runtime/src/main/java/org/corfudb/runtime/view/StreamOptions.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+public class StreamOptions {
+    public final boolean ignoreTrimmed;
+
+    public StreamOptions(boolean ignoreCheckpoints) {
```
no2chem: ignoreTrimmed and ignoreCheckpoints appear to be different.

runtime/src/main/java/org/corfudb/runtime/view/StreamOptions.java
Maithem: Good idea.

runtime/src/main/java/org/corfudb/runtime/view/StreamOptions.java
Maithem: done

Maithem:
@no2chem @vjeko addressed comments, please review.

runtime/src/main/java/org/corfudb/runtime/view/StreamOptionsBuilder.java
```diff
@@ -0,0 +1,22 @@
+package org.corfudb.runtime.view;
+
+
+/**
+ * Created by maithem on 6/21/17.
+ */
+public class StreamOptionsBuilder {
+    private boolean ignoreTrimmed;
+
+    public StreamOptionsBuilder() {
```
no2chem: Ok, I guess one more nit, if we're going with the builder paradigm

StreamOptionsBuilder should just be a static class inside StreamBuilder
StreamOptions.builder() should give you a StreamOptionsBuilder

This is the paradigm that Guava follows (like ImmutableMap.builder().)

Maithem:
@no2chem ok I addressed your comments, hope its better now.

PR #752, opened 2017-06-21
Runtime hint for stream tails to the Sequencer

rogermichoud:
This API is principally meant to be used for recovery purposes. The runtime
can propose hit stream tails to the sequencer. If the sequencer doesn't have
a better information, it will update its stream tails to the hint.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -316,6 +316,34 @@ public class SequencerServer extends AbstractServer {
     }
 
     /**
+     * Consider a hint given by a runtime about the loction of the stream tails
+     *
+     * If the stream tail hinted is lower that the current one, the hint will be discarded.
+     *
+     * @param msg
+     * @param ctx
+     * @param r
+     * @param isMetricsEnabled
+     */
+    @ServerHandler(type=CorfuMsgType.TAILS_HINT, opTimer=metricsPrefix + "tails-hint")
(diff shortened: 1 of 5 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **0e6dc91**.

 [Pull Request #752 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/fmYw5HY2z6KaIuFoeJ1BE6GkZtLqA41e)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -316,6 +316,34 @@ public synchronized void resetServer(CorfuPayloadMsg<Long> msg, ChannelHandlerCo
...
+     * Consider a hint given by a runtime about the loction of the stream tails
+     *
+     * If the stream tail hinted is lower that the current one, the hint will be discarded.
+     *
+     * @param msg
+     * @param ctx
+     * @param r
+     * @param isMetricsEnabled
+     */
+    @ServerHandler(type=CorfuMsgType.TAILS_HINT, opTimer=metricsPrefix + "tails-hint")
```
dahliamalkhi: please change this to a "conservative" approach: send the global log tail along with the hint. If the current global tail moved beyond it, ignore this hint.

I am concerned about a slow hint that arrives after more entries are added to a stream, but the sequencer fails. Example:

stream has entries 1, 2
system crashes
on recovery, client discovers stream tail is 2 , but goes to sleep
entry 3 is added
sequencer fails and recovers
client sends hint 2 to new sequencer
entry 4 is added with backpointer to 2

rogermichoud:
Correct. Discard this PR. We are moving this in the reset sequencer message. Management server will reconstruct the tails on sequencer failover.

Issue #753, opened 2017-06-21
Fix hashing schme


no2chem:
Currently, the scheme for hashing uses the Java hashCode, which may result in a high number of collisions. The hashing scheme needs to be updated to one which is more resistant to collisions.

no2chem:
This was partially addressed by #759, but we are still using hashCode, need to provide a stronger hash function.

PR #759, opened 2017-06-22
Use a two-level map for conflicts, support poisoning

no2chem:
This PR partially addresses #753 and reduces the number of
false conflicts. 

It introduces a two-level conflict map, the first level
maps stream ids to a new per-stream data class. The 
per-stream data class contains a map of conflict parameters
(which now can be longs) to addresses, as well as a
single long which represents a minimum "poisoned" value.

The implementation still uses a Java hashCode, (32 bits),
but this is much improved over the previous implementation
which attempted to squeeze the stream ID and a java hashCode
of the conflict param into 32 bits. In order to use a
longer hash code, we'll need to find a generic way to
generate a long hashCode from an object without serializing
it.

Poisoned streams are streams affected by operations with
fine-grained conflict information, for example, operations
such as clear(). Previously, these operations would
not conflict at all, when they should have conflicted with
all operations.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -2,23 +2,21 @@ package org.corfudb.infrastructure;
 
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.MetricRegistry;
-import com.github.benmanes.caffeine.cache.Cache;
-import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.google.common.collect.ImmutableMap;
 
 import io.netty.channel.ChannelHandlerContext;
 
 import java.lang.invoke.MethodHandles;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 
+import lombok.Data;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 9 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **14ac84f**.

 [Pull Request #759 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/12vs4xxU3PpclcyRvcTy6F2PhpBAldpS)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -125,20 +123,43 @@
...
+         *                              the conflict parameter.
+         */
+        Long getLatestUpdate(long conflictParameter) {
+            Long mapAddress = conflictMap.get(conflictParameter);
+            if (mapAddress == null) {
+                if (minPoisoned != Address.NEVER_READ) {
+                    // No conflict info, but the stream was poisoned.
+                    return minPoisoned;
+                } else {
+                    // Otherwise, the code expects to get a null,
```
dahliamalkhi: this comment is meaningless "the code expects to get a null"

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -485,26 +495,40 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
...
         // update the cache of conflict parameters
         if (req.getTxnResolution() != null) {
+            req.getTxnResolution().getPoisonedStreams()
+                    .stream()
+                    .forEach(id -> {
+                        SequencerStreamInfo streamInfo =
+                                conflictToGlobalTailCache
+                                        .computeIfAbsent(id,
+                                                k -> new SequencerStreamInfo());
+                        streamInfo.minPoisoned = newTail - 1;
```
dahliamalkhi: how about clearing the conflict map here to reclaim the space? this way, when a map is cleared, the sequencer space is compacted. otherwise, it grows forever.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TxResolutionInfo.java
```diff
@@ -46,6 +50,7 @@ public TxResolutionInfo(UUID txId, long snapshotTimestamp) {
         this.snapshotTimestamp = snapshotTimestamp;
         this.conflictSet = Collections.emptyMap();
         this.writeConflictParams = Collections.emptyMap();
+        this.poisonedStreams = Collections.emptySet();
```
dahliamalkhi: isn't it better to make this null (cheaper serialization), and check if it is null on the receiving end? token request performance matters.

brajneesh:
@no2chem can we get this PR in today after addressing the review comments

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -204,7 +213,7 @@ public int getConflictHashCode(UUID streamId, int conflictParam) {
      *     cause.
      */
     public TokenType txnCanCommit(TxResolutionInfo txInfo, /** Input. */
```
no2chem: Yeah sure, not in the scope of the PR (and not my code) but we can address that in another PR.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
no2chem: Sure, that sounds like a good idea, I'll add that.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TxResolutionInfo.java
no2chem: Cheaper serialization? Doubt it. The cost of Collections.emptySet is pretty close to just returning null (it returns **the** immutable empty set) and it keeps the code modular, as the implementation will just serialize a map with size 0. On the other hand we probably need to special case the deserializer when it gets a map/set of size 0, but that's a relatively easy fix.

PR #778, opened 2017-06-28
Update logback, netty, caffeine, slf4j and surefire

no2chem:
This PR updates the versions of various components to their
latest versions.


no2chem:
mapsAsMqsTest has been intermittently failing for awhile. Perhaps a fix would be to disable these faulty tests on Travis for the time being.

dahliamalkhi:
note that useMapsAsMQs test has (a) surfaced important issues in the past, and (b) tries to emulate MP use case.

no2chem:
@dahliamalkhi It's not a particularly useful test if we can't tell if the failures are due to a bug in the test or the code. Either way it seems that @slfritchie may have fixed the test, could you please check his PR?

slfritchie:
It's PR #783, which fixes the bug described by #780, "useMapsAsMQs fails intermittently".

Issue #773, opened 2017-06-27
Logging file (not log data) is erased after restart corfu server


bitmingw:
Corfu uses [logback](https://logback.qos.ch/) to manage the logging generated by application. Currently, the logging file is erased after each restart of corfu server, which makes the debugging process difficult.

The configuration of logback is defined in `runtime/src/main/resources/logback.xml`. From the [docs](https://logback.qos.ch/manual/appenders.html) of logback, when `STDOUT` is used as the appender, a default file will be opened (e.g. `/var/log/corfu.9000.log`) and there is no way to preserver the logging from last run. As a result, we will need to switch to `FILE` appender. This process, thereafter, brings up two questions.

First, there are a couple of places use `System.out.print` to log message. Those usages need to clean up when we switch to file appender.

Second, we need to think about a good way to name logging files when there are multiple instances of corfu servers.

Maithem:
Most of the system.out.printl statements are in the sample code, they dont run on the server.
There are a couple of system out statements that are in CorfuServer.java that print the server options and logo, which should be easy to migrate to logging.

bitmingw:
Resolved by #782

PR #781, opened 2017-06-28
Persist logging after server reboot (#773)

bitmingw:
This patch modifies the logback configuration to save corfu
server logging after corfu server is restarted.

In the old configuration, all logging are treated to write to
stdout. Logback will create (and erase) a default logging file
at `/var/log/corfu.9000.log` and put all info there.

In the new configuration, `/var/log/corfu.log` is specified as
the place to hold logging info. By default FileAppender is persistent
after server reboot.

This patch also modifies the format such that date is included.

**Future works**

1. If the amount of logging is big over the time, we need to add a rolling
   policy to handle them.
2. Mac users benefits from colors. But it does harm to users with traditional
   vi, which is probably to be the log reader in a real server. We can discuss
   and see if it is better to remove colors.
3. After this patch, `/var/log/corfu.9000.log` will still be created to hold
   stdout stream. To make it cleaner we should remove `System.out.print` gradually.

### DIFF
--- a/runtime/src/main/resources/logback.xml
+++ b/runtime/src/main/resources/logback.xml
@@ -1,12 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <appender name="FILE" class="ch.qos.logback.core.FileAppender">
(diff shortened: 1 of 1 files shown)


bitmingw:
Don't review this. I have found a much cleaner implementation.

PR #797, opened 2017-07-03
Fix: LogUnit FD/Memory leak

Maithem:
This patch evicts segment handles that correspond to addresses
that have been prefix trimmed. As a result of this eviction,
memory and open file descriptors are cleaned.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -254,6 +254,8 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
             }
         } catch (IOException e) {
             throw new RuntimeException(e.getMessage(), e);
+        } finally {
+            sh.release();
         }
 
         lastSegment = tailSegment;
@@ -327,24 +329,28 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
     @Override
     public void trim(long address) {
         SegmentHandle handle = getSegmentHandleForAddress(address);
-        if (!handle.getKnownAddresses().containsKey(address)
-                || handle.getPendingTrims().contains(address)) {
-            return;
-        }
-        TrimEntry entry = TrimEntry.newBuilder()
-                .setChecksum(getChecksum(address))
-                .setAddress(address)
-                .build();
-
-        // TODO(Maithem) possibly move this to SegmentHandle. Do we need to close and flush?
-        OutputStream outputStream = Channels.newOutputStream(handle.getPendingTrimChannel());
         try {
+            if (!handle.getKnownAddresses().containsKey(address)
+                    || handle.getPendingTrims().contains(address)) {
+                return;
+            }
+
+            TrimEntry entry = TrimEntry.newBuilder()
+                    .setChecksum(getChecksum(address))
+                    .setAddress(address)
+                    .build();
+
+            // TODO(Maithem) possibly move this to SegmentHandle. Do we need to close and flush?
+            OutputStream outputStream = Channels.newOutputStream(handle.getPendingTrimChannel());
+
             entry.writeDelimitedTo(outputStream);
             outputStream.flush();
             handle.pendingTrims.add(address);
             channelsToSync.add(handle.getPendingTrimChannel());
         } catch (IOException e) {
             log.warn("Exception while writing a trim entry {} : {}", address, e.toString());
+        } finally {
+            handle.release();
         }
     }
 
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1008,11 +1046,24 @@ public LogHeader getLogHeader() {
...
         private Set<Long> trimmedAddresses = Collections.newSetFromMap(new ConcurrentHashMap<>());
         private Set<Long> pendingTrims = Collections.newSetFromMap(new ConcurrentHashMap<>());
+        private volatile int refCount = 0;
 
 
+        public synchronized void retain() {
+            refCount++;
+        }
+
+        public synchronized void release() {
```
slfritchie: I've a question about the refcounting and `StreamLogFiles::close`, which isn't paying attention to the refcounts.  AFAICT, that method is only being used by unit test code, and honoring the refcounts in those tests probably doesn't matter?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -372,20 +379,37 @@ private void trimPrefix() {
...
 
-        for (File file : files) {
-            if (!file.delete()) {
-                log.error("Couldn't delete/trim file {}", file.getName());
+                File logFile = new File(sh.getFileName());
+                File trimmedFile = new File(getTrimmedFilePath(sh.getFileName()));
+                File pendingFile = new File(getPendingTrimsFilePath(sh.getFileName()));
+
+                if (!logFile.delete()) {
+                    log.error("Couldn't delete/trim file {}", logFile.getName());
```
no2chem: Style: trimPrefix: Couldn't ...

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
```diff
@@ -417,6 +429,13 @@ public void testPrefixTrim() {
             }
         }
 
+        // Verify that the trimmed segment channels are close
```
no2chem: are closed

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -937,22 +965,27 @@ public void append(long address, LogData entry) {
         } catch (IOException e) {
             log.error("Disk_write[{}]: Exception", address, e);
             throw new RuntimeException(e);
+        } finally {
```
medhavidhawan: hopefully you have done the release in all places.

Maithem: I tested this with the l2 tests and lsof doesn't show any deleted entries. Let's see what happens on the other system

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: Done.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -329,22 +331,27 @@ public void trim(long address) {
         SegmentHandle handle = getSegmentHandleForAddress(address);
         if (!handle.getKnownAddresses().containsKey(address)
                 || handle.getPendingTrims().contains(address)) {
+            handle.release();
```
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: You are right, I do that intentionally. The rational is, by the time the segment is closed, the ref count should be 0 and if it is not, then "forcefully" close the segment. Close will take precedence over waiting for the refcount to reach 0.

Maithem:
Addressed all comments.

Maithem:
Addresses https://github.com/CorfuDB/CorfuDB/pull/797

Maithem:
@no2chem @medhavidhawan @slfritchie are you guys satisfied with the changes? can I merge?

PR #801, opened 2017-07-05
Add logical time expiry policy to AddressSpaceView's Caffeine cache

### LINKED ISSUE #790: Trimmed addresses should be evicted from the cache
When AddressSpaceView gets a trimmed exception, it should evict all trimmed entries from the cache.

slfritchie:
Fixes #790

Add a custom Ticker class to AddressSpaceView's Caffeine cache,
based on a notion of logical time specified by global log address.
The clock advances when a TrimmedException is witnessed.  See #790
for discussion.

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/LogicalTicker.java
+++ b/runtime/src/main/java/org/corfudb/util/LogicalTicker.java
@@ -1,0 +1,58 @@
+package org.corfudb.util;
+
+import com.github.benmanes.caffeine.cache.Expiry;
+import com.github.benmanes.caffeine.cache.Ticker;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.corfudb.protocols.wireprotocol.ILogData;
+
+/**
+ * A Ticker subclass for manipulating Caffeine cache expiry
+ * policy using logical time.
+ */
+
+public class LogicalTicker implements Ticker {
+    AtomicLong time;
+
+    // Caffeine's algorithm assumes fairly "wide" distances between
+    // times, e.g. nanoseconds, see commentary on GitHub PR 801.
+    private static final long scaleFactor = 1_000_000_000;
+
+    public LogicalTicker(AtomicLong time) {
+        this.time = time;
+    }
+
+    @Override
+    public long read() {
+        long time = this.time.get();
+        if (time < 0) {
+            return 0; // Don't report a negative clock
+        } else {
+            return time * scaleFactor;
+        }
+    }
+
+    /** Return an expiry policy based on logical time. */
+    public static Expiry<Long, ILogData> caffeineLogicalExpiryPolicy() {
+        // Use a policy of logical time, where the logical clock advances
+        // each time that we witness a TrimmedException.  The "time" for
+        // a cache entry's expiry is its log address.
+
+        return new Expiry<Long, ILogData>() {
+            public long expireAfterCreate(Long key, ILogData l, long currentTime) {
+                return (key * scaleFactor) - currentTime;
+            }
+
+            public long expireAfterUpdate(Long key, ILogData l,
+                                          long currentTime, long currentDuration) {
+                return (key * scaleFactor) - currentTime;
+            }
+
+            public long expireAfterRead(Long key, ILogData l,
+                                        long currentTime, long currentDuration) {
+                return (key * scaleFactor) - currentTime;
+            }
+        };
+    }
+}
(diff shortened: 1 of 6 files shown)


test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -382,4 +382,31 @@ public void prefixTrimThrowsException()
...
+        getRuntime().getAddressSpaceView().invalidateClientCache();
+        getRuntime().getAddressSpaceView().invalidateServerCaches();
+
+        // First witness of TrimmedException by AddressSpaceView::read()
+        assertThatThrownBy(() -> svB.next().getPayload(getRuntime()))
+                .isInstanceOf(TrimmedException.class);
+        // Test short-cut in AddressSpaceView::read() by trying
+        // to read an earlier log address than svB's attempt.
+        assertThatThrownBy(() -> svA.next().getPayload(getRuntime()))
+                .isInstanceOf(TrimmedException.class);
```
dahliamalkhi: shouldn't this succeed? we only trimmed from 1 onward

slfritchie: A's append is at address 0, B's append is at address 1.  This line is A's first attempt to read the stream, so the trim at 1 should trigger TrimmedException.

slfritchie:
After realizing there's effectively a 2nd place where this policy should be enforced, I've force-pushed commit 46473b2ed8829596defade972b69a24050800be9 to add the same caching policy to the server side cache in LogUnitServer.

runtime/src/main/java/org/corfudb/util/LogicalTicker.java
```diff
@@ -0,0 +1,49 @@
...
+            }
+
+            public long expireAfterUpdate(Long key, ILogData l,
+                                          long currentTime, long currentDuration) {
+                return key;
+            }
+
+            public long expireAfterRead(Long key, ILogData l,
+                                        long currentTime, long currentDuration) {
+                return key;
```
ben-manes: The return value is the duration until the item expires from the current time. So this extends the lifetime of the entry on every read or write by N. From your description, I think you might be only wanting it to be on creation so that entries expire after the clock advances past the threshold. I _think_ you might have thought of this as a point-in-time instead of duration, but am not familiar enough to know for certain.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **46473b2**.

 [Pull Request #801 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/BTbYMKcOFSphQF4bzxkNEP1ABFJ6sirl)

ben-manes:
I think you should multiply the logical time by a larger step size. The policy you are using relies on a [TimerWheel](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/TimerWheel.java) which uses time + hashing to be O(1). However that assumes time is moving at a reasonable pace, so the minimum bucket size is ~1 second in nanos. Since your increments are minuscule, the optimization breaks down to not turn the wheels and evict. This limitation isn't in a tree based solution like a heap, but those require O(lg n) time instead. Using a second or more per logical time unit would resolve this.

slfritchie:
Hi, @ben-manes, thanks for the advice.  I was indeed thinking about absolute time instead of a relative duration; I'll return to using duration.  I'll also add a multiplying factor as you suggest.

slfritchie:
I've put the changes discussed above into a new squashed commit, 001ec5a.  Thanks again, @ben-manes.

To follow up on the use of relative durations for expiry policy ... I returned to the very ugly hack at https://github.com/slfritchie/CorfuDB/blob/598f830092ac97c53c40da108c91987df9313585/test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java#L409-L509 and patched it up with the new LogicalTicker.  The patched-up hack no longer exhibits the "WTF, you're ignoring expiry!" behavior that I saw earlier this week, hooray.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **001ec5a**.

 [Pull Request #801 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/lG4fZMVf0Nmb2vWeJRnjm0reqqbBkh7x)

runtime/src/main/java/org/corfudb/util/LogicalTicker.java
```diff
@@ -0,0 +1,58 @@
...
+        // a cache entry's expiry is its log address.
+
+        return new Expiry<Long, ILogData>() {
+            public long expireAfterCreate(Long key, ILogData l, long currentTime) {
+                return (key * scaleFactor) - currentTime;
+            }
+
+            public long expireAfterUpdate(Long key, ILogData l,
+                                          long currentTime, long currentDuration) {
+                return (key * scaleFactor) - currentTime;
```
ben-manes: I think this should be the same as currentDuration, which is the remaining time to live. Its provided if you don't want to modify the expiration time.

xnull:
Not relevant anymore

PR #803, opened 2017-07-05
Fix sequencer recovery with fastSMRLoader

Maithem:
This patch moves important data for checkpoint entries from the payload into the metadata map. This is neaded because when the corfu server restart, it may not be able to deserialize the payload.

### DIFF
--- a/format/proto/types.proto
+++ b/format/proto/types.proto
@@ -32,7 +32,9 @@ message LogEntry {
     optional int64 checkpointId_most_significant = 11;
     optional int64 checkpointId_least_significant = 12;
     optional int64 checkpointedStreamId_most_significant = 13;
-    optional int64 checkpointedStreamdId_least_significant = 14;
+    optional int64 checkpointedStreamId_least_significant = 14;
+    //  Tail of the stream at the time of taking the checkpoint snapshot.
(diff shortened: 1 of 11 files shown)


zalokhan:
I have added the verification step.
I still need to add exception handling in the necromancer to avoid recovery in case of irrecoverable exceptions.

format/proto/types.proto
```diff
@@ -32,7 +32,9 @@ message LogEntry {
     optional int64 checkpointId_most_significant = 11;
     optional int64 checkpointId_least_significant = 12;
     optional int64 checkpointedStreamId_most_significant = 13;
-    optional int64 checkpointedStreamdId_least_significant = 14;
+    optional int64 checkpointedStreamId_least_significant = 14;
+    //  Tail of the stream at the time of taking the checkpoint snapshot.
```
dahliamalkhi: I don't understand this comment. Is it the global tail or the stream tail?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **c5a4f56**.

 [Pull Request #803 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/8ZpmqoE4Hy2M73hEqsSDn9wMbIJLAWfr)

PR #804, opened 2017-07-05
Return empty immutable collections on deserialization

no2chem:


no2chem:
This is not needed, because the Guava immutable collection builders correctly handle this case.

Issue #802, opened 2017-07-05
ckCommitAtomicity fails intermittently


slfritchie:
See https://travis-ci.org/CorfuDB/CorfuDB/builds/250437301#L3839 for context of this `org.corfudb.runtime.concurrent.StreamSeekAtomicityTest` unit test suite failure.

Particularly noteworthy:

    Caused by: java.lang.NullPointerException
    	at org.corfudb.runtime.object.transactions.OptimisticTransactionalContext.lambda$access$0(OptimisticTransactionalContext.java:101)
    	at org.corfudb.runtime.object.VersionLockedObject.access(VersionLockedObject.java:180)
    	at org.corfudb.runtime.object.transactions.OptimisticTransactionalContext.access(OptimisticTransactionalContext.java:95)
    	at org.corfudb.runtime.object.CorfuCompileProxy.accessInner(CorfuCompileProxy.java:162)
    	... 8 more

Maithem:
The problem seems to have been fixed.

PR #805, opened 2017-07-06
Fix getOptimisticStreamUnsafe() race between optimistic & locked executions

slfritchie:
Prior to this patch, a NullPointerException was possible with use of
getOptimisticStreamUnsafe()

1. checking getOptimisticStreamUnsafe() for null
2. using getOptimisticStreamUnsafe()'s result to call a chained method

This race condition can happen during optimistic txn execution.
Between 1 & 2, a second thread can try optimistic txn execution,
fail in a race, then *lock* the object and re-execute.  During
the re-execution, the optimistic stream pointer is set to null.

When the first thread resumes execution, step 2 will throw
NullPointerException.

I spent a while rummaging around to see if there's a deeper problem with this race, but I don't think so.  If the optimistic txn takes some action that races with the optimistic stream pointer being set to null by some other thread, that action will be undone when the optimistic txn discovers that its optimistic lock isn't valid.  Reviewers, please double-check.


slfritchie:
Fixes #802

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **39502ae**.

 [Pull Request #805 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/2SFRnpiZChwhRr2ewbuSLffbftEh71ky)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **b431a66**.

 [Pull Request #805 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/znTqBy1vqV2w8909iTAwa6iAmE0wOw7t)

slfritchie:
@dahliamalkhi I have a unit test case crafted by hand with manually-inserted scheduler yield statements.  I haven't yet crafted an automagic annotation scheme using AspectJ.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -92,14 +92,15 @@
...
-                                || o.getOptimisticStreamUnsafe()
-                                        .isStreamCurrentContextThreadCurrentContext())
-                ),
+                .access(o -> {
+                            WriteSetSMRStream stream = o.getOptimisticStreamUnsafe();
+                            return (
+                                getWriteSetEntrySize(proxy.getStreamID()) == 0 && // No updates
+                                        // And at the correct timestamp
+                                        o.getVersionUnsafe() == getSnapshotTimestamp()
+                                        && (stream == null
```
medhavidhawan: good catch.

PR #809, opened 2017-07-06
Recovery logging & retaining max value in fastLoader

zalokhan:
Added logging.
Retaining max value in stream tail map.

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
@@ -183,12 +183,14 @@ public class FastSmrMapsLoader {
         // On checkpoint, we also need to track the stream tail of the checkpoint
         if (isCheckPointEntry(logData)) {
             if (logData.getCheckpointType() == CheckpointEntry.CheckpointEntryType.END) {
-                streamTails.put(logData.getCheckpointedStreamId(),
-                        getStartAddressOfCheckPoint(logData));
+                streamTails.compute(logData.getCheckpointedStreamId(),
+                        (uuid, value) -> (value == null) ? getStartAddressOfCheckPoint(logData)
+                            : Math.max(value, getStartAddressOfCheckPoint(logData)));
             }
         }
         for (UUID streamId : logData.getStreams()) {
-            streamTails.put(streamId, address);
+            streamTails.compute(streamId,
+                    (uuid, value) -> (value == null) ? address : Math.max(value, address));
         }
     }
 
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
```diff
@@ -183,12 +183,14 @@ public void updateStreamTails(long address, ILogData logData) {
         // On checkpoint, we also need to track the stream tail of the checkpoint
         if (isCheckPointEntry(logData)) {
             if (logData.getCheckpointType() == CheckpointEntry.CheckpointEntryType.END) {
-                streamTails.put(logData.getCheckpointedStreamId(),
-                        getStartAddressOfCheckPoint(logData));
+                streamTails.compute(logData.getCheckpointedStreamId(),
+                        (uuid, value) -> (value == null) ? getStartAddressOfCheckPoint(logData)
```
Maithem: You should also check that addresses don't go backwards, get the global address from the entry itself?

zalokhan: The fastSMRLoader already enforces that in the order of the ranges it processes on bulk reads.

Maithem: That's not true because you also consider checkpoint entries that have previous addresses.

Maithem: Dahlia's example:

1 2 3 4 5 6 7
A A A A A A

Where 5 6 7 have CP-start, CP-Cont and CP-End  and then any non-cp entries after that will overwrite the stream tail with a lower address because the backpoints point to an earlier address.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **47e1d7f**.

 [Pull Request #809 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/GuKTyCYuCu14veHAwhE2e0qs5Ru1PEWd)

runtime/src/main/java/org/corfudb/recovery/FastSmrMapsLoader.java
dahliamalkhi: @Maithem I don't understand your concern. are you saying that the log-address of the checkpoint should be considered? I don't see why, only the logical checkpoint start-address should matter.

In the example above, say that the checkpoint start-address is 2; hence 3 and 4 have -higher- log addresses than the checkpoint, although CP records are in higher log addresses (5, 6, 7).

Maithem: @dahliamalkhi I don't have a concern. Initially, when I asked Zeeshan to max/ceil the stream addresses instead of the overwriting them, I also asked him to have an explicit check that will fail if addresses will go backwards, but after your example I realized that check is not necessary because addresses can go backwards, but we shouldn't throw an error instead we should just take the max.

PR #811, opened 2017-07-07
Increase logging for NoRollbackException and TransactionExceptions

no2chem:
This PR introduces additional logging, for TransactionAbortedException
the id of the stream is added whenever possible, for NoRollbackException
the cause of the failed undo is added to the exception message.
1

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/exceptions/TransactionAbortedException.java
+++ b/runtime/src/main/java/org/corfudb/runtime/exceptions/TransactionAbortedException.java
@@ -1,7 +1,10 @@
 package org.corfudb.runtime.exceptions;
 
+import java.util.UUID;
+
 import lombok.Getter;
 import org.corfudb.protocols.wireprotocol.TxResolutionInfo;
+import org.corfudb.runtime.object.transactions.AbstractTransactionalContext;
 
 /**
  * Created by mwei on 1/11/16.
@@ -18,6 +21,9 @@ public class TransactionAbortedException extends RuntimeException {
     Integer conflictKey;
 
     @Getter
+    UUID conflictStream;
+
+    @Getter
     Throwable cause;
 
     /**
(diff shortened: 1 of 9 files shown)


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -555,7 +556,13 @@ protected void rollbackStreamUnsafe(ISMRStream stream, long rollbackVersion) {
                     applyUndoRecordUnsafe(it.previous());
                 }
             } else {
-                throw new NoRollbackException();
+                Optional<SMREntry> entry = entries.stream().findFirst();
+                if (entry.isPresent()) {
```
medhavidhawan: why not contain this check inside the Exception

runtime/src/main/java/org/corfudb/runtime/exceptions/TransactionAbortedException.java
```diff
@@ -28,22 +34,28 @@
...
-        this(txResolutionInfo, conflictKey, abortCause, null);
+            Integer conflictKey, AbortCause abortCause, AbstractTransactionalContext context) {
+        this(txResolutionInfo, conflictKey, null, abortCause, null, context);
     }
 
     public TransactionAbortedException(
             TxResolutionInfo txResolutionInfo,
-            Integer conflictKey, AbortCause abortCause, Throwable cause) {
+            Integer conflictKey, UUID conflictStream,
+            AbortCause abortCause, Throwable cause, AbstractTransactionalContext context) {
```
medhavidhawan: retain the context as part of the exception and also change the toString of the AbstractTransactionalContext to print out more information.

no2chem: Fixed

PR #813, opened 2017-07-07
Fix Version Object Trim Race

Maithem:
This patch fixes a race when a transaction starts with a time
stamp that is greater than the trim mark, but then fails with
a TrimmedException.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
+++ b/runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
@@ -1,10 +1,22 @@
 package org.corfudb.runtime.exceptions;
 
+import lombok.Getter;
+import lombok.Setter;
+import org.corfudb.runtime.view.Address;
+
 /**
  * This exception is thrown when a client tries to read an address
  * that has been trimmed.
  */
 public class TrimmedException extends LogUnitException {
+    @Getter
+    @Setter
+    long address = Address.NON_ADDRESS;
+
+    @Getter
+    @Setter
+    long snapshot = Address.NON_ADDRESS;
+
     public TrimmedException() {
 
     }
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -26,6 +26,7 @@ import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.exceptions.TrimmedUpcallException;
 import org.corfudb.runtime.object.transactions.AbstractTransactionalContext;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
+import org.corfudb.runtime.view.Address;
 import org.corfudb.util.MetricsUtils;
 import org.corfudb.util.Utils;
 import org.corfudb.util.serializer.ISerializer;
@@ -161,6 +162,20 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
                 return TransactionalContext.getCurrentContext()
                         .access(this, accessMethod, conflictObject);
             } catch (Exception e) {
+                if (e instanceof TransactionAbortedException) {
+                    TransactionAbortedException tae = (TransactionAbortedException) e;
+                    if (tae.getCause() instanceof TrimmedException) {
+                        TrimmedException te = (TrimmedException) tae.getCause();
+                        AbstractTransactionalContext context = TransactionalContext.getCurrentContext();
+                        System.out.println(te.getSnapshot() + "   " + TransactionalContext
+                                .getCurrentContext().getSnapshotTimestamp());
+                        if (context.getRetry() < 1 || te.getSnapshot() != Address.NON_ADDRESS
+                                && te.getSnapshot() <= context.getSnapshotTimestamp()) {
+                            context.incrementRetry();
+                            return accessInner(accessMethod, conflictObject, isMetricsEnabled);
+                        }
+                    }
+                }
                 log.error("Access[{}] Exception: {}", this, e);
                 this.abortTransaction(e);
             }
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
@@ -98,6 +98,10 @@ public abstract class AbstractTransactionalContext implements
     private final long snapshotTimestamp = obtainSnapshotTimestamp();
 
 
+    @Getter
+    private volatile int retry = 0;
+
+
     /**
      * The address that the transaction was committed at.
      */
@@ -133,6 +137,10 @@ public abstract class AbstractTransactionalContext implements
         parentContext = TransactionalContext.getCurrentContext();
 
         AbstractTransactionalContext.log.debug("TXBegin[{}]", this);
+    }
+
+    public synchronized void incrementRetry() {
+        retry++;
     }
 
     /**
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -154,7 +154,9 @@ public class AddressSpaceView extends AbstractView {
                 throw new RuntimeException("Unexpected return of empty data at address "
                         + address + " on read");
             } else if (data.isTrimmed()) {
-                throw new TrimmedException();
+                TrimmedException te = new TrimmedException();
+                te.setAddress(address);
+                throw te;
             }
             return data;
         }
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -202,7 +202,16 @@ public abstract class AbstractQueuedStreamView extends
      *
      * @param address       The address to read.
      */
-    protected abstract @Nonnull ILogData read(final long address);
+    protected @Nonnull ILogData read(final long address) {
+        try {
+            return readImpl(address);
+        } catch (TrimmedException e) {
+            e.setSnapshot(getCurrentContext().checkpointSnapshotAddress);
+            throw e;
+        }
+    }
+
+    protected abstract @Nonnull ILogData readImpl(final long address);
 
     /**
      * Given a list of addresses, retrieve the data as a list in the same
@@ -212,9 +221,18 @@ public abstract class AbstractQueuedStreamView extends
      *                      addresses given.
      */
     protected @Nonnull List<ILogData> readAll(@Nonnull final List<Long> addresses) {
+        try {
+            return readAllImpl(addresses);
+        } catch (TrimmedException e) {
+            e.setSnapshot(getCurrentContext().checkpointSnapshotAddress);
+            throw e;
+        }
+    }
+
+    protected  @Nonnull List<ILogData> readAllImpl(@Nonnull final List<Long> addresses) {
         return addresses.parallelStream()
-                        .map(this::read)
-                        .collect(Collectors.toList());
+                .map(this::read)
+                .collect(Collectors.toList());
     }
 
     /**
@@ -231,8 +249,18 @@ public abstract class AbstractQueuedStreamView extends
      * @return              True, if entries were added to the read queue,
      *                      False, otherwise.
      */
-    protected abstract boolean fillReadQueue(final long maxGlobal,
-                                          final QueuedStreamContext context);
+    protected boolean fillReadQueue(final long maxGlobal,
+                                          final QueuedStreamContext context) {
+        try {
+            return fillReadQueueImpl(maxGlobal, context);
+        } catch (TrimmedException e) {
+            e.setSnapshot(getCurrentContext().checkpointSnapshotAddress);
+            throw e;
+        }
+    }
+
+    protected abstract boolean fillReadQueueImpl(final long maxGlobal,
+                                             final QueuedStreamContext context);
 
     /**
      * {@inheritDoc}
--- a/test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
+++ b/test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
@@ -100,4 +100,54 @@ public class CheckpointTrimTest extends AbstractViewTest {
         assertThat(newTestMap)
                  .containsKeys("a", "b", "c");
     }
+
+    @Test
+    public void testCheckpointTrimDuringPlayback2() throws Exception {
+        Map<String, String> testMap = getDefaultRuntime().getObjectsView().build()
+                .setTypeToken(new TypeToken<SMRMap<String, String>>() {})
+                .setStreamName("test")
+                .open();
+
+        Map<String, String> testMap2 = getDefaultRuntime().getObjectsView().build()
+                .setTypeToken(new TypeToken<SMRMap<String, String>>() {})
+                .setStreamName("test2")
+                .open();
+
+        // Place entries into the map
+        testMap.put("a", "a");
+        testMap.put("b", "a");
+        testMap.put("c", "a");
+        testMap2.put("a", "c");
+        testMap2.put("b", "c");
+        testMap2.put("c", "c");
+
+        // Insert a checkpoint
+        MultiCheckpointWriter mcw = new MultiCheckpointWriter();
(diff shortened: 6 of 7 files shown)


runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
```diff
@@ -5,11 +5,25 @@
  * that has been trimmed.
  */
 public class TrimmedException extends LogUnitException {
+    private final long address;
     public TrimmedException() {
-
+        address = -1;
```
no2chem: Can you use the Address.NON_EXIST constant?

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -161,6 +161,22 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
                 return TransactionalContext.getCurrentContext()
                         .access(this, accessMethod, conflictObject);
             } catch (Exception e) {
+                if (e instanceof TrimmedException) {
```
no2chem: This doesn't seem needed - the TXcontext will reset the object, and under a lock since it is done inside VersionLockedObject::access’s call to updateFunction. 

Perhaps though, rather than aborting the TXContext itself needs to retry at least one to pick up the checkpoint.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -161,6 +161,22 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
...
+                    TrimmedException te = (TrimmedException) e;
+                    long snapShotTime = TransactionalContext.getCurrentContext()
```
no2chem: snapShotTime -> snapshotTime

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **2e9a93b**.

 [Pull Request #813 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/1zVt6No91cB3K1qHJvyeEAHKKlR6OvZz)

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
slfritchie: "retry at least one" meaning the recursive call to `accessInner` that Maithem has added here?

runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
```diff
@@ -5,11 +5,25 @@
...
+    public TrimmedException(long address) {
+        this.address = address;
+    }
+
+    public long getAddress() {
+        return address;
+    }
+
+    public boolean hasAddress() {
+        return this.address > -1;
```
dahliamalkhi: there is a method for this, Address.isAddress()

Maithem:
Hopefully I have addressed all comments.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -161,6 +162,20 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
                 return TransactionalContext.getCurrentContext()
                         .access(this, accessMethod, conflictObject);
             } catch (Exception e) {
+                if (e instanceof TransactionAbortedException) {
+                    TransactionAbortedException tae = (TransactionAbortedException) e;
+                    if (tae.getCause() instanceof TrimmedException) {
+                        TrimmedException te = (TrimmedException) tae.getCause();
+                        AbstractTransactionalContext context = TransactionalContext.getCurrentContext();
+                        System.out.println(te.getSnapshot() + "   " + TransactionalContext
```
Maithem: Remove system.out

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -161,6 +162,20 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
...
+                                .getCurrentContext().getSnapshotTimestamp());
+                        if (context.getRetry() < 1 || te.getSnapshot() != Address.NON_ADDRESS
```
medhavidhawan: add a comment as this is the crux of the whole thing.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -154,7 +154,9 @@ public void write(IToken token, Object data) throws OverwriteException {
                 throw new RuntimeException("Unexpected return of empty data at address "
                         + address + " on read");
             } else if (data.isTrimmed()) {
-                throw new TrimmedException();
+                TrimmedException te = new TrimmedException();
```
medhavidhawan: why not have another constructor with address as a parameter.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
```diff
@@ -100,4 +100,54 @@ public void testCheckpointTrimDuringPlayback() throws Exception {
         assertThat(newTestMap)
                  .containsKeys("a", "b", "c");
     }
+
+    @Test
+    public void testCheckpointTrimDuringPlayback2() throws Exception {
```
medhavidhawan: High level comment is useful as it eases people into reading the test in future.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
dahliamalkhi: this counter is wrong. it counts retries per transaction-context, but a TX might access more than one object, and each might encounter a trimException.

The retry counter should be for the loop here only. 

Also, you don't want to retry even once if te.getSnapshot() > context.getSnapshotTS() . The retry counter should be used for two purposes:

1. bound the recursion depth
2. print a warning message if we retry twice in the loop here.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -98,6 +98,10 @@
     private final long snapshotTimestamp = obtainSnapshotTimestamp();
 
 
+    @Getter
+    private volatile int retry = 0;
```
dahliamalkhi: see above; I think this counter is wrong.

Maithem: Do you suggest that I maintain a counter per-stream?

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
Maithem: yes, I had that in my initial patch, but missed it here. Will add it.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **186f670**.

 [Pull Request #813 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/wsMRUPcR9xwHkHAVCkCcVWzBUSYaw8gY)

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
no2chem: Also, rename this test to what it actually tests now (transactionalReadAfterCheckpoint would work).

runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
```diff
@@ -1,10 +1,22 @@
...
  * that has been trimmed.
  */
 public class TrimmedException extends LogUnitException {
+    @Getter
+    @Setter
+    long address = Address.NON_ADDRESS;
+
+    @Getter
+    @Setter
+    long snapshot = Address.NON_ADDRESS;
```
no2chem: This isn't quite what we discussed... I guess it would be okay to store this address, which BTW, is not a snapshot, but the minimum object "version"), but it should provide a boolean (in this case, perhaps just via a method, called isNonrecoverable(), for example), so the client doesn't haven't do compute whether or not it is recoverable.

Frther

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -161,6 +162,20 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
                 return TransactionalContext.getCurrentContext()
                         .access(this, accessMethod, conflictObject);
             } catch (Exception e) {
+                if (e instanceof TransactionAbortedException) {
```
no2chem: This continues to be the wrong place to do this. Do NOT catch the TransactionAbortedException, handle the TrimException in the transactional context.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
no2chem: And yeah, remove System.out.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
no2chem: Why does the context have retry state?

no2chem: Agree, counter should be for loop only. Also, handling this error is still in the wrong place.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -135,6 +139,10 @@
         AbstractTransactionalContext.log.debug("TXBegin[{}]", this);
     }
 
+    public synchronized void incrementRetry() {
```
no2chem: What is the point of this method?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -202,7 +202,16 @@ protected ILogData getNextEntry(QueuedStreamContext context,
      *
      * @param address       The address to read.
      */
-    protected abstract @Nonnull ILogData read(final long address);
+    protected @Nonnull ILogData read(final long address) {
+        try {
+            return readImpl(address);
```
no2chem: Let's not create xxImpl methods, please just catch the exception in the BackpointerStreamView. Other implementations (if we ever have any) may have very different ways of recovering.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
medhavidhawan: I see good catch! I had missed that in the review.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
medhavidhawan: Ok I have a question/comment around this.
1. Can we not unify how we handle trimmed exception ? Or is the handling for nontransactional case very different? (Hopefully we are not missing anything in that case) It is ok to use the same implementation at different places.

dahliamalkhi: It -is- different. In a regular access, you may want to keep retrying, because you get a new stream-tail from the sequencer each time. 

in a TX context, on the other hand, we only want to try once against each trim-offset. If that fails, it means that the TX is trying to get to a snapshot which has been trimmed, and will just keep failing.

medhavidhawan: I see, but that also has a potential for infinite recursion in case of any bugs. Also if there are no checkpoints between the start version and sync to version or if we already have tried syncing with the existing checkpoint we will just stack overflow without any meaningful logging or information ? Is that true or am I missing something.

dahliamalkhi: agreed. We can add an assertion that trimExceptions in normal "access()" return a different trim each time.

medhavidhawan:
Also just want to understand why we need to fetch a new tail from sequencer
on every recursion for this particular access instance. This access
happened at a particular point in time, why not just sync object up to this
time point for this request? I am just asking questions for more clarity
around this

runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
Maithem: This snapshot value is populated from checkpointSnapshotAddress as the exception is propagated upwards.

Just to recapitulate, a non-recoverable exception happens when the TransactionSnapshotTime is less than the checkpointSnapshotAddress after the first retry (because the checkpointSnapshotAddress is not guaranteed to point to the latest checkpoint on the first access). Do you agree?

Those two values come from two places, one, AbstractQueuedStreamView and AbstractTransactionalContext, how would you determine whether a Trimmed Exception is non-recoverable from BackpointerStreamView if you don't have access to the retry counter?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
Maithem: Initially I intended to catch and populate the trimmed exception in BackpointerStreamView, but then it didnt seem to make sense because BackpointerStreamView is the "concrete" implementation and seemed like it need to be in a higher layer. I don't like the xxImpl methods too, but that's just how the code is structured.

I can move it to BackpointerStreamView, I don't have a strong preference on this issue @medhavidhawan @dahliamalkhi what do you guys think?

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
Maithem: @no2chem I put it in the CorfuCompileProxy because I didn't want to duplicate code in each TransactionalContext implementation (i.e. both snapshot and optimistic). Are you suggesting code duplication code here? or are you thinking of putting this in AbstractTransactionalContext?
If it's the latter, it will look like the xxImpl methods in AbstractQueuedStreamView. What do you think?

medhavidhawan:
I am not at my computer to visualize the different layers. High level you
need to use the principle to populate data from the source. Also logic
should lie in the most relevant layer. On my phone I cannot see the code
properly hence the obtuse answer

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
medhavidhawan: Ok I looked at the code now. The choices we have
1. create xxImpl style methods and wrap them with try catch .  
2. try catch down into backpointer stream view because no other implementation will use these abstractions.
3. move the read and readAll method implementations into the abstract class  and wrap them there.

At the moment the methods are so small, I do not have a big preferrence. If 3 is correct and makes sense to everybody we can do 3. else we can do 2.

dahliamalkhi: Agree with Medhavi, both 2 and 3 are viable, since we really only have one implementation for now; slight preference to 3 as well.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
no2chem: No, this continues to be the wrong place to put this. We can talk about combining the code that snapshotTransactionalContext and optimisticTransactionalContext uses, but an abort exception here is not the right place.

runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
no2chem: No, I don't agree. Whether a trimmed exception is recoverable or not has nothing to do with the number of retries.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
no2chem: You need to eliminate this counter variable.

Maithem:
Addressed by #815

PR #815, opened 2017-07-10
Fix Version Object Trim Race

Maithem:
This patch fixes a race when a transaction starts with a time
stamp that is greater than the checkpoint snapshot time, but
then fails with a TrimmedException.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
+++ b/runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
@@ -1,15 +1,27 @@
 package org.corfudb.runtime.exceptions;
 
+import lombok.Getter;
+import lombok.Setter;
+
 /**
  * This exception is thrown when a client tries to read an address
  * that has been trimmed.
  */
 public class TrimmedException extends LogUnitException {
-    public TrimmedException() {
 
+    /*
+     * This flag determins whether the operation that caused this exception
+     * can retry or not.
+     */
+    @Getter
+    @Setter
+    private boolean retriable = true;
+
+    public TrimmedException() {
     }
 
     public TrimmedException(String message) {
         super(message);
+        retriable = true;
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
@@ -111,20 +111,27 @@ public class OptimisticTransactionalContext extends AbstractTransactionalContext
                             // inside syncObjectUnsafe, depending on the object
                             // version, we may need to undo or redo
                             // committed changes, or apply forward committed changes.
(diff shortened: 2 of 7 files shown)


Maithem:
@no2chem  @medhavidhawan  @dahliamalkhi  ok hopefully this is a much cleaner patch. I created a different PR so I dont truncate the code with push force. I've addressed most comments (still didnt write comments on why we need retry logic).

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **f5a0b34**.

 [Pull Request #815 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/XEsodbJcfUhOlMHS0xUFkhGHogH05zFH)

runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
```diff
@@ -1,15 +1,27 @@
...
 public class TrimmedException extends LogUnitException {
-    public TrimmedException() {
 
+    /*
+     * This flag determins whether the operation that caused this exception
+     * can retry or not.
+     */
+    @Getter
+    @Setter
+    private boolean retriable = true;
```
no2chem: I would consider renaming this "resolved" or something similar, a "resolved" trimmed exception means that we're certain this address is permanently trimmed and no longer accessible, a "unresolved" trimmed exception does not.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -111,20 +111,27 @@
...
-                                o.syncObjectUnsafe(getSnapshotTimestamp());
-                            } catch (TrimmedException te) {
-                                // If a trim is encountered, we must reset the object
-                                o.resetUnsafe();
-                                // and abort the transaction
-                                TransactionAbortedException tae =
-                                        new TransactionAbortedException(
-                                                new TxResolutionInfo(getTransactionID(),
-                                                        getSnapshotTimestamp()), null,
+                            for (int x = 0; x < this.builder.getRuntime().getTrimRetry(); x++) {
```
no2chem: Okay, this is the right place now. If you wanted code reuse, you could stuff this into a function like AbstractTransactionalContext::doTransactionalSync(VersionLockedObject object)

runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
dahliamalkhi: I disgree. retriable conveys exactly what this is. resolved is ambiguous.

no2chem: No, I still disagree. What is a retriable exception? The exception is not retriable, the read of the object at the version is, and that is precisely because whether or not the address is trimmed is unresolved.

PR #826, opened 2017-07-13
Release referenced message object in NettyServerRouter

zalokhan:
Releasing referenced message object held by NettyServerRouter.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
@@ -3,6 +3,7 @@ package org.corfudb.infrastructure;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.util.ReferenceCountUtil;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -148,6 +149,9 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter
             }
         } catch (Exception e) {
             log.error("Exception during read!", e);
+        } finally {
+            // Releasing referenced counted objects passed to the handler.
+            ReferenceCountUtil.release(msg);
         }
     }
 


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **03df6e1**.

 [Pull Request #826 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/9knUllZTNAAuz203W53XrLfek8KZchOt)

zalokhan:
No, I don't have tests. These messages are being released in the ClientRouter explicitly in the parent class.

slfritchie:
Before running a simple in-memory single node server plus Maithem's load generator (in #838), I added a `ReferenceCountUtil.refCnt(msg)` call to check the reference count.  According to https://netty.io/4.1/api/index.html?io/netty/util/ReferenceCountUtil.html, the count is already -1, namely, not reference counted.

Maithem:
@slfritchie in other words, this is not needed?

slfritchie:
@Maithem Hm, well, I didn't look to see if there are other paths or other allocators for the `msg` object.  Applying this patch doesn't crash anything -- the call appears to be a noop.

Maithem:
@slfritchie if its a no-op, then why is it needed?

slfritchie:
Hi, Maithem, sorry, It isn't needed.  At best, the noop would be a reminder or already-taken-care-of-in-this-one-spot if we switch back to refcounted allocators, right?

zalokhan:
I ran the server with a modification in the NettyServerRouter to cause a Runtime Exception on 80% of the requests:
```
    public void channelRead(ChannelHandlerContext ctx, Object msg) {
        try {
            // The incoming message should have been transformed to a CorfuMsg earlier in the
            // pipeline.
            CorfuMsg m = ((CorfuMsg) msg);
            // We get the handler for this message from the map
            AbstractServer handler = handlerMap.get(m.getMsgType());
            if (handler == null) {
                // The message was unregistered, we are dropping it.
                log.warn("Received unregistered message {}, dropping", m);
            } else {
                if (validateEpoch(m, ctx)) {
                    // Route the message to the handler.
                    if (log.isTraceEnabled()) {
                        log.trace("Message routed to {}: {}", handler.getClass().getSimpleName(),
                                msg);
                    }
                    handlerWorkers.submit(() -> handler.handleMessage(m, ctx, this));
                }
            }
            if (random.nextInt(100) < 80) {
                throw new RuntimeException("drop");
            }
        } catch (Exception e) {
            log.error("Exception during read!", e);
        }
    }
```
After hitting the server with several thousand requests over a period of 30 minutes I do not see any significant increase in non-heap memory.

PR #829, opened 2017-07-14
Two level conflict hash, set hash to 32 bits

no2chem:
This PR creates a two-level conflict hash cache, the first level
is organized by UUID, and also uses the full 32 bits provided
by Java's .hashCode() for the hash.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -14,6 +14,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
@@ -128,7 +129,9 @@ public class SequencerServer extends AbstractServer {
     private long maxConflictWildcard = Address.NOT_FOUND;
 
     private final long defaultCacheSize = Long.MAX_VALUE;
-    private final Cache<Integer, Long> conflictToGlobalTailCache;
+    private final long defaultPerStreamCacheSize = Long.MAX_VALUE;
+    private long perStreamCacheSize = defaultPerStreamCacheSize;
+    private final Cache<UUID, Cache<Integer, Long>> conflictToGlobalTailCache;
 
     /**
      * Handler for this server.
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -128,7 +129,9 @@
     private long maxConflictWildcard = Address.NOT_FOUND;
 
     private final long defaultCacheSize = Long.MAX_VALUE;
-    private final Cache<Integer, Long> conflictToGlobalTailCache;
+    private final long defaultPerStreamCacheSize = Long.MAX_VALUE;
+    private long perStreamCacheSize = defaultPerStreamCacheSize;
+    private final Cache<UUID, Cache<Integer, Long>> conflictToGlobalTailCache;
```
medhavidhawan: I worry about these cache in cache kind of things. The cache is not passive entiry it does it's own evictions, metrics etc. Is it possible to keep a simpler passive data structure like a HashMap of Caches or just a cache<Stream, Pair<Stream, conflictHash>> or something to that effect. This 2 level nesting of the cache data structure  makes me uncomfortable.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -171,33 +174,40 @@ public SequencerServer(ServerContext serverContext) {
...
             cacheSize = Long.parseLong((String) opts.get("--sequencer-cache-size"));
         }
+        if (opts.get("--sequencer-per-stream-cache-size") != null
+                && Long.parseLong((String) opts.get("--sequencer-per-stream-cache-size")) != 0) {
+            perStreamCacheSize =
+                    Long.parseLong((String) opts.get("--sequencer-per-stream-cache-size"));
+        }
+
 
         conflictToGlobalTailCache = Caffeine.newBuilder()
```
medhavidhawan: Do you anticipate the streams to be in such large number that you need a loading cache for them ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -171,33 +174,40 @@ public SequencerServer(ServerContext serverContext) {
...
                 .maximumSize(cacheSize)
-                .removalListener((Integer k, Long v, RemovalCause cause) -> {
+                .removalListener((UUID k, Cache<Integer, Long> v, RemovalCause cause) -> {
                     if (!RemovalCause.REPLACED.equals(cause)) {
-                        log.trace("Updating maxConflictWildcard. Old value = '{}', new value='{}'"
-                                        + " conflictParam = '{}'. Removal cause = '{}'",
-                                maxConflictWildcard, v, k, cause);
-                        maxConflictWildcard = Math.max(v, maxConflictWildcard);
+                        if (v != null) {
```
medhavidhawan: Do we need more new tests as this functionality has changes in the data structures used ? Where is that covered ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
medhavidhawan: I also think there is a general problem in this scheme (Not because of this code change). The object.hashCode() generated at the client side will not match for the same object from different clients unless they are forced to generate a consistent hashcode across all client jvm instances. This will lead to a lot of problems in a cluster scenario.

no2chem: Sure, we could make the first level a cache and the second level a hashtable.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
no2chem: I don't know. It's really workload dependent.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
no2chem: Well, the previous tests cover this code, but we could add some additional tests to make sure the wildcard is properly loaded on eviction.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
no2chem: Yes, that _could_ be a problem. Fortunately for all Java library objects, .hashCode() is stable across all JVMs...
I think if you wanted to solve this the right way you would add some hook where we could generate the hash code post serialization. That's probably going to require some invasive updates, but I can give it a stab. BTW, Guava's Hasher can only deal with primitive types, as I suspected.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a8a8f1a**.

 [Pull Request #829 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/BWrjGSQqxCcyPDJX9Ts5T76nfCusY0ks)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
medhavidhawan: Is there some spec that guarantees that hashCode() is stable across JVMs. please forward the link if you have.  
Yes I just looked the Guava Hasher does not deal with objects that easily. However it does have
the ObjectFunnel thing to pipe in all the relevant members to hash. 

class TestObject {
        String str = "new";
        final int i = 24;
        @Override
        public int hashCode() {
            return Hashing.md5().hashObject(this, new Funnel<TestObject>() {
                @Override
                public void funnel(TestObject testObject, PrimitiveSink primitiveSink) {
                    primitiveSink.putString(str, Charset.defaultCharset());
                    primitiveSink.putInt(i);
                }
            }).hashCode();
        }
    }
    @Test
    public void testHash() {
        System.out.println("THE HASH CODE IS " + new TestObject().hashCode());
    }


You are right, hashCode can be generated post serialization as well and that could be an approach.
Or hashCode can be generated using introspection to create the set of relevant values in the object and then hashing them.

If we had the ConflictParam/Key as a first class interface, we could have our own hash Code which is larger than an integer and also have an equals method.

dahliamalkhi: I am also somewhat concerned but for a different, concrete reason. 

Evicting conflict information on entire-stream basis seems somewhat crude. Whenever the cache limit is hit, I would prefer to evict many ancient entries from multiple streams, rather than all entries from one stream.

This is the case where I prefer to keep the conflict map "flat", and not sorted per stream. Yes, it is twice the space, but the cache eviction policy is more sane.

no2chem: It's true, but the problem is that this would involve an object creation per lookup, which I'm trying to avoid. I guess there's not really a great way around it.

no2chem: There's no spec at all, just that the core Java library happens to specify how hashCodes are implemented, and by luck that hasn't changed... Yes, it probably is perilous to continue using hashCodes. I'm converging more and more that we should just dump this approach and figure out how to tell the serializer to generate our hash code for us.

no2chem:
Okay, I think I have a better design for a two level table, closing this PR.

PR #820, opened 2017-07-12
Ignore Trimming Trimmed Addresses

Maithem:
This patch changes the behavior of prefixTrim, instead of throwing
an exception on prefix trimming a trimmed address, it just ignores
the request.

PR #830, opened 2017-07-14
CP Empty Maps

Maithem:
This patch logically revers PR#808 and adds a new test case
which encodes an edge case where accessing an empty map
doesn't cause a StackOverflowException.

### DIFF
--- a/test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
+++ b/test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
@@ -61,8 +61,9 @@ public class CheckpointSmokeTest extends AbstractViewTest {
         IStreamView sv = r.getStreamsView().get(CorfuRuntime.getStreamID("S1"));
         final int objSize = 100;
         long a1 = sv.append(new byte[objSize]);
-        final long firstAddress = 0L;
-        assertThat(a1).isEqualTo(firstAddress);
+        final long cpEndAddress = 2L;
+        // Verify that the start/end records have been written for empty maps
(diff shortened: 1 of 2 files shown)


slfritchie:
I haven't fully caught up yet on mid-week events, sorry.  #820 also looks good to me and should be merged as long as Robert doesn't mind.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -525,5 +526,48 @@ public void MultiCheckpointWriterTestInner(boolean consecutiveCkpoints) throws E
             assertThat(m2B.get(keyB)).isEqualTo(i);
         }
     }
+
+    @Test
+    public void emptyCheckPoint() throws Exception {
```
medhavidhawan: It would be good to write comments.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -525,5 +526,48 @@ public void MultiCheckpointWriterTestInner(boolean consecutiveCkpoints) throws E
...
+        final String author = "CPWriter";
+        final int iter = 1000;
+
+        for (int x = 0; x < iter; x++) {
+            mA.put(Integer.toString(x), (long) x);
+            mB.put(Integer.toString(x), (long) x);
+        }
+
+        for (String key : mA.keySet()) {
+            mA.remove(key);
```
medhavidhawan: Do we need to test the case where put and remove are done in transactions.

PR #833, opened 2017-07-14
32 bit sequencer hash

no2chem:
This PR is a slight improvement over #829. It uses a string consisting
of the hash code plus the stream ID as a key, we will have to monitor
it carefully to see if this causes additional GC pressure or performance
regression.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -128,7 +128,7 @@ public class SequencerServer extends AbstractServer {
     private long maxConflictWildcard = Address.NOT_FOUND;
 
     private final long defaultCacheSize = Long.MAX_VALUE;
-    private final Cache<Integer, Long> conflictToGlobalTailCache;
+    private final Cache<String, Long> conflictToGlobalTailCache;
 
     /**
      * Handler for this server.
@@ -174,7 +174,7 @@ public class SequencerServer extends AbstractServer {
 
         conflictToGlobalTailCache = Caffeine.newBuilder()
                 .maximumSize(cacheSize)
-                .removalListener((Integer k, Long v, RemovalCause cause) -> {
+                .removalListener((String k, Long v, RemovalCause cause) -> {
                     if (!RemovalCause.REPLACED.equals(cause)) {
                         log.trace("Updating maxConflictWildcard. Old value = '{}', new value='{}'"
                                         + " conflictParam = '{}'. Removal cause = '{}'",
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -236,7 +236,7 @@ public TokenType txnCanCommit(TxResolutionInfo txInfo, /** Input. */
                 // for each key pair, check for conflict;
                 // if not present, check against the wildcard
                 conflictParamSet.forEach(conflictParam -> {
-                    int conflictKeyHash = getConflictHashCode(entry.getKey(),
+                    String conflictKeyHash = getConflictHashCode(entry.getKey(),
```
medhavidhawan: Looks fine but it is not solving Anny's case. Do you plan to introduce the bigger hash in the next PR.
Also do you need more tests as the hashcode is now a string ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -193,8 +193,8 @@ public SequencerServer(ServerContext serverContext) {
     * @param conflictParam The conflict parameter.
     * @return A conflict hash code.
     */
-    public int getConflictHashCode(UUID streamId, int conflictParam) {
-        return Objects.hash(streamId, conflictParam);
+    public String getConflictHashCode(UUID streamId, int conflictParam) {
+        return streamId.toString() + Integer.toString(conflictParam);
```
dahliamalkhi: I think that by default, toStrong() prints the Java hashcode of the object, so this brings us back to where we started, no? 

if the intention is that this is just one step towards the full solution, I buy it.

no2chem: The hashcode Java uses here doesn't matter what matters is during a hashcode collision equals will be called and the correct conflict param will be used, unlike before. I'll write a test which shows this.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **710d657**.

 [Pull Request #833 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/oVDfTzX0htR1sLX2uJaP3Iz4cf5YVbjC)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
no2chem: It does solve the situation described in the unit test added.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **373a742**.

 [Pull Request #833 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/A62xRRd5bABj8AzanX8Ffvqx77gtVTQa)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
medhavidhawan: ok maybe I did not understand the test case correctly. In my conversation with Anny,
she mentioned that the hash for 2 UUIDs (keys) collided and so the streamId did not matter in the scheme. We should check with her once.
@annym could you please provide input.

PR #840, opened 2017-07-16
Change conflict parameters to pass byte arrays

no2chem:
This patch changes the underlying type sent to the
sequencer for conflict resolution from integers
to primitive byte arrays.

It still uses .hashCode to generate the byte
array, changing the generation mechanism will
be addressed in the next PR on this topic.


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -193,8 +193,8 @@ public SequencerServer(ServerContext serverContext) {
     * @param conflictParam The conflict parameter.
     * @return A conflict hash code.
     */
-    public String getConflictHashCode(UUID streamId, int conflictParam) {
-        return streamId.toString() + Integer.toString(conflictParam);
+    public String getConflictHashCode(UUID streamId, byte[] conflictParam) {
+        return streamId.toString() + Utils.bytesToHex(conflictParam);
```
dahliamalkhi: eventually, why not make the entire result a byte array?

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -122,7 +121,7 @@
     private final WriteSetInfo writeSetInfo = new WriteSetInfo();
 
     @Getter
-    private final ReadSetInfo readSetInfo = new ReadSetInfo();
+    private final ConflictSetInfo readSetInfo = new ConflictSetInfo();
```
dahliamalkhi: this naming is going to be confusing. I am not saying the previous situation was better.. but still: ConflictSet is for conflicts, but readSet may not actually contain conflicts in the case of write-write transactions.

AccessSetInfo?

no2chem: In the case of write-write transactions, I don't think "ReadSetInfo" is even necessary. It happens that AbstractTransactionContext keeps track of them. AccessSetInfo isn't very descriptive.

I still believe ConflictSet best describes what it is.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
no2chem: Eventually, I think this needs to be removed in general so I didn't really touch it much. Allocating a whole byte array just to do a lookup every time seems pricey.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **307d328**.

 [Pull Request #840 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/0yVrGhCIY4al99WC00PU6I3N3aty7SOY)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
dahliamalkhi: I meant: why not return byte[] instead of a String, and work with byte[] types for conflict throughput?

no2chem: Because that involves generating a byte buffer and a byte array, but ok, if you insist...

no2chem: Also I'll add that byte[] uses reference equality, making this change super messy...

no2chem: Yeah no. Using a byte[] as a key would not by trivial. You're free to attempt to implement it, but I warn you, it'll involve generating a bytebuffer to create the key, then putting the byte[] into a custom wrapper class because caffeine doesn't support custom comparators.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -45,6 +45,36 @@
...
+    private final static char[] hexArray = "0123456789ABCDEF".toCharArray();
+
+    /** Convert a byte array to a hex string.
+     * Source:
+     * https://stackoverflow.com/questions/9655181/
+     * how-to-convert-a-byte-array-to-a-hex-string-in-java
+     * @param bytes Byte array to convert
+     * @return      Hex string representation.
+     */
+    public static String bytesToHex(byte[] bytes) {
```
medhavidhawan: You could use a standard library for this if you want 
Google:BaseEncoding.base16()
Apache:org.apache.commons.codec.binary.Hex.encodeHexString(byteArray)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **e243147**.

 [Pull Request #840 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/LmlZYfXat3DVRuelSV4zmiwKEp4dTejz)

PR #848, opened 2017-07-16
Support custom hash code scheme registration plus interface

no2chem:
This PR introduces support for registering custom hash functions on a
per type basis, as well as a custom interface.
 When the serializer is asked for a hash byte array,
the serializer consults with a map of custom hashers, if present
it executes the custom hasher. If threre is no custom hasher, the serializer
checks if the type implements type ICorfuHashable, if it does, it uses
that method provided on the type.

The custom hasher can choose to do anything, from passing an empty
byte array to using all data - for performance reason the hash size
should be as small as possible.


test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
```diff
@@ -131,6 +135,194 @@ public void customConflictObjectsNoConflictNoAbort()
...
+     * with the serializer to always conflict, as it always hashes to
+     * and empty byte array.
+     */
+    @Test
+    public void customSameHashAlwaysConflicts() {
+        // Register a custom hasher which always hashes to an empty byte array
+        Serializers.JSON.registerCustomHasher(CustomSameHashConflictObject.class,
+                o -> new byte[0]);
+
+        CustomSameHashConflictObject c1 = new CustomSameHashConflictObject("a", "a");
```
slfritchie: The two keys here placeholders for the stream ID and map key, right?  if so, then please change their names and/or constants to be more descriptive than `k1` or `a`.

no2chem: No, they are just internal keys that have no meaning, other than being equal. They are not related to the stream or map at all.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **6a75a71**.

 [Pull Request #848 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/VYkdY2kvUBFGQMtaQZyNtm6qc7VA33Jc)

test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
```diff
@@ -131,6 +135,194 @@ public void customConflictObjectsNoConflictNoAbort()
...
+                .open();
+
+        t(1, this::OptimisticTXBegin);
+        t(2, this::OptimisticTXBegin);
+        t(1, () -> map.put(c1 , "v1"));
+        t(2, () -> map.put(c2 , "v2"));
+        t(1, this::TXEnd);
+        t(2, this::TXEnd)
+                .assertThrows()
+                .isInstanceOf(TransactionAbortedException.class);
```
medhavidhawan: Good tests also verify the abort cause to be hash conflict.

runtime/src/main/java/org/corfudb/util/serializer/ICorfuHashable.java
```diff
@@ -0,0 +1,10 @@
+package org.corfudb.util.serializer;
+
+/** Represents an object which the serializer may hash.
+ * Created by mwei on 7/16/17.
+ */
+public interface ICorfuHashable {
+
+    /** Generate a hash for this object to be used by Corfu. */
+    byte[] generateCorfuHash();
```
medhavidhawan: Does the method name need to have Corfu in it. Is there a significance to that ?

no2chem: I guess I didn't want the name to clash with any other existing implementation. We can take it out if it's preferable.

medhavidhawan: It was a nit, it is your choice. Since the hash is not specific to Corfu so i asked.

Issue #853, opened 2017-07-18
Stream Sync IndexOutOfBoundsException


Maithem:
There seems to be a concurrency issue when running operations concurrently. An IndexOutOfBoundsException is thrown. The problem can be reproduced using the load generator. 

Stacktrace
java.lang.IndexOutOfBoundsException: Index: 0
at java.util.Collections$EmptyList.get(Collections.java:4454)
at org.corfudb.runtime.object.transactions.WriteSetSMRStream.current(WriteSetSMRStream.java:177)
at org.corfudb.runtime.object.transactions.WriteSetSMRStream.previous(WriteSetSMRStream.java:215)
at org.corfudb.runtime.object.VersionLockedObject.rollbackStreamUnsafe(VersionLockedObject.java:563)
at org.corfudb.runtime.object.VersionLockedObject.optimisticRollbackUnsafe(VersionLockedObject.java:620)
at org.corfudb.runtime.object.VersionLockedObject.syncObjectUnsafe(VersionLockedObject.java:312)
at org.corfudb.runtime.object.CorfuCompileProxy.lambda$getUpcallResultInner$5(CorfuCompileProxy.java:284)
at org.corfudb.runtime.object.VersionLockedObject.update(VersionLockedObject.java:243)
at org.corfudb.runtime.object.CorfuCompileProxy.getUpcallResultInner(CorfuCompileProxy.java:283)
at org.corfudb.runtime.object.CorfuCompileProxy.getUpcallResult(CorfuCompileProxy.java:236)
at org.corfudb.runtime.collections.SMRMap$CORFUSMR.put(SMRMap$CORFUSMR.java:173)
at org.corfudb.generator.operations.WriteOperation.execute(WriteOperation.java:22)
at org.corfudb.generator.main.lambda$main$1(main.java:44)
at java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1386)
at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)

Maithem:
Fixed by https://github.com/CorfuDB/CorfuDB/pull/854

PR #854, opened 2017-07-18
Synchronize MultiObjectSMREntry

Maithem:
Synchronize reads and writes on a MultiObjectSMREntry.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
@@ -6,9 +6,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 
-import lombok.Getter;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.ILogData;
@@ -18,7 +18,7 @@ import org.corfudb.util.serializer.Serializers;
 
 
 /**
- * A log entry sturcture which contains a collection of multiSMRentries,
+ * A log entry structure which contains a collection of multiSMREntries,
  * each one contains a list of updates for one object.
  */
 @Deprecated // TODO: Add replacement method that conforms to style
@@ -28,11 +28,11 @@ import org.corfudb.util.serializer.Serializers;
 public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
 
     // map from stream-ID to a list of updates encapsulated as MultiSMREntry
-    @Getter
-    public Map<UUID, MultiSMREntry> entryMap = new HashMap<>();
+    private Map<UUID, MultiSMREntry> entryMap;
 
     public MultiObjectSMREntry() {
         this.type = LogEntryType.MULTIOBJSMR;
+        entryMap = new HashMap<>();
     }
 
     public MultiObjectSMREntry(Map<UUID, MultiSMREntry> entryMap) {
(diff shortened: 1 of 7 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **dfa6110**.

 [Pull Request #854 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/PxalEQIOGlLQkQ83i02XkYYyBy3XC5Pf)

slfritchie:
This looks like it's trying to address #853.

Maithem:
@slfritchie Yes, I think its a timing issue thats related to memory, I don't think I can write an explicit  test case. I believe it happens by the 3rd get. Also, apparently current is not called unless that list is not empty ?

Maithem:
@slfritchie Also, I dont always see a ConcurrentModificationException, Its mostly seems that stale reads of that map are causing the out of bound exception

slfritchie:
Hm, are there two problems here?

1. Stale read. This PR seems good but I can't figure out if its sufficient for all stale read cases.
2. `getWriteSetEntryList()` returning an empty list, which is what I've interpreted #853 to be.  If returning an empty list is possible, then this PR isn't enough to avoid WriteSetSMRStream::current throwing IOOB exceptions.

Maithem:
@slfritchie It seems like other places of the code expect it to return a null, if the index doesnt exist, but also returning a null would mean throwing a runtime exception NoRollbackException(stream.pos(), rollbackVersion);

runtime/src/main/java/org/corfudb/runtime/object/transactions/WriteSetSMRStream.java
```diff
@@ -167,10 +167,17 @@ void mergeTransaction() {
         if (Address.nonAddress(currentContextPos)) {
             currentContextPos = -1;
         }
-        return Collections.singletonList(contexts
-                .get(currentContext)
-                .getWriteSetEntryList(id)
-                .get((int)(currentContextPos)));
+
+        try {
+            return Collections.singletonList(contexts
```
Maithem: remove

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a61f79d**.

 [Pull Request #854 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/a8MtAJek1zra7WpVrh3FCilxBa2VZcsU)

slfritchie:
Did your work last night succeed in finding the reason for why synchronizing the map in this class works as it does?

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -28,11 +28,15 @@
 public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
 
     // map from stream-ID to a list of updates encapsulated as MultiSMREntry
-    @Getter
-    public Map<UUID, MultiSMREntry> entryMap = new HashMap<>();
+    private Map<UUID, MultiSMREntry> entryMap;
+
+    synchronized public Map<UUID, MultiSMREntry> getEntryMap(){
```
medhavidhawan: do we really need the getter for this map anymore ?

runtime/src/main/java/org/corfudb/runtime/object/transactions/WriteSetSMRStream.java
```diff
@@ -167,10 +167,17 @@ void mergeTransaction() {
...
+                    .get(currentContext)
+                    .getWriteSetEntryList(id)
+                    .get((int) (currentContextPos)));
+        } catch (Exception e) {
+            e.printStackTrace();
```
medhavidhawan: remove these .

runtime/src/main/java/org/corfudb/runtime/object/transactions/WriteSetSMRStream.java
```diff
@@ -167,10 +167,17 @@ void mergeTransaction() {
...
         }
-        return Collections.singletonList(contexts
-                .get(currentContext)
-                .getWriteSetEntryList(id)
-                .get((int)(currentContextPos)));
+
+        try {
+            return Collections.singletonList(contexts
+                    .get(currentContext)
+                    .getWriteSetEntryList(id)
```
medhavidhawan: if I were to tighten this up, I would just have getWriteSetEntry(id, currentContextPost) or something to that effect. This leaves a lose end for other bugs.

Maithem:
@slfritchie Yes, it was what I described earlier. You have a reader and writer in two different threads and access is no synchronized, so some of the reads were causing an out of bound exception

Maithem:
Addressed all comments.

PR #862, opened 2017-07-21
Collection sync

Maithem:
Use synchronized collections for MultiObjectSMREntry and
MultiSMREntry. This fixes a race condition, where while a thread
is writing another thread can observe stale reads.


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **44b53f8**.

 [Pull Request #862 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/9H3pNP32pw2RPdXAJ6M72Cl1UT12EsIh)

medhavidhawan:
Are we going to look into restructuring this code or just using
synchronized data stuctures is final fix. To me this looks like duct tape.

Maithem:
@medhavidhawan Ive been thinking about this for some time and i dont think you can protect against concurrent modification unless your objects are immutable or you use heavy synchronization. Even when moving the state into the object class, you can leak references. Anyways lets sync up on this because the code is convoluted and we need to start addressing this problem proactively.

PR #871, opened 2017-07-24
Buffer Checkpoints

Maithem:
A Checkpoint should be admitted to the QueuedStreamContext only
when the whole checkpoint has been read. This patch prevents
partial checkpoint from being considered. This is required,
because partial checkpoints can cause a false TrimmedException
to be thrown.

PR #876, opened 2017-07-26
GH871 alternative patch

slfritchie:
This is an alternative to the patch to #871.  I didn't have a complete understanding of the circumstances that #871 is supposed to fix, so I made my best guess.  If I'm wrong, then this PR should be closed.

My understanding was that @Maithem was running into a case where `fillReadQueue` would encounter at `TrimmedException` before a checkpoint `START` record was found, and then the map state was corrupted because the checkpoint's data was not discarded.  There are a few cases where a `TrimmedException` can be thrown by runtime without seeing the exception from the server; see AbstractQueuedStreamView.java.

This PR will forcibly clear the stream context's `readCpQueue` if `START` is not seen.

### DIFF
--- a/test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
+++ b/test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
@@ -90,7 +90,26 @@ public class CheckpointSmokeTest extends AbstractViewTest {
      */
     @Test
 	public void smoke1Test() throws Exception {
-        final String streamName = "mystream";
+        boolean[] b = new boolean[]{true, false};
+        int counter = 0;
+
+        for (int i = 0; i < 2; i++) {
+            for (int j = 0; j < 2; j++) {
+                for (int k = 0; k < 2; k++) {
+                    // Avoid the case of writing START & END but not the
+                    // checkpoint data in the middle CONTINUATION (index=j).
+                    if (! (b[i] && !b[j] && b[k])) {
+                        System.err.printf("try %s %s %s %d\n", b[i], b[j], b[k], counter);
+                        smoke1Test(b[i], b[j], b[k], counter++);
+                    }
+                }
+            }
+        }
+    }
+
+    public void smoke1Test(boolean write1, boolean write2, boolean write3, int counter)
+            throws Exception {
+        final String streamName = "mystream" + counter;
         final UUID streamId = CorfuRuntime.getStreamID(streamName);
         final String key1 = "key1";
         final long key1Val = 42;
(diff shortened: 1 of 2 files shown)


test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -90,7 +90,26 @@ public void testEmptyMapCP() throws Exception {
...
+        boolean[] b = new boolean[]{true, false};
+        int counter = 0;
+
+        for (int i = 0; i < 2; i++) {
+            for (int j = 0; j < 2; j++) {
+                for (int k = 0; k < 2; k++) {
+                    // Avoid the case of writing START & END but not the
+                    // checkpoint data in the middle CONTINUATION (index=j).
+                    if (! (b[i] && !b[j] && b[k])) {
+                        System.err.printf("try %s %s %s %d\n", b[i], b[j], b[k], counter);
```
no2chem: System.err doesn't belong in a test

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -90,7 +90,26 @@ public void testEmptyMapCP() throws Exception {
      */
     @Test
 	public void smoke1Test() throws Exception {
```
no2chem: smoke1test is not a good name for a test

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -90,7 +90,26 @@ public void testEmptyMapCP() throws Exception {
...
-        final String streamName = "mystream";
+        boolean[] b = new boolean[]{true, false};
+        int counter = 0;
+
+        for (int i = 0; i < 2; i++) {
+            for (int j = 0; j < 2; j++) {
+                for (int k = 0; k < 2; k++) {
+                    // Avoid the case of writing START & END but not the
+                    // checkpoint data in the middle CONTINUATION (index=j).
+                    if (! (b[i] && !b[j] && b[k])) {
```
no2chem: It's pretty hard to determine what this test does - what is i, j and k?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **3742ed7**.

 [Pull Request #876 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/TNK4yknF2FWBPLlWUkx97ckC1O4eqqaT)

slfritchie:
Addressed by #889

PR #878, opened 2017-07-26
Reduce unit test execution time from ~750 seconds to ~200 seconds

slfritchie:
1. Adjust CorfuRuntime retry timeout units & mutability for unit tests.

Recent changes to the Sequencer have created a race condition that
many unit tests lose: when the Sequencer is in not ready state,
loser tests must loop & wait 5 seconds before resuming.  On average,
100-110 tests are delayed by this wait.  The new timeout is 25 msec.

2. Shut down ManagementServer's thread pool but don't wait for
it to finish before continuing with JUnit's @After cleanup tasks.

3. Use smallest parameter (`TIMEOUT_VERY_SHORT`) for sleep constant
in CheckpointTest.java.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
@@ -544,12 +544,7 @@ public class ManagementServer extends AbstractServer {
             corfuRuntime.shutdown();
         }
 
-        try {
-            failureDetectorService.awaitTermination(serverContext.SHUTDOWN_TIMER.getSeconds(),
-                    TimeUnit.SECONDS);
-        } catch (InterruptedException ie) {
-            log.debug("failureDetectorService awaitTermination interrupted : {}", ie);
-        }
+        failureDetectorService.shutdownNow();
         log.info("Management Server shutting down.");
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -46,6 +46,9 @@ public class CorfuRuntime {
 
     static final long DEFAULT_BATCH_FOR_FAST_LOADER = 5;
     static final int DEFAULT_TIMEOUT_MINUTES_FAST_LOADING = 30;
+    @Getter
+    @Setter
+    static int defaultRetryMilliseconds = 5000;
 
     @Data
     public static class CorfuRuntimeParameters {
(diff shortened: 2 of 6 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **ff03230**.

 [Pull Request #878 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/I4t5PuBr3ikg4r9L1Kc6mez1qNPNK35Q)

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -544,12 +544,7 @@ public void shutdown() {
             corfuRuntime.shutdown();
         }
 
-        try {
-            failureDetectorService.awaitTermination(serverContext.SHUTDOWN_TIMER.getSeconds(),
-                    TimeUnit.SECONDS);
-        } catch (InterruptedException ie) {
-            log.debug("failureDetectorService awaitTermination interrupted : {}", ie);
-        }
+        failureDetectorService.shutdownNow();
```
zalokhan: We are already performing a `failureDetectorService.shutdownNow()` in the beginning of this method.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -46,6 +46,9 @@
 
     static final long DEFAULT_BATCH_FOR_FAST_LOADER = 5;
     static final int DEFAULT_TIMEOUT_MINUTES_FAST_LOADING = 30;
+    @Getter
+    @Setter
+    static int defaultRetryMilliseconds = 5000;
```
dahliamalkhi: shouldn't this be all upper case and final, and the getter/setter on retryRate?

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
slfritchie: Whoops, sorry about overlooking that.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
slfritchie: Upper case, sure.  Final, no, not without also chasing down many/all instances of tests creating their own CorfuRuntime objects and having all of instances configure non-default retry times.   A mutable class static var "hammer" is the right tool here, I think, to keep test times from creeping upward yet again.

xnull:
outdated

PR #884, opened 2017-07-27
Fix Bulk Read

Maithem:
Removed Range Reads and added chucking to readAll.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -184,13 +184,10 @@ public class LogUnitServer extends AbstractServer {
     @ServerHandler(type = CorfuMsgType.READ_REQUEST, opTimer = metricsPrefix + "read")
     private void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r,
                       boolean isMetricsEnabled) {
-        log.trace("log read: {} {}", msg.getPayload().getStreamID() == null
-                        ? "global" : msg.getPayload().getStreamID(),
-                msg.getPayload().getRange());
+        log.trace("read: {}", msg.getPayload().getList());
         ReadResponse rr = new ReadResponse();
         try {
-            for (Long l = msg.getPayload().getRange().lowerEndpoint();
-                    l < msg.getPayload().getRange().upperEndpoint() + 1L; l++) {
+            for (Long l : msg.getPayload().getList()) {
                 ILogData e = dataCache.get(l);
                 if (e == null) {
                     rr.put(l, LogData.EMPTY);
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
@@ -1,10 +1,9 @@
 package org.corfudb.protocols.wireprotocol;
 
-import com.google.common.collect.Range;
-
 import io.netty.buffer.ByteBuf;
 
-import java.util.UUID;
+import java.util.Arrays;
+import java.util.List;
 
 import lombok.AllArgsConstructor;
 import lombok.Data;
@@ -16,9 +15,7 @@ import lombok.Data;
 @AllArgsConstructor
 public class ReadRequest implements ICorfuPayload<ReadRequest> {
 
-    final Range<Long> range;
-    @SuppressWarnings({"checkstyle:abbreviationaswordinname", "checkstyle:membername"})
-    final UUID streamID;
+    final List<Long> list;
 
     /**
      * Deserialization Constructor from ByteBuf to ReadRequest.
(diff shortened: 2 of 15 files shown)


Maithem:
Needs more clean up, will do it tomorrow.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **8a2b7d8**.

 [Pull Request #884 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Bgic7Pp7t3VzQn8t6W4cCLzDqo68nybJ)

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -300,17 +303,25 @@ public void invalidateClientCache() {
...
      */
     public @Nonnull Map<Long, ILogData> cacheFetch(Iterable<Long> addresses) {
-        //turn the addresses into Set for now to satisfy signature requirement down the line
-        Set<Long> readAddresses = new TreeSet<>();
-        Iterator<Long> iterator = addresses.iterator();
-        while(iterator.hasNext()){
-            readAddresses.add(iterator.next());
+        Map<Long, ILogData> temp = new HashMap<>();
+
+        int batchSize = 20;
```
slfritchie: This constant seems like something that someone soon will want to make configurable/adjustable?

test/src/test/java/org/corfudb/integration/StreamIT.java
```diff
@@ -1,68 +1,53 @@
...
 import java.util.UUID;
+import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * A set integration tests that exercise the stream API.
  */
 
-public class StreamIT extends AbstractIT {
```
slfritchie: Hm, the surgery here appears to have changed the integration style test into a unit test.  I think it'd be better if this test had adaptations made within the IT style, then, if you wish, add a new unit test(s).

Also, there are two large blocks of code commented out.  If they aren't useful, then perhaps it's better to remove this entire test and/or convert to unit test style only?

Maithem: im discarding the test related changes, I just reused that test to run on a live system

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -184,13 +184,10 @@ public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx,
     @ServerHandler(type = CorfuMsgType.READ_REQUEST, opTimer = metricsPrefix + "read")
     private void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r,
                       boolean isMetricsEnabled) {
-        log.trace("log read: {} {}", msg.getPayload().getStreamID() == null
-                        ? "global" : msg.getPayload().getStreamID(),
-                msg.getPayload().getRange());
+        log.trace("log read: {}", msg.getPayload().getSet());
```
no2chem: read: is fine, the logger will print the calling class

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
```diff
@@ -16,36 +18,24 @@
 @AllArgsConstructor
 public class ReadRequest implements ICorfuPayload<ReadRequest> {
 
-    final Range<Long> range;
-    @SuppressWarnings({"checkstyle:abbreviationaswordinname", "checkstyle:membername"})
-    final UUID streamID;
+    final Set<Long> set;
```
no2chem: Hm, let's not use set here - use a list. If the caller requests multiple addresses twice, we'll just return it, no need to enforce this.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -300,17 +303,25 @@ public void invalidateClientCache() {
...
+        Iterable<List<Long>> batches = Iterables.partition(addresses, batchSize);
+
+        for (List<Long> batch : batches) {
+            //turn the addresses into Set for now to satisfy signature requirement down the line
+            Set<Long> readAddresses = new TreeSet<>();
```
no2chem: See above comment to keep the data structure a list down to the message

Maithem:
@no2chem I will introduce a range read in another iteration. Let's phase this in one at a time, we dont have code that uses a ranged read.

no2chem:
@Maithem well, the existing code uses a ranged read, so I would prefer if we preserved it...

Maithem:
@no2chem I didn't say we shouldn't. @medhavidhawan recommended that we phase them in two different patches.

no2chem:
@Maithem well, in this patch, create a new API (READ_MULTIPLE)? Instead of getting rid of the previous

Maithem:
@no2chem So introduce a new api and use it while leaving the other API broken?

no2chem:
@Maithem unless I'm mistaken, the other API works properly, it was just (1) used improperly (2) lacks an argument/bound check

Maithem:
@no2chem The READ_REQUEST api, conflates two apis, read a single address and read range.

no2chem:
@Maithem Ok, in a future PR, rename it READ_RANGE. But leave it intact, unless you can provide evidence to your claim that it is broken.

Maithem:
Will open a new pr

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **b753944**.

 [Pull Request #884 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/MnFIfC5X5h22LnEMVClqj6rEykdNlWQ2)

PR #888, opened 2017-07-31
Implicit transactional methods are not Nested

rogermichoud:
If we already are in a transaction when we call implicit
transactional method, we don't have to nest it. The transactional
method will be executed in the current transactional context.

Since Nested Transaction code path is not very stable yet, let's minimize the risk and not use them implicitly.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -330,6 +330,15 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
     @Deprecated // TODO: Add replacement method that conforms to style
     @SuppressWarnings({"checkstyle:membername", "checkstyle:abbreviation"}) // Due to deprecation
     private <R> R TXExecuteInner(Supplier<R> txFunction, boolean isMetricsEnabled) {
+        // Don't nest transactions if we are already running transactionally
+        if (TransactionalContext.isInTransaction()) {
+            try {
+                return txFunction.get();
+            } catch (Exception e) {
+                log.warn("TXExecute[{}] Abort with Exception: {}", this, e);
+                this.abortTransaction(e);
+            }
+        }
         long sleepTime = 1L;
         final long maxSleepTime = 1000L;
         int retries = 1;
--- a/test/src/test/java/org/corfudb/runtime/object/transactions/ImplicitTransactionsTest.java
+++ b/test/src/test/java/org/corfudb/runtime/object/transactions/ImplicitTransactionsTest.java
@@ -1,0 +1,89 @@
+package org.corfudb.runtime.object.transactions;
+
+import com.google.common.reflect.TypeToken;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.SMRMap;
+import org.corfudb.runtime.object.CorfuCompileProxy;
+import org.corfudb.runtime.object.ICorfuSMR;
+import org.corfudb.runtime.object.TransactionalObject;
+import org.corfudb.runtime.view.ObjectsView;
+import org.junit.Test;
+
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Created by rmichoud on 7/28/17.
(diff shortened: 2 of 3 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **fd9e5bc**.

 [Pull Request #888 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/38spDls0oaw54rDdwB3ctxARA7oJwPAU)

test/src/test/java/org/corfudb/runtime/object/transactions/ImplicitTransactionsTest.java
```diff
@@ -0,0 +1,86 @@
...
+                .setStreamName("test-1")
+                .setTypeToken(new TypeToken<SMRMap<String,String>>() {})
+                .open();
+
+
+        TXBegin();
+        m1.put("k1", "v1");
+
+        CorfuCompileProxy cp = getProxy(rt, "test-1");
+        boolean isInNestedTx = (boolean) cp.TXExecute(TransactionalContext::isInNestedTransaction);
```
no2chem: I realize the TransactionContext doesn't require a runtime reference. Why not create a custom object to test this condition instead (with isNestedTransaction as a method)?

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -330,6 +330,10 @@ public UUID getStreamID() {
     @Deprecated // TODO: Add replacement method that conforms to style
     @SuppressWarnings({"checkstyle:membername", "checkstyle:abbreviation"}) // Due to deprecation
     private <R> R TXExecuteInner(Supplier<R> txFunction, boolean isMetricsEnabled) {
+        // This should not nest transactions
```
no2chem: Maybe a better comment would be, "don't nest transactions if we are already running transactionally"

slfritchie:
@rogermichoud @no2chem In the other execution paths of `TXExecuteInner`, any exception will abort the transaction initiated by itself.  Roger's patch checks for an existing txn but does not abort that existing txn when an exception is thrown.  Also, the other callbacks such as `getUpcallResultInner` are aborting the existing transaction when an exception is hit.

This lack of symmetry looks odd.  For example, if I plant a fake `ArithmeticException` in `SMRMap::putIfAbsent`, I see that arithmetic error leak out of this patch's code *and also* the transaction has not been aborted.  In contrast, if I plant the same exception in `getUpcallResultInner`, I see what feels to be correct: an `TransactionAbortedException` with cause of `ArithmeticException`.

test/src/test/java/org/corfudb/runtime/object/transactions/ImplicitTransactionsTest.java
rogermichoud: What do you mean by a custom object ?  A test Corfu Object that have an isInNestedTransaction @transactional method ? What is the benefit of doing it with a custom object instead of with the CorfuCompileProxy ?

rogermichoud:
@slfritchie Is this any different that if you call a function within a transaction and this function yield an exception ? If I throw an ArithmeticException during a Transaction, does it get aborted ?

test/src/test/java/org/corfudb/runtime/object/transactions/ImplicitTransactionsTest.java
no2chem: It keeps us from overspecifying the test.

We really want to test if an object with an @TransactionalMethod annotation really doesn't nest,  not whether TXExecute nests or not.

no2chem:
@slfritchie You're right, we should catch the exception and abort the transaction if there is one, and if there is not, we should throw the exception. @rogermichoud can you add that bit as well?

rogermichoud:
@slfritchie @no2chem 
Addressed the comments

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **1665263**.

 [Pull Request #888 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/L2YtI4zv2mBvi64hLb9QlqUHYPl6thd9)

PR #896, opened 2017-08-06
Configurable corfu metrics reporting

zalokhan:
Adding metrics registry as a configurable parameter so that the application can pass its own registry and report the metrics in its own convenient way when required.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -132,11 +132,10 @@ public class LogUnitServer extends AbstractServer {
                 .maximumWeight(maxCacheSize)
                 .removalListener(this::handleEviction)
                 .writer(batchWriter)
-                .recordStats()
                 .build(this::handleRetrieval);
 
         MetricRegistry metrics = serverContext.getMetrics();
-        MetricsUtils.addCacheGauges(metrics, metricsPrefix + "cache.", dataCache);
+//        MetricsUtils.addCacheGauges(metrics, metricsPrefix + "cache.", dataCache);
 
         Runnable task = () -> streamLog.compact();
         compactor = scheduler.scheduleAtFixedRate(task, 10, 45, TimeUnit.MINUTES);
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
@@ -78,8 +78,8 @@ public class ServerContext {
         String mp = "corfu.server.";
         synchronized (metrics) {
             if (!isMetricsReportingSetUp(metrics)) {
-                addJvmMetrics(metrics, mp);
-                MetricsUtils.addCacheGauges(metrics, mp + "datastore.cache.", dataStore.getCache());
+//                addJvmMetrics(metrics, mp);
+//                MetricsUtils.addCacheGauges(metrics, mp + "datastore.cache.", dataStore.getCache());
                 MetricsUtils.metricsReportingSetup(metrics);
             }
         }
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -16,6 +16,7 @@ import java.util.stream.Collectors;
 
 import lombok.Data;
 import lombok.Getter;
+import lombok.NonNull;
 import lombok.Setter;
 import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
@@ -211,7 +212,14 @@ public class CorfuRuntime {
     @Getter
     private static final String mpObj = mp + "object.";
     @Getter
-    private static final MetricRegistry metrics = new MetricRegistry();
+    private static MetricRegistry defaultMetrics = new MetricRegistry();
+    @Getter
+    private MetricRegistry metrics = new MetricRegistry();
+
+    public CorfuRuntime setMetrics(@NonNull MetricRegistry metrics) {
+        this.metrics = metrics;
+        return this;
+    }
 
     /**
      * When set, overrides the default getRouterFunction. Used by the testing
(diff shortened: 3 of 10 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -211,7 +211,9 @@
     @Getter
     private static final String mpObj = mp + "object.";
     @Getter
-    private static final MetricRegistry metrics = new MetricRegistry();
+    private static MetricRegistry defaultMetrics = new MetricRegistry();
```
Maithem: You probably don't want it to be static. Different Runtimes should have different registries

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -270,13 +273,18 @@ public CorfuRuntime() {
...
+        this(configurationSring, getDefaultMetrics());
+    }
+
     /**
      * Parse a configuration string and get a CorfuRuntime.
      *
      * @param configurationString The configuration string to parse.
      */
-    public CorfuRuntime(String configurationString) {
+    public CorfuRuntime(String configurationString, MetricRegistry metricRegistry) {
```
Maithem: Don't pass this as a argument. Use setters like other configurable options.

zalokhan: I need to pass this as a parameter because it needs to pass this metrics registry before it instantiates the client routers(as they need the registry too).
Else as you are saying we can get away with a setter if we make sure to set it before we call connect (that is when the routers are created.)

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
zalokhan: We tried that. But then the address space view requires the metrics registry which it obtains statically. If we make it an object parameter, we would need to pass the registry to the address space view too on creation. Would that be fine?

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **917ff5d**.

 [Pull Request #896 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/w0uJdaPBPYHfmvoNfnhXAuw9VXApWkPn)

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
slfritchie: @zalokhan I think that Maithem's suggestion is good.  It seems to me reasonable to assume that if you wish to use a custom registry, then it must be configured prior to calling connect().  The same goes for many of the other set'able runtime values and also configuring the use of TLS.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **c72a032**.

 [Pull Request #896 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/SmhN5Q0zZW81CnLU1V70T5Mg7ftcRDHG)

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -136,7 +136,7 @@ public LogUnitServer(ServerContext serverContext) {
                 .build(this::handleRetrieval);
 
         MetricRegistry metrics = serverContext.getMetrics();
-        MetricsUtils.addCacheGauges(metrics, metricsPrefix + "cache.", dataCache);
+//        MetricsUtils.addCacheGauges(metrics, metricsPrefix + "cache.", dataCache);
```
dahliamalkhi: yes, but then also turn off 'recordStats()' on line 135

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -78,8 +78,8 @@ public ServerContext(Map<String, Object> serverConfig, IServerRouter serverRoute
         String mp = "corfu.server.";
         synchronized (metrics) {
             if (!isMetricsReportingSetUp(metrics)) {
-                addJvmMetrics(metrics, mp);
-                MetricsUtils.addCacheGauges(metrics, mp + "datastore.cache.", dataStore.getCache());
+//                addJvmMetrics(metrics, mp);
+//                MetricsUtils.addCacheGauges(metrics, mp + "datastore.cache.", dataStore.getCache());
```
dahliamalkhi: same

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -244,7 +252,8 @@
                 try {
                     router.addClient(new LayoutClient())
                             .addClient(new SequencerClient())
-                            .addClient(new LogUnitClient())
+                            .addClient(new LogUnitClient().setMetricRegistry(metrics != null
+                                            ? metrics : CorfuRuntime.getDefaultMetrics()))
```
dahliamalkhi: not sure why LogUnitClient is the only one that needs to maintain the MetricsRegistry in its own local field; anyway, this is legacy, not new.

dahliamalkhi:
@no2chem I think we actually want to keep it for now. there is no evidence that metering actually causes a performance degradation, we are just playing it safe for now.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zalokhan: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
zalokhan: Disabled DataStore in ServerContext cache recordStats()

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
zalokhan: Yes long term plan is to introduce annotations like the ones used in MP. @Timed, @Metered, etc.

zalokhan:
@dahliamalkhi Made the changes.
@no2chem Yes, I think we need to keep them, will work on a later patch to filter these metrics on demand without a lot of checks causing performance degradation.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **7c80c1d**.

 [Pull Request #896 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/2WYU7j25OyhXD1QzTtX5txwH3U80SQVw)

PR #900, opened 2017-08-08
Distribution Sample

Maithem:
A simple implementation of reservoir sampling.

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/SamplingUtils.java
+++ b/runtime/src/main/java/org/corfudb/util/SamplingUtils.java
@@ -1,0 +1,53 @@
+package org.corfudb.util;
+
+import static com.carrotsearch.sizeof.RamUsageEstimator.sizeOf;
+
+import java.util.Arrays;
+
+/**
+ * Select a random subset k, from a stream that has size n,
+ * but n is unknown (i.e. online sampling).
+ *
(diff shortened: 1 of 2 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a48d63a**.

 [Pull Request #900 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/u7Cx1F0vfJsk13GyBrsw2scGjR1Z1AlH)

runtime/src/main/java/org/corfudb/util/SamplingUtils.java
```diff
@@ -0,0 +1,53 @@
...
+ * Created by maithem on 8/7/17.
+ */
+public class SamplingUtils {
+
+    private int k;
+    private long n;
+    private long[] sample;
+    private int x;
+
+    public SamplingUtils(int k, long p) {
```
annym: what is this p used for?

Maithem: Initially, I wanted p to be the percentage of n samples that should be considered, but in this case n is not known ahead of time. I will remove, since we just need a random subset of size k to consider.

PR #902, opened 2017-08-10
CorfuTable can modify the indexer at runtime

rogermichoud:
Main change is that we can modify the indexer of an existing
CorfuTable. This enable FastObjectLoader to reload the maps in
memory and postpone the computation of the index at instatiation
time. Also include minor fixes.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -22,6 +22,7 @@ import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
 
+import lombok.Getter;
 import org.corfudb.annotations.Accessor;
 import org.corfudb.annotations.ConflictParameter;
 import org.corfudb.annotations.CorfuObject;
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -146,6 +148,12 @@ public boolean isEmpty() {
         return mainMap.isEmpty();
     }
 
+    @Accessor
+    public boolean hasIndices() {return !indexFunctions.isEmpty();}
+
+    @Getter
+    public Class indexerClass;
```
no2chem: I think this should be a ```Class<F>```

rogermichoud: it doesn't "compile". CorfuMap#CORFUSMR.java is complaining about not knowing the symbol class F

no2chem: Ah. That sounds like a annotationProcessor issue. Never mind for now then.

rogermichoud:
@no2chem @medhavidhawan Before merging, medhavi wants to review it as well. He also asked for a fallback on getByindex. I will do the following, if there is no indexer, fallback to scan and filter.

rogermichoud:
@no2chem @medhavidhawan On second thought, this would be a bit more than just a small fix. We would need to generate a predicate from the functions and degenerate to the scan and filter with the predicate. We can postpone this, I opened an Issue #903

no2chem:
Sounds good. I think it would be ok to merge this into the main PR, since @medhavidhawan can just review that.

rogermichoud:
doing the review with him now

rogermichoud:
Good to merge in your branch

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **e31646b**.

 [Pull Request #902 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/FC2sKziZwFSTCC1UrO501egYZdH0eRfA)

PR #907, opened 2017-08-11
Use ArrayList::new in CorfuTable

no2chem:
This patches getByIndexAndFilter to collect using ArrayList::new, which performs significantly
better than Collections.toList.


annym:
Looks good to me!

no2chem:
@annym Can you approve so it can be merged?

no2chem:
@annym Ok, take a look again

PR #912, opened 2017-08-16
Increase logging in case of dropped indexes

Maithem:


medhavidhawan:
In general can we project outside the accessor lock then objects will be
locked for more reasonable time

PR #923, opened 2017-08-21
Print Write Size

Maithem:
Print the write size on an OverwriteException.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
@@ -11,6 +11,9 @@ import javax.annotation.Nullable;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.protocols.logprotocol.StreamCOWEntry;
+import org.corfudb.protocols.wireprotocol.DataType;
+import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.protocols.wireprotocol.TokenType;
 import org.corfudb.protocols.wireprotocol.TxResolutionInfo;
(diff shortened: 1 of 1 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d2d2e4f**.

 [Pull Request #923 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/S0Y73aIZ4fvdw4tq7UKSS3EPl8IpMOgw)

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -142,9 +145,17 @@ public long append(@Nonnull Set<UUID> streamIDs, @Nonnull Object object,
                 // If we're here, we succeeded, return the acquired token
                 return tokenResponse.getTokenValue();
             } catch (OverwriteException oe) {
+                int sizeOfWrite = 0;
+                try {
+                    ILogData.SerializationHandle handle = new LogData(DataType.DATA, object).getSerializedForm();
+                    sizeOfWrite = handle.getSerialized().getSizeEstimate();
+                    handle.close();
+                } catch (Exception e) {
+                    log.warn("Couldnt get size", e);
```
no2chem: Log message format not correct

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -142,9 +145,17 @@ public long append(@Nonnull Set<UUID> streamIDs, @Nonnull Object object,
...
+                }
                 // We were overwritten, get a new token and try again.
-                log.warn("append[{}]: Overwrite , streams {}", tokenResponse.getTokenValue(),
-                        streamIDs.stream().map(Utils::toReadableId).collect(Collectors.toSet()));
+                log.warn("append[{}]: Overwrite , streams {}, size {}", tokenResponse.getTokenValue(),
```
no2chem: Does this really need to be warn level?

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -142,9 +145,17 @@ public long append(@Nonnull Set<UUID> streamIDs, @Nonnull Object object,
                 // If we're here, we succeeded, return the acquired token
                 return tokenResponse.getTokenValue();
             } catch (OverwriteException oe) {
+                int sizeOfWrite = 0;
```
no2chem: This calculation should only be done if the log is set to output warn level messages

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -142,9 +145,17 @@ public long append(@Nonnull Set<UUID> streamIDs, @Nonnull Object object,
...
+                try {
+                    ILogData.SerializationHandle handle = new LogData(DataType.DATA, object).getSerializedForm();
```
no2chem: Probably should be a static method inside of ILogData.

Maithem:
addressed by https://github.com/CorfuDB/CorfuDB/pull/925

PR #943, opened 2017-09-13
Fix stream previous/seek calls

no2chem:
A bug in the implementation of previous could result in incorrect data being returned by a stream, if previous is called right after a seek.
This PR fixes this issue and adds a unit test which demonstrates the bug.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -274,7 +274,7 @@ public abstract class AbstractQueuedStreamView extends
         final QueuedStreamContext context = getCurrentContext();
         final long oldPointer = context.globalPointer;
 
-        log.trace("Previous[{}] max={} min={}", this,
+        log.trace("previous[{}]: max={} min={}", this,
                 context.maxResolution,
                 context.minResolution);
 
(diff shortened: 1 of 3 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **59224ca**.

 [Pull Request #943 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Iiap8VqL58qsMuEsMy3cjv6nx4R9ZdKd)

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -235,13 +235,6 @@ protected boolean followBackpointers(final UUID streamId,
 
         // Loop until we have reached the stop address.
         while (currentAddress > stopAddress  && Address.isAddress(currentAddress)) {
-            // The queue already contains an address from this
```
medhavidhawan: what was this for earlier. I cannot understand it  now.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -297,20 +299,23 @@ public synchronized ILogData previous() {
...
             remainingUpTo(context.minResolution);
             context.minResolution = Address.NON_ADDRESS;
             context.globalPointer = oldPointer;
             prevAddress = context
                     .resolvedQueue.lower(context.globalPointer);
-            log.trace("Previous[}] updated queue {}", this, context.resolvedQueue);
+            log.trace("previous[{}]: updated resolved queue {}", this, context.resolvedQueue);
         }
+
+        // Clear the read queue, it may no longer be valid
```
medhavidhawan: can we have some sort of picture or ascii art for this. I keep forgetting the whole mechanism and it is hard for me to review it after sometime. Would be nice if we can have a picture.

PR #957, opened 2017-10-05
Modifying version to 0.1.1

zalokhan:
Modifying version numbers for v0.1.1

### DIFF
--- a/annotationProcessor/pom.xml
+++ b/annotationProcessor/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>corfu</artifactId>
         <groupId>org.corfudb</groupId>
-        <version>0.1-SNAPSHOT</version>
+        <version>0.1.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>corfu</artifactId>
         <groupId>org.corfudb</groupId>
-        <version>0.1-SNAPSHOT</version>
+        <version>0.1.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/cmdlets/pom.xml
+++ b/cmdlets/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>corfu</artifactId>
         <groupId>org.corfudb</groupId>
-        <version>0.1-SNAPSHOT</version>
+        <version>0.1.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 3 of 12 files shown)


bitmingw:
Hi Zeeshan,

This patch is generally good. I want to mention two things.

1. By dropping `SNAPSHOT` it is going to affect the behavior of [Corfu Repos](https://github.com/CorfuDB/Corfu-Repos) for public use. You may want to keep `SNAPSHOT`. And what ever changes you made here need to be reflected in that repo.
2. Please cherry-pick this change to master branch after merge.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **74df9d4**.

 [Pull Request #957 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/VupUwQaews4gLJ1mO5xSYh7e8Q02RayP)

no2chem:
Hi, I'm not sure that this should go into master. Probably master needs to continue to be tagged snapshot, as @bitmingw points out. We'll create a branch for 0.1.1 and tag it so. Perhaps a better patch would be to use a tag in the build, if present?

zalokhan:
@bitmingw I'll keep this branch as it is for now.
You are right and I'll create a new PR to change the versioning of master branch to move to v0.2.0.
Only the master branch pushes and updates the Corfu Repo and the mvn repo. So this branch should not affect it.

bitmingw:
I believe we shouldn't tag 0.1.1 right now. Since tag should be immutable, it can only be used as we formally declare "release".

In the meanwhile, changing version number can be done earlier even if we could put other changes after that.

no2chem:
@bitmingw but a 0.1.1 artifact should only be generated as a result of a release. Anything else should be a snapshot, right?

bitmingw:
@no2chem Then how about the idea of 0.1.1-SNAPSHOT?

no2chem:
@bitmingw Yeah, that seems reasonable. But we should also make sure to make the tag automatically somehow a well, eventually.

bitmingw:
I agree that we need a formal process to approve a tag at some point.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **5b25616**.

 [Pull Request #957 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/1N9NfNtFpn3IXCcrCVBNuxxC8VqzqoHZ)

bitmingw:
@no2chem If I understand correctly, we can add a hook to git. If a maven project version number without SNAPSHOT is detected & that version is not tagged, a tag is generated automatically.

Is that what you want?

no2chem:
@bitmingw That would be nice (and maybe we can do that in another PR, seems like it will require some testing). For this immediate PR though, I'll be okay if the version # is taken from a variable.

bitmingw:
@zalokhan Can you make the corresponding change (0.1.1-SNAPSHOT)? We are scheduling a upgrade either today or tomorrow.

PR #962, opened 2017-10-12
Add node and Workflow management (First Draft)

zalokhan:
First draft for Add Node and Workflow management.

### Step
- Each workflow consists of a series of ordered steps.
- A step consists of 
  - Name and UUID for identification
  - process - function executed by this thread.
  - The step can take parameters to be passed to the process.

### Workflow See PR #965 
- Uses a builder to add steps (in sequence to be executed)
- Creates an immutable list of these steps.
- On executeAsync, these steps are executed in order (cannot be run in parallel). Currently there is no linking of steps (passing one's result as parameter to the next one)

### Status
- The status of the workflow is updated with the execution of its steps.

## Add Node
Add node comprises of 2 major workflows:

### Workflow 1: Add new node
- This comprises of 2 steps:
  - **Step 1:** Set up new node. This bootstraps the new node with the existing layout and initializes the management server.
  -  **Step 2:** Add new node as a part of the layout. If the new node is a part of the log unit chain or quorum, a split segment has to be executed which seals the existing segment and adds this new node as a part of the new segment. (No down time for data path operations.)

### Workflow 2: Merge segment
- This workflow is executed only if the new node to be added is going to be a log unit server.
- Steps of these workflow are dependent on the type of replication protocol being used. The current implementation strongly assumes Chain replication only. Abstraction to support other replication modes might be missing.
- Comprises of 3 steps:
  - **Step 1:** Catch-up segment. 
    - This step fetches a list of all known addresses from the existing head and tail of the chain.
    - The diff of all the addresses between these 2 sets is calculated.
    - A read is performed (on the tail) for all the missing addresses (diff set) which forces hole fills.

  - **Step 2:** Segment replication
    - At the source (any node in the existing chain as all nodes have the same state after catch-up) the files are copied into a buffer and sent to the destination node. **(No lock or snapshot is taken on the file being copied)**
    - At the destination, the file is opened by a temporary segment handler and copied to the existing file to preserve existing data.

  - **Step 3:** Merge Segment
    - The layout is now reconfigured to merge the last 2 segments.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ConservativeFailureHandlerPolicy.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ConservativeFailureHandlerPolicy.java
@@ -31,6 +31,7 @@ public class ConservativeFailureHandlerPolicy implements IFailureHandlerPolicy {
         LayoutWorkflowManager layoutManager = new LayoutWorkflowManager(originalLayout);
         Layout newLayout = layoutManager
                 .moveResponsiveSequencerToTop(failedNodes)
+                .clearUnResponsiveServers()
                 .addUnresponsiveServers(failedNodes)
                 .build();
         newLayout.setRuntime(corfuRuntime);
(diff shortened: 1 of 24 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **37847e8**.

 [Pull Request #962 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ziIBZ0Z7CucI6MvL9vWrA6lwYhqaYLf7)

zalokhan:
Will split this into smaller patches.

PR #970, opened 2017-10-20
Split segment and add node

zalokhan:
## Add Node

Prerequisites:
- #945 

#### Workflow 1: Add new node

This comprises of 2 steps:
- Step 1: Set up new node. This bootstraps the new node with the existing layout and initializes the management server.
- Step 2: Add new node as a part of the layout. If the new node is a part of the log unit chain or quorum, a split segment has to be executed which seals the existing segment and adds this new node as a part of the new segment. (No down time for data path operations.)

#### Workflow 2: Merge segment #971

PR #971, opened 2017-10-20
 Catchup, replicate and Merge segments

zalokhan:
### Workflow: Merge segment

This workflow is executed only if the new node to be added is going to be a log unit server.
Steps of these workflow are dependent on the type of replication protocol being used. The current implementation strongly assumes Chain replication only. Abstraction to support other replication modes might be missing.
Comprises of 3 steps:
- Step 1: Catch-up segment.
  - This step fetches a list of all known addresses from the existing head and tail of the chain.
  - The diff of all the addresses between these 2 sets is calculated.
  - A read is performed (on the tail) for all the missing addresses (diff set) which forces hole fills.

- Step 2: Segment replication
  - At the source (any node in the existing chain as all nodes have the same state after catch-up) the 
    files are copied into a buffer and sent to the destination node. (No lock or snapshot is taken on 
    the file being copied)
  - At the destination, the file is opened by a temporary segment handler and copied to the existing 
    file to preserve existing data.

- Step 3: Merge Segment
  - The layout is now reconfigured to merge the last 2 segments.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -28,6 +28,7 @@ import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.MultipleReadRequest;
+import org.corfudb.protocols.wireprotocol.RawDataMsg;
 import org.corfudb.protocols.wireprotocol.ReadRequest;
 import org.corfudb.protocols.wireprotocol.ReadResponse;
 import org.corfudb.protocols.wireprotocol.TrimRequest;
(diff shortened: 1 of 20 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **9e40dc7**.

 [Pull Request #971 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/11Laas5LJh5McOAvHHeOjQ6htBx4DACQ)

Maithem:
This is no longer required, correct?

zalokhan:
This is required.
This follows #970 Split segment and add node.
This PR contains the data replication and merge segment.
I closed the irrelevant PRs

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -286,6 +287,20 @@ private void flushCache(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r
         r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
     }
 
+    /**
+     * Handles requests for replicating
```
Maithem: We can't use this, the time it takes to transfer a segment is equal to the time it takes to write it to the address space. In other words, its slow. Also, there are many corner cases that are not considered by the replicateRawData implementation, for example the interleaving of trimmed addresses and non-trimmed addresses. 

I re-implemented it here: https://github.com/CorfuDB/CorfuDB/pull/1004/files

zalokhan:
Removing state transfer code as it will use batch writes.
Merging the rest of the "merge segments" code to #970

PR #973, opened 2017-10-24
FastLoader have a white list mode

rogermichoud:
A new API has been provided for the FastLoader to
selectively choose the streams it wants to load. It also
adds checkpoint streams for each stream added to the
whitelist.

Along, some refactoring of fastLoader testing.

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
@@ -9,7 +9,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
-import java.util.concurrent.*;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
@@ -43,6 +47,12 @@ import static org.corfudb.runtime.view.Address.isAddress;
  * be applied before the normal entries starting after the checkpoint start address.
  *
  * If used in the recoverSequencer mode, it will reconstruct the stream tails.
+ *
+ * There are two main modes, blacklist and whitelist. These two modes are mutually exclusive:
+ * In blacklist mode, we will process every streams as long as they are not in the streamToIgnore
+ * list. In whitelist mode, only the streams present in streamsToLoad will be loaded. We make
+ * sure to also include the checkpoint streams for each of them.
+ *
  *
  * Created by rmichoud on 6/14/17.
  */
@@ -89,6 +99,32 @@ public class FastObjectLoader {
     @Setter
     @Getter
     private boolean logHasNoCheckPoint = false;
+
+    private boolean whiteList = false;
+    private List<UUID> streamsToLoad = new ArrayList<>();
+
+    /**
+     * Enable whiteList mode where we only reconstruct
+     * the streams provided through this api. In this mode,
+     * we will only process the streams present in streamsToLoad.
+     * All the others are ignored.
+     * @param streamsToLoad
+     */
+    public FastObjectLoader addStreamsToLoad(List<String> streamsToLoad) {
+        if (streamsToIgnore.size() != 0) {
+            throw new IllegalStateException("Cannot add a whitelist when there are already streams to ignore");
+        }
+
+        whiteList = true;
+
+        streamsToLoad.forEach(streamName -> {
+            this.streamsToLoad.add(CorfuRuntime.getStreamID(streamName));
+            // Generate the streamsCheckpointId (we need to allow them as well)
+            this.streamsToLoad.add(CorfuRuntime.getCheckpointStreamIdFromName(streamName));
+        });
+
+        return this;
+    }
 
     /**
      * We can add streams to be ignored during the
@@ -140,7 +176,15 @@ public class FastObjectLoader {
     }
 
     public void addStreamToIgnore(String streamName) {
+        // In a whitelist mode, we cannot add streams to the blacklist
+        if (whiteList) {
+            throw new IllegalStateException("Cannot add a stream to the blacklist (streamsToIgnore)" +
+                    "in whitelist mode.");
+        }
+
         streamsToIgnore.add(CorfuRuntime.getStreamID(streamName));
+        // Ignore also checkpoint of this stream
+        streamsToIgnore.add(CorfuRuntime.getCheckpointStreamIdFromName(streamName));
     }
 
     /**
@@ -221,29 +265,85 @@ public class FastObjectLoader {
         }
     }
 
+    /**
+     * It is a valid state to have entries that are checkpointed but the log was not fully
+     * trimmed yet. This means that some entries are at the same time in their own slot and in
+     * the checkpoint. We must avoid to process them twice.
+     *
+     * This comes especially relevant when the operation order affects the end result.
+     * (e.g. clear() operation)
+     *
+     * @param streamId stream id to validate
+     * @param entry entry under scrutinization.
+     * @return if the entry is already part of the checkpoint we started from.
+     */
+    private boolean entryAlreadyContainedInCheckpoint(UUID streamId, SMREntry entry) {
+        return streamsMetaData.containsKey(streamId) && entry.getEntry().getGlobalAddress() <
+                streamsMetaData.get(streamId).getHeadAddress();
+    }
 
     /**
      * Check if this entry is relevant
      *
-     * Entries before checkpoint starts are irrelevant since they are
-     * contained in the checkpoint.
+     * There are 3 cases where an entry should not be processed:
+     *   1. In whitelist mode, if the stream is not in the whitelist (streamToLoad)
+     *   2. In blacklist mode, if the stream is in the blacklist (streamToIgnore)
+     *   3. If the entry was already processed in the previous checkpoint.
      *
      * @param streamId identifies the Corfu stream
      * @param entry entry to potentially apply
      * @return if we need to apply the entry.
      */
-    private boolean shouldEntryBeApplied(UUID streamId, SMREntry entry) {
+    private boolean shouldEntryBeApplied(UUID streamId, SMREntry entry, boolean isCheckpointEntry) {
+        // 1.
+        // In white list mode, ignore everything that is not in the list (the list contains the streams
+        // passed by the client + derived checkpoint streams).
+        if (whiteList && !streamsToLoad.contains(streamId)) {
+            return false;
+        }
+
+        // 2.
         // We ignore the transaction stream ID because it is a raw stream
         // We don't want to create a Map for it.
         if (streamId == runtime.getObjectsView().TRANSACTION_STREAM_ID || streamsToIgnore.contains(streamId)) {
             return false;
         }
 
-        if (!streamsMetaData.containsKey(streamId) || entry.getEntry().getGlobalAddress() >
-                streamsMetaData.get(streamId).getHeadAddress()) {
-            return true;
+        // 3.
+        // If the entry was already processed with the previous checkpoint.
+        if (!isCheckpointEntry && entryAlreadyContainedInCheckpoint(streamId, entry)) {
+            return false;
         }
-        return false;
+
+        return true;
+    }
+
+    private boolean shouldStreamBeProcessed(UUID streamId) {
+        if (whiteList) {
+            return streamsToLoad.contains(streamId);
+        }
+
+        return !streamsToIgnore.contains(streamId);
+    }
+
+    /**
+     * If none of the streams in the logData should be processed, we
+     * can simply ignore this logData.
+     *
+     * In the case of a mix of streams we need to process and other that we don't,
+     * we will still go ahead with the process.
+     *
+     * @param logData
+     * @return
+     */
+    private boolean shouldLogDataBeProcessed(ILogData logData) {
+        boolean shouldProcess = false;
+        for (UUID id : logData.getStreams()) {
+            if (shouldStreamBeProcessed(id)){
+                shouldProcess = true;
+            }
+        }
+        return shouldProcess;
     }
 
 
@@ -257,7 +357,7 @@ public class FastObjectLoader {
      */
     private void applySmrEntryToStream(UUID streamId, SMREntry entry,
                                        long globalAddress, boolean isCheckPointEntry) {
-        if (isCheckPointEntry || shouldEntryBeApplied(streamId, entry)) {
+        if (shouldEntryBeApplied(streamId, entry, isCheckPointEntry)) {
 
             // Get the serializer type from the entry
             ISerializer serializer = Serializers.getSerializer(entry.getSerializerType().getType());
@@ -417,7 +517,7 @@ public class FastObjectLoader {
         switch (logData.getType()) {
             case DATA:
                 // Checkpoint should have been processed first
-                if (!isCheckPointEntry(logData)) {
+                if (!isCheckPointEntry(logData) && shouldLogDataBeProcessed(logData)) {
                     updateCorfuObject(logData);
                 }
                 break;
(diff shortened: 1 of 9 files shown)


runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -90,6 +90,28 @@
...
+     * Enable whiteList mode where we only reconstruct
+     * the streams provided through this api.
+     * @param streamsToLoad
+     */
+    public FastObjectLoader addStreamsToLoad(List<UUID> streamsToLoad) {
+        whiteList = true;
+        this.streamsToLoad.addAll(streamsToLoad);
+
+        // Generate the streamsCheckpointId (we need to allow them as well)
+        streamsToLoad.forEach((streamId) ->
```
corfu-repo: MINOR Remove the parentheses around the "streamId" parameter [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1611)

test/src/test/java/org/corfudb/recovery/Helpers.java
```diff
@@ -0,0 +1,107 @@
...
+    static VersionLockedObject getVersionLockedObject(CorfuRuntime cr, String streamName, Class type) {
+        CorfuCompileProxy cp = getCorfuCompileProxy(cr, streamName, type);
+        return cp.getUnderlyingObject();
+    }
+
+    static void assertThatMapIsNotBuilt(CorfuRuntime rt, String streamName, Class type) {
+        ObjectsView.ObjectID mapId = new ObjectsView.
+                ObjectID(CorfuRuntime.getStreamID(streamName), type);
+
+        assertThat(!rt.getObjectsView().getObjectCache().containsKey(mapId));
```
corfu-repo: BLOCKER Complete the assertion. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2970)

corfu-repo:
* ![BLOCKER][BLOCKER] 1 blocker
* ![MINOR][MINOR] 1 minor

[BLOCKER]: https://sonarsource.github.io/sonar-github/severity-blocker.png 'Severity: BLOCKER'

bitmingw:
@rogermichoud We are having inconsistent interface for `addStreamToIgnore` and `addStreamsToLoad`. The ignore one takes table name and does a translate from table name to stream ID internally. I expect the add one to have a similar behavior: takes List<String> and do the transform.

rogermichoud:
@bitmingw Will do the change. Also, there is another issue I have to fix before you can test it.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d079081**.

 [Pull Request #973 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/X8TfRZcHPlQ47SPZ3JYsZQO0Kjl6hsWt)

bitmingw:
@rogermichoud I think it basically works but ERROR logs are still printed out, which is annoying.

The problem can be described below.

```
loadMaps()
  - recoverRuntime()
    - processLogData()
      - updateCorfuObject()  <--- here
```

In `updateCorfuObject()` the first thing is constructing `logEntry` from `logData`. The deserailization will fail for a unknown type. Then the log `Cannot deserialize log entry` and its stack trace is printed out.

I purpose to suppress this ERROR message since it will confuse people.

rogermichoud:
I will fix it this way. There are 2 cases:

1. No stream in the logData (it can contain multiple streams), will just completely ignore the logData.
2. We have a mix of streams (some are in the whitelist, some are not), I will go ahead with the deserialization and log an error if we cannot deserialize it. This means that a client wrote a transaction with a mix of tables we are able to deserialize and some that we are not.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **ef8dad1**.

 [Pull Request #973 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/4yeIAA3pq7rDmcJMne09ESH2ahMx9Nju)

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -90,6 +90,26 @@
...
 
+    private boolean whiteList = false;
+    private List<UUID> streamsToLoad = new ArrayList<>();
+
+    /**
+     * Enable whiteList mode where we only reconstruct
+     * the streams provided through this api.
+     * @param streamsToLoad
+     */
+    public FastObjectLoader addStreamsToLoad(List<String> streamsToLoad) {
```
Maithem: What's the definition of white lists here, only load streams that are in the white list? 
Make sure that they are mutually exclusive (i.e. can't use both lists at the same time).

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -232,20 +254,53 @@ public void updateStreamTails(long address, ILogData logData) {
      * @param entry entry to potentially apply
      * @return if we need to apply the entry.
      */
-    private boolean shouldEntryBeApplied(UUID streamId, SMREntry entry) {
+    private boolean shouldEntryBeApplied(UUID streamId, SMREntry entry, boolean isCheckpointEntry) {
+        // If white list mode, ignore everything that is not in the list or checkpoints
```
Maithem: The comment is incorrect?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -232,20 +254,53 @@ public void updateStreamTails(long address, ILogData logData) {
...
+        if (whiteList && !streamsToLoad.contains(streamId)) {
```
Maithem: This is kind of superfluous, only the second clause is needed. Also, why negate/negate?

If (streamsToLoad.contains(streamId)) return true;

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -232,20 +254,53 @@ public void updateStreamTails(long address, ILogData logData) {
...
+            return false;
+        }
+
         // We ignore the transaction stream ID because it is a raw stream
         // We don't want to create a Map for it.
         if (streamId == runtime.getObjectsView().TRANSACTION_STREAM_ID || streamsToIgnore.contains(streamId)) {
```
Maithem: You can either add this stream to the black list, or not include it in the white list.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -232,20 +254,53 @@ public void updateStreamTails(long address, ILogData logData) {
...
             return false;
         }
 
-        if (!streamsMetaData.containsKey(streamId) || entry.getEntry().getGlobalAddress() >
+        if (isCheckpointEntry || !streamsMetaData.containsKey(streamId) || entry.getEntry().getGlobalAddress() >
```
Maithem: Just use static boolean isCheckPointEntry(ILogData logData) ?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -373,6 +373,14 @@ public static UUID getStreamID(String string) {
         return UUID.nameUUIDFromBytes(string.getBytes());
     }
 
+    public static UUID getCheckpointStreamIdFromId(UUID streamId) {
+        return getStreamID(streamId.toString() + "_cp");
```
Maithem: Probably want to make the string lateral a static definition in streamview

test/src/test/java/org/corfudb/recovery/FastObjectLoaderTest.java
```diff
@@ -813,233 +472,134 @@ public void canReadWithEntriesInterleavingCPS() throws Exception {
...
-        assertThatMapIsBuilt(rt1, rt2, "Map2", map2);
-        assertThatMapIsBuilt(rt1, rt2, "Map3", map3);
-        assertThatObjectCacheIsTheSameSize(rt1, rt2);
-
+        assertThatMapsAreBuilt(rt2);
+        assertThatObjectCacheIsTheSameSize(getDefaultRuntime(), rt2);
     }
 
     @Test
     public void ignoreStreamForRuntimeButNotStreamTails() throws Exception {
```
Maithem: What happens when you white list and ignore a stream?

Maithem:
Also, rebase.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -90,6 +90,26 @@
     @Getter
     private boolean logHasNoCheckPoint = false;
 
+    private boolean whiteList = false;
```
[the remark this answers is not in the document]
rogermichoud: I used it as the knob for the mutual exclusion between the whitelist mode and the blacklist mode. If the client of the API wants to do whitelist but because of its logic he doesn't have any table in the provided streamsToLoad, it should still be a whitelist. 

As is, I could not infer from the streasToLoad if we are in whitelist mode with no streams given or if we are in blacklist mode.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
rogermichoud: Will add in the comments that they are mutually exclusive. The way to enforce it is by using the whiteList flag.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
rogermichoud: is poorly written.  I'll rephrase it. What I wanted to emphasize was that in this list there are both the streams passed by the client and their derived checkpoint streams.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
rogermichoud: The logic is a bit more complicated. The last condition should always be checked, whether we are in whitelist or blacklist. It checks that we are not processing an entry that is already contained in a checkpoint we already processed.

This is why if streamsToLoad.contains(streamId) return true is actually not enough. I had this at the beginning, and it was not passing the tests.

I will re-work this function to make it more self-explanatory. I reckon it is a bit hard to read.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
rogermichoud: I am too deep for that. I don't have access to the logData directly (or I would need to get it backward from the logEntry). Here I am processing an entry inside the CheckpointEntry. Easiest way to know if we are processing a checkpoint or a normal entry is to pass it along the parameters.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
rogermichoud: sound right.

test/src/test/java/org/corfudb/recovery/FastObjectLoaderTest.java
rogermichoud: The expectation is that you should not be doing this. Modes are mutually exclusive, so adding a stream to ignore in the whitelist mode should be a prohibited operation. I added an exception when this happens.

rogermichoud:
@Maithem I addressed the various comments. 

I agree it is debatable if the way I structured my testing is the best one, but I will let it like this for the 2 following reasons:
1. There is a lot of corner cases in these tests, refactoring them in bigger tests is kind of dangerous. I might miss some of these corner cases.
2. The advantage of having multiple small tests is that it is easier to understand what went wrong when some of them are not passing (and it happened many times). When refactoring the code, I saw that only the one related to checkpoints with extra data after them were failing. It was easier to locate where was the bug.

Beyond the style/small changes, I rework the shouldEntryBeApplied workflow to make it easier for the reader. Also added some test to ensure mutual exclusivity of whitelist/blacklist mode.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -90,6 +100,32 @@
...
+
+    /**
+     * Enable whiteList mode where we only reconstruct
+     * the streams provided through this api. In this mode,
+     * we will only process the streams present in streamsToLoad.
+     * All the others are ignored.
+     * @param streamsToLoad
+     */
+    public FastObjectLoader addStreamsToLoad(List<String> streamsToLoad) {
+        if (streamsToIgnore.size() != 0) {
```
corfu-repo: MINOR Use isEmpty() to check whether the collection is empty or not. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1155)

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -221,29 +265,85 @@ public void updateStreamTails(long address, ILogData logData) {
...
+     * the checkpoint. We must avoid to process them twice.
+     *
+     * This comes especially relevant when the operation order affects the end result.
+     * (e.g. clear() operation)
+     *
+     * @param streamId stream id to validate
+     * @param entry entry under scrutinization.
+     * @return if the entry is already part of the checkpoint we started from.
+     */
+    private boolean entryAlreadyContainedInCheckpoint(UUID streamId, SMREntry entry) {
```
corfu-repo: MINOR Remove this use of "SMREntry"; it is deprecated. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3ACallToDeprecatedMethod)

corfu-repo:
* ![MINOR][MINOR] 2 minor

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **b49359b**.

 [Pull Request #973 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ojOnvJP7YQDRSstZAYAj6Sxh0f6AIb5Y)

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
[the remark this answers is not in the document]
Maithem: whiteList is set to true when you add a stream to a white list, so it can be inferred from the list size. Setting it to true even if the white list is empty is another issue.

Also, my initial point was its kind of redundant and it doesn't seem to work. Consider this example:

@Test
    public void whiteListMode() throws Exception {
        populateMaps(MORE, getDefaultRuntime(), CorfuTable.class, true, SOME);

        List<String> streamsToLoad = new ArrayList<>();
        streamsToLoad.add("Map1");
        streamsToLoad.add("Map3");

        CorfuRuntime rt2 = new CorfuRuntime(getDefaultConfigurationString()).connect();
        FastObjectLoader loader = new FastObjectLoader(rt2)
                .setDefaultObjectsType(CorfuTable.class);
        loader.addStreamToIgnore("a");
        loader.addStreamsToLoad(streamsToLoad);
        loader.loadMaps();

        assertThatMapsAreBuiltWhiteList(rt2, streamsToLoad);
    }

rogermichoud:
@Maithem 
The two modes are mutually exclusive. The operation you mentioned is prohibited. These wrong usage of the loader (mixing modes) are tested in blackListAndWhiteListAreMutuallyExclusive.

Better be safe than sorry, I copied verbatim your example and it has the expected behavior, throwing a IllegalStateException: Cannot add a whitelist when there are already streams to ignore

Your concern on robustness and readibility is valid. The whiteMode flag actually reduce code in the addStreasToLoad method. Since an empty list is already initialized, I don't need to add extra logic to check if the list has been initialized (or use lombok lazy initialization that is sometimes a bit opaque).  I will argue that this is more code style than code functionality. Also, the alternative is to have a method isWhiteListModeEnable that infers from the list (and from some other mechanism to see if we passed a empty list) if we are in whiteListMode. I am not sure this would make anything more readable.

Anyway, as long as functionaly works and readibility is okay, I agree we should proceed.

Maithem:
Ok, I see the test failing now. Didn't see the force-push.

PR #976, opened 2017-10-26
Fetch KnownAddressSet from LogUnit

zalokhan:
API to fetch the set of all known addresses from the log unit server.
This will be used to hole fill to equalize the chain before node catchup and segment replication.
Prerequisite for #971 (Catchup, replicate & merge segments)

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -38,6 +38,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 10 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -735,6 +736,41 @@ private LogData  readRecord(SegmentHandle sh, long address)
     }
 
     /**
+     * Creates and returns a new segment handle for a given existing file path for a segmentIndex.
+     *
+     * @param filePath     New segment handle points to this file path.
+     * @param segmentIndex Segment handle reads and writes addresses falling in this segmentIndex.
+     * @return
+     */
+    SegmentHandle createSegmentHandle(String filePath, long segmentIndex) {
```
Maithem: Is this just refactoring?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1041,6 +1052,33 @@ public LogHeader getLogHeader() {
...
+    public Set<Long> getKnownAddressesInRange(long startAddress, long endAddress) {
+
+        long startSegment = startAddress / RECORDS_PER_LOG_FILE;
+        long endSegment = endAddress / RECORDS_PER_LOG_FILE;
+
+        Set<Long> knownAddressSet = new HashSet<>();
+        for (long segmentIndex = startSegment; segmentIndex <= endSegment; segmentIndex++) {
+            String filePath = logDir + File.separator + segmentIndex + ".log";
+
+            // Filters all addresses in range.
```
Maithem: writeChannels is an in-memory map, it can be empty when the logunit restarts, or can be trimmed when a prefix trim happens, which seems to violate the contract.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/AddNodeRequest.java
```diff
@@ -0,0 +1,38 @@
+package org.corfudb.protocols.wireprotocol;
```
Maithem: Move this to the PR that actually use/test it.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **4f167a4**.

 [Pull Request #976 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/kKrrXGeRMquWJ3l11H85TTymphdEQQOQ)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
zalokhan: Yes just refactoring

zalokhan:
@Maithem 
There is a minor difference between this approach and on-the-go inference approach you suggested. In this approach, the replication is targeted to the head of the chain. The head does not service normal reads and is comparatively less congested compared to the tail.
Whereas, in the on-the-go approach, the replication would have to be targeted to the tail to infer the missing entries so as to hole-fill them and then replicate them too.

Maithem:
@zalokhan I think you need to fill holes in the frozen segment in the whole chain, not just per node, before you start the copy.

Maithem:
@zalokhan  another possible solution is to coordinate the hole fills using the management server, then do a direct rsync from source to destination at a segment boundary. If rsync is really much faster than your transfer implementation, but if the hole-filling step is expensive, then it might not be that advantageous.

zalokhan:
Will infer and handle holes during bulk data replication.

PR #983, opened 2017-11-06
Using slf4j interface logger instead of logback implementation

rogermichoud:
This breaks MP dependency. Let's try not to break existing client applications if we can avoid to.

### DIFF
--- a/generator/src/main/resources/logback-test.xml
+++ b/generator/src/main/resources/logback-test.xml
@@ -30,7 +30,7 @@
     <logger name="io.netty.buffer" level="INFO"/>
 
     <!-- Correctness -->
-    <logger name="correctness" level="INFO" additivity="false">
+    <logger name="correctness" level="TRACE" additivity="false">
         <appender-ref ref="CORRECTNESS"/>
     </logger>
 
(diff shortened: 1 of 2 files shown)


bitmingw:
This works as a temporary fix. Need a formal fix because the correctness log shouldn't appear at the MP side.

rogermichoud:
@bitmingw What do you mean? Correctness log is only activated when we use the LoadGenerator. This should not impact any other application.

bitmingw:
@Maithem By using Slf4j we end up losing the control of logging configuration, because configuration is only specified at MP side.

As @zalokhan pointed out, this logging is only for load generator, then we need some mechanism such that other components won't consume it by accident.

rogermichoud:
I put the level of logging in corfuCompile proxy to trace.

Also, Ming will add this in in the MP log4j2.xml. `<logger name=“correctness” level=“OFF”/>`

rogermichoud:
@no2chem 
Sure. This is for the correctness validation layer. I need to know all reads/write at which version they were accessed. For Transactional operation, I can get this version from the TransactionContext, but for non-transaction operations, this is the only place where I can gather it.

When we run the LoadGenerator, it creates a history of all operations so we can validate if there is any inconsistencies. This version is used to detect at which version non-transaction operations were executed.

This is the less intrusive way I found to get this info.

Maithem:
@no2chem  what do you mean, are you talking about slf4j, or the way the versions are logged ?

rogermichoud:
I agree, we should have a more "elegant" way of getting the version of a non-transactional access/write. Let's keep this in the backlog. Probably great for the dev branch !

PR #985, opened 2017-11-08
Fixed Metrics Reporting

Maithem:


Maithem:
@zalokhan is this necessary, or do you have a better approach ?

zalokhan:
@Maithem Just wondering if csv reporting is the right way to go. If the application plans to keep metrics reporting on forever, this will definitely result in space issues right ?

PR #977, opened 2017-10-26
Orchestrator Service

Maithem:
Implementation of a workflow execution service that runs in the
management server.

PR #988, opened 2017-11-09
Add Node workflow

Maithem:
A multi-step workflow that adds a new node to the cluster.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/AddNodeWorkflow.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/AddNodeWorkflow.java
@@ -1,0 +1,109 @@
+package org.corfudb.infrastructure.orchestrator;
+
+import org.corfudb.protocols.wireprotocol.orchestrator.AddNodeRequest;
+import org.corfudb.protocols.wireprotocol.orchestrator.Request;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.view.Layout;
+
+import javax.annotation.Nonnull;
+import java.util.List;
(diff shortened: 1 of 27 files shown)


Maithem:
This PR has two dependencies #970 and #977, will rebase after they go in.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/AddNodeWorkflow.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    class AddNodeToLayout implements Action {
+        @Override
+        public String getName() {
+            return "AddNodeToLayout";
+        }
+
+        @Override
+        public Map<String, Object> execute(@Nonnull CorfuRuntime runtime) throws Exception {
+            Layout currentLayout = runtime.getLayoutView().getLayout();
```
zalokhan: You will need to clone this layout before you pass it to the addNode method.
The add node method tries to seal this layout which will modify it.

Maithem: There's a race between the invalidate layout and clone.

zalokhan: Actually `runtime.getLayoutView().getLayout()` should block on the future and wait for the runtime to return the layout after the `invalidateLayout`.
But you are right, there is some race here, I have seen it too but have not been able to spot the exact reason.

Maithem:
subsumed by https://github.com/CorfuDB/CorfuDB/pull/1014

Issue #991, opened 2017-11-09
Config Robustness


Maithem:
The runtime exposes a couple of setters to change different configurations that different runtime views rely on, therefore the views should only be initialized after the configurations have been set. If they interleave, then some configurations can be ignored. It might be a good idea to force an order.

no2chem:
Probably should use the existing ```CorfuRuntimeParameters``` class, and set the parameters at initialization time by passing ```CorfuRuntimeParameters``` through the constructor, and deprecate all the setters that are exposed in the runtime.

Issue #916, opened 2017-08-17
Metrics is causing possible race condition in CI


no2chem:
This is appearing in CI logs, causing failed builds:

[INFO] --- clojure-maven-plugin:1.8.1:compile (compile) @ cmdlets ---
Compiling org.corfudb.shell to /tmp/classes72337861294533364.dir
01:05:15,665 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback-test.xml]
01:05:15,665 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback.groovy]
01:05:15,665 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found resource [logback.xml] at [jar:file:/home/travis/build/CorfuDB/CorfuDB/runtime/target/runtime-0.1-4385.jar!/logback.xml]
01:05:15,681 |-INFO in ch.qos.logback.core.joran.spi.ConfigurationWatchList@517d4a0d - URL [jar:file:/home/travis/build/CorfuDB/CorfuDB/runtime/target/runtime-0.1-4385.jar!/logback.xml] is not of type file
01:05:15,827 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - debug attribute not set
01:05:15,830 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - About to instantiate appender of type [ch.qos.logback.core.rolling.RollingFileAppender]
01:05:15,840 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - Naming appender as [MetricsRollingFile]
01:05:15,880 |-INFO in ch.qos.logback.core.rolling.FixedWindowRollingPolicy@7862f56 - No compression will be used
01:05:15,886 |-INFO in ch.qos.logback.core.joran.action.NestedComplexPropertyIA - Assuming default type [ch.qos.logback.classic.encoder.PatternLayoutEncoder] for [encoder] property
01:05:15,923 |-INFO in ch.qos.logback.core.rolling.RollingFileAppender[MetricsRollingFile] - Active log file name: /var/log/corfu-metrics.log
01:05:15,923 |-INFO in ch.qos.logback.core.rolling.RollingFileAppender[MetricsRollingFile] - File property is set to [/var/log/corfu-metrics.log]
[...]

PR #992, opened 2017-11-09
AddressSpaceView Initialization Bug

Maithem:
The current initialization order of the AddressSpaceView causes
the cache configuration by the runtime to be ignored. This patch
fixes this order so that the numCacheEntries config is picked up
by the cache.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -279,8 +279,6 @@ public class CorfuRuntime {
         nodeRouters = new ConcurrentHashMap<>();
         retryRate = 5;
 
-        getAddressSpaceView().setMetrics(metrics != null
-                ? metrics : CorfuRuntime.getDefaultMetrics());
         synchronized (metrics) {
             if (metrics.getNames().isEmpty()) {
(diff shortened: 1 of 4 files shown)


Maithem:
Also addresses #916

Maithem:
@zalokhan Its possible that's why I opened this issue https://github.com/CorfuDB/CorfuDB/issues/991
This fix follows our current assumption which is all configurations will happen before any operation on any of the views. In this case, the AddressSpaceView is lazily initialized, so when you call setMetrics then try get the  AddressSpaceView, it will pick up the metrics registry you set. 


@no2chem No, the patch addresses the cache config problem. Currently on master, when you setNumCacheEntries, it won't be picked up by the address space view because of the order of initialization.
The AddressSpaceView was being initialized in the CorfuRuntime() constructor, and as a consequence it always initialized the loading cache with  the default numCacheEntries value. For example,

CorfuRuntime rt = new CorfuRuntime("localhost:9000);  <--- Sets the AddressSpaceView to 5000
rt.setNumCacheEntries(1_000_000); <--- Ignored, AddressSpaceView is already initialized
rt.connect();

or 

CorfuRuntime rt = new CorfuRuntime("localhost:9000) <--- Sets the AddressSpaceView to 5000
.setNumCacheEntries(1_000_000)  <--- Ignored, AddressSpaceView is already initialized
.connect();

PR #994, opened 2017-11-13
Track Max Write Size

Maithem:
Track and print the max write buffer size.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
@@ -6,11 +6,17 @@ import io.netty.handler.codec.MessageToByteEncoder;
 
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.concurrent.atomic.LongAccumulator;
+import java.util.function.LongBinaryOperator;
+
 /**
  * Created by mwei on 10/1/15.
  */
 @Slf4j
 public class NettyCorfuMessageEncoder extends MessageToByteEncoder<CorfuMsg> {
+
+
+    final LongAccumulator maxValue = new LongAccumulator(Math::max, Long.MIN_VALUE);
 
     @Override
     protected void encode(ChannelHandlerContext channelHandlerContext,
@@ -18,8 +24,18 @@ public class NettyCorfuMessageEncoder extends MessageToByteEncoder<CorfuMsg> {
                           ByteBuf byteBuf) throws Exception {
         try {
             corfuMsg.serialize(byteBuf);
+            if(log.isDebugEnabled()) {
+                long prev = maxValue.get();
+                maxValue.accumulate(byteBuf.readableBytes());
+                long curr = maxValue.get();
+                // The max value has been updated.
+                if (prev < curr) {
+                    log.debug("encode: New max write buffer found {}", curr);
+                }
+            }
+
         } catch (Exception e) {
-            log.error("Error during serialization!", e);
+            log.error("encode: Error during serialization!", e);
         }
     }
 }


runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
```diff
@@ -12,12 +12,24 @@
...
+
+    Integer max = 0;
+
     @Override
     protected void encode(ChannelHandlerContext channelHandlerContext,
                           CorfuMsg corfuMsg,
                           ByteBuf byteBuf) throws Exception {
         try {
             corfuMsg.serialize(byteBuf);
+            byteBuf.readableBytes();
```
Maithem: This will impact performance, only sync if log.debug is enabled.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
```diff
@@ -12,12 +12,26 @@
...
                           ByteBuf byteBuf) throws Exception {
         try {
             corfuMsg.serialize(byteBuf);
+            if(log.isDebugEnabled()) {
+                byteBuf.readableBytes();
+                synchronized (max) {
+                    int readableBytes = byteBuf.readableBytes();
+                    if (max < readableBytes) {
+                        max = readableBytes;
+                        log.debug("New max write buffer found {}", max);
```
no2chem: encode:

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
```diff
@@ -12,12 +12,26 @@
...
+                    }
+                }
+            }
+
         } catch (Exception e) {
             log.error("Error during serialization!", e);
```
no2chem: encode:

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
```diff
@@ -12,12 +12,26 @@
 @Slf4j
 public class NettyCorfuMessageEncoder extends MessageToByteEncoder<CorfuMsg> {
 
+
+    Integer max = 0;
```
no2chem: This doesn't seem to be safe for concurrent execution. Try using LongAccumulator.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
```diff
@@ -12,12 +12,26 @@
...
                           CorfuMsg corfuMsg,
                           ByteBuf byteBuf) throws Exception {
         try {
             corfuMsg.serialize(byteBuf);
+            if(log.isDebugEnabled()) {
+                byteBuf.readableBytes();
+                synchronized (max) {
+                    int readableBytes = byteBuf.readableBytes();
+                    if (max < readableBytes) {
+                        max = readableBytes;
```
zalokhan: On this assignment, the synchronized lock will be lost as it is acquired on the Integer object. Right? You have to use a final object for synchrony,

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
```diff
@@ -12,12 +12,26 @@
...
+
     @Override
     protected void encode(ChannelHandlerContext channelHandlerContext,
                           CorfuMsg corfuMsg,
                           ByteBuf byteBuf) throws Exception {
         try {
             corfuMsg.serialize(byteBuf);
+            if(log.isDebugEnabled()) {
+                byteBuf.readableBytes();
+                synchronized (max) {
```
no2chem: Yikes. Object synchronization is going to be overkill. Please use LongAccumulator, or at least AtomicLong, as suggested.

no2chem: Also as @zalokhan points out, using a synchronization monitor lock here is not going to work since the integer (and monitor) will be replaced on assignment...

Maithem: Right, I should have used an AtomicInteger. I'll just replace it with an accumulator .

Maithem:
@zalokhan @no2chem addressed your comments.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageEncoder.java
```diff
@@ -6,20 +6,36 @@
...
+import java.util.function.LongBinaryOperator;
+
 /**
  * Created by mwei on 10/1/15.
  */
 @Slf4j
 public class NettyCorfuMessageEncoder extends MessageToByteEncoder<CorfuMsg> {
 
+
+    final LongAccumulator accumulator = new LongAccumulator(Math::max, Long.MIN_VALUE);
```
no2chem: rename this field to maxValue, as its more descriptive

Maithem: Done.

PR #996, opened 2017-11-13
Change Loggingn Level

Maithem:
Only emit thread termination logs on errors.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
@@ -57,7 +57,7 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter
                             exception.getMessage(),
                             exception);
                 } else {
-                    log.warn("onTermination: Thread terminated (completed normally).");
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
```diff
@@ -57,7 +57,7 @@ protected void onTermination(Throwable exception) {
                             exception.getMessage(),
                             exception);
                 } else {
-                    log.warn("onTermination: Thread terminated (completed normally).");
+                    log.debug("onTermination: Thread terminated (completed normally).");
```
medhavidhawan: what is the reason for changing this log level ? Looks good but do not know the context.

Maithem: It pollutes the server side logs. When you have more client writer threads than the server handler threads, forkjoin will create "temporary" threads, if some of the initial threads block. After the requests finish executing, it will terminate the extra threads and log a message that the thread terminated successfully. We should only log a message if it exceptionally terminates.

medhavidhawan:
ok Looks goog to me.

PR #534, opened 2017-04-10
Cluster uuid

zalokhan:
Added a cluster ID check on the serverRouters.
Every message now with the clientID, requestID and epoch also contains the clusterID.
The clusterID is set only once during the bootstrap of a node. All subsequent `setClusterId()` calls are ignored.

An example of the new layout with the cluster uuid:
```
{
  "clusterId": "d81d3808-fdd2-4f42-82d5-04f73da55845",
  "layoutServers": [
    "localhost:9000"
  ],
  "sequencers": [
    "localhost:9000"
  ],
  "segments": [
    {
      "replicationMode": "CHAIN_REPLICATION",
      "start": 0,
      "end": -1,
      "stripes": [
        {
          "logServers": [
            "localhost:9000"
          ]
        }
      ]
    }
  ],
  "unresponsiveServers": [],
  "epoch": 0
}
```

Will resolve #240

PR #999, opened 2017-11-14
Add cluster Id support

no2chem:
This PR adds a cluster ID field, which should remain consistent
during reconfigurations.

It does not add verification of this field, which is left
for a future PR.


zalokhan:
You can have a look at #534 
Won't have to reimplement everything again. I have rebased the branch and it fails 2 tests which should be easy to fix.

no2chem:
Ok, lets do that then.
Closed in favor of #534.

PR #1002, opened 2017-11-15
Longevity gets a hard stop at the end of run

rogermichoud:
Ensure that even if the application is hung, we still terminate the run.

### DIFF
--- a/generator/src/main/java/org/corfudb/generator/LongevityApp.java
+++ b/generator/src/main/java/org/corfudb/generator/LongevityApp.java
@@ -61,19 +61,27 @@ public class LongevityApp {
     }
 
     /**
-     * Let the workers finish naturally (thanks to the timer) and then kill
+     * Give a chance to the workers to finish naturally (thanks to the timer) and then kill
      * the producer and the checkpointer.
      *
      * At the end of the duration, we give some margin for the workers to finish their task before shutting
      * down the thread pool.
-     * If the application is hung (which can happen when something goes wrong) we will still terminate.
+     *
+     * If the application is hung (which can happen when something goes wrong) we do a hard exit. If any thread
+     * is not playing nice with Interrupted exceptions (which can be a bug as well), this is the only way to
+     * be sure we terminate.
      */
     private void waitForAppToFinish() {
         workers.shutdown();
         try {
-            workers.awaitTermination(durationMs + APPLICATION_TIMEOUT_IN_MS, TimeUnit.MILLISECONDS);
+            boolean finishedInTime = workers.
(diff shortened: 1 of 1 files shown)


generator/src/main/java/org/corfudb/generator/LongevityApp.java
```diff
@@ -61,19 +61,27 @@ public LongevityApp(long durationMs, int numberThreads, String configurationStri
...
     private void waitForAppToFinish() {
         workers.shutdown();
         try {
-            workers.awaitTermination(durationMs + APPLICATION_TIMEOUT_IN_MS, TimeUnit.MILLISECONDS);
+            boolean finishedInTime = workers.
+                    awaitTermination(durationMs + APPLICATION_TIMEOUT_IN_MS, TimeUnit.MILLISECONDS);
+            String livenessState = finishedInTime ? "Success" : "Fail";
+            Correctness.recordOperation("Liveness, " + livenessState, false);
+            if (!finishedInTime) {
+                System.exit(0);
```
no2chem: Shouldn't you return an error code in this case?

rogermichoud: Sure. Fixed.

rogermichoud:
@no2chem Done.

generator/src/main/java/org/corfudb/generator/LongevityApp.java
```diff
@@ -61,19 +61,27 @@ public LongevityApp(long durationMs, int numberThreads, String configurationStri
...
     private void waitForAppToFinish() {
         workers.shutdown();
         try {
-            workers.awaitTermination(durationMs + APPLICATION_TIMEOUT_IN_MS, TimeUnit.MILLISECONDS);
+            boolean finishedInTime = workers.
+                    awaitTermination(durationMs + APPLICATION_TIMEOUT_IN_MS, TimeUnit.MILLISECONDS);
+            String livenessState = finishedInTime ? "Success" : "Fail";
+            Correctness.recordOperation("Liveness, " + livenessState, false);
+            if (!finishedInTime) {
+                System.exit(1);
```
Maithem: Shouldn't this be a negative error code?

rogermichoud: I think it doesn't really matter. I have been looking around for conventions and, as always, there are several of them.

Unix convention says that returning 1 is "generic error"
Somebody on stack overflow says that returning -1 is undefined.
Someone else says that negative numbers are for system error and positive for application errors.

Anyway, using negative code will translate into its 2's complement. Since we are not doing anything with the number and a non-zero number is enough to mark the execution as not successful, I think it is okay if we let it like this and merge the code in.

PR #1004, opened 2017-11-16
Batch Write Implementation

Maithem:
Implemented a StreamLog batch write API that writes a series of
consecutive addresses in one shot.

PR #1006, opened 2017-11-17
Populate LogData Entries

Maithem:
This patch populates the global address for all entry types on
reads.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -20,9 +20,6 @@ import org.corfudb.util.serializer.Serializers;
 @Slf4j
 public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
 
-    public static final LogData EMPTY = new LogData(DataType.EMPTY);
-    public static final LogData HOLE = new LogData(DataType.HOLE);
-    public static final LogData TRIMMED = new LogData(DataType.TRIMMED);
     public static final int NOT_KNOWN = -1;
 
     @Getter
@@ -36,6 +33,24 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     private int lastKnownSize = NOT_KNOWN;
 
     private final transient AtomicReference<Object> payload = new AtomicReference<>();
+
+    public static LogData getTrimmed(long address) {
+        LogData logData = new LogData(DataType.TRIMMED);
+        logData.setGlobalAddress(address);
+        return logData;
+    }
+
+    public static LogData getHole(long address) {
+        LogData logData = new LogData(DataType.HOLE);
+        logData.setGlobalAddress(address);
+        return logData;
+    }
+
+    public static LogData getEmpty(long address) {
+        LogData logData = new LogData(DataType.EMPTY);
+        logData.setGlobalAddress(address);
+        return logData;
+    }
 
     /**
      * Return the payload.
@@ -208,6 +223,17 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
         }
         if (type.isMetadataAware()) {
             ICorfuPayload.serialize(buf, metadataMap);
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        } else if (!(o instanceof LogData)) {
+            return false;
+        } else {
+            return compareTo((LogData) o) == 0;
         }
     }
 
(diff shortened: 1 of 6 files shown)


Maithem:
@no2chem is there a typo in your example, because I don't see it. You are comparing two different slots in the address space, so I expect them to be different (i.e they are not equal). Also, we need to agree on the definition of equality here. Do you define equality solely based on the type of entry ?

Maithem:
@rogermichoud are you referring to the fact that some interfaces pass the address long and the LogData entry, hence redundancy ? Can you point of instances of the workarounds. Maybe they should be addressed in this patch.

no2chem:
@Maithem Well, you added an address field to an empty return type, before, an empty address would equal an empty address no matter where the read was from. 

As in,

```java
LogData.EMPTY.equals(LogData.EMPTY);
```
What are you trying to fix? Where is there code that is querying an empty log data for its address? I'd rather fix that than add additional new-object creation just to pass the address (which the reader should already know!)...

Maithem:
@no2chem perhaps we need to override the equality logic to compare on the address (or all the contents if we are paranoid) because its only based on references right now. For example, some logic might want to expect that LogData.getEmpty(0).equals(LogData.getEmpty(0)) is true.

Maithem:
@no2chem this change was based on my assumption of the modelling, where a LogData type represents a slot in the address space. 

Essentially, my intention is to bind trimmed and hole addresses to a specific address. I did it for those two types and extended it to the empty type for the sake of uniformity.

no2chem:
Why should empty and hole be bound to a specific address? Have you considered checkpoints, which also generate ILogData, as well?

Maithem:
When manipulating LogData collections, it is useful to have LogData entries know their global address, otherwise they have to be wrapped with different structures to include their address. 

Yes, all checkpoint entry types populate the global address metadata correctly, or as expected.

no2chem:
I don't know, this increases the size of serialization as well... for example if I return a map of addresses to logdatas, and I read a range of 0 to 4

I get
0 -> empty(0)
1 -> empty(1)
2 -> empty(2)
3 -> empty(3)

When empty(0) semantically is the same as empty(1).
When are we manipulating a log data collection in a way where we need to know the address of an empty log data?  When is it useful?

Maithem:
@no2chem How is that any different than?
0 -> empty(-1)
1 -> empty(-1)
2 -> empty(-1)
3 -> empty(-1)

Here: https://github.com/CorfuDB/CorfuDB/pull/1004

That's just an artifact of how the address space is modeled, having a single instance of LogData.Empty
on the server side is not an optimization and on the client size we don't deal with many Empty entries anyways, because we fill them up.

no2chem:
@Maithem because empty without an address can be represented as a single bit, but with an address will require 8 bytes. 

More importantly, I'll repeat my question again:

When are we manipulating a log data collection in a way where we need to know the address of an empty log data? When is it useful?

Maithem:
@no2chem We never persist empty entries. So, if you think that 8 extra bytes will incur a significant perf impact in terms of transmission, serialization and memory footprint, then we can settle this by running some perf tests. 

In my last response I linked the PR that does make use of the populated entries, here's the link **again** #1004

no2chem:
@Maithem You just referenced #1004 without specifying why you referenced it.

Either way, this PR seems to be about writes, and you can't write EMPTY, I guess you could write 'hole', but IMO hole should be stored in a bitmap.

Maithem:
@no2chem I referenced because you asked me where this behavior is needed. The added code depends on having uniformity between all LogData types. See, its not hard to code around this and avoid all of this discussion together, but my point is, thats how we modeled the address space and therefore I should be able to use it as such. Having to code around these special cases makes the code messy. If you dont like the model let's change it, but for now this change seems reasonable with almost no impact.

Anyways, Ill fix the equality issue and resubmit this patch, let me know if you have more concerns.

no2chem:
@Maithem Ok, I think this is fine, but if we go this route, we should eliminate the use of maps in ReadResponse and other places, because that information is redundant.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -211,6 +226,17 @@ void doSerializeInternal(ByteBuf buf) {
         }
     }
 
+    @Override
+    public boolean equals(Object o) {
```
corfu-repo: MINOR This class overrides "equals()" and should therefore also override "hashCode()". [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1206)

corfu-repo:
* ![MINOR][MINOR] 1 minor

rogermichoud:
@Maithem What I meant is maybe there is some code out there that is especially cautious about using getGlobalAddress on LogData because it was expecting to get a -1 instead of the real address. I was skimming through the getGlobalAddress usage and I didn't find anything obvious.

I recall that I had to use a workaround in the FastLoader to pass the addresses with the list of LogData because I couldn' t safely get the address from each instance. I think this is too tedious to search for these occurrences and we just going to live with the fact that some code out there was built "working around" this bug. Shouldn't be harmful.

Maithem:
@no2chem this PR is ready

Issue #995, opened 2017-11-13
RPC Timeout


Maithem:
Make this value configurable from the runtime.

PR #1027, opened 2017-11-30
RPC Timeout

Maithem:
Make the RPC timeout configurable via the CorfuRuntime.

Addresses https://github.com/CorfuDB/CorfuDB/issues/995

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -146,6 +146,13 @@ public class CorfuRuntime {
     public int writeRetry = 3;
 
     /**
+     * Set the timeout for rpc requests
+     */
+    @Getter
+    @Setter
+    public long timeoutResponse = 5000;
+
+    /**
      * Sets expireAfterAccess and expireAfterWrite in seconds.
      */
     @Getter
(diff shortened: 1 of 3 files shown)


Maithem:
Hmm, maybe I should address https://github.com/CorfuDB/CorfuDB/issues/991 before checking this in.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -145,6 +145,13 @@
     @Setter
     public int writeRetry = 3;
 
+    /**
+     * Set the timeout for rpc requests
+     */
+    @Getter
+    @Setter
+    public long timeoutResponse = 5000;
```
no2chem: Make this a duration

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -235,14 +235,14 @@ public NettyClientRouter(String host, Integer port, Boolean tls,
     public NettyClientRouter(String host, Integer port, Boolean tls,
                              String keyStore, String ksPasswordFile, String trustStore,
                              String tsPasswordFile, Boolean saslPlainText, String usernameFile,
-                             String passwordFile, MetricRegistry metricRegistry) {
+                             String passwordFile, MetricRegistry metricRegistry, long responseTimeout) {
```
no2chem: Can we avoid adding more fields to pass information (the current amount is kind of crazy).
Make a ConnectionSettings (or similar class) to pass this data....

Maithem: Yea, its pretty bad. Will address #991 first.

no2chem:
Superseded by #1090.

PR #940, opened 2017-09-11
Optimize Read Channels

Maithem:
Use a single channel per segment for uncached reads.

PR #1040, opened 2017-12-05
Memory mapped read/address-space implementation

### LINKED ISSUE #986: Server JVM OutOfMemory on startup with large log files
Sever fails to startup when a log file has size more than 2GB (JVM heap limit)

```
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
	at java.nio.HeapByteBuffer.<init>(HeapByteBuffer.java:57)
	at java.nio.ByteBuffer.allocate(ByteBuffer.java:335)
	at org.corfudb.infrastructure.log.StreamLogFiles.readAddressSpace(StreamLogFiles.java:633)
	at org.corfudb.infrastructure.log.StreamLogFiles.lambda$getSegmentHandleForAddress$2(StreamLogFiles.java:770)
	at org.corfudb.infrastructure.log.StreamLogFiles$$Lambda$43/1899223686.apply(Unknown Source)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
	at org.corfudb.infrastructure.log.StreamLogFiles.getSegmentHandleForAddress(StreamLogFiles.java:751)
	at org.corfudb.infrastructure.log.StreamLogFiles.initializeMaxGlobalAddress(StreamLogFiles.java:245)
	at org.corfudb.infrastructure.log.StreamLogFiles.<init>(StreamLogFiles.java:113)
	at org.corfudb.infrastructure.LogUnitServer.<init>(LogUnitServer.java:124)
	at org.corfudb.infrastructure.CorfuServer.addLogUnit(CorfuServer.java:487)
	at org.corfudb.infrastructure.CorfuServer.main(CorfuServer.java:278)
```

### LINKED ISSUE #1048: BufferUnderflowException on large files
## Overview
 **Bug report.** After writing 10,000 500KB entries into the log, even with -Xmx16G, read requests fail with an error after successful server startup:

```
2017-12-07 03:17:21,221 TRACE [org.corfudb.runtime.clients.NettyClientRouter$2event-0] o.c.r.c.NettyClientRouter - Message routed to BaseClient: CorfuMsg(clientID=e929ab99-53b3-4e18-940d-f45ad371c0ee, requestID=129, epoch=39, buf=PooledSlicedByteBuf(freed), msgType=ERROR_SERVER_EXCEPTION)
 2017-12-07 03:17:21,222 WARN  [org.corfudb.runtime.clients.NettyClientRouter$2event-0] o.c.r.c.BaseClient - Server threw exception for request 129
 java.nio.BufferUnderflowException: null
	at java.nio.HeapByteBuffer.get(HeapByteBuffer.java:151)
```

The logunit remains unresponsive to any requests and throws BufferUnderflowException. Possibly fixed by #1040.

no2chem:
## Overview

Description: This PR replaces the implementation of read and readAddressSpace with versions that used MemoryMappedByteBuffers, instead of allocating and reading into a secondary byte buffer, then constructing the protobuf. A new inputstream (MappedByteBufInputStream) which automatically allocates a buffer and maintains a view over a file channel, is introduced, which allows the log unit to read over a file in a memory-mapped fashion. Verification in StreamLogFiles has been merged into StreamLogFiles::scanLogAndPerformAction. This change unifies code which needs to scan over an entire log file. I've tested the PR on a 6GB log file.

Why should this be merged: Issue #986 reported that a Corfu server is unable to startup with a . large log file (>800MB) due to heap space. Usually this shouldn't be a issue since log files are limited to 10k entries each, but the server does not place a limit to how large each entry is, so this could result in an unstartable Corfu server if large entries were written (for example, checkpoints of ~100KB per entry). This was due to StreamLogFiles::readAddressSpace and StreamLogFiles::getGlobalAdddressTail requiring twice as much memory as the file to read it. It should also significantly reduce memory pressure during startup. Unfortunately, generating a memory mapping for each read is also likely to be potentially costly, #940 may help, but transferring a FileRegion (instead of converting into a protobuf) should be preferred.

Related issue(s) (if applicable): Fixes #1048, #986, related to #940 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -12,7 +12,6 @@ import io.netty.buffer.Unpooled;
 
 import java.io.File;
 import java.io.FileFilter;
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -30,22 +29,22 @@ import java.util.Collection;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicLong;
 
+import java.util.concurrent.atomic.LongAccumulator;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import lombok.extern.slf4j.Slf4j;
 
-import org.apache.commons.io.FileUtils;
 import org.corfudb.format.Types;
 import org.corfudb.format.Types.LogEntry;
+import org.corfudb.format.Types.LogEntryMetadataOnly;
 import org.corfudb.format.Types.LogHeader;
 import org.corfudb.format.Types.Metadata;
 import org.corfudb.format.Types.TrimEntry;
@@ -55,6 +54,7 @@ import org.corfudb.protocols.wireprotocol.IMetadata;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.runtime.exceptions.DataCorruptionException;
 import org.corfudb.runtime.exceptions.OverwriteException;
+import org.corfudb.util.MappedByteBufInputStream;
 
 
 /**
@@ -80,7 +80,7 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
     public final String logDir;
     private final boolean noVerify;
     private final ServerContext serverContext;
-    private final AtomicLong globalTail = new AtomicLong(0L);
+    private final LongAccumulator globalTail = new LongAccumulator(Long::max, -1L);
     private Map<String, SegmentHandle> writeChannels;
     private Set<FileChannel> channelsToSync;
     private MultiReadWriteLock segmentLocks = new MultiReadWriteLock();
@@ -100,15 +100,27 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
             dir.mkdirs();
         }
 
-        writeChannels = new ConcurrentHashMap();
+        writeChannels = new ConcurrentHashMap<>();
         channelsToSync = new HashSet<>();
         this.noVerify = noVerify;
         this.serverContext = serverContext;
-        verifyLogs();
+
         // Starting address initialization should happen before
         // initializing the tail segment (i.e. initializeMaxGlobalAddress)
-        initializeStartingAddress();
-        initializeMaxGlobalAddress();
+        startingAddress = serverContext.getStartingAddress();
+        long tailSegment = serverContext.getTailSegment();
+        long addressInTailSegment = (tailSegment * RECORDS_PER_LOG_FILE) + 1;
+        SegmentHandle sh = getSegmentHandleForAddress(addressInTailSegment);
+        try {
+            scanLogAndPerformAction(sh.fileName, (offset, metadata, le) ->
+                    globalTail.accumulate(le.getGlobalAddress()));
+        } catch (IOException e) {
+            throw new RuntimeException(e.getMessage(), e);
+        } finally {
+            sh.release();
+        }
+
+        lastSegment = tailSegment;
 
         // This can happen if a prefix trim happens on
         // addresses that haven't been written
@@ -201,7 +213,7 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
         // TODO(Maithem) since writing a record and setting the tail segment is not
         // an atomic operation, it is possible to set an incorrect tail segment. In
         // that case we will need to scan more than one segment
-        globalTail.getAndUpdate(maxTail -> address > maxTail ? address : maxTail);
+        globalTail.accumulate(address);
         long segment = address / RECORDS_PER_LOG_FILE;
         if (lastSegment < segment) {
             serverContext.setTailSegment(segment);
@@ -231,85 +243,6 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
             return true;
         }
         return false;
-    }
-
-    private void initializeStartingAddress() {
-        startingAddress = serverContext.getStartingAddress();
-    }
-
-    private void initializeMaxGlobalAddress() {
-        long tailSegment = serverContext.getTailSegment();
-        long addressInTailSegment = (tailSegment * RECORDS_PER_LOG_FILE) + 1;
-        SegmentHandle sh = getSegmentHandleForAddress(addressInTailSegment);
-        try {
-            Collection<LogEntry> segmentEntries = (Collection<LogEntry>)
-                    getCompactedEntries(sh.getFileName(), new HashSet()).getEntries();
-
-            for (LogEntry entry : segmentEntries) {
-                long currentAddress = entry.getGlobalAddress();
-                globalTail.getAndUpdate(maxTail -> currentAddress > maxTail
-                        ? currentAddress : maxTail);
-            }
-        } catch (IOException e) {
-            throw new RuntimeException(e.getMessage(), e);
-        } finally {
-            sh.release();
-        }
-
-        lastSegment = tailSegment;
-    }
-
-    private void verifyLogs() {
-        String[] extension = {"log"};
-        File dir = new File(logDir);
-
-        if (dir.exists()) {
-            Collection<File> files = FileUtils.listFiles(dir, extension, true);
-
-            for (File file : files) {
-                try (FileInputStream fsIn = new FileInputStream(file)) {
-                    FileChannel fc = fsIn.getChannel();
-
-
-                    ByteBuffer metadataBuf = ByteBuffer.allocate(METADATA_SIZE);
-                    fc.read(metadataBuf);
-                    metadataBuf.flip();
-
-                    Metadata metadata = Metadata.parseFrom(metadataBuf.array());
-
-                    ByteBuffer headerBuf = ByteBuffer.allocate(metadata.getLength());
-                    fc.read(headerBuf);
-                    headerBuf.flip();
-
-                    LogHeader header = LogHeader.parseFrom(headerBuf.array());
-
-                    fc.close();
-                    fsIn.close();
-
-                    if (metadata.getChecksum() != getChecksum(header.toByteArray())) {
-                        log.error("Checksum mismatch detected while trying to read "
-                                + "header for logfile {}", file);
-                        throw new DataCorruptionException();
-                    }
-
-                    if (header.getVersion() != VERSION) {
-                        String msg = String.format("Log version {} for {} should match "
-                                + "the logunit log version {}",
-                                header.getVersion(), file.getAbsoluteFile(), VERSION);
-                        throw new RuntimeException(msg);
-                    }
-
-                    if (!noVerify && !header.getVerifyChecksum()) {
-                        String msg = String.format("Log file {} not generated with "
-                                + "checksums, can't verify!", file.getAbsoluteFile());
-                        throw new RuntimeException(msg);
-                    }
-
-                } catch (IOException e) {
-                    throw new RuntimeException(e.getMessage(), e);
-                }
-            }
-        }
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **7bd4da3**.

 [Pull Request #1040 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/ezIgWhIfjP3DFhQBwLla6uMJFdZDovnB)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **8399417**.

 [Pull Request #1040 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/7i3fmvLS53cZqiVlARFVp0yRdAZgnimw)

no2chem:
Unfortunately, this issue is much deeper than previously thought. The CompactedLog class holds a reference to each LogData during the initial scan, which is unnecessary during startup as only the maximum address needs to be collected. Since a reference is held for each LogData, garbage collection fails to reclaim memory. It would be easy enough to just collect data during the scan, but the same code path is used to trim files, so the LogUnit would unexpectedly crash during trim in this case.

Looking at a potential solution which involves generating the compacted file in place.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -42,6 +49,7 @@
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicLong;
```
corfu-repo: MINOR Remove this unused import 'java.util.concurrent.atomic.AtomicLong'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -506,64 +510,76 @@ private void trimLogFile(String filePath, Set<Long> pendingTrim) throws IOExcept
...
 
+    /** Scan all entries, updating the LogUnit structures and verifying the log file integrity.
+     *
+     * @param filePath  The log file (segment) to scan
+     */
+    private void scanAndVerifyEntries(String filePath) throws IOException {
+        FileChannel fc = getChannel(filePath, true);
+        try (MappedByteBufInputStream stream = new MappedByteBufInputStream(fc)) {
+            Metadata headerMetadata = Metadata.parseFrom(stream.limited(METADATA_SIZE));
+            LogHeader header = LogHeader.parseFrom(stream.limited(headerMetadata.getLength()));
```
corfu-repo: MINOR Remove this unused "header" local variable. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1481)

corfu-repo:
* ![MINOR][MINOR] 2 minor

zalokhan:
I tested a restart of a server with a 1.8G log file.
Without this patch the server crashes on startup. 
Now the restart is successful.
I will now try to test with a file >2GB and post my findings here.

no2chem:
@zalokhan Thanks! I re-implemented the trim function so it completes in a single pass without buffering as well, so that shouldn't crash the server either.

Unfortunately, it seems that we will run into another problem: The JVM only supports mapping 2GB at a time, currently, so I suspect we will crash if we have log files that are >2GB of physical space.

Maithem:
You can map more than 2GB, but I dont think this problem should be solved this way. File sizes should be smaller, from my tests smaller segments exhibit better performance. The file sizes can be controlled by parameters like (rpc wait time, batch writes). Another way to solve this problem is to allocate an array once and read/process multiple chunks at a time. In addition to the mentioned work around, you can just increase the -Xmx value and you wouldn't see this issue. In other words, I don't think this change is necessary now. Also, @zalokhan  only observed this problem when he was deliberately writing large files. 

The reason I don't think this PR is necessary right now is because 1) there are easy workarounds 2) we haven't seen this in production 3) we want to introduce more memory mapped buffers in the logging unit

Because of  #3 I think we should have another implementation that heavily uses memory mapped files (i.e. a different StreamLog implementation) and this change would make sense in the new implementation.

no2chem:
I disagree. Currently, file sizes cannot be controlled, because log entry sizes are not controlled and just because "you don't run into the issue now", doesn't mean that it won't happen in the future. Changing the segment size doesn't help much, because you can still have some unpredictably large writes. 

The JVM doesn't natively support mapping more than 2GB. There are unsafe hacks, or you can map the same file multiple times, but a >2GB mapping is unsupported. This issue causes a Corfu instance to crash in an unrecoverable way. If you think that is a non-issue/non-bug, then perhaps you should comment in the issue (#986).

Maithem:
My remarks with respect to segment size is regarding the size of the file not number of entries inside a file.
We have ways of mitigating this risk, the underlying problem is we don't control the maximum write size for a particular address, we talked about before but haven't concluded anything, maybe you want to fix that first. Allowing arbitrary parameters and trying to "fix it" with the method you propose is an incomplete solution and currently not necessary. If you are solving this for the general case, then this PR doesn't work (For example it work work for files larger than 2gb as you pointed out). If you are trying to solve the problem for the 2Gb case, then this seems arbitrary and we need to talk more about write sizes.

My argument is not based on the fact that we don't see it today and therefore we shouldn't worry about it. My point is we have controls to mitigate this issue (and we have used these controls in production to fix this very problem before).

no2chem:
Well, this PR is incomplete (if you notice, it is in the In Progress pipeline), and extending the PR to solve the solution for 2GB is not difficult. In fact, I'm almost done with that. Obviously, I didn't realize the Java 2GB solution until later.

If your statement is that you don't think that this is a bug, then we can discuss that in the issue. This appears to be what you are indicating.

If your statement is that we shouldn't use memory mapped files to solve this bug, then we can discuss this here. I don't believe we should have a "memory mapped" and non-"memory mapped" implementation. How do we tell users which one to pick?

Maithem:
Incomplete in terms of generality of the solution, not the status of this particular PR. 

I don't know if a bug is the right classification for this issue, since it is contingent on limits of a single write which are not well defined. I think we should talk about the write size limits, but if you want to ignore that for now, which is fine, then I suggest you do chunking instead of memory mapping, because chunking solves the general problem, whereas mapping solves part of the problem. 

Unless we have a complete implementation of StreamLog files that uses mapped files, we can't tell which one is better under different scenarios. I think creating a new StreamLog based on StreamLogFiles and morphing it over time is a better solution, because then you can cleanly compare the performance characteristics of two different implementations. Obviously, there is no silver bullet, so we might end up recommending different implementations for different workloads.

no2chem:
>Incomplete in terms of generality of the solution, not the status of this particular PR.
Once the PR supports >2GB, I'm not sure what you mean by "lack of generality"

>I don't know if a bug is the right classification for this issue, since it is contingent on limits of a single write which are not well defined. 
The server starts up, encounters a log file of 1.8GB, and crashes. How is that not a bug?

Maithem:
If you will support >2gb, then I think this is fine.

"The server starts up, encounters a log file of 1.8GB, and crashes. How is that not a bug?", it can also be seen as misconfiguration (i.e. you need more memory), not necessarily a bug in the code. That's why I am saying its not clear. Ultimately, segments sizes are a function of single address writes, which is undefined. Since its not defined, its hard to claim this as a bug.

no2chem:
> it can also be seen as misconfiguration (i.e. you need more memory), not necessarily a bug in the code

Not really. It should tell the user they need more memory so its still a bug. 
Besides, needing as much memory as the size of the file just to start up is a bug. Reading a file just to verify it should not requiring reading it into memory. 
Worse yet, there's no way to recover from such a big file even if you accidentally caused the condition, since trim will fail as well. Even if we had write size controls, we still need to be able to recover from large files. And 1.8 GB is not a particularly large file. It's not even 2GB.

Maithem:
> Not really. It should tell the user they need more memory (rather than crash with a heap allocation error) so its still a bug.

Its nice to print "You need more memory", many other systems dont, but if you think that's the fix then you can do it.

> Besides, needing as much memory as the size of the file just to start up is a bug. Reading a file just to verify it should not requiring reading it all into memory.

That's your opinion, which I think is an unreasonable generalization. Also, the readaddress space is not just "verification", it builds an index.

> Worse yet, there's no way to recover from such a big file even if you accidentally caused the condition, since trim will fail as well.

False. as i mentioned its not a simple as you put it, like i mentioned before you can bump up the -xmx parameter. Also, if you fail after restart, trim doesn't require segments to be open for the trim to happen as shown below:

`File dir = new File(logDir);
        FileFilter fileFilter = new FileFilter() {
            public boolean accept(File file) {
                try {
                    String segmentStr = file.getName().split("\\.")[0];
                    return Long.parseLong(segmentStr) < endSegment;
                } catch (Exception e) {
                    log.warn("trimPrefix: ignoring file {}", file.getName());
                    return false;
                }
            }`

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **b1f60ac**.

 [Pull Request #1040 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/KQoB2KV5vnnBU3DvgwNYXTJscAMy78eE)

no2chem:
>That's your opinion, which I think is an unreasonable generalization. Also, the readaddress space is not just "verification", it builds an index.

The same applies to indexes as well. It's not an unreasonable generalization, I think that if SQL server crashed if there was a log file larger than memory, it would be considered a bug.

> False. as i mentioned its not a simple as you put it, like i mentioned before you can bump up the -xmx parameter.

Okay, that'll work for a file that fits in memory, perhaps, but what about one which does not? You only can bump Xmx up so far. The comment about trim is irrelevant if the server crashes on startup, right?

This is a issue where there is a reasonable solution. I could understand if there were technical reasons why you don't think this is a good solution, but you seem to be against solving it altogether. Why is that?

Maithem:
> The comment about trim is irrelevant if the server crashes on startup
It is irrelevant, I was just correcting a fact that you made in the previous post. 

> This is a issue where there is a reasonable solution. I could understand if there were technical reasons why you don't think this is a good solution, but you seem to be against solving it altogether. Why is that?

No I think it is a problem that needs to be addressed. My previous posts were to make a point that this problem is a consequence of another pending problem and solving that problem first can simplify the solution to this problem.

Once you complete this PR, then we can objectively evaluate it against other solutions and see if it is the best one.

no2chem:
> It is irrelevant, I was just correcting a fact that you made in the previous post.

You mean that trim will fail? The server can't startup, so you can't trim. What are you correcting? Also, even if the server manages to startup, memory allocation failures will start happening if you try to trim, won't they? Catching the exception at the trim doesn't help.

Maithem:
> Also, even if the server manages to startup, memory allocation failures will start happening if you try to trim, won't they?

No, trimPrefix doesn't open segments because it doesn't invoke readAddressSpace, so the trim can succeed. 

Also, If other requests result in a operation that invokes readAddressSpace, then the allocation will fail on that request's thread, but I don't think that it bricks the server such that a subsequent trims are blocked since they will be serviced on a different thread.

no2chem:
@Maithem the problem is not just readAddressSpace, but also ```getCompactedEntries``` , which is called by ```trimLogFile```. If ```trimLogFile``` is called on a file which has a large untrimmed portion, then the server will run out of memory because ```CompactedEntry``` buffers every entry before writing it out. Even if the server doesn't run out of memory, it seems that keeping the entries in a map generates memory pressure during compaction which can be avoided.

Maithem:
@no2chem Correct, but that's only for sparse trimming (not used), not prefix trim. Also, If you are optimizing for memory pressure, then also consider not reading the whole file to build the index, just put the index at the top of the file. For example, {header | index | records}, that way you can read the index directly without reading all the records. This will address both problems, that is, memory pressure and large segments. Just consider it, think about it, it might make sense to do that.

no2chem:
I'm just trying to fix the bug in issue #986. I've reproduced it locally, it seems that you need about 2x the heap space of the file at the last segment. So with a 900MB file (produced using 100KB raw entries), you need 2GB of heap space.

no2chem:
Now updated to support files >2GB. I've tested with a 6GB log file, and though startup takes a few seconds (a potential improvement may be to scan in a multi-threaded fashion), the server no longer crashes and I'm no longer seeing the bufferunderflow errors in #1048 either.

I doubt this PR will have that much of an effect on performance.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **70e9ce8**.

 [Pull Request #1040 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/92rpuWDRpZQs48JaI1S6q2gz1lZgeGuQ)

Issue #1044, opened 2017-12-06
CorfuRuntime does not update router epoch when NetworkException


zalokhan:
## Overview
The CorfuRuntime does not update the client router's epoch if there is a Network Exception while setting the epoch from the server list in fetchLayout.

- ( ) **Bug report.** Testing add node cmdlet.

PR #982, opened 2017-11-03
Prevent NetworkException from causing a double-write

rogermichoud:
The Runtime has a settable timeout for how long it will wait for
an unavailable system. If we are not able to progress (due to NetworkExcpetion)
during the timeout, the handler provided by the client (or default handler,
that shutdown the Runtime) will be triggered.

Fixes #974

PR #1046, opened 2017-12-06
Fix set router epoch on runtime invalidate.

zalokhan:
## Overview

Description: CorfuRuntime: fetchLayout fetches the latest layout from a random layout server and sets the client router epochs to the newer, higher epoch number in a loop.
However, if the first router in the loop throws a network exception the loop is aborted and the other router epochs remain unset.

Why should this be merged: To avoid wrong epoch exceptions and set the router epochs correctly. 

Related issue(s) (if applicable): #1044 

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/runtime/CorfuRuntimeTest.java
+++ b/test/src/test/java/org/corfudb/runtime/CorfuRuntimeTest.java
@@ -1,20 +1,33 @@
 package org.corfudb.runtime;
 
 import org.corfudb.infrastructure.TestLayoutBuilder;
+
+import org.corfudb.runtime.clients.BaseClient;
+import org.corfudb.runtime.clients.LayoutClient;
 import org.corfudb.runtime.clients.LogUnitClient;
+import org.corfudb.runtime.clients.ManagementClient;
+import org.corfudb.runtime.clients.SequencerClient;
+import org.corfudb.runtime.clients.TestClientRouter;
 import org.corfudb.runtime.clients.TestRule;
 import org.corfudb.runtime.exceptions.unrecoverable.SystemUnavailableError;
+
+import org.corfudb.infrastructure.TestServerRouter;
+import org.corfudb.runtime.exceptions.NetworkException;
 import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.view.AbstractViewTest;
 import org.corfudb.runtime.view.Layout;
 import org.corfudb.runtime.view.stream.IStreamView;
 import org.corfudb.util.CFUtils;
+import org.junit.Before;
 import org.junit.Test;
 
 import java.time.Duration;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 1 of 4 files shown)


corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **d1fc0f5**.

 [Pull Request #1046 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/1GwIiS1vSUnWFujzkXQyZGv4HkysIJhd)

zalokhan:
@rogermichoud Added a test. 
Verification: The test fails on master.

test/src/test/java/org/corfudb/runtime/CorfuRuntimeTest.java
```diff
@@ -169,8 +195,70 @@ public void doesNotAllowReadsAfterSealAndBeforeNewLayout() throws Exception {
...
+     * @throws Exception
+     */
+    @Test
+    public void testFailedRouterInFetchLayout() throws Exception {
+
+        final Map<String, TestClientRouter> routerMap = new ConcurrentHashMap<>();
+        final AtomicReference<String> failedNode = new AtomicReference<>();
+
+        CorfuRuntime.overrideGetRouterFunction = (corfuRuntime, endpoint) -> {
+            if (failedNode.get() != null && endpoint.equals(failedNode.get())) {
```
no2chem: Can you just move this logic to AbstractViewTest::getRouterFunction?

test/src/test/java/org/corfudb/runtime/view/AbstractViewTest.java
```diff
@@ -350,6 +350,16 @@ public String getEndpoint(int port) {
         return "test:" + port;
     }
 
+    /**
```
no2chem: If you move the logic as suggested in the comment above, this function is no longer necessary.

test/src/test/java/org/corfudb/runtime/CorfuRuntimeTest.java
no2chem: Ah, I see what is happening here. Is there a reason to throw these exceptions (it seems like they aren't picked up in your test?)

no2chem: Never mind, I see the logic. It's an unfortunate side effect of the test router not being able to simulate a closed connection. Add that to the comments (that this is simulating a failed node), and I think we'll be okay.

zalokhan: Firstly the test routers never throw Network Exceptions and I needed to test them specifically.
These exceptions are caught by the CorfuRuntime while updating the router epochs.

This is very specific to this test and I wonder if anyone else would ever want to throw this exception at this point in the code.

rogermichoud: I introduced a "simulated network exception" at the test router level. Can you not leverage this? It is way less fancy than that, so not sure it serves your purpose. That is what this method does, throwing a network exception from the router.

The method is simulateDisconnectedEndpoint(), in TestClientRouter.

zalokhan: Sounds good I'll add a dependency on #982 and consume your test addition.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **0ec14b0**.

 [Pull Request #1046 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/h2stK4gS0KbA8jVXxGU2T6xDbXmo0sxo)

zalokhan:
@rogermichoud @no2chem 
Sorry for realizing this so late.
I am not able to reproduce the error using the new `simulateEndpointDisconnect` feature in the TestClientRouter.
This is because the NettyClientRouter unlike the TestClientRouter calls the `start()` method in the constructor. This at times can throw a NetworkException if the connection cannot be made.
Simulating this will require a different approach.

So for now, I'll stick to my existing test and make a few changes which @no2chem suggested in the review.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **c51489f**.

 [Pull Request #1046 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/E60Hza0lil80uR2hvqNVZfMi71AP6cWX)

no2chem:
@zalokhan did you update this? I think it needs a comment describing what you're doing in CorfuRuntimeTest (simulating failed connections).

zalokhan:
@no2chem Added the comment.

PR #1059, opened 2017-12-11
Implement Node Id on Server

### LINKED ISSUE #1050: Generate a unique ID for each node
## Overview
**Feature request.**  Currently, the layout uses IP addresses to identify nodes, which can be problematic if a node is restarted, since the node may no longer contain the same data. We should generate a unique ID for each node and advertise it. Ideally, this ID should be human readable because it will go in the layout.

## Deliverables
- [ ] Generate a unique node ID for each node

no2chem:
## Overview

Description: This PR introduces a node Id field, which is generated on first start and stored in the data store. The node Id is a 128-bit UUID which is represented as a Base64 URL-safe string. The planned use of a base64 string is to append it to a server url, in the form ```<host>:<port>:<node-id>```, which will enable the node id to be checked at connection time during a handshake to be implemented in the future. Currently, the Node Id can be retrieved in using a versionRequest call to a server, and the server can retrieve the node Id from a serverContext.

Why should this be merged: This PR makes it possible to disambiguate servers which have failed, but restarted at the same IP address, eliminating configuration errors from allowing clients to connect to incorrect servers.

Related issue(s) (if applicable): Fixes #1050 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
@@ -271,9 +271,6 @@ public class CorfuServer {
         // Create a common Server Context for all servers to access.
         serverContext = new ServerContext(opts, router);
 
-        // Generate a node ID if necessary.
-        generateNodeId(serverContext);
-
         // Add each role to the router.
         router.addServer(new BaseServer(serverContext));
         addSequencer();
@@ -480,22 +477,5 @@ public class CorfuServer {
     public static void addManagementServer() {
         managementServer = new ManagementServer(serverContext);
         router.addServer(managementServer);
-    }
-
-    /** Generate a Node Id if not present.
-     *
-     * @param context   The server context to use.
-     */
-    private static void generateNodeId(@Nonnull ServerContext context) {
-        String currentId = context.getDataStore().get(String.class, "",
-                ServerContext.NODE_ID);
-        if (currentId == null) {
-            String idString = UuidUtils.asBase64(UUID.randomUUID());
-            log.info("No Node Id, setting to new Id={}", idString);
-            context.getDataStore().put(String.class, "", ServerContext.NODE_ID, idString);
-
-        } else {
-            log.info("Node Id = {}", currentId);
-        }
     }
 }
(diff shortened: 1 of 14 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -275,12 +271,15 @@ public static void main(String[] args) {
         // Create a common Server Context for all servers to access.
         serverContext = new ServerContext(opts, router);
 
+        // Generate a node ID if necessary.
+        generateNodeId(serverContext);
```
zalokhan: Should this be generated here?
Or can we move it down in the ServerContext creation.
This way we will have a similar behaviour between our tests and real servers.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -480,4 +481,21 @@ public static void addManagementServer() {
         managementServer = new ManagementServer(serverContext);
         router.addServer(managementServer);
     }
+
+    /** Generate a Node Id if not present.
+     *
+     * @param context   The server context to use.
+     */
+    private static void generateNodeId(@Nonnull ServerContext context) {
+        String currentId = context.getDataStore().get(String.class, "",
```
zalokhan: Can we move the setter and getter of this NODE_ID to the server context.
Every component will try to consume it.

no2chem: I think this is not needed now that there is a helper function.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
no2chem: Ok - moved it to server context. Going to wait for #1061 to make meaningful tests.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **4ae0bbe**.

 [Pull Request #1059 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/1nxikniF253rXLMtv8ciPEFdoVpv4lpm)

Issue #1077, opened 2017-12-14
Handshake


annym:
## Overview

**Feature.**  Upon client connection we need to initiate a handshake between client and server node, which makes validations based on the node id. Once handshake is completed, data can be sent between client-server.

- (x) **Feature request.**

## Deliverables

- Implementation of a handshake mechanism between client and server.

Maithem:
PR merged.

PR #1078, opened 2017-12-14
Handshaking

annym:
## Overview

Description:

An implementation of Corfu client-server handshaking. Overall:
- Upon connection, the client will initiate a handshake.
- Both (server and client) will start a handshake timeout timer.
- The client sends a message that contains its own id and the (asserted) server's id.
- The server validates the handshake message and replies with its own server id
  and corfu version.
- The client makes its own validation and in case of is success,
  message exchange is enabled.
- If any validation fails client/server will close the connection.

Why should this be merged:  This is a Feature Request. The fact that a node is no longer tied to an IP and instead is uniquely identified by a 16Byte Identifier raises the need of actually validating upon connection that we are connected to the right node. 

Related issue(s) (if applicable): #1077 

- [x] Public API has Javadoc

### DIFF
--- a/generator/src/main/java/org/corfudb/generator/LongevityApp.java
+++ b/generator/src/main/java/org/corfudb/generator/LongevityApp.java
@@ -125,8 +125,6 @@ public class LongevityApp {
 
             exitStatus = checkpointHasFinished ? 0 : 1;
             System.exit(exitStatus);
-
-
         }
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
@@ -1,0 +1,254 @@
+package org.corfudb.infrastructure;
+
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+import io.netty.handler.timeout.ReadTimeoutException;
+import io.netty.handler.timeout.ReadTimeoutHandler;
+import io.netty.util.AttributeKey;
+
+import java.util.ArrayDeque;
+import java.util.Queue;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.corfudb.protocols.wireprotocol.CorfuMsg;
+import org.corfudb.protocols.wireprotocol.CorfuMsgType;
+import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
+import org.corfudb.protocols.wireprotocol.HandshakeMsg;
+import org.corfudb.protocols.wireprotocol.HandshakeResponse;
+
+/**
+ * The ServerHandshakeHandler waits for the handshake message, validates and sends
+ * a response to the client. This reply contains its node id and current version of Corfu.
+ *
+ * Created by amartinezman on 12/11/17.
+ */
+@Slf4j
+public class ServerHandshakeHandler extends ChannelDuplexHandler {
+
+    private final UUID nodeId;
+    private final String corfuVersion;
+    private final AtomicBoolean handshakeComplete;
+    private final AtomicBoolean handshakeFailed;
+    private final int timeoutInSeconds;
+    private final Queue<CorfuMsg> messages = new ArrayDeque();
+    private static final  AttributeKey<UUID> clientIdAttrKey = AttributeKey.valueOf("ClientID");
+    private static final String READ_TIMEOUT_HANDLER = "readTimeoutHandler";
+
+
+    /**
+     * Creates a new ServerHandshakeHandler which will handle the handshake--initiated by a client
+     * on the server side.
+     *
+     * @param nodeId Current Server Node Identifier.
+     * @param corfuVersion Version of Corfu in Server Node.
+     */
+    public ServerHandshakeHandler(UUID nodeId, String corfuVersion, int timeoutInSeconds) {
+        this.nodeId = nodeId;
+        this.corfuVersion = corfuVersion;
+        this.timeoutInSeconds = timeoutInSeconds;
+        this.handshakeComplete = new AtomicBoolean(false);
+        this.handshakeFailed = new AtomicBoolean(false);
+    }
+
+    /**
+     * Read data from the Channel.
+     *
+     * @param ctx channel handler context
+     * @param m object received in inbound buffer
+     * @throws Exception
+     */
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
+
+        if (this.handshakeFailed.get()) {
+            return;
+        }
+
+        if (this.handshakeComplete.get()) {
+            // If handshake completed successfully, but still a message came through this handler,
+            // send on to the next handler in order to avoid message loss.
+            super.channelRead(ctx, m);
+            return;
+        }
+
+        CorfuPayloadMsg<HandshakeMsg> handshake;
+
+        try {
+            handshake = (CorfuPayloadMsg<HandshakeMsg>) m;
+            log.debug("channelRead: Handshake Message received. Removing " + READ_TIMEOUT_HANDLER +
+                    " from pipeline.");
+            ctx.pipeline().remove(READ_TIMEOUT_HANDLER);
+        } catch (ClassCastException e) {
+            log.warn("channelRead: Non-handshake message received by handshake handler." +
+                    " Send upstream only if handshake succeeded.");
+            if (this.handshakeComplete.get()) {
+                // Only send upstream if handshake is complete.
+                super.channelRead(ctx, m);
+            } else {
+                // Otherwise, drop message.
+                try {
+                    log.debug("channelRead: Dropping message: " +
+                            ((CorfuMsg) m).getMsgType().name());
+                } catch (Exception ex) {
+                    log.error("channelRead: Message received by Server is not a valid " +
+                            "CorfuMsg type.");
+                }
+            }
+            return;
+        }
+
+        UUID clientId = handshake.getPayload().getClientId();
+        UUID serverId = handshake.getPayload().getServerId();
+
+        // Validate handshake, but first verify if node identifier is set to default (all 0's)
+        // which indicates node id matching is not required.
+        if (serverId.equals(UUID.fromString("00000000-0000-0000-0000-000000000000"))) {
+            log.info("channelRead: node id matching is not requested by client.");
+        } else {
+            if (!serverId.equals(this.nodeId)) {
+                log.error("channelRead: Invalid handshake: this is " + this.nodeId +
+                        " and client is trying to connect to " + serverId);
+                this.fireHandshakeFailed(ctx);
+                return;
+            }
+        }
+
+        // Store clientID as a channel attribute.
+        ctx.channel().attr(clientIdAttrKey).set(clientId);
+        log.info("channelRead: Handshake validated by Server.");
+        log.debug("channelRead: Sending handshake response: Node Id: " + this.nodeId +
+                " Corfu Version: " + this.corfuVersion);
+
+        CorfuMsg handshakeResponse = CorfuMsgType.HANDSHAKE_RESPONSE
+                .payloadMsg(new HandshakeResponse(this.nodeId, this.corfuVersion));
+        ctx.writeAndFlush(handshakeResponse);
+
+        // Flush messages in queue
+        log.debug("channelRead: There are {" + this.messages.size() + "} messages in queue to " +
+                "be flushed.");
+        for (CorfuMsg message : this.messages) {
+            ctx.writeAndFlush(message);
+        }
+
+        // Remove this handler from the pipeline; handshake is completed.
+        log.info("channelRead: Removing handshake handler from pipeline.");
+        ctx.pipeline().remove(this);
+        this.fireHandshakeSucceeded();
+    }
+
+    /**
+     * Channel event that is triggered when a new connected channel is created.
+     *
+     * @param ctx channel handler context
+     * @throws Exception
+     */
+    @Override
+    public void channelActive(final ChannelHandlerContext ctx) throws Exception {
+        log.info("channelActive: Incoming connection established from: " +
+                ctx.channel().remoteAddress() + " Start Read Timeout.");
+        ctx.pipeline().addBefore(ctx.name(), READ_TIMEOUT_HANDLER,
+                new ReadTimeoutHandler(this.timeoutInSeconds));
+    }
+
+    /**
+     * Channel event that is triggered when the channel is closed.
+     *
+     * @param ctx channel handler context
+     * @throws Exception
+     */
+    @Override
+    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+        log.debug("channelInactive: Channel closed.");
+        if (!this.handshakeComplete.get()) {
+            this.fireHandshakeFailed(ctx);
+        }
+    }
+
+    /**
+     * Channel event that is triggered when an exception is caught.
+     *
+     * @param ctx channel handler context
+     * @param cause exception cause
+     * @throws Exception
+     */
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
+            throws Exception {
+        log.error("exceptionCaught: Exception caught: " + cause.getMessage());
+
+        if (cause instanceof ReadTimeoutException) {
+            // Read timeout: no inbound traffic detected in a period of time.
+            if (handshakeFailed.get()) {
+                log.debug("exceptionCaught: Handshake timeout checker: already failed.");
+                return;
+            }
+
+            if (!handshakeComplete.get()) {
+                log.error("exceptionCaught: Handshake timeout checker: timed out. Close Connection.");
+                handshakeFailed.set(true);
+                ctx.channel().close();
+            } else {
+                log.debug("exceptionCaught: Handshake timeout " +
+                        "checker: discarded (handshake OK)");
+            }
+        } else {
+            super.exceptionCaught(ctx, cause);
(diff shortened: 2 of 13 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+
+
+    @Override
+    public void channelActive(final ChannelHandlerContext ctx) throws Exception {
+        this.group.add(ctx.channel());
+        log.info("Incoming connection established from: " +
+                ctx.channel().remoteAddress());
+
+        // Start timeout checker. Wait for handshake to complete, otherwise close connection.
+        new Thread() {
```
no2chem: Hm. I think we want to avoid creating a new thread here. I think we have two options here, we can schedule on the executor for the context (```ctx.executor()```), or we can use the ```ReadTimeoutHandler``` and uninstall it from the pipeline as well (I think you can uninstall by name). I'm sort of leaning towards the ReadTimeoutHandler.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+
+            @Override
+            public void run() {
+                try {
+                    log.debug("Wait for the handshake to complete, then disconnect");
+                    latch.await(timeoutInMillis, TimeUnit.MILLISECONDS);
+                } catch (InterruptedException e) {
+                    log.error("Handshake timeout checker: interrupted!");
+                    e.printStackTrace();
```
no2chem: Should re-throw the exception :).

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+ */
+@Slf4j
+@ChannelHandler.Sharable
+public class ServerHandshakeHandler extends ChannelDuplexHandler {
+
+    private final long timeoutInMillis;
+    private final UUID nodeId;
+    private final String corfuVersion;
+    private final ChannelGroup group;
+    private final AtomicBoolean handshakeComplete;
```
no2chem: volatile boolean should be sufficient, right?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+import org.corfudb.protocols.wireprotocol.HandshakeResponse;
+
+/**
+ * The ServerHandshakeHandler waits for the handshake message, validates and sends
+ * a response to the client. This reply contains its node id and current version of Corfu.
+ *
+ * Created by amartinezman on 12/11/17.
+ */
+@Slf4j
+@ChannelHandler.Sharable
```
no2chem: Is this really Sharable? handshakeComplete looks like state.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -0,0 +1,223 @@
...
+
+        // Write the handshake & add a timeout listener.
+        CorfuMsg handshake = CorfuMsgType.HANDSHAKE_INITIATE
+                .payloadMsg(new HandshakeMsg(this.clientId, this.nodeId));
+
+        log.info("Initiate handshake. Send handshake message.");
+        ChannelFuture f = ctx.writeAndFlush(handshake);
+        f.addListener(new ChannelFutureListener() {
+            @Override
+            public void operationComplete(ChannelFuture future)
```
no2chem: Same re: ReadTimeoutHandler on the server side.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -422,7 +422,7 @@ public static UUID getCheckpointStreamIdFromName(String streamName) {
     }
 
     /**
-     * Get corfy runtime version.
+     * Get corfu runtime version.
```
no2chem: Heh, good catch

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
annym: Totally agree, I thought I had removed it. It is not sharable.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+                super.channelRead(ctx, m);
+                return;
+            }
+
+            CorfuPayloadMsg<HandshakeMsg> handshake;
+
+            try {
+                handshake = (CorfuPayloadMsg<HandshakeMsg>) m;
+            } catch (ClassCastException e) {
+                log.warn("Non-handshake message received by handshake handler. Send upstream only " +
```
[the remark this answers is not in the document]
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+            // Validate handshake
+            if (!serverId.equals(this.nodeId)) {
+                log.error("Invalid handshake: this is " + this.nodeId +
+                        " and client is trying to connect to " + serverId);
+                this.fireHandshakeFailed(ctx);
+                return;
+            }
+
+            // Store clientID as a channel attribute.
+            AttributeKey<UUID> clientIdAttrKey = AttributeKey.valueOf("ClientID");
```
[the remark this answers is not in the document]
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+            ctx.channel().attr(clientIdAttrKey).set(clientId);
+            log.info("Handshake validated by Server.");
```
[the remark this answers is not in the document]
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,213 @@
...
+
+            // Flush messages in queue
+            log.debug("There are {" + this.messages.size() + "} messages in queue to " +
+                    "be flushed.");
+            for (CorfuMsg message : this.messages) {
+                ctx.writeAndFlush(message);
+            }
+
+            // Remove this handler from the pipeline; handshake is completed.
+            log.info("Removing handshake handler from pipeline.");
```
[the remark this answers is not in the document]
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
annym: I wasn't aware of the ReadTimeoutHandler, it is really useful. I also lean towards this approach. What I am doing now is adding the ReadTimeoutHandler upon handshake initiation, and removing it on ChannelRead.. This also removes the latch and the mutex

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
annym: Done

annym:
@no2chem I have addressed your comments.  Please let me know if there is any other comment. The only thing remaining is replacing to the actual NodeId, so there is a dependency on these other PRs.

generator/src/main/java/org/corfudb/generator/LongevityApp.java
```diff
@@ -114,17 +114,17 @@ private void waitForAppToFinish() {
...
-            try {
-                checkpointHasFinished = checkpointer.awaitTermination(APPLICATION_TIMEOUT_IN_MS, TimeUnit.MILLISECONDS);
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-                exitStatus = 1;
-            }
-
-            exitStatus = checkpointHasFinished ? 0 : 1;
-            System.exit(exitStatus);
+//            boolean checkpointHasFinished = false;
```
no2chem: Is there a reason this is part of this PR?

annym: oops.. will remove, this was part of another test I was doing... plus I'll add the javadocs

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,197 @@
...
+        ctx.pipeline().remove(this);
+        this.fireHandshakeSucceeded();
+    }
+
+    @Override
+    public void channelActive(final ChannelHandlerContext ctx) throws Exception {
+        log.info("channelActive: Incoming connection established from: " +
+                ctx.channel().remoteAddress() + " Start Read Timeout.");
+        ctx.pipeline().addBefore(ctx.name(), "readTimeoutHandler",
+                new ReadTimeoutHandler(10));
```
medhavidhawan: should 10 be a constant or a config property ?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,197 @@
...
+        if (this.handshakeComplete.get()) {
+            log.debug("write: Handshake already completed, not appending corfu message to queue");
+            super.write(ctx, msg, promise);
+        } else {
+            this.messages.offer((CorfuMsg) msg);
+        }
+    }
+
+
+    private void fireHandshakeFailed(ChannelHandlerContext ctx) {
```
medhavidhawan: do we need 2 variables "handshakeComplete and handshakeFailed" to maintain state ?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -0,0 +1,197 @@
...
+            return;
+        }
+
+        CorfuPayloadMsg<HandshakeResponse> handshakeResponse;
+
+        try {
+           handshakeResponse = (CorfuPayloadMsg<HandshakeResponse>) m;
+            log.info("channelRead: Handshake Response received. Removing readTimeoutHandler " +
+                    "from pipeline.");
+            ctx.pipeline().remove("readTimeoutHandler");
```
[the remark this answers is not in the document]
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
no2chem: Hm, yeah it should be a config property. Can you pull this from the options map in ServerContext?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -0,0 +1,197 @@
...
+     */
+    public ServerHandshakeHandler(UUID nodeId, String corfuVersion) {
+        this.nodeId = nodeId;
+        this.corfuVersion = corfuVersion;
+        this.handshakeComplete = new AtomicBoolean(false);
+        this.handshakeFailed = new AtomicBoolean(false);
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
```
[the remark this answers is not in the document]
no2chem: I think for once this is rebased against master with the CorfuServer branch, integration testing will be possible. Unit testing will have to unfortunately come a little later.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
annym: I believe we should use 2. The fact that handshake hasn't failed (default) does not imply it has completed. Only if it has completed we might forward incoming messages in the pipeline when received in the window of handshake validation when handler has not been removed yet from the pipeline.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -0,0 +1,197 @@
...
+ * otherwise, the handshake times out, and either server or client close the connection.
+ *
+ * Created by amartinezman on 12/8/17.
+ */
+@Slf4j
+public class ClientHandshakeHandler extends ChannelDuplexHandler {
+
+    private final UUID clientId;
+    private final UUID nodeId;
+    private final AtomicBoolean handshakeComplete;
```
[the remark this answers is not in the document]
annym: As before, you might enter a state in which handshake hasn't failed (the default value) but not because it actually completed but because nothing has happened yet. Or it might have failed not because it completed but because it timed out.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -0,0 +1,197 @@
...
+     */
+    public ClientHandshakeHandler(@NonNull UUID clientId, @NonNull UUID serverId) {
+        this.clientId = clientId;
+        this.nodeId = serverId;
+        this.handshakeComplete = new AtomicBoolean(false);
+        this.handshakeFailed = new AtomicBoolean(false);
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object m)
```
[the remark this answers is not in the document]
annym: I have included comments on all methods. Yes, we do plan to have integration tests.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -0,0 +1,197 @@
...
+
+        try {
+           handshakeResponse = (CorfuPayloadMsg<HandshakeResponse>) m;
+            log.info("channelRead: Handshake Response received. Removing readTimeoutHandler " +
+                    "from pipeline.");
+            ctx.pipeline().remove("readTimeoutHandler");
+        } catch (ClassCastException e) {
+            log.warn("channelRead: Non-handshake message received by handshake handler. Send upstream only " +
+                    "if handshake succeeded.");
+            if (this.handshakeComplete.get()) {
```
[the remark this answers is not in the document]
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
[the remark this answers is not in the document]
annym: Done

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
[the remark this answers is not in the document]
medhavidhawan: So the variable you use is an object. It has a third "construed" state: uninitialized which can be used for that, But as long as you are comfortable, I am ok. My suggestion would be to couple these variables together in one object so that they mutate together and there is less chance of mistakes for later as well.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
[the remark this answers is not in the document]
medhavidhawan: ok cool.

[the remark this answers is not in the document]
medhavidhawan: ok cool

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
medhavidhawan: The variables you use are java objects. These variables have a derived state "initialized/null" and you can use that for the signal you are looking for. I am ok with your approach and do not have a strong opinion. If 2 variables make it easier for you, I would suggest to couple them together in a state object so that they mutate together and there is little chance of making mistakes where we set one and forgot the other. The code that you have does not have any issues, the logic looks good , this is more of suggestion for future proofing.

Maithem:
The commit history is a bit off, please rebase.

no2chem:
@annym 
It looks like its failing because the option wasn't set on the TestServerContext
```
                    context.getServerConfig(Integer.class, "--HandshakeTimeout")));
```
is null. 

In the test package :  org.corfudb.infrastructure.ServerContextBuilder
Create a new handshakeTimeout field and set that field in .ServerContextBuilder::Build

no2chem:
@annym Also might want to double check that you get an integer (seems like docopt will probably actually give you a string).

annym:
@medhavidhawan I get your point. I have created a Handshake State Object so both mutate together. 

@no2chem Good catch, already added it to ServerContextBuilder as a String and made the required changes.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **aed31d9**.

 [Pull Request #1078 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/sGKfPzApcJNo454tOh58vCKMWpkRSlwS)

PR #1084, opened 2017-12-14
Fix Orchestrator Query

Maithem:
Properly add/remove workflow ids when tracking workflows.

## Overview

Description:

This PR fixes a bug introduced in #940. Workflows were being added by a different key than they were being removed, as a consequence querying a workflow's id would always return "active" even if the workflow wasn't running.

Why should this be merged: 
Fixes CI on master.

- [x] Public API has Javadoc

### DIFF
--- a/format/proto/types.proto
+++ b/format/proto/types.proto
@@ -27,7 +27,7 @@ enum OrchestratorResponseType {
     // The status of a workflow
     WORKFLOW_STATUS = 0;
     // Id of a created workflow
-    WORKFLOW_ID = 1;
+    WORKFLOW_CREATE = 1;
 }
 
 message DataRank {
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
@@ -7,7 +7,7 @@ import org.corfudb.infrastructure.IServerRouter;
 import org.corfudb.protocols.wireprotocol.CorfuMsgType;
 import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
 import org.corfudb.protocols.wireprotocol.orchestrator.AddNodeRequest;
-import org.corfudb.protocols.wireprotocol.orchestrator.AddNodeResponse;
+import org.corfudb.protocols.wireprotocol.orchestrator.CreateWorkflowResponse;
 import org.corfudb.protocols.wireprotocol.orchestrator.OrchestratorRequest;
 import org.corfudb.protocols.wireprotocol.orchestrator.OrchestratorResponse;
 import org.corfudb.protocols.wireprotocol.orchestrator.QueryRequest;
@@ -19,7 +19,6 @@ import org.corfudb.runtime.view.Layout;
 
 import javax.annotation.Nonnull;
 import java.util.Map;
-import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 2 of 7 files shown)


no2chem:
Supersedes #1081

Maithem:
Let CI run a couple of times before we merge.

Maithem:
Fixes #1084

format/proto/types.proto
```diff
@@ -27,7 +27,7 @@ enum OrchestratorResponseType {
     // The status of a workflow
     WORKFLOW_STATUS = 0;
     // Id of a created workflow
```
no2chem: Comment doesn't seem to match anymore?

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
```diff
@@ -33,7 +32,7 @@
  * that is specified by Workflow.getActions() to achieve a bigger goal. For example, growing the
  * cluster. Initiated through RPC, the orchestrator will create a workflow instance and attempt
  * to execute all its actions.
- *
+ * <p>
```
no2chem: empty p tag

test/src/test/java/org/corfudb/runtime/clients/ManagementClientTest.java
```diff
@@ -80,13 +77,6 @@ public void handleBootstrap()
                 .isInstanceOf(ExecutionException.class);
     }
 
-    @Test
-    public void addNodeWorkflowRPCTest() throws Exception {
```
no2chem: Is this test not valid anymore or?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/orchestrator/CreateWorkflowResponse.java
```diff
@@ -6,29 +6,29 @@
...
 
-import static org.corfudb.format.Types.OrchestratorResponseType.WORKFLOW_ID;
+import static org.corfudb.format.Types.OrchestratorResponseType.WORKFLOW_CREATE;
 
 /**
  * AddNodeResponse returns the UUID of the add node workflow that was requested.
  * @author Maithem
  */
-public class AddNodeResponse implements Response {
+public class CreateWorkflowResponse implements Response {
```
no2chem: Also... this name doesn't match anymore...

test/src/test/java/org/corfudb/runtime/clients/ManagementClientTest.java
Maithem: It is, I just moved the logic to the other test.

Maithem:
Ok, let me address the comments about the comments and try a second run.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
Maithem: removed.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/orchestrator/CreateWorkflowResponse.java
Maithem: done.

format/proto/types.proto
Maithem: Workflow_id is used as a response for all create workflow operations. It will be shared, not just for create node workflow.

no2chem: But why is the enum called WORKFLOW_CREATE? I'm lost here.

Maithem: I'm not sure I understand your question. WORKFLOW_CREATE doesn't refer to create node workflow, it refers to any workflow that is created .

no2chem:
Looks like CI is passing now. Merging, thanks!

format/proto/types.proto
no2chem: The comments should probably say that then.

Maithem: Ok, will change it in the remove node PR.

PR #1089, opened 2017-12-14
Regenerate CorfuTable indexes if failed

no2chem:
## Overview

Description: This PR regenerates CorfuTable indexes if they have failed, building on top of #1088. This allows queries after the failed index to have the same performance as if the indexes were present the whole time.

Why should this be merged: This PR builds on top of #1088, improving performance on subsequent queries when the index fails.

Related issue(s) (if applicable): Fixes #1087

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #1093, opened 2017-12-14
Object Open Race

Maithem:
## Overview

Description:

When an object is opened, it is added to the cache right after
its creation, as a consequence other threads can retreive the object
from the object cache after its creation and before additional
initialization. For example, register index. This patch moves
the extra initialization into an atomic operation (i.e. computeIfAbsent.

Why should this be merged: 
Fixes a race bug in object open

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -159,11 +159,12 @@ public class CorfuTable<K ,V, F extends Enum<F> & CorfuTable.IndexSpecification,
         indexerClass = indexFunctionEnumClass;
         indexFunctions.addAll(EnumSet.allOf(indexFunctionEnumClass));
         indexFunctions.forEach(f -> indexMap.put(f, new HashMap<>()));
+        log.info("CorfuTable: creating CorfuTable with {} as indexer class", indexFunctionEnumClass);
     }
 
     /** Default constructor. Generates a table without any secondary indexes. */
     public CorfuTable() {
-        log.debug("CorfuTable: Creating a table without secondary indexes! Secondary index lookup"
+        log.info("CorfuTable: Creating a table without secondary indexes! Secondary index lookup"
             + " will DEGRADE to a full scan");
     }
 
@@ -284,23 +285,6 @@ public class CorfuTable<K ,V, F extends Enum<F> & CorfuTable.IndexSpecification,
 
         return projectionFunction.generateProjection(index, entryStream.filter(entryPredicate))
                 .collect(Collectors.toCollection(ArrayList::new));
-    }
-
-
-    /**
-     * Register new index class
-     *
-     * This replaces the current index.
-     *
-     * @param indexFunctionEnumClass
-     */
-    public void registerIndex(Class<F> indexFunctionEnumClass) {
-        indexerClass = indexFunctionEnumClass;
-        indexMap.clear();
-        indexFunctions.clear();
-        indexFunctions.addAll(EnumSet.allOf(indexFunctionEnumClass));
-        indexFunctions.forEach(f -> indexMap.put(f, new HashMap<>()));
-        mainMap.forEach(this::mapSecondaryIndexes);
     }
 
     /** {@inheritDoc} */
--- a/runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
@@ -61,7 +61,7 @@ public class ObjectBuilder<T> implements IObjectBuilder<T> {
 
     @SuppressWarnings("unchecked")
     public <R> ObjectBuilder<R> setTypeToken(TypeToken<R> typeToken) {
-        this.type = (Class<T>)typeToken.getRawType();
+        this.type = (Class<T>) typeToken.getRawType();
         return (ObjectBuilder<R>) this;
     }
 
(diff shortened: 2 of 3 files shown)


rogermichoud:
This will break the fast loader. 

At fastloading time, we don't have the index. This is when the table is created. Then when the client application open it, it won't register. This mechanism was there to ensure that we can register a index at open, even though the table already exists.

With this change, it means we cannot register an index after the table is created. We can go down that route, but we need to provide an API in the FastLoader to give the index. For this table.

Maithem:
@rogermichoud I see, that's because the fast loader loads streams and not objects. One way to solve this, is to also have a stream cache and let the fast loader populate that. The current mechanism has logic that conflates two things, opening a stream and binding an object to a stream.

no2chem:
Looking at the code more carefully now, I don't think there is a race here. The problem is probably somewhere else...

no2chem:
It looks like registerIndex can be running even though another thread may be already using the map. This would be a concurrency issue.

no2chem:
The problem is (I think) that registerIndex could be called multiple times, and I don’t think it operates under a lock. It will reset the indexFunctions while another thread is calling getByIndex.

Maithem:
We can mark it as a mutator to exclusively lock the object, but also we want to guard against returning the table without a table.

no2chem:
@Maithem - I suggest emitting a log message when registerIndex is called, and performing under a lock:

```
 CorfuTable currentCorfuTable = ((CorfuTable) result);
synchronized(currentCorfuTable) {
    if (currentCorfuTable.hasSecondaryIndices()) {
      ...
    }
}
```

no2chem:
@Maithem That's not sufficient, you need atomicity across the ```hasSecondaryIndices()``` and ```registerIndex()``` calls.

rogermichoud:
I can do a pull request with an api to provide an index to fastLoader.

Maithem:
@no2chem registerIndex registers a class of indexers, and it seems like only one class of indexers can be registered, isnt that sufficient ?

Maithem:
@rogermichoud I feel like the fastloader should be object agnostic, in that case you open all necessary streams before opening any object. If we are time constraint I think piping that down all the way to the Fastloader is fine, but we should consider the stream cache.

no2chem:
No, the problem is

Thread 1 calls open()
Thread 2 calls open()
Thread 1 retrieves fast-loaded (non-indexed) object
Thread 2 retrieves fast-loaded (non-indexed) object
Thread 1 checks if table has secondary indexes, sees none
Thread 2 checks if table has secondary indexes, sees none
Thread 1 and Thread 2 call registerIndex simultaneously.

I think. Race could also be somewhere else....

Maithem:
The register index logic is protected under computeIfAbsent atomicity and only one object can be created. Moreover, registerIndex doesn't make a lot of sense, it doesnt add an index, but overwrites. I will move it to the constructor

no2chem:
I don't think you're getting the problem.. that code block is only necessary because the fast loader doesn't have the index since it's not persisted. Moving it to the constructor won't work. I recommend just generating the index at query time as in #1089

no2chem:
And the computeifabsent doesn't help because it's already added to the cache by the fast loader

medhavidhawan:
I think what Michael is saying is correct that it does not help the cache problem. But it does make sure that construction of the objects is not done in parts and that too not atomically. The earlier construction was not right no matter what.

Maithem:
@medhavidhawan This PR addresses the problem you describe in the latter part of your statement. This is one part of a multi-part fix to address the whole problem.

Maithem:
@rogermichoud @medhavidhawan please review.

Maithem:
@rogermichoud It's a dependency, it will go away once I rebase.

rogermichoud:
@Maithem Yes, I saw afterward. removed the comment.

corfu-repo:

1. ![MINOR][MINOR] [ObjectBuilder.java#L17](https://github.com/CorfuDB/CorfuDB/blob/75c03ce9c4bb756b1aa16ac4892be68815bd83ee/runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java#L17): Remove this unused import 'org.corfudb.runtime.collections.CorfuTable'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

Maithem:
As Roger has stated, the FastLoader will create the tables with the index on first invocation.  

> this check should go away too. 

Which check are you talking about?

no2chem:
The stuff you moved into computeifabsent. It doesn't do anything there, right?

Maithem:
@no2chem are you referring to the serializer check ?

no2chem:
I see it was rebased. It didn't load. Please restore registerIndex, however. You can mark it as a MutatorAccessor to eliminate races.

Maithem:
> I don't think you're getting the problem.. that code block is only necessary because the fast loader doesn't have the index since it's not persisted. Moving it to the constructor won't work. I recommend just generating the index at query time as in #1089

Michael, making strong statements while being misinformed is not very judicious. I hope it is clear to you now what the scope of this patch is.

Also, the solution you recommend in #1089 doesn't seem like a good one. It attempts to address the side affect of a concurrency problem, not the core issue (i.e. proper synchronization). Furthermore, the table's index management control flow should be separated from the write/read path, but #1089 conflates them. Building the index from scratch on the fly during a read operation can cause unpredictable performance behavior.  I also haven't seen systems that do this, maybe you can reference one or two so that I can learn more about them.

PR #1098, opened 2017-12-15
Custom type initialization fix

rogermichoud:
## Overview
Ensure that custom types are built correctly in the FastLoader.

When passing customType objects to the fastloader, upon creation of the tables, it checks if it needs to reconstruct the table using the custom object loader or using the default one. The check used to be if the custom constructor is different than the default constructor.

Since now the default type used in MP is CorfuTable, when we were passing a custom constructor that happens to also be of CorfuTable type, we would just used the default type. Because of this, we were losing the indexer passed in the custom object builder.

This change the condition for constructing a table with the custom object builder. If the stream ID is in customTypes, we will reconstruct the table with the custom object builder.

Why should this be merged: 
This is a critical issue, that impact recreation of maps with indexer. It has a very low impact (low risk bugwise) and was tested on a live real system (MP)

- [x] Public API has Javadoc


rogermichoud:
@no2chem Sorry for lack of clearness in the description. This need is "somewhat" urgently needed for meeting our MP deadlines.

no2chem:
@rogermichoud Do you need to backport this as well?

PR #1102, opened 2017-12-17
Move seal epoch to Base Server

### LINKED ISSUE #1012: Move Seal functionality to Base Server
Currently SET_EPOCH is only handled by Layout Servers and thus a node cannot run an independent LogUnit/Sequencer. Moving out the Sealing functionality to Base Server will allow reconfigurations with independent running components.

zalokhan:
## Overview

Description: Moves the seal functionality to the Base Server from the Layout Server.

Why should this be merged: This allows corfu servers with log unit component to function independent of a layout server.

Related issue(s) (if applicable): Fixes #1012

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
@@ -3,18 +3,19 @@ package org.corfudb.infrastructure;
 import io.netty.channel.ChannelHandlerContext;
 
 import java.lang.invoke.MethodHandles;
-import java.util.HashMap;
 import java.util.Map;
 
 import javax.annotation.Nonnull;
 import lombok.Getter;
-import lombok.Setter;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.protocols.wireprotocol.CorfuMsg;
 import org.corfudb.protocols.wireprotocol.CorfuMsgType;
+import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
 import org.corfudb.protocols.wireprotocol.JSONPayloadMsg;
 import org.corfudb.protocols.wireprotocol.VersionInfo;
+import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.util.Utils;
 
 /**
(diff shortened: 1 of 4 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -67,6 +67,43 @@ private void getVersion(CorfuMsg msg, ChannelHandlerContext ctx,
         r.sendResponse(ctx, msg, new JSONPayloadMsg<>(vi, CorfuMsgType.VERSION_RESPONSE));
     }
 
+    boolean checkBootstrap(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
```
no2chem: This probably doesn't need to be here, right?

zalokhan: It still needs to check for bootstrap right ? Would it be safe to proceed with incrementing epoch even if the layout server is not bootstrapped?

no2chem: I think it should be safe to increment epoch (what if there is no layout service)?

We just don't want the server to serve a layout before being bootstrapped.... With clusterId that requirement should go away as well

zalokhan: I see. I'll make the change

zalokhan:
@no2chem There is race condition in this I just realized.
Now that "seal" and "commit" handling messages are in different classes, they are no longer synchronized. They both access the setServerEpoch methods.
A scenario can be set up where the seal is completed with epoch "x", followed by phase 1 & 2.
This is followed by another seal with epoch "x+1" which sets the server Epoch.
Finally the previous round commits with the epoch "x" setting the server epoch back to x which is incorrect.

no2chem:
@zalokhan shouldn't the layout server reject commits from the previous epoch, preventing this scenario?

zalokhan:
So in the code the check is made in the method but the setting of the server epoch is much later and is not atomic. There is a race possible.

no2chem:
@zalokhan would this be fixed by forcing all clients to disconnect after an epoch change (seal)? We shouldn't be accepting messages from previous epochs.

zalokhan:
No from the current code this wouldn't fix this race.

no2chem:
@zalokhan It seems like it would? You would need to change the LAYOUT_COMMITTED epoch to respect the epoch check. 

Alternatively, you could synchronize on the serverContext object, which should also fix the problem.

zalokhan:
@no2chem I'm still not sure Let me know if I get this right.
Let's consider the same scenario:
Seal is completed with epoch "x", followed by phase 1 & 2. We receive a commit message.
The handler receives it but has not yet set the epoch.
Now another seal comes in. Disconnects all the clients and sets the epoch to "x+1"

Now the on-going commit still completes setting the epoch back to "x". Its client has been disconnected but the handler will still complete executing the rest of the code right.

In this case the client disconnection did not help prevent the epoch from moving backwards. Neither will the synchronized - unless we do an epoch check in the synchronized block of the server context.

no2chem:
@zalokhan Good point. We need to wait for all clients to complete before ACKing the seal. We could do this perhaps by shutting down the event executor somehow, but okay, that will be a bit more complex. I would say that's probably what we want _eventually_.

So let's stick with the synchronized option then. In layout server during commit, take a lock on serverContext's object monitor. In base server's seal, take a lock on the same object monitor. 

That should force the layout commit to either observe the new epoch X+1 and fail, or force the new epoch seal to happen after the commit, fixing the race.

zalokhan:
Yes, that sounds good

Issue #1114, opened 2017-12-24
getRouterFunction Race


Maithem:
## Overview

**Bug report.** 
getRouterFunction is not thread safe, it can be called by multiple threads resulting in creating more than one router per endpoint. This can cause inconsistencies when invalidateLayout tries to update the epoch on each endpoint. One side effect that I observe is unexpected wrong epoch exceptions.

zalokhan:
This has now been fixed.
- We no more update the router epochs in the CorfuRuntime invalidateLayout loop.
- getRouter is an atomic operation now.  It only creates if its absent in the ConcurrentHashMap in the nodeRouterPool.
Please reopen if you still feel this is not fixed.

PR #1115, opened 2017-12-25
getRouter Race

Maithem:
Fixed a race in the getRouter method.

## Overview

Description:
This patch synchronizes the creation of node routers. This PR is needed to fix random WrongEpochException failures.  

Why should this be merged: 
Fixes https://github.com/CorfuDB/CorfuDB/issues/1114

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/management/PeriodicPollPolicy.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/management/PeriodicPollPolicy.java
@@ -102,7 +102,7 @@ public class PeriodicPollPolicy implements IFailureDetectorPolicy {
                     historyStatus.put(allServers[i], true);  // Assume it's up until we think it
                     // isn't.
                 }
-                historyRouters[i] = corfuRuntime.getRouterFunction.apply(allServers[i]);
+                historyRouters[i] = corfuRuntime.getRouter(allServers[i]);
                 historyPollFailures[i] = 0;
                 historyPollEpochExceptions[i] = 0;
             }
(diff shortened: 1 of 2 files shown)


Maithem:
> I've a separate fix for this race, but it occurs due to the test getRouterFunction not populating nodeRouters properly

There seems to be two possible races. The first one is when there is no override function (which this PR tries to fix). The second one is by the test supplied override functions. Does your fix address the first, or the second problem? Also, which branch is it on ? 

In both cases, node routers are being created and used before invalidate layout invokes setEpoch on the routers, which causes WrongEpochException if the system is on an epoch greater than 0 (default uninitialized long value)

>  I don't think WrongException is an actual exception (maybe you meant WrongEpochException)?

Yes.

no2chem:
It's on a branch which hasn't been pushed which changes how layouts are retrieved in general. I can push it if you really want to take a look at it, it's failing because of a bug in how WrongEpochException is handled in ChainReplicationView at the moment.

A node router shouldn't be retrieved until the runtime is instantiated. overrideGetRouterFunction needs to be set before the router is instantiated. If not, obviously that is a bug in the code and should be addressed there, by setting that function BEFORE getting a runtime. Exactly where are you seeing this race?

And of course, Id note I don't like the overridegetRouterFunction setter much. I'd like to deprecate overriding the router function soon, but I don't think this is the way to do it...

Maithem:
> I can push it if you really want to take a look at it

Yea, I can take a look. 

> Exactly where are you seeing this race?

When I did the Exception clean up I removed a couple of try/catch blocks that were swallowing exceptions thrown by reconfigureSequencerServers, so now I see WrongEpochException failures in random tests, its kind of rare too. 

I think this PR should be sufficient to fix the race issue for the case where overrideGetRouterFunction is not used. This won't fix the WrongEpochException failures in the unit tests, your branch might fix that problem.

no2chem:
See the layoutThread branch but it's still a WIP.

Is this a IT problem? How could you get WrongEpochException if you attempt to use a NettyClientRouter instead of a testrouter in test?

Maithem:
> Is this a IT problem? 
Nope, I was seeing it in the unit tests.

The race doesn't just pertain to the test router. This PR tries to fix a race that can happen when adding entries to nodeRouters. It actually doesn't fix the problem with the test routers. I think there are multiple issues here.

Issue #1106, opened 2017-12-19
Reloading Certs On the Client


Maithem:
## Overview

 **Bug report.**/**Feature request.** 
The server uses a corfu client in two places, the management server and the Orchestrator. They should be able to detect cert changes and reload them appropriately.

PR #1119, opened 2017-12-28
Use same CorfuRuntimeParameters

Maithem:
## Overview
The Orchestrator creates a new runtime for each workflow, this PR
makes the orchestrator use the same CorfuRuntimeParameters as the
management server's runtime. As a result, the security and auth
settings are carried over.

Related issue(s) (if applicable): #1106

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
@@ -35,6 +35,7 @@ import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
 import org.corfudb.protocols.wireprotocol.FailureDetectorMsg;
 import org.corfudb.protocols.wireprotocol.orchestrator.OrchestratorMsg;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuRuntime.CorfuRuntimeParameters;
 import org.corfudb.runtime.clients.LayoutClient;
 import org.corfudb.runtime.clients.ManagementClient;
 import org.corfudb.runtime.clients.SequencerClient;
@@ -181,7 +182,7 @@ public class ManagementServer extends AbstractServer {
             log.error("Error scheduling failure detection task, {}", err);
         }
 
-        orchestrator = new Orchestrator(this::getCorfuRuntime);
+        orchestrator = new Orchestrator(this::getCorfuRuntime, serverContext);
     }
 
     private void bootstrapPrimarySequencerServer() {
@@ -399,18 +400,8 @@ public class ManagementServer extends AbstractServer {
     public synchronized CorfuRuntime getCorfuRuntime() {
 
         if (corfuRuntime == null) {
-            corfuRuntime = new CorfuRuntime();
-            if ((Boolean) opts.get("--enable-tls")) {
-                corfuRuntime.enableTls((String) opts.get("--keystore"),
-                        (String) opts.get("--keystore-password-file"),
-                        (String) opts.get("--truststore"),
-                        (String) opts.get("--truststore-password-file"));
-                if ((Boolean) opts.get("--enable-sasl-plain-text-auth")) {
-                    corfuRuntime.enableSaslPlainText(
-                            (String) opts.get("--sasl-plain-text-username-file"),
-                            (String) opts.get("--sasl-plain-text-password-file"));
-                }
-            }
+            CorfuRuntimeParameters params = serverContext.getDefaultRuntimeParameters();
+            corfuRuntime = CorfuRuntime.fromParameters(params);
             // Runtime can be set up either using the layout or the bootstrapEndpoint address.
             if (latestLayout != null) {
                 latestLayout.getLayoutServers().forEach(ls -> corfuRuntime.addLayoutServer(ls));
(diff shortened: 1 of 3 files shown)


Maithem:
@medhavidhawan I still don't have a good way of testing this. I manually tested and it seems to be working. I'm looking for ways on how to include this in our automated testing. Note that this patch is a trivial change, the reloadable trust store logic has been committed earlier by @zjohnny226

runtime/src/main/java/org/corfudb/util/NodeLocator.java
```diff
@@ -37,6 +37,10 @@
 
     /** The port number on the host the node is located on. */
     final int port;
+    
+    public String getConnectionString() {
```
no2chem: no javadoc. Also, why do we need this? .toString() gets the correct connection string, which may have a node id?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -336,7 +336,7 @@ private CorfuRuntime(@Nonnull CorfuRuntimeParameters parameters) {
         this.parameters = parameters;
         layoutServers = new ArrayList<>();
         layoutServers.addAll(this.parameters.getLayoutServers().stream()
-                                .map(NodeLocator::toString)
+                                .map(NodeLocator::getConnectionString)
```
no2chem: see other comment

test/src/test/java/org/corfudb/runtime/CorfuRuntimeTest.java
```diff
@@ -121,6 +121,16 @@ private Layout get3NodeLayout() throws Exception {
         return l;
     }
 
+    //@Test
```
no2chem: test is commented out?

Maithem:
@no2chem 
You can see the problem by running the following test

`    @Test
    public void simpleStreamTest() throws Exception {
        CorfuRuntime rt = new CorfuRuntime("localhost:9000").connect();
        rt.invalidateLayout();
        System.out.println(rt.getLayoutView().getLayout());


        CorfuRuntime.CorfuRuntimeParameters parameters = CorfuRuntime.CorfuRuntimeParameters.builder()
                .layoutServer(NodeLocator.parseString("localhost;9000")).build();
        
        CorfuRuntime rt2 =  CorfuRuntime.fromParameters(parameters).connect();
        rt2.invalidateLayout();
        System.out.println(rt2.getLayoutView().getLayout());
    }`


I'll clean this and #1121 just want feedback.

Maithem:
@medhavidhawan ok, CI was failing because the orcheterator ITs actually test this change. #1123 fixed a bug and now it seems to be working. 

@no2chem your old comments are no longer applicable, take a look one more time.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
```diff
@@ -143,12 +147,29 @@ synchronized void dispatch(CorfuPayloadMsg<OrchestratorMsg> msg,
...
+            getRuntime.call().invalidateLayout();
+            Layout currLayout = getRuntime.call().getLayoutView().getLayout();
+
+            List<NodeLocator> servers = currLayout.getAllServers().stream()
+                    .map(NodeLocator::parseString)
+                    .collect(Collectors.toList());
+
+            CorfuRuntimeParameters currParams = getRuntime.call().getParameters();
+            CorfuRuntimeParameters params = CorfuRuntimeParameters.builder()
+                    .cacheDisabled(true)
```
no2chem: Sorry, missed this on my first pass. Can you add comments as to why you're disabling the cache, etc. Why do the parameters differ?

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -399,18 +400,17 @@ public void handleHeartbeatRequest(CorfuMsg msg, ChannelHandlerContext ctx, ISer
...
-                        (String) opts.get("--keystore-password-file"),
-                        (String) opts.get("--truststore"),
-                        (String) opts.get("--truststore-password-file"));
-                if ((Boolean) opts.get("--enable-sasl-plain-text-auth")) {
-                    corfuRuntime.enableSaslPlainText(
-                            (String) opts.get("--sasl-plain-text-username-file"),
-                            (String) opts.get("--sasl-plain-text-password-file"));
-                }
-            }
+            CorfuRuntimeParameters params = CorfuRuntimeParameters.builder()
```
no2chem: I'd put this in serverContext, in case anything else happens to need a runtime. Maybe call it getDefaultRuntimeParameters or something?

Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
Maithem: Added comments.

The workflows don't use the cache and just to make sure that the orchestrator doesn't use extra memory. 
It also doesn't open streams, so it doesn't need the fast loader.

Maithem:
@no2chem addressed the comments.

PR #1128, opened 2018-01-03
Clear Node

Maithem:
## Overview
Description:
When the add node workflow runs, it should clear the state of the
endpoint that it is trying to add.

Why should this be merged: 
This PR fixes a bug in the add node workflow. 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -99,7 +99,7 @@ public class LogUnitServer extends AbstractServer {
 
     private final StreamLog streamLog;
 
-    private final BatchWriter<Long, ILogData> batchWriter;
+    private volatile BatchWriter<Long, ILogData> batchWriter;
 
     /**
      * Returns a new LogUnitServer.
@@ -271,6 +271,7 @@ public class LogUnitServer extends AbstractServer {
     private void flushCache(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
         try {
             dataCache.invalidateAll();
+            log.info("flushCache: flushed cache");
         } catch (RuntimeException e) {
             log.error("Encountered error while flushing cache {}", e);
         }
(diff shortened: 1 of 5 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -294,9 +295,15 @@ private void rangeWrite(CorfuPayloadMsg<RangeWriteMsg> msg,
...
     @ServerHandler(type = CorfuMsgType.RESET_LOGUNIT)
-    private void resetLogUnit(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+    private synchronized void resetLogUnit(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+        // Waits for all current writes to get flushed
+        // This method is called after a seal and before consensus
+        // therefore writes wouldn't interleave with the reset
+        log.info("resetLogUnit: reset server msg1 {}", msg);
         streamLog.reset();
         dataCache.invalidateAll();
+        batchWriter = new BatchWriter(streamLog);
```
zalokhan: Can't we pass the flush and reset through the batchwriter as it ensures single threaded execution.

no2chem: Why do you need a new BatchWriter instance? Can we avoid this? Or clean shutdown the writerService, at least.

Maithem: hmm, I initially had batchWriter.close() then batchWriter = new BatchWriter(streamLog);
I'll add that back. The reason i create a new instance is because I close the current one. The reason I close the current one is because it waits for all the requests to get flushed.

Maithem: this logic will be called after a seal, so I don't think you can have interleaving if you clear/wait for the batchwriter to finish all its work then do the reset.

no2chem: I think cleanly shutting down the writerService and restarting it should suffice for flushing it right (you should be guaranteed the writerService thread exits if it is running if you call cleanShutdown) --- 

but I see that
```
                    } catch (Exception e) {
                        currOp.setException(e);
                        res.add(currOp);
                    }
```
In the batchWriterProcessor() method could swallow an InterruptedException, so you need to fix that.

no2chem: Sorry, meant
```
executorService.shutdownNow();
executorService.awaitTermination();
```
was thinking about netty thread groups there

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **bda23f4**.

 [Pull Request #1128 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/9sjllUMBVs2di0Ek0nY70jgk3Hduuw5c)

Maithem:
Not needed anymore.

PR #1144, opened 2018-01-10
Code Re-org

Maithem:
## Overview
Description:

Moved orchestrator types from types.proto to the wireprotocol
package and add clients for the orchestrator.

Why should this be merged: 
Required for clustering robustness 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/AddNodeWorkflow.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/AddNodeWorkflow.java
@@ -1,12 +1,15 @@
-package org.corfudb.protocols.wireprotocol.orchestrator;
+package org.corfudb.infrastructure.orchestrator;
 
 import com.google.common.collect.ContiguousSet;
 import com.google.common.collect.DiscreteDomain;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Range;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.LogData;
+import org.corfudb.protocols.wireprotocol.orchestrator.AddNodeRequest;
+import org.corfudb.protocols.wireprotocol.orchestrator.Request;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.clients.LogUnitClient;
 import org.corfudb.runtime.exceptions.AlreadyBootstrappedException;
@@ -14,7 +17,6 @@ import org.corfudb.runtime.view.Layout;
 
 import javax.annotation.Nonnull;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -47,6 +49,9 @@ public class AddNodeWorkflow implements IWorkflow {
     @Getter
     final UUID id;
 
+    @Getter
+    final List<Action> actions;
+
     /**
      * Creates a new add node workflow from a request.
      *
@@ -55,19 +60,15 @@ public class AddNodeWorkflow implements IWorkflow {
     public AddNodeWorkflow(Request request) {
         this.id = UUID.randomUUID();
         this.request = (AddNodeRequest) request;
+        actions = ImmutableList.of(new BootstrapNode(),
+                new AddNodeToLayout(),
+                new StateTransfer(),
+                new MergeSegments());
     }
 
     @Override
     public String getName() {
         return ADD_NODE.toString();
-    }
-
-    @Override
-    public List<Action> getActions() {
-        return Arrays.asList(new BootstrapNode(),
-                new AddNodeToLayout(),
-                new StateTransfer(),
-                new MergeSegments());
     }
 
     /**
@@ -82,8 +83,6 @@ public class AddNodeWorkflow implements IWorkflow {
 
         @Override
         public void impl(@Nonnull CorfuRuntime runtime) throws Exception {
-            changeStatus(ActionStatus.STARTED);
-
             try {
                 runtime.getLayoutManagementView().bootstrapNewNode(request.getEndpoint());
             } catch (Exception e) {
@@ -91,11 +90,9 @@ public class AddNodeWorkflow implements IWorkflow {
                     log.info("BootstrapNode: Node {} already bootstrapped, skipping.", request.getEndpoint());
                 } else {
                     log.error("execute: Error during bootstrap", e);
-                    changeStatus(ActionStatus.ERROR);
+                    throw e;
                 }
             }
-
-            changeStatus(ActionStatus.COMPLETED);
         }
     }
 
@@ -114,16 +111,7 @@ public class AddNodeWorkflow implements IWorkflow {
 
         @Override
         public void impl(@Nonnull CorfuRuntime runtime) throws Exception {
-            changeStatus(ActionStatus.STARTED);
             Layout currentLayout = new Layout(runtime.getLayoutView().getLayout());
-
-            if (currentLayout.getAllServers().contains(request.getEndpoint())) {
-                log.info("Node {} already exists in the layout, skipping.", request.getEndpoint());
-                newLayout = currentLayout;
-                changeStatus(ActionStatus.COMPLETED);
-                return;
-            }
-
             runtime.getLayoutManagementView().addNode(currentLayout, request.getEndpoint(),
                     true, true,
                     true, false,
@@ -131,7 +119,6 @@ public class AddNodeWorkflow implements IWorkflow {
 
             runtime.invalidateLayout();
             newLayout = new Layout(runtime.getLayoutView().getLayout());
-            changeStatus(ActionStatus.COMPLETED);
             return;
 
         }
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/orchestrator/OrchestratorResponseType.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/orchestrator/OrchestratorResponseType.java
@@ -1,0 +1,42 @@
+package org.corfudb.protocols.wireprotocol.orchestrator;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ *
+ * The type of responses that the orchestrator replies with.
+ *
+ * Created by Maithem on 1/2/18.
+ */
+@AllArgsConstructor
+public enum OrchestratorResponseType {
+    /**
+     * The status of a workflow
+     */
+    WORKFLOW_STATUS(0, QueryResponse::new),
+
+    /**
+     * Id of a created workflow
+     */
+    WORKFLOW_CREATED(1, CreateWorkflowResponse::new);
+
+    @Getter
+    public final int type;
+
+    @Getter
+    final Function<byte[], Response> responseGenerator;
+
+    /**
+     * Map an int to an enum.
+     */
+    static final Map<Integer, OrchestratorResponseType> typeMap =
+            Arrays.stream(OrchestratorResponseType.values())
+                    .collect(Collectors.toMap(OrchestratorResponseType::getType,
+                            java.util.function.Function.identity()));
+}
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -259,6 +259,11 @@ public class CorfuRuntime {
          */
         UncaughtExceptionHandler uncaughtExceptionHandler;
         //endregion
+
+        /**
+         * The number of times to retry invalidate when a layout change is expected.
+         */
+        @Default int invalidateRetry = 3;
     }
 
     /**
--- a/runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
@@ -201,7 +201,8 @@ public class LayoutBuilder {
      */
     public LayoutBuilder mergePreviousSegment(int segmentIndex) {
         if (segmentIndex < 1) {
-            throw new LayoutModificationException("No segments to merge.");
+            log.warn("mergePreviousSegment: No segments to merge.");
+            return this;
         }
 
         List<LayoutSegment> layoutSegmentList = layout.getSegments();
--- a/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
@@ -1,17 +1,21 @@
 package org.corfudb.runtime.view;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.TimeoutException;
 
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.protocols.wireprotocol.orchestrator.CreateWorkflowResponse;
-import org.corfudb.protocols.wireprotocol.orchestrator.QueryResponse;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.clients.ManagementClient;
+import org.corfudb.runtime.exceptions.NetworkException;
+import org.corfudb.runtime.exceptions.WorkflowException;
 import org.corfudb.util.Sleep;
-import org.corfudb.util.Utils;
+
+import javax.annotation.Nonnull;
 
 /**
  * A view of the Management Service to manage reconfigurations of the Corfu Cluster.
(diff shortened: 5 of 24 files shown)


corfu-repo:

1. ![MINOR][MINOR] [ManagementView.java#L11](https://github.com/CorfuDB/CorfuDB/blob/332a4805828bae1e66b58b9132284a7678fc71f1/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java#L11): Remove this unused import 'org.corfudb.protocols.wireprotocol.orchestrator.QueryResponse'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -201,7 +201,9 @@ public LayoutBuilder addLogunitServer(int stripeIndex, long globalLogTail,
      */
     public LayoutBuilder mergePreviousSegment(int segmentIndex) {
         if (segmentIndex < 1) {
-            throw new LayoutModificationException("No segments to merge.");
+            log.warn("mergePreviousSegment: No segments to merge.");
+            return this;
+            //throw new LayoutModificationException("No segments to merge.");
```
corfu-repo: MAJOR This block of commented-out lines of code should be removed. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3ACommentedOutCodeLine)

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -1,7 +1,9 @@
 package org.corfudb.runtime.view;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.TimeoutException;
```
corfu-repo: MINOR Remove this unused import 'java.util.concurrent.TimeoutException'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -10,6 +12,7 @@
 import org.corfudb.protocols.wireprotocol.orchestrator.QueryResponse;
```
corfu-repo: MINOR Remove this unused import 'org.corfudb.protocols.wireprotocol.orchestrator.QueryResponse'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

corfu-repo:
SonarQube analysis reported 3 issues
* ![MAJOR][MAJOR] 1 major
* ![MINOR][MINOR] 2 minor

[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +24,77 @@
 @Slf4j
 public class ManagementView extends AbstractView {
 
-    private final long layoutRefreshTimeout = 500;
+    static final Duration WORKFLOW_TIMEOUT = Duration.ofMinutes(10);
```
no2chem: these constants don't belong here. They actually look like they need to be part of the servercontext

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +24,77 @@
...
-    public boolean isWorkflowActive(String server, UUID workflowId) {
-        return layoutHelper(l -> {
-            QueryResponse response = runtime.getRouter(server)
-                    .getClient(ManagementClient.class)
-                    .queryRequest(workflowId);
-            log.info("isWorkflowActive: {} on server:{} = {}",
-                workflowId, server, response.isActive());
-            return response.isActive();
-        });
+    void waitForWorkflow(UUID workflow, ManagementClient client) {
```
no2chem: You should mark this as private

no2chem: Also, shouldn't it return whether the workflow is finished or whether it failed to query?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +24,77 @@
...
+            try {
+                if (!client.queryRequest(workflow).isActive()) {
+                    break;
+                }
+            } catch (NetworkException e) {
+                log.warn("waitForWorkflow: Network exception while waiting for {} on try {}", workflow, x);
+            }
+            Sleep.sleepUninterruptibly(QUERY_INTERVAL);
+            log.info("waitForWorkflow: waiting for {} on try {}", workflow, x);
+        }
```
no2chem: Seems like it should throw a TimeoutException here?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +24,77 @@
...
+    void waitForWorkflow(UUID workflow, ManagementClient client) {
+        long tries = WORKFLOW_TIMEOUT.getSeconds() / QUERY_INTERVAL.getSeconds();
+        log.info("waitForWorkflow: num tries {}", tries);
+        for (long x = 0; x < tries; x++) {
+            try {
+                if (!client.queryRequest(workflow).isActive()) {
+                    break;
+                }
+            } catch (NetworkException e) {
+                log.warn("waitForWorkflow: Network exception while waiting for {} on try {}", workflow, x);
```
no2chem: Why swallow the NetworkException?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +24,77 @@
...
+            }
+            Sleep.sleepUninterruptibly(QUERY_INTERVAL);
+            log.info("waitForWorkflow: waiting for {} on try {}", workflow, x);
+        }
+    }
+
+    Layout getMaxLayout(CorfuRuntime rt) {
```
no2chem: Should be private? Also, what is it for? It probably doesn't do what you expect it to do

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +24,77 @@
...
+            }
+        }
+        return max;
     }
 
     /**
      * Add a new node to the existing cluster.
      *
-     * @param endpoint Endpoint of the new node to be added to the cluster.
+     * @param endpointToAdd Endpoint of the new node to be added to the cluster.
```
no2chem: Why not just have it throw an exception instead of returning true/false?

test/src/test/java/org/corfudb/integration/WorkflowIT.java
```diff
@@ -50,42 +46,42 @@ public void AddNodeIT() throws Exception {
...
+        CorfuRuntime n1Rt = new CorfuRuntime(getConnectionString(n1Port)).setCacheDisabled(true).connect();
 
         CorfuTable table = n1Rt.getObjectsView()
                 .build()
                 .setType(CorfuTable.class)
                 .setStreamName(streamName)
                 .open();
 
-        final int numEntries = 12_000;
+        final int numEntries = 30_000;
```
no2chem: Please use a parameter instead of choosing an arbitrary number

test/src/test/java/org/corfudb/integration/WorkflowIT.java
```diff
@@ -98,46 +94,19 @@ public void AddNodeIT() throws Exception {
...
-                    Thread.sleep(sleepTime);
-                } else {
-                    break;
-                }
-            } catch (Exception e) {
-                rt.invalidateLayout();
-                Thread.sleep(sleepTime);
-            }
-        }
+        System.out.println("stop");
```
no2chem: println

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: servercontext or runtime parameters ?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: I'll rethrow the exception, that way if waitForWorkflow finishes then it finished waiting, otherwise if it throws an exception then it means that if failed to query. Based on that i'll retry.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Decrease the chance of the layout's epoch to regress and makes the operation less likely to retry.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: I will throw a workflow failure exception

test/src/test/java/org/corfudb/integration/WorkflowIT.java
Maithem: Ok.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Throw a TimeoutException if the workflow takes more than the max wait time?

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
no2chem: probably from a ServerContext. Which means that WORKFLOW_TIMEOUT Duration should be provided to addNode as a parameter.=

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
no2chem: or retries?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
no2chem: This seems to be a hack, maybe we should keep the layout from regressing in the first place instead.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: It makes sense to throw a TimeoutException when waitForWorkflow executes without the "break" ever being called, I don't see why you would throw a TimeoutException on retrying within the same method.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: I'll move this to invalidate layout

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: This is on the client side why would you want it to be part of the server context?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
no2chem: That's what I meant. Throw an exception if the workflow is still active after X retries, don't exit silently.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
no2chem: As in a client should pass these constants when the methods are called, I'm guessing from an orchestrator the constants should come from the server context.

Maithem: Yes, but the orchestrator doesn't invoke this method. Its the "application" that calls this through the runtime.

no2chem: Then the application should provide these parameters

Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Not anymore. Done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Done.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Done.

Maithem:
I addressed your comments @medhavidhawan . Core orchestrator types have been moved to the infrastructure package.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/orchestrator/OrchestratorResponseType.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Created by box on 1/2/18.
+ */
+@AllArgsConstructor
+public enum OrchestratorResponseType {
```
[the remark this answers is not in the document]
Maithem: done.

Maithem:
@no2chem I addressed all your comments.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/AddNodeWorkflow.java
```diff
@@ -55,21 +60,17 @@
     public AddNodeWorkflow(Request request) {
         this.id = UUID.randomUUID();
         this.request = (AddNodeRequest) request;
+        actions = Arrays.asList(new BootstrapNode(),
```
no2chem: Can you use a ImmutableList (Guava)?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -259,6 +259,15 @@
          */
         UncaughtExceptionHandler uncaughtExceptionHandler;
         //endregion
+
+        /** {@link Duration} before a workflow times out. */
+        @Default Duration workflowTimeout = Duration.ofMinutes(10);
```
no2chem: This isn't quite what I asked for. Workflow timeouts don't belong here.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +22,65 @@
...
      * Add a new node to the existing cluster.
      *
-     * @param endpoint Endpoint of the new node to be added to the cluster.
-     * @return True if completed successfully.
+     * @param endpointToAdd Endpoint of the new node to be added to the cluster.
+     * @throws WorkflowException throws a workflow exception on failure
      */
-    public boolean addNode(String endpoint) {
-        return layoutHelper(l -> {
+    public void addNode(String endpointToAdd) {
```
no2chem: addNode should take the timeouts as a parameter. Requiring a different timeout based on the workflow or caller seems like a common scenario.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +22,65 @@
...
-    public boolean isWorkflowActive(String server, UUID workflowId) {
-        return layoutHelper(l -> {
-            QueryResponse response = runtime.getRouter(server)
-                    .getClient(ManagementClient.class)
-                    .queryRequest(workflowId);
-            log.info("isWorkflowActive: {} on server:{} = {}",
-                workflowId, server, response.isActive());
-            return response.isActive();
-        });
+    void waitForWorkflow(UUID workflow, ManagementClient client) throws NetworkException,
```
no2chem: workflow timeouts should be a parameter here. Probably the number of retries, for that matter too.

test/src/test/java/org/corfudb/CorfuTestParameters.java
```diff
@@ -111,7 +111,7 @@ public CorfuTestParameters(){
         NUM_ITERATIONS_VERY_LOW = TRAVIS_BUILD ? 1 : 10;
         NUM_ITERATIONS_LOW = TRAVIS_BUILD ?  10 : 100;
         NUM_ITERATIONS_MODERATE = TRAVIS_BUILD ? 100: 1000;
-        NUM_ITERATIONS_LARGE = TRAVIS_BUILD ? 1_000 : 10_000;
+        NUM_ITERATIONS_LARGE = TRAVIS_BUILD ? 1_000 : 15_000;
```
no2chem: Why is this value being modified?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -21,64 +22,65 @@
...
-                Sleep.MILLISECONDS.sleepUninterruptibly(layoutRefreshTimeout);
-            }
-
-            runtime.invalidateLayout();
+            try {
+                CreateWorkflowResponse resp = client.addNodeRequest(endpointToAdd);
+                log.info("addNode: Workflow id {} for {}", resp.workflowId, endpointToAdd);
+                waitForWorkflow(resp.getWorkflowId(), client);
+                Layout newLayout = null;
+                for (int y = 0; y < 3; y++) {
```
no2chem: hard-coded constant

test/src/test/java/org/corfudb/CorfuTestParameters.java
Maithem: To use it in places where we want to write more entries than the size of segment.

no2chem: This changes the value for ALL the tests. This is NOT the place to change. Where is the segment size being set. Seems like a simple change would be to set the segment size to moderate rather than changing EVERY test.

Maithem: Then I'll just make it test variable instead. The segment size is not easily configurable right now.

no2chem: Or, you could make a NUM_ITERATIONS_VERY_LARGE. but I guess that should be 50k or 100k

Maithem: That's too large, some of the tests were timing out on 20k, i think.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/AddNodeWorkflow.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: Ok, will pass it through the method signature.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Number of retries for the timeout to expire, or number of retries of the workflow ? The former is just a function of the query retry and total time to wait.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Yes.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
Maithem: Ill just put this in the corfu runtime parameters. invalidating and check the layout a couple of times is something that we need to do in a couple of places.

no2chem: No. Retries is instance specific. Sometimes I might want to not retry and sometimes I might want to retry forever, on two parallel calls.

Make it a function parameter.

no2chem: I'd contrast this with the global rpc timeout. While it is  true that some functions may want a shorter or longer RPC time out, 99% of the functions are agnostic of the RPC timeout.

Maithem: This is a specific case, the invalidateRetry is only used where a layout is expected to change (i.e. currently attemptConsensus and after a workflow finishes executing), it won't be used for all layout invalidation. In that case, it seems a bit contrived that attempt consensus wants to invalidate the layout 3 times, but addnode wants to do it for for 5 times while both code paths are trying to achieve the exact thing. 

Let me know if you have a particular use case, otherwise I'll leave it as is, its a small change anyways.

Maithem:
@no2chem What are you not thrilled about exactly, the way retries are implemented, or the way the retry parameter is passed around ? A retry function can be passed around to reuse the retry logic. You can also make suggestions and I'll see if I can incorporate them in the remove node PR.

PR #1146, opened 2018-01-11
Sequencer Ready API

Maithem:
## Overview

Description:
The reconfigureSequencerServers doesn't consider the sequencer's status when it makes a decision without to push out configurations or not. This PR adds changes the logic to conisder the sequencer status by retrieving the status of the leading sequencer and checking if its not ready, in that case it will force push a configuration. 

Why should this be merged: 
Required for all workflows. While workflows execute the chances of failure are high, if a workflow changes the layout, but fails before the sequencer is configured problem the system freezes. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
@@ -51,6 +51,7 @@ public enum CorfuMsgType {
     TOKEN_RES(21, new TypeToken<CorfuPayloadMsg<TokenResponse>>(){}),
     BOOTSTRAP_SEQUENCER(22, new TypeToken<CorfuPayloadMsg<SequencerTailsRecoveryMsg>>(){}),
     SEQUENCER_TRIM_REQ(23, new TypeToken<CorfuPayloadMsg<Long>>() {}),
+    SEQUENCER_STATUS_REQ(24, TypeToken.of(CorfuMsg.class)),
 
     // Logging Unit Messages
     WRITE(30, new TypeToken<CorfuPayloadMsg<WriteRequest>>() {}),
--- a/runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
@@ -47,6 +47,32 @@ public class SequencerClient implements IClient {
         return msg.getPayload();
     }
 
+    /**
+     * Handle an ACK response from the server.
+     *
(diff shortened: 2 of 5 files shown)


runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
```diff
@@ -51,6 +51,7 @@
     TOKEN_RES(21, new TypeToken<CorfuPayloadMsg<TokenResponse>>(){}),
     BOOTSTRAP_SEQUENCER(22, new TypeToken<CorfuPayloadMsg<SequencerTailsRecoveryMsg>>(){}),
     SEQUENCER_TRIM_REQ(23, new TypeToken<CorfuPayloadMsg<Long>>() {}),
+    SEQUENCER_STATUS_REQ(24, TypeToken.of(CorfuMsg.class)),
```
no2chem: toIs this API really necessary? Shouldn't it be enough just to make a sequencer query and check if it fails?

runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
```diff
@@ -47,6 +47,32 @@ private static Object handleTokenResponse(CorfuPayloadMsg<TokenResponse> msg,
...
 
+    /**
+     * Handle an ACK response from the server.
+     *
+     * @param msg The ping request message
+     * @param ctx The context the message was sent under
+     * @param r   A reference to the router
+     * @return Always True, since the ACK message was successful.
+     */
+    @ClientHandler(type = CorfuMsgType.ACK)
```
no2chem: I don't think this is necessary, these messages are handled by baseclient already

Maithem: Apparently, the test router rails if these definitions are not here.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
Maithem: Hmm, I think so.

runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
no2chem: It shouldn't unless you don't have a base client? The handlers are mapped by type, so declaring them twice would overwrite the previous registration...

Maithem:
Ok, now that you point it out, I think this is redundant. Will query and check if it responds with NOT_READY

PR #1167, opened 2018-01-24
Fix channel race

no2chem:
## Overview

Description: When a channel is connected, there is a race in NettyClientRouter between when the channel is connected and when the completionFuture for the channel is fulfilled. As a result, a client can call sendMessageAndGetCompletableFuture() and encounter a NullPointerException if the channel is set after the completionFuture is fulfilled.

This PR also removes a deprecated ChannelHandlerContext that was being passed around, which is also vulnerable to the same race.

Why should this be merged: Resolves a NPE bug in NettyClientRouter.

- [x] Public API has Javadoc


medhavidhawan:
Is there a way to create a test for this ?

Issue #1016, opened 2017-11-23
CorfuTable Index


Maithem:
The CorfuTable interface to adding secondary indices is superfluous, we should think of a cleaner interface.

zalokhan:
Resolved by #1199

PR #1175, opened 2018-01-26
Flexible CorfuTable

Maithem:
## Overview
Introduced a new model that enables a flexible way to introduce
different indexes with a single CorfuTable instance.

Why should this be merged: 
Currently indexerClass enforces the same types for all secondary index functions, which is kind of restrictive. For example, two secondary indexes cannot map to two different types for a single CorfuTable. This FCorfuTable (i.e Flexable CorfuTable) type allows the secondary indexes to map to different types.

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/FCorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/FCorfuTable.java
@@ -1,0 +1,704 @@
+package org.corfudb.runtime.collections;
+
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.annotations.Accessor;
+import org.corfudb.annotations.ConflictParameter;
+import org.corfudb.annotations.CorfuObject;
+import org.corfudb.annotations.DontInstrument;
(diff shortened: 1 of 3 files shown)


no2chem:
Isn't this achieved by taking the index generic out of CorfuTable (or replacing it with Object)?

Maithem:
@no2chem that's one solution for the index function type, the other issue related to a use case we the registry is dynamically constructed at runtime, so an enum won't work. Looking into making the indexer class an iterable.

Maithem:
@medhavidhawan I will make these changes on CorfutTable once we agree on the index specification.

runtime/src/main/java/org/corfudb/runtime/collections/FCorfuTable.java
```diff
@@ -0,0 +1,704 @@
...
+    public @Nonnull
+    <I extends Comparable<I>>
+    Collection<Entry<K, V>> getByIndex(@Nonnull IndexName indexName, I indexKey) {
+        String name = indexName.get();
+        Map<K, V> res = secondaryIndexes.get(name).get(indexKey);
+        if (res == null) {
+            return Collections.emptySet();
+        }
+
+        return secondaryIndexes.get(name).get(indexKey).entrySet();
```
Maithem: need to wrap with immutable collection.

Maithem:
Addresses #1016

PR #1184, opened 2018-02-01
Fixing intermittently failing killAndHealNode

zalokhan:
## Overview

Description:
Fixing intermittently failing killAndHealNode IT.
Reducing waiting times on unit tests.

Why should this be merged: 
Fixes intermittent failures.

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
+++ b/test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
@@ -15,6 +15,7 @@ import java.util.Map;
 import java.util.Random;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
 
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.ReadResponse;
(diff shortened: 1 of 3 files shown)


rogermichoud:
Sometimes I had issues running tests on my machine because timeouts were too aggressive. Am I the only one experiencing this? Also, we might start to see some timeout exceptions as well in our CI. This is also especially relevant for IT tests.

Maithem:
ClusterReconfigIT doesn't actually pass on my machine because the timeout is too short.

zalokhan:
I see. I only changed the retry timeouts. Probably I made them too aggressive.
I'll remove the timeout changes and keep the test fixes.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -287,7 +287,8 @@ public void restartTest() throws Exception {
 
         restartServer(corfuRuntime, DEFAULT_ENDPOINT);
 
-        assertThat(corfuRuntime.getLayoutView().getLayout().getEpoch()).isEqualTo(l.getEpoch() + 1);
+        assertThat(corfuRuntime.getLayoutView().getLayout().getEpoch())
+                .isGreaterThanOrEqualTo(l.getEpoch() + 1);
```
Maithem: This doesn't match the comment. Why do you expect the epoch to be greater than or equal, is it because recovery can fail and the epoch can increase multiple times?

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -466,8 +467,6 @@ public void killAndHealNode() throws Exception {
                 .open();
         final String data = createStringOfSize(1_000);
         Random r = getRandomNumberGenerator();
-        final AtomicBoolean moreDataToBeWritten = new AtomicBoolean(true);
```
Maithem: Why did you remove this?

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -515,16 +514,12 @@ public void killAndHealNode() throws Exception {
         refreshedLayout = runtime.getLayoutView().getLayout();
         final long epochAfterHealingNode = 3L;
         // Waiting node to be healed and added back to the layout.
-        while (refreshedLayout.getEpoch() != epochAfterHealingNode) {
+        while (refreshedLayout.getEpoch() < epochAfterHealingNode) {
```
Maithem: It seems like you have these while loops in many places. I dont think this is a good idea, probably need to bound the retries.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
zalokhan: Yes that's the reason. I'll update the comment to reflect this.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
zalokhan: This is creating a daemon writer thread initially introduced to check if we can write after killing or adding. I anyways have a write after these 2 events (in the main test thread) and fail if these writes fail.
So I feel this extra thread is not required now.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
zalokhan: Done.

Maithem:
@annym is the injection-framework CI working, should we wait for it?

annym:
@Maithem yes it is working, I restarted them and this one is already complete.

Maithem:
@annym thanks, i see it now.

PR #1209, opened 2018-03-02
Accept network interface to bind.

zalokhan:
## Overview

Description: Introduce a new flag to accept network interface name to which the corfu server can bind to. Example:
`./corfu_server -q eth0 -ms 9000` will bind to interface `eth0`.

Why should this be merged: In scenarios where the public ip is not known, the server startup script can accept an interface to bind to whenever is starts.

- [x] Public API has Javadoc


zalokhan:
Oh I did not see this. Will fix this. Thanks.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -225,6 +238,34 @@ public static void main(String[] args) {
...
+     * Fetches the IP address given an interface name.
+     * Throws an unrecoverable error and aborts if the server is unable to find a valid address.
+     *
+     * @param interfaceName Network interface name.
+     * @return address for the interface name.
+     */
+    private static String getAddressFromInterfaceName(String interfaceName) {
+        try {
+            Enumeration<InetAddress> inetAddress = NetworkInterface.getByName(interfaceName)
+                    .getInetAddresses();
```
Maithem: why do you care about ipv4 and loopback devices? can't you just take the SocketAddress object and pass it to netty ?

zalokhan: The interface has both inet (ipv4) and inet6 (v6) addresses. So I need the check for ipv4 addresses. I can remove the loopback check.

Maithem: why do you care about that? just let it fail on the bind ?

zalokhan: Each interface has ipv4 and ipv6 addresses and we are interested to bind to the v4 address only.

zalokhan:
I have defaulted the binding to localhost in absence of interface or address.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -225,6 +245,33 @@ public static void main(String[] args) {
...
     }
 
+    /**
+     * Fetches the IP address given an interface name.
+     * Throws an unrecoverable error and aborts if the server is unable to find a valid address.
+     *
+     * @param interfaceName Network interface name.
+     * @return address for the interface name.
+     */
+    private static String getAddressFromInterfaceName(String interfaceName) {
```
no2chem: This doesn't belong in CorfuServer. Move it to a utility class

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -174,6 +181,11 @@
                     + " --version                                                                "
                     + "              Show version\n";
 
+    private static final String MEMORY = "--memory";
```
no2chem: These really don't need to be static constants

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -442,9 +491,9 @@ public static void restartServer(ServerContext serverContext, boolean resetData)
 
         corfuServerThread = new Thread(() -> {
             cleanShutdown((NettyServerRouter) serverContext.getServerRouter());
-            if (resetData && !(Boolean) serverContext.getServerConfig().get("--memory")) {
+            if (resetData && !(Boolean) serverContext.getServerConfig().get(MEMORY)) {
```
no2chem: We don't need the constants

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -307,7 +356,7 @@ ChannelFuture startAndListen(@Nonnull EventLoopGroup bossGroup,
             bootstrapConfigurer.configure(bootstrap);
 
             bootstrap.childHandler(getServerChannelInitializer(context, router));
-            return bootstrap.bind(port).sync();
+            return bootstrap.bind(address, port).sync();
```
no2chem: I don't think this is necessary and it seems that it will break the previous behavior which binded to all addresses.

zalokhan: We don't want that right? We are not using the `--address` field then.
Shouldn't it bind only to one address.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
zalokhan: Ill remove these

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
zalokhan: Shouldn't they be constants? They are used in a lot of places (eg. --address, --memory).
ServerContext, ManagementServer, LogunitServer.

no2chem: I think we should only use constants for numeric values or strings where a value may easily be changed (e.g. timeout) or there is some performance reason.

Here I think it's unlikely that these values will change and it's cumbersome to define these constants in addition to having them in the docopt string.

Maybe if Java supported template literals we could use the constant in the docopt string too...

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
no2chem: The address field is currently being used only to populate the layout (when -s is specified), nothing else. Is there a use case for listening on only one address?

zalokhan: The current implementation would be bug prone if we attempt to run 2 corfu servers on different interfaces with the same port. We can bind to an address if provided or default to binding to all addresses if nothing is specified. Would this be reasonable?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
zalokhan: I see. I'll remove them.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
no2chem: I'm concerned about keeping the current behavior.

Does the use case you're concerned about actually exist?
Are there cases where you have a server with two interfaces and want to serve TWO different servers on the same port?

zalokhan:
@no2chem @Maithem Addressed the reviews.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -176,6 +180,8 @@
 
     private static volatile Thread corfuServerThread;
     private static volatile Thread shutdownThread;
+    // Bind to all interfaces only if no address or interface specified by the user.
+    private static boolean bindToAllInterfaces = false;
```
no2chem: This shouldn't be a static var

zalokhan: Its accessed only by static methods.
I have to make it a class level variable as it is reused on restarts/resets.

no2chem: Isn't that passed around in the context anyway?

zalokhan:
@no2chem Now passing the flag in the context. Also needed to pass as a parameter to support restart reset cases.

runtime/src/main/java/org/corfudb/util/NetworkUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
+
+/**
+ * Network utility methods.
+ *
+ * <p>Created by zlokhandwala on 3/6/18.
+ */
+@Slf4j
+public class NetworkUtils {
```
corfu-repo: MAJOR Add a private constructor to hide the implicit public one. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1118)

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -199,6 +203,23 @@ public static void main(String[] args) {
 
         log.debug("Started with arguments: " + opts);
 
+        // Bind to all interfaces only if no address or interface specified by the user.
+        final boolean bindToAllInterfaces;
+        // Fetch the address if given a network interface.
+        if (opts.get("--network-interface") != null) {
+            opts.put("--address",
```
corfu-repo: CRITICAL Define a constant instead of duplicating this literal "--address" 4 times. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1192)

corfu-repo:
* ![CRITICAL][CRITICAL] 1 critical
* ![MAJOR][MAJOR] 1 major

[CRITICAL]: https://sonarsource.github.io/sonar-github/severity-critical.png 'Severity: CRITICAL'
[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

Maithem:
LGTM. Forgot to approve before I merged.

PR #1220, opened 2018-03-14
Using VM interval property to enable metrics

StrikingThirteen:
## Overview

Description:
This commit enables metrics collection and reporting based on the VM's metricsInterval. It also removes some of the obsolete code in MetricsUtils.

Why should this be merged: 
To enable collecting and reporting metrics.

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/MetricsUtils.java
+++ b/runtime/src/main/java/org/corfudb/util/MetricsUtils.java
@@ -12,16 +12,11 @@ import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
 import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
 import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
 import com.github.benmanes.caffeine.cache.Cache;
-
-import java.io.File;
-import java.io.FileInputStream;
-import java.util.Properties;
-import java.util.concurrent.TimeUnit;
-
-import org.slf4j.LoggerFactory;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.slf4j.LoggerFactory;
 
+import java.util.concurrent.TimeUnit;
 
 @Slf4j
 public class MetricsUtils {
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/util/MetricsUtils.java
```diff
@@ -31,43 +27,30 @@
     private static final MetricSet metricsJVMMem = new MemoryUsageGaugeSet();
     private static final MetricSet metricsJVMThread = new ThreadStatesGaugeSet();
 
-    private static Properties metricsProperties = new Properties();
+    private static int interval;
```
Maithem: why is this static ?

StrikingThirteen: One of the properties that similar to metricsCollectionEnabled, and so on being used by other static methods of this util class. Mostly replaces: (private static Properties metricsProperties = new Properties();)

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
```diff
@@ -31,43 +26,37 @@
...
-            } catch (Exception e) {
-                log.error("Error processing METRICS_PROPERTIES {}: {}", propPath,
-                        e.toString());
-            }
+    private static void loadVmProperties() {
+        try {
+            interval = Integer.valueOf(System.getProperty("metricsInterval"));
+        } catch (NumberFormatException e) {
+            // interval defaults to -1 as disabled when the value cannot be
+            // extracted from VM property
```
Maithem: log error

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
```diff
@@ -31,43 +26,37 @@
...
+            interval = -1;
+        }
+
+        if (interval > 0) {
```
Maithem: what's the difference between metrics collection/reporting?

I think an application would want collection without reporting, or collection and reporting, but a +/- interval only enables/disables both options. Maybe use 0 for collection, but not reporting, neg-interval would be to disable everything and a pos-number would mean a reporting interval.

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
```diff
@@ -85,22 +74,19 @@ public static boolean isMetricsReportingSetUp(MetricRegistry metrics) {
...
      * Start metrics reporting via the Dropwizard 'CsvReporter' file writer.
      * Reporting can be turned on and off via the properties file described
-     * in loadPropertiesFile()'s docs.  The report interval and report
+     * in loadVmProperties()'s docs.  The report interval and report
      * directory cannot be altered at runtime.
      *
      * @param metrics Metrics registry
      */
     public static void metricsReportingSetup(MetricRegistry metrics) {
         metrics.counter(mpTrigger);
```
Maithem: rename this, what is a mpTrigger ?

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
StrikingThirteen: Done.

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
StrikingThirteen: Done.
Collection without reporting might have limited value considering that dropwizard limited expressiveness at debugging at each point of time. However, it's a good idea to have that ability if application wants to have this ability for introducing a manual reporting. Added interval=0 for collection without reporting.

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
StrikingThirteen: mpTrigger was intended by the author for filtering out some of the metrics. The logic requires investigating & clean-up yet avoiding side effects. That's a task that I will do outside the scope of this enabling metrics PR.

Maithem:
@StrikingThirteen I see, please squash the history and lets get this in.

StrikingThirteen:
@Maithem Squashed.

Maithem:
@StrikingThirteen do we know why njection-framework  is failing ?

StrikingThirteen:
@Maithem It was stale. New pre-checkin is in progress

PR #1192, opened 2018-02-08
Stamp writes with expected epoch

zalokhan:
## Overview
Description: Stamp all epoch aware messages with epoch the user of the client expects it to be in.

Why should this be merged: Race in writes with epoch changes causes correctness issues.

Related issue(s): #1191 

## Checklist (Definition of Done):
- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #1229, opened 2018-03-20
Fix object creation performance regression #1192

zalokhan:
## Overview

Description:
Fixing performance. Instead of creating deep copies of the layout, reusing the layout reference in the RuntimeLayout constructor.
In the AbstractView, we now keep an atomic reference of the RuntimeLayout used and create new only if there is an epoch change. 

Why should this be merged: Fixes performance drop introduced by #1192 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -3,6 +3,8 @@ package org.corfudb.runtime.view;
 import java.time.Duration;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
+
 import javax.annotation.Nonnull;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
@@ -59,6 +61,11 @@ public abstract class AbstractView {
         return layoutHelper(function, false);
     }
 
+    /**
+     * Maintains reference of the RuntimeLayout which gets invalidated if the AbstractView
+     * encounters a newer layout. (with a greater epoch)
+     */
+    private AtomicReference<RuntimeLayout> runtimeLayout = new AtomicReference<>();
 
     /**
      * Helper function for view to retrieve layouts.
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/RuntimeLayout.java
```diff
@@ -49,7 +49,7 @@
      */
     public RuntimeLayout(@Nonnull Layout layout, @Nonnull CorfuRuntime corfuRuntime) {
 
-        this.layout = new Layout(layout);
+        this.layout = layout;
```
Maithem: I thought the whole point of a runtime layout is to have your own layout.

zalokhan: It still does right. It's just the management agent which modifies layouts. We can make sure all modifications are made on a copy.

no2chem: Yeah, this is fine. a runtime layout is a runtime+layout so you can easily get new clients.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -59,6 +61,7 @@ public Layout getCurrentLayout() {
         return layoutHelper(function, false);
     }
 
+    private volatile AtomicReference<RuntimeLayout> runtimeLayout = new AtomicReference<>();
```
no2chem: I'm pretty sure this doesn't need to be an AtomicReference.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -86,7 +89,13 @@ public Layout getCurrentLayout() {
         final Duration retryRate = runtime.getParameters().getConnectionRetryRate();
         while (true) {
             try {
-                return function.apply(new RuntimeLayout(runtime.layout.get(), runtime));
+                final Layout layout = runtime.layout.get();
```
no2chem: Place proper synchronization around this so only a single retrieval is permitted, use a stamped lock so that optimistic reads are possible.

zalokhan: I used the atomic reference to provide synchronization. So that single retrieval is permitted.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **a85bc12**.

 [Pull Request #1229 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/nVoMByk2S4OWaaZPA26XLOwmDhXbKKhP)

runtime/src/main/java/org/corfudb/runtime/view/RuntimeLayout.java
medhavidhawan: change the old comment.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
medhavidhawan: remove volatile

zalokhan:
@no2chem @medhavidhawan I have addressed the comments.
Michael, we agreed on AtomicReference.
I have confirmed with Anny that the performance numbers have improved matching the commits before the regression.

PR #1235, opened 2018-03-21
Layout migration

zalokhan:
## Overview

Description: Introduction of network interface binding by #1209 allows the user to bind to a specific IP address. To support migration from version 0.1.1, we should transform the layout configuration files (LAYOUT_CURRENT and MANAGEMENT_LAYOUT) so that they replace the old ip:port with the new binding ip:port.

Why should this be merged: To allow the user to migrate from 0.1.1 to the latest build.

- [x] Public API has Javadoc

### DIFF
--- a/migration/src/main/java/org/corfudb/migration/LogFormat1to2.java
+++ b/migration/src/main/java/org/corfudb/migration/LogFormat1to2.java
@@ -23,7 +23,7 @@ import static com.google.common.collect.Sets.newHashSet;
 
 /**
  * This migration tool will migrate the local data store files and the log segment files
- * fro, version 1 to 2. This migration will include remove an extra delimiter (i.e. redundant magic byte)
+ * from version 1 to 2. This migration will include remove an extra delimiter (i.e. redundant magic byte)
  * and add a new field to the metadata type. As a result, checksums will be recomputed.
  *
  * To run this tool, execute the following steps:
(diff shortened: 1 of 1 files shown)


migration/src/main/java/org/corfudb/migration/LogFormat1to2.java
```diff
@@ -0,0 +1,303 @@
...
+import java.nio.file.StandardOpenOption;
+import java.util.EnumSet;
+import java.util.Set;
+
+import org.corfudb.format.Types;
+import org.corfudb.runtime.view.Layout;
+
+/**
+ * This migration tool will migrate the local data store files and the log segment files
+ * fro, version 1 to 2. This migration will include remove an extra delimiter (i.e. redundant magic byte)
```
medhavidhawan: r/fro/from

Maithem:
Im not sure this belongs here. Migration usually refers to data format changes, not "config" changes and this PR just replaces ips. Changing layout addresses after creating the layout seems like an orthogonal issue.

migration/src/main/java/org/corfudb/migration/LogFormat1to2.java
```diff
@@ -83,15 +104,37 @@ public boolean accept(File dir, String name) {
...
 
-    public static void migrateDsFile(String path) throws IOException {
+    public static void migrateDsFile(String path, String oldEndpoint, String newEndpoint)
+            throws IOException {
         Path srcPath = Paths.get(path);
         Path destPath = Paths.get(path + ".ds");
 
         byte[] bytes = Files.readAllBytes(srcPath);
+
+        if (layoutStateRecoveryFiles.stream().anyMatch(path::contains)) {
```
Maithem: This method processes files based on certian prefixes, if you want to specifically process particular files you can create another method. Also, make sure you match the suffix and not "any" on the full file patch.

Why cant you use replaceall on the json string instead of all of this?

zalokhan:
You are right, this is a config modifying script. Where should this ideally be?
There would be a strict order on how to run these 2 scripts then. First, migrate and then modify the new config files or vice versa.

migration/src/main/java/org/corfudb/migration/LogFormat1to2.java
zalokhan: I'll make the matching suffix change. 
I considered replaceAll at first but thought this would be a cleaner way. I can change it back.

Maithem: hmm, isn't that just one line instead of iterating the different types?

Also, that would remove the dependency on the runtime. This module already has a proto file, so depending on runtime would also make it depend on the format/proto file ... The two proto files are not the same.

Maithem:
@zalokhan Maybe create a layout utils ? you need to do this independently from the migration tool and running the migration tool every time you want to change the ip addresses doesn't seem right.

Also, what is the use case here? it seems like this can lead to inconsistencies in cluster mode.

migration/src/main/java/org/corfudb/migration/LogFormat1to2.java
zalokhan: Done.

migration/src/main/java/org/corfudb/migration/LogFormat1to2.java
zalokhan: Yes, true.
I've made the change.

zalokhan:
The use case is that a cluster running with version 0.1.1, binds to localhost:9000 yet advertises private-IP:9000. So now we face issues in upgrades. The server binds to the eth0 interface with the newly introduced flag but contains localhost in the layout. So the user would want to change this advertised IP address to the private binding (to which the server binds) IP.

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **784594d**.

 [Pull Request #1235 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/5u16ZSB5JU6KLB8AjsYPTbTttivc7g2e)

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **92b5542**.

 [Pull Request #1235 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/Jye69w3Q8P5kny5Knas0zTH6a0DshbSW)

Issue #1207, opened 2018-03-01
Redundant Sequencer Calls


Maithem:
## Overview

**Bug report.** 

calling next() on the stream always results in a sequencer call even if the next address is already resolved.
The following example illustrates this issue(both next() calls will emit a sequencer call where they shouldn't)

`@Test
    public void seqCalls() throws Exception {
        CorfuRuntime rt = new CorfuRuntime("localhost:9000").connect();

        final int iter = 50;
        final int payload = 100;
        for (int x = 0; x < iter; x++) {
            rt.getStreamsView().get(CorfuRuntime.getStreamID("s1")).append(new byte[payload]);
        }
       
        IStreamView s1 = rt.getStreamsView().get(CorfuRuntime.getStreamID("s1"));

        s1.seek(100);
        s1.next();
        s1.previous();
        s1.previous();
        s1.next();
    }`

PR #1247, opened 2018-04-03
Optimisitic TX RPC Call out of the lock

annym:
To avoid waiting for RPC calls to be executed under the VLO lock,
snapshot timestamp can be obtained right before acquiring the lock.

## Overview

Description: Get Snapshot Timestamp before the lock is acquired. This typically represents a single RPC call per transaction, therefore, the performance gains are higher for smaller transactions. 

Why should this be merged: depending on the workload pattern we have found performance improvements of up to 44%. Performance outperforms as the transaction size decreases and the number of threads in the client increases. This percentage (44%) was found for the case of Single Write Transactions and 40 client threads. We have also tested this thoroughly on Rishi's System and not found any regression. 

Related issue(s) (if applicable): #1207

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -196,7 +196,7 @@ public class VersionLockedObject<T> {
                     }
                 }
             } catch (Exception e) {
-                // If we have an exception, we didn't get a chance to validate the the lock.
+                // If we have an exception, we didn't get a chance to validate the lock.
                 // If it's still valid, then we should re-throw the exception since it was
                 // on a correct view of the object.
                 if (lock.validate(ts)) {
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
@@ -93,6 +93,8 @@ public class OptimisticTransactionalContext extends AbstractTransactionalContext
         // Next, we sync the object, which will bring the object
         // to the correct version, reflecting any optimistic
         // updates.
+        // Get snapshot timestamp in advance so it is not performed under the VLO lock
+        long ts = getSnapshotTimestamp();
         return proxy
                 .getUnderlyingObject()
                 .access(o -> {
@@ -101,7 +103,7 @@ public class OptimisticTransactionalContext extends AbstractTransactionalContext
                             // Obtain the stream position as when transaction context last
                             // remembered it.
                             long streamReadPosition = getKnownStreamPosition()
-                                    .getOrDefault(proxy.getStreamID(), getSnapshotTimestamp());
+                                    .getOrDefault(proxy.getStreamID(), ts);
 
                             return (
                                     (stream == null || stream.isStreamCurrentContextThreadCurrentContext())
@@ -152,9 +154,11 @@ public class OptimisticTransactionalContext extends AbstractTransactionalContext
             return wrapper.getUpcallResult();
         }
         // Otherwise, we need to sync the object
+        // Get snapshot timestamp in advance so it is not performed under the VLO lock
+        long ts = getSnapshotTimestamp();
         return proxy.getUnderlyingObject().update(o -> {
             log.trace("Upcall[{}] {} Sync'd", this,  timestamp);
-            syncWithRetryUnsafe(o, getSnapshotTimestamp(), proxy, this::setAsOptimisticStream);
+            syncWithRetryUnsafe(o, ts, proxy, this::setAsOptimisticStream);
             SMREntry wrapper2 = getWriteSetEntryList(proxy.getStreamID()).get((int)timestamp);
             if (wrapper2 != null && wrapper2.isHaveUpcallResult()) {
                 return wrapper2.getUpcallResult();


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -211,6 +211,15 @@ public VersionLockedObject(Supplier<T> newObjectFn,
         // Next, we just upgrade to a full write lock if the optimistic
         // read fails, since it means that the state of the object was
         // updated.
+        // Note (to remove): This can only tell us it was changed by the current client, even in the case that
```
no2chem: This TODO can be removed. I think the main risk is getting the timestamp too early, but that's fine, it will just slightly increase the abort rate.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -93,6 +93,8 @@
         // Next, we sync the object, which will bring the object
         // to the correct version, reflecting any optimistic
         // updates.
+        // Get snapshot timestamp in advance so it is not performed under the VLO lock
+        long ts = getSnapshotTimestamp();
```
Maithem: hmm, can't you just remove the lazy evaluation here
https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java#L105 ?

Maithem:
@no2chem I think the perf tests uses default transactions (i.e. OptimisticTransaction) which doesn't evaluate the timesnapshot in the constructor.

https://github.com/CorfuDB/CorfuDB/blob/44816933e17d7e7c8de8ebaf580380e2f49a05bd/runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java#L64

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
annym: Well actually now that I see @Maithem comment, this is not an actual concern because the snapshotTimestamp has a lazy getter so even though the updateFunction is applied this timestamp is already computed, in this case are we really able to sync?? if the timestamp should not have been updated?

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
annym: If we remove the lazy getter we would be having a second RPC call in the access function whenever an update function is applied, but still if you see my previous comment I'm wondering if it ain't the actual right thing to do in order to sync the object..

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **e1e8c41**.

 [Pull Request #1247 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/TD6oDAstmrJIY6IhK4tfzD1eZQEmWj3Y)

no2chem:
Ok, summary of discussion on this PR:
- 40% performance increase on small transactions which use read-write transactions
- no significant effect on write-write transactions

@annym -- overall great find -- could you remove the todo, then we can merge this?

annym:
@no2chem Done!

corfudb-performance:
Results automatically generated by _CorfuDB Benchmark Framework_ to assess the performance of this pull request for commit **108e039**.

 [Pull Request #1247 Graphs ](https://snapshot.raintank.io/dashboard/snapshot/MBX4RFxFcf8zW6IMO4ooCaMryaOtv2qZ)

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
no2chem: If you removed the lazy evaluation, the value would no longer be caches, and therefore result in a rpc on _every_ object access

no2chem: Cached* not to mention it would break correctness

Maithem: I don't see how this is related to caching, lombok's lazy eval dictates when a class field is initialized (i.e. construction vs first access time initialization). If you delombok the code you'll see this:

private final long snapshotTimestamp = obtainSnapshotTimestamp();

public long getSnapshotTimestamp() {
        return this.snapshotTimestamp;
} 

In other words, the transaction snapshot would be evaluate once the class is created and then subsequent calls to getSnapshotTimestamp wouldn't actually result in an rpc, which is what we wanted to do, or am I missing something?

annym: I see your point, you say moving it out of the lock is equivalent to removing the lazy evaluation and removing the lombok getter annotation. I agree, however, wouldn't this increase our chances of aborted transactions? as it might be computed way in advance to the access (according to any client business logic running within a TX). 

On the other hand, either solution we adopt which both attempt to do a single RPC, would you agree that the syncWithRetryUnsafe is not really doing anything? we are passing the same snapshotTimestamp, so no way to sync back or forward.

Maithem: Yes, I think they are equivalent and "cleaner". True, that can happen, but I think its also fine to assume transactional semantics right after tx.begin is called as opposed to first read/write. 

Why not? Notice that there are two time snapshots, one for the tx and the other that belongs to the VLO's stream. Consider the case where you have a stream at version 10 and you have a transaction that starts at version 5, then the direct access will evaluate to false since o.getVersionUnsafe() != streamReadPosition , then the update function is going to be applied, which will call syncWithRetryUnsafe(o (stream VLO ts 10) , 5 (ts of tx), .....) and that will call syncObjectUnsafe which will keep rolling back the smr stream from 10 (vlo stream version) to 5 (tx ts). Does this seem right or did i misunderstand your point?

annym: Regarding the first point,I'm fine with this if we all agree this is the right semantic for an Optimistic Tx, I'll submit a PR with this change.

Thanks for the explanation, this example does give me a use case in which I see it.

The scenario I was thinking of was this. Let's assume you have a TX that starts at V5 and the stream is also at v5, since the direct access function is evaluated within the lock, suppose we are not able to return because we cannot validate the lock (optimistic read), the only reason why we wouldn't be able to do so is because the object was modified (by this client), so upgrading to a writelock and failing to do the direct access again would lead to syncWithRetryUnsafe which I thought would attempt to bring the object to the latest version (let's say v6 as it was updated by another thread and we are aware of it). But know that I see your example, I imagine in this case we just rollback to v5 and eventually this TX will be aborted...

no2chem: The reason we obtain tx snapshot lazily for R-W transactions is that a write-only r-w transaction  doesn't need a snapshot, getting a snapshot would be a wasted RPC to the sequencer. We should leave as is.

Maithem: I was just responding to an earlier comment before the merge, I think its fine for now.

For your example, let's say T1 tries to read V5 and T2 move the object to V6, then:
T1: start optimistic read
T2: acquire write lock and apply the update and relinquish the lock.
T1: apply direct access function and fail the lock validation. 
T1: acquire the write lock and apply the update function (i.e. syncWithRetryUnsafe -> rollbackStreamUnsafe)
T1: apply the access function (after lock upgrade)
 
Notice after the failed validation (with direct access function), it will apply the update function which synchronizes the versions (i.e. roll back from V6 to V5). I think the key point here is that optimistic reading path never changes the objects state/version.

Maithem: @no2chem I see, but that's not happening in the current code base, since transactions need TxResolutionInfo, which depend on getSnapshotTimestamp.

Issue #1262, opened 2018-04-11
Implement the option to export metrics to CSV files 


StrikingThirteen:
## Overview
In order to facilitate the analysis of the collected metrics, implement the option to export metrics as csv files.

StrikingThirteen:
The corresponding patch is merged. closing the issue.

PR #1267, opened 2018-04-16
Implements possibility of exporting corfu metrics to CSV files.

StrikingThirteen:
## Overview

Description:
This commit, implements the option of reporting corfu metrics to CSV files.

Why should this be merged: 
Creating CSV files for different metrics facilitates users' investigation of Corfu performance by analyzing and aggregating different metrics offline. Similarly developers can assess the performance implication of introduced changes by crunching results capturing different Corfu performance indicators along side each other.

Related issue(s) (if applicable): #1262 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/MetricsUtils.java
+++ b/runtime/src/main/java/org/corfudb/util/MetricsUtils.java
@@ -1,6 +1,7 @@
 package org.corfudb.util;
 
 import com.codahale.metrics.Counter;
+import com.codahale.metrics.CsvReporter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.JmxReporter;
 import com.codahale.metrics.MetricFilter;
@@ -18,6 +19,7 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.util.concurrent.TimeUnit;
 
 @Slf4j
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/util/MetricsUtils.java
```diff
@@ -139,6 +160,29 @@ private static void setupSlf4jReporting(MetricRegistry metrics) {
...
+        if (!metricsCsvReportingEnabled) return;
+
+        final File directory = new File(metricsCsvFolder);
+
+        if (!directory.isDirectory() ||
+            !directory.exists() ||
+            !directory.canWrite()) {
+            log.warn("Provided CSV directory : {} doesn't exist, isn't a directory, or not accessible " +
+                    "for writes. Disabling CSV Reporting.");
+            metricsCsvReportingEnabled = false;
```
Maithem: Why fail silently here? if this is enabled then its expected to work no ?

StrikingThirteen: Logging the failure to collect CSV metrics due to non existing directory or lack of write privilege seems to be reasonable response to the failure. It is probably excessive to consider it as a show stopper.

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
```diff
@@ -139,6 +160,29 @@ private static void setupSlf4jReporting(MetricRegistry metrics) {
...
+            return;
+        }
+
+        CsvReporter csvReporter = CsvReporter.forRegistry(metrics)
+                .convertDurationsTo(TimeUnit.MICROSECONDS)
+                .convertRatesTo(TimeUnit.SECONDS)
+                .filter(ADDRESS_SPACE_FILTER)
+                .build(directory);
```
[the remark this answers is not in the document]
StrikingThirteen: It creates a CSV per metric in the provided folder given its flag is on. They are geared towards easy trouble shooting and are generated when the CSV flag is on. Rather than scraping the logs, this provides enables technical and non-technical users to simply open the CSV files in excel or similar tools and analyze the metrics and correlations between them.

PR #1278, opened 2018-04-27
Stream View Layer Simplification

annym:
## Overview

Description:

Stream View Layer redesign for code simplification. The notion of a
StreamAddressSpace is introduced as a means to represent the space of
global addresses containing log entries/updates for the corresponding
stream. This new design is compatible with the idea of disabling the
checkpoint notion.

Why should this be merged: Design simplification and implications on future performance improvements in discussion. 

- [ ] Public API has Javadoc


Maithem:
Does this fix the stream reset problem, where interleaving access to different versions can reset the stream, which results in a rebuild?

runtime/src/main/java/org/corfudb/runtime/view/stream/CompositeStreamAddressSpace.java
```diff
@@ -0,0 +1,338 @@
...
+    public void reset() {
+        regularSAS.reset();
+        cpSAS.reset();
+        pointerOnRegularStream = true;
+    }
+
+    @Override
+    public void seek(long address) {
+        // Seek for the address on the regular stream, if it is not available it means
+        // it has been checkpoint(ed) and trimmed.
```
Maithem: I think this changes the behavior of seek. Currently, on master, when you seek to an address that doesn't belong to the stream, the stream layer will choose the "closest" (i.e. greatest stream address less than the seek address) address to the non-stream address and not throw an Exception.

There's code in mp that queries the global tail and does stream.seek(globalTail) in that case if the stream doesn't have the global tail then it will break.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -0,0 +1,288 @@
...
+    public void seek(long address) {
+
+        int tmpPtr = addresses.indexOf(address);
+
+        if (tmpPtr != -1) {
+            currInd = tmpPtr;
+            return;
+        }
+
+        throw new RuntimeException("Couldn't seek " + streamId + " to address " + address);
```
Maithem: probably need to make this more specific, maybe call it NonSeekableAddressException ?

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamAddressSpace.java
```diff
@@ -0,0 +1,169 @@
...
+     */
+    long lower(long globalAddress, boolean forward);
+
+    /**
+     * Defines how the space of addresses is traversed to find new addresses in the range between
+     * newTail and oldTail (non-exclusive).
+     *
+     * @return the number of addresses found for this stream.
+     */
+    int findAddresses(long oldTail, long newTail, Function<Long, ILogData> readFn);
```
Maithem: I don't think this is being used, can we remove it ?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -0,0 +1,288 @@
...
+ * The abstract method 'findAddresses' defines how to traverse the space of unexplored addresses,
+ * i.e, addresses between the latest synced point (oldTail) and the latest update (newTail).
+ * For example, the implementation of BackpointerAddressSpace assumes the use of backpointers
+ * to determine the previous log entry for a stream.
+ *
+ * Created by amartinezman on 4/23/18.
+ */
+public abstract class StreamAddressSpace implements IStreamAddressSpace {
+
+    protected CopyOnWriteArrayList<Long> addresses;
```
Maithem: Why is this needed?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -0,0 +1,288 @@
...
+    protected int maxInd;
+    protected int currInd;
+    protected StreamOptions options;
+
+    public StreamAddressSpace(UUID id, CorfuRuntime runtime) {
+        this.streamId = id;
+        this.runtime = runtime;
+        this.addresses = new CopyOnWriteArrayList();
+        this.currInd = -1;
+        this.maxInd = -1;
```
Maithem: maybe create an uninitialized constant for all the -1s ?

runtime/src/main/java/org/corfudb/runtime/view/stream/CompositeStreamAddressSpace.java
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return this.regularSAS.isEmpty() && this.cpSAS.isEmpty();
+    }
+
+    @Override
+    public void setPointerToPosition(long address) {
+        this.regularSAS.setPointerToPosition(address);
```
Maithem: It looks like this method is used to advance the stream pointer, which is usually close to the tail, but calling indexOf on addresses will do a linear scan from the beginning of the array. Maybe you want to do it backwards instead if the array is large.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -308,20 +311,26 @@ private boolean entryAlreadyContainedInCheckpoint(UUID streamId, SMREntry entry)
      */
     private boolean shouldEntryBeApplied(UUID streamId, SMREntry entry, boolean isCheckpointEntry) {
         // 1.
+        // If the entry is null
+        if (entry == null) {
+            return false;
```
Maithem: Why didn't we have to account for this before?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -325,20 +326,23 @@ public void syncObjectUnsafe(long timestamp) {
...
-                    log.warn("Rollback[{}] to {} failed {}", this, timestamp, nre);
-                    resetUnsafe();
-                }
-            }
+            this.rollBackToTimestamp(timestamp);
+            // Attempt to sync stream to timestamp, it might have to sync to the tail to build the
+            // history
             syncStreamUnsafe(smrStream, timestamp);
+            // Check if stream was synced ahead and roll back to timestamp if necessary
+            this.rollBackToTimestamp(timestamp);
```
Maithem: the object is rolled back, then synced forward. Why do you need an extra roll back after the sync ?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -325,20 +326,23 @@ public void syncObjectUnsafe(long timestamp) {
...
+        }
+    }
+
+    private void rollBackToTimestamp(long timestamp) {
```
Maithem: this can also be used in line 304

runtime/src/main/java/org/corfudb/runtime/view/stream/SingleStreamView.java
```diff
@@ -0,0 +1,291 @@
...
+@Slf4j
+public class SingleStreamView implements IStreamView {
+
+    final private UUID id;
+    final private CorfuRuntime runtime;
+    private IStreamAddressSpace sas;
+    final StreamOptions options;
+
+    public SingleStreamView(UUID id, CorfuRuntime runtime,
+                            BiFunction<UUID, CorfuRuntime, IStreamAddressSpace> streamAddressSpace,
```
Maithem: I feel like its more natural to have IStreamAddressSpace as the second parameter as opposed to passing a BiFunction

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamAddressSpace.java
```diff
@@ -0,0 +1,63 @@
...
+import org.corfudb.runtime.exceptions.TrimmedException;
+
+/**
+ * The BackpointerStreamAddressSpace follows backpointers in order to
+ * traverse a stream in a given range (defined by the space between oldTail and newTail).
+ *
+ * Created by amartinezman on 4/18/18.
+ */
+@Slf4j
+public class BackpointerStreamAddressSpace extends StreamAddressSpace {
```
Maithem: why not push this up to StreamAddressSpace as you already consume IStreamAddressSpace in other places.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamAddressSpace.java
```diff
@@ -0,0 +1,63 @@
...
+                    break;
+                } else {
+                    throw te;
+                }
+            }
+        }
+        
+        List<Long> revList = Lists.reverse(addressesToAdd);
+        this.addAddresses(revList);
+        return revList.size();
```
Maithem: It might be cleaner to actually return revList and the caller can use it to add the addressed to addAddresses.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -0,0 +1,288 @@
...
+            res.add(addresses.get(currInd + 1));
+            currInd++;
+        }
+        return res;
+    }
+
+    @Override
+    public void addAddresses(List<Long> addresses) {
+        this.addresses.addAll(addresses);
+        Collections.sort(this.addresses);
```
Maithem: I believe this is unnecessary expensive, why do you need to sort every time you resolve a new delta? 

Also, can you add a sanity check to make sure the boundaries are correct (addresses[0] > this.addresses[last ind])
and maybe during tail traversal you can make sure that the backpointer is always decreasing.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -0,0 +1,288 @@
...
+            return addresses.get(currInd);
+        }
+    }
+
+    @Override
+    public boolean hasNext() {
+        if (currInd < maxInd) {
+            return true;
+        } else {
+            return false;
```
Maithem: Would this return the true if currInd == maxInd and the tail is greater than maxInd ?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    @Override
+    public void removeAddresses(long upperBound) {
+        int removedAddresses = 0;
+        int index = this.addresses.indexOf(upperBound);
+        if (index != -1) {
+            for (int i = index; i >= 0; i--){
+                removedAddresses ++;
+                this.addresses.remove(i);
```
Maithem: These removes are O(n) it might be expensive. I'm not sure if a new array allocation would be cheaper.
All you need to do is find the index of upperBound and then slice the array from index + 1 to this.addresses.size()

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -0,0 +1,288 @@
...
+                currInd = currInd - removedAddresses;
+            }
+
+            maxInd = this.addresses.size() - 1;
+        }
+    }
+
+    @Override
+    public void syncUpTo(long globalAddress, long newTail, long lowerBound,
+                         Function<Long, ILogData> readFn) {
```
Maithem: why do you need readFn ? AddressSpaceView.read already defines the reading protocol.

runtime/src/main/java/org/corfudb/runtime/object/ISMRStream.java
```diff
@@ -74,4 +74,6 @@ long append(SMREntry entry,
     default UUID getID() {
         return new UUID(0L, 0L);
     }
+
+//    public boolean isSnapshotAvailable(long timestamp);
```
no2chem: Did you intend to remove this from the interface?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
no2chem: It would be useful to have a description of this function
It also should probably be suffixed with -Unsafe, since it calls unsafe functions without a lock

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -601,10 +606,11 @@ protected void rollbackStreamUnsafe(ISMRStream stream, long rollbackVersion) {
      * @param stream    The stream to sync forward
      * @param timestamp The timestamp to sync up to.
      */
-    protected void syncStreamUnsafe(ISMRStream stream, long timestamp) {
+    protected void  syncStreamUnsafe(ISMRStream stream, long timestamp) {
```
no2chem: spacing

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamAddressSpace.java
```diff
@@ -0,0 +1,63 @@
...
+@Slf4j
+public class BackpointerStreamAddressSpace extends StreamAddressSpace {
+
+    public BackpointerStreamAddressSpace(UUID id, CorfuRuntime runtime) {
+        super(id, runtime);
+    }
+
+    @Override
+    public int findAddresses(long oldTail, long newTail, Function<Long, ILogData> readFn) {
+        List<Long> addressesToAdd = new ArrayList<>(100);
```
no2chem: Is there a reason this is initially 100? It should be at least defined in a constant somewhere.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamAddressSpace.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Slf4j
+public class BackpointerStreamAddressSpace extends StreamAddressSpace {
+
+    public BackpointerStreamAddressSpace(UUID id, CorfuRuntime runtime) {
+        super(id, runtime);
+    }
+
+    @Override
+    public int findAddresses(long oldTail, long newTail, Function<Long, ILogData> readFn) {
```
no2chem: Needs a bit of a high-level description of the function here

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamAddressSpace.java
```diff
@@ -0,0 +1,63 @@
...
+                } else {
+                    // When a hole is encountered we downgrade to single step
+                    newTail = oldTail - 1;
+                }
+            } catch (TrimmedException te) {
+                if (options.ignoreTrimmed) {
+                    log.warn("followBackpointers: Ignoring trimmed exception for address[{}]," +
+                            " stream[{}]", newTail, streamId);
+                    this.removeAddresses(newTail);
+                    break;
```
no2chem: Hm, why do you break here? Shouldn't we be ignoring the trim exception?

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamAddressSpace.java
```diff
@@ -0,0 +1,63 @@
...
+                } else {
+                    throw te;
+                }
+            }
+        }
+        
+        List<Long> revList = Lists.reverse(addressesToAdd);
```
no2chem: Why do you need to reverse this list? addAddresses doesn't seem to have any ordering requirement.

runtime/src/main/java/org/corfudb/runtime/view/stream/CheckpointStreamAddressSpace.java
```diff
@@ -0,0 +1,175 @@
...
+ * The Checkpoint Stream Address Space resolves the space of addresses of a checkpoint(ed) stream,
+ * therefore it is capable of following backpointers and resolving the checkpoint at the same time,
+ * i.e., determine how to traverse the cp stream
+ *
+ * Created by amartinezman on 4/18/18.
+ */
+public class CheckpointStreamAddressSpace extends StreamAddressSpace {
+
+    /** Info on checkpoint we used for initial stream replay,
+     *  other checkpoint-related info & stats.  Hodgepodge, clarify.
```
no2chem: be sure to clarify this in the final rev

runtime/src/main/java/org/corfudb/runtime/view/stream/CheckpointStreamAddressSpace.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Override
+    public long getCurrentPointer() {
+        // The current pointer refers to the current version of the object, which is in fact related
+        // to the last global address synced for this checkpoint stream.
+        return getLastAddressSynced();
+    }
+
+    @Override
+    public int findAddresses(long oldTail, long newTail, Function<Long, ILogData> readFn) {
```
no2chem: This seems to be somewhat redundant to the code in BackpointerStreamAddressSpace::findAddresses (?). Can we make sure that we reuse the code? (This BackpointerOp logic used to handle checkpoint/noncheckpoint streams).

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamAddressSpace.java
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Seeks for a specific address, and moves the pointer in the stream to this position.
+     *
+     */
+    void seek(long address);
+
+    /**
+     * Get maximum resolved address.
+     *
+     * @return maximum address for this stream (address is global)
```
no2chem: So, all addresses are global now, there's no need to distinguish.

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamAddressSpace.java
```diff
@@ -0,0 +1,169 @@
...
+     * @return addresses in the stream up to a limit (addresses are global)
+     */
+    List<Long> remainingUpTo(long limit);
+
+    /**
+     * Add list of addresses to the space of addresses of this stream.
+     *
+     * @param addresses global addresses to be added.
+     */
+    void addAddresses(List<Long> addresses);
```
no2chem: If you want to use the List<> collection, I would recommend at least annotating this parameter @NonNull.

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamAddressSpace.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * Determines if there is a valid address in the stream while traversing forward
+     * from the current pointer.
+     *
+     * @return true if exists, false otherwise.
+     */
+    boolean hasNext
+
+    ();
```
no2chem: on previous line

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamAddressSpace.java
```diff
@@ -0,0 +1,169 @@
...
+     * @return true if empty. false, otherwise.
+     */
+    boolean isEmpty();
+
+    /**
+     * Set Pointer of Stream to the position of 'address'
+     *
+     * @param address address where the stream space should be pointing to.
+     */
+    void setPointerToPosition(long address);
```
no2chem: Is this somehow different then 'seek'?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
no2chem: Moreover, shouldn't a collection with range query abilities (say, SkipList) be used?

Maithem: Also, it seems like inserts are expensive on this container type (i.e. O(n) inserts):

List                 | Add  | Remove | Get  | Contains | Next | Data Structure
---------------------|------|--------|------|----------|------|---------------
ArrayList            | O(1) |  O(n)  | O(1) |   O(n)   | O(1) | Array
LinkedList           | O(1) |  O(1)  | O(n) |   O(n)   | O(1) | Linked List
CopyOnWriteArrayList | O(n) |  O(n)  | O(1) |   O(n)   | O(1) | Array

Maithem:
subsumed by the latest ##1818

PR #1279, opened 2018-05-02
Complete unmapping secondary indices

zalokhan:
## Overview

Description: Unmap all secondary indexes.

Why should this be merged: This checks and returns if the slot is null and does not continue to other indexes in the indexSpec.

- [x] Public API has Javadoc


zalokhan:
@Maithem I cannot come up with a negative test case to prove the requirement of this change.
This code path is safe to remove making the following assumption that no indices are added dynamically after the creation of the object.
Should I keep the code path to support dynamic index registration in the future with a comment describing the same?

Maithem:
Your call. I would prefer that we remove it and possibly reintroduce it when we want to support dynamic addition of indices.

zalokhan:
#1290 addresses this issue.

PR #1256, opened 2018-04-10
junit microbenchmark framework

no2chem:
## Overview

Description: This PR introduces a microbenchmarking framework which can be invoked from within junit. It uses local netty channels, so should simulate closely the performance of netty without the full network stack.

The changes for the local netty channels are taking from the testFramework2 branch. They are changes which enable a new simulated local protocol, which could be reused in the test framework for more accurate unit tests.

This provides an easy way to invoke the microbenchmark from within the IDE, just as with any junit test. A simple benchmark, SMRMap benchmark is provided which illustrates its use. Overall, this should greatly simplify the task of writing and running accurate microbenchmarks.

Sample output:
```
SMRMapBenchmark.getThenInsert                            thrpt   10  3.825 ± 0.657  ops/ms
SMRMapBenchmark.getThenInsertMultithreaded               thrpt   10  0.807 ± 0.216  ops/ms
SMRMapBenchmark.getThenInsertTransactional               thrpt   10  3.731 ± 0.139  ops/ms
SMRMapBenchmark.getThenInsertTransactionalMultithreaded  thrpt   10  9.427 ± 2.771  ops/ms
SMRMapBenchmark.insert                                   thrpt   10  5.012 ± 0.828  ops/ms
SMRMapBenchmark.insertMultiThreaded                      thrpt   10  0.259 ± 0.070  ops/ms
```

Why should this be merged: Allows for microbenchmarking from within the unit test framework.

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

Issue #989, opened 2017-11-09
Too Many Threads


Maithem:
The server is creating too many handler threads to service requests. ForkJoinPool uses the number of cpus to set the parallelisim, but that number doesn't seem to be strictly enforced (because of the fork join pool implementation) . This can be illustrated by setting the parallelisim to 4 and creating 20 writer threads. The server will create a new thread per write request and ForkJoin pool will spawn more than 4 threads at a time. 

@no2chem since all write operations are blocking, do we really want this behavior: "unbounded" growth of threads?

no2chem:
1) When is this a problem
2) Is the # of threads really unbounded? A FJ pool will only create a new thread if there aren't any threads available (probably the available threads are waiting for I/O).

Maithem:
1) Possible starvation issues, fundamentally incoming requests can affect prior requests. Shouldn't the server reject requests that it cannot service?

2) From the implementation, I can see that there is an upper bound, but that bound is not the parallelism number that is passed to the constructor. A thread per request seems to be expensive. Also, in a FJ pool a new thread is created for every request (it seems), which seems expensive. In the profiler, I can see that the number of threads is almost equal to the number of concurrent client threads.

no2chem:
1. Starvation issues? You mean running out of threads? Can you actually show some evidence for your speculation?  You would need 32k threads total to hit the limit, but perhaps you are referring to a memory issue? Again, you need some evidence for this claim.

Also, *fundamentally* previous requests will affect future requests. This is what happens when we have limited resources like I/O, no?

2. You seem to think threads are expensive. Can you ground your intuition in some evidence?

Maithem:
tl;dr based on the latency/throughput stats that I have collected it seems like the forkjoin pool is slower than fixed sized thread pools, and my claim is the FJ pool is slower because it creates more threads than the parallelism parameter (which is set to the # of CPUs) and it because it keeps recreating threads.

Its more of an observation based on my perf analysis than pure speculation. I ran the experiment 3 days ago, but I just re-ran it to gather some numbers for you. In this test I run both the client and the server (in-memory) on the same machine. I ran two tests. One on a server that has a forkjoin pool and the other on a fixed sized pool. In each test I start with 1 writer and double the number of threads until it reaches 64 writer threads.

The data below shows data concerning latency for a write request.

Forkjoin, 1 writer thread

#[Mean    =         47.8, StdDeviation   =        107.6]
[...]

PR #1281, opened 2018-05-05
Seperate Sequencer Threads

Maithem:
## Overview
Allow different servers to use their own thread pools, with that, the Sequencer requests are now serviced on a different thread pool.

Why should this be merged: This allows us to dedicate threads for particular servers without having the negative effects of sharing same thread pool. Having granular control over each server threading will also pave the way for future optimizations. 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
@@ -6,6 +6,9 @@ import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.CorfuMsg;
 
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
 /**
  * Created by mwei on 12/4/15.
  */
@@ -15,6 +18,11 @@ public abstract class AbstractServer {
     @Getter
     @Setter
     volatile boolean shutdown;
+
+    static final ExecutorService sharedExecutor = Executors
+            .newFixedThreadPool(BatchWriter.BATCH_SIZE + Runtime.getRuntime().availableProcessors(),
+                    new ServerThreadFactory("SharedServerThread-",
+                            new ServerThreadFactory.ExceptionHandler()));
 
     public AbstractServer() {
         shutdown = false;
@@ -41,6 +49,10 @@ public abstract class AbstractServer {
         if (!getHandler().handle(msg, ctx, r)) {
             log.warn("Received unhandled message type {}", msg.getMsgType());
         }
+    }
+
+    public ExecutorService getExecutor() {
+        return sharedExecutor;
     }
 
     /**
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
@@ -7,12 +7,6 @@ import io.netty.channel.ChannelInboundHandlerAdapter;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.ForkJoinWorkerThread;
-import java.util.concurrent.atomic.AtomicInteger;
-
-import javax.annotation.Nonnull;
 
 import lombok.Getter;
 import lombok.Setter;
@@ -33,53 +27,6 @@ import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
 public class NettyServerRouter extends ChannelInboundHandlerAdapter
         implements IServerRouter {
 
-    public static class ServerThreadFactory
-            implements ForkJoinPool.ForkJoinWorkerThreadFactory {
-
-        public static final String THREAD_PREFIX = "ServerRouter-";
-        final AtomicInteger threadNumber = new AtomicInteger(0);
-
-        public static class ServerWorkerThread extends ForkJoinWorkerThread {
-
-            protected ServerWorkerThread(final ForkJoinPool pool, final String threadName) {
-                super(pool);
-                this.setName(threadName);
-                this.setUncaughtExceptionHandler(NettyServerRouter::handleUncaughtException);
-            }
-
-            @Override
-            protected void onTermination(Throwable exception) {
-                if (exception != null) {
-                    log.error("onTermination: Thread terminated due to {}:{}",
-                            exception.getClass().getSimpleName(),
-                            exception.getMessage(),
-                            exception);
-                } else {
-                    log.debug("onTermination: Thread terminated (completed normally).");
-                }
-                super.onTermination(exception);
-            }
-        }
-
-        @Override
-        public ForkJoinWorkerThread newThread(ForkJoinPool pool) {
-            return new ServerWorkerThread(pool,
-                    THREAD_PREFIX + threadNumber.getAndIncrement());
-        }
-    }
-
-    protected static void handleUncaughtException(Thread t, @Nonnull Throwable e) {
-        log.error("handleUncaughtException[{}]: Uncaught {}:{}",
-                t.getName(),
-                e.getClass().getSimpleName(),
-                e.getMessage(),
-                e);
-    }
-
-    protected final ExecutorService handlerWorkers =
-            new ForkJoinPool(Runtime.getRuntime().availableProcessors() * 2,
-                    new ServerThreadFactory(),
-                    NettyServerRouter::handleUncaughtException, true);
 
     /**
      * This map stores the mapping from message type to netty server handler.
@@ -179,7 +126,8 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter
                         log.trace("Message routed to {}: {}", handler.getClass().getSimpleName(),
                                 msg);
                     }
-                    handlerWorkers.submit(() -> {
+
+                    handler.getExecutor().submit(() -> {
                         try {
                             handler.handleMessage(m, ctx, this);
                         } catch (Throwable t) {
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -14,6 +14,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import lombok.Getter;
@@ -140,6 +143,16 @@ public class SequencerServer extends AbstractServer {
             return false;
         }
         return true;
+    }
+
+    ThreadFactory threadFactory = new ServerThreadFactory("sequencer-",
+            new ServerThreadFactory.ExceptionHandler());
+
+    ExecutorService executor = Executors.newSingleThreadExecutor(threadFactory);
+
+    @Override
+    public ExecutorService getExecutor() {
+        return executor;
     }
 
     /**
@@ -547,6 +560,12 @@ public class SequencerServer extends AbstractServer {
         r.sendResponse(ctx, msg, CorfuMsgType.TOKEN_RES.payloadMsg(new TokenResponse(
                 TokenType.NORMAL, TokenResponse.NO_CONFLICT_KEY, token,
                 backPointerMap.build())));
+    }
+
+    @Override
+    public void shutdown() {
+        super.shutdown();
+        executor.shutdownNow();
     }
 
     @VisibleForTesting
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerThreadFactory.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerThreadFactory.java
@@ -1,0 +1,44 @@
+package org.corfudb.infrastructure;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Slf4j
+public class ServerThreadFactory implements ThreadFactory {
+
+    private static final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
+
+    private final Thread.UncaughtExceptionHandler handler;
+
+    private final String threadPrefix;
+
+    final AtomicInteger threadNumber = new AtomicInteger(0);
+
+    public ServerThreadFactory(String threadPrefix, Thread.UncaughtExceptionHandler handler) {
+        this.threadPrefix = threadPrefix;
+        this.handler = handler;
+    }
+
+    @Override
+    public Thread newThread(Runnable run) {
+        Thread thread = defaultFactory.newThread(run);
+        thread.setName(threadPrefix + threadNumber.getAndIncrement());
+        thread.setUncaughtExceptionHandler(handler);
+        thread.setDaemon(true);
+        return thread;
+    }
+
+    public static class ExceptionHandler implements Thread.UncaughtExceptionHandler {
+        @Override
+        public void uncaughtException(Thread t, Throwable e) {
+            log.error("handleUncaughtException[{}]: Uncaught {}:{}",
+                    t.getName(),
+                    e.getClass().getSimpleName(),
+                    e.getMessage(),
+                    e);
+        }
+    }
+}
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -12,9 +12,11 @@ import io.netty.channel.ChannelHandlerContext;
 import java.lang.invoke.MethodHandles;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
 import lombok.Getter;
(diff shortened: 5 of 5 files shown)


corfu-repo:

1. ![MINOR][MINOR] [NettyServerRouter.java#L10](https://github.com/CorfuDB/CorfuDB/blob/16378445e780aafd165224ce892eb23b9d133a98/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java#L10): Remove this unused import 'java.util.concurrent.ExecutorService'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -139,6 +143,15 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
         return true;
     }
 
+    ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat("sequencer-%d").build();
+
+    ExecutorService executor = Executors.newSingleThreadExecutor(threadFactory);
```
zalokhan: Can you add a shutdown method to gracefully terminate this executor service?

zalokhan: Also, the NettyServerRouter thread pool registers the exception handler `handleUncaughtException`. Is there any way to add it here too to maintain the same functionality?

Maithem: Done.

Maithem:
@zalokhan let's leave the synchronization for now just in case we want to experiment with more threads later.

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
```diff
@@ -168,20 +131,21 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
             // pipeline.
             CorfuMsg m = ((CorfuMsg) msg);
             // We get the handler for this message from the map
-            AbstractServer handler = handlerMap.get(m.getMsgType());
-            if (handler == null) {
+            AbstractServer server = handlerMap.get(m.getMsgType());
```
StrikingThirteen: Isn't handler a more precise description of what an AbstractServer does? otherwise maybe the handlerMap should also changed to serverMap.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerThreadFactory.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Slf4j
+public class ServerThreadFactory implements ThreadFactory {
+
+    private static final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
+
+    private final Thread.UncaughtExceptionHandler handler;
+
+    public final String threadPrefix;
```
StrikingThirteen: nit: can it be converted to private?

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -16,6 +20,11 @@
     @Setter
     volatile boolean shutdown;
 
+    static final ExecutorService sharedExecutor = Executors
```
medhavidhawan: Can we try to separate datapath (sequencer/log unit) thread pool(s) from control path (Layout/Management server) thread pool(s). The control path can run on single or at most 2 threads. Since their queues will be separate they will not be able interfere so much with datapath.

medhavidhawan: Do we have any thoughts around any metrics on thread starvation/contention we want  to expose for further tuning this ?

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -16,6 +20,11 @@
...
+            .newFixedThreadPool(Runtime.getRuntime().availableProcessors() * 2,
```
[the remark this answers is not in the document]
Maithem: I can only think of the sequencer threads and those are separated out.

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
Maithem: Changed back.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerThreadFactory.java
Maithem: done.

medhavidhawan:
So this will improve performance on one side. The other change you are asking for is something that is already planned. I think Maithem is already working on it. So can we have both changes in ?

no2chem:
@medhavidhawan Do we really know it will improve performance? It's not clear that it will, it introduces several extra thread pools, including for the management server, and without actually running the code, it seems that there is no way to shutdown the static thread pool in abstractServer safely. This is concerning because there's not a way to determine that all log unit threads have been completed, for example.

Also, I haven't run this, but since the fixedThreadPool is not daemon, it seems to also prevent the server from cleanly exiting. The ForkJoinPool doesn't have this problem, because it starts all threads as daemon [1](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html).

I stand by my concern that this PR simply masks the problem, we don't actually know whether it will improver performance, and furthermore, it may introduce correctness issues relating to cleanup and shutdown.

medhavidhawan:
Re Performance: In the microbenchmarks for sequencer we saw throughput gains. The high level tests that we have in our performance framework did not regress with the sequencer change. 

You are correct in pointing out that we should benchmark with the other thread changes. Maybe it has been done and I am not aware, I will check. 

Re Different thread pools: The idea behind different thread pools is to let different servers run more independently and not use the same threadpool and queue for all functions.

FixedThreadPool vs ForkJoinPool: I do not have a preference but if one is performing better than the other, we should pick that one and if Fixed threadpool has shutdown issues we should fix them or use ForkJoin pool.
So in the end what I see might help:
One Sequencer thread to reduce synchronization overhead on the sequencer processing path.
One thread for LayoutServer
One thread for ManagementServer
Maintain as many threads as the (number of cores  - 1  for sequencer ) for logunit. 

I do not think this is masking any problems.This is an independent activity in which we are trying to separate threads for some throughput gain. If log unit is slower for some other reason, we should fix it.

medhavidhawan:
@no2chem "We are not adding threadpools for kicks. If it appears so, I apologise and assure you that we are not trying any such thing.". If you are showing a better alternative, we will all gladly agree to it. Please provide more fine grained details to the nothreadpool work you have. Just providing a PR to look at for comparison is hard. Maybe we can look at it in person. 
The design/architecture does not preclude the possibility of running servers independently.
The main justification for independent threadpools in my mind was this:
We can have a mix of datplane and controlplane functionality in a node, so it would be good that we have separate processing queues and threadpools for these functions. We will have reduced impact of any one slow function on another one. A slow layout or management server should not affect the sequencer or the logunit. A slow logunit should not affect the sequencer. 

If you believe this intuition is counter to the emperical evidence that you have, I would say let us go with what the data is providing us.

no2chem:
@medhavidhawan It couldn't be possibly that hard to understand which only adds 3 lines of code (and deletes 59): https://github.com/CorfuDB/CorfuDB/pull/1288/files compared to the 83 lines (and 44 deletions) this PR adds.

Have you actually been able to create starvation issues in practice? I haven't actually observed any starvation issues with a reasonable number of cores using a few clients, you need hundreds of clients to actually see the effects. And that's not why you're seeing a performance improvement anyway.
 
What this PR does is force sequencer requests to be served by the same single thread, which does have some value as it decreases the amount of context switches and sharing. Since the sequencer is single threaded, you probably could remove locks too.

I'm not just saying your intuition here is wrong, I'm also saying that you misunderstand Netty's threading model here. You are putting multiple thread pools in front of a dispatch pool already provided by Netty. If you fix the blocking calls in LogUnit, then theres **no point in re-dispatching to a secondary pool**, and it's not clear that the blocking calls need to be immediately fixed anyway, since I at least don't observe they are an issue (they may be an issue with hard disk). Either way, in that case the correct fix in that case would be to fix the blocking calls in log unit.

You can see the PR for the performance yourself, but to compare against this PR:
```
Benchmark                      (CONFLICTS_PER_STREAM)  (STREAMS)   Mode  Cnt   Score   Error   Units
SequencerBenchmark.getToken                       N/A        N/A  thrpt   10  71.633 ± 4.670  ops/ms
SequencerBenchmark.queryToken                     N/A        N/A  thrpt   10  73.937 ± 7.076  ops/ms
SequencerBenchmark.txToken                          1          1  thrpt   10  47.810 ± 1.414  ops/ms
SequencerBenchmark.txToken                          1         10  thrpt   10  16.801 ± 2.315  ops/ms
SequencerBenchmark.txToken                         10          1  thrpt   10  26.649 ± 1.117  ops/ms
SequencerBenchmark.txToken                         10         10  thrpt   10   5.316 ± 0.496  ops/ms
SequencerBenchmark.txToken                        100          1  thrpt   10   5.358 ± 0.375  ops/ms
SequencerBenchmark.txToken                        100         10  thrpt   10   0.611 ± 0.075  ops/ms
```

No pools:
```
Benchmark                      (CONFLICTS_PER_STREAM)  (STREAMS)   Mode  Cnt   Score   Error   Units
SequencerBenchmark.getToken                       N/A        N/A  thrpt   10  76.968 ± 9.034  ops/ms
SequencerBenchmark.queryToken                     N/A        N/A  thrpt   10  78.844 ± 7.246  ops/ms
SequencerBenchmark.txToken                          1          1  thrpt   10  45.043 ± 2.411  ops/ms
SequencerBenchmark.txToken                          1         10  thrpt   10  15.409 ± 2.495  ops/ms
SequencerBenchmark.txToken                         10          1  thrpt   10  25.650 ± 1.515  ops/ms
SequencerBenchmark.txToken                         10         10  thrpt   10   4.887 ± 0.442  ops/ms
SequencerBenchmark.txToken                        100          1  thrpt   10   5.270 ± 0.412  ops/ms
SequencerBenchmark.txToken                        100         10  thrpt   10   0.598 ± 0.045  ops/ms
```

Baseline:
```
Benchmark                      (CONFLICTS_PER_STREAM)  (STREAMS)   Mode  Cnt   Score   Error   Units
SequencerBenchmark.getToken                       N/A        N/A  thrpt   10  51.167 ± 8.090  ops/ms
SequencerBenchmark.queryToken                     N/A        N/A  thrpt   10  52.468 ± 4.853  ops/ms
SequencerBenchmark.txToken                          1          1  thrpt   10  38.195 ± 2.092  ops/ms
SequencerBenchmark.txToken                          1         10  thrpt   10  16.096 ± 2.491  ops/ms
SequencerBenchmark.txToken                         10          1  thrpt   10  25.556 ± 0.999  ops/ms
SequencerBenchmark.txToken                         10         10  thrpt   10   5.385 ± 0.236  ops/ms
SequencerBenchmark.txToken                        100          1  thrpt   10   5.450 ± 0.356  ops/ms
SequencerBenchmark.txToken                        100         10  thrpt   10   0.568 ± 0.114  ops/ms
```

And transactions:
```Benchmark                                                     (NUM_KEYS)  (NUM_MAPS)  (NUM_OPERATIONS)   Mode  Cnt  Score   Error   Units
OptimisticTransactionBenchmark.readWriteDiffMap                       10          10                 1  thrpt   10  0.384 ± 0.113  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10          10               100  thrpt   10  0.048 ± 0.007  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10         100                 1  thrpt   10  2.257 ± 0.477  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10         100               100  thrpt   10  0.009 ± 0.001  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100          10                 1  thrpt   10  0.409 ± 0.132  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100          10               100  thrpt   10  0.049 ± 0.010  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100         100                 1  thrpt   10  2.747 ± 0.320  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100         100               100  thrpt   10  0.009 ± 0.001  ops/ms
```

No pools:
```
Benchmark                                                     (NUM_KEYS)  (NUM_MAPS)  (NUM_OPERATIONS)   Mode  Cnt  Score   Error   Units
OptimisticTransactionBenchmark.readWriteDiffMap                       10          10                 1  thrpt   10  0.579 ± 0.099  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10          10               100  thrpt   10  0.037 ± 0.006  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10         100                 1  thrpt   10  2.587 ± 0.280  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10         100               100  thrpt   10  0.009 ± 0.001  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100          10                 1  thrpt   10  0.495 ± 0.105  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100          10               100  thrpt   10  0.097 ± 0.034  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100         100                 1  thrpt   10  2.921 ± 0.394  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100         100               100  thrpt   10  0.009 ± 0.001  ops/ms
```

 baseline:
```
Benchmark                                                     (NUM_KEYS)  (NUM_MAPS)  (NUM_OPERATIONS)   Mode  Cnt  Score   Error   Units
OptimisticTransactionBenchmark.readWriteDiffMap                       10          10                 1  thrpt   10  0.415 ± 0.093  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10          10               100  thrpt   10  0.132 ± 0.073  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10         100                 1  thrpt   10  2.691 ± 0.413  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                       10         100               100  thrpt   10  0.009 ± 0.001  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100          10                 1  thrpt   10  0.413 ± 0.114  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100          10               100  thrpt   10  0.142 ± 0.097  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100         100                 1  thrpt   10  2.466 ± 0.172  ops/ms
OptimisticTransactionBenchmark.readWriteDiffMap                      100         100               100  thrpt   10  0.009 ± 0.001  ops/ms
```

no2chem:
BTW, this isn't to say we shouldn't have per-server thread pools period, what I'm saying is:

(1) message dispatch and handling should be fast and non-blocking. Then there are no issues of starvation on the message handling pool and no need to do double-dispatch

(2) when message handling needs to do I/O, it should be asynchronously queued and handled. The LogUnit is a good example of this - instead of doing the I/O on the message pool and holding a thread, the handler should insert the request into a queue, which could be completed by an alternate thread pool. Of course not all LogUnit requests are going to be blocking, so for example a message which hits the cache or to do a tail query can be handled by the message handler.

no2chem:
@medhavidhawan Of course that kind of QoS would be ideal, but I'm not sure how you would achieve that in practice without significant overhead. This PR doesn't achieve that - while it guarantees that there is always a single thread to service the sequencer, messages are still processed in the order they arrive. So if you have 6 logunit requests and a token after it, the log unit requests will be processed first, and will have to wait. Unless there are more processors than requests of course, and the sequencer pool has affinity for it's own dedicated processor.

In resource constrained situations though, you'll have to dump everything in a priority queue or a linked blocking queue of some sort. And I'm not sure that's worth the overhead.

It seems to me that an easier initial design is to treat all nonblocking messages fairly and to queue blocking messages.

medhavidhawan:
I think this PR might not be perfect but it is a simple attempt to move messages from the incoming queue to the different thread pool queues. The assumption is that this moving of messages is relatively fast. After that, the problem is to make sure that there is an opportunity for sequencer thread to keep running such that the log unit threads are not taking away all its cycles. The io bound threads (log unit threads ) can be on a fork-join pool such that we maintain enough parallelism without taking away all the cpus. 
Again I am not sure what we have proven empirically yet but let's go with what the agreed upon tests say. 

I agree priority queuing would be great but as you said we might not want the overhead (we can test it if we have cycles).

no2chem:
@medhavidhawan @Maithem Diving further into this, here are some microbenchmarks showing the performance degradation:

```
Benchmark                                         Mode  Cnt   Score    Error   Units
EventLoopBenchmark.send                          thrpt   10  90.148 ± 31.248  ops/ms
EventLoopBenchmark.sendWithCachedExecutor        thrpt   10  79.922 ± 13.754  ops/ms
EventLoopBenchmark.sendWithFixedExecutor         thrpt   10  77.494 ± 14.158  ops/ms
EventLoopBenchmark.sendWithWorkStealingExecutor  thrpt   10  83.868 ± 21.360  ops/ms
```

This is just with a single PING/PONG. You can see that just passing the message to be handled in a work stealing executor reduces performance by 7%, approximately, worse with a fixed thread pool 15%, where there is excessive queueing.

This is somewhat correlated with the same results and issues I pointed out in #989. So I stand by my statements and continue to believe that #1288 is a better fix, perhaps in addition to fixing the behavior of the logunit handlers.

zalokhan:
Running the Benchmark tests from PR #1256 to compare the performance of #1281  and #1288 
The benchmark tests connected to a standalone persisted CorfuServer.

16 Threads | getToken | queryToken | txToken
-- | -- | -- | --
**Master** | 18.559 | 18.271 | 18.61
**NoThread** | 15.954 | 15.833 | 12.541
**1-Thread** | 23.506 | 24.317 | 22.817

32 Threads | getToken | queryToken | txToken
-- | -- | -- | --
**Master** | 17.684 | 18.061 | 17.858
**NoThread** | 15.765 | 15.689 | 13.509
**1-Thread** | 22.105 | 22.945 | 22.079

64 Threads | getToken | queryToken | txToken
-- | -- | -- | --
**Master** | 20.458 | 21.094 | 19.357
**NoThread** | 16.008 | 15.474 | 12.782
**1-Thread** | 22.535 | 23.179 | 21.789

Issue #1222, opened 2018-03-15
Race between H.O.G and Paxos


zalokhan:
## Overview
There is a race between the force commit layout by the Hand Of God and the regular Paxos run.
There can be 2 clients able to commit a layout to the same epoch using the above 2 code paths.

Found by intermittently failing `AddAndRemoveNodeIT` test.

### Proposed solution
In case of a `force layout`, we should also block the `phase1` and `phase2` slots for this epoch.
At the same time, we should check if there is already `phase2` data present for this epoch, we cannot force the layout on this epoch as this slot is filled. So we will have to abort and then attempt to occupy the next epoch slot.

PR #1295, opened 2018-05-18
Fix failing AddAndRemoveNodeIT

zalokhan:
## Overview

Description: Fixes intermittent test failure. This fix shuts down the removed nodes.
NOTE: The main reason for this failure is the HoG and Paxos race #1222 

Why should this be merged: Avoids intermittent failure on CI.

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/WorkflowIT.java
+++ b/test/src/test/java/org/corfudb/integration/WorkflowIT.java
@@ -8,6 +8,7 @@ import org.corfudb.runtime.view.Layout;
 import org.corfudb.runtime.view.LayoutBuilder;
 import org.junit.Test;
 
+import java.io.IOException;
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


test/src/test/java/org/corfudb/integration/WorkflowIT.java
```diff
@@ -37,35 +38,30 @@ String getConnectionString(int port) {
     final Duration pollPeriod = Duration.ofMillis(50);
     final int workflowNumRetry = 3;
 
+    private Process runServer(int port, boolean single) throws IOException {
```
StrikingThirteen: If there is also a parameter for host, this helper method would be more flexible.

zalokhan: We already have a generic function as CorfuRunner. I just wanted a common function for this class.

Issue #1298, opened 2018-05-21
Use lazy log argument evaluations in place of eager log message preparations


StrikingThirteen:
## Overview

In order to avoid the performance penalty of indiscriminate log message creation, It is recommended to evaluate log message arguments lazily. 
The advantage of lazy log argument evaluation is that the message won't be prepared if the logging level doesn't match the logging configuration.

StrikingThirteen:
Already merged. Closing the issue.

PR #1299, opened 2018-05-21
Replacing string concatenations with lazy argument evaluations

StrikingThirteen:
## Overview

Description:
This PR, replaces the eager log message preparations with lazy argument evaluations. 

Why should this be merged: 
Eager message preparation creates log messages regardless of the logging level which leads to unnecessary message creation. Lazy evaluation however only prepares the messages when logging level requires a message to be logged.

Related issue(s) (if applicable): #1298 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
@@ -1,8 +1,6 @@
 package org.corfudb.infrastructure;
 
 import io.netty.channel.ChannelHandlerContext;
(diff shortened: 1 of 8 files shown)


StrikingThirteen:
@no2chem Thanks. It would be something like a regex that catches a '+' after a logger while that '+' is followed by a character excluding a quotation)
So it should catch: 
log.debug("Message" + msg) 

But doesn't bother with (as this will be concatenated  at compile time):
log.debug("Recovery write loop for [long line]" +
                   "second line of log")

Not sure about the concrete regex expression that catches all the loggers though.

PR #1301, opened 2018-05-23
Refactoring: emptying handler's queues and using singleton instead of list

StrikingThirteen:
## Overview

Description of refactoring:
This is a refactoring on Server and Client HandshakeHandlers and MultipleReadRequest for the following purposes:
- Removing messages from queue at the time of writing and flushing them by the HandshakeHandlers
- Fixing unchecked assignment
- Replacing implementation of Queue (from ArrayDeque to LinkedList) for memory allocation and consistent o(1) 
- Using singleton instead of creating a list of one

Why should this be merged: 
- Avoiding to unnecessarily keep messages in queue of HandshakeHandler
- Consistent o(1) add/remove from queue and avoiding ArrayDeque's memory overhead
- Replacing unchecked assignment
- Avoiding list creation for one member

Related issue(s) (if applicable): #1316 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
@@ -8,7 +8,7 @@ import io.netty.handler.timeout.ReadTimeoutHandler;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
-import java.util.ArrayDeque;
+import java.util.LinkedList;
 import java.util.Queue;
 import java.util.UUID;
 
@@ -30,7 +30,7 @@ public class ClientHandshakeHandler extends ChannelDuplexHandler {
     private final UUID nodeId;
     private final int handshakeTimeout;
     private final HandshakeState handshakeState;
-    private final Queue<CorfuMsg> messages = new ArrayDeque();
+    private final Queue<CorfuMsg> messages = new LinkedList<>();
     private static final String READ_TIMEOUT_HANDLER = "readTimeoutHandler";
 
     /** Events that the handshaker sends to downstream handlers.
@@ -40,7 +40,6 @@ public class ClientHandshakeHandler extends ChannelDuplexHandler {
         CONNECTED,  /* Connection succeeded. */
         FAILED      /* Handshake failed. */
     }
-
 
     /**
      * Creates a new ClientHandshakeHandler which will handle the handshake between the
@@ -130,8 +129,8 @@ public class ClientHandshakeHandler extends ChannelDuplexHandler {
         log.info("channelRead: Handshake succeeded. Server Corfu Version: [{}]", corfuVersion);
         log.debug("channelRead: There are [{}] messages in queue to be flushed.", this.messages.size());
         // Flush messages in queue
-        for (CorfuMsg message : this.messages) {
-            ctx.writeAndFlush(message);
+        while (!messages.isEmpty()) {
+            ctx.writeAndFlush(messages.poll());
         }
 
         // Remove this handler from the pipeline; handshake is completed.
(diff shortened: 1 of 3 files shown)


StrikingThirteen:
@annym Thanks Anny. Yes, it was green but I'll make sure that it passes again. The last I checked it failed on intermittent KillAndHeal.

StrikingThirteen:
Travis tests are passing. Rebasing and squashing commits to prepare for merge.

Maithem:
@StrikingThirteen  performance tests still running ?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -30,7 +30,7 @@
     private final UUID nodeId;
     private final int handshakeTimeout;
     private final HandshakeState handshakeState;
-    private final Queue<CorfuMsg> messages = new ArrayDeque();
+    private final Queue<CorfuMsg> messages = new LinkedList<>();
```
Maithem: I think I'm missing the point, but why is this change needed? A linked list implementation would be better if we remove elements that are not at the tail/head, but I don't see any logic that does that.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -130,8 +129,8 @@ public void channelRead(ChannelHandlerContext ctx, Object m)
         log.info("channelRead: Handshake succeeded. Server Corfu Version: [{}]", corfuVersion);
         log.debug("channelRead: There are [{}] messages in queue to be flushed.", this.messages.size());
         // Flush messages in queue
-        for (CorfuMsg message : this.messages) {
-            ctx.writeAndFlush(message);
+        while (!messages.isEmpty()) {
+            ctx.writeAndFlush(messages.poll());
```
Maithem: The messages are just being iterated, don't you actually want to drain the queue?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
StrikingThirteen: LinkedList does provide a consistent time (without the amortization of ArrayDeque or without any additional backing array assigned). Also the messages queue is not being randomly accessed, it's purely a FIFO making LinkedList a natural pick. That being said, it's not required per se so if you're against it, I'll keep the ArrayDeque.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
StrikingThirteen: Actually the queue is being drained and hence the poll(). Can you elaborate on why this is just an iteration?

Maithem: Hmm, this was suppose to be a comment on the diff:

for (CorfuMsg message : this.messages) {
            ctx.writeAndFlush(message);
}

Is this leaking actually a leak on master?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
Maithem: I was referring to better caching behavior for arrays, but it seems like you introduced a remove (i.e. poll) which didn't exist in the old code. Also, this doesn't seem to be a critical path, so I'm in different.

StrikingThirteen: Sounds good. Thanks

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
StrikingThirteen: I see. correct, master still is not draining.

Issue #1320, opened 2018-06-13
Snapshot Transaction


Maithem:
**Bug report.**

When a SNAPSHOT transaction is started without a version number, it defaults to -1 as its snapshot, this yields and empty object. I think its more natural to start at the current global tail if a snapshot version isn't provided.

PR #1323, opened 2018-06-18
Snapshot Transaction Default Timestamp

Maithem:
## Overview
Changed the default snapshot transaction's timestamp to the global
tail.

Why should this be merged: The default timestamp is -1, this results in reading an empty object. Using the global tail as the default timestamp hides the need to explicitly call the sequencer view from app code, which is convenient. 

Related issue(s) (if applicable): #1320

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/object/transactions/SnapshotTransactionalContext.java
```diff
@@ -91,7 +92,11 @@ public void addTransaction(AbstractTransactionalContext tc) {
 
     @Override
     public long obtainSnapshotTimestamp() {
-        return getBuilder().getSnapshot();
+        if (getBuilder().getSnapshot() == Address.NON_ADDRESS) {
```
StrikingThirteen: Is there a new possibility introduced where SnapshotTransactionContext's obtainSnapshotTimestamp() is called and the getBuilder().getSanpshot() is a non address?

no2chem: That would seem to be a bug in application logic (the application provided an illegal address in the builder). Perhaps we could check for that in the constructor in the future, but I don't see any behavior change.

Maithem: I tried something like : 

+    public TransactionBuilder setSnapshot(long snapshot) {
+        if (snapshot <= Address.NON_ADDRESS) {
+            throw new IllegalArgumentException("Invalid snapshot: the timestamp has to be greater than -1");
+        }
+
+        this.snapshot = snapshot;
+        return this;
+    }

but some tests were failing, because -1 is actually a valid snapshot (i.e. empty). We can't force > -1 because [0, inf+) might have not been written.

runtime/src/main/java/org/corfudb/runtime/object/transactions/TransactionBuilder.java
```diff
@@ -27,7 +28,7 @@
     /** For snapshot transactions, the address the
      * snapshot will start at.
      */
-    public long snapshot = -1L;
+    public long snapshot = Address.NON_ADDRESS;
```
corfu-repo: MINOR Make snapshot a static final constant or non-public and provide accessors if needed. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

corfu-repo:
* ![MINOR][MINOR] 1 minor

Issue #1322, opened 2018-06-15
Quorum replication gets incorrect minimum replication factor


zalokhan:
## Overview

Minimum replication factor calculated based on the layout servers. Rather it should be based on the number of log unit servers in the stripe[s].

https://github.com/CorfuDB/CorfuDB/blob/0b3164e8d08553db28f80ca8130e6c413b2af8af/runtime/src/main/java/org/corfudb/runtime/view/Layout.java#L369-L371

zalokhan:
This `minReplicationFactor` is intended for log unit servers only. And this min replication factor is different and needs to be calculated per stripe for the given segment.

PR #1326, opened 2018-06-18
Chain min replication factor set to 1.

### LINKED ISSUE #1325: Chain replication minimum factor 2 causes deadlock.
## Overview
Consider the following scenario with 3 nodes A, B and C using chain replication A -> B -> C.
The cluster is bootstrapped with the following layout:
```
Epoch:0
Segments: [0 to -1: A->B->C]
```
Node B fails causing B to be marked unresponsive and removed from the chain
```
Epoch:1
Segments: [0 to -1: A->C]
Unresponsive: B
```
Now Node B recovers. The healing workflow kicks in splitting the segment at address 10 to add node B.
```
Epoch:2
Segments: [0 to 10: A->C], [10 to -1: A->C->B]
```
The next step would be to restore redundacy in node B for the first segment.
However, before this can be done, node C crashes.
This creates a deadlock as node B cannot recover as it cannot read the first segment from the tail of the chain (Node C). At the same time, Node C cannot be removed from the chain as it violates the minimum replication factor.

zalokhan:
## Overview
Description: Change the min replication factor for Chain replication to 1.
Also corrects the Quorum replication min factor to calculate the quorum based on log servers in each stripe.

Why should this be merged: Resolves liveness issue caused by #1325

Related issue(s) (if applicable): Resolves #1325 and #1322 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/Layout.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/Layout.java
@@ -329,8 +329,8 @@ public class Layout {
             }
 
             @Override
-            public int getMinReplicationFactor(Layout layout) {
-                return 2;
+            public int getMinReplicationFactor(Layout layout, LayoutStripe stripe) {
+                return 1;
             }
 
             @Override
@@ -366,8 +366,8 @@ public class Layout {
             }
 
             @Override
-            public int getMinReplicationFactor(Layout layout) {
-                return (layout.getLayoutServers().size() / 2) + 1;
+            public int getMinReplicationFactor(Layout layout, LayoutStripe stripe) {
+                return (stripe.getLogServers().size() / 2) + 1;
             }
 
             @Override
@@ -421,7 +421,7 @@ public class Layout {
             }
 
             @Override
-            public int getMinReplicationFactor(Layout layout) {
+            public int getMinReplicationFactor(Layout layout, LayoutStripe stripe) {
                 return 1;
             }
 
@@ -447,12 +447,14 @@ public class Layout {
                 throws QuorumUnreachableException;
 
         /**
-         * Compute the min replication factor for replication protocol
+         * Compute the min replication factor for the log unit servers in the replication protocol
+         * for a specific stripe.
          *
-         * @param layout the layout to compute the min replication factor for
+         * @param layout the layout to compute the min replication factor for.
+         * @param stripe The stripe for which the minimum replication factor is needed.
          * @return the minimum amount of nodes required to maintain replication
          */
-        public abstract int getMinReplicationFactor(Layout layout);
+        public abstract int getMinReplicationFactor(Layout layout, LayoutStripe stripe);
 
         public abstract IStreamView getStreamView(CorfuRuntime r, UUID streamId, StreamOptions options);
 
--- a/test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
+++ b/test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
@@ -1478,9 +1478,10 @@ public class ManagementViewTest extends AbstractViewTest {
         // STEP 3.
         clearClientRules(getCorfuRuntime());
         addServerRule(SERVERS.PORT_0, new TestRule().drop().always());
+        getManagementServer(SERVERS.PORT_0).shutdown();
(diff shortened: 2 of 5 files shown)


test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1500,11 +1501,13 @@ public void queryClusterStatus() throws Exception {
...
+                        return true;
                     }
-                    return true;
-                }));
+                    return false;
+                }).drop());
         addServerRule(SERVERS.PORT_1, new TestRule().drop().always());
-        latch.tryAcquire(PARAMETERS.TIMEOUT_NORMAL.toMillis(), TimeUnit.MILLISECONDS);
+        getManagementServer(SERVERS.PORT_1).shutdown();
+        latch.acquire();
```
Maithem: I think this should be time bounded.

zalokhan: Right, I'll put the tryAcquire back.

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
```diff
@@ -452,7 +452,7 @@ public abstract void validateSegmentSeal(LayoutSegment layoutSegment,
          * @param layout the layout to compute the min replication factor for
```
StrikingThirteen: can we have the stripe added to comments too?

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
```diff
@@ -329,8 +329,8 @@ public void validateSegmentSeal(LayoutSegment layoutSegment,
             }
 
             @Override
-            public int getMinReplicationFactor(Layout layout) {
-                return 2;
+            public int getMinReplicationFactor(Layout layout, LayoutStripe stripe) {
```
annym: This might be a bit unrelated, but why isn't the LayoutStripe part of the Layout definition?

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
```diff
@@ -329,8 +329,8 @@ public void validateSegmentSeal(LayoutSegment layoutSegment,
...
+                return 1;
```
annym: Would it make somehow sense for the replication factor to be defined/changed in the layout file? so it is not actually hard coded, and default to this if not present?

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
zalokhan: Layout stripe is a part of the layout. We can just pass the stripe index and infer the stripe from the passed layout object. I felt this would be more intuitive since we already have the stripe.

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
zalokhan: The minimum replication factor should not change that often for it to be configurable. We hit a deadlock condition if the minimum replication factor is 2 in case of Chain replication.
On the other hand, the minimum replication factor would not be constant for quorum replication. It would increase as the number of nodes in the stripe increase and decrease when the nodes are removed.

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
annym: Ok, I see I was a bit confused because both were passed separately. But then I agree passing the stripe is more intuitive than the stripe index.

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
annym: Ok, I was only thinking it for the chain replication but if this leads to potential deadlock then it would make sense to leave it out of the scope of the layout. 

Just to be sure, stripes are updated as a server becomes unresponsive? I'm just trying to make sure that stripes are not accounting for log servers that are down.

zalokhan: Yes, the stripes are updated. If a node is down it is removed from the stripe. Provided the number of nodes remaining in the stripe is greater than or equal to the `minReplicationFactor`.
If not, the system just hangs until the condition is satisfied.

zalokhan:
@Maithem @StrikingThirteen I have addressed your comments.

PR #1328, opened 2018-06-19
Using named generic instead of wildcard generic of Comparable for IndexRegistry

StrikingThirteen:
## Overview

Description:
Using named generic instead of wildcard generic of Comparable in IndexRegistry to address casting issues on the client side.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -161,7 +161,7 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
 
         CorfuTable.IndexRegistry<?, ?> EMPTY = new CorfuTable.IndexRegistry<Object, Object>() {
             @Override
-            public Optional<CorfuTable.Index<Object, Object, ? extends Comparable<?>>> get(CorfuTable.IndexName name) {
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -161,7 +161,7 @@ public int hashCode() {
 
         CorfuTable.IndexRegistry<?, ?> EMPTY = new CorfuTable.IndexRegistry<Object, Object>() {
             @Override
-            public Optional<CorfuTable.Index<Object, Object, ? extends Comparable<?>>> get(CorfuTable.IndexName name) {
+            public <I extends Comparable<?>> Optional<CorfuTable.Index<Object, Object, I>> get(CorfuTable.IndexName name) {
```
Maithem: Why do you need this. Are you trying to force the compiler to error out the index function mismatch ?

StrikingThirteen: It used to be a named generic like this before I changed it to wildcard. My change however requires client to specifically cast to a type (e.g. I) and suppressing warnings. However in the old form (current PR brings it back) provided type in the definition (e.g. I) is chosen as the return value from registry's get method and no additional casting or suppressing warning is required on the client side.

Issue #1329, opened 2018-06-19
Don't Cache Checkpoint Entries


Maithem:
## Overview

Client's that write checkpoints don't require the checkpoint data during normal operation. Checkpoint's don't generate undo records and they are only required to build the object's state on initialization, not caching these entries would greatly relieve memory pressure. Particularity because checkpointing is done in-place.

PR #1330, opened 2018-06-19
Added Cache Options Per Write

Maithem:
Description:
Now appends can specify a caching option per append. Also, changed
the CheckpointWriter to no cache checkpoint entries.

Why should this be merged: This helps relieve memory pressure.

Related issue(s) (if applicable): #1329

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -5,6 +5,7 @@ import com.codahale.metrics.MetricRegistry;
 import com.github.benmanes.caffeine.cache.CacheLoader;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Iterables;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.DataType;
@@ -117,14 +118,15 @@ public class AddressSpaceView extends AbstractView {
      * has been adopted, or a WrongEpochException if the
      * token epoch is invalid.
      *
-     * @param token     The token to use for the write.
-     * @param data      The data to write.
+     * @param token        The token to use for the write.
+     * @param data         The data to write.
+     * @param cacheOption  The caching behaviour for this write
      * @throws OverwriteException   If the globalAddress given
      *                              by the token has adopted
      *                              another value.
      * @throws WrongEpochException  If the token epoch is invalid.
      */
-    public void write(IToken token, Object data) throws OverwriteException {
+    public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOption cacheOption) {
         final ILogData ld = new LogData(DataType.DATA, data);
 
         layoutHelper(e -> {
@@ -158,9 +160,19 @@ public class AddressSpaceView extends AbstractView {
         }, true);
 
         // Cache the successful write
-        if (!runtime.getParameters().isCacheDisabled()) {
+        if (!runtime.getParameters().isCacheDisabled() && cacheOption == CacheOption.WRITE_THROUGH) {
             readCache.put(token.getTokenValue(), ld);
         }
+    }
+
+    /**
+     * Write the given log data and then add it to the address
+     * space cache (i.e. WRITE_THROUGH option)
+     *
+     * @see AddressSpaceView#write(IToken, Object, CacheOption)
+     */
+    public void write(IToken token, Object data) throws OverwriteException {
+        write(token, data, CacheOption.WRITE_THROUGH);
     }
 
     /** Directly read from the log, returning any
@@ -376,5 +388,10 @@ public class AddressSpaceView extends AbstractView {
                 .getReplicationProtocol(runtime)
                 .read(e, address)
         );
+    }
+
+    @VisibleForTesting
+    LoadingCache<Long, ILogData> getReadCache() {
+        return readCache;
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
@@ -69,13 +69,13 @@ public class StreamsView extends AbstractView {
      * @param streamIDs    The streams to append to.
      * @param object       The object to append to each stream.
      * @param conflictInfo Conflict information for the sequencer to check.
+     * @param cacheOption  The caching mode for write/append
      * @return The address the entry was written to.
      * @throws TransactionAbortedException If the transaction was aborted by
      *                                     the sequencer.
      */
     public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictInfo,
-                       @Nonnull UUID ... streamIDs) {
-
+                       @Nonnull CacheOption cacheOption, @Nonnull UUID ... streamIDs) {
         // Go to the sequencer, grab an initial token.
         TokenResponse tokenResponse = conflictInfo == null
                 ? runtime.getSequencerView().next(streamIDs) // Token w/o conflict info
@@ -112,7 +112,7 @@ public class StreamsView extends AbstractView {
 
             // Attempt to write to the log
             try {
-                runtime.getAddressSpaceView().write(tokenResponse, object);
+                runtime.getAddressSpaceView().write(tokenResponse, object, cacheOption);
                 // If we're here, we succeeded, return the acquired token
                 return tokenResponse.getTokenValue();
             } catch (OverwriteException oe) {
@@ -162,5 +162,15 @@ public class StreamsView extends AbstractView {
                 Arrays.stream(streamIDs).map(Utils::toReadableId).collect(Collectors.toSet()),
                 ILogData.getSerializedSize(object));
         throw new AppendException();
+    }
+
+    /**
+     * Append to multiple streams and caches the result.
+     *
+     * @see StreamsView#append(Object, TxResolutionInfo, CacheOption, UUID...)
+     */
+    public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictInfo,
+                       @Nonnull UUID ... streamIDs) {
+       return append(object, conflictInfo, CacheOption.WRITE_THROUGH, streamIDs);
     }
 }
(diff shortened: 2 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -158,11 +160,18 @@ public void write(IToken token, Object data) throws OverwriteException {
...
 
         // Cache the successful write
-        if (!runtime.getParameters().isCacheDisabled()) {
+        if (!runtime.getParameters().isCacheDisabled() && cacheOption == CacheOption.WRITE_THROUGH) {
             readCache.put(token.getTokenValue(), ld);
         }
     }
 
+    /**
+     * @see AddressSpaceView#write(IToken, Object, CacheOption)
```
no2chem: Add in the comment that this uses WRITE_THROUGH

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -117,14 +118,15 @@ private void validateStateOfWrittenEntry(long address, @Nonnull ILogData ld) {
...
+     * @param token        The token to use for the write.
+     * @param data         The data to write.
+     * @param cacheOption  The caching behaviour for this write
      * @throws OverwriteException   If the globalAddress given
      *                              by the token has adopted
      *                              another value.
      * @throws WrongEpochException  If the token epoch is invalid.
      */
-    public void write(IToken token, Object data) throws OverwriteException {
+    public void write(IToken token, Object data, CacheOption cacheOption) throws OverwriteException {
```
no2chem: Can you add the null/nonnulll annotations to the signature

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -158,11 +160,18 @@ public void write(IToken token, Object data) throws OverwriteException {
         }, true);
 
         // Cache the successful write
-        if (!runtime.getParameters().isCacheDisabled()) {
+        if (!runtime.getParameters().isCacheDisabled() && cacheOption == CacheOption.WRITE_THROUGH) {
```
no2chem: Interestingly, this is not technically a write through cache, since the write happens to the log unit first, then is added into the cache. But, close enough

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -8,6 +8,7 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import lombok.NonNull;
```
corfu-repo: MINOR Remove this unused import 'lombok.NonNull'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
corfu-repo: MINOR Remove the declaration of thrown exception 'org.corfudb.runtime.exceptions.OverwriteException' which is a runtime exception. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3ARedundantThrowsDeclarationCheck)

corfu-repo:
* ![MINOR][MINOR] 2 minor

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: True! I'm actually not aware of any terminology that captures this ordering. I thought about WRITE_AFTER , but I'm not sure if that would be more confusing. Let me know if you have a better term.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -69,13 +70,13 @@ public IStreamView get(UUID stream, StreamOptions options) {
...
      * @param conflictInfo Conflict information for the sequencer to check.
+     * @param cacheOption  The caching mode for write/append
      * @return The address the entry was written to.
      * @throws TransactionAbortedException If the transaction was aborted by
      *                                     the sequencer.
      */
     public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictInfo,
-                       @Nonnull UUID ... streamIDs) {
-
+                       @Nonnull CacheOption cacheOption, @Nonnull UUID ... streamIDs) {
```
StrikingThirteen: Maybe use lombok's NonNull (already imported) rather than javax's. Otherwise, lombok's import is unnecessary and can be removed

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -163,4 +164,14 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
                 ILogData.getSerializedSize(object));
         throw new AppendException();
     }
+
+    /**
+     * Append to multiple streams and caches the result.
+     *
+     * @see StreamsView#append(Object, TxResolutionInfo, CacheOption, UUID...)
+     */
+    public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictInfo,
```
StrikingThirteen: same here with Nonnull and NonNull

test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
```diff
@@ -84,6 +85,30 @@ public void ensureStripingWorks()
                 .isEmptyAtAddress(0);
     }
 
+    @Test
+    public void testUncachedWrites() {
```
StrikingThirteen: Helpful test, thanks.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
Maithem: done,

test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
annym: For completeness should we add a test that shows that regardless of the Cache Option if the cache is disabled no writes will be cached?

Maithem: Done.

annym: I can't see the case where we r.setCacheDisabled(true); did you push the change?

Maithem: Hmm, I think you are asking for the default write behavior (i.e. without passing a CacheOption, so I added : 

// write with the default write method
        r.getAddressSpaceView().write(new TokenResponse(2, epoch,
                        Collections.singletonMap(CorfuRuntime.getStreamID("stream1"), Address.NO_BACKPOINTER)),
                "payload".getBytes());

Maithem: I'll remove r.setCacheDisabled(false); because that's the point of this change, cache by-pass

Maithem:
********* Scenario report *********
Correctness success: True
Liveness success: True
Number of maps: 50
Number of keys/map: 100
Number of operations: 6341067
Number of Inconsistencies: 0
Inconsistency percentage: 0.0 %
Exceptions counts:

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
ben-manes: You could make it write-through by using `cache.asMap().compute` to perform the operation under the cache's hash-bin lock (~ per key). For example, assuming `doWrite` returns the `LogData` param this could be,

```java
boolean cachable = !runtime.getParameters().isCacheDisabled() 
    && (cacheOption == CacheOption.WRITE_THROUGH);
if (cachable) {
  readCache.asMap().compute(token.getTokenValue(), (key, oldValue) -> doWrite(token, data));
} else {
  doWrite(token, ld);
}
```

If you are okay with always invoking through the cache, then you can simplify it down to,
```java
boolean cachable = !runtime.getParameters().isCacheDisabled() 
    && (cacheOption == CacheOption.WRITE_THROUGH);
readCache.asMap().compute(token.getTokenValue(), (key, oldValue) -> {
    LogData newValue = doWrite(token, data);
    return cachable ? newValue : oldValue;
});
```
where `oldValue` may be `null` if absent.

The benefits of a computation is that no thread interleaving can stampede the entry with a stale value on the `put`. But that's likely already handled by Corfu, so just an fyi rather than recommendation.

Maithem: @ben-manes Thanks for the fyi

is this any different than just using a CacheWriter like:

```java
final LoadingCache<Long, ILogData> readCache = Caffeine.<Long, ILogData>newBuilder()
            .maximumSize(runtime.getParameters().getNumCacheEntries())
            .recordStats()
            .writer(new CacheWriter<Long, ILogData>() {
                @Override
                public void write(@Nonnull Long key, @Nonnull ILogData value) {
                    writeLU(value);
                }

                @Override
                public void delete(@Nonnull Long key,
                                   @Nullable ILogData value,
                                   @Nonnull RemovalCause cause) {

                }
            })
```

ben-manes: You're right, that's likely the same. The writer is called under the lock, too.

The potential difference is that your `layoutHelper` method (aka `writeLU(value)`) is using parameters that the cache might not have (depending on your implementation details). The `token` is set on the `LogData` during the `layoutHelper` call, but the cache key is `token.getTokenValue()`. Most likely you could set it earlier or recreate the token, but I don't know. So you have a capturing lambda, whereas the `CacheWriter` is non-capturing by only knowing the direct inputs. If you need extra state from the environment then the `compute` is more flexible, but if not then it is identical.

PR #1342, opened 2018-06-24
Select responsive failover sequencer.

### LINKED ISSUE #1339: Reassign responsive sequencer on failover.
## Overview
On reassigning a failover sequencer on a node failure, the unresponsive servers list is not considered causing the management server to choose to choose an unresponsive server as the primary sequencer.

https://github.com/CorfuDB/CorfuDB/blob/ee3e8d6b685059b19bba736c1b3460a0f4ae42f6/runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java#L293-L306

zalokhan:
## Overview

Description: 
This PR includes:
-Changed the logic to only select a responsive sequencer server on failover.
-Added UT to show that unresponsive sequencer servers won't be picked on failover.
-Clean up in test (restricting visibility and modifier of test parameters)

Why should this be merged: In the current code, we do not check for the unresponsive servers list and can failover to an unresponsive sequencer which can lead to a cluster liveness issue.

Related issue(s) : Resolves #1339

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
@@ -1,20 +1,16 @@
 package org.corfudb.runtime.view;
 
 import com.google.common.collect.Sets;
-
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-
-import java.util.UUID;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.runtime.exceptions.LayoutModificationException;
 import org.corfudb.runtime.view.Layout.LayoutSegment;
 import org.corfudb.runtime.view.Layout.LayoutStripe;
 
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 /**
  * Allows us to make modifications to a layout.
@@ -295,13 +291,21 @@ public class LayoutBuilder {
         List<String> modifiedSequencerServers = new ArrayList<>(layout.getSequencers());
         for (int i = 0; i < modifiedSequencerServers.size(); i++) {
             String sequencerServer = modifiedSequencerServers.get(i);
-            if (!endpoints.contains(sequencerServer)) {
+
+            // Add a sequencer server given sequencerServer is not already included in
+            // endpoints AND it is not included in the unresponsive nodes of the layout.
+            if (!endpoints.contains(sequencerServer) &&
+                !layout.getUnresponsiveServers().contains(sequencerServer)) {
                 modifiedSequencerServers.remove(sequencerServer);
                 modifiedSequencerServers.add(0, sequencerServer);
                 layout.setSequencers(modifiedSequencerServers);
                 return this;
             }
         }
+
+        log.warn("Failover to a responsive sequencer server failed. Each of the layout " +
+                 "sequencers:{} is either a failed endpoint or an unresponsive server",
+                 modifiedSequencerServers.toString());
         throw new LayoutModificationException("All sequencers failed.");
     }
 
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -295,7 +294,8 @@ public LayoutBuilder assignResponsiveSequencerAsPrimary(Set<String> endpoints) {
         List<String> modifiedSequencerServers = new ArrayList<>(layout.getSequencers());
         for (int i = 0; i < modifiedSequencerServers.size(); i++) {
             String sequencerServer = modifiedSequencerServers.get(i);
-            if (!endpoints.contains(sequencerServer)) {
+            if (!endpoints.contains(sequencerServer)
```
StrikingThirteen: Maybe a one line plain text description would help the maintainers to understand the condition. (e.g. a sequencer server should not be contained or unresponsive one, or similar)

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
```diff
@@ -245,4 +246,55 @@ public void checkAdditionOfNodes() throws Exception {
...
+     * Reassign a failover sequencer server from the set of responsive nodes.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void reassignResponsiveSequencer() throws Exception {
+
+        Layout originalLayout = new TestLayoutBuilder()
+                .setEpoch(1L)
+                .addLayoutServer(SERVERS.PORT_0)
```
StrikingThirteen: Although it is a test, it might be nice to have  one or two helper methods to avoid the repetition of layout building. Your decision as it is worth it or not.

StrikingThirteen:
This PR needs to be merged in order to resolve the bug where unresponsive sequencers being picked at failover. @zalokhan has resolved the issue and created this PR, in his absence I'm doing the final changes, addressing the comments, rebasing and making it ready for merge.

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
StrikingThirteen: Done.

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
StrikingThirteen: I assessed the suggestion of adding helper method. It doesn't carry its weight at current state. This is due to different variations that it has to support for creating different Layouts used in different tests.

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -295,7 +291,11 @@ public LayoutBuilder assignResponsiveSequencerAsPrimary(Set<String> endpoints) {
         List<String> modifiedSequencerServers = new ArrayList<>(layout.getSequencers());
         for (int i = 0; i < modifiedSequencerServers.size(); i++) {
             String sequencerServer = modifiedSequencerServers.get(i);
-            if (!endpoints.contains(sequencerServer)) {
+
+            // Add a sequencer server given sequencerServer is not already included in
+            // endpoints AND it is not included in the unresponsive nodes of the layout.
+            if (!endpoints.contains(sequencerServer) &&
```
medhavidhawan: I think we need a log warning or error when we cannot choose a sequencer ? It is hard to debug if we have a problem in this area.

StrikingThirteen: Thanks for the good point. Added a warning log with relevant information.

StrikingThirteen:
- Addressed @medhavidhawan comments for adding log.
- Added another UT to verify exception when failover is unsuccessful due to servers being unresponsive.

PR #1348, opened 2018-07-05
Expose the node and cluster status.

zalokhan:
## Overview

Description: Change the access modifiers for  Node and Cluster status enums to public.

Why should this be merged: Status not accessible by clients.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/ClusterStatusReport.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ClusterStatusReport.java
@@ -17,17 +17,37 @@ public class ClusterStatusReport {
     /**
      * Connectivity to the node.
      */
-    enum NodeStatus {
+    public enum NodeStatus {
+
+        /**
+         * Node is reachable.
+         */
         UP,
+
+        /**
+         * Node is not reachable.
+         */
         DOWN
     }
 
     /**
      * Health of the cluster.
      */
-    enum ClusterStatus {
+    public enum ClusterStatus {
+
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/ClusterStatusReport.java
```diff
@@ -17,15 +17,15 @@
...
+    public enum NodeStatus {
         UP,
         DOWN
     }
 
     /**
      * Health of the cluster.
      */
-    enum ClusterStatus {
+    public enum ClusterStatus {
```
StrikingThirteen: This also exposes the health value to the clients. Are we ok with committing to stick with these health values?
If not, we have to at least highlight this in their java docs.

StrikingThirteen: Also these enums are becoming part the exposed api, so regardless of the above decision maybe consider adding more comments to their java docs briefly describing each state.

zalokhan:
@StrikingThirteen I have added additional comments to describe each state.

zalokhan:
Yes, it shows up in Javadoc.

Maithem:
Please dont check this in until we have the pre-checkin and perf tests results.

Rtremblay:
This is a very simple fix  that passed all test on the previsou TOT. We are press with time as we have 1350 and 1349  so it is agree with Ragnar to push this one directly

PR #1352, opened 2018-07-06
Avoiding unnecessary copying of servers during server removals

StrikingThirteen:
## Overview

Description:
Currently we copy the layout servers examine them and only remove the suggested endpoints when at least one server remains.
This can be done more concisely and without copying the servers.

Why should this be merged: 
Avoid copying the list of servers
More concise and readable
Update:
Added more comments to all the public api on LayoutBuilder
Added null checks on arguments of public api on LayoutBuilder
Added a Unit Test checking that null arguments are not being accepted on public api of LayoutBuilder

Related issue(s) (if applicable): #1362 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -99,17 +99,17 @@ public LayoutBuilder removeUnresponsiveServer(@Nonnull String endpoint) {
      */
     public LayoutBuilder removeLayoutServer(String endpoint) {
 
-        List<String> layoutServers = layout.getLayoutServers();
```
no2chem: If we're going to change this function, I think we should properly document behavior.

First, endpoint should have a @Nonnull annotation.

Second, What's the behavior supposed to be if the endpoint is not in the layout? The current behavior is that the function silently succeeds. If that's what is desired, could you add that to the comment?

Third, add to the javadoc @throws that LayoutModificationException is thrown if it is the last remaining server.
 
That comment would make the code below more clear.

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -99,17 +99,17 @@ public LayoutBuilder removeUnresponsiveServer(@Nonnull String endpoint) {
...
-                if (layoutServers.size() == 1) {
-                    throw new LayoutModificationException(
-                            "Attempting to remove all layout servers.");
-                }
-                layoutServers.remove(i);
-                return this;
-            }
+        // Only remove the endpoint if it does not attempt to remove the last remaining layout server
+        if (layout.getLayoutServers().size() == 1 &&
+            layout.getLayoutServers().get(0).equals(endpoint)) {
```
no2chem: perhaps contains() would be more clear here.

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
StrikingThirteen: @no2chem The change was intended to keep the behaviors as it was and just improve the internal process. However as you mentioned I think it is a good idea to change the API so that we do not allow null argument. I have done it for this and other public methods on the builder.

RE first point: **Done** (also done for other public methods)
RE seocond point: No changes in behavior happened here. your summarization is correct and adding comment is **Done**.
RE third point: **Done** also added comment for all the missing java docs.

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
StrikingThirteen: I'm a bit hesitant on using contains as I feel **equals** is more expressive in this case. We have just **one** server and we check its equality against endpoint. However, let me know if you feel strongly about it and I'll change it as requested.

StrikingThirteen:
@no2chem addressed the comments. Can you please review.

PR #1350, opened 2018-07-06
Seal before reset of healing node.

zalokhan:
## Overview

Description: Reset the logunit node while `HealNodeWorkflow` after we have sealed. Also, make the reset epoch unaware to allow a stale node to recover. The reset epoch is persisted by the node to prevent multiple resets from wiping the data on the same epoch.

Why should this be merged: A node with a very old epoch can never heal as it fails on reset with a wrong epoch exception.

Related issue(s) (if applicable): Resolves #1340 

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #1359, opened 2018-07-10
Create Actions once - HealNodeWorkflow

### LINKED ISSUE #1357: Create Actions only once per Workflow invocation.
## Overview

`getActions` creates a new Action object every time its invoked.

zalokhan:
## Overview

Description: Avoid creating multiple actions on every`getActions` invocation.

Related issue(s) (if applicable): Resolves #1357 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/AddNodeWorkflow.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/AddNodeWorkflow.java
@@ -55,7 +55,7 @@ public class AddNodeWorkflow implements IWorkflow {
     final UUID id;
 
     @Getter
-    final List<Action> actions;
+    protected List<Action> actions;
 
     /**
      * Creates a new add node workflow from a request.
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/HealNodeWorkflow.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/HealNodeWorkflow.java
@@ -2,10 +2,9 @@ package org.corfudb.infrastructure.orchestrator.workflows;
 
 import static org.corfudb.protocols.wireprotocol.orchestrator.OrchestratorRequestType.HEAL_NODE;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Sets;
 
-import java.util.Arrays;
-import java.util.List;
 import java.util.Set;
 
 import javax.annotation.Nonnull;
@@ -33,18 +32,13 @@ public class HealNodeWorkflow extends AddNodeWorkflow {
     public HealNodeWorkflow(HealNodeRequest healNodeRequest) {
         super(new AddNodeRequest(healNodeRequest.getEndpoint()));
         this.request = healNodeRequest;
+        this.actions = ImmutableList.of(new HealNodeToLayout(),
+                new RestoreRedundancyAndMergeSegments());
     }
 
     @Override
     public String getName() {
         return HEAL_NODE.toString();
-    }
-
-    @Override
-    public List<Action> getActions() {
-        return Arrays.asList(
-                new HealNodeToLayout(),
-                new RestoreRedundancyAndMergeSegments());
     }
 
     /**


infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/AddNodeWorkflow.java
```diff
@@ -55,7 +55,7 @@
     final UUID id;
 
     @Getter
-    final List<Action> actions;
+    protected List<Action> actions;
```
StrikingThirteen: why not keeping it package private?

StrikingThirteen: I know that L46 (protected Layout newLayout) is outside the scope of your change but even that can be package private instead of protected. If you're fine with limiting the visibilities, can you change that too?

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/HealNodeWorkflow.java
```diff
@@ -33,20 +32,15 @@
...
+                new RestoreRedundancyAndMergeSegments());
     }
 
     @Override
     public String getName() {
         return HEAL_NODE.toString();
     }
 
-    @Override
-    public List<Action> getActions() {
```
StrikingThirteen: Did you see any significant time/resource on multiple list creation or it was just proactive improvement?

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/HealNodeWorkflow.java
```diff
@@ -33,20 +32,15 @@
     public HealNodeWorkflow(HealNodeRequest healNodeRequest) {
         super(new AddNodeRequest(healNodeRequest.getEndpoint()));
         this.request = healNodeRequest;
+        this.actions = ImmutableList.of(new HealNodeToLayout(),
```
StrikingThirteen: Good change, thanks.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/AddNodeWorkflow.java
zalokhan: HealNodeWorkflow extends this class and sets this variable. We can't do that if this is private.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/HealNodeWorkflow.java
zalokhan: It was a required item in #1350 
I should have done it in the same PR but overestimated the number of changes required.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/AddNodeWorkflow.java
StrikingThirteen: I was suggesting "package private" or default (i.e. no modifier) not "private". so all the classes in the package have access which is the case here.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/HealNodeWorkflow.java
StrikingThirteen: Thanks. I was just curious.

StrikingThirteen:
@zalokhan Replied on the visibilities comments.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/HealNodeWorkflow.java
Maithem: This is not a performance issue. Its incorrect behavior. Calling getActions twice would return a new set of actions, i.e. the state of the previous actions is lost.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/workflows/AddNodeWorkflow.java
Maithem: actions should be instantiated when the workflow is created, why does it need to be changed to protected ?

PR #1361, opened 2018-07-10
Surface Timeout Exceptions

Maithem:
## Overview
The client workflows were recasting TimeoutException to
RuntimeException as a result retries don't get triggered.

Why should this be merged: Makes workflow retries more robust. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/clients/ManagementClient.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/ManagementClient.java
@@ -4,6 +4,7 @@ import java.util.Collections;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
 
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/ManagementClient.java
```diff
@@ -84,12 +85,12 @@ public ManagementClient(IClientRouter router, long epoch) {
      * @param endpoint the endpoint to add to the cluster
      * @return create workflow response that contains the uuid of the workflow
      */
-    public CreateWorkflowResponse addNodeRequest(@Nonnull String endpoint) {
+    public CreateWorkflowResponse addNodeRequest(@Nonnull String endpoint) throws TimeoutException {
```
StrikingThirteen: Can we update the public APIs javadoc to include the throws as well?

runtime/src/main/java/org/corfudb/runtime/clients/ManagementClient.java
```diff
@@ -116,33 +117,33 @@ public CreateWorkflowResponse healNodeRequest(@Nonnull String endpoint,
...
+        return (CreateWorkflowResponse) CFUtils.getUninterruptibly(resp, TimeoutException.class).getResponse();
     }
 
     /**
      * Query the state of a workflow on a particular orchestrator.
      * @param workflowId the workflow to query
      * @return Query response that contains whether the workflow is running or not
      */
-    public QueryResponse queryRequest(@Nonnull UUID workflowId) {
+    public QueryResponse queryRequest(@Nonnull UUID workflowId) throws TimeoutException {
```
StrikingThirteen: Same here and also in the removeNode

StrikingThirteen:
@Maithem it's outside the scope of this PR but it's nice that at some point in future we go towards having UT/IT using mocks to exercise for instance throwing such exceptions. What do you think?

PR #1364, opened 2018-07-11
Fixed Verify Order

Maithem:
Description:

After a Network/Timeout Exception, the runtime should veirfy
the result of the workflow before retrying.

Why should this be merged: After a timeout exception, the action might have completed, so verifying the result before retrying is more efficient. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
@@ -116,20 +116,18 @@ public abstract class WorkflowRequest {
                 Layout requestLayout = new Layout(runtime.getLayoutView().getLayout());
                 UUID workflowId = sendRequest(requestLayout);
                 ManagementClient orchestrator = getOrchestrator(requestLayout);
-
                 waitForWorkflow(workflowId, orchestrator, timeout, pollPeriod);
-
-                for (int y = 0; y < runtime.getParameters().getInvalidateRetry(); y++) {
-                    runtime.invalidateLayout();
-                    Layout layoutToVerify = new Layout(runtime.getLayoutView().getLayout());
-                    if (verifyRequest(layoutToVerify)) {
-                        log.info("WorkflowRequest: Successfully completed {}", this);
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -116,20 +116,18 @@ public void invoke() {
...
-                for (int y = 0; y < runtime.getParameters().getInvalidateRetry(); y++) {
-                    runtime.invalidateLayout();
-                    Layout layoutToVerify = new Layout(runtime.getLayoutView().getLayout());
-                    if (verifyRequest(layoutToVerify)) {
-                        log.info("WorkflowRequest: Successfully completed {}", this);
-                        return;
-                    }
+            for (int y = 0; y < runtime.getParameters().getInvalidateRetry(); y++) {
+                runtime.invalidateLayout();
+                Layout layoutToVerify = new Layout(runtime.getLayoutView().getLayout());
```
StrikingThirteen: I know you haven't touched this logic but this defensive copying of layout might be eliminated. It seems that none of the verfiyRequest implementations are mutating the layout (as expected from a verifying method).

Maithem: The layout from line 116 can't be used because it will be the same one before the operation happens, thus missing one try. For example, copy layout1 -> execute operation -> verify layout1. The verification will run on the same layout before the operation, which doesn't make sense.

StrikingThirteen: yes, L116 doesn't make sense. I was referring to L126's creation of layoutToVerify. so rather than copying a layout from runtime.getLayoutView().getLayout() and verifying the copy, couldn't we have verifyRequest(runtime.getLayoutView().getLayout())?

Maithem: I see. I think its fine since this code path should execute frequently.

Issue #1366, opened 2018-07-12
Dirty Transactions In ThreadLocal


Maithem:
**Bug report.**
Its possible for transactions to be nested, where each transactional context refers to a different CorfuRuntime. For example, an executor thread that abruptly fails before clearing the thread local stack, a new job uses the same thread and starts a transaction with a new CorfuRuntime. This will cause the new Txn to be nested and the parent will refer to previously failed transaction (which refers to the old runtime). This can cause consistency issues and perpetually fail subsequent transactions on that thread (if that runtime is shutdown).

I think we should only nest transactions that belong to the same runtime. For example, If a transaction has a parent with a different Runtime, then we can pop all the transactional contexts and replace them with the more recent (with the new runtime) transaction context, or we can just throw a unrecoverable exception.

zalokhan:
Merged.

PR #1367, opened 2018-07-12
Prevent Incorrect Nesting

Maithem:
## Overview
Prevent transactions from different runtimes to be nested.

Why should this be merged: If this condition is not prevented then transactional consistency can be broken.  

Related issue(s) (if applicable): #1366

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/runtime/view/ObjectsViewTest.java
+++ b/test/src/test/java/org/corfudb/runtime/view/ObjectsViewTest.java
@@ -9,6 +9,7 @@ import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.collections.SMRMap;
 import org.corfudb.runtime.exceptions.TransactionAbortedException;
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
 import org.corfudb.runtime.view.stream.IStreamView;
 import org.corfudb.util.serializer.Serializers;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


test/src/test/java/org/corfudb/runtime/view/ObjectsViewTest.java
```diff
@@ -122,6 +123,19 @@ public void abortedTransactionDoesNotConflict()
         assertThat((String) args[1]).isEqualTo("v2");
     }
 
+    @Test
+    public void incorrectNestingTest() {
```
medhavidhawan: Please add comments as this is subtle stuff.

zalokhan:
@Maithem Should we attempt to address this by each Runtime having its own stack and then not having to worry about this scenario?

Maithem:
@zalokhan Maybe, but what would be the use case? I think that is more confusing: being able to start two disjoint transactions to two systems at the same time.

zalokhan:
I think this would be a common scenario. An application can have a system which has 2 Runtimes in 2 threads and attempting their respective transactions for reasons such as the threads needing to operate on their own streams.

Maithem:
@zalokhan I see, but this change doesn't prevent the case you just mentioned. You can have two threads, starting two transactions on separate clients. What this PR prevents. Is starting two disjoint transactions on separate clients at the same time (i.e. starting transactions concurrently).

zalokhan:
@Maithem Right the transaction context is local to the thread.

test/src/test/java/org/corfudb/runtime/view/ObjectsViewTest.java
```diff
@@ -122,6 +123,25 @@ public void abortedTransactionDoesNotConflict()
...
+        CorfuRuntime r1 = getDefaultRuntime();
+        CorfuRuntime r2 = CorfuRuntime.fromParameters(CorfuRuntime.CorfuRuntimeParameters.builder()
+                .nettyEventLoop(NETTY_EVENT_LOOP)
+                .build());
+        SMRMap<String, String> map = getDefaultRuntime().getObjectsView()
```
StrikingThirteen: If we use the already declared r1 here (L132) to call the getObjectsView, the code would be more expressive.

test/src/test/java/org/corfudb/runtime/view/ObjectsViewTest.java
```diff
@@ -122,6 +123,25 @@ public void abortedTransactionDoesNotConflict()
...
+                .setStreamName("mapa")
+                .setTypeToken(new TypeToken<SMRMap<String, String>>() {})
+                .open();
+
+        r1.getObjectsView().TXBegin();
+        map.get("key1");
+
+        // Try to start a new nested transaction with a different runtime
+        assertThatThrownBy(() -> r2.getObjectsView().TXBegin()).isInstanceOf(UnrecoverableCorfuError.class);
+    }
```
StrikingThirteen: maybe ending the transaction on r1 after the assert?

Issue #1371, opened 2018-07-13
Node Failure Cause Transactions to Fail


Maithem:
**Bug report.** 

Node failure cause transactions to fail without being retried. 

0;3901:02:20.467 1;31ERROR0;39 [Thread-1] 36o.c.r.v.ObjectsView0;39 - TXEnd[TX[6197]]: Unexpected exception
0;39 org.corfudb.runtime.exceptions.ShutdownException: nullNORMAL_OUTPUT
0;39	at org.corfudb.runtime.clients.BaseHandler.handleShutdownException(BaseHandler.java:155)NORMAL_OUTPUT
0;39	at org.corfudb.runtime.clients.ClientMsgHandler.handle(ClientMsgHandler.java:76)NORMAL_OUTPUT
0;39	at org.corfudb.runtime.clients.IClient.handleMessage(IClient.java:40)NORMAL_OUTPUT
ERROR_OUTPUTException in thread "Thread-1" org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError: Unexpected exception during commit
ERROR_OUTPUT	at org.corfudb.runtime.view.ObjectsView.TXEnd(ObjectsView.java:172)
ERROR_OUTPUT	at org.corfudb.integration.ClusterRandomFaultsIT.lambda$spawnWriterThread$0(ClusterRandomFaultsIT.java:83)
ERROR_OUTPUT	at java.lang.Thread.run(Thread.java:748)
ERROR_OUTPUTCaused by: org.corfudb.runtime.exceptions.ShutdownException
ERROR_OUTPUT	at org.corfudb.runtime.clients.BaseHandler.handleShutdownException(BaseHandler.java:155)
ERROR_OUTPUT	at org.corfudb.runtime.clients.ClientMsgHandler.handle(ClientMsgHandler.java:76)
ERROR_OUTPUT	at org.corfudb.runtime.clients.IClient.handleMessage(IClient.java:40)
ERROR_OUTPUT	at org.corfudb.runtime.clients.NettyClientRouter.channelRead0(NettyClientRouter.java:570)
ERROR_OUTPUT	at org.corfudb.runtime.clients.NettyClientRouter.channelRead0(NettyClientRouter.java:65)
ERROR_OUTPUT	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
[...]

PR #1376, opened 2018-07-16
RPC Failure On Unavailable Servers

Maithem:
## Overview
This fix addresses the case when an rpc fails without retrying
when it reaches a server that is shutting down.

Why should this be merged: Prevents rpc failure some nodes are unavailable. 

Related issue(s) (if applicable): #1371

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuMsgHandler.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuMsgHandler.java
@@ -219,7 +219,7 @@ public class CorfuMsgHandler {
         return (msg, ctx, r) -> {
             if (server.isShutdown()) {
                 log.warn("Server received {} but is shutdown.", msg.getMsgType().toString());
-                r.sendResponse(ctx, msg, CorfuMsgType.ERROR_SHUTDOWN_EXCEPTION.msg());
+                r.sendResponse(ctx, msg, CorfuMsgType.ERROR_SERVER_UNAVAILABLE.msg());
                 return;
             }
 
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -114,6 +115,8 @@ public Layout getCurrentLayout() {
                             + "invalidate view", we.getCorrectEpoch());
                 } else if (re instanceof NetworkException) {
                     log.warn("layoutHelper: System seems unavailable", re);
+                } else if (re instanceof ShutdownException) {
```
Maithem: Should we introduce a different exception (i.e. ServerUnavailable?) because this one is shared by the CorfuRuntime shutdown exception (i.e. shouldn't be retried)

StrikingThirteen: Setting aside the TimeoutException, why can't we use catch clauses for catching and dealing with different exceptions (instead of manually checking the types)?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuMsgHandler.java
```diff
@@ -219,7 +219,7 @@ private void registerMethod(@Nonnull final MethodHandles.Lookup caller,
         return (msg, ctx, r) -> {
             if (server.isShutdown()) {
                 log.warn("Server received {} but is shutdown.", msg.getMsgType().toString());
-                r.sendResponse(ctx, msg, CorfuMsgType.ERROR_SHUTDOWN_EXCEPTION.msg());
+                r.sendResponse(ctx, msg, CorfuMsgType.ERROR_SERVER_UNAVAILABLE.msg());
```
Maithem: Should we collapse not ready and server unavailable ?

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -114,6 +115,8 @@ public Layout getCurrentLayout() {
                             + "invalidate view", we.getCorrectEpoch());
                 } else if (re instanceof NetworkException) {
                     log.warn("layoutHelper: System seems unavailable", re);
+                } else if (re instanceof UnavailableServerException) {
```
Maithem: can we somehow log the layout on failures?

Issue #1384, opened 2018-07-19
Force Remove Fails


Maithem:
**Bug report.** 

Force remove fails when removing a node that is shutdown. For example, remove 2 dead nodes from a 3 node cluster and removing a dead node from a 2 node cluster.

PR #1385, opened 2018-07-19
Force remove bug

Maithem:
## Overview
The workflow requests should select a random orchestrator to prevent
selecting dead nodes (in the case of a stale layout).

Why should this be merged: Fixes a bug that blocks force remove requests

Related issue(s) (if applicable): #1384

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
@@ -3,6 +3,7 @@ package org.corfudb.runtime.view.workflows;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.orchestrator.CreateWorkflowResponse;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.clients.ManagementClient;
 import org.corfudb.runtime.view.Layout;
 
 import javax.annotation.Nonnull;
@@ -31,12 +32,12 @@ public class AddNode extends WorkflowRequest {
     }
 
     @Override
-    protected UUID sendRequest(@Nonnull Layout layout) throws TimeoutException {
+    protected UUID sendRequest(@Nonnull ManagementClient managementClient) throws TimeoutException {
         // Select the current tail node and send an add node request to the orchestrator
-        CreateWorkflowResponse resp = getOrchestrator(layout).addNodeRequest(nodeForWorkflow);
-        log.info("sendRequest: requested to add {} on orchestrator {}:{}, layout {}",
-                nodeForWorkflow, getOrchestrator(layout).getRouter().getHost(),
-                getOrchestrator(layout).getRouter().getPort(), layout);
+        CreateWorkflowResponse resp = managementClient.addNodeRequest(nodeForWorkflow);
+        log.info("sendRequest: requested to add {} on orchestrator {}:{}",
+                nodeForWorkflow, managementClient.getRouter().getHost(),
+                managementClient.getRouter().getPort());
         return resp.getWorkflowId();
     }
 
--- a/runtime/src/main/java/org/corfudb/runtime/view/workflows/ForceRemoveNode.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/workflows/ForceRemoveNode.java
@@ -3,6 +3,7 @@ package org.corfudb.runtime.view.workflows;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.orchestrator.CreateWorkflowResponse;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.clients.ManagementClient;
 import org.corfudb.runtime.view.Layout;
 
 import javax.annotation.Nonnull;
@@ -27,12 +28,12 @@ public class ForceRemoveNode extends RemoveNode {
     }
 
     @Override
-    protected UUID sendRequest(@Nonnull Layout layout) throws TimeoutException {
+    protected UUID sendRequest(@Nonnull ManagementClient managementClient) throws TimeoutException {
         // Select the current tail node and send an add node request to the orchestrator
-        CreateWorkflowResponse resp = getOrchestrator(layout).forceRemoveNode(nodeForWorkflow);
-        log.info("sendRequest: requested to force remove {} on orchestrator {}:{}, layout {}",
-                nodeForWorkflow, getOrchestrator(layout).getRouter().getHost(),
-                getOrchestrator(layout).getRouter().getPort(), layout);
+        CreateWorkflowResponse resp = managementClient.forceRemoveNode(nodeForWorkflow);
+        log.info("sendRequest: requested to force remove {} on orchestrator {}:{}",
+                nodeForWorkflow, managementClient.getRouter().getHost(),
+                managementClient.getRouter().getPort());
         return resp.getWorkflowId();
     }
 
--- a/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
@@ -3,6 +3,7 @@ package org.corfudb.runtime.view.workflows;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.clients.BaseClient;
 import org.corfudb.runtime.clients.ManagementClient;
 import org.corfudb.runtime.exceptions.NetworkException;
 import org.corfudb.runtime.exceptions.WorkflowException;
@@ -14,6 +15,7 @@ import javax.annotation.Nonnull;
 import java.time.Duration;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
@@ -39,15 +41,19 @@ public abstract class WorkflowRequest {
     /**
      * Send a workflow request
      *
-     * @param layout current layout
+     * @param client the client to the selected orchestrator
      * @return a uuid that corresponds to the created workflow
      */
-    protected abstract UUID sendRequest(Layout layout) throws TimeoutException;
+    protected abstract UUID sendRequest(ManagementClient client) throws TimeoutException;
 
     /**
      * Select an orchestrator and return a client. Orchestrator's that
      * are on unresponsive nodes and the affected endpoint by the workflow
-     * will not be selected.
+     * will not be selected. The layout might not reflect the state
+     * of responsive servers, so we ping the endpoint before we select it.
+     * An example of this would be a 3 node cluster, with two nodes that
+     * die immediately, the layout won't have those two nodes as unresponsive
+     * because it can't commit to a quorum.
      *
      * @param layout the layout that is used to select an orchestrator
      * @return a management client that is connected to the selected
@@ -63,8 +69,22 @@ public abstract class WorkflowRequest {
             throw new WorkflowException("getOrchestrator: no available orchestrators " + layout);
         }
 
-        return runtime.getLayoutView().getRuntimeLayout(layout)
+        // Select an available orchestrator
+        ManagementClient managementClient = runtime.getLayoutView().getRuntimeLayout(layout)
                 .getManagementClient(activeLayoutServers.get(0));
+
+        for (String endpoint : activeLayoutServers) {
+            BaseClient client = runtime.getLayoutView().getRuntimeLayout(layout)
+                    .getBaseClient(endpoint);
+            if (client.pingSync()) {
+                log.info("getOrchestrator: orchestrator selected {}, layout {}", endpoint, layout);
+                managementClient = runtime.getLayoutView().getRuntimeLayout(layout)
+                        .getManagementClient(endpoint);
+                break;
+            }
+        }
+
+        return managementClient;
     }
 
     /**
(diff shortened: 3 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -39,10 +40,10 @@
...
      *
-     * @param layout current layout
+     * @param client the client to the selected orchestrator
      * @return a uuid that corresponds to the created workflow
      */
-    protected abstract UUID sendRequest(Layout layout) throws TimeoutException;
+    protected abstract UUID sendRequest(ManagementClient client) throws TimeoutException;
 
     /**
      * Select an orchestrator and return a client. Orchestrator's that
```
zalokhan: Can you add a comment saying that we pick the Orchestrator randomly?

test/src/test/java/org/corfudb/integration/Harness.java
```diff
@@ -0,0 +1,98 @@
...
+ * <p>
+ * <p>Created by maithem on 7/18/18.
+ */
+
+public class Harness {
+
+    String localAddress = "localhost";
+
+    private String getAddressForNode(int port) {
+        return "localhost:900" + port;
```
medhavidhawan: nit:i would just keep it as localhost  and ports can be 9000, 9001, 9002. Just easier to read.

test/src/test/java/org/corfudb/integration/Harness.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private Layout getLayoutForNodes(int n) {
```
medhavidhawan: can you provide a comment so that we know what the layout should look like.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
```diff
@@ -273,4 +276,27 @@ public void clusterResizingTest2() throws Exception {
 
         shutdownCorfuServer(p0);
     }
+
+    @Test
+    public void forceRemoveTwoDeadNodes() throws Exception {
```
zalokhan: This is very similar to clusterResizingTest2. Can we merge them?

test/src/test/java/org/corfudb/integration/Harness.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return a list of provisioned nodes in the cluster
+     * @throws IOException
+     */
+    public List<Node> deployCluster(int n) throws IOException {
+        List<Node> nodes = new ArrayList<>(n);
+        String conn = getClusterConnectionString(n);
+        for (int i = 0; i < n; i++) {
+            Process proc = new AbstractIT.CorfuServerRunner()
+                    .setHost(localAddress)
+                    .setPort(9000 + i)
```
medhavidhawan: i think the port is off. [90000, 90001,] etc. Will be difficult to read.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
medhavidhawan: great test can you provide high level comments.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
```diff
@@ -273,4 +276,27 @@ public void clusterResizingTest2() throws Exception {
...
+
+        assertThat(rt.getLayoutView().getLayout().getAllServers().size()).isEqualTo(numNodes);
+
+        // Shutdown two nodes
+        run(n0.shutdown, n1.shutdown);
+
+        rt.getManagementView().forceRemoveNode(n0.getAddress(), workflowNumRetry, timeout, pollPeriod);
+        rt.getManagementView().forceRemoveNode(n1.getAddress(), workflowNumRetry, timeout, pollPeriod);
+
+        assertThat(rt.getLayoutView().getLayout().getAllServers().size()).isEqualTo(1);
```
medhavidhawan: could you also make sure the correct expected layout is present.

test/src/test/java/org/corfudb/integration/cluster/Harness/Node.java
```diff
@@ -0,0 +1,43 @@
...
+    @Getter
+    final String address;
+
+    @Getter
+    final String clusterAddress;
+
+    @Getter
+    final String logPath;
+
+    final public Action shutdown;
```
medhavidhawan: Another idea would be to make these Actions enums. You can then define all the actions in one place and easier to read the implementation in one place. Do it if you find it easier that way. I am ok either ways.

test/src/test/java/org/corfudb/integration/cluster/Harness/PauseAction.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+public class PauseAction extends Action {
+
+    public PauseAction(Node node) {
+        this.node = node;
+    }
+
+    @Override
+    public void run() {
```
medhavidhawan: no implementation ?

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -63,8 +64,14 @@ protected ManagementClient getOrchestrator(Layout layout) {
             throw new WorkflowException("getOrchestrator: no available orchestrators " + layout);
         }
 
+        int ind = ThreadLocalRandom.current()
```
medhavidhawan: could you please add comments to this. Especially to the fact that activeLayoutServers list might not be correct as it is not possible to change layout once we have lost quorum. Also another possible approach could be to ping and check which server is up before assigning it.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
zalokhan: And I think these 2 tests are different because they force remove different sets of nodes. We can leave both tests in that case.

Maithem: The problem our current tests didnt catch this case was because the force remove was sensitive to the node order in the layout.

test/src/test/java/org/corfudb/integration/Harness.java
Maithem: done

test/src/test/java/org/corfudb/integration/Harness.java
Maithem: done.

test/src/test/java/org/corfudb/integration/Harness.java
Maithem: this is an int addition and not a string concatenation. It will create 9000, 9001, 9002 etc..

test/src/test/java/org/corfudb/integration/cluster/Harness/PauseAction.java
Maithem: Removed. Can be implemented later.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
Maithem: done.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
Maithem: done.

Maithem:
Addressed all comments.

runtime/src/main/java/org/corfudb/runtime/view/workflows/ForceRemoveNode.java
```diff
@@ -27,12 +28,12 @@ public ForceRemoveNode(@Nonnull String endpointToRemove, @Nonnull CorfuRuntime r
...
     @Override
-    protected UUID sendRequest(@Nonnull Layout layout) throws TimeoutException {
+    protected UUID sendRequest(@Nonnull ManagementClient managementClient) throws TimeoutException {
         // Select the current tail node and send an add node request to the orchestrator
-        CreateWorkflowResponse resp = getOrchestrator(layout).forceRemoveNode(nodeForWorkflow);
+        CreateWorkflowResponse resp = managementClient.forceRemoveNode(nodeForWorkflow);
         log.info("sendRequest: requested to force remove {} on orchestrator {}:{}, layout {}",
-                nodeForWorkflow, getOrchestrator(layout).getRouter().getHost(),
-                getOrchestrator(layout).getRouter().getPort(), layout);
+                nodeForWorkflow, managementClient.getRouter().getHost(),
```
annym: Remove layout from log message.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -63,8 +68,16 @@ protected ManagementClient getOrchestrator(Layout layout) {
...
         }
 
+        // Randomly select an orchestrator, just in-case the unresponsive servers
+        // doesn't reflect the current state of the system
+        int ind = ThreadLocalRandom.current()
+                .nextInt(activeLayoutServers.size()) % activeLayoutServers.size();
+
+        log.info("getOrchestrator: orchestrator selected {}, layout {}",
+                activeLayoutServers.get(ind), layout);
+
```
annym: will we consider verifying in advance if the selected node is responsive?

Maithem: Yea, I think as the number of unresponsive nodes increase, selecting one at random might not be the best strategy. @medhavidhawan  suggests that we ping.

runtime/src/main/java/org/corfudb/runtime/view/workflows/ForceRemoveNode.java
Maithem: done.

Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -63,8 +69,22 @@ protected ManagementClient getOrchestrator(Layout layout) {
             throw new WorkflowException("getOrchestrator: no available orchestrators " + layout);
         }
 
-        return runtime.getLayoutView().getRuntimeLayout(layout)
+        // Select an available orchestrator
+        ManagementClient managementClient = runtime.getLayoutView().getRuntimeLayout(layout)
                 .getManagementClient(activeLayoutServers.get(0));
```
annym: Why do we select an orchestrator in advance? if it will be picked further in the loop..

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -63,8 +69,22 @@ protected ManagementClient getOrchestrator(Layout layout) {
...
+
+        for (String endpoint : activeLayoutServers) {
```
annym: should we shuffle this list? or we might always end with the same orchestrator...

Maithem: If the orchestrator is not available (i.e. pingSync == false) then how can this happen?
This loop will ping through the active servers set and select the first one that will respond to the ping.

annym: I thought we wanted to have some randomness in the selection of the orchestrator, it is true that it will pick the one which pingSync == true, but chances are that if nothing goes wrong you will always be picking the first one, instead of giving the chance to any other..

zalokhan: @annym We would like to avoid randomness for simplifying debugging with deterministic behaviour.

Maithem: The randomness was introduced to discover a node that is available, using the ping achieves that and a better strategy when you have more nodes that are dead (i.e. less retries) .

annym: ok perfect, I thought that we wanted to make the selection somehow random.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
Maithem: just a "default" orchestrator instead of null;

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
```diff
@@ -31,12 +32,12 @@ public AddNode(@Nonnull String endpointToAdd, @Nonnull CorfuRuntime runtime,
...
     @Override
-    protected UUID sendRequest(@Nonnull Layout layout) throws TimeoutException {
+    protected UUID sendRequest(@Nonnull ManagementClient managementClient) throws TimeoutException {
         // Select the current tail node and send an add node request to the orchestrator
-        CreateWorkflowResponse resp = getOrchestrator(layout).addNodeRequest(nodeForWorkflow);
-        log.info("sendRequest: requested to add {} on orchestrator {}:{}, layout {}",
-                nodeForWorkflow, getOrchestrator(layout).getRouter().getHost(),
-                getOrchestrator(layout).getRouter().getPort(), layout);
+        CreateWorkflowResponse resp = managementClient.addNodeRequest(nodeForWorkflow);
+        log.debug("sendRequest: requested to add {} on orchestrator {}:{}",
```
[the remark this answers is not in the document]
Maithem: done.

PR #1388, opened 2018-07-20
Use Native Implementation for SSL

Maithem:
## Overview
This patch changes the ssl context constructor to use a native
engine implementation. My test show 4.5x improvement on server
and client restart.

Why should this be merged: Improves client/sequencer reboot performance when TLS is used.

- [x] Public API has Javadoc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,12 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-tcnative-boringssl-static</artifactId>
+            <version>2.0.7.Final</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.codehaus.mojo</groupId>
             <artifactId>findbugs-maven-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -149,6 +149,12 @@
             <version>4.1.18.Final</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
bitmingw: In https://netty.io/wiki/forked-tomcat-native.html it has a `classifier` field. Why do you think it can be missing here?

bitmingw: Refer this as the full instruction. https://netty.io/wiki/forked-tomcat-native.html#how-to-download-netty-tcnative-boringssl-static

bitmingw:
I think it works like this.

https://netty.io/wiki/forked-tomcat-native.html#how-to-download-netty-tcnative-boringssl-uber-jar is the instruction of how corfu should be setup.

https://netty.io/wiki/forked-tomcat-native.html#how-to-download-netty-tcnative-boringssl-static is the instruction of how NSX should be setup.

Issue #1389, opened 2018-07-23
Remove FastObjectLoader's dependency on Sequencer


zalokhan:
## Overview
The FastObjectLoader depends on the SequencerView to fetch the globalLogTail. This can be achieved by polling the log unit servers' tail(s) (Similar to the implementation in the LayoutManagementView.)

This can provide an improvement in startup times in cases where the server and client both restart at the same time. Currently, the client's fastObjectLoader would wait for the sequencer recovery, then get the tails and then recover. However, with this fix, the client can recover in parallel with the sequencer recovery.

PR #1392, opened 2018-07-25
Remove FastObjectLoader Dependency on Sequencer

annym:
## Overview

Description: Now the FastObjectLoader will obtain the log globalTail by polling the Log Unit servers instead of relying on the sequencers view, which in turn needs to wait for the sequencer to
recover. This will allow clients and servers restarting at the same time to run in parallel rather than having to wait for the sequencer to totally recover.

Why should this be merged: This fixes a performance issue (#1389) and is required also in the effort of optimizing the stream tail recreation on sequencer bootstrap, which is currently lead by the Layout Servers and will be moved into the sequencer server.

Related issue(s) (if applicable): #1389

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
@@ -1,6 +1,16 @@
 package org.corfudb.recovery;
 
-import javax.annotation.Nonnull;
+import static org.corfudb.recovery.RecoveryUtils.createObjectIfNotExist;
+import static org.corfudb.recovery.RecoveryUtils.deserializeLogData;
+import static org.corfudb.recovery.RecoveryUtils.getCorfuCompileProxy;
+import static org.corfudb.recovery.RecoveryUtils.getLogData;
+import static org.corfudb.recovery.RecoveryUtils.getSnapShotAddressOfCheckPoint;
+import static org.corfudb.recovery.RecoveryUtils.getStartAddressOfCheckPoint;
+import static org.corfudb.recovery.RecoveryUtils.isCheckPointEntry;
+import static org.corfudb.runtime.view.Address.isAddress;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -14,12 +24,14 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import javax.annotation.Nonnull;
+
 import lombok.Data;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
+
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.LogEntry;
 import org.corfudb.protocols.logprotocol.MultiObjectSMREntry;
@@ -38,9 +50,6 @@ import org.corfudb.util.CFUtils;
 import org.corfudb.util.Utils;
 import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
-
-import static org.corfudb.recovery.RecoveryUtils.*;
-import static org.corfudb.runtime.view.Address.isAddress;
 
 /** The FastObjectLoader reconstructs the coalesced state of SMRMaps through sequential log read
  *
@@ -78,13 +87,11 @@ public class FastObjectLoader {
     @Getter
     private boolean loadInCache;
 
-    @Setter
     @Getter
-    private long logHead = -1;
+    private long logHead = Address.NON_EXIST;
 
-    @Setter
     @Getter
-    private long logTail = -1;
+    private long logTail = Address.NON_EXIST;
 
     @Setter
     @Getter
@@ -104,6 +111,12 @@ public class FastObjectLoader {
 
     private boolean whiteList = false;
     private List<UUID> streamsToLoad = new ArrayList<>();
+
+    @VisibleForTesting
+    void setLogHead(long head) { this.logHead = head; }
+
+    @VisibleForTesting
+    void setLogTail(long tail) { this.logTail = tail; }
 
     /**
      * Enable whiteList mode where we only reconstruct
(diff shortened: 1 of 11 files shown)


runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -460,14 +418,19 @@ public void reconfigureSequencerServers(Layout originalLayout, Layout newLayout,
...
                     fastObjectLoader.setRecoverSequencerMode(true);
                     fastObjectLoader.setLoadInCache(false);
 
                     // FastSMRLoader sets the logHead based on trim mark.
-                    fastObjectLoader.setLogTail(maxTokenRequested);
+
+                    // This is specifically set to 0 if maxTokenRequested is -1 because we do not
+                    // want the FastLoader to calculate its own tail based on the Runtime's layout
+                    // and want to calculate it on the newLayout achieved by consensus.
+                    long tail = maxTokenRequested == Address.NON_ADDRESS ? 0L : maxTokenRequested;
```
zalokhan: You would want to remove this right. The FastLoader has a default of -6 now to support this. Also update the comment accordingly.
And also initialize `maxTokenRequested` to -1 instead?

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -454,4 +463,47 @@ public static void printLogAnatomy(CorfuRuntime runtime, ILogData logData) {
                     logData.getGlobalAddress());
         }
     }
+
+
+    /**
+     * Fetches the max global log tail from the log unit cluster. This depends on the mode of
+     * replication being used.
+     * CHAIN: Block on fetch of global log tail from the head log unitin every stripe.
```
zalokhan: unitin -> unit in

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -248,7 +258,10 @@ private void findAndSetLogHead() {
     }
 
     private void findAndSetLogTail() {
-        logTail = runtime.getSequencerView().query().getTokenValue();
+        Layout newLayout = runtime.getLayoutView().getLayout();
+        Layout.LayoutSegment latestSegment = newLayout.getSegments()
+                .get(newLayout.getSegments().size() - 1);
+        logTail = getMaxGlobalTail(newLayout, latestSegment, runtime);
```
zalokhan: Can we add a log statement here? Displaying both the log tail and the epoch at which was calculated.
And one for the log head calculated too. This can help us catch bugs like the hole-filling one.

annym: Done

annym:
@zalokhan Addressed all comments.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -248,7 +258,11 @@ private void findAndSetLogHead() {
     }
 
     private void findAndSetLogTail() {
-        logTail = runtime.getSequencerView().query().getTokenValue();
+        Layout newLayout = runtime.getLayoutView().getLayout();
```
Maithem: this logic (i.e. get last segment) seems to be in multiple places, can we push it to layout ?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -473,11 +487,11 @@ private void updateCorfuObject(ILogData logData) {
      *
      */
     private void initializeHeadAndTails() {
-        if (logHead < 0) {
+        if (logHead == Address.NON_EXIST) {
```
Maithem: why do we even have this logic? shouldn't we always initialize the head/tail from the LUs ? Why are there setters on these variables. Can we clean this up?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
Maithem: can getMaxGlobalTail throw a network exception ? where is this retried ?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
annym: Setters seem to be used in the LayoutManagementView, to take advantage of the fact that the globalTail was requested already (I did change this to compute it within the FOL) and then retrieve the value which is further used for the sequencer reset. But it is also used to force tail and head in FOL Unit Tests. So this logic is aiming to bypass the check against log units.

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -1,11 +1,11 @@
 package org.corfudb.runtime.view;
 
+import static org.corfudb.util.Utils.getMaxGlobalTail;
```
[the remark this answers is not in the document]
annym: Done

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -460,14 +418,16 @@ public void reconfigureSequencerServers(Layout originalLayout, Layout newLayout,
                         .equals(newLayout.getPrimarySequencer())) {
                     Layout.LayoutSegment latestSegment = newLayout.getSegments()
```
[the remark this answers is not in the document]
annym: Done

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
annym: Done

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
annym: I removed the setters and declared these apart annotated as @VisibleForTesting

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
annym: Yes it could, moved getMaxGlobalTail logic to AddressView so we could rely on the layouthelper to retry and run systemdownHandler in case we fail to ever connect and fetch the log tail...

annym:
@Maithem @zalokhan addressed all comments.

corfu-repo:

1. ![CRITICAL][CRITICAL] [Utils.java#L287](https://github.com/CorfuDB/CorfuDB/blob/4450806d92ed2c295f151d13b09b5b5ee4f9e57d/runtime/src/main/java/org/corfudb/util/Utils.java#L287): Make sure deserializing objects is safe here. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS4508)

[CRITICAL]: https://sonarsource.github.io/sonar-github/severity-critical.png 'Severity: CRITICAL'

Maithem:
probably need to rebase.

Issue #1399, opened 2018-07-27
Transaction Aborted Exception Overload of Conflict Cause


annym:
## Overview

If a transaction is aborted given that we got an overwrite exception on all 'retry' attempts (maybe because the writer is slow and a faster reader is hole filling) we report a TransactionAbortedException with CONFLICT cause and null conflict key/stream id. This is not properly handled by the application layer. We should properly report the cause of this abort to OVERWRITE. Meaning what actually happened we got overwritten on all our attempts to write to the log.

Maithem:
Overwrites should be handled by the stream layer. I don't think transactions should fail with an overwrite cause.

PR #1401, opened 2018-07-28
Transaction Abort Exception Cause for slow writers

annym:
## Overview

The cause CONFLICT is reported whenever two transaction conflict over
the same keys for a given stream(s). This change corrects the overload
of this abort cause for the case of slow writers, i.e., whenever a
writer fails to write given that the designated address is continuously
overwritten (perhaps hole filled) by a faster reader. This is due to the
fact that it is not really conflicting with any other stream, the cause
of abort is that you are overwritten by a fast reader in the system.

Why should this be merged: this is an issue that has been reported by the Application Layer as conflicting keys/streams show up being NULL which makes no sense out of a Conflicting Transaction for upper layers.

Related issue(s) (if applicable): #1399

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.util.List;
 
 import org.corfudb.protocols.wireprotocol.LogData;
+import org.corfudb.runtime.exceptions.OverwriteCause;
 
 /**
  * An interface definition that specifies an api to interact with a StreamLog.
@@ -86,4 +87,34 @@ public interface StreamLog {
      * Clears all data and resets all segment handlers.
      */
     void reset();
+
+    /**
+     * Get overwrite cause for a given address.
+     *
+     * @param address global log address
+     * @param entry entry which would cause the overwrite
+     * @return (OverwriteCause) Cause of the overwrite
+     */
+    default OverwriteCause getOverwriteCauseForAddress(long address, LogData entry) {
+        LogData currentEntry = read(address);
+        OverwriteCause cause = OverwriteCause.DIFF_DATA;
+
+        if (currentEntry != null) {
+            if (currentEntry.isHole()) {
+                cause = OverwriteCause.HOLE;
+            } else if (entry.getData() != null && currentEntry.getData() != null &&
+                    currentEntry.getData().length == entry.getData().length) {
+                // If the entry is already present and it is not a hole, the write
+                // might have been propagated by a fast reader from part of the chain.
+                // Compare based on data length. Based on this info client will do an actual
+                // verification on the data
+                cause = OverwriteCause.SAME_DATA;
+            }
+        } else {
+            // No actual entry is found in this address, there is no apparent cause
+            // for the overwrite exception
+            cause = OverwriteCause.NONE;
+        }
+        return cause;
+    }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/exceptions/OverwriteException.java
+++ b/runtime/src/main/java/org/corfudb/runtime/exceptions/OverwriteException.java
@@ -1,7 +1,15 @@
 package org.corfudb.runtime.exceptions;
 
+import lombok.Getter;
+
 /**
  * Created by mwei on 12/14/15.
  */
+@Getter
 public class OverwriteException extends LogUnitException {
+   private OverwriteCause overWriteCause;
+
+    public OverwriteException(OverwriteCause cause) {
+        this.overWriteCause = cause;
+    }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -7,13 +7,26 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Iterables;
+
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
 import lombok.extern.slf4j.Slf4j;
+
 import org.corfudb.protocols.wireprotocol.DataType;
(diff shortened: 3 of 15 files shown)


test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
```diff
@@ -109,6 +111,50 @@ public void customConflictObjectsConflictAborts()
...
+                    .matches(m -> {
+                        if (m.getMsgType().equals(CorfuMsgType.WRITE)) {
+                            rtReader.getStreamsView().get(streamID).next();
+                            return true;
+                        } else {
+                            return false;
+                        }
+                    });
+            addClientRule(rtWriter, testRule);
+            TXEnd();
```
zalokhan: Add a fail after TXEnd. `Assert.fail()` So that the test fails if the transaction completes without aborting.
You can then get rid of your caughtException flag.

test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
```diff
@@ -109,6 +111,50 @@ public void customConflictObjectsConflictAborts()
...
+        Map<String, String> map = rtWriter
+                .getObjectsView().build()
+                .setStreamID(streamID)
+                .setType(SMRMap.class)
+                .open();
+
+        try {
+            OptimisticTXBegin();
+            map.put("k1", "v1");
+            TestRule testRule = new TestRule()
```
zalokhan: Create and add the rule before the put after the map creation probably.

annym: Done

test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
annym: Done

annym:
@zalokhan Addressed all comments.

annym:
@Maithem Yes, it is ready for review.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
```diff
@@ -86,4 +87,22 @@
      * Clears all data and resets all segment handlers.
      */
     void reset();
+
+    default OverwriteCause verifyOverwriteCauseForAddress(long address, LogData entry) {
```
Maithem: Should this really be part of the interface?

Maithem: consider moving this to the logunit class?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
```diff
@@ -86,4 +87,22 @@
...
+        LogData oldEntry = read(address);
```
Maithem: oldEntry -> currentEntry ?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
```diff
@@ -86,4 +87,22 @@
...
+        OverwriteCause cause = OverwriteCause.DIFF_DATA;
+
+        if (oldEntry.isHole()) {
+            cause = OverwriteCause.HOLE;
+        } else if (entry.getData() != null && oldEntry.getData() != null) {
+            if (oldEntry.getData().length == entry.getData().length) {
```
Maithem: What if you are writing a hole in an address that has already been hold filled?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
annym: I thought of making it part of a utils file since I need it both from the StreamLogFiles and the InMemoryStreamLog, but it relies on the read function, so I thought it was maybe better to be part of the base definition and not duplicate code. Or if preferred I can move it to a utils file.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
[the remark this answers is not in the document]
annym: Done

annym: The cause of overwrite exception would be HOLE, regardless of what you are trying to write, which won't go through due to this exception.

runtime/src/main/java/org/corfudb/runtime/exceptions/OverwriteException.java
```diff
@@ -1,7 +1,15 @@
 package org.corfudb.runtime.exceptions;
 
+import lombok.Getter;
+
 /**
  * Created by mwei on 12/14/15.
  */
+@Getter
 public class OverwriteException extends LogUnitException {
+   private OverwriteCause overWriteCause;
```
corfu-repo: MINOR Make this "overWriteCause" field final. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1165)

corfu-repo:
* ![MINOR][MINOR] 1 minor

Maithem:
@annym you can remove "half" of the changes in this PR, by moving the logic (detect if its a hole/same entry/different entry) to the address space. Although that case will be a bit slower (i.e. one extra rpc), I think it is a better trade off.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -148,11 +152,21 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
                 l.getReplicationMode(token.getTokenValue())
                         .getReplicationProtocol(runtime)
                         .write(e, ld);
-            } catch (OverwriteException | WriteSizeException ex) {
-                // If we have an Overwrite exception, it is already too late for trying
-                // to validate the state of the write, we know that the write didn't complete.
-                // Large writes are also rejected right away.
-                throw ex;
+            } catch (OverwriteException ex) {
```
Maithem: If you have a chain longer than the number of retries, would it be possible for the transaction to fail even though the right went through?

PR #1403, opened 2018-07-28
Handle UnavailableServerException on CorfuRuntime

annym:
Fix issue on CorfuRuntime.connect to properly handle
unavailableServerExceptions that are thrown in between the fetchLayout
and the checkVersion.

## Overview

Description:

Fixes a bug on CorfuRuntime.connect. In the case a server is reset right
in the middle of the fetchLayout and checkVersion of the CorfuRuntime
connect we are not catching this server exception and this is thrown up
to the client. Upon this exception we do not retry as the check version
is a complete informational step, no actual implications. If the server
is actually done this will be caught when attempting to perform any
operation against Corfu.

Why should this be merged: Bug Fix

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -3,8 +3,27 @@ package org.corfudb.runtime;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
+
+import java.lang.Thread.UncaughtExceptionHandler;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeoutException;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nonnull;
+
 import lombok.Builder;
 import lombok.Builder.Default;
 import lombok.Data;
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -782,8 +787,13 @@ private void checkVersion() {
                             getVersionString(), version.getVersion());
                 }
             }
-        } catch (TimeoutException | NetworkException | ShutdownException e) {
-            log.error("connect: failed to get version", e);
+        } catch (TimeoutException | NetworkException | ShutdownException | UnavailableServerException e) {
```
Maithem: Why do you catch Exception? aren't the new exceptions coming from adding UnavailableServerException?

annym: Since the checkVersion is purely informational, regardless of whatever exception occurs, if it passes or not, no further action is taken... If we don't catch any generic Exception we could be bubbling up other type of exceptions we are not aware of, and that are irrelevant to this step. In this case the separation was only meant to further provide detailed log info, but at this point we care of not throwing up any kind of exception...

StrikingThirteen: checkVersion is being called only from connect() on Runtime. Even if we are sure that important exceptions as ShutdownException or UnavailabeServerException are again during execution of connect(), would it be better re-throw them here in checkVersion so that connect() deals with them?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -782,8 +787,13 @@ private void checkVersion() {
...
+            log.error("connect: failed to get version. Couldn't connect to server.", e);
+        } catch (Exception ex) {
+            // Because checkVersion is just an informational step (log purpose), we don't need to retry
```
StrikingThirteen: These comments apply to the other catch clause as well. Maybe bring put them before both or in the comment for the method. Also the only difference between the catch clauses is a sentence in the log message (e.g. **Couldn't connect to server.**). If it is not critical combine them to simplify.

PR #1414, opened 2018-08-02
Force Timeout On Connection

Maithem:
## Overview

The tcp connection can get in a state where the client doesn't
receive a termination signal, but the server dies abruptly. In this
case, the client can hang until the OS times out the connection.
This can take a very long time (in our case 15min) until the system
becomes available again.

Why should this be merged: Fixes a bug where the system can hang indefinitely 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
@@ -16,6 +16,7 @@ import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
 import io.netty.handler.codec.LengthFieldPrepender;
 import io.netty.handler.ssl.SslContext;
+import io.netty.handler.timeout.ReadTimeoutHandler;
 import io.netty.util.concurrent.GlobalEventExecutor;
 import lombok.Getter;
 import lombok.NonNull;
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -277,6 +278,7 @@ private ChannelInitializer getChannelInitializer() {
         return new ChannelInitializer() {
             @Override
             protected void initChannel(@Nonnull Channel ch) throws Exception {
+                ch.pipeline().addLast(new ReadTimeoutHandler(parameters.getHandshakeTimeout()));
```
StrikingThirteen: I see that we are using the handshake timeout here. But do we want handshake timeout and connection timeout be fundamentally the same? If yes, then this would fine, however if they may differ in the future we might want to introduce a new a corfu runtime parameter such as connection timeout.

zalokhan:
ClientHandshakeHandler adds this filter too. Will we end up with multiple filters? Should we not add during the handshake then?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
StrikingThirteen: Better naming would be good. however my concern was about a scenario that we want to set-up different timeouts for inactivity and for handshake. It might be a contrived example but we might set an aggressive handshake time out of 10 sec (as it is now in the parameters) but want to allow a lenient timeout of 30 sec for inactivity.

Maithem: the timeout added to the pipeline by the handshake will be removed, if the handshake succeed. @annym this is correct right?

annym: Yes it will be removed. I think Sam is trying to say what if we want different values for the handshake timeout than for the inactiveConnectionTimeout. Also, the default value for the handshakeTimeout is 10 seconds, couldn't we end up closing the connection too frequently?

Maithem: addressed all comments. I think 30 seconds should suffice. Its configurable anyways.

StrikingThirteen: Thanks @Maithem.

Issue #1402, opened 2018-07-28
Server Threads Being Exhausted 


Maithem:
**Bug report.**

Somehow the shared server thread pool can get exhausted which results in blocking any writes to the layout servers thus blocking the healing process. We have to find out why the shared pool is getting exhausted, but we should consider allocating a separate thread pool  for the layout servers.

Maithem:
Merged.

PR #1418, opened 2018-08-03
Streamlog compaction

xnull:
## Overview
Improve Log compaction functional

Description:
Log cleaner extracted from LogUnitServer into a separated class StreamLogCompaction. 

Why should this be merged: 
Added try-catch block to prevent scheduled pool corruption if the task throws an exception. MetricsUtil small improvement.

Related issue(s) (if applicable): #1402 

- [x] Public API has Javadoc

### DIFF
--- a/annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
+++ b/annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
@@ -385,18 +385,14 @@ public class ObjectAnnotationProcessor extends AbstractProcessor {
                             }
                         } else if (method.getAnnotation(InterfaceOverride.class) != null) {
                             interfacesToAdd.add(ParameterizedTypeName.get(ifaceElement.asType()));
-                            if (overwrite.isPresent()) {
-                                methodSet.remove(overwrite.get());
-                            }
-                            methodSet.add(new SmrMethodInfo(method,
-                                        ifaceElement));
+                            overwrite.ifPresent(methodSet::remove);
+                            methodSet.add(new SmrMethodInfo(method, ifaceElement));
                         } else if (
                                 // if we have a implementation, and we aren't already
                                 // implemented by the object.
                                 method.getModifiers().contains(Modifier.DEFAULT)
                                         &&
-                                methodSet.stream().noneMatch(x -> x.method.toString()
-                                .equals(method.toString()))) {
+                                methodSet.stream().noneMatch(x -> x.method.toString().equals(method.toString()))) {
                             methodSet.add(new SmrMethodInfo(method,
                                             ifaceElement, true));
                         }
--- a/checks.xml
+++ b/checks.xml
@@ -86,7 +86,7 @@
             <property name="allowNonPrintableEscapes" value="true"/>
         </module>
         <module name="LineLength">
-            <property name="max" value="100"/>
+            <property name="max" value="120"/>
             <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
         </module>
         <module name="AvoidStarImport"/>
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
@@ -1,0 +1,63 @@
+package org.corfudb.infrastructure.log;
+
+import com.codahale.metrics.Counter;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.util.MetricsUtils;
+
+import java.util.concurrent.*;
+
+/**
+ * Scheduled log compaction manager.
+ * Execute log compaction safely (prevent scheduled executor corruption if the task throws an exception)
+ * https://stackoverflow.com/questions/6894595/scheduledexecutorservice-exception-handling
+ */
+@Slf4j
+public class StreamLogCompaction {
+    private static final String METRIC_NAME = StreamLogCompaction.class.getName();
+
+    private final ThreadFactory threadFactory = new ThreadFactoryBuilder()
+            .setDaemon(true)
+            .setNameFormat("LogUnit-Maintenance-%d")
+            .build();
+
+    /**
+     * How many times log compaction executed
+     */
+    private final Counter gc;
+
+    /**
+     * A scheduler, which is used to schedule periodic garbage collection.
+     */
+    private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor(threadFactory);
+    private final ScheduledFuture<?> compactor;
+
+    public StreamLogCompaction(StreamLog streamLog, long initialDelay, long period, TimeUnit timeUnit) {
+        gc = MetricsUtils.metrics.counter(String.format("%s-stream[%d]", METRIC_NAME, streamLog.hashCode()));
+
+        Runnable task = () -> {
+            gc.inc();
+            log.debug("Start log compaction. Counter: {}", gc.getCount());
+            try {
+                streamLog.compact();
+            } catch (Throwable ex) {
+                log.error("Can't compact stream log. Counter: {}", gc.getCount());
+            }
+        };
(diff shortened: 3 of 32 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
```diff
@@ -0,0 +1,63 @@
...
+    public StreamLogCompaction(StreamLog streamLog, long initialDelay, long period, TimeUnit timeUnit) {
+        gc = MetricsUtils.metrics.counter(String.format("%s-stream[%d]", METRIC_NAME, streamLog.hashCode()));
+
+        Runnable task = () -> {
+            gc.inc();
+            log.debug("Start log compaction. Counter: {}", gc.getCount());
+            try {
+                streamLog.compact();
+            } catch (Throwable ex) {
+                log.error("Can't compact stream log. Counter: {}", gc.getCount());
```
Maithem: Why happens if the thread is interrupted?
Also, if an exception happens here, then we might need to bail.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
```diff
@@ -0,0 +1,63 @@
+package org.corfudb.infrastructure.log;
```
Maithem: Is all of this really necessary ? it seems like your just need to add try/catch in the runnable  task that calls stream.compact

xnull: I believe yes, reasons: 
 - I don't want to pollute LogUnit logic 
 - I also want to measure how many times the compaction happened. I wouldn't like to put that measurement logic into LogUnit. 
 - It's easier to write a test if compaction is in a separate class.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
xnull: Any exception just breaks the task and scheduledPool can be stopped if the shutdown method is called.

I believe we can't do anything in the task if an exception happens, we just write that fact in a log and try to compact the log after a timeout. If you have any ideas what we can do in this case, please let me know.

StrikingThirteen: Expose the exception rather than counter in the catch clause

xnull:
@Maithem ok, then I will split my PR on - `Compaction thread` and `Metrics clean up`

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
```diff
@@ -0,0 +1,63 @@
...
+        Runnable task = () -> {
+            gc.inc();
+            log.debug("Start log compaction. Counter: {}", gc.getCount());
+            try {
+                streamLog.compact();
+            } catch (Throwable ex) {
+                log.error("Can't compact stream log. Counter: {}", gc.getCount());
+            }
+        };
+        compactor = scheduler.scheduleAtFixedRate(task, initialDelay, period, timeUnit);
```
StrikingThirteen: Consider with fixed delay

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
```diff
@@ -0,0 +1,63 @@
...
+        compactor.cancel(true);
+        scheduler.shutdownNow();
+    }
+
+    /**
+     * How many times log gc was executed (the log was compacted)
+     *
+     * @return counter
+     */
+    public long getGcCounter() {
```
StrikingThirteen: counter is already registered to MetricsUtil. wouldn't exposing this here be redundant?

annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
```diff
@@ -385,18 +385,14 @@ public void generateProxy(TypeElement classElement)
                             }
                         } else if (method.getAnnotation(InterfaceOverride.class) != null) {
                             interfacesToAdd.add(ParameterizedTypeName.get(ifaceElement.asType()));
-                            if (overwrite.isPresent()) {
-                                methodSet.remove(overwrite.get());
-                            }
-                            methodSet.add(new SmrMethodInfo(method,
-                                        ifaceElement));
+                            overwrite.ifPresent(methodSet::remove);
```
StrikingThirteen: It doesn't seem to be the previous behavior where we were passing the content of overwrite (if present) to be removed from the set. method references in lambda apply on the value presented to them which is boolean in your statement not the content of overwrite.

checks.xml
```diff
@@ -86,7 +86,7 @@
             <property name="allowNonPrintableEscapes" value="true"/>
         </module>
         <module name="LineLength">
-            <property name="max" value="100"/>
+            <property name="max" value="120"/>
```
StrikingThirteen: I agree that 120 is more preferable to 100 but can we defer this for now and this PR?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.concurrent.*;
+
+/**
+ * Scheduled log compaction manager.
+ * Execute log compaction safely (prevent scheduled executor corruption if the task throws an exception)
+ * https://stackoverflow.com/questions/6894595/scheduledexecutorservice-exception-handling
+ */
+@Slf4j
+public class StreamLogCompaction {
+    private static final String METRIC_NAME = StreamLogCompaction.class.getName();
```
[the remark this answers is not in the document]
xnull: What name would you recommend? I am ready to change to any other name. 

P.S. I gave this name because this const is a local and it doesn't really matter if it generic or nor, so I decided to use a generic name.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
xnull: ok

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
xnull: ok

annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
xnull: Intellij Idea just highlights this code and offers to replace to ifPresent.
Anyway, I won't bring this change in the other PR

checks.xml
xnull: why? is that can affect anything?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogCompaction.java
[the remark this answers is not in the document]
StrikingThirteen: Maybe something with "stream compaction" in name (e.g. STREAM_COMPACT_METRIC)?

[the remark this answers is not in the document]
xnull: Sounds good, I like it, will change

annotationProcessor/src/main/java/org/corfudb/annotations/ObjectAnnotationProcessor.java
StrikingThirteen: Intellij sometimes suggest equivalent syntax changes that might lead to sematic differences. I think Intellij thought that the method (ifPresent) returns the actual value while in reality it returns a boolean (not useful for the method reference). But you can check more, I might be wrong.

checks.xml
StrikingThirteen: its related changes was cluttering this PR. also bring idea to a consensus in the team before making the change as most of our codebase is based on 100. It will lead to great style change across the code base.

xnull: Ok, let's discuss on Monday.

Issue #1419, opened 2018-08-03
Remaining Up To to non-existing address


annym:
## Overview

When attempting to retrieve all entries of a stream up to a non-existing address (negative address) a TrimmedException is thrown. The right behavior would be to actually retrieve 0 entries for this case.

PR #1421, opened 2018-08-03
Remaining up to behavior non-existing addresses

annym:
## Overview

Fix behavior of remaining up to when attempting to retrieve entries up
to a non-existing address (negative address).

Description:

Why should this be merged: wrong behavior. Currently it throws a Trimmed Exception which is fundamentally wrong, because it is not that the address was trimmed, but instead by being an invalid address no entries are retrieved. 

Related issue(s) (if applicable): #1419

- [x ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -135,7 +135,7 @@ public abstract class AbstractQueuedStreamView extends
 
         // If maxGlobal is before the checkpoint position, throw a
         // trimmed exception
-        if (maxGlobal < context.checkpointSuccessStartAddr) {
+        if (maxGlobal < context.checkpointSuccessStartAddr && Address.isAddress(maxGlobal)) {
             throw new TrimmedException();
         }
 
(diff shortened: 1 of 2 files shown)


zalokhan:
Should this be valid for only non-existent addresses or any address less than the current stream pointer?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -135,7 +135,7 @@ protected ILogData getNextEntry(QueuedStreamContext context,
 
         // If maxGlobal is before the checkpoint position, throw a
         // trimmed exception
-        if (maxGlobal < context.checkpointSuccessStartAddr) {
+        if (maxGlobal < context.checkpointSuccessStartAddr && Address.isAddress(maxGlobal)) {
```
StrikingThirteen: Maybe swap the conditions. We should bail right away if it is not a valid address

Maithem: why not just
` if (maxGlobal < context.checkpointSuccessStartAddr || Address.isNonAddress(maxGlobal))` ?

Issue #1423, opened 2018-08-03
UnavailableServerException Leaking


Maithem:
**Bug report.** 
When a client invokes a workflow, a UnavailableServerException can leak and break the retry mechanism.

PR #1424, opened 2018-08-04
Retry on UnavailableServerException

Maithem:
## Overview
Description: When invoking a WorkflowRequest retry on UnavailableServerException.

Why should this be merged: If a workflow requests reaches a server during shutdown, the workflow request will fail and not retry. This fix catch the exception so that the request is retried on a different
orchestrator. 

Related issue(s) (if applicable): #1423

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
@@ -6,6 +6,7 @@ import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.clients.BaseClient;
 import org.corfudb.runtime.clients.ManagementClient;
 import org.corfudb.runtime.exceptions.NetworkException;
+import org.corfudb.runtime.exceptions.UnavailableServerException;
 import org.corfudb.runtime.exceptions.WorkflowException;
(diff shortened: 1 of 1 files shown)


Maithem:
I'm not sure why we even respond to an rpc during shutdown, maybe just ignore the message and the client should get a timeout/socket exception and the client would retry appropriately. What do you guys think? For now this is the most minimal change I can come up with

zalokhan:
Yes, I agree, that would simplify a lot of our code too. There is no need for this special case.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -137,7 +138,7 @@ public void invoke() {
                 ManagementClient orchestrator = getOrchestrator(requestLayout);
                 UUID workflowId = sendRequest(orchestrator);
                 waitForWorkflow(workflowId, orchestrator, timeout, pollPeriod);
-            } catch (NetworkException | TimeoutException e) {
+            } catch (NetworkException | TimeoutException | UnavailableServerException e) {
```
StrikingThirteen: Would it be possible or useful if we have a backing off mechanism so in case of UnavailableServerException we won't exhaust all the retries in a quick succession?

Issue #1429, opened 2018-08-08
Sequencer Overflow Exception


Maithem:
**Bug report.**
Is it possible for transactions to abort with SEQUENCER_OVERFLOW on sequencer roll-over?

PR #1435, opened 2018-08-09
Fix TAE abort cause on sequencer failover

annym:
## Overview

Correct the cause of abort for transactions that cannot commit whenever
there is a sequencer failover from Sequencer Overflow to New Sequencer. 

Why should this be merged: this is a semantic problem that can lead to wrong solutions, like resizing the cache under the believe that transactions are aborting due to sequencer overflow.

Related issue(s) (if applicable): #1429

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -1,7 +1,7 @@
 package org.corfudb.infrastructure;
 
+import static org.corfudb.protocols.wireprotocol.TokenType.TX_ABORT_NEWSEQ;
 import static org.corfudb.protocols.wireprotocol.TokenType.TX_ABORT_SEQ_OVERFLOW;
-
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.RemovalCause;
@@ -121,11 +121,19 @@ public class SequencerServer extends AbstractServer {
      *
      * {@link SequencerServer::maxConflictWildcard} :
      * a "wildcard" representing the maximal update timestamp of
-     * all the confict keys which were evicted from the cache
+     * all the conflict keys which were evicted from the cache
+     *
+     * * {@link SequencerServer::maxConflictNewSequencer} :
+     * represents the max update timestamp of all the conflict keys
+     * which were evicted from the cache by the time this server is elected
+     * the primary sequencer. This means that any snapshot timestamp below this
+     * actual threshold would abort due to NEW_SEQUENCER cause.
      */
+    private final Cache<String, Long> conflictToGlobalTailCache;
+
     private long maxConflictWildcard = Address.NOT_FOUND;
 
-    private final Cache<String, Long> conflictToGlobalTailCache;
+    private long maxConflictNewSequencer = Address.NOT_FOUND;
 
     /**
      * Handler for this server.
@@ -180,12 +188,11 @@ public class SequencerServer extends AbstractServer {
             cacheSize = Long.parseLong((String) opts.get("--sequencer-cache-size"));
 
         }
-
         conflictToGlobalTailCache = Caffeine.newBuilder()
                 .maximumSize(cacheSize)
                 .removalListener((String k, Long v, RemovalCause cause) -> {
                     if (!RemovalCause.REPLACED.equals(cause)) {
-                        log.trace("Updating maxConflictWildcard. Old value = '{}', new value='{}'"
+                         log.trace("Updating maxConflictWildcard. Old value = '{}', new value='{}'"
                                         + " conflictParam = '{}'. Removal cause = '{}'",
                                 maxConflictWildcard, v, k, cause);
                         maxConflictWildcard = Math.max(v, maxConflictWildcard);
@@ -259,6 +266,21 @@ public class SequencerServer extends AbstractServer {
                         break;
                     }
 
+                    // The maxConflictNewSequencer is modified whenever a server is elected
+                    // as the 'new' sequencer, we immediately set its value to the max timestamp
+                    // evicted from the cache at that time. If a txSnapshotTimestamp falls
+                    // under this threshold we can report that the cause of abort is due to
+                    // a NEW_SEQUENCER (not able to hold these in its cache).
+                    if (txSnapshotTimestamp < maxConflictNewSequencer) {
+                        log.debug("ABORT[{}] snapshot-ts[{}] WILDCARD New Sequencer ts=[{}]",
+                                txInfo, txSnapshotTimestamp, maxConflictNewSequencer);
+                        response.set(TX_ABORT_NEWSEQ);
+                        break;
+                    }
+
+                    // If the txSnapshotTimestamp did not fall under the new sequencer threshold
+                    // but it does fall under the latest evicted timestamp we report the cause of
+                    // abort as SEQUENCER_OVERFLOW
                     if (txSnapshotTimestamp < maxConflictWildcard) {
                         log.debug("ABORT[{}] snapshot-ts[{}] WILDCARD ts=[{}]",
                                 txInfo, txSnapshotTimestamp, maxConflictWildcard);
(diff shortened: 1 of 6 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -256,6 +267,13 @@ private TokenType txnCanCommit(TxResolutionInfo txInfo, /** Input. */
                         break;
                     }
 
+                    if (txSnapshotTimestamp < maxConflictNewSequencer) {
```
zalokhan: Can you add a brief comment here explaining when is this `maxConflictNewSequencer` modified and why are we aborting this?

zalokhan:
There is a possibility for false aborts.
Note: This is present in the current code and not introduced by this PR.
Consider a sequencer node_A moving from epoch 0 to 1 retaining its position as the primary sequencer. However, the delta bootstrap (without the stream tails map) message gets dropped. Now the management agent re-runs the fast loader and sends a full bootstrap message (with stream tails map) to the sequencer thus resetting the maxConflictWildcard variables and aborting all transactions.
We can avoid such conditions if the management agent can infer the layout history and bootstrap accordingly. This suggestion is a riskier change and we can look into this in a later patch.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -118,12 +121,21 @@
      *
      * {@link SequencerServer::maxConflictWildcard} :
      * a "wildcard" representing the maximal update timestamp of
-     * all the confict keys which were evicted from the cache
+     * all the conflict keys which were evicted from the cache
      */
     private long maxConflictWildcard = Address.NOT_FOUND;
```
xnull: Looks like this variable should be volatile or AtomicLong

StrikingThirteen: @xnull why? it is only accessed in synchronized methods. @annym will there be unlocked access/write to this?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -118,12 +121,21 @@
...
 
+    /**
+     * Represent the max update timestamp of all the conflict keys
+     * which were evicted from the cache by the time this server was elected
+     * the primary sequencer. This means that any snapshot timestamp below this
+     * actual threshold would abort due to NEW_SEQUENCER cause.
+     */
+    private long maxConflictNewSequencer = Address.NOT_FOUND;
```
Maithem: How are you detecting this, are you just marking the initial token (i.e. during initialization) and throwing new sequencer whenever you see a timestamp less than that? I don't think thats unique enough to identify this issue. Can we use the sequencer epoch?

Maithem: A timestamp is tied to a sequencer epoch, if the epoch != the current sequencer's epoch then reject as a different sequencer. In this case, the epoch also marks the sequencers in memory state.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
StrikingThirteen: I was talking about anny's "maxConflictNewSequencer" being accessed in synchronized method. This one however might and needs investigation.

xnull: Ok, but it seems dangerous anyway

xnull: Just realized - we read maxConflictNewSequencer in txCanCommit method which is not synchronized, so if I am not wrong - it's possible we could update maxConflictNewSequencer in synchronized method but read stale value in  txCanCommit method.
Am I right?

Maithem: It doesn't really matter. The sequencer server is single threaded anyways.

xnull: This fact is specific only for current SequenceServer, what if tomorrow we decide to replace `newSingleThreadExecutor` with another one?

zalokhan:
@xnull All calls to the sequencer are synchronized via `tokenRequest` method. Even `txCanCommit` is called via this method. So there can only be one thread reading or updating at a time.

xnull:
Ok

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -118,12 +121,21 @@
      *
      * {@link SequencerServer::maxConflictWildcard} :
      * a "wildcard" representing the maximal update timestamp of
-     * all the confict keys which were evicted from the cache
+     * all the conflict keys which were evicted from the cache
      */
     private long maxConflictWildcard = Address.NOT_FOUND;
 
+    /**
```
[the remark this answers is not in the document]
annym: Done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
annym: I am detecting this upon a full bootstrap sequencer message (resetServer method) which implies that this is a newly elected Sequencer. As soon as we determine that this is a newly elected sequencer we set this boundary, to the same limit of the last evicted entry in the cache meaning that we are not able to resolve anything below this point but the actual reason is because it is a NEW_SEQUENCER and not actually because OVERFLOW. As soon as you start removing things from the cache the maxConflictWildcard distances away from this boundary and may lead to aborts due to overflow.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -398,7 +424,7 @@ public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg>
         bootstrapEpoch = readyEpoch;
         serverContext.setSequencerEpoch(readyEpoch);
 
-        log.info("Sequencer reset with token = {}, streamTailToGlobalTailMap = {},"
+        log.info("Sequencer reset with token = {}, seamTailToGlobalTailMap = {},"
```
zalokhan: seamTailToGlobalTailMap -> streamTailToGlobalTailMap

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
Maithem: I think that will be a bigger change, not as seemless as just swapping the thread executor.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -118,11 +121,19 @@
...
+     * which were evicted from the cache by the time this server is elected
+     * the primary sequencer. This means that any snapshot timestamp below this
+     * actual threshold would abort due to NEW_SEQUENCER cause.
      */
+    private final Cache<String, Long> conflictToGlobalTailCache;
+
     private long maxConflictWildcard = Address.NOT_FOUND;
 
-    private final Cache<String, Long> conflictToGlobalTailCache;
+    private long maxConflictNewSequencer = Address.NOT_FOUND;
```
Maithem: This name can be improved, maybe make initialToken (from the bootstrap method) global and re-use it to detect timestamps lower than that?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -256,6 +266,21 @@ private TokenType txnCanCommit(TxResolutionInfo txInfo, /** Input. */
                         break;
                     }
 
+                    // The maxConflictNewSequencer is modified whenever a server is elected
```
Maithem: Move this before Line 253?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
annym: Done

Maithem:
Are we still pushing this ?

Maithem:
I think we need this, been seeing a lot of overflow exceptions on some bugs. Can you rebase?

annym:
@Maithem This has been rebased, I have also seen it recently being reported.

Issue #1135, opened 2018-01-04
Invalidate Layout Parameters


Maithem:
**Feature request.**

For some code paths we don't want fetchLayout to spin indefinitely. For instance, if a user specifies incorrect TLS params connect() will spin indefinitely. Maybe we can pass a timeout/retry parameter.

no2chem:
And what should we do if connect fails after X retries? throw an Error?...

no2chem:
Also, this functionality is already implemented in ```registerSystemDownHandler```.

Maithem:
Yea fail with a timeout exception. Let me look into registerSystemDownHandler, I'm not sure if it can be used to specify different behavior for different code paths.

Maithem:
outdated.

PR #1436, opened 2018-08-09
Rethrow systemDownHandler exceptions

### LINKED ISSUE #1433: SystemDownHandler swallowed in AbstractView thrown by fetchLayout
## Overview
While executing the `layoutHelper` in `AbstractView` a systemDownHandler is thrown by the `fetchLayout` method while getting the layout. This Exception is wrapped around by the ExecutionException. This leads to a deadlock in the `AbstractView`.

zalokhan:
## Overview

Description: Rethrow SystemDownHandler Exceptions in AbstractView thrown by fetchLayout.

Why should this be merged: AbstractView gets stuck in an infinite loop if `layout.get()` is unable to fetch a layout.

Related issue(s) (if applicable): Fixes #1433 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -6,12 +6,12 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import javax.annotation.Nonnull;
+
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.NetworkException;
 import org.corfudb.runtime.exceptions.ServerNotReadyException;
 import org.corfudb.runtime.exceptions.WrongEpochException;
-import org.corfudb.runtime.exceptions.unrecoverable.SystemUnavailableError;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.util.Sleep;
 
@@ -121,29 +121,33 @@ public abstract class AbstractView {
                     throw new RuntimeException(re);
                 }
 
-                // Invoking the systemDownHandler if the client cannot connect to the server.
-                if (--systemDownTriggerCounter <= 0) {
-                    runtime.systemDownHandler.run();
-                }
-                runtime.invalidateLayout();
-                Sleep.sleepUninterruptibly(retryRate);
-
             } catch (InterruptedException ie) {
                 throw new UnrecoverableCorfuInterruptedError("Interrupted in layoutHelper", ie);
             } catch (ExecutionException ex) {
-                log.warn("Error executing remote call, invalidating view and retrying in {} ms",
-                        retryRate, ex);
-
-                // If SystemUnavailable exception is thrown by the layout.get() completable future,
-                // the exception will materialize as an ExecutionException. In that case, we need to propagate
-                // this exception.
-                if (ex.getCause() instanceof SystemUnavailableError) {
-                    throw (SystemUnavailableError) ex.getCause();
+                // If an error or an unchecked exception is thrown by the layout.get() completable
+                // future, the exception will materialize as an ExecutionException. In that case,
+                // we need to propagate this Error or unchecked exception.
+                if (ex.getCause() instanceof Error) {
+                    log.error("layoutHelper: Encountered error. Aborting layoutHelper", ex);
+                    throw (Error) ex.getCause();
                 }
 
-                runtime.invalidateLayout();
-                Sleep.sleepUninterruptibly(retryRate);
+                if (ex.getCause() instanceof RuntimeException) {
+                    log.error("layoutHelper: Encountered unchecked exception. "
+                            + "Aborting layoutHelper", ex);
+                    throw (RuntimeException) ex.getCause();
+                }
+
+                log.warn("layoutHelper: Error executing remote call, invalidating view and "
+                        + "retrying in {} ms", retryRate, ex);
             }
+
+            // Invoking the systemDownHandler if the client cannot connect to the server.
+            if (--systemDownTriggerCounter <= 0) {
+                runtime.systemDownHandler.run();
+            }
+            runtime.invalidateLayout();
+            Sleep.sleepUninterruptibly(retryRate);
         }
     }
 


runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -131,14 +131,21 @@ public Layout getCurrentLayout() {
...
             } catch (ExecutionException ex) {
-                log.warn("Error executing remote call, invalidating view and retrying in {} ms",
-                        retryRate, ex);
-
-                // If SystemUnavailable exception is thrown by the layout.get() completable future,
-                // the exception will materialize as an ExecutionException. In that case, we need to propagate
-                // this exception.
-                if (ex.getCause() instanceof SystemUnavailableError) {
-                    throw (SystemUnavailableError) ex.getCause();
+                log.warn("layoutHelper: Error executing remote call, invalidating view and "
```
StrikingThirteen: Maybe move this log to after the two if clauses. (as if we encounter Error or RuntimeException, there won't be any invalidating and retrying)

zalokhan: Done.

StrikingThirteen: Thanks

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -131,16 +131,23 @@ public Layout getCurrentLayout() {
...
             } catch (ExecutionException ex) {
-                log.warn("Error executing remote call, invalidating view and retrying in {} ms",
-                        retryRate, ex);
-
-                // If SystemUnavailable exception is thrown by the layout.get() completable future,
-                // the exception will materialize as an ExecutionException. In that case, we need to propagate
-                // this exception.
-                if (ex.getCause() instanceof SystemUnavailableError) {
-                    throw (SystemUnavailableError) ex.getCause();
+                // If an error or an unchecked exception is thrown by the layout.get() completable
```
Maithem: Just timeout on the future, this is not necessary.

zalokhan:
Timing out might not be trivial. Since the `fetchLayout` traverses through all the layout servers, it can take a variable amount of time until it returns back the layout. We cannot keep the timeout a function of the number of layout servers too as we don t have the layout yet.

Maithem:
Why not? 

Why even differentiate between long wait periods on layout discovery vs request completion? The period to detect a fault (i.e. detect an unresponsive system) should be independent for both cases. Depending on the fetchlayout to invoke the system down handler can inflate the abstracts view time before detecting the system is unresponsive.

zalokhan:
So there is a difference between waiting on `fetchLayout` for a specific time vs waiting for it to attempt requesting the layout from all Layout servers. Currently, we bound on the retry on all layout servers.
Assume we have 10 layout servers. Due to the fetchLayout choosing layout servers at random, there can be a chance that we contact 3 unresponsive servers and timeout multiple times causing our system down handler to be invoked. This would be incorrect as we are declaring the cluster to be unresponsive due to our short timeout.

xnull:
Yep, I agree, I mean entire `layoutHelper` method, it's overcomplicated

Maithem:
I understand that layout discovery can take time, if it takes more time than we expect, then we should invoke the handler.

> Due to the fetchLayout choosing layout servers at random, there can be a chance that we contact 3 unresponsive servers and timeout multiple times causing our system down handler to be invoked.

fetchLayout doesn't report progress, its binary. Either it blocks, or invokes the system down handler. Having another thread waiting on that binary decision doesn't seem to be necessary (can't tell progress). From the application's point of view, a request hasn't completed for a time duration and thats what constitutes an "unavailable system".

zalokhan:
If the API waits for all the responses it would have done its due diligence and then reported the systemDownHandler. 
The timeout can even be added by the user by wrapping its RPC calls in a timebound future. But what if the customer does want to wait/block until the server is up?

StrikingThirteen:
@zalokhan @Maithem Currently fetchLayout is going through the shuffled layoutServers consequentially. and in the case of encountering unresponsive servers it delays (or will lead to timeout in case of bounding the entire op as suggested above). Would it be possible or desirable to have fetchLayout spawn several thread and attempt to fetch the layout from several layout servers in parallel? In that case, it can improve performance and in the case of adding timeout as mentioned in the comments and avoid unnecessarily invoking system down handler.

Maithem:
>  Plus changing the behaviour of the systemDownHandler from being retry bound to timebound would be an interface change.

I'm not suggesting to change the behavior of the systemDownHandler. Its a lambda anyways, how woud that change the interface?

> The timeout can even be added by the user by wrapping its RPC calls in a timebound future.

Consider a map.put, are you suggesting wrapping that with a future and applying a timeout to that? Doesn't seem necessary, it would also make app code superfluous

> But what if the customer does want to wait/block until the server is up?

Then they shouldn't register a system down handler. Also, this patch has the oppiste behavior (since it will break from the abstract views loop)

Maithem:
@StrikingThirteen How would this optimization remove the need to not have the system down handler. Even if all the parallel requests timeout, you might want to retry.

StrikingThirteen:
@Maithem Where did I say that this optimization removes the need of system down handler / retries? The point was that we might avoid unnecessary invocations, that is when by chance the first layout servers are being queried in the loop are unresponsive. Attempting in parallel increases the chance of getting a layout from responsive layout servers sooner.

Maithem:
@StrikingThirteen oh, I see your point now, but getting a response faster is not what this PR is trying to solve.

StrikingThirteen:
@Maithem agreed it is outside the scope of this PR. That being said, in the future it might alleviate @zalokhan 's concerns regarding a sequence of unresponsive layout servers being queried first. So when implementing the time bounded fetch you suggested, we can have higher likelihood of getting a response faster which lowers likelihood of getting the TimeoutException (give there are responsive layout servers in the collection).

apavan:
I like the idea of timeout but to me I think we should have a wholistic approach as I see potential for more enhancements in this space. We should open a new issue for this and track it under that. For what this change is attempting fix, this seems to be the minimalistic change.

apavan:
LGTM

zalokhan:
Also, refer #1135 thread where we had this discussion earlier. But I guess we did not conclude that either.

Maithem:
> I like the idea of timeout but to me I think we should have a wholistic approach as I see potential for more enhancements in this space.

What does this mean, which space/enhancements are you referring to?

> For what this change is attempting fix, this seems to be the minimalistic change.

Not true. The timeout is a single line change and cleaner (i.e. eliminates blindly waiting on a different thread and provides a tighter timeout bound) from a code organization point of view.

PR #1440, opened 2018-08-09
Terminating sequencerBootstrap thread on shutdown

### LINKED ISSUE #1439: SequencerBootstrapTask cannot be shutdown.
## Overview

The asynchronous task executing the sequencer recovery by running the FastObjectLoader cannot be interrupted since its submitted to the ForkJoinPool. This task self-aborts using the systemDownHandler after 60 seconds.
However, several tests are timing out due to this behaviour causing CI to fail intermittently.
We can use a `cachedThreadPool` which gives us a better control over this thread.

zalokhan:
## Overview

Description: Submitting sequencer bootstrap tasks to the detectionWorker thread pool instead of ForkJoinPool to better control the termination of this task.

Why should this be merged: On shut down the task runs asynchronously. 

Related issue(s) (if applicable): Fixes #1439 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
@@ -18,6 +18,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
@@ -84,7 +85,7 @@ public class ManagementAgent {
      * To dispatch initialization tasks for recovery and sequencer bootstrap.
      */
     @Getter
-    private Thread initializationTaskThread;
+    private final Thread initializationTaskThread;
     /**
      * Detection Task Scheduler Service
      * This service schedules the following tasks every policyExecuteInterval (1 sec):
@@ -97,7 +98,7 @@ public class ManagementAgent {
      * To dispatch tasks for failure or healed nodes detection.
      */
     @Getter
-    private ExecutorService detectionTaskWorkers;
+    private final ExecutorService detectionTaskWorkers;
     /**
      * Future for periodic failure and healed nodes detection task.
      */
@@ -113,8 +114,10 @@ public class ManagementAgent {
     /**
      * Future which is reset every time a new task to bootstrap the sequencer is launched by the
      * ManagementAgent. This is to avoid multiple bootstrap requests.
+     * Using AtomicReference here to avoid multiple sequencer recovery tasks being triggered.
      */
-    private volatile CompletableFuture<Boolean> sequencerRecoveryFuture;
+    private final AtomicReference<Future<Boolean>> sequencerRecoveryFuture
+            = new AtomicReference<>(CompletableFuture.completedFuture(true));
 
     /**
      * The management agent attempts to bootstrap a NOT_READY sequencer if the
@@ -198,7 +201,7 @@ public class ManagementAgent {
         this.reconfigurationEventHandler = new ReconfigurationEventHandler();
 
         final int managementServiceCount = 1;
-        final int detectionWorkersCount = 2;
+        final int detectionWorkersCount = 3;
 
         this.detectionTasksScheduler = Executors.newScheduledThreadPool(
                 managementServiceCount,
@@ -223,8 +226,6 @@ public class ManagementAgent {
                         .setNameFormat(serverContext.getThreadPrefix() + "LocalMetricsPolling")
                         .build());
 
-        this.sequencerRecoveryFuture = CompletableFuture.completedFuture(true);
-
         // Creating the initialization task thread.
         // This thread pool is utilized to dispatch one time recovery and sequencer bootstrap tasks.
         // One these tasks finish successfully, they initiate the detection tasks.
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -223,6 +228,15 @@
                         .setNameFormat(serverContext.getThreadPrefix() + "LocalMetricsPolling")
                         .build());
 
+        // Creating a cached thread pool to spawn a task to bootstrap the sequencer. The number of
+        // tasks submitted to this thread pool is limited to one. We check if we have an ongoing
+        // task before we launch a new one.
+        this.sequencerBootstrapWorker = Executors.newCachedThreadPool(
```
Maithem: This class has too many thread pools, why is this needed?  Why not reuse the detectionTaskWorkers pool? Also, why is this a cched thread pool, if all you want to do is cancel a task, you can use a single threaded executor.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -242,12 +256,12 @@
...
      *
      * @param layout Layout to use to bootstrap the primary sequencer.
-     * @return Completable future which completes when the task completes successfully or with a
-     * failure.
+     * @return Future which completes when the task completes successfully or with a failure.
      */
-    public CompletableFuture<Boolean> triggerSequencerBootstrap(@NonNull Layout layout) {
+    public Future<Boolean> triggerSequencerBootstrap(@NonNull Layout layout) {
         if (sequencerRecoveryFuture.isDone()) {
-            this.sequencerRecoveryFuture = CompletableFuture.supplyAsync(() -> {
```
Maithem: I see, this spawns non-daemon threads.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
zalokhan: I can reuse the detectionWorkers pool by supplying it an extra thread.
I used the cached thread pool to optimize for thread resources and return the thread once the task is completed.
In a good scenario where we have minimal epoch changes, this thread will be wasted.

Maithem: Optimize as needed. 

Are you saying that newCachedThreadPool will destroy the thread if its not being used?

Maithem: I see. I don't think this optimization is necessary.

zalokhan: Ok, I'll add a thread to detectionWorkersPool then.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -242,12 +256,12 @@
      * a task in progress, this is a no-op.
      *
      * @param layout Layout to use to bootstrap the primary sequencer.
-     * @return Completable future which completes when the task completes successfully or with a
-     * failure.
+     * @return Future which completes when the task completes successfully or with a failure.
      */
-    public CompletableFuture<Boolean> triggerSequencerBootstrap(@NonNull Layout layout) {
+    public Future<Boolean> triggerSequencerBootstrap(@NonNull Layout layout) {
```
StrikingThirteen: I know that this method is not being used outside ManagementAgent, yet this is a breaking change of interface. Can we avoid it? Also if it is not needed to be public, now that the interface is changing maybe consider narrowing the visibility of triggerSequenceBootstrap method

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
StrikingThirteen: @Maithem If we are going to increase the detection thread pool size why not just using SingleThreadExecutor? @zalokhan Although cached thread pool shed their non-core threads, the Executor's default has unlimited max. to take the same benefit maybe consider:
`this.sequencerBootstrapWorker = new ThreadPoolExecutor(0, 1,  
                                                      60, TimeUnit.SECONDS, 
                                                       aQueue, aThreadFactory); `

It has an upper limit of 1 which is semantically inline with logic of the bootstrapping (no more than 1 bootstrapping task) and it has 0 core thread.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -792,6 +807,7 @@ public void shutdown() {
         detectionTasksScheduler.shutdownNow();
         detectionTaskWorkers.shutdownNow();
         localMetricsPollingService.shutdownNow();
+        sequencerBootstrapWorker.shutdownNow();
```
StrikingThirteen: Not sure, if you are going to keep all four. but in that case, maybe a method that takes a collection of thread pools and shut them might be more readable. (same for awaitTermination)

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -256,8 +256,9 @@
                 }
                 return true;
             });
+        } else {
+            log.info("triggerSequencerBootstrap: a bootstrap task is already in progress.");
```
medhavidhawan: this is predicated on the fact that the function is being called in a single threaded manner. Please improve this.

zalokhan:
@medhavidhawan Fixed the sequencer trigger on a node to be atomic.

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
```diff
@@ -1,102 +1,78 @@
...
-    /** Generate a new {@link SingletonResource}.
+    /**
+     * Generate a new {@link SingletonResource}.
      *
-     * @param generator     A method to be called when a new {@link T} is needed.
+     * @param generator A method to be called when a new {@link T} is needed.
      */
     private SingletonResource(Supplier<T> generator) {
-        lock = new StampedLock();
+        this.resource = new AtomicReference<T>();
```
corfu-repo: MINOR Replace the type specification in this constructor call with the diamond operator ("<>"). [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2293)

corfu-repo:
* ![MINOR][MINOR] 1 minor

annym:
We might need to update the overview description of this PR (commit message) so it actually matches with the final implementation and avoid discrepancies or quick assumptions in the future.

PR #1441, opened 2018-08-09
Reject empty seq bootstrap msg on failover.

zalokhan:
## Overview

Description: Reject empty/delta sequencer bootstrap messages on the failover sequencer.

Why should this be merged: Causes a correctness issue where the sequencer dispatches stale tokens and regresses the token count.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -7,7 +7,9 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+
 import io.netty.channel.ChannelHandlerContext;
+
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -21,6 +23,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
@@ -134,14 +137,14 @@ public class SequencerServer extends AbstractServer {
 
     @Getter
     @Setter
-    private volatile long bootstrapEpoch = Layout.INVALID_EPOCH;
+    private volatile long sequencerEpoch = Layout.INVALID_EPOCH;
 
     @Override
     public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
-        if ((bootstrapEpoch != serverContext.getServerEpoch())
+        if ((sequencerEpoch != serverContext.getServerEpoch())
                 && (!msg.getMsgType().equals(CorfuMsgType.BOOTSTRAP_SEQUENCER))) {
             log.warn("Rejecting msg at sequencer : sequencerStateEpoch:{}, serverEpoch:{}, "
-                    + "msg:{}", bootstrapEpoch, serverContext.getServerEpoch(), msg);
+                    + "msg:{}", sequencerEpoch, serverContext.getServerEpoch(), msg);
             return false;
         }
         return true;
@@ -300,16 +303,16 @@ public class SequencerServer extends AbstractServer {
         Token token;
         if (req.getStreams().isEmpty()) {
             // Global tail query
-            token = new Token(globalLogTail.get() - 1, bootstrapEpoch);
+            token = new Token(globalLogTail.get() - 1, sequencerEpoch);
             streamTails = Collections.emptyList();
         } else if (req.getStreams().size() == 1) {
             // single stream query
-            token = new Token(streamTailToGlobalTailMap.getOrDefault(streams.get(0), Address.NON_EXIST), bootstrapEpoch);
+            token = new Token(streamTailToGlobalTailMap.getOrDefault(streams.get(0), Address.NON_EXIST), sequencerEpoch);
             streamTails = Collections.emptyList();
         } else {
             // multiple stream query, the token is populated with the global tail and the tail queries are stored in
             // streamTails
-            token = new Token(globalLogTail.get() - 1, bootstrapEpoch);
+            token = new Token(globalLogTail.get() - 1, sequencerEpoch);
             streamTails = new ArrayList<>(streams.size());
             for (int x = 0; x < streams.size(); x++) {
                 streamTails.add(streamTailToGlobalTailMap.getOrDefault(streams.get(x), Address.NON_EXIST));
@@ -351,7 +354,7 @@ public class SequencerServer extends AbstractServer {
                                          ChannelHandlerContext ctx, IServerRouter r) {
         long initialToken = msg.getPayload().getGlobalTail();
         final Map<UUID, Long> streamTails = msg.getPayload().getStreamTails();
-        final long readyEpoch = msg.getPayload().getReadyStateEpoch();
+        final long bootstrapMsgEpoch = msg.getPayload().getSequencerEpoch();
 
         // Boolean flag to denote whether this bootstrap message is just updating an existing
         // primary sequencer with the new epoch (if set to true) or bootstrapping a currently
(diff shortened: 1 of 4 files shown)


test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1664,4 +1658,97 @@ public void unblockSequencerRecoveryOnDeadlock() throws Exception {
...
     }
+
+    /**
+     * Scenario to verify that we do not regress the token count when the layout switches primary
+     * sequencers.
+     * We assert that the failover sequencer should always receive a full bootstrap message rather
+     * than an empty bootstrap message (without streamTailsMap)
+     */
+    @Test
+    public void failoverSeqDoesNotRegressTokenValue() throws Exception {
```
StrikingThirteen: Thanks for the helpful testcase/scenario.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1664,4 +1658,97 @@ public void unblockSequencerRecoveryOnDeadlock() throws Exception {
...
+        // Append data 10 - 19
+        for (int i = 0; i < num; i++) {
+            streamView.append(payload);
+        }
+
+        // Remove partition and allow PORT_0 to be healed.
+        clearServerRules(SERVERS.PORT_0);
+
+        // Add rules so that Full Sequencer bootstrap messages are not sent to PORT_0.
+        TestRule dropFullBootstrapRule = new TestRule()
```
annym: To be on the safe side do you think we should add these rules before even clearing the server rules on PORT_0? I know there shouldn't be a change in the sequencer but if it were , we are sure it will not be taking a full bootstrap message.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1664,4 +1658,97 @@ public void unblockSequencerRecoveryOnDeadlock() throws Exception {
...
+        final byte[] payload = "test".getBytes();
+        final int num = 10;
+        // 0 - 9
+        for (int i = 0; i < num; i++) {
+            streamView.append(payload);
+        }
+
+        // Adding rule to partition out PORT_0.
+        addServerRule(SERVERS.PORT_0, new TestRule().always().drop());
+
```
annym: should we wait for PORT_0 to show as unresponsive?

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1664,4 +1658,97 @@ public void unblockSequencerRecoveryOnDeadlock() throws Exception {
...
+        // Wait until PORT_1 is responsive and healed in the layout.
+        for (int i = 0; i < PARAMETERS.NUM_ITERATIONS_MODERATE; i++) {
+            if (corfuRuntime.getLayoutView().getLayout().getUnresponsiveServers().isEmpty()) {
+                break;
+            }
+            Sleep.sleepUninterruptibly(PARAMETERS.TIMEOUT_SHORT);
+            corfuRuntime.invalidateLayout();
+        }
+        assertThat(corfuRuntime.getLayoutView().getLayout().getUnresponsiveServers().isEmpty())
```
annym: Right after PORT_1 comes back and to verify functionality, is there a way we could proof that the sequencer received a delta bootstrap and it rejected it? before allowing full sync messages back?

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: It doesn't matter where we add it as long as we add it before we partition PORT_1 and mark it unresponsive. Once PORT_1 is partitioned out then PORT_0 should not receive full bootstrap messages for the test case to work. Does this explanation help?

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: The next statement is the append. The append takes care of this and will only proceed if PORT_0 is removed from the log unit chain. So we achieve this indirectly.

annym: ok, perfect!

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: I can intercept the delta bootstrap message but cannot verify that the sequencer rejects it. However, this would change in the future if we move the state creation to the sequencer.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
annym: yes that's why I said there shouldn't be a sequencer change even when PORT_0 rules are cleared, but was wondering if coming back could in any way trigger a change of sequencer even when PORT_! is still up. If there is no possible way for this to happen, I agree the order does not matter.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
annym: If you don't consider this check adds any value then it's ok, we can leave it as it is...

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: Right. I'm moving this to the top just to be safe.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: I have added the message interception and check that we receive a delta bootstrap message and is responded with a NACK. Let me know if this looks good and if you have any other suggestions.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -359,10 +362,13 @@ public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg>
         final boolean bootstrapWithoutTailsUpdate = msg.getPayload()
                 .getBootstrapWithoutTailsUpdate();
 
-        // If bootstrapEpoch is -1 (startup), the sequencer should not accept
-        // bootstrapWithoutTailsUpdate bootstrap messages.
-        if (bootstrapEpoch == Layout.INVALID_EPOCH && bootstrapWithoutTailsUpdate) {
-            log.warn("Cannot update existing sequencer. Require full bootstrap.");
+        // If bootstrapEpoch is -1 (startup) OR readyEpoch is not the consecutive epoch of the
```
Maithem: I think we missed this because this method's logic is convoluted and the language(description of states) is unclear. Since we are here, let's clean this up. The main points: 

- I think sequencerEpoch is more appropriate than readyEpoch
- bootstrapWithoutTailsUpdate should probably be called epochUpdate
- explicitly encode the states (unbootstrapped | streamTailsUpdate )
- reduce the branching complexity (simplify multiple independent if-clauses)

zalokhan: bootstrapWithoutTailsUpdate is a boolean denoting whether it is a full or a delta bootstrap message. epochUpdate doesn't seem the right name to replace this.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -359,10 +362,13 @@ public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg>
...
+        // bootstrapEpoch then the sequencer should not accept bootstrapWithoutTailsUpdate
+        // bootstrap messages.
+        if (bootstrapWithoutTailsUpdate
+                && (bootstrapEpoch == Layout.INVALID_EPOCH || bootstrapEpoch != readyEpoch - 1)) {
```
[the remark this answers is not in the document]
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
zalokhan: I need the multiple if-statements for appropriate logging. Combining multiple statements will cause difficulty in debugging in cases of bootstrap message rejection. I haven't added any more than the current code.

zalokhan: I have made the change to sequencerEpoch.

zalokhan: `bootstrapWithoutTailsUpdate should probably be called epochUpdate` Can you please explain?

Maithem: Is this not just a message that just updates the sequencer's epoch? how is delta bootstrap and bootstrapWithoutTailsUpdate more intuitive than epochUpdate. I get what you are trying to say, but I'm pointing out that the language is really overloaded  :)

zalokhan: I see. But all bootstrap messages update the epoch - with full bootstrap messages and the delta ones too. Wont it cause confusion?  How about resetSequencerState?

zalokhan:
@Maithem CI and pre-checkin have passed on this.
Kindly let me know if you need any more changes else I can merge this.

PR #1445, opened 2018-08-14
Fix singleton resource locking.

### LINKED ISSUE #1446: Singleton resource deadlock
## Overview

Singleton Resource used by the Servers does not release the lock correctly in case of upgrading the lock from a read to write level.
The threads attempt to unlock with an incorrect timestamp causing it to crash with an exception and hence never releasing the lock.

https://github.com/CorfuDB/CorfuDB/blob/2678f7c858d9105666b176cdc29904de3b71d08c/runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java#L53-L82

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/StampedLock.html#tryConvertToWriteLock-long-

zalokhan:
## Overview

Description: SingletonResource does not release the correct lock in case of lock upgrade from readLock to writeLock.

Why should this be merged: Lock is never released and thread crashes.

Related issue(s) (if applicable): Fixes #1446 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
+++ b/runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
@@ -1,102 +1,78 @@
 package org.corfudb.util.concurrent;
 
-import java.util.concurrent.locks.StampedLock;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+
 import javax.annotation.Nonnull;
 
-/** Utility class which implements a singleton resource pattern.
+/**
+ * Utility class which implements a singleton resource pattern.
  *
- *  <p>A {@link SingletonResource} is a common resource pattern where the first thread that
- *  needs to use a resource instantiates it. Subsequent threads should re-use the resource
- *  instantiated by the first thread. The {@link SingletonResource#cleanup(Consumer)} allows
- *  the developer to release resources if the resource has been instantiated (doing nothing
- *  if no thread ever created the resource in the first place).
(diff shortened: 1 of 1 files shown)


zalokhan:
@Maithem Stamped lock has been used so that we can use the read write lock levels.

@StrikingThirteen Yes, I will make the change.

medhavidhawan:
I have similar concerns as Maithem. this is unnecessarily complicated. Can we simplify this ? This is for singleton pattern so there are no efficiency concerns here. Why such a complicated code structure to achieve this ?

StrikingThirteen:
@zalokhan I have another concern about the thread safety of this class although not related to changes on this commit. It seems that SingletonResource might be unsafely publishing/modifying a resource. Consider this scenario: a thread can acquire a readLock (or write for that matters) and gets a resource in a stable state. So far so good, however there is a chance that after publishing the instance, another thread comes along to cleanup, gets a writeLock, and starts the clean up on the already published instance (L99). so the previous thread has an instance in an unstable state (under cleaning op). Maybe @medhavidhawan 's suggestion toward simplification can address this too.

zalokhan:
@Maithem @medhavidhawan I have simplified the code to use an AtomicReference. This greatly simplifies the code. Kindly review.

zalokhan:
@StrikingThirteen I have changed the code. Kindly review again.
To answer your previous question, that scenario is possible. But in our implementation, we shut down all Executor services which make use of the CorfuRuntime before calling the cleanup method. However, if any thread holds a reference to a stale shut down runtime, it will encounter a shutdown exception and the thread should exit.

Issue #1442, opened 2018-08-10
Learn trim mark


Maithem:
**Feature request.**
Periodically try to learn the trim mark by querying the system. This will help flush CorfuRuntime caches.

zalokhan:
Merged #1457

zalokhan:
Merged in 0.2.1. Will keep this open until it is ported to master too.

Maithem:
Doesn't address stream layer objects.

PR #1448, opened 2018-08-14
Learn trim mark (#1442)

cjqhenry14:
## Overview

Description:
1. Periodically try to learn the trim mark by querying the system.
This will help flush CorfuRuntime caches.

2. Clear cache when runtime is shutdown.

Why should this be merged: 

Related issue(s) (if applicable): #1442 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -125,6 +125,9 @@ public class CorfuRuntime {
 
         /** Sets expireAfterAccess and expireAfterWrite in seconds. */
         @Default long cacheExpiryTime = Long.MAX_VALUE;
+
+        /** Sets the period of retrieving the latest trim mark in minutes. */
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -125,6 +125,9 @@
 
         /** Sets expireAfterAccess and expireAfterWrite in seconds. */
         @Default long cacheExpiryTime = Long.MAX_VALUE;
+
+        /** Sets the period of retrieving the latest trim mark in minutes. */
+        @Default long trimMarkSyncPeriod = 10;
```
xnull: why not private?

cjqhenry14: Now I follow the existing access levels for these parameters. I agree that it should be private as the public getter/setter will be generated.

zalokhan:
Merged in #1457

PR #1476, opened 2018-09-11
Move systemDownHandler to CorfuRuntimeParameters.

zalokhan:
## Overview

Description:
Moving the systemDownHandler to CorfuParameters. Making it volatile and adding additional logging.

Why should this be merged: Adding logging for debuggability.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -122,9 +122,9 @@ public abstract class AbstractView {
                                                                function,
                                                        boolean rethrowAllExceptions)
             throws A, B, C, D {
-        runtime.beforeRpcHandler.run();
+        runtime.getParameters().getBeforeRpcHandler().run();
         final Duration retryRate = runtime.getParameters().getConnectionRetryRate();
-        int systemDownTriggerCounter = runtime.getParameters().getSystemDownHandlerTriggerLimit();
+        int systemDownTriggerCounter = 0;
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -159,8 +159,12 @@ private Layout getLayoutUninterruptibly() {
 
             // Invoking the systemDownHandler if the client cannot connect to the server.
             if (--systemDownTriggerCounter <= 0) {
-                runtime.systemDownHandler.run();
+                log.info("layoutHelper: Invoking the systemDownHandler.");
+                runtime.getParameters().getSystemDownHandler().run();
             }
+            // Minimum counter value should be 0.
+            systemDownTriggerCounter = Math.max(0, systemDownTriggerCounter);
```
medhavidhawan: if we are going to do it this way, then why not do this way
if(Math.max(0, --systemDownTriggerCounter) == 0) {

}

zalokhan: We will not update the counter with the value then. This causes us to log negative numbers in the log statements.

medhavidhawan: yes or you can do a better job by updating the number of attempts you have made so far.

zalokhan: Done.

Issue #1493, opened 2018-10-08
Incorrect SingletonResource Implementation 


Maithem:
**Bug report.**
The singleton implementation can create multiple instances of the resource. The implementation is being used to create CorfuRuntimes that can't be GC's and thus leaking resources (i.e. netty channels).

PR #1494, opened 2018-10-08
Proper Singleton Implementation

Maithem:
## Overview
The current implementation was incorrectly using an atomic primitive
that was leading to creating multiple resources per singleton. In
this particular case, it was leaking netty channels through
CorfuRuntime instances. The new implementation uses double
synchronization check to implement acquisition.

Why should this be merged: Fixes a resource leak 

Related issue(s) (if applicable): #1493

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
+++ b/runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
@@ -1,6 +1,5 @@
 package org.corfudb.util.concurrent;
 
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
@@ -11,9 +10,7 @@ import javax.annotation.Nonnull;
  *
  * <p>A {@link SingletonResource} is a common resource pattern where the first thread that
  * needs to use a resource instantiates it. Subsequent threads should re-use the resource
- * instantiated by the first thread. The {@link SingletonResource#cleanup(Consumer)} allows
- * the developer to release resources if the resource has been instantiated (doing nothing
- * if no thread ever created the resource in the first place).
+ * instantiated by the first thread.
  *
  * @param <T> The type of resource this {@link SingletonResource} holds.
  */
@@ -27,7 +24,7 @@ public class SingletonResource<T> {
     /**
      * The resource to be held.
      */
-    private final AtomicReference<T> resource;
+    private volatile T resource;
 
     /**
      * Factory method with similar semantics as a {@link ThreadLocal}.
@@ -46,7 +43,6 @@ public class SingletonResource<T> {
      * @param generator A method to be called when a new {@link T} is needed.
      */
     private SingletonResource(Supplier<T> generator) {
-        this.resource = new AtomicReference<T>();
         this.generator = generator;
 
     }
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
```diff
@@ -264,6 +264,7 @@ public void shutdown() {
             log.debug("Orchestrator executor awaitTermination interrupted : {}", ie);
             Thread.currentThread().interrupt();
         }
+        getRuntime.cleanup(CorfuRuntime::shutdown);
```
zalokhan: The management server already shuts this down. It shuts down the services using this first i.e., the orchestrator and the management service and then finally the runtime.
Shutting this down here will only create a new instance by the management service if that hasn't been shut down yet.

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
```diff
@@ -66,13 +75,10 @@ public T get() {
...
-    public void cleanup(@Nonnull Consumer<T> cleaner) {
-        resource.updateAndGet(t -> {
-            if (t != null) {
-                cleaner.accept(t);
-            }
-            return null;
-        });
+    public synchronized void cleanup(@Nonnull Consumer<T> cleaner) {
+        if (resource != null) {
+            cleaner.accept(resource);
```
zalokhan: Set it back to null if we want the object to be reusable.

corfu-repo:

1. ![MINOR][MINOR] [SingletonResource.java#L3](https://github.com/CorfuDB/CorfuDB/blob/9ae9ef1d278d77585867bbfa1d56da61bb2658f1/runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java#L3): Remove this unused import 'java.util.concurrent.atomic.AtomicReference'. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AUselessImportCheck)

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
```diff
@@ -57,7 +56,17 @@ private SingletonResource(Supplier<T> generator) {
...
+        if (temp == null) {
+            synchronized (this) {
+                temp = resource;
+                if (temp == null) {
+                    temp = generator.get();
+                    resource = temp;
+                }
+            }
+        }
+        return temp;
```
StrikingThirteen: There is a possibility of a race condition in which get() returning an instance where resource is already cleaned up.

Maithem: How is that different from calling get from one thread and then calling destroy from another on the same object?

Having cleanup as a part of the singleton seems wrong, it breaks the semantics of a singleton (i.e. recycling it with a new object). The resource should be cleaned up externally. In other words, acquire the resource and then call clean up on it.

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
Maithem: see @StrikingThirteen 's comment.   I think its best we remove this method, it doesn't make a lot of sense.

zalokhan: Rather than removing, we can change the name to destroy. We do need to clean up the resource - in this case, shut down the runtime.
Once destroyed the resource cannot be recreated/re-instantiated.

Maithem: > We do need to clean up the resource - in this case

I understand, the last paragraph of my comment says:

> The resource should be cleaned up externally. In other words, acquire the resource and then call clean up on it.

The Singleton itself doesn't need to have a destroy, shutdown should be called on the object after it is acquired. Trying to enforce a single call to  "close/shutdown" through a singleton just doesn't seem right.

zalokhan: Ok, that sounds good.

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
StrikingThirteen: The race condition happens in the context you mentioned. However temp still holding a reference to the object even if cleaner has nulled or cleaned the resource. So even though the singleton class's resource is null. Get returns its ref to the object through temp. What about returning resource? (it's volatile so either going through initialization or not would be fine).
To you second point, I agree that cleanup breaks the semantic of singleton as we know it.

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
StrikingThirteen: If we do null out the resource the race problem that I mentioned in the other comment will be present. If we don't, probably there is no point of keeping the cleanup around.

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
```diff
@@ -66,13 +75,10 @@ public T get() {
...
-            }
-            return null;
-        });
+    public synchronized void cleanup(@Nonnull Consumer<T> cleaner) {
+        if (resource != null) {
+            cleaner.accept(resource);
+        }
     }
 
 }
```
StrikingThirteen: nit: add a newline

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
Maithem: Done.

runtime/src/main/java/org/corfudb/util/concurrent/SingletonResource.java
Maithem: Removed the cleanup logic.

Maithem:
@annym is this test stuck ?

PR #1507, opened 2018-10-14
Re-implemented the Generator class for supporting chunk-by-chunk verification

mcheng-vmware:
Provided APIs for running workloads that can be divided into chunks, of which transactions are decoupled from each other.

## Overview

Description:

Why should this be merged: 

This can help do corrrectness veification for Longevity test chunk-by-chunk on the fly.

### DIFF
--- a/generator/src/main/java/org/corfudb/generator/Generator.java
+++ b/generator/src/main/java/org/corfudb/generator/Generator.java
@@ -1,15 +1,22 @@
 package org.corfudb.generator;
 
+import lombok.extern.slf4j.Slf4j;
 import java.util.List;
+import java.time.Duration;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ArrayBlockingQueue;
 
 import org.corfudb.generator.operations.CheckpointOperation;
 import org.corfudb.generator.operations.Operation;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.unrecoverable.SystemUnavailableError;
+import org.corfudb.util.Sleep;
 
 import static java.util.concurrent.Executors.newWorkStealingPool;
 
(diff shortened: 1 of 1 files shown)


generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,196 @@
...
+
+    public Generator(String endPoint, long cpPeriod, Duration totalRuntime) {
+
+        this.cpPeriod = cpPeriod;
+        this.totalRuntime = totalRuntime;
+
+        rt = new CorfuRuntime(endPoint).connect();
+
+        try {
+            tryToConnectTimeout(TIME_TO_WAIT_FOR_RUNTIME_TO_CONNECT);
```
WenbinZhu: Is runtime already connected at line 65?

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,196 @@
...
+    static final int numThreads = 10;
+
+    BlockingQueue<Operation> operationQueue;
+    CorfuRuntime rt;
+    State state;
+    ScheduledExecutorService scheduler;
+    ExecutorService workers;
+
+    // The lock prevents consuming the queue while it's being filled.
+    Boolean operationQueueLock;
```
WenbinZhu: This variable should probably be `volatile`. Also you can make it `boolean`instead of `Boolean`
Actually I think we can remove this lock variable since `BlockingQueue` implementations are all thread safe, including `ArrayBlockingQueue`. What do you think?

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,196 @@
...
+                        Correctness.recordOperation("Liveness, " + "Success", false);
+                        break;
+                    }
+                    fillOperationQueue();
+                    lastPollTime = System.currentTimeMillis();
+                } else {
+                    // If it's been more than LIVENESS_TIMEOUT, and the queue is still not drained,
+                    // then the system might have got stuck and we record its 'Liveness' as 'Fail'.
+                    if (System.currentTimeMillis() - lastPollTime > LIVENESS_TIMEOUT.toMillis()) {
+                        Correctness.recordOperation("Liveness, " + "Fail", false);
```
WenbinZhu: Not sure if we need to continue running the remaining operations after we found some chunk already failed? If not, we should probably jump out of the whole loop here.

generator/src/main/java/org/corfudb/generator/Generator.java
mcheng-vmware: oh, forgot to remove the connect() from the line that created the runtime, will fix this.

generator/src/main/java/org/corfudb/generator/Generator.java
mcheng-vmware: Didn't notice there is a difference between 'boolean' and 'Boolean'... will change it to a boolean. I know BlockingQueue is thread safe, but it's for doing one thing at a time. However, while filling the queue, we are putting many operations one by one into the queue (see the for loop in fillOperationQueue), so I added this lock for more safety. Anyway, we can discuss this more offline.

generator/src/main/java/org/corfudb/generator/Generator.java
mcheng-vmware: I have discussed this with Anny last Friday. Should we go on running the longevity test if we already found some chunks have correctness issue? And the answer is 'yes': let's allow the test go on even some chunks failed, and it'll be an interesting phenomenon if we found only one/two chunks failed in some middle point and the following chunks all proceed well. We should leave the program running and let human to decide whether to kill a test or not.

WenbinZhu: Yeah make sense to me

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,196 @@
...
+
+    /**
+     * Each workers in the worker pool will consume an operation from the blocking queue.
+     * Once the operation de-queued, it is executed.
+     */
+    private void consumeOperationQueue() {
+        // Assign a worker for each thread in the pool
+        for (int i = 0; i < numThreads; i++) {
+            workers.execute(() -> {
+                while (!operationQueueLock && !operationQueue.isEmpty()) {
```
WenbinZhu: There might be a serious problem here: once any of these two checks failed, the thread will finish and exit, but we expect it to block until the fillOperationQueue put more operations into the queue and continue consuming, right? I think that's why we are using a  blocking queue.

mcheng-vmware: Fixed.

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -163,11 +171,23 @@ public void runLongevityTestForVerification() {
...
+
+            if (interrupted) {
+                break;
+            }
             // One round of 'chunk' starts
             long cumulativeRuntime = 0;
             while (true) {
                 Sleep.sleepUninterruptibly(POLL_TIME);
+
+                interrupted = Thread.currentThread().isInterrupted();
```
WenbinZhu: The sleepUninterruptibly method will throw an UnrecoverableCorfuInterruptedError if interrupted, so you can remove all the `interrupted` checks.

mcheng-vmware: @xnull since Wenbin has pointed out, I'll remove this.

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,201 @@
...
+                    // Break this while loop if it reaches CHUNK_RUNTIME
+                    if (cumulativeRuntime >= CHUNK_RUNTIME.toMillis()) {
+                        // At the end of the chunk, give some margin
+                        // for the workers to finish their operations
+                        Sleep.sleepUninterruptibly(CHUNK_TIMEOUT);
+                        Correctness.recordOperation("Liveness, " + "Success", false);
+                        break;
+                    }
+                    fillOperationQueue();
+                    lastPollTime = System.currentTimeMillis();
```
WenbinZhu: I was thinking about your comment on`operationQueueLock` in my first review and I found there is a race condition here. Imagine there is only one operation left in the queue, and more than one consumer thread have passed the `if (!operationQueueLock && !operationQueue.isEmpty())` check, then only one thread can take the only operation in the queue, and all the other threads will block on `operationQueue.take()`. After some time, `fillOperationQueue` starts and adding operations one by one, then some threads blocking on `take()` will consume the operations just added, even if `fillOperationQueue()` is not finished since `if (!operationQueueLock...)` check is already done before the thread is blocked. This race condition leads to the wrong calculation of totalRuntime.

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,201 @@
...
+                } else {
+                    // If it's been more than LIVENESS_TIMEOUT, and the queue is still not drained,
+                    // then the system might have got stuck and we record its 'Liveness' as 'Fail'.
+                    if (System.currentTimeMillis() - lastPollTime > LIVENESS_TIMEOUT.toMillis()) {
+                        Correctness.recordOperation("Liveness, " + "Fail", false);
```
WenbinZhu: We need to stop these consumers that are failed, otherwise the failed consumers are still running and cannot be scheduled to do the operations in the next chunk.

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -1,15 +1,15 @@
...
 
+import lombok.extern.slf4j.Slf4j;
 import java.util.List;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.time.Duration;
+import java.util.concurrent.*;
```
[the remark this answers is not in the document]
mcheng-vmware: will do.

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,201 @@
...
- *
  * Created by maithem on 7/14/17.
+ *
+ * Modified by Miao Cheng on 10/15/18:
+ *
+ * Re-implemented the Generator class to provide APIs for running workloads that can be
+ * 'divided' into chunks, of which transactions are decoupled from each other.
+ * This can help do correctness verification for Longevity test chunk-by-chunk on the fly.
  */
+@Slf4j
```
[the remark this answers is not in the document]
mcheng-vmware: Yes, you are right on this. When I first saw this file, I also think it has some duplication with the LongevityRun, I think we should merge the functionality. will sync this with Wenbin and Maithem.

generator/src/main/java/org/corfudb/generator/Generator.java
mcheng-vmware: yes, we need to fix this. will discuss with you offline.

mcheng-vmware: At second thought, I don't think what you describe will happen. will explain to you offline.

generator/src/main/java/org/corfudb/generator/Generator.java
mcheng-vmware: As I explained before, we should let the human to decide whether to stop a test or not. And here may not be the consumers' fault, i.e. the system itself got stuck can lead to the problem. We will have a correctness report reflecting the 'liveness failure', which gives enough indication for the person who run the test.

generator/src/main/java/org/corfudb/generator/Generator.java
mcheng-vmware: anyway, will discuss this offline.

generator/src/main/java/org/corfudb/generator/Generator.java
```diff
@@ -19,55 +19,201 @@
...
-            } catch (Exception e) {
-                throw new RuntimeException(e);
+            // Minus one CHUNK_RUNTIME from the TOTAL_RUNTIME
+            totalRuntime = totalRuntime.minus(CHUNK_RUNTIME);
+            if (totalRuntime.toHours() == 0) {
+                break;
             }
+            // Fill up the queue for next round of 'chunk' run
+            fillOperationQueue();
+            lastPollTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
mcheng-vmware: Anny, today I discussed your doubt with Maithem and added a 'TODO: checkpoint all the existing maps of last chunk and do prefix trim' after shipping one chunk of log to verification. Our idea is that instead of recording the last chunk's state on the verification side, we 'patched' the last state to the head of next chunk, i.e. before running the next chunk, we checkpoint all the existing maps of last chunk and do prefix trim, in this case, when next chunk starts running, it'll also have the old data.

xnull:
Not relevant anymore

PR #1511, opened 2018-10-17
Lookup by Index when Index is not specified

annym:
## Overview

The use of the getByIndex API should fail whenever the index has not
been specified for the given table. This should be followed by the
proper exception and logging. In addition we add a message field to
the transaction aborted exception info in case a message for the cause
of abort is available.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -191,6 +191,8 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
             CorfuTable.IndexRegistry<K, V> result = (CorfuTable.IndexRegistry<K, V>) EMPTY;
             return result;
         }
+
+
     }
 
     /** Helper function to get a map (non-secondary index) Corfu table.
@@ -279,14 +281,12 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
             secondaryIndexes.put(index.getName().get(), new HashMap<>());
             indexSpec.add(index);
         });
-        log.info("CorfuTable: creating CorfuTable with {} as indexRegistry", indices);
+        log.info("CorfuTable: creating CorfuTable with the following indexes: {}", secondaryIndexes.keySet().toString());
     }
 
     /** Default constructor. Generates a table without any secondary indexes. */
     public CorfuTable() {
         this(IndexRegistry.empty());
-        log.trace("CorfuTable: Creating a table without secondary indexes! Secondary index lookup"
-            + " will DEGRADE to a full scan");
     }
 
     /** {@inheritDoc} */
@@ -346,11 +346,20 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
     <I extends Comparable<I>>
     Collection<Entry<K, V>> getByIndex(@Nonnull IndexName indexName, I indexKey) {
         String secondaryIndex = indexName.get();
-        Map<K, V> res = secondaryIndexes.get(secondaryIndex).get(indexKey);
+        Map<Comparable, Map<K, V>> secondaryMap;
+        if (secondaryIndexes.containsKey(secondaryIndex) &&
+                ((secondaryMap = secondaryIndexes.get(secondaryIndex)) != null)) {
+            // If secondary index exists and function for this index is not null
+            Map<K, V> res = secondaryMap.get(indexKey);
 
-        return res == null ?
-                Collections.emptySet():
-                new HashSet<>(res.entrySet());
+            return res == null ?
+                    Collections.emptySet() :
+                    new HashSet<>(res.entrySet());
+        }
+
+        // If index is not specified, the lookup by index API must fail.
+        log.error("CorfuTable: secondary index " + secondaryIndex + " does not exist for this table, cannot complete the get by index.");
+        throw new IllegalArgumentException("Secondary Index " + secondaryIndex + " is not defined.");
     }
 
     /**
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -346,11 +346,16 @@ public V get(@ConflictParameter Object key) {
     <I extends Comparable<I>>
     Collection<Entry<K, V>> getByIndex(@Nonnull IndexName indexName, I indexKey) {
         String secondaryIndex = indexName.get();
-        Map<K, V> res = secondaryIndexes.get(secondaryIndex).get(indexKey);
-
-        return res == null ?
-                Collections.emptySet():
-                new HashSet<>(res.entrySet());
+        try {
+            Map<K, V> res = secondaryIndexes.get(secondaryIndex).get(indexKey);
```
Maithem: Can't you determine this ahead of time (i.e. before the access) instead of catching the exception post access?

annym: I'm aware of the PassThrough annotation but I'm not actually sure on how to combine both, to run this check previous to the accessor.

test/src/test/java/org/corfudb/runtime/object/transactions/StreamTest.java
```diff
@@ -289,4 +290,20 @@ public void testConcurrentStreamRW() {
...
+        CorfuTable<String, String> map = instantiateCorfuObject(CorfuTable.class, "Test");
+
+        try {
+            TXBegin();
+            map.put("ak", "av");
+            map.getByIndex(StringIndexer.BY_FIRST_LETTER, "a");
+            TXEnd();
+        } catch (TransactionAbortedException tae) {
+            assertThat(tae.getAbortCause()).isEqualTo(AbortCause.UNDEFINED);
+            assertThat(tae.getCause().equals(IllegalArgumentException.class));
```
corfu-repo: BLOCKER Complete the assertion. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2970)

corfu-repo:
* ![BLOCKER][BLOCKER] 1 blocker

[BLOCKER]: https://sonarsource.github.io/sonar-github/severity-blocker.png 'Severity: BLOCKER'

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
annym: Do we have an annotation that supports this behavior, or would I need to modify the annotationProcessor and add a new annotation for this?

Maithem: Hmm, "before the access" as in before the java object access and not SMR access. Do you catch a NPE because secondaryIndexes.get(secondaryIndex) can return null ?

annym: I thought you were referring to the SMR access, which also made sense to me, to avoid going all the way through if no actual index exists. I did the change.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -285,8 +285,8 @@ public CorfuTable(IndexRegistry<K, V> indices) {
     /** Default constructor. Generates a table without any secondary indexes. */
     public CorfuTable() {
         this(IndexRegistry.empty());
-        log.trace("CorfuTable: Creating a table without secondary indexes! Secondary index lookup"
-            + " will DEGRADE to a full scan");
+        log.warn("CorfuTable: Creating a table without secondary indexes! Secondary index lookup"
```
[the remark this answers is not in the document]
annym: I also thought of this but we do not have that info available at this level. With regards to the consolidate still the levels of log are different.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -346,11 +346,20 @@ public V get(@ConflictParameter Object key) {
...
-        return res == null ?
-                Collections.emptySet():
-                new HashSet<>(res.entrySet());
+                return res == null ?
+                        Collections.emptySet() :
+                        new HashSet<>(res.entrySet());
+            }
+        }
+
+        // If index is not specified, the lookup by index API must fail.
```
StrikingThirteen: A suggestion: what about doing the secondaryMap check on top of the method (as it is really the pre condition) and throwing the exceptions right away if it doesn't exist. This eliminates the need for other if condition at L349 and is more straightforward.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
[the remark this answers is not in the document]
StrikingThirteen: Why warning? The callers of this constructor are not recommended one way or other. Info or trace level seem more suitable.

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
```diff
@@ -62,6 +62,22 @@ public void canReadFromEachIndex() {
                 .containsExactly("ab");
     }
 
+    @Test (expected = IllegalArgumentException.class)
```
StrikingThirteen: Please add one line comment explaining the purpose of test.

test/src/test/java/org/corfudb/runtime/object/transactions/StreamTest.java
```diff
@@ -289,4 +290,20 @@ public void testConcurrentStreamRW() {
     }
 
 
+    @Test (expected = TransactionAbortedException.class)
```
StrikingThirteen: Please a one line comment explaining the purpose of test.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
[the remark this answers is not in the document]
Maithem: I suggest that this message is change to "table created with the following indexes" as opposed to printing the warning.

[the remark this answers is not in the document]
annym: Then I'm just removing the warning message because this is already printed in the other constructor.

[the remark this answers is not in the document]
Maithem: Just make sure that the toString() prints something reasonable.

Issue #1510, opened 2018-10-17
Faulty Nested Transactions


Maithem:
**Bug report.**
When nesting transactions, it is possible for a child transaction to have a snapshot of an older time than the parent. For example, the parent transaction starts at T=10, but the child transaction starts at T=5, this can cause data loss issues. In particular, the checkpoint writer will trim based on the parent's timestamp, if this time regression happen data can be lost. 

` @Test
    public void largeStreamWrite() throws Exception {
        CorfuRuntime rt = new CorfuRuntime("localhost:9000").connect();

        Map<String, String> map = rt.getObjectsView()
                .build()
                .setStreamName("s1")
                .setType(CorfuTable.class)
                .open();

        map.put("k1", "v1");
        map.put("k1", "v2");
        map.put("k1", "v3");
        map.put("k1", "v4");
        map.put("k1", "v5");

        rt.getObjectsView().TXBuild()
                .setType(TransactionType.SNAPSHOT)
                .setSnapshot(2L)
                .begin();
        rt.getObjectsView().TXBuild()
                .setType(TransactionType.SNAPSHOT)
                .setSnapshot(1L)
                .begin();
        System.out.println(map.get("k1"));
        rt.getObjectsView().TXEnd();
        System.out.println(map.get("k1"));
        rt.getObjectsView().TXEnd();
        System.out.println(map.get("k1"));
    }`

PR #1515, opened 2018-10-18
Transaction Timestamps Cleanup

Maithem:
## Overview
Enforces child transactions to inherit the root/parent
transaction timestamp.

Why should this be merged: Prevents nested transactions to have different views.

Related issue(s) (if applicable): #1510

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -137,12 +137,12 @@
...
 
-    AbstractTransactionalContext(TransactionBuilder builder) {
+    AbstractTransactionalContext(Transaction transaction) {
         transactionID = UUID.randomUUID();
-        this.builder = builder;
+        this.transaction = transaction;
 
         startTime = System.currentTimeMillis();
-
+        // TODO(Maithem) set snapshot from parent if it exists
```
annym: this TODO needs to be solved for this PR, right?

runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
```diff
@@ -76,12 +70,12 @@ public void TXBegin() {
 
         /* If it is a nested transaction, inherit type of parent */
         if (TransactionalContext.isInTransaction()) {
-            type = TransactionalContext.getCurrentContext().getBuilder().getType();
+            type = TransactionalContext.getCurrentContext().getTransaction().getType();
```
annym: Is it that we don't inherit snapshot from this point because in the case of a TX different than snapshot we might have not yet accessed an object, is this assumption right? should we add a comment on why we do not directly inherit snapshot timestamp at this point?

Maithem:
@annym This was a bad attempt to do the clean up, I addressed that here #1517 , we keep the lazy ts evaluation behavior

Maithem:
hmm, some tests seem to be failing any idea why ? https://travis-ci.org/CorfuDB/CorfuDB/builds/443833806?utm_source=github_status&utm_medium=notification

PR #1520, opened 2018-10-22
Enable Epoch Checks

Maithem:
## Overview
At some point we needed to bypass epoch checks to recover the
sequencer, now it is not the case.

Why should this be merged: Reading state during configuration can be
dangerous if the epochs are not verified. 

- [x] Public API has Javadoc


Maithem:
@zalokhan looked back at the code and I believe its ok to enforce these epochs. Calls made to LogUnitClient.getTail come from the AddressSpaceView and Utils. Only calls from the latter are used for reconfiguration and the method signature accepts a layout. Also, its usually called after a seal, or a correct layout, so the rpc ends up being stamped with a valid epoch.

Maithem:
@zalokhan see my first comment.

Issue #1528, opened 2018-10-25
Properly Seal Servers


Maithem:
**Bug report.**

We have logic that depends on the seal to flush all pending requests on the system from previous epochs. Since the seal only sets the epoch on the router, but not flush requests requests from an older epoch can interleave with requests from the new epoch resulting in undesired behavior. 

To be more concrete, when the the faster loader will request the log tail from the logging units after a seal operation. If a write operation from an older epoch executes after the seal and before the tail request happens, then the faster loader will build state based on an incorrect tail. We need to make sure to flush operations on a seal. Lets focus on the logging unit and think about this can apply to other servers.

WenbinZhu:
Addressed in #1545

PR #1545, opened 2018-11-08
Correctly seal servers on epoch change

WenbinZhu:
## Overview

The current seal logic is not correct as described in #1528. This patch fix this issue by setting remote epoch on all servers and execute correct sealing logic on each server types. For now only the log unit servers has the needs to do the additional sealing logic, which flushes all the operations before the seal and fail all the operations after this, but stamped with an older epoch.

This patch also fix the incorrect behavior when flushing and failing operations on log unit servers with BatchWriter, which could cause tests to fail intermittently.

Related issue(s) (if applicable): #1528

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
@@ -28,10 +28,21 @@ public abstract class AbstractServer {
         shutdown = false;
     }
 
-    /** Get the message handler for this instance.
-     * @return  A message handler.
+    /**
+     * Get the message handler for this instance.
+     *
+     * @return A message handler.
      */
     public abstract CorfuMsgHandler getHandler();
+
+    /**
+     * Seal the server with the epoch.
+     *
+     * @param epoch Epoch to seal with
+     */
+    public void sealServerWithEpoch(long epoch) {
+        // Overridden in log unit to flush operations stamped with an old epoch
+    }
 
     public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
         // Overridden in sequencer to mark ready/not-ready state.
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
@@ -31,8 +31,8 @@ public class BaseServer extends AbstractServer {
     private final CorfuMsgHandler handler =
             CorfuMsgHandler.generateHandler(MethodHandles.lookup(), this);
 
-
-    /** Respond to a ping message.
+    /**
+     * Respond to a ping message.
      *
      * @param msg   The incoming message
      * @param ctx   The channel context
@@ -43,7 +43,8 @@ public class BaseServer extends AbstractServer {
         r.sendResponse(ctx, msg, CorfuMsgType.PONG.msg());
     }
 
-    /** Respond to a version request message.
+    /**
+     * Respond to a version request message.
      *
      * @param msg   The incoming message
      * @param ctx   The channel context
@@ -58,6 +59,7 @@ public class BaseServer extends AbstractServer {
 
     /**
      * Respond to a epoch change message.
+     * This method also executes sealing logic on each individual server type.
      *
      * @param msg The incoming message
      * @param ctx The channel context
@@ -68,9 +70,10 @@ public class BaseServer extends AbstractServer {
                                                    ChannelHandlerContext ctx,
                                                    @NonNull IServerRouter r) {
         try {
-            log.info("handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch {}",
-                    msg.getPayload());
-            serverContext.setServerEpoch(msg.getPayload(), r);
+            long epoch = msg.getPayload();
+            log.info("handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch {}", epoch);
+            serverContext.setServerEpoch(epoch, r);
+            serverContext.getServers().forEach(s -> s.sealServerWithEpoch(epoch));
             r.sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.ACK));
         } catch (WrongEpochException e) {
             log.debug("handleMessageSetEpoch: Rejected SET_EPOCH current={}, requested={}",
@@ -80,7 +83,8 @@ public class BaseServer extends AbstractServer {
         }
     }
 
-    /** Reset the JVM. This mechanism leverages that corfu_server runs in a bash script
+    /**
+     * Reset the JVM. This mechanism leverages that corfu_server runs in a bash script
      * which monitors the exit code of Corfu. If the exit code is 100, then it resets
      * the server and DELETES ALL EXISTING DATA.
      *
@@ -95,7 +99,8 @@ public class BaseServer extends AbstractServer {
         CorfuServer.restartServer(serverContext, true);
     }
 
-    /** Restart the JVM. This mechanism leverages that corfu_server runs in a bash script
+    /**
+     * Restart the JVM. This mechanism leverages that corfu_server runs in a bash script
      * which monitors the exit code of Corfu. If the exit code is 200, then it restarts
      * the server.
      *
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
@@ -45,24 +45,22 @@ public class BatchWriter<K, V> implements CacheWriter<K, V>, AutoCloseable {
                     .build());
 
     /**
-     * The epochWaterMark is the epoch up to which all operations have been sealed. Any
-     * BatchWriterOperation arriving after the epochWaterMark with an epoch less than the epochWaterMark, is
-     * completed exceptionally with a WrongEpochException.
+     * The sealEpoch is the epoch up to which all operations have been sealed. Any
+     * BatchWriterOperation arriving after the sealEpoch with an epoch less than the sealEpoch
+     * is completed exceptionally with a WrongEpochException.
      * This is persisted in the ServerContext by the LogUnitServer to withstand restarts.
      */
-    private volatile long epochWaterMark;
+    private long sealEpoch;
 
     /**
      * Returns a new BatchWriter for a stream log.
      *
-     * @param streamLog      stream log for writes (can be in memory or file)
-     * @param epochWaterMark All operations stamped with epoch less than the epochWaterMark are
-     *                       discarded.
-     * @param streamLog stream log for writes (can be in memory or file)
-     * @param doSync    If true, the batch writer will sync writes to secondary storage
+     * @param streamLog  Stream log for writes (can be in memory or file)
+     * @param sealEpoch  All operations stamped with epoch less than the sealEpoch are discarded
+     * @param doSync     If true, the batch writer will sync writes to secondary storage
      */
-    public BatchWriter(StreamLog streamLog, long epochWaterMark, boolean doSync) {
-        this.epochWaterMark = epochWaterMark;
+    public BatchWriter(StreamLog streamLog, long sealEpoch, boolean doSync) {
+        this.sealEpoch = sealEpoch;
         this.doSync = doSync;
         this.streamLog = streamLog;
         operationsQueue = new LinkedBlockingQueue<>();
@@ -136,22 +134,17 @@ public class BatchWriter<K, V> implements CacheWriter<K, V>, AutoCloseable {
     }
 
     /**
-     * Insert epochWaterMark in queue and wait for queue to process all preceding operations.
-     * All operations in the queue after the epochWaterMark operation, that have epoch less than
-     * the epochWaterMark epoch are discarded and their futures are completed exceptionally with a
-     * WrongEpochException. The epochWaterMark is used in case of:
-     * Reset - All operations need to be flushed before a reset and no new operations for the
-     * previous epoch should be accepted.
-     * SealAndFlush - Similarly on a seal, all operations need to be flushed before a seal and no
-     * new operations for the previous epoch should be accepted.
+     * Insert seal operation in queue and wait for queue to process all preceding operations.
+     * All operations in the queue after the sealEpoch operation, that have epoch less than
+     * the sealEpoch epoch are discarded and their futures are completed exceptionally with a
+     * WrongEpochException.
      *
-     * @param epoch Epoch to epochWaterMark with.
+     * @param epoch Epoch to seal with.
      */
-    public void waitForEpochWaterMark(@Nonnull long epoch) {
+    public void waitForSealComplete(long epoch) {
         try {
             CompletableFuture<Void> cf = new CompletableFuture<>();
-            epochWaterMark = epoch;
-            operationsQueue.add(new BatchWriterOperation(Type.EPOCH_WATER_MARK, null, null, epoch, null, cf));
+            operationsQueue.add(new BatchWriterOperation(Type.SEAL, null, null, epoch, null, cf));
             cf.get();
         } catch (Exception e) {
             throw new RuntimeException(e);
@@ -192,7 +185,7 @@ public class BatchWriter<K, V> implements CacheWriter<K, V>, AutoCloseable {
         try {
             BatchWriterOperation lastOp = null;
             int processed = 0;
-            List<BatchWriterOperation> res = new LinkedList();
+            List<BatchWriterOperation> res = new LinkedList<>();
 
             while (true) {
                 BatchWriterOperation currOp;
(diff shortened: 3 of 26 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -28,11 +28,22 @@ public AbstractServer() {
...
+     * @return A message handler.
      */
     public abstract CorfuMsgHandler getHandler();
 
+    /**
+     * Seal the server with the epoch.
+     *
+     * @param epoch Epoch to seal with
+     */
+    public void sealServerWithEpoch(long epoch, IServerRouter r) {
```
xnull: Can you make it abstract?

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -150,7 +150,6 @@ public void prefixTrim(@Nonnull long address, @Nonnull long epoch) {
     public void waitForEpochWaterMark(@Nonnull long epoch) {
```
xnull: please delete `@Nonnull` it's useless for primitives

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -521,10 +521,8 @@ public static void restartServer(ServerContext serverContext, boolean resetData)
     /**
      * Attempt to cleanly shutdown all the servers.
      */
-    public static void cleanShutdown(@Nonnull NettyServerRouter router) {
+    public static void cleanShutdown(@Nonnull List<AbstractServer> servers) {
```
xnull: It seems to me that use `NettyServerRouter` is better than a list of abstract servers. What was the reason, why you changed it?

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
```diff
@@ -24,4 +26,9 @@
      * @param server    The server to route messages to
      */
     void addServer(AbstractServer server);
+
+    /**
+     * Get a list of registered servers.
+     */
+    List<AbstractServer> getServers();
```
xnull: ImmutableList

xnull: Why did you decide to expose a list of servers?

xnull: Even `addServer` method is deprecated

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
Maithem: Can you get rid of the IServerRouter and just call it Seal ?

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
WenbinZhu: Also the current code already exposed getServers() logic in NettyServerRouter (lombok getter), I just pushed the logic to interface level.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -285,12 +285,35 @@ private void rangeWrite(CorfuPayloadMsg<RangeWriteMsg> msg,
...
+     * Seal the server with the epoch by setting epochWaterMark.
+     *
+     * - The epochWaterMark is persisted to withstand restarts.
+     * - The epochWaterMark is inserted in the queue and then we wait to flush all operations
+     *   in the queue before this operation.
+     * - All operations after this operation but stamped with an older epoch will be failed.
+     */
+    @Override
+    public synchronized void sealServerWithEpoch(long epoch, IServerRouter r) {
+        if (epoch > serverContext.getLogUnitEpochWaterMark()) {
```
Maithem: Just pass the epoch the router is not needed. Only the base server needs to know about the router (i.e. set in mem epoch)

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: Done, I forgot to remove it.

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
WenbinZhu: As discussed, we keep it a normal method.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -285,12 +285,35 @@ private void rangeWrite(CorfuPayloadMsg<RangeWriteMsg> msg,
...
+    /**
+     * Seal the server with the epoch by setting epochWaterMark.
+     *
+     * - The epochWaterMark is persisted to withstand restarts.
+     * - The epochWaterMark is inserted in the queue and then we wait to flush all operations
+     *   in the queue before this operation.
+     * - All operations after this operation but stamped with an older epoch will be failed.
+     */
+    @Override
+    public synchronized void sealServerWithEpoch(long epoch, IServerRouter r) {
```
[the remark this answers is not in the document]
WenbinZhu: To prevent concurrent call with resetLogUnit

runtime/src/main/java/org/corfudb/runtime/view/RuntimeLayout.java
```diff
@@ -57,18 +56,18 @@ public RuntimeLayout(@Nonnull Layout layout, @Nonnull CorfuRuntime corfuRuntime)
...
-    public void moveServersToEpoch()
+    public void sealMinServerSet()
             throws WrongEpochException, QuorumUnreachableException {
         log.debug("Requested move of servers to new epoch {} servers are {}", layout.getEpoch(),
                 layout.getAllServers());
 
         // Set remote epoch on all servers in layout.
-        Map<String, CompletableFuture<Boolean>> resultMap =
-                SealServersHelper.asyncSetRemoteEpoch(this);
+        Map<String, CompletableFuture<Boolean>> resultMap = SealServersHelper.asyncSealServers(this);
```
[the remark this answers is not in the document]
WenbinZhu: We need to get BaseClient in this method from RuntimeLayout, have to pass itself.

runtime/src/main/java/org/corfudb/runtime/view/SealServersHelper.java
```diff
@@ -24,16 +25,18 @@
...
      * @param runtimeLayout RuntimeLayout stamped with the layout to be sealed.
      * @return A map of completableFutures for every remoteSetEpoch call.
      */
-    public static Map<String, CompletableFuture<Boolean>> asyncSetRemoteEpoch(
+    public static Map<String, CompletableFuture<Boolean>> asyncSealServers(
             RuntimeLayout runtimeLayout) {
         Layout layout = runtimeLayout.getLayout();
         Map<String, CompletableFuture<Boolean>> resultMap = new HashMap<>();
-        // Seal layout servers
+        // Seal all servers
```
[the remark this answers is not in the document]
WenbinZhu: where is empty try block?

test/src/test/java/org/corfudb/infrastructure/TestServerRouter.java
```diff
@@ -30,6 +30,9 @@
     @Getter
     public Map<CorfuMsgType, AbstractServer> handlerMap;
 
+    @Getter
+    public List<AbstractServer> servers;
```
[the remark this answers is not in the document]
WenbinZhu: close as discussed

runtime/src/main/java/org/corfudb/runtime/view/SealServersHelper.java
```diff
@@ -94,13 +93,12 @@ public static void waitForChainSegmentSeal(Layout.LayoutSegment layoutSegment, M
...
             }
 
-            int reachableServers = (int) Arrays.stream(completableFutures)
-                    .filter(booleanCompletableFuture ->
-                            !booleanCompletableFuture.isCompletedExceptionally()).count();
-
             if (!success) {
-                throw new QuorumUnreachableException(reachableServers,
-                        completableFutures.length);
+                int reachableServers = (int) Arrays.stream(completableFutures)
```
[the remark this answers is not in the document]
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
WenbinZhu: As discussed, keep the current behavior, router shouldn't be aware of sealing.

WenbinZhu:
Addressed comments and Added a unit test.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -285,12 +284,35 @@ private void rangeWrite(CorfuPayloadMsg<RangeWriteMsg> msg,
...
+     * - The epochWaterMark is persisted to withstand restarts.
+     * - The epochWaterMark is inserted in the queue and then we wait to flush all operations
+     *   in the queue before this operation.
+     * - All operations after this operation but stamped with an older epoch will be failed.
+     */
+    @Override
+    public synchronized void sealServerWithEpoch(long epoch) {
+        if (epoch > serverContext.getLogUnitEpochWaterMark()) {
+            serverContext.setLogUnitEpochWaterMark(epoch);
+            batchWriter.waitForEpochWaterMark(epoch);
```
Maithem: Can you get rid of the getLogUnitEpochWaterMark? the server epoch is enough. This is bad code that we need to clean up. When the base server starts it can can seal on the other servers with the current server epoch.

WenbinZhu: As discussed, will remove the getLogUnitEpochWaterMark from this method, but keep it in resetLogUnit to prevent multiple reset in same epoch.

WenbinZhu: @zalokhan @Maithem  I removed epochWaterMark from sealing, removed flush logic from reset and also did some renaming. Please review my latest commit in this pr. I'll squash them later.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -45,24 +45,22 @@
...
-     * The epochWaterMark is the epoch up to which all operations have been sealed. Any
-     * BatchWriterOperation arriving after the epochWaterMark with an epoch less than the epochWaterMark, is
-     * completed exceptionally with a WrongEpochException.
+     * The sealEpoch is the epoch up to which all operations have been sealed. Any
+     * BatchWriterOperation arriving after the sealEpoch with an epoch less than the sealEpoch
+     * is completed exceptionally with a WrongEpochException.
      * This is persisted in the ServerContext by the LogUnitServer to withstand restarts.
      */
-    private volatile long epochWaterMark;
+    private volatile long sealEpoch;
```
Maithem: I know you didn't write this, but just an fyi this doesnt need to be a volatile.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -285,25 +284,37 @@ private void rangeWrite(CorfuPayloadMsg<RangeWriteMsg> msg,
...
 
+    /**
+     * Seal the server with the epoch.
+     *
+     * - A seal operation is inserted in the queue and then we wait to flush all operations
+     *   in the queue before this operation.
+     * - All operations after this operation but stamped with an older epoch will be failed.
+     */
+    @Override
+    public synchronized void sealServerWithEpoch(long epoch) {
```
Maithem: why is this synchronized ?

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
```diff
@@ -70,6 +69,14 @@ public void addServer(AbstractServer server) {
         throw new UnsupportedOperationException("No longer supported");
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<AbstractServer> getServers() {
+        return ImmutableList.copyOf(servers);
```
Maithem: What's the point of returning immutable copies, why not just make the class field immutable and return that?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -77,7 +77,7 @@
 
     // LogUnit Server
     private static final String PREFIX_LOGUNIT = "LOGUNIT";
-    private static final String EPOCH_WATER_MARK = "EPOCH_WATER_MARK";
+    private static final String EPOCH_WATER_MARK = "SEAL";
```
Maithem: resetLayoutDatastore uses the old string (i.e. EPOCH_WATER_MARK)

test/src/test/java/org/corfudb/runtime/view/LogUnitSealTest.java
```diff
@@ -0,0 +1,105 @@
...
+        client = runtimeLayout.getLogUnitClient(SERVERS.ENDPOINT_0);
+
+        // Using the new client to send messages stamped with epoch 2
+        for (int i = 0; i < numOp; i++) {
+            LogData ld = LogData.getEmpty(address++);
+            successOpFutures2.add(client.write(ld));
+        }
+
+        // Before sealing, all operations should complete normally
+        successOpFutures1.forEach(f -> assertThatCode(f::join).doesNotThrowAnyException());
```
Maithem: Does this actually wait on the futures to complete? 
If not, then there could be a race condition, where this check is executed before the future is completed.

test/src/test/java/org/corfudb/runtime/view/LogUnitSealTest.java
```diff
@@ -0,0 +1,105 @@
...
+        // After sealing, operations stamped with old epoch should fail
+        failedOpFutures.forEach(f -> assertThatThrownBy(f::join)
+                .isInstanceOf(CompletionException.class)
+                .hasCauseExactlyInstanceOf(WrongEpochException.class));
```
Maithem: How do you know that this exception is thrown by BatchWriter and not the base server?

test/src/test/java/org/corfudb/runtime/view/LogUnitSealTest.java
WenbinZhu: Yes, assertThatCode() will wait on future to complete, I also did an experiment to verify it.

test/src/test/java/org/corfudb/runtime/view/LogUnitSealTest.java
WenbinZhu: I manually reset the router's epoch: https://github.com/CorfuDB/CorfuDB/blob/5af64b5457606f4f6bf46c18812871945cb84bc4/test/src/test/java/org/corfudb/runtime/view/LogUnitSealTest.java#L77 

Also I verify it through the test logs.

test/src/test/java/org/corfudb/runtime/view/LogUnitSealTest.java
```diff
@@ -0,0 +1,105 @@
...
+        final int numOp = 3;
+        long address = 0L;
+
+        for (int i = 0; i < numOp; i++) {
+            LogData ld = LogData.getEmpty(address++);
+            successOpFutures1.add(client.write(ld));
+        }
+
+        // Seal server
+        layout.setEpoch(layout.getEpoch() + 1);
```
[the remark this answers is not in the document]
WenbinZhu: Done. Thanks for clarifying.

WenbinZhu:
@Maithem Clarified and addressed, thanks for reviewing.

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -68,9 +70,10 @@ public synchronized void handleMessageSetEpoch(@NonNull CorfuPayloadMsg<Long> ms
                                                    ChannelHandlerContext ctx,
                                                    @NonNull IServerRouter r) {
         try {
-            log.info("handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch {}",
-                    msg.getPayload());
-            serverContext.setServerEpoch(msg.getPayload(), r);
+            long epoch = msg.getPayload();
+            log.info("handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch {}", epoch);
+            serverContext.setServerEpoch(epoch, r);
+            serverContext.getServers().forEach(s -> s.sealServerWithEpoch(epoch));
```
Maithem: Why didn't we push this logic into setServerEpoch (instead of doing it in the handler) ?

PR #1550, opened 2018-11-13
Make integration tests more generic

xnull:
## Overview

Description:
Make integration tests more generic:
 - open random `open` port for a corfu server, which allows running integration tests in parallel in the future
 - get rid of hard-coded names for nodes in tests

Why should this be merged: 
 - allows running integration tests in parallel in the future

- [x] Public API has Javadoc

### DIFF
--- a/it/src/main/java/org/corfudb/universe/group/cluster/CorfuCluster.java
+++ b/it/src/main/java/org/corfudb/universe/group/cluster/CorfuCluster.java
@@ -1,6 +1,8 @@
 package org.corfudb.universe.group.cluster;
 
+import org.corfudb.universe.node.NodeException;
 import org.corfudb.universe.node.client.LocalCorfuClient;
+import org.corfudb.universe.node.server.CorfuServer;
 
 /**
  * Provides a Corfu specific cluster of servers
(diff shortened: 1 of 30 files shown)


it/src/main/java/org/corfudb/universe/group/cluster/CorfuCluster.java
```diff
@@ -9,8 +11,38 @@
...
+     *
+     * @param index corfu server position
+     * @return a corfu server
+     */
+    default CorfuServer getServerByIndex(int index) {
+        return nodes()
+                .values()
+                .stream()
+                .skip(index)
+                .findFirst()
```
WenbinZhu: You can just use `return (CorfuServer) nodes().values().asList().get(index)` since .values() is already sorted by key.  This is more clear.

it/src/main/resources/logback.xml
```diff
@@ -19,7 +19,7 @@
     <logger name="org.corfudb.runtime.view.workflows.WorkflowRequest" level="off"/>
 
     <!-- Control logging levels for individual components here. -->
-    <root level="info">
+    <root level="debug">
```
WenbinZhu: Please revert.

it/src/main/java/org/corfudb/universe/node/server/CorfuServer.java
```diff
@@ -63,7 +63,7 @@
     default String getEndpoint() {
         return getNetworkInterface() + ":" + getParams().getPort();
     }
-
+
```
WenbinZhu: NIT: remove leading space.

it/src/main/java/org/corfudb/universe/group/cluster/CorfuCluster.java
xnull: Then I have to check if the value is null and throw an exception, in my implementation it's already covered

it/src/main/java/org/corfudb/universe/node/server/CorfuServer.java
xnull: done

it/src/main/resources/logback.xml
xnull: done

it/src/main/java/org/corfudb/universe/group/cluster/CorfuCluster.java
xnull: Then I have to check if the value is null and throw an exception, in my implementation it's already covered

it/src/main/resources/logback.xml
xnull: done

it/src/main/java/org/corfudb/universe/node/server/CorfuServer.java
xnull: done

PR #1562, opened 2018-11-16
Implemented Garbage Collection For CorfuRuntime Objects

Maithem:
## Overview
Implemented garbage collection for the runtime objects.

Why should this be merged: Fixes a memory leak issue in the stream layer

Related issue(s) (if applicable): #1442

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -33,6 +50,8 @@
 import org.corfudb.runtime.exceptions.WrongClusterException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
+import org.corfudb.runtime.object.CorfuCompileProxy;
+import org.corfudb.runtime.object.ICorfuSMR;
```
Maithem: push this down to the ObjectsView class, this class doesn't need to know about these types.

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamView.java
```diff
@@ -39,6 +39,8 @@
      */
     void reset();
 
+    void gc(long trimMark);
```
Maithem: Add comments and add a test.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -928,9 +893,49 @@ public synchronized CorfuRuntime connect() {
...
     }
 
+    /**
+     * Over time garbage is created on the global log and clients clients sync
+     * parts of the log therefore clients can accumulate garbage overtime. This
+     * method runs a garbage collection process on the different client views,
+     * which discards data before the trim mark (i.e. the point that demarcates
+     * the beginning of the active log).
+     */
+    private void runRuntimeGC() {
```
xnull: Can you extract GC logic? Create a class for that.
What I am worried about - CorfuRuntime comes bigger again. And also runtime shouldn't be in charge of gc logic

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -928,9 +893,49 @@ public synchronized CorfuRuntime connect() {
...
+                    continue;
+                }
+
+                log.debug("runRuntimeGC: starting gc cycle, removing {} to {}", trimMark,
+                        currTrimMark);
+                long startTs = System.currentTimeMillis();
+                getObjectsView().gc(currTrimMark);
+                getAddressSpaceView().gc(currTrimMark);
+                long endTs = System.currentTimeMillis();
+                trimMark = currTrimMark;
```
xnull: You don't need to have `trimMark` as a field in this class, you can pass it as a parameter in your method

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -55,6 +56,22 @@ public BackpointerStreamView(final CorfuRuntime runtime,
...
+    public void gc(long trimMark) {
+        NavigableSet<Long> q1 = getCurrentContext().readCpQueue;
+        NavigableSet<Long> q2 = getCurrentContext().readQueue;
+        NavigableSet<Long> q3 = getCurrentContext().resolvedQueue;
+        // TODO(Maithem): set slicing might be more efficiency than this
+        q1.removeIf(e -> e < trimMark);
+        q2.removeIf(e -> e < trimMark);
+        q3.removeIf(e -> e < trimMark);
+
+        if (!q3.isEmpty()) {
```
Maithem: If this is empty should it be reset to non_address instead?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: It is thought. We GC the runtime views. Adding a class just for one small method seems superfluous.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -928,15 +992,57 @@ public synchronized CorfuRuntime connect() {
...
+                log.info("runRuntimeGC: completed gc in {}ms on {} object(s), new trimMark {}",
+                        endTs - startTs, getObjectsView().getObjectCache().size(), trimMark);
+            } catch (Exception e) {
+                if (e.getCause() instanceof InterruptedException) {
+                    Thread.currentThread().interrupt();
+                    throw new RuntimeException(e);
```
xnull: Do not throw a generic exception, please use something else

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
xnull: Don't agree, CorfuRuntime is gigantic, adding more and more methods in CorfuRuntime that the runtime not in charge of leads us to troubles. 
If you extract Gc part you can add metrics for that, write a unit test, check that gc actually works, you can move a couple of fileds from CorfuRuntime to you gc class. 
It will be much better and have so many methods in corfuruntime

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -55,6 +56,22 @@ public BackpointerStreamView(final CorfuRuntime runtime,
         this.options = StreamOptions.DEFAULT;
     }
 
+    @Override
+    public void gc(long trimMark) {
+        NavigableSet<Long> q1 = getCurrentContext().readCpQueue;
+        NavigableSet<Long> q2 = getCurrentContext().readQueue;
+        NavigableSet<Long> q3 = getCurrentContext().resolvedQueue;
+        // TODO(Maithem): set slicing might be more efficiency than this
```
zalokhan: Yes, you can explore using something like
`resolvedQueue.headSet(trimMark).clear();`

test/src/test/java/org/corfudb/runtime/view/RuntimeViewsTest.java
```diff
@@ -0,0 +1,63 @@
...
+        MultiCheckpointWriter mcw = new MultiCheckpointWriter();
+        mcw.addMap(table);
+        Token trimMark = mcw.appendCheckpoints(rt, "cp1");
+        rt.getAddressSpaceView().prefixTrim(trimMark.getSequence());
+        rt.getParameters().setRuntimeGCPeriod(Duration.ofMinutes(0));
+        rt.runRuntimeGC();
+        long sizeOfTableAfterGc = sizeOf.deepSizeOf(table);
+
+        assertThat(sizeOfTableAfterGc).isLessThan(sizeOfTableAfterWrite);
+        assertThat(rt.getAddressSpaceView().getReadCache().asMap()).isEmpty();
```
zalokhan: Can we also assert on the resolvedQueue being trimmed in the StreamsView?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamView.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: Not sure if you've seen the latest update, but there is a test.

test/src/test/java/org/corfudb/runtime/view/RuntimeViewsTest.java
Maithem: Exposing the context to the test is very messy. Line 60 tests this indirectly. I made the condition tighter:
assertThat(sizeOfTableAfterGc).isLessThan(sizeOfTableAfterWrite - (numWrites * Long.BYTES));

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: Say no more fam. Done.

test/src/test/java/org/corfudb/runtime/view/ViewsGarbageCollectorTest.java
```diff
@@ -0,0 +1,63 @@
...
+            table.put(String.valueOf(x), String.valueOf(x));
+        }
+
+        long sizeOfTableAfterWrite = sizeOf.deepSizeOf(table);
+
+        MultiCheckpointWriter mcw = new MultiCheckpointWriter();
+        mcw.addMap(table);
+        Token trimMark = mcw.appendCheckpoints(rt, "cp1");
+        rt.getAddressSpaceView().prefixTrim(trimMark.getSequence());
+        rt.getParameters().setRuntimeGCPeriod(Duration.ofMinutes(0));
```
WenbinZhu: Since the scheduler is already started, is this useful here?

test/src/test/java/org/corfudb/runtime/view/ViewsGarbageCollectorTest.java
```diff
@@ -0,0 +1,63 @@
...
+        rt.getGarbageCollector().runRuntimeGC();
+        long sizeOfTableAfterGc = sizeOf.deepSizeOf(table);
+        final int numOfSetsStreamSets = 3;
```
WenbinZhu: Please rename this and add a comment.

test/src/test/java/org/corfudb/runtime/view/ViewsGarbageCollectorTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.time.Duration;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Created by maithem on 11/16/18.
+ */
+
+public class ViewsGarbageCollectorTest extends AbstractViewTest {
```
WenbinZhu: please add a javadoc comment here to illustrate how gc is tested against memory leak.

PR #1565, opened 2018-11-22
LZ4 compression

Maithem:
## Overview
Introduce compression to the transport. 

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
@@ -25,6 +25,8 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.channel.ServerChannel;
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
 import io.netty.handler.codec.LengthFieldPrepender;
+import io.netty.handler.codec.compression.Lz4FrameDecoder;
+import io.netty.handler.codec.compression.Lz4FrameEncoder;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslHandler;
 
@@ -43,6 +45,8 @@ import javax.net.ssl.SSLException;
 
 import lombok.extern.slf4j.Slf4j;
 
+import net.jpountz.lz4.LZ4Factory;
+import net.jpountz.xxhash.XXHashFactory;
 import org.apache.commons.io.FileUtils;
 import org.corfudb.protocols.wireprotocol.NettyCorfuMessageDecoder;
 import org.corfudb.protocols.wireprotocol.NettyCorfuMessageEncoder;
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -274,6 +276,8 @@ private ChannelInitializer getChannelInitializer() {
             protected void initChannel(@Nonnull Channel ch) throws Exception {
                 ch.pipeline().addLast(new IdleStateHandler(parameters.getIdleConnectionTimeout(),
                         parameters.getKeepAlivePeriod(), 0));
+                ch.pipeline().addLast(new Lz4FrameDecoder());
+                ch.pipeline().addLast(new Lz4FrameEncoder());
```
Maithem: This doesn't make sense: compression after encryption will probably not benefit from size reduction (since tls will produce a "random" stream of bytes). Change the order Compress -> encrypt.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -477,6 +479,10 @@ protected void initChannel(@Nonnull Channel ch) throws Exception {
                     ch.pipeline().addLast("sasl/plain-text", new
                             PlainTextSaslNettyServer());
                 }
+
+                ch.pipeline().addLast(new Lz4FrameDecoder());
+                ch.pipeline().addLast(new Lz4FrameEncoder());
```
Maithem: The test is failing because the block size defaults to 64kb, so at the 64kb boundary the buffer is is compressed and sent even if the message is greater than 64kb this breaks the NettyCorfuMessageEncoder because it assumes the buffer is not segmented (i.e. contains the whole message).

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageDecoder.java
```diff
@@ -19,7 +19,9 @@
     @Override
     protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteBuf,
                           List<Object> list) throws Exception {
+        System.out.println();
```
medhavidhawan: You need to remove the system out.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -477,6 +481,16 @@ protected void initChannel(@Nonnull Channel ch) throws Exception {
                     ch.pipeline().addLast("sasl/plain-text", new
                             PlainTextSaslNettyServer());
                 }
+
+                Lz4FrameEncoder encoder = new Lz4FrameEncoder(LZ4Factory.fastestInstance(),
```
medhavidhawan: This initialization is repeated on client side as well . Can we do it in one place. Are there any common utils or does the packaging prevent it ?

PR #1566, opened 2018-11-27
Flaky Test

Maithem:
## Overview
Relaxed the size check of ViewsGarbageCollectorTest, there seems
to be a time issue that causes the test to fail intermittently.

Why should this be merged: Fixes a test that fails intermittently.   

- [x] Public API has Javadoc


Maithem:
@xnull I think its related to some internal state being created which is creating a time race. I tried to expose the internal data structures so that they can be checked directly from the test, but its really messy. Maybe you can propose something?

Maithem:
@xnull The test's failure rate is not that high, I'll hold of this PR while I think of a cleaner way to expose the stream's internal data structures.

xnull:
Not needed anymore

PR #1572, opened 2018-11-28
Failed Authentication

annym:
## Overview

- Corfu Server log statement to report the target IP that failed to authenticate.
- Client exponential back off for reconnects upon authentication failure.


infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -462,10 +460,29 @@ protected void initChannel(@Nonnull Channel ch) throws Exception {
...
-                    ch.pipeline().addLast("ssl", new SslHandler(engine));
+
+                    SslHandler sslHandler = new SslHandler(engine);
+
+                    // Add listener to handle SSL Handshake Exceptions in order to properly log
+                    // the target IP failing to authenticate against this server.
+                    sslHandler.handshakeFuture().addListener(channelFuture -> {
+                        if (channelFuture.cause() != null && channelFuture.cause().getClass()
+                                .equals(SSLHandshakeException.class)) {
+                            String peerHost;
```
Maithem: Can't you just use 
`ch.remoteAddress()`
instead of casting?

Example:
https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java#L153

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -462,10 +460,29 @@ protected void initChannel(@Nonnull Channel ch) throws Exception {
...
+                            try {
+                                peerHost = ((java.net.InetSocketAddress) ch.remoteAddress()).getAddress().getHostAddress();
+                            } catch (Exception e) {
+                                peerHost = "Not determined.";
+                            }
+                            log.error("CorfuServer authentication failed for target IP: {}", peerHost);
```
Maithem: Can you revise the wording? The clients try to authenticate with the server so is the ip really a target?
Consider:
`Client authentication failed (remote address {address})`

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -191,6 +209,10 @@ public NettyClientRouter(@Nonnull NodeLocator node,
         outstandingRequests = new ConcurrentHashMap<>();
         shutdown = true;
 
+        sslAuthenticationFailure = new AtomicBoolean();
+        retryPolicy = new ExponentialBackoffRetry(null);
+        retryPolicy.setMaxRetryThreshold(maxRetryThreshold);
```
Maithem: Can you make the ExponentialBackoffRetry object immutable and pass the the threshold in the constructor ?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -330,6 +352,12 @@ private void addReconnectionOnCloseFuture(@Nonnull Channel channel,
             // If we aren't shutdown, reconnect.
             if (!shutdown) {
                 log.info("addReconnectionOnCloseFuture[{}]: reconnecting", node);
+                if (sslAuthenticationFailure.get()) {
+                    retryPolicy.nextWait();
+                } else {
+                    // reset exponential time retry policy, to start from beginning on next failure.
+                    retryPolicy.reset();
```
Maithem: This seems a bit strange if you have an object that encapsulates the exponential backoff logic why do you need to invoke reset externally?  Shouldn't retryPolicy.nextWait reset itself internally ?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -330,6 +352,12 @@ private void addReconnectionOnCloseFuture(@Nonnull Channel channel,
             // If we aren't shutdown, reconnect.
             if (!shutdown) {
                 log.info("addReconnectionOnCloseFuture[{}]: reconnecting", node);
+                if (sslAuthenticationFailure.get()) {
+                    retryPolicy.nextWait();
```
Maithem: Can we make nextWait return the next value, so that we can invoke the sleep logic directly? 
That way we can add logging with the right context.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -629,6 +662,16 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                 ctx.close();
             } else if (e.state() == IdleState.WRITER_IDLE) {
                 keepAlive();
+            } else if (evt instanceof SslHandshakeCompletionEvent) {
+                // Event fired once the SSL handshake is complete, which may have succeeded or not.
+                // If SSL handshake failed, activate flag to exponentially backoff reconnection time.
+                SslHandshakeCompletionEvent sslEvt = (SslHandshakeCompletionEvent) evt;
+                if (sslEvt.cause() instanceof SSLHandshakeException) {
```
Maithem: I don't think this fix should be specific to ssl handshake failures. We should should solve the general problem (i.e. clients spinning on errors). Any exception can cause the channel to be closed, for example:
1. client/server handshake 
2. authentication failures
3. encode/decode failures
4. other errors

In all of those cases, we should have an exponential backoff.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
annym: Done. It also prints the port info but no harm in this.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
annym: This reset happens in between authentication failures and successes to reset the timer calculation from the beginning and exponentially backoff from the start. Otherwise, in an scenario failure, success, failure, this last would never restart it would continue from the rate of the last backoff.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
apavan: Good points! Is your suggestion to handle all these as part of this PR or create issues that could be added to our backlog.

annym: From the logs I could observe these exceptions are wrapped by the SSLHandshakeException.

Maithem: I don't think these cases are handled differently. We just need to apply the exponential backoff to the reconnect logic.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -128,6 +136,14 @@
      */
     public volatile boolean shutdown;
 
+    /**
+     * Indicates if SSL authentication has failed or not, in order to apply a backoff mechanism for
+     * exponential time reconnection.
+     */
+    private AtomicBoolean sslAuthenticationFailure;
```
StrikingThirteen: isSslAuthenticationFailed is more expressive (and then optionally remove the variable comment)

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -606,6 +636,16 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                 ctx.close();
             } else if (e.state() == IdleState.WRITER_IDLE) {
                 keepAlive();
+            } else if (evt instanceof SslHandshakeCompletionEvent) {
+                // Event fired once the SSL handshake is complete, which may have succeeded or not.
+                // If SSL handshake failed, activate flag to exponentially backoff reconnection time.
+                SslHandshakeCompletionEvent sslEvt = (SslHandshakeCompletionEvent) evt;
+                if (sslEvt.cause() instanceof SSLHandshakeException) {
```
StrikingThirteen: There are several levels of nested if else in the code. Can you refactor?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
StrikingThirteen: > From the logs I could observe these exceptions are wrapped by the SSLHandshakeException.

Why endcode/decode failures are wrapped by SshHandshakeException?

runtime/src/main/java/org/corfudb/util/retry/ExponentialBackoffRetry.java
```diff
@@ -83,4 +88,8 @@ public void nextWait() {
         Sleep.MILLISECONDS.sleepUninterruptibly(sleepTime);
     }
 
+    public void reset() {
```
StrikingThirteen: Why not reseting the retryCounter in the reset method?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -363,7 +388,12 @@ private void channelConnectionFutureHandler(@Nonnull ChannelFuture future,
             // a sleep period.
             if (!shutdown) {
                 log.info("connectAsync[{}]: Channel connection failed, reconnecting...", node);
-                Sleep.sleepUninterruptibly(parameters.getConnectionRetryRate());
+                if (sslAuthenticationFailure.get()) {
+                    retryPolicy.nextWait();
+                } else {
+                    retryPolicy.reset();
```
StrikingThirteen: Consider turning ExponentialBackoffRetry to an immutable class and assigning a new instance rather than reseting.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
Maithem: When an exception happens during encode/decode we close the channel and disconnect, then subsequent requests will try to open a new channel thus triggering the reconnect logic.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
annym: Done

runtime/src/main/java/org/corfudb/util/retry/ExponentialBackoffRetry.java
annym: Good catch. Done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -312,6 +313,9 @@
         @Default
         int systemDownHandlerTriggerLimit = 20;
 
+        @Default
+        Duration maxRetryThreshold = Duration.ofMinutes(5);
```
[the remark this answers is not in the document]
annym: This is already in the CorfuRuntimeParameters, same as RPC timeout.

annym:
@Maithem I have addressed all comments, please review. Exponential backoff for reconnections is now performed regardless of the cause that triggered the reconnection and resetting the retry connection time calculator is only performed on connection success.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
Maithem: On re-connection the client will bind to a new port, that's how I was able to tell that the client was spinning on reconnecting (the ports kept changing on every request)

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -186,7 +187,7 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx,
         java.lang.Throwable cause) throws Exception {
-        log.error("exceptionCaught: Exception {} caught.", cause.getClass().getSimpleName(), cause);
+        log.error("exceptionCaught: Exception {} caught. Cause: {}", cause.getClass().getSimpleName(), cause);
```
Maithem: Does this make https://github.com/CorfuDB/CorfuDB/pull/1572/files#diff-74369a0020e71f6d710749351f0ba98eR475 redundant ?

annym: Not really right? the other code is on server logs.

runtime/src/main/java/org/corfudb/util/retry/ExponentialBackoffRetry.java
```diff
@@ -60,8 +63,13 @@ public ExponentialBackoffRetry(IRetryable runFunction) {
...
     }
 
+    public ExponentialBackoffRetry(IRetryable runFunction, Duration maxRetryThreshold) {
+        super(runFunction);
+        this.maxRetryThreshold = maxRetryThreshold;
+    }
+
     @Override
-    public void nextWait() {
+    public long nextWait() {
```
zalokhan: Can we return Duration?

Issue #1575, opened 2018-11-29
CorfuRuntime::shutdown Never Returns


Maithem:
**Bug report.** 
The shutdown method intermittently freezes and never returns.

PR #1577, opened 2018-11-30
Shutdown EventLoop Once

Maithem:
Description:
Only shutdown the event loop once and don't wait for the shut
down future to complete.

Why should this be merged: the shutdown method is called from both the application side and our servers, if this method never return it can block the server from cleanly shutting down restart (i.e partial shutdown)

Related issue(s) (if applicable): #1575

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
@@ -118,11 +118,6 @@ public class NettyClientRouter extends SimpleChannelInboundHandler<CorfuMsg>
      */
     private volatile Channel channel = null;
 
-    /**
-     * The {@link EventLoopGroup} for this router which services requests.
-     */
-    public final EventLoopGroup eventLoopGroup;
-
     /** Whether to shutdown the {@code eventLoopGroup} or not. Only applies when
      *  a deprecated constructor (which generates its own {@link EventLoopGroup} is used.
      */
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -386,23 +380,6 @@ public void stop() {
         log.debug("stop: Shutting down router for {}", node);
         shutdown = true;
         connectionFuture.completeExceptionally(new ShutdownException());
-        try {
-            channel.disconnect();
-            channel.close().syncUninterruptibly();
```
WenbinZhu: Can we keep these two channel closing lines but modify the second one:
```
channel.disconnect();
channel.close();
```
I think we should still request the channel to close, but just don't wait on it.

Maithem: Closing the event loops will call all child channels. 
Look here:
https://stackoverflow.com/questions/37902844/netty-serversocket-and-children-graceful-shutdown

Maithem:
@WenbinZhu are you convinced ?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
WenbinZhu: Sorry I may not get your point, since we remove the codes to shutdown the event loops, how can they call all child channels?

Maithem: shutting down the event loop will shut down the channels.
https://netty.io/wiki/user-guide-for-4.x.html#shutting-down-your-application

Maithem: no its still there:
https://github.com/CorfuDB/CorfuDB/pull/1577/files#diff-ec92d60e991399ec1c89e259082b7965R693

WenbinZhu: I see, thanks for clarification.

WenbinZhu:
> @WenbinZhu are you convinced ?

Yes, thanks

Issue #1573, opened 2018-11-29
TRANSACTION_STREAM leaking Memory


Maithem:
**Bug report.**
Since the ViewsGarbageCollector only runs on opened objects it doesn't trim he transaction stream. We keep track of opened streams(like SMRObjects) and run gc on them too.

PR #1578, opened 2018-11-30
Raw Streams Garbage Collection

Maithem:
## Overview
Made all raw streams (i.e. streams that are not consumed by the VLO)
thread-safe and extended the garbage collection process to include
the CorfuRuntime::StreamView.

Why should this be merged: Fixes memory leaks created by streams that don't belong to SMRObjects (i.e. transaction stream and any stream created by the CorfuRuntime consumer directly). 

Related issue(s) (if applicable): #1573

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
@@ -471,6 +471,7 @@ public class FastObjectLoader {
     private void cleanUpForRetry() {
         runtime.getAddressSpaceView().invalidateClientCache();
         runtime.getObjectsView().getObjectCache().clear();
+        runtime.getStreamsView().getStreamCache().clear();
         nextRead = logHead;
         resetAddressProcessed();
 
--- a/runtime/src/main/java/org/corfudb/runtime/ViewsGarbageCollector.java
+++ b/runtime/src/main/java/org/corfudb/runtime/ViewsGarbageCollector.java
@@ -75,6 +75,7 @@ public class ViewsGarbageCollector {
                     currTrimMark);
             long startTs = System.currentTimeMillis();
             runtime.getObjectsView().gc(currTrimMark);
+            runtime.getStreamsView().gc(currTrimMark);
             runtime.getAddressSpaceView().gc(currTrimMark);
             long endTs = System.currentTimeMillis();
             trimMark = currTrimMark;
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -144,8 +144,9 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
         this.args = args;
         this.serializer = serializer;
 
+        // Since the VLO is thread safe we don't need to use a thread safe stream implementation
         underlyingObject = new VersionLockedObject<T>(this::getNewInstance,
-                new StreamViewSMRAdapter(rt, rt.getStreamsView().get(streamID)),
+                new StreamViewSMRAdapter(rt, rt.getStreamsView().getUnsafe(streamID)),
                 upcallTargetMap, undoRecordTargetMap,
                 undoTargetMap, resetSet);
 
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -15,6 +15,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nonnull;
@@ -272,11 +273,24 @@ public class AddressSpaceView extends AbstractView {
                             .flatMap(stripe -> stripe.getLogServers().stream())
                             .map(e::getLogUnitClient)
                             .map(LogUnitClient::getTrimMark)
-                            .map(CFUtils::getUninterruptibly)
+                            .map(future -> {
+                                // This doesn't look nice, but its required to trigger
+                                // the retry mechanism in AbstractView. Also, getUninterruptibly
+                                // can't be used here because it throws a UnrecoverableCorfuInterruptedError
+                                try {
+                                    return future.join();
+                                } catch (CompletionException ex) {
+                                    Throwable cause = ex.getCause();
+                                    if (cause instanceof RuntimeException) {
+                                        throw (RuntimeException) cause;
+                                    } else {
+                                        throw new RuntimeException(cause);
+                                    }
+                                }
+                            })
                             .max(Comparator.naturalOrder()).get();
                     return new Token(e.getLayout().getEpoch(), trimMark);
                 });
-
     }
 
     /**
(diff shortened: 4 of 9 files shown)


runtime/src/main/java/org/corfudb/runtime/ViewsGarbageCollector.java
```diff
@@ -75,6 +75,7 @@ public void runRuntimeGC() {
                     currTrimMark);
```
Maithem: getTrimMark is not a safe-call remove Unrecoverable exceptions from its call path.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -471,6 +471,7 @@ private void initializeHeadAndTails() {
     private void cleanUpForRetry() {
         runtime.getAddressSpaceView().invalidateClientCache();
         runtime.getObjectsView().getObjectCache().clear();
+        runtime.getStreamsView().getStreamCache().clear();
```
Maithem: Why is this needed? the fast loader only creates unsafe streams.

runtime/src/main/java/org/corfudb/runtime/ViewsGarbageCollector.java
Maithem: Done.

test/src/test/java/org/corfudb/runtime/view/stream/BackpointerStreamViewTest.java
```diff
@@ -158,7 +158,7 @@ public void headOfStreamBackpointerTermination() {
 
         // Fetch Stream B and verify backpointer count (which requires 1 read = 1 entry)
         svB.remainingUpTo(totalEntries);
-        assertThat(((BackpointerStreamView) svB).getBackpointerCount()).isEqualTo(1L);
+        assertThat(((ThreadSafeStreamView) svB).getUnderlyingStream().getBackpointerCount()).isEqualTo(1L);
```
Maithem: Add an explicit test.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -272,11 +273,10 @@ public Token getTrimMark() {
                             .flatMap(stripe -> stripe.getLogServers().stream())
                             .map(e::getLogUnitClient)
                             .map(LogUnitClient::getTrimMark)
-                            .map(CFUtils::getUninterruptibly)
+                            .map(CompletableFuture::join)
```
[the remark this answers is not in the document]
Maithem: unrecoverable exceptions are escaping the try/catch logic and causing some threads to unexpectedly disappear. Don't want to change the getUninterruptibly logic in this PR.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -147,7 +148,7 @@ private Layout getLayoutUninterruptibly() {
                     WrongEpochException we = (WrongEpochException) re;
                     log.warn("Got a wrong epoch exception, updating epoch to {} and "
                             + "invalidate view", we.getCorrectEpoch());
-                } else if (re instanceof NetworkException) {
+                } else if (re instanceof NetworkException || re.getCause() instanceof NetworkException) {
```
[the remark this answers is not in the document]
Maithem: Its wrapped by CompletionException, I think the java streaming logic wraps it, I'll take a look.
Also, need to retry on TimeoutException.

test/src/test/java/org/corfudb/runtime/view/stream/BackpointerStreamViewTest.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -50,15 +58,37 @@ public IStreamView get(UUID stream) {
         return this.get(stream, StreamOptions.DEFAULT);
     }
 
+    public IStreamView getUnsafe(UUID stream) {
```
annym: Can we add a comment on how getUnsafe differentiates from get.. maybe implications/risks of using one versus the other or why we use this for streams consumed by the VLO.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -145,7 +145,7 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
         this.serializer = serializer;
 
         underlyingObject = new VersionLockedObject<T>(this::getNewInstance,
-                new StreamViewSMRAdapter(rt, rt.getStreamsView().get(streamID)),
+                new StreamViewSMRAdapter(rt, rt.getStreamsView().getUnsafe(streamID)),
```
annym: thinking, we could add a comment on why we do not need a thread-safe call on the streams view here...

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -1,6 +1,7 @@
 package org.corfudb.runtime.view;
 
 import java.time.Duration;
+import java.util.concurrent.CompletionException;
```
annym: is this needed?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -15,6 +15,9 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
```
annym: is ExecutionException needed?

runtime/src/main/java/org/corfudb/runtime/view/stream/ThreadSafeStreamView.java
```diff
@@ -0,0 +1,106 @@
...
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.view.StreamOptions;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.UUID;
+import java.util.function.Function;
+
+/**
+ * A thread-safe implementation of IStreamView.
```
annym: Can we add a comment on which methods are made thread-safe with regards to BackpointerStreamView, it would ease comparison of both.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: removed.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
Maithem: done

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/view/stream/ThreadSafeStreamView.java
Maithem: BackpointerStreamView is hidden, the IStreamView interface is thread-safe (i.e. all methods).

Maithem:
@annym I see that you have re-enabled the cron (after) i have disabled it. Is it possible to prioritize this?

zalokhan:
@Maithem triggered injection framework.

Maithem:
Thanks for pushing this through guys.

PR #1599, opened 2018-12-07
Cleanup seq cache before setting maxConflictWildcard.

zalokhan:
## Overview

Description: Cleanup sequencer cache before we set the new bootstrapped values on the sequencer.

Why should this be merged: The invalidate cache is async and is not same as cache eviction. The Eviction should take place before the value of maxConflictWildcard is set.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -421,10 +421,12 @@ public class SequencerServer extends AbstractServer {
         // Note, this is correct, but conservative (may lead to false abort).
         // It is necessary because we reset the sequencer.
         if (!bootstrapWithoutTailsUpdate) {
+            // Evict all entries from the cache. This eviction triggers the callback modifying the maxConflictWildcard.
+            conflictToGlobalTailCache.cleanUp();
+
             globalLogTail.set(initialToken);
             maxConflictWildcard = initialToken - 1;
             maxConflictNewSequencer = maxConflictWildcard;
-            conflictToGlobalTailCache.invalidateAll();
 
             // Clear the existing map as it could have been populated by an earlier reset.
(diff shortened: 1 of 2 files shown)


test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1761,4 +1769,75 @@ public void testSealedDegradedClusterHealing() {
...
+        corfuRuntime.invalidateLayout();
+        Layout layout = new Layout(corfuRuntime.getLayoutView().getLayout());
+        layout.setEpoch(layout.getEpoch() + 1);
+        corfuRuntime.getLayoutView().getRuntimeLayout(layout).sealMinServerSet();
+        corfuRuntime.getLayoutView().updateLayout(layout, 1L);
+        return layout;
+    }
+
+    /**
+     * Test scenario where the sequencer bootstrap triggers cache cleanup causing maxConflictWildcard to be reset.
```
StrikingThirteen: Can you add a bit more description of steps to the test comments.

StrikingThirteen: Actually we don't seem to directly verify on reset maxConflicWildcard but instead making the conclusion from lack of exception and being able to do data operation.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -421,10 +421,10 @@ public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg>
         // Note, this is correct, but conservative (may lead to false abort).
         // It is necessary because we reset the sequencer.
         if (!bootstrapWithoutTailsUpdate) {
+            conflictToGlobalTailCache.cleanUp();
```
StrikingThirteen: Can you add a sentence to comment regarding this eviction?

PR #1604, opened 2018-12-07
Change existing IT test cases to fit the new fault detecor

WenbinZhu:
## Overview

Description:

Choose the node with the largest endpoint to stop/pause and
change the way of assertion to prevent jittering.

- [x] Public API has Javadoc


it/src/test/java/org/corfudb/universe/scenario/OneNodeDownIT.java
```diff
@@ -44,15 +45,14 @@ public void oneNodeDownTest() {
             }
 
             testCase.it("Should stop one node and then restart", data -> {
-                CorfuServer server0 = corfuCluster.getFirstServer();
+                CorfuServer server2 = corfuCluster.getServerByIndex(2);
```
StrikingThirteen: Why changing the servers?

it/src/test/java/org/corfudb/universe/scenario/OneNodeDownIT.java
```diff
@@ -44,15 +45,14 @@ public void oneNodeDownTest() {
...
 
                 // Stop one node and wait for layout's unresponsive servers to change
-                server0.stop(Duration.ofSeconds(10));
-                waitForUnresponsiveServersChange(size -> size == 1, corfuClient);
-
-                // Verify layout, unresponsive servers should contain only the stopped node
-                Layout layout = corfuClient.getLayout();
-                assertThat(layout.getUnresponsiveServers()).containsExactly(server0.getEndpoint());
+                // We choose the node with largest endpoint to prevent jittering
+                // TODO: rotate the node to stop to fully test jittering and deadlock cases
```
StrikingThirteen: Maybe remove this TODO comment if the change is not coming very soon.

it/src/test/java/org/corfudb/universe/scenario/OneNodeDownIT.java
```diff
@@ -44,15 +45,14 @@ public void oneNodeDownTest() {
...
+                server2.stop(Duration.ofSeconds(10));
+                waitForLayoutChange(layout -> layout.getUnresponsiveServers()
```
StrikingThirteen: It's a refactoring but what was the problem that it is trying to fix?

it/src/test/java/org/corfudb/universe/scenario/OneNodePausedIT.java
```diff
@@ -44,29 +45,28 @@ public void oneNodePausedTest() {
             }
 
             testCase.it("Should pause one node and then resume", data -> {
-                CorfuServer server1 = corfuCluster.getServerByIndex(1);
+                CorfuServer server2 = corfuCluster.getServerByIndex(2);
```
StrikingThirteen: The test should work for the original servers. Why changing the servers?

it/src/test/java/org/corfudb/universe/scenario/OneNodeDownIT.java
WenbinZhu: Because servers are already sorted by their endpoints. If we take down server0, with the new algorithm, server2 could first become unresponsive because of a timing issue (server1 has a stale heartbeat before server0 down, thus sending this view to server2, so server2 think the link between server0 and server1 is health, but server2 itself has a new heartbeat and knows about server0 down. So server2 make decision locally and mark itself unresponsive since it has larger endpoint).

Using server2 here make test stable because we  make sure the node we taken down is the node becomes unresponsive later.

WenbinZhu: However once we introduce the stabilizer, we can safely take down any node, that's why I put a TODO below.

it/src/test/java/org/corfudb/universe/scenario/OneNodeDownIT.java
WenbinZhu: See comments above.

it/src/test/java/org/corfudb/universe/scenario/OneNodeDownIT.java
WenbinZhu: This is a better way to do assert, to eliminate any possibility of jittering.

it/src/test/java/org/corfudb/universe/scenario/OneNodePausedIT.java
WenbinZhu: See comments above.

Maithem:
@WenbinZhu is this still valid, if not please close the PR.

PR #1623, opened 2018-12-17
Add Management Server Configuration

StrikingThirteen:
## Overview

Description:
Add Management Server Configuration

Why should this be merged: 
Avoid directly dealing with opts.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -95,8 +92,10 @@
      */
     public ManagementServer(ServerContext serverContext) {
 
-        this.opts = serverContext.getServerConfig();
-        this.localEndpoint = this.opts.get("--address") + ":" + this.opts.get("<port>");
+        this.managementServerConfig = ManagementServerConfig.parse(serverContext.getServerConfig());
+
+        this.localEndpoint = managementServerConfig.getAddress() + ":" +
```
xnull: It's much better to use NodeLocator instead of host+port, would be great if you replace host+port to NodeLocator here and in ManagementServerConfig

StrikingThirteen: At this point I don't want to change the fields of ManagementServer so I'll keep the endpoint as is. The ManagementServerConfiguration however is changed to be based on NodeLocator.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -95,8 +93,9 @@
      */
     public ManagementServer(ServerContext serverContext) {
 
-        this.opts = serverContext.getServerConfig();
-        this.localEndpoint = this.opts.get("--address") + ":" + this.opts.get("<port>");
+        this.managementServerConfig = ManagementServerConfig.parse(serverContext.getServerConfig());
+
+        this.localEndpoint = NodeLocator.getLegacyEndpoint(managementServerConfig.getNodeLocator());
```
xnull: Looks like localEndpoint is not used at all, if I am right please get rid of it

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -325,4 +324,22 @@ public void shutdown() {
...
+    @Getter
+    public static class ManagementServerConfig {
+        private final NodeLocator nodeLocator;
+
+        public static ManagementServerConfig parse(Map<String, Object> opts) {
+            final String address = (String) opts.get("--address");
+            final Integer port = (Integer) opts.get("<port>");
+
+            return ManagementServerConfig.builder()
+                    .nodeLocator(NodeLocator.parseString(address + ":" + port))
```
xnull: Please check if this entire class actually used by ManagementServer, looks like localEndpoint variable not used by any one.

If it's not used then just delete this class and localEndpoint parameter.
It it's used by some one then you could use `NodeLocator.builder().host().port().build()` instead of parseString

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -113,15 +114,16 @@ public ExecutorService getExecutor() {
      */
     public ManagementServer(ServerContext serverContext) {
 
-        this.opts = serverContext.getServerConfig();
-        this.localEndpoint = this.opts.get("--address") + ":" + this.opts.get("<port>");
+        managementServerConfig = ManagementServerConfig.parse(serverContext.getServerConfig());
```
vjeko: Please keep the style consistent. It is generally a good practice to prefix member variables in a constructor with "this".

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -113,15 +114,16 @@ public ExecutorService getExecutor() {
...
+        localEndpoint = managementServerConfig.getAddress() + ":" +
+                             managementServerConfig.getPort();
         this.serverContext = serverContext;
 
         executor = Executors.newFixedThreadPool(serverContext.getManagementServerThreadCount(),
                 new ServerThreadFactory("management-", new ServerThreadFactory.ExceptionHandler()));
 
-        this.failureHandlerPolicy = serverContext.getFailureHandlerPolicy();
-        this.healingPolicy = serverContext.getHealingHandlerPolicy();
+        failureHandlerPolicy = serverContext.getFailureHandlerPolicy();
```
vjeko: Why was this changed?

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -113,15 +114,16 @@ public ExecutorService getExecutor() {
      */
     public ManagementServer(ServerContext serverContext) {
 
-        this.opts = serverContext.getServerConfig();
-        this.localEndpoint = this.opts.get("--address") + ":" + this.opts.get("<port>");
+        managementServerConfig = ManagementServerConfig.parse(serverContext.getServerConfig());
+
```
zalokhan: Server context already has a method to getLocalEndpoint. You can use that directly.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: done

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: moved all to sever context.
done.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: I'm using "this" just when the variable in local scope shadows the fields. It seems using this adds some clutter without much value and so limiting it usage to when scope mandates its usage. Let me know if you feel strongly and will add them back.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: healingPolicy wasn't being used.
if you're referring to "this", I removed unnecessary keyword as per previous comments. similarly, if you feel strongly about the inclusion of this keyword, I'll bring them that. That being said there are other constructors in the code base where "this" only used for dealing with shadowed variables. let me know.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: yes, I think it is much cleaner to move the responsibility to ServerContext. Moved all of them there.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
xnull: I join to Vjeco and also vote for using `this` in a constructor,  just make it consistent and lined up the code. For any other methods don't use `this`

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: done

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: done

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
StrikingThirteen: done

test/src/test/java/org/corfudb/infrastructure/ReconfigurationEventHandlerTest.java
```diff
@@ -58,7 +58,7 @@ public void updateLayoutOnFailure() {
         failedServers.add(getEndpoint(SERVERS.PORT_2));
```
zalokhan: Would it be possible to remove SERVERS.PORT_0 instead of 2.
This way we can test 2 more items in this test with the given policy:
1. Sequencer rollover.
2. LogUnit removal.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -139,7 +142,6 @@ public ServerContext(Map<String, Object> serverConfig) {
         this.dataStore = new DataStore(serverConfig, this::dataStoreFileCleanup);
         generateNodeId();
         this.failureHandlerPolicy = new ConservativeFailureHandlerPolicy();
```
zalokhan: Do we want to remove this from here and move it to the ManagementServer as we discussed?

PR #1631, opened 2018-12-19
Adding more logs for failure detection

StrikingThirteen:
## Overview

Description:
Adding more log statements to the recommendation enginer

Why should this be merged: 
For better debugging.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/management/FullyConnectedClusterRecommendationEngine.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/management/FullyConnectedClusterRecommendationEngine.java
@@ -1,5 +1,7 @@
 package org.corfudb.infrastructure.management;
 
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.ClusterState;
 import org.corfudb.protocols.wireprotocol.NodeState;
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/management/FullyConnectedClusterRecommendationEngine.java
```diff
@@ -176,6 +184,9 @@ public ClusterRecommendationStrategy getClusterRecommendationStrategy() {
                                     .add(currentNode));
         }
 
+        log.debug("Converted view of the cluster to a symmetric view of connectivity failures: {}",
```
medhavidhawan: another way would be to log.info whenever there is any output. On that trigger you can log both input and output. Hope is that would be seldom enough that you can have it as info.

StrikingThirteen: @medhavidhawan If we just log when healed/failed nodes are being proposed, it can be INFO. However my thought was to provide more logs at DEBUG level even the alg doesn't output any healed or failed. Do you think it's ok to keep the DEBUG logs and add two INFO level logs for when outputs are not empty?

StrikingThirteen: As we discussed offline, added the info level logs for when here are non empty proposed healed or failed nodes.

Maithem:
I thought we are not pushing any changes until CI is fixed.

PR #1632, opened 2018-12-19
Remove Parse Trimming Logic

Maithem:
## Overview
We currently only support prefix trimming, removed all the logic
related to sparse trimming, which will also prevent the related
metadata files from being created.

Why should this be merged: Prevents unnecessary metadata from being generated. 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -223,14 +223,6 @@ public class LogUnitServer extends AbstractServer {
         }
     }
 
-    @ServerHandler(type = CorfuMsgType.TRIM)
-    private void trim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        TrimRequest req = msg.getPayload();
-        batchWriter.trim(req.getAddress().getSequence(), req.getAddress().getEpoch());
-        //TODO(Maithem): should we return an error if the write fails
-        r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
-    }
-
     @ServerHandler(type = CorfuMsgType.PREFIX_TRIM)
     private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx,
                             IServerRouter r) {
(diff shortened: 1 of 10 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -223,14 +223,6 @@ private void fillHole(CorfuPayloadMsg<FillHoleRequest> msg, ChannelHandlerContex
...
-    @ServerHandler(type = CorfuMsgType.TRIM)
-    private void trim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        TrimRequest req = msg.getPayload();
-        batchWriter.trim(req.getAddress().getSequence(), req.getAddress().getEpoch());
-        //TODO(Maithem): should we return an error if the write fails
-        r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
-    }
-
     @ServerHandler(type = CorfuMsgType.PREFIX_TRIM)
     private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx,
```
annym: Shouldn't we remove the trim logic from the LogUnitClient as well?
And also clean up, by removing the trim CorfuMsgType.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
```diff
@@ -543,7 +453,7 @@ public void testPrefixTrim() {
 
         // Write 50 segments and trim the first 25
         final long numSegments = 50;
-        final long filesPerSegment = 3;
+        final long filesPerSegment = 1;
```
annym: why did you change this?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
Maithem: Currently on master, we have 3 files per segment. For example, for segment 1, we have 1.log          1.log.pending  and 1.log.trimmed. This test used to verify that the 3 files were being generated. Now that we removed those 2 files, the assertion would fail.

Issue #1598, opened 2018-12-07
CI Failing


Maithem:
**Bug report.**
This test is failing often please try to fix it:
[ERROR] Errors: 
[ERROR]   NodeDownAndLinkFailureIT.nodeDownAndLinkFailureTest:39->lambda$nodeDownAndLinkFailureTest$4:49->lambda$null$3:69 » TestTimedOut
[INFO]

WenbinZhu:
Test client log:
[NodeDownAndLinkFailureTest.log](https://github.com/CorfuDB/CorfuDB/files/2659242/NodeDownAndLinkFailureTest.log)

Server0 log:
[wlehey-corfu-node64449.log](https://github.com/CorfuDB/CorfuDB/files/2659243/wlehey-corfu-node64449.log)

Server1 log:
[wlehey-corfu-node64450.log](https://github.com/CorfuDB/CorfuDB/files/2659244/wlehey-corfu-node64450.log)

Server2 log:
[wlehey-corfu-node64451.log](https://github.com/CorfuDB/CorfuDB/files/2659246/wlehey-corfu-node64451.log)


Test branch name: triageFailedTests

Maithem:
It seems like ever since the fault detector related code went in, the master branch has been really unstable.

More tests failing:
https://travis-ci.org/CorfuDB/CorfuDB/jobs/467265276
https://travis-ci.org/CorfuDB/CorfuDB/jobs/467265277

Can you guys resolve this so that we don't have to revert the recent fault detector changes.

PR #1635, opened 2018-12-19
nodeUpAndPartitionedTest Keeps Failing

Maithem:
## Overview
Revert "Add Unresponisve Node Up AND Responsive Partitioned Test (#1593)"

This reverts commit 8b4cbb14c174d13884c3b0bcc068d352f195ba75.

Why should this be merged: The test is timing out and failing CI at a very high rate (~40% observed) 

Related issue(s) (if applicable): #1598

- [ ] Public API has Javadoc


StrikingThirteen:
@Maithem I don't think it should be reverted as the problem is not being introduced by this test. As you can see on the CI of this patch, the ITs still failing on the next long running integration test: NodeDownAndLinkFailureIT. Maybe ignore the test till the root cause is resolved?

Maithem:
@StrikingThirteen I get that there's a real bug, but I'm trying to revert everything related to this.

PR #1637, opened 2018-12-20
Avoid unnecessary layout updates

StrikingThirteen:
## Overview

Description:
Avoids updating layout when unnecessary. There is a race condition and it happens when a node detect failedNodes (or healedNode) but by the time that ManagementServer receives the MANAGMENT_FAILING_DETECTED (or MANAGEMENT_HEALING_DETECTED) the appropriate action (ie. add to responsive / remove from unresponsive) is already taken. So currently management server just bumps up the layout with a new epoch without making any useful change to the layout. 

Why should this be merged: 
To avoid updating layout where responsive and unresponsive won't effectively changed as the result of detected failures and healings.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
@@ -1,31 +1,29 @@
 package org.corfudb.infrastructure;
 
 import io.netty.channel.ChannelHandlerContext;
-
-import java.lang.invoke.MethodHandles;
-import java.time.Duration;
-import java.util.Map;
-
-import javax.annotation.Nonnull;
-
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.infrastructure.management.ClusterStateContext;
 import org.corfudb.infrastructure.management.ReconfigurationEventHandler;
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -216,11 +214,29 @@ public void handleFailureDetectedMsg(CorfuPayloadMsg<DetectorMsg> msg,
...
+        // layout
+        final Set<String> allActiveServers = layout.getAllActiveServers();
+        final Set<String> responsiveFailedNodes = detectorMsg.getFailedNodes()
+                .stream()
+                .filter(allActiveServers::contains)
+                .collect(Collectors.toSet());
+
+        // If there is no need to update the layout return without any reconfiguration
+        if (responsiveFailedNodes.isEmpty()) {
+            log.warn("handleFailureDetectedMsg: No action is taken as none of the failed nodes " +
```
xnull: Please use JsonUtils

Maithem:
Can you update the description, its not clean when this can happen and how this patch addresses the issue. I'm trying to reproduce the issue.

Maithem:
So we don;t know how the problem happens, but somehow this fixes the scale issues?

I'm trying to write a test case based on the description.

PR #1638, opened 2018-12-21
Prevent jitter by increasing sleep time for failure detection

StrikingThirteen:
## Overview

Description:
Avoiding quick iterations in failure detection to prevent jittery decision.

Why should this be merged: 
Prevents a succession of quick iteration lead to immature failure detection and consequent jitter in the layout.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -153,6 +149,9 @@ private PollReport pollRound(long epoch,
         // In each iteration we poll all the servers in the members list.
         for (int iteration = 0; iteration < failureThreshold; iteration++) {
 
+            final long iterationStartTime = System.nanoTime();
+            long interIterationSleepTime = interIterationInterval;
+
```
Maithem: make interIterationInterval final and make sure its >1ms, otherwise you can have the opposite behavior of what this patch is trying to do. In linux, I don't think you can put a thread to sleep for less than 1ms, so if it gets downcast to 0, the thread won't sleep. If you need sub-millisecond precision you can use busy-waiting, but the sleep utils used here to even use that.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -174,13 +173,22 @@ private PollReport pollRound(long epoch,
                 // Timeout is increased anytime a failure is encountered.
                 // Failure includes both, unresponsive and outOfPhaseEpoch nodes.
                 if (failuresDetected) {
+                    final long iterationElapsedTime = TimeUnit.MILLISECONDS.convert(
+                            System.nanoTime() - iterationStartTime, TimeUnit.NANOSECONDS);
+
+                    // Increase the inter iteration sleep time if the period time increases
+                    interIterationSleepTime = Math.max(interIterationInterval,
+                                                       period - iterationElapsedTime);
```
Maithem: If you can only sleep in milliseconds and you ceil to 1ms, whats the point of tracking nanoseconds and convertin ns->ms ?

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -174,13 +173,22 @@ private PollReport pollRound(long epoch,
...
                     tuneRoutersResponseTimeout(members, routerMap, new HashSet<>(members), period);
                 }
                 final int pollIteration = iteration;
                 responses.forEach(s -> responsesMap.put(s, pollIteration));
             }
-            Sleep.MILLISECONDS.sleepUninterruptibly(interIterationInterval);
+            log.debug("Time taken for iteration's failure detection: {}",
+                    TimeUnit.MILLISECONDS.convert(System.nanoTime() - iterationStartTime,
+                            TimeUnit.NANOSECONDS));
+            Sleep.MILLISECONDS.sleepUninterruptibly(interIterationSleepTime);
```
Maithem: Why sleep again here, if we tune the timeouts on every iteration for pollOnceAsync, shouldn't the timeouts on those requests also reflect interIterationSleepTime?

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
StrikingThirteen: interIterationInterval has setter/getters and currently being used. Its default is set at 1ms and the patch keeping max of the default and potentially increased time. so it won't go bellow 1ms except if it is specifically set through the setter.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
StrikingThirteen: nanoTime provides relative time rather than wall clock and is a better choice for micro benchmarking. its better accuracy is the secondary benefit.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
StrikingThirteen: The sleep at the end of each iteration is for poll iteration intervals when a failure observed and separate from response timeout. changed variable names and added comments to better describe it in the code.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -153,6 +149,9 @@ private PollReport pollRound(long epoch,
         // In each iteration we poll all the servers in the members list.
         for (int iteration = 0; iteration < failureThreshold; iteration++) {
 
+            final long pollIterationStartTime = System.nanoTime();
```
Maithem: I think my comment got lost from the last review. But why is this nanoTime?

StrikingThirteen: I think I replied to the initial comments. It's because of using relative time elapsed rather than time passed from epoch. all times are converted/used in millisecond though. Also more accurate.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -76,15 +72,15 @@
     private long period = initPeriodDuration;
 
     /**
-     * Interval between iterations in a pollRound.
+     * Poll interval between iterations in a pollRound in milliseconds.
      */
     @Getter
     @Setter
-    private long interIterationInterval = 1_000;
+    private long initialPollInterval = 1_000;
```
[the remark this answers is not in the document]
StrikingThirteen: Name is changed to be more expressive but not refactoring the code in this patch.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -174,13 +173,25 @@ private PollReport pollRound(long epoch,
                 // Timeout is increased anytime a failure is encountered.
                 // Failure includes both, unresponsive and outOfPhaseEpoch nodes.
                 if (failuresDetected) {
+                    final long iterationElapsedTime = TimeUnit.MILLISECONDS.convert(
```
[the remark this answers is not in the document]
StrikingThirteen: it's calculating time elapsed at different points.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
Maithem: Ok, I remember now. But why is it necessary to have that precision (inter-millisecond), if you can't even use it? Converting back and forth just doesn't seem necessary.

xnull: I agree with Maithem, we don't need nano time for network communication, it makes things more complicated

StrikingThirteen: @Maithem as mention on the first reply as well as above: 

> it's because of using relative time elapsed rather than time passed from epoch.

If you think that there is no difference between using nanoTime and currentTimeMillis in one being timer and the other wall clock, let me know.

(better accuracy of nanoTime even at the millisecond granularity is a side point and hence "also" in reply which was also mentioned as a side benefit but NOT the rationale for picking nanoTime the first reply).

StrikingThirteen: @xnull where in this code are we using nano time for network communication? Accuracy as mentioned several times is NOT the reason for using nanoTime. please read the comments above.

Maithem: I'm not asking you about the difference between System.currentTimeMillis and System.nanoTime, I think I understand that part.

Let me rephrase the question, how would using System.currentTimeMillis negatively affect the algorithm?

If the accuracy is the "side point", then what is the main point. Your argument is not clear. Furthermore, referring to previous points doesn't help either because you force pushed and they are no longer there for reference.

StrikingThirteen: @Maithem Ok, thanks for clarifying the concern. The main point is that using a timer I think is correct choice over using a wall clock. Example of negative effect: wrong benchmark when using currentTimeMillis after a clock sync which won't be the case with nanoTime.

The previous reply was addressing the same question (here it is, in case):
> nanoTime provides relative time rather than wall clock and is a better choice for micro benchmarking.

all in all, if you still prefer currentTimeMillis, I'll change nanoTime to it.

StrikingThirteen: The currentTimeMillis provides a wall clock time and is non monotonic which may lead to a regress as a result of clock adjustment.
nanoTime is a monotonic timer and won't regress.
We had an offline discussion to keep the nanoTime in order to avoid regression.

xnull: > which may lead to a regress
BTW I didn't find anything about the regression in documentation http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/System.java

Maithem: Related: https://bugs.java.com/bugdatabase/view_bug.do;jsessionid=c0af3bf36ff7fffffffffdbd3372166fe34d?bug_id=6311057

StrikingThirteen: @xnull 
from https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#currentTimeMillis--
>....Returns the **current time** in milliseconds. ....

hence a wall clock. As a result different calls to currentTimeMillis can potentially jump backward (or forward). 
on the other hand:
from https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--
> ...can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time.

StrikingThirteen: @xnull what I quoted was copy pasted from java doc (links above).
Are you suggesting that using currentTimeMillis for timing can not lead to negative timers?

PR #1639, opened 2018-12-21
Querying all layout servers to find the latest layout

StrikingThirteen:
## Overview

Description:
Instead of randomly querying the list of layout servers to find the latest layout. All layout servers will be iterated through and the latest layout will be picked.

Why should this be merged: 
To take the latest layout amongst all the provided layout servers.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -879,46 +873,32 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
...
-
-                        return l;
-                    } catch (InterruptedException ie) {
-                        throw new UnrecoverableCorfuInterruptedError(
-                                "Interrupted during layout fetch", ie);
-                    } catch (Exception e) {
-                        log.warn("Tried to get layout from {} but failed with exception:", s, e);
+                try {
+                    // Fetch the latest layout we can possibly fetch from the list of layout servers.
+                    CompletableFuture<Layout> layoutFuture = getLatestLayout(layoutServersCopy);
```
Maithem: Why is the runtime tracking two layouts (i.e. layout and latestLayout), remove latestLayout and just update the this.layout as you discover layouts with higher epochs.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -879,46 +873,32 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
             int systemDownTriggerCounter = 0;
 
             while (true) {
```
Maithem: The behavior changed from "probing" to a broadcast, this while(true) is unnecessary and can spin on exceptions that you want to fail right away. I think we should remove it, especially because getLatestLayout swallows all network exceptions.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -879,46 +873,32 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
...
+                                latestLayout.getEpoch());
+                        continue;
                     }
+                    checkClusterId(l);
+
+                    layout = layoutFuture;
+                    latestLayout = l;
+
+                    // Prune away removed node routers from the nodeRouterPool.
+                    pruneRemovedRouters(l);
```
Maithem: pruneRemovedRouters leaks channels, make sure you close the router's channel.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -939,6 +919,53 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
...
     }
 
+    /**
+     * Helper function to fetch the latest layout from provided list of {@param layoutServers}
+     *
+     * @param layoutServers list layout servers
+     * @return Completed Future with the latest layout or an exception if no layout was found.
+     */
+    private CompletableFuture<Layout> getLatestLayout(List<String> layoutServers) {
+        CompletableFuture<Layout> latestLayoutFuture = null;
```
Maithem: Just make getLatestLayout return a Layout, if its null then return a dummy layout with -1 epoch. Then in fetchLayout you can wrap this with a future and return.

PR #1642, opened 2018-12-21
Ignore intermittent failing test

StrikingThirteen:
## Overview

Description:
The test NodeUpAndPartitioned is failing intermittently but frequently.  Suggesting to ignore until the underlying timeout on travis is addressed. 

Why should this be merged: 
Avoid intermittent failing test slows down unrelated checkins.

- [x] Public API has Javadoc


StrikingThirteen:
I'm suggesting to ignore rather than revert until the underlying issue is addressed as the same issue less frequently causes the failure on other test such as NodesDownAndPartitionedFailure. suggestions?

Maithem:
I was trying to revert the test case + the code path it was testing. Only disabling the test to have a greener CI is not better than just retrying CI until it passes.

StrikingThirteen:
@Maithem sure. I have seen assert failures only when it caught an actual issue but might has bugs itself. Let's keep it and debug the root cause.

PR #1646, opened 2018-12-26
Change Default Backoff Threshold

annym:
## Overview

Decrease max threshold for client disconnects to 500milliseconds.

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/retry/ExponentialBackoffRetry.java
+++ b/runtime/src/main/java/org/corfudb/util/retry/ExponentialBackoffRetry.java
@@ -2,6 +2,7 @@ package org.corfudb.util.retry;
 
 import java.time.Duration;
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/util/retry/ExponentialBackoffRetry.java
```diff
@@ -84,15 +97,15 @@ public Duration nextWait() {
 
         if (System.currentTimeMillis() + sleepTime > nextBackoffTime) {
             nextBackoffTime = 0;
-            retryCounter = 1;
+            retryCounter.set(1);
```
StrikingThirteen: I think it should be set to startBasePower

annym:
@StrikingThirteen Exponential backoff is an available utility used in other parts of the code, which can be tuned to the desired behavior (no extra work). If only two points are wanted (like suggested to avoid numerous reconnection attempts in a small scale) it can be tuned to behave in this way, or we can later reuse it to tune it differently. As explained in other opportunities, the upper bound set in PR #1572 was not based on any criteria, it was randomly picked and we later saw the implications, so comparing them is not the case. I agree we should have a strong case to decide on this value, however, as I indicated before we are bounded by our failure/healing detection frequency. If we attempt to react to failures/healing in the order of a second, our reconnect frequency needs to be lower than this, or otherwise we won't be able to react to failures/healing and we'll start experiencing other kinds of problems like timeouts of many failure/healing related messages queuing up on a channel that has not been connected. If the reconnection upper bound is higher than our failure/healing detection frequency then our ability to react to changes is bounded by this number, what is the best number seen this way? I believe our confidence is as much as ours with all related failure link failing PRs, as these two are pretty closely related and dependent. Any suggestions on how to increase the confidence of intermittent failures?

annym:
We have decided to revert and go for a simpler solution (#1649), add 1 second sleep between reconnection attempts for a established connection failed to succeed (for instance because of certificate validation).

PR #1659, opened 2019-01-04
Get Cluster Status API

annym:
## Overview

Simplify and remove dependency with failure detector.
Introduce DB_SYNCING as a cluster state.

Why should this be merged: this API currently relies on the failure detector to determine the state of the cluster which is rather complex. We should be reporting the cluster state based on the layout.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/ClusterStatusReport.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ClusterStatusReport.java
@@ -32,7 +32,15 @@ public class ClusterStatusReport {
         /**
          * Node is not reachable.
          */
-        DOWN
+        DOWN,
+
+        /**
+         * Whenever the state of a node cannot be determined.
+         * If a client looses connection to all nodes
+         * it is wrong to report them DOWN, nodes might be up and clients are unable
+         * to connect.
+         */
+        NA;
     }
 
     /**
@@ -46,14 +54,20 @@ public class ClusterStatusReport {
         STABLE(0),
 
         /**
+         * The cluster is operational but one or several nodes are syncing in the background.
+         * i.e., (replica catch up)
+         */
+        DB_SYNCING(1),
+
+        /**
          * The cluster is operational but working with reduced redundancy.
          */
-        DEGRADED(1),
+        DEGRADED(2),
 
         /**
          * The cluster is not operational.
          */
-        UNAVAILABLE(2);
+        UNAVAILABLE(3);
 
         @Getter
         final int healthValue;
@@ -64,17 +78,71 @@ public class ClusterStatusReport {
     }
 
     /**
+     * Represents the connectivity status of this runtime to a node.
+     */
+    public enum ConnectivityStatus {
+        /**
+         * The node responds to pings from client.
+         */
+        RESPONSIVE(true),
+
+        /**
+         * The node does not respond to pings from client.
+         */
+        UNRESPONSIVE(false);
+
+        @Getter
+        final boolean ping;
+
+        ConnectivityStatus(boolean ping) {
+            this.ping = ping;
+        }
+    }
+
+    /**
+     * Represents the sources for cluster status computation.
+     * This aims to provide an indicative of status reliability.
+     */
+    public enum ClusterStatusReliability {
+        /**
+         * Cluster Status reported based on quorum.
+         */
+        STRONG_QUORUM,
+
+        /**
+         * Cluster Status reported based on highest layout in the system.
+         */
+        WEAK_NO_QUORUM,
+
+        /**
+         * Unavailable, whenever communication fails to ALL nodes, the source of reliability is
+         * not available.
+         */
+        UNAVAILABLE
+    }
+
+    /**
      * Layout at which the report was generated.
      */
     private final Layout layout;
 
     /**
-     * Map of connectivity of the report generator client to the cluster nodes.
-     */
-    private final Map<String, NodeStatus> clientServerConnectivityStatusMap;
-
-    /**
      * Cluster health.
      */
     private final ClusterStatus clusterStatus;
+
+    /**
+     * Cluster Status Reliability (source of information)
+     */
+    private final ClusterStatusReliability clusterStatusReliability;
+
+    /**
+     * Individual Node Status (within cluster view).
+     */
+    private final Map<String, NodeStatus> clusterNodeStatusMap;
+
+    /**
+     * Map of connectivity of the client to the each cluster node.
+     */
+    private final Map<String, ConnectivityStatus> clientServerConnectivityStatusMap;
 }
--- a/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
@@ -1,18 +1,15 @@
 package org.corfudb.runtime.view;
 
-import com.google.common.collect.Sets;
-
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -21,8 +18,6 @@ import javax.annotation.Nonnull;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
-import org.corfudb.protocols.wireprotocol.ClusterState;
-import org.corfudb.protocols.wireprotocol.NodeState;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.clients.IClientRouter;
 import org.corfudb.runtime.clients.LayoutClient;
@@ -30,8 +25,9 @@ import org.corfudb.runtime.exceptions.NetworkException;
 import org.corfudb.runtime.exceptions.WorkflowException;
 import org.corfudb.runtime.exceptions.WorkflowResultUnknownException;
 import org.corfudb.runtime.exceptions.WrongEpochException;
-import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.runtime.view.ClusterStatusReport.ClusterStatus;
+import org.corfudb.runtime.view.ClusterStatusReport.ClusterStatusReliability;
+import org.corfudb.runtime.view.ClusterStatusReport.ConnectivityStatus;
 import org.corfudb.runtime.view.ClusterStatusReport.NodeStatus;
 import org.corfudb.runtime.view.Layout.LayoutSegment;
 import org.corfudb.runtime.view.workflows.AddNode;
@@ -175,7 +171,7 @@ public class ManagementView extends AbstractView {
         // stateTransfer and is in process of achieving full redundancy.
         ClusterStatus logUnitRedundancyStatus = peerResponsiveNodes.stream()
                 .anyMatch(s -> getLogUnitNodeStatusInLayout(layout, s) == NodeStatus.DB_SYNCING)
-                ? ClusterStatus.DEGRADED : ClusterStatus.STABLE;
+                ? ClusterStatus.DB_SYNCING : ClusterStatus.STABLE;
         // Check the availability of the log servers in all segments as reads to all addresses
         // should be accessible.
         ClusterStatus logunitClusterStatus = layout.getSegments().stream()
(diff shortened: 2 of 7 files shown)


runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -292,6 +277,20 @@ private Layout getHighestEpochLayout(List<String> layoutServers) {
         return layout;
     }
 
+    private Layout getHighestEpochLayout(Map<String, Layout> layoutMap) {
+        Layout layout  = null;
```
Maithem: you can use discoverLayout from #1660

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1496,24 +1508,36 @@ public void queryClusterStatus() throws Exception {
...
+        assertThat(nodeConnectivityMap.get(SERVERS.ENDPOINT_0)).isEqualTo(ConnectivityStatus.RESPONSIVE);
+        assertThat(nodeConnectivityMap.get(SERVERS.ENDPOINT_1)).isEqualTo(ConnectivityStatus.RESPONSIVE);
+        assertThat(nodeConnectivityMap.get(SERVERS.ENDPOINT_2)).isEqualTo(ConnectivityStatus.RESPONSIVE);
         assertThat(nodeStatusMap.get(SERVERS.ENDPOINT_0)).isEqualTo(NodeStatus.UP);
         assertThat(nodeStatusMap.get(SERVERS.ENDPOINT_1)).isEqualTo(NodeStatus.UP);
         assertThat(nodeStatusMap.get(SERVERS.ENDPOINT_2)).isEqualTo(NodeStatus.UP);
+        assertThat(clusterStatusReliability).isEqualTo(ClusterStatusReport.ClusterStatusReliability.STRONG_QUORUM);
         assertThat(clusterStatus.getClusterStatus()).isEqualTo(ClusterStatus.STABLE);
 
         // STEP 2.
```
Edilmo: As it's seen in this test, with the new implementation is possible to have Server Nodes with status UP but connectivity status UNRESPONSIVE. Maybe it will be good to have a comment in this code explaining this.

runtime/src/main/java/org/corfudb/runtime/view/ClusterStatusReport.java
```diff
@@ -64,17 +78,70 @@
...
-     * Map of connectivity of the report generator client to the cluster nodes.
+     * Represents the sources for cluster status computation.
+     * This aims to provide an indicative of status reliability.
      */
-    private final Map<String, NodeStatus> clientServerConnectivityStatusMap;
+    public enum ClusterStatusReliability {
+        /**
+         * Cluster Status reported based on quorum.
+         */
+        STRONG_QUORUM,
```
xnull: just QUORUM and NO_QUORUM ?

runtime/src/main/java/org/corfudb/runtime/view/ClusterStatusReport.java
```diff
@@ -64,17 +78,70 @@
...
+
+        /**
+         * Cluster Status reported based on highest layout in the system.
+         */
+        WEAK_NO_QUORUM,
+
+        /**
+         * Unavailable
+         */
+        UNAVAILABLE
```
xnull: UNAVAILABLE means? Can you please add some details

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
+            // The initial list of layoutServers is obtained from the runtime, this list might
+            // not be complete as it can be based on a a minimum of servers required for bootstrap,
+            // by setting the discoverLayoutServers flag, we request layouts from all available
+            // layout servers present in the actual layout file.
+            Map<String, Layout> layoutMap = getLayouts(layoutServers, true);
+
+            int serversLayoutResponses = layoutMap.size();
+            int quorum = runtime.getLayoutView().getQuorumNumber();
+
+            if (serversLayoutResponses == 0) {
```
xnull: you could simplify it: layoutMap.isEmpty()

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
annym: Anyway we need it for the second if condition, < quorum.

xnull: yes, just use that number for second condition

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
-     *
-     * @return ClusterStatusReport
-     */
-    public ClusterStatusReport getClusterStatus() {
+    private ClusterStatusReport getUnavailableClusterStatusReport(List<String> layoutServers, ClusterStatusReliability reliability, @Nonnull Set<String> responsiveServers, Layout layout) {
+        return new ClusterStatusReport(layout,
+                ClusterStatus.UNAVAILABLE,
+                reliability,
+                layoutServers.stream().collect(Collectors.toMap(
+                        endpoint -> endpoint,
```
xnull: just for info, you could use Function.identity()

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
+                        node -> NodeStatus.NA)),
+                layoutServers.stream().collect(Collectors.toMap(
+                        endpoint -> endpoint,
+                        endpoint -> responsiveServers.contains(endpoint) ? ConnectivityStatus.RESPONSIVE : ConnectivityStatus.UNRESPONSIVE)));
+    }
+
+    private ClusterStatusReport getUnavailableClusterStatusReport(List<String> layoutServers) {
+        return getUnavailableClusterStatusReport(layoutServers, ClusterStatusReliability.UNAVAILABLE, Collections.emptySet(), null);
+    }
```
xnull: can we pass something else but not null?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
-                            node -> NodeStatus.DOWN)),
-                    ClusterStatus.UNAVAILABLE);
+    private Map<String, Layout> getLayouts(List<String> layoutServers, Boolean discoverLayoutServers) {
+        Map<String, CompletableFuture<Layout>> layoutFuturesMap = getLayoutFutures(layoutServers);
+        Map<String, Layout> layoutMap = new HashMap<>();
+
+        // Wait on the Completable futures
+        for (Entry<String, CompletableFuture<Layout>> entry : layoutFuturesMap.entrySet()) {
+            try {
+                Layout nodeLayout = entry.getValue().get();
```
xnull: It takes linear time to connect to all servers, it should be run in parallel

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
+                        entry.getKey(), e);
+            }
         }
 
-        RuntimeLayout runtimeLayout = new RuntimeLayout(layout, runtime);
+        // If discoverLayoutServers is set, fetch layouts from all other layout servers
+        // discovered in the first search which are not contained in the initial list.
+        if (discoverLayoutServers) {
+            List<String> discoveredLayoutServers = layoutMap.values().stream()
+                    .map(l -> l.getLayoutServers())
```
xnull: can you please rename all `l`,`s` and other letters to words

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
+                    .flatMap(s -> s.stream())
+                    .distinct()
+                    .filter(s -> !layoutServers.contains(s))
+                    .collect(Collectors.toList());
+
+            if (!discoveredLayoutServers.isEmpty()) {
+                Map <String, Layout> recursiveLayouts = getLayouts(new ArrayList<>(discoveredLayoutServers), false);
```
xnull: Can we use an enum instead of boolean?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
+    private Layout getLayoutFromQuorum(Map<String, Layout> layoutMap, int quorum) {
+        Layout layout = null;
+        Map<Layout, Integer> uniqueLayoutMap = null;
+
+        // Find if layouts are the same in all nodes (at least quorum nodes should agree in the
+        // same layout) - count occurrences of each layout in cluster nodes.
+        for (Entry<String, Layout> entry : layoutMap.entrySet()) {
+            if (uniqueLayoutMap == null) {
+                uniqueLayoutMap = new HashMap<Layout, Integer>() {{
+                    put(entry.getValue(), 1);
```
xnull: Findbugs won't be happy. You create (compiler creates) an anonymous class, which is not so good. You can use ImmutableMap.of() or just create and then put the value

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
annym: what do you propose to pass? whenever this is the case, you could not connect to any layout server, hence there is no layout to take a decision on. If we pass a fake layout with all nodes unresponsive, I'm not sure how accurate this is, cause 1. it was not the layout on which this status is based on and 2. how could you know if this is not a client that came back from days down and is based on a very old layout, so we are showing out of date info...

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
annym: all calls are async and were done in the previous step, anyways we need to wait on all responses. Anyways, this part of the code will change when we have the quorum read API

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
annym: why? a enum with a unique value? Besides, remember this will change to read from quorum api.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
annym: I won't address this as this is the part of the code likely to be simplified by the read quorum API.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
xnull: As discussed: please find out if we can use Optional instead of null

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
xnull: then, ok

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
xnull: Boolean doesn't describe what is that, so, if it requires changing API, then nevermind

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
+        }
+
+        // Filter layouts present in quorum number of nodes
+        Map<Layout, Integer> uniqueLayoutsQuorumMap = uniqueLayoutMap.entrySet().stream()
+                .filter(x -> x.getValue() >= quorum)
+                .collect(Collectors.toMap(x -> x.getKey(), x -> x.getValue()));
+
+        // Select layout with greater number of occurrences.
+        if (uniqueLayoutsQuorumMap.size() > 0) {
+            Map.Entry<Layout, Integer> maxEntry = null;
```
xnull: It can be simplified to: 
Put all values from the Map to a TreeSet then `getFirst()`

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -320,125 +238,303 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
+                }
+            }
+            layout = maxEntry.getKey();
+        }
+
+        return layout;
+    }
+
+    private Layout getHighestEpochLayout(Map<String, Layout> layoutMap) {
+        Layout layout  = null;
```
xnull: Please use TreeSet and a comparator for layouts

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
annym: I've added a description, however as I said this method is likely to  be replaced.

annym:
Replaced by #1695

PR #1663, opened 2019-01-07
Report offending TX during conflict resolution

vjeko:
Description:

When reporting aborted transactions, signal to the client which
transaction ID caused the transaction to get aborted (in case of a true
conflict). This is necessary as currently there is no easy way for the
consumers of Corfu to troubleshoot these types of issues.

Why should this be merged: Transaction conflict debuggability/troubleshooting is currently lacking and very important for scale/production systems.

Related issue(s) (if applicable): None


vjeko:
Address is something that we already return.

We can read the address associated with the conflict, but this will only give us the data that was written from the other transaction -- which is marginally more useful. For example, if we have an update and and a delete, and update fails, what we will see as written to that address is "REMOVE <key>". We really need a method that can be used for troubleshooting conflicts that does not involve any reverse-engineering or guessing.

The other solution is to embed this kind of metadata within MultiSMREntry, but since that requires a change of the data-format (migration) and probably more discussion, I was reluctant to go down that path.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -131,7 +127,7 @@
      * the primary sequencer. This means that any snapshot timestamp below this
      * actual threshold would abort due to NEW_SEQUENCER cause.
      */
-    private final Cache<String, Long> conflictToGlobalTailCache;
+    private final Cache<String, ConflictTuple> conflictToGlobalTailCache;
```
annym: please update description for conflictToGlobalTailCache... also maybe ConflictTuple does not convey the actual meaning, wouldn't it be more suitable something like CommitTuple, which refers to the address and the transaction ID of this committed operation...

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenResponse.java
```diff
@@ -58,6 +66,8 @@ public TokenResponse(Token token, Map<UUID, Long> backpointerMap) {
     public TokenResponse(ByteBuf buf) {
         respType = TokenType.values()[ICorfuPayload.fromBuffer(buf, Byte.class)];
         conflictKey = ICorfuPayload.fromBuffer(buf, byte[].class);
+        conflistStream = ICorfuPayload.fromBuffer(buf, UUID.class);
```
annym: typo.. conflist...

test/src/test/java/org/corfudb/integration/TransactionAbortedIT.java
```diff
@@ -0,0 +1,108 @@
...
+                    e.getContext().getWriteSetInfo().getConflicts().keySet().stream().findFirst().get();
+            final byte[] keyHash = ConflictSetInfo.generateHashFromObject(proxyInternal, key);
+            Assertions.assertThat(Arrays.equals(keyHash, e.getConflictKey())).isTrue();
+
+            endLatch.countDown();
+        }
+    }
+
+    @Test
+    public void largeTransaction() throws InterruptedException, IOException {
```
annym: can you please add a description to this test.

Maithem:
@vjeko There's no reverse engineering. I was alluding to the later. The transaction id can be part of the ILogdata metadata.

vjeko:
@Maithem I agree with your sentiment.

How about we go ahead and go with this patch for now, and I will start working on ILogData approach (this might take a bit longer since it requires changes on Corfu, MP and Migration)? This patch also does a fail amount of cleanup related to transaction reporting, so I would prefer not to abandon the change.

Also, Is this change too intrusive given how late in the release cycle we are?

Maithem:
@vjeko Ok, that's fine. Its not that it's intrusive, my suggestion was based on the fact that we are trying to remove streamTailToGlobalTailMap from the sequencer. The sequencer ideally should just be ditching out tokens and answering tail queries (i.e. not do conflict resolution). The rest of the information can be inferred from the log.

annym:
@vjeko Can you please rebase and resolve conflicts? so we can merge this in!

vjeko:
Done!

Issue #1576, opened 2018-11-30
getOrchestrator can't make progress


Maithem:
**Bug report.**
If the layout doesn't reflect the actual state of nodes in the system workflows can get stuck, so the orchestrator should not fully rely on the layouts state, but use it as a hint.

zalokhan:
Fixed by #1665.
Please re-open if needed.

PR #1665, opened 2019-01-08
getOrchestrator Bug

Maithem:
## Overview
Since the layout's view can reflect incorrect state (i.e. responsive
servers marked as unresponsive), all nodes should be considered
when selecting an orchestrator, not just responsive ones (based
on the layout)

Why should this be merged: Unblocks the cluster when the layout has marked all the responsive nodes as unresponsive and the unresponsive node as responsive. 

Related issue(s) (if applicable): #1576

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
@@ -59,20 +59,19 @@ public abstract class WorkflowRequest {
      * orchestrator
      */
     protected ManagementClient getOrchestrator(@NonNull Layout layout) {
-        List<String> activeLayoutServers = layout.getLayoutServers().stream()
-                .filter(s -> !layout.getUnresponsiveServers().contains(s)
-                        && !s.equals(nodeForWorkflow))
+        List<String> availableLayoutServers = layout.getLayoutServers().stream()
+                .filter(s -> !s.equals(nodeForWorkflow))
                 .collect(Collectors.toList());
 
-        if (activeLayoutServers.isEmpty()) {
+        if (availableLayoutServers.isEmpty()) {
             throw new WorkflowException("getOrchestrator: no available orchestrators " + layout);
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -59,20 +59,19 @@
...
+                .filter(s -> !s.equals(nodeForWorkflow))
                 .collect(Collectors.toList());
 
-        if (activeLayoutServers.isEmpty()) {
+        if (availableLayoutServers.isEmpty()) {
             throw new WorkflowException("getOrchestrator: no available orchestrators " + layout);
         }
 
         // Select an available orchestrator
         ManagementClient managementClient = runtime.getLayoutView().getRuntimeLayout(layout)
```
annym: what if the selected orchestrator node is actually unresponsive? couldn't we fall into an infinite loop and never complete the workflows?

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -59,20 +59,19 @@
      * orchestrator
      */
     protected ManagementClient getOrchestrator(@NonNull Layout layout) {
-        List<String> activeLayoutServers = layout.getLayoutServers().stream()
-                .filter(s -> !layout.getUnresponsiveServers().contains(s)
-                        && !s.equals(nodeForWorkflow))
+        List<String> availableLayoutServers = layout.getLayoutServers().stream()
+                .filter(s -> !s.equals(nodeForWorkflow))
```
WenbinZhu: Would this approach degrade the performance of workflow since we `pingSync` the endpoint everytime before we select it? Maybe it's better to first try to getOrchestrator from the responsiveServers as the layout suggests and if none of them are actually responsive, then we try to select from the unresponsiveServers?

Issue #1668, opened 2019-01-09
Restore the interrupted status


vjeko:
Ensure that all code paths that deal with interrupts re-enable the interrupt flag. In order to solve this issue systematically, during the construction of UnrecoverableCorfuInterruptedError we call Thread.currentThread().interrupt().

PR #1669, opened 2019-01-09
Restore the interrupted status

vjeko:
Ensure that all code paths that deal with interrupts re-enable the interrupt flag. In order to solve this issue systematically, during the construction of UnrecoverableCorfuInterruptedError we call Thread.currentThread().interrupt().

After capturing InterruptedException, we need to re-enable the interrupt
flag, since InterruptedException is wrapped in an unchecked exception.

Related issue(s) (if applicable): #1668

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/ViewsGarbageCollector.java
+++ b/runtime/src/main/java/org/corfudb/runtime/ViewsGarbageCollector.java
@@ -4,6 +4,7 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.exceptions.GarbageCollectorException;
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.runtime.view.Address;
 
 import javax.annotation.concurrent.NotThreadSafe;
@@ -83,8 +84,7 @@ public class ViewsGarbageCollector {
                     endTs - startTs, runtime.getObjectsView().getObjectCache().size(), trimMark);
         } catch (Exception e) {
             if (e.getCause() instanceof InterruptedException) {
-                Thread.currentThread().interrupt();
-                throw new GarbageCollectorException(e);
+                throw new UnrecoverableCorfuInterruptedError((InterruptedException) e.getCause());
             } else {
                 log.error("Encountered an error while running runtime GC", e);
             }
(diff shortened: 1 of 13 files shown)


StrikingThirteen:
@vjeko I have a question/concern. I understand the rationale of centralizing the "reseting of interrupt flag" in one place was the reason for having UnrecoverableCorfuInterruptedError reset the flag. But do you think it is intuitive that throwing a new exception has a side effect of setting the interrupt flag? If you think it's justified then that's fine. In this case can you please add comments UnrecoverableCorfuInterruptedError to clarifying that it has the mentioned effect of reseting the flag?

runtime/src/main/java/org/corfudb/runtime/exceptions/unrecoverable/UnrecoverableCorfuInterruptedError.java
```diff
@@ -14,6 +14,8 @@
      */
     public UnrecoverableCorfuInterruptedError(@Nonnull InterruptedException cause) {
         super(cause);
+        // Restore the interrupted status.
```
StrikingThirteen: Exceptions are also part of api, can you please add comments to UnrecoverableCorfuInterruptedError about its behavior of setting the interrupt flag?

StrikingThirteen:
also consider changing waitForAppToFinish in LongevityApp's use of Thread.currentThread.interrupt() similar to above changes.

vjeko:
@StrikingThirteen As much as I love functional programing and side-effect free functions, there is a more practical concern -- under no circumstances should Corfu be allowed to catch Interrupted Exceptions without re-setting the interrupt flag (it really screws up the client). Right now, the interrupt flag is being set out-of-bound which is pretty error prone. I just wanted to make sure, as we keep adding more code that we do not end up in the same predicament.

And yes, I will amend to comments.

runtime/src/main/java/org/corfudb/runtime/ViewsGarbageCollector.java
```diff
@@ -83,8 +84,7 @@ public void runRuntimeGC() {
                     endTs - startTs, runtime.getObjectsView().getObjectCache().size(), trimMark);
         } catch (Exception e) {
             if (e.getCause() instanceof InterruptedException) {
-                Thread.currentThread().interrupt();
-                throw new GarbageCollectorException(e);
```
zalokhan: Throw GarbageCollectorException instead of UnrecoverableCorfuInterruptedError?

vjeko: I would like the client to have only one un-checked exception type that signals interrupts. Is there a specific reason why GarbageCollectorException was proffered here? Also looks like this is the sole usage of GarbageCollectorException.

zalokhan: I see.

> Is there a specific reason why GarbageCollectorException was proffered here? 
I'm not sure.

Just make sure you know that UnrecoverableCorfuInterruptedError is an Error which would require you to catch a throwable now on the client side code, while GarbageCollectorException is an Exception.

vjeko:
I've addressed all the comments.

StrikingThirteen:
LGTM

Issue #1673, opened 2019-01-14
Retry Trim


annym:
## Overview

We currently do not retry Trim requests. In the event of a WrongEpochException (given to changes in the system) we fetch the new layout and return, skipping this task (unless user manually retries). In the case of applications with scheduled/periodic trim/cp tasks, this could lead to eventual skips of any of these tasks for every cycle.

PR #1675, opened 2019-01-16
Retry prefixTrim for Runtime on wrong epoch

annym:
## Overview

If client is in a stale epoch, prefixTrim is retried after a layout
invalidate.

Why should this be merged: prevents prefixTrim tasks failing due to clients on the wrong epoch. This is relevant as trimming is an operation typically performed periodically, if we do not retry, trim will have to wait for the next cycle unnecessarily. 

Related issue(s) (if applicable): #1673

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -8,6 +8,9 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.collect.Iterables;
+import com.google.common.util.concurrent.UncheckedExecutionException;
+
+import io.netty.handler.timeout.TimeoutException;
 
 import java.time.Duration;
 import java.util.Comparator;
@@ -22,9 +25,6 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-import com.google.common.util.concurrent.ExecutionError;
-import com.google.common.util.concurrent.UncheckedExecutionException;
-import io.netty.handler.timeout.TimeoutException;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.protocols.wireprotocol.DataType;
@@ -348,9 +348,12 @@ public class AddressSpaceView extends AbstractView {
                                     .flatMap(stripe -> stripe.getLogServers().stream())
                                     .map(e::getLogUnitClient)
                                     .map(client -> client.prefixTrim(address))
-                                    .forEach(CFUtils::getUninterruptibly);
-                            return null;    // No return value
-                        }, true);
+                                    .forEach(cf -> {CFUtils.getUninterruptibly(cf,
+                                            NetworkException.class, TimeoutException.class,
+                                            WrongEpochException.class);
+                                    });
+                            return null;
+                }, true);
                 // TODO(Maithem): trimCache should be epoch aware?
                 runtime.getSequencerView().trimCache(address.getSequence());
                 break;
@@ -358,6 +361,21 @@ public class AddressSpaceView extends AbstractView {
                 log.warn("prefixTrim: encountered a network error on try {}", x, e);
                 Duration retryRate = runtime.getParameters().getConnectionRetryRate();
                 Sleep.sleepUninterruptibly(retryRate);
+            } catch (WrongEpochException wee) {
+                long serverEpoch = wee.getCorrectEpoch();
+                // Retry if wrongEpochException corresponds to message epoch (only)
+                if (address.getEpoch() == serverEpoch) {
+                    long runtimeEpoch = runtime.getLayoutView().getLayout().getEpoch();
+                    log.warn("prefixTrim[{}]: wrongEpochException, runtime is in epoch {}, " +
+                            "while server is in epoch {}. Invalidate layout for this client " +
+                            "and retry, attempt: {}/{}", address, runtimeEpoch, serverEpoch, x+1, numRetries);
+                    runtime.invalidateLayout();
+                } else {
+                    // wrongEpochException corresponds to a stale trim address (prefix trim token on the wrong epoch)
+                    log.error("prefixTrim[{}]: stale prefix trim. Prefix trim on wrong epoch {}, " +
+                            "while server on epoch {}", address, address.getEpoch(), serverEpoch);
+                    throw wee;
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -349,15 +349,32 @@ public void prefixTrim(final Token address) {
...
+                            return null;
+                }, true);
                 // TODO(Maithem): trimCache should be epoch aware?
                 runtime.getSequencerView().trimCache(address.getSequence());
                 break;
             } catch (NetworkException | TimeoutException e) {
                 log.warn("prefixTrim: encountered a network error on try {}", x, e);
                 Duration retryRate = runtime.getParameters().getConnectionRetryRate();
                 Sleep.sleepUninterruptibly(retryRate);
+            } catch (RuntimeException re) {
```
Maithem: I believe you catch RuntimeException and not WrongEpochException because the server doesn't return the correct error message, but fails and the "uncaught exception handler" wraps the throwable in a RuntimeException and sends it back and then you need to unwrap it as such.

I think you should just fix the server side and just catch WrongEpochException

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
```diff
@@ -183,6 +191,63 @@ public void testCheckpointTrimDuringPlayback() throws Exception {
...
+
+        // Initialize map.
+        Map<String, String> testMap = getDefaultRuntime().getObjectsView().build()
+                .setTypeToken(new TypeToken<SMRMap<String, String>>() {
+                })
+                .setStreamName("test")
+                .open();
+
+        // Populate map with 5 entries.
+        testMap.put("k1", "1");
```
Maithem: I think you can omit writing, just exercise the ::prefixTrim ?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: I'm not clear with this. Even if we catch this on the server side, we can't do anything as it's the client who must move to the new epoch and retry. Are you proposing sending a new kind of message?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -348,16 +348,33 @@ public void prefixTrim(final Token address) {
...
+                long runtimeEpoch = runtime.getLayoutView().getLayout().getEpoch();
+                long serverEpoch = wee.getCorrectEpoch();
+                // Only retry if wrongEpochException corresponds to message epoch
+                if (runtimeEpoch != serverEpoch) {
+                    log.warn("prefixTrim[{}]: wrongEpochException, runtime is in epoch {}, " +
+                            "while server is in epoch {}. Invalidate layout for this client " +
+                            "and retry, attempt: {}/{}", address, runtimeEpoch, serverEpoch, x+1, numRetries);
+                    runtime.invalidateLayout();
+                } else {
+                    // wrongEpochException corresponds to a stale trim address (prefix trim token on the wrong epoch)
```
WenbinZhu: If it's a stale trim address, we should get a `StaleTokenException`, right?  If our current code is actually throwing a `WrongEpochException` in this case, I think it's not appropriate and should be fixed. what do you think?

StrikingThirteen: Maybe log what you have commented here.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -348,16 +348,33 @@ public void prefixTrim(final Token address) {
...
             } catch (NetworkException | TimeoutException e) {
                 log.warn("prefixTrim: encountered a network error on try {}", x, e);
                 Duration retryRate = runtime.getParameters().getConnectionRetryRate();
                 Sleep.sleepUninterruptibly(retryRate);
+            } catch (WrongEpochException wee) {
+                log.warn("prefixTrim[{}]: wrongEpochException", address);
+                long runtimeEpoch = runtime.getLayoutView().getLayout().getEpoch();
+                long serverEpoch = wee.getCorrectEpoch();
+                // Only retry if wrongEpochException corresponds to message epoch
+                if (runtimeEpoch != serverEpoch) {
```
Maithem: It seems like a wrong epoch exception can originate from the server router, or the batch writer. 
The batch writer check the epoch of address.getEpoch, bit the server epoch will check the msg.getEpoch, which is stamped by the client's layout. So why not only retry if address.getEpoch == serverEpoch? 

Because the current check can retry and fail, where as it could have failed on the first retry.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Stale token and wrong epoch are not the same thing. The prior is specific to transaction tokens.

WenbinZhu: @Maithem I saw the only place to throw `StaleTokenException` is in `AddressSpaceView::write()`, so I think it's not transaction specific, right?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
annym: Done

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -348,16 +348,33 @@ public void prefixTrim(final Token address) {
...
+                    log.warn("prefixTrim[{}]: wrongEpochException, runtime is in epoch {}, " +
+                            "while server is in epoch {}. Invalidate layout for this client " +
+                            "and retry, attempt: {}/{}", address, runtimeEpoch, serverEpoch, x+1, numRetries);
+                    runtime.invalidateLayout();
+                } else {
+                    // wrongEpochException corresponds to a stale trim address (prefix trim token on the wrong epoch)
+                    throw new RuntimeException(wee);
```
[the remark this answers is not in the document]
annym: Is this correct? I thought it is cleaner to always throw RuntimeExceptions, so the client always knows what to expect and the cause is what is more specific. But this was just my assumption.

annym:
@Maithem @zalokhan @StrikingThirteen comments addressed, please review.

annym:
The only test failing before merge is a well-known failed test: nodeDownAndLinkFailureTest

Issue #1676, opened 2019-01-17
Transaction Abort Conflict report Stream ID


annym:
## Overview

When a transaction abort exception is raised due to a conflict, though the KEY is reported, the stream id is not provided for complete debugging. This issue is different from #1663 which aims to furthermore provide additional info of the other conflicting transaction.

PR #1677, opened 2019-01-17
Transaction Abort Exception Conflict Stream ID

annym:
## Overview

Description:

Add information of conflicting stream ID to TAE.

Why should this be merged: improves debugging.

Related issue(s) (if applicable): #1676 

- [x ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -2,17 +2,34 @@ package org.corfudb.infrastructure;
 
 import static org.corfudb.protocols.wireprotocol.TokenType.TX_ABORT_NEWSEQ;
 import static org.corfudb.protocols.wireprotocol.TokenType.TX_ABORT_SEQ_OVERFLOW;
-
 import com.codahale.metrics.Timer;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+
 import io.netty.channel.ChannelHandlerContext;
+
+import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+
 import org.corfudb.protocols.wireprotocol.CorfuMsg;
 import org.corfudb.protocols.wireprotocol.CorfuMsgType;
 import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -129,10 +125,17 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
 
             // Is our token a valid type?
             if (tokenResponse.getRespType() == TokenType.TX_ABORT_CONFLICT) {
+                UUID conflictStream = null;
+                byte[] conflictKey = tokenResponse.getConflictKey();
+                if (conflictKey != null && conflictKey.length > 0) {
+                    conflictStream = conflictInfo.getStreamIDForConflictKey(conflictKey);
```
Maithem: What if the same key exists in two streams, it can match the wrong stream right?

annym: Yes, you're right. I was trying to avoid sending more info on the TokenResponse, I guess there is no way to go around this, I'll just make the change on the sequencer server side.

vjeko:
Anny, my change is already addressing that issue of missing streamID. If you think it is not, can you please re-base on top of my change in order to avoid conflicts?

annym:
@vjeko yes, I had this discussion with Maithem today. Because yes this small part is already addressed on your change. However, that PR also includes the conflicting transaction ID which further makes changes on the sequencer's cache, and there is an ongoing discussion on wether to go for this solution or persist this info in the log metadata and retrieve the address. We agreed to separate it, and when that discussion is finalized we take either your change or push for the other solution. Is that ok? we can then rebase your PR.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -248,7 +250,8 @@ private String getConflictHashCode(UUID streamId, byte[] conflictParam) {
      *     cause.
      */
     private TokenType txnCanCommit(TxResolutionInfo txInfo, /** Input. */
-                                  AtomicReference<byte[]> conflictKey /** Output. */) {
+                                  AtomicReference<byte[]> conflictKey, /** Output. */
+                                  AtomicReference<UUID> conflictStreamID /** Output. */) {
```
Maithem: It seems like we just return a conflict and not all conflicts. Do we care about all conflicts?

vjeko:
There are several concerns I have witch that approach.

1) The information cannot be fully captured. What happens when one of the clients dies and the log entry gets hole-filled? This seems like a best effort-solution, and not a real one.
2) The solution you are proposing, can it be implemented fully on the Corfu side? Is it enough to do transaction ID resolution during TXEnd(), or are there any other code paths?
3) Not technical, but tactical -- the data-migration aspect is going to be a bigger issue than what you expect. Given the influx of issues, and the time constrains I am not sure we will be able to pull it off.

Shipping the next release without this feature is a pretty bad idea, especially if we run into problems during production.

Maithem:
See comments inline:

> The information cannot be fully captured. What happens when one of the clients dies and the log entry gets hole-filled? This seems like a best effort-solution, and not a real one.

This is actually another reason why we shouldn't proceed with #1663 .  In case of a hole, you'll even return more wrong information. A hole, is a no-op, so ideally it shouldn't even abort the transaction. 
The source of truth is the materialized log and not the sequencer. Therefore, performing the reads as I suggested initially is the "real solution" and what's being proposed in #1663 is an intermediate solution, which is also faulty. 


> The solution you are proposing, can it be implemented fully on the Corfu side? Is it enough to do transaction ID resolution during TXEnd(), or are there any other code paths?

I'm not sure if you know, but transaction resolution used to be fully implemented on the client side. Mapping transaction IDs to conflicting transactions can be achieved easily via application-level logging, no extra Corfu support is needed, which is why I don't support an intermediate solution. 

> Not technical, but tactical -- the data-migration aspect is going to be a bigger issue than what you expect. Given the influx of issues, and the time constrains I am not sure we will be able to pull it off.

Since I'm involved with the migration efforts, I fully understand the implications and that's why I didn't push for any data format changes in #1663 even thought it is the right approach. 

> Shipping the next release without this feature is a pretty bad idea, especially if we run into problems during production.

The feature can be achieve by other means, its not predicated on the changes in #1663. I'm just suggesting a different and less invasive approach as a intermediate solution. 

The alternative approach is to log (on the application side) the id and address of committed transactions. Since the "abort"/conflict address is equal to the address of the transaction causing the abort, you can correlate the two transactions by inspecting the logs (application logs, not corfu's address space).

vjeko:
Right, we can correlate log addresses and transaction IDs on the client side. Right now, that information is being printed, but not percolated to the client as part of the token response. I want to be able to deal with it programmatically.

```
                if (v != null && v > txSnapshotTimestamp.getSequence()) {
                    log.debug("ABORT[{}] conflict-key[{}](ts={})", txInfo, conflictParam, v);
                    conflictKey.set(conflictParam);
                    response.set(TokenType.TX_ABORT_CONFLICT);
                    break;
                }
```

vjeko:
@annym @Maithem I have updated my patch, so now we do not add any extra state on the sequencer.

The main purpose of the patch is to propagate the conflicting address to the client.

annym:
This change is already included in PR #1663

PR #1678, opened 2019-01-17
Bounded Reads

Maithem:
## Overview
When the untrimmed range of the address space is large, the FastObjectLoader
can run out of memory, because it keeps submitting read requests without
taking into account the number of pending jobs. This patch adds a limit
on the number of pending jobs.

Why should this be merged: Fixes an issue where the fast loader can't load a very large address space, because it runs out of memory

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
@@ -17,9 +17,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 
@@ -41,6 +42,7 @@ import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.collections.CorfuTable;
 import org.corfudb.runtime.collections.CorfuTable.IndexRegistry;
 import org.corfudb.runtime.collections.SMRMap;
+import org.corfudb.runtime.exceptions.FastObjectLoaderException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.runtime.object.CorfuCompileProxy;
 import org.corfudb.runtime.view.Address;
@@ -75,6 +77,7 @@ public class FastObjectLoader {
     static final int DEFAULT_TIMEOUT_MINUTES_FAST_LOADING = 30;
     static final int NUMBER_OF_ATTEMPT = 3;
     static final int STATUS_UPDATE_PACE = 10000;
+    static final int DEFAULT_NUMBER_OF_PENDING_FUTURES = 1_000;
 
     private CorfuRuntime runtime;
 
@@ -85,6 +88,10 @@ public class FastObjectLoader {
     @Setter
     @Getter
     private boolean loadInCache;
+
+    @Setter
+    @Getter
+    int numberOfPendingFutures = DEFAULT_NUMBER_OF_PENDING_FUTURES;
 
     @Getter
     private long logHead = Address.NON_EXIST;
@@ -112,6 +119,9 @@ public class FastObjectLoader {
 
     @VisibleForTesting
     void setLogTail(long tail) { this.logTail = tail; }
+
+    // A future to track the last submitted read request
+    volatile private Future lastReadRequest;
 
     /**
      * Enable whiteList mode where we only reconstruct
@@ -187,8 +197,6 @@ public class FastObjectLoader {
     private int retryIteration = 0;
     private long nextRead;
 
-    private List<Future> futureList;
-
     public FastObjectLoader(@Nonnull final CorfuRuntime corfuRuntime) {
         this.runtime = corfuRuntime;
         loadInCache = !corfuRuntime.getParameters().isCacheDisabled();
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -216,8 +222,15 @@ public void addStreamToIgnore(String streamName) {
...
+        // Note that the queue implementation requires the corePoolSize to
+        // be equal to maximumPoolSize, so this should be fine for a single threaded
+        // executor
+        necromancer = new ThreadPoolExecutor(1, 1,
+                0L, TimeUnit.MILLISECONDS,
+                new BoundedQueue<>(numberOfPendingFutures),
+                new ThreadFactoryBuilder()
+                        .setNameFormat("FastObjectLoaderReaderThread-%d").build());
+
         futureList = new ArrayList<>();
```
medhavidhawan: how will the futureList be cleaned up ? That one keeps growing in size with completed futures ?

Maithem: I removed the array, its actually not needed.

Maithem: I updated the patch.

Maithem:
@medhavidhawan I did test it with 2 million entries.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -758,5 +775,30 @@ public void updateLatestCheckpointIfLater(UUID checkPointId) {
             }
         }
     }
+
+    /**
+     * This queue implementation is to be used by single threaded exeuctors
```
StrikingThirteen: exeuctors -> executor

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -758,5 +775,30 @@ public void updateLatestCheckpointIfLater(UUID checkPointId) {
...
+     * to restrict the amount of pending job submissions.
+     */
+    public class BoundedQueue<E> extends ArrayBlockingQueue<E> {
+
+        public BoundedQueue(int size) {
+            // This queue will be used to processes a consecutive range
+            // of elements: FIFO order is needed. Thus, fair=true
+            super(size, true);
```
StrikingThirteen: I think fair's FIFO relates to threads access order not queue's elements. Doesn't this data structure specifically mentioned to be used single threaded?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -758,5 +775,30 @@ public void updateLatestCheckpointIfLater(UUID checkPointId) {
             }
         }
     }
+
+    /**
+     * This queue implementation is to be used by single threaded exeuctors
+     * to restrict the amount of pending job submissions.
+     */
+    public class BoundedQueue<E> extends ArrayBlockingQueue<E> {
```
StrikingThirteen: Why capacity of ArrayBlockingQueue or LinkedBlockingQueue is not enough for the bounding purpose?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -758,5 +775,30 @@ public void updateLatestCheckpointIfLater(UUID checkPointId) {
...
+
+        public BoundedQueue(int size) {
+            // This queue will be used to processes a consecutive range
```
annym: processes -> process

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
annym: Can't we just use the public ArrayBlockingQueue(int capacity,
                  boolean fair) directly?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -691,7 +708,7 @@ private void applyForEachAddress(BiConsumer<Long, ILogData> logDataProcessor) {
                 long address = entry.getKey();
                 ILogData logData = entry.getValue();
                 if (address != addressProcessed + 1) {
-                    fail("We missed an entry. It can lead to correctness issues.");
+                    throw new IllegalStateException("We missed an entry. It can lead to correctness issues.");
```
annym: should we add the address that was missed?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
Maithem: A queue ADT is by definition FIFO. I though there were multiple producers, but I realized that its a single producer, single consumer, so it really doesn't matter.

I'm thinking about removing the executor, I don't feel like buffering is really necessary here. What do you think ?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
Maithem: I don't believe so.

When the executor consumes the queue, it will use Queue::offer to enqueue jobs. If the queue has reached its limit, offer will return false and the ExecutorService::submit will throw a RejectedExecutionException. That will break the FastObjectLoader's logic. Ideally, we want to wait (i.e. block) until some of the pending jobs complete and free slots on the queue. By bounding the number of pending tasks, we restrict the memory usage, which is required to fix the memory "leak" issue that this PR attempts to address.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -216,32 +224,50 @@ public void addStreamToIgnore(String streamName) {
...
         }
-        for (Future future : futureList) {
-            CFUtils.getUninterruptibly(future);
 
+        if (lastReadRequest == null) {
+            log.info("killNecromancer: no read requests have been processed.");
+            return;
         }
+
+        // Only waiting on the last requests seems like a hack, but it should
```
lixinchengdu: last requests -> last request

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
StrikingThirteen: @Maithem I think there is a little extra overhead when fair is true but shouldn't matter much considering it is used in single threads. 
As for removing the executor, keeping it around adds the resilience of keeping the pool size to exactly one without extra effort in case of a necromancer thread terminating. A mild positive point for keeping it. I'm guessing you're concerned with executor's overhead that may not add much value?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
StrikingThirteen: Yeah, ADQ / LBQ wouldn't work as you want to block the offers when full. So +1 for the bounded queue. The only thing is that core/max should be the same in the executor which they are.

PR #1679, opened 2019-01-17
Fix incorrect sanity check for docker container restart

WenbinZhu:
## Overview

Fix incorrect sanity check for docker container restart which currently prevents normal restart attempts.

- [x] Public API has Javadoc

### DIFF
--- a/it/src/main/java/org/corfudb/universe/util/DockerManager.java
+++ b/it/src/main/java/org/corfudb/universe/util/DockerManager.java
@@ -134,8 +134,8 @@ public class DockerManager {
 
         try {
             ContainerInfo container = docker.inspectContainer(containerName);
-            if (container.state().running() || container.state().paused()) {
-                log.warn("The container `{}` already running, should stop before restart", container.name());
+            if (container.state().restarting()) {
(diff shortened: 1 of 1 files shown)


it/src/main/java/org/corfudb/universe/util/DockerManager.java
```diff
@@ -134,8 +134,8 @@ public void restart(String containerName) {
 
         try {
             ContainerInfo container = docker.inspectContainer(containerName);
-            if (container.state().running() || container.state().paused()) {
-                log.warn("The container `{}` already running, should stop before restart", container.name());
+            if (container.state().restarting()) {
```
StrikingThirteen: Logically it makes sense to be able to restart a running container. The old one however was preventing the restart of running or paused containers. Was that by design or a bug?

WenbinZhu: It's a bug. We have both `start()` and `restart()`. `start()` is for stopped containers, while `restart()` is for running ones. But the old logic is same for these two, so it should be a bug.

Issue #1651, opened 2019-01-01
Periodic Object Sync Option


Maithem:
**Feature request.** 
Add an option to auto-sync objects periodically. This can be very helpful in a cluster setting to minimize the cost of clients syncing shared streams.

Maithem:
outdated.

PR #1680, opened 2019-01-17
Auto Object Sync

Maithem:
## Overview
This patch adds the ability to periodically sync all opened
objects to the global tail. This is required to amortize
sync costs for clients that have irregular access patterns.

Related issue(s) (if applicable): #1651

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
```diff
@@ -23,39 +24,71 @@
...
     private boolean started;
 
-    final ScheduledExecutorService gcThread = Executors.newSingleThreadScheduledExecutor(
+    @Getter
+    @Setter
+    long objectSyncDelayInSeconds = 30;
+
+    @Getter
+    @Setter
+    long objectGCPeriodInSeconds = 20 * 60;
```
medhavidhawan: add comments as to what these variables are for.

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
```diff
@@ -23,39 +24,71 @@
 
 @Slf4j
 @NotThreadSafe
-public class ViewsGarbageCollector {
+public class ViewsMaintenance {
```
medhavidhawan: Please update the class level comments.

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
```diff
@@ -23,39 +24,71 @@
...
-    final ScheduledExecutorService gcThread = Executors.newSingleThreadScheduledExecutor(
+    @Getter
+    @Setter
+    long objectSyncDelayInSeconds = 30;
+
+    @Getter
+    @Setter
+    long objectGCPeriodInSeconds = 20 * 60;
+
+    final ScheduledExecutorService maintenance = Executors.newSingleThreadScheduledExecutor(
```
medhavidhawan: hopefully running both tasks on the same single threaded executor is ok. Could you just please add a comment.

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
```diff
@@ -23,39 +24,71 @@
...
 
     private long trimMark = Address.NON_ADDRESS;
 
     @Getter
     private boolean started;
 
-    final ScheduledExecutorService gcThread = Executors.newSingleThreadScheduledExecutor(
+    @Getter
+    @Setter
+    long objectSyncDelayInSeconds = 30;
```
lixinchengdu: One purpose of periodical sync is to avoid expensive syncing due to running out of cache. 
Is it better if objectSyncDelayInSeconds is a function (probably just as simple as a linear function) of the runtime cache size?

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
Maithem: Updated the class comment.

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
Maithem: Its not just a function of the cache, but also the time it takes to resolve the backpointers. I also want to keep the parameters simple.

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
Maithem: done.

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
Maithem: I think it should be OK. GC is just purely in memory operation and if object syncing takes soo long that the in memory GC becomes a problem, then the sync it self will probably be an issue before the side effects of not doing GC becomes an issue.

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
```diff
@@ -13,49 +14,90 @@
...
     public void start() {
-        gcThread.scheduleAtFixedRate(() -> runRuntimeGC(),
-                runtime.getParameters().getRuntimeGCPeriod().toMinutes(),
-                runtime.getParameters().getRuntimeGCPeriod().toMinutes(),
-                TimeUnit.MINUTES);
+        maintenance.scheduleAtFixedRate(() -> runRuntimeGC(),
+                objectGCPeriodInSeconds, objectGCPeriodInSeconds,
+                TimeUnit.SECONDS);
+        maintenance.scheduleAtFixedRate(() -> runSyncObjects(),
+                objectSyncDelayInSeconds, objectGCPeriodInSeconds,
```
annym: is this right? initial delay = 30sec, and then run every 20min? or is it intended to run every 30secs? with such a big gap wouldn't it be less risk to separate both tasks?

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
```diff
@@ -13,49 +14,90 @@
...
     public void stop() {
-        gcThread.shutdownNow();
+        maintenance.shutdownNow();
         this.started = false;
     }
 
+    /**
+     * Objects are synced on access, if a client access an object irregularly, then over time
+     * it can experience unpredictable latencies because of syncing. When used, the ViewsMaintinance
+     * will periodically sync the object ao amortize the sync costs.
```
annym: ao -> typo

runtime/src/main/java/org/corfudb/runtime/ViewsMaintenance.java
```diff
@@ -13,49 +14,90 @@
...
+        try {
+            long s1 = System.currentTimeMillis();
+            runtime.getObjectsView().syncObjects();
+            long s2 = System.currentTimeMillis();
+            log.info("syncObjects: sync took {} ms, total maps {}", s2 - s1, runtime.getObjectsView()
+                    .getObjectCache().size());
+        } catch (Exception e) {
+            if (e.getCause() instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+                throw new GarbageCollectorException(e);
```
annym: Is this the right error to throw? It is not a GC error, should we introduce a SyncObjectException or viewsMaintenanceException?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -302,6 +302,19 @@ public void seek(long globalAddress) {
         smrStream.seek(globalAddress);
     }
 
+    /**
```
annym: Can we add a comment that suggests the difference from syncObjectUnsafe

test/src/test/java/org/corfudb/runtime/view/ViewsMaintenanceTest.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        // Notice that we don't access table1Rt2's map, but we expect it to be
+        // synced
+        final int numRefreshCycles = 3;
+        Sleep.sleepUninterruptibly(Duration.ofSeconds(objectRefreshPeriod * numRefreshCycles));
+        long rt2TableVersionAfterSync = ((CorfuCompileProxy) ((ICorfuSMR) table1Rt2).getCorfuSMRProxy())
+                .getUnderlyingObject().getVersionUnsafe();
+        // Verify that it was synced up to the tail
+        assertThat(rt2TableVersionAfterSync).isEqualTo(rt2.getSequencerView().query().getToken().getSequence());
```
annym: Though it is the same, to avoid potential errors on  rt2, wouldn't it be safer to compare to rt1?

Issue #1681, opened 2019-01-18
MultiCheckpointWriter Cache Thrashing 


Maithem:
**Bug report.**
For deployments where the CorfuRuntime is shared between the application and MultiCheckpointWriter, the cache can be severely thrashed causing unpredictable latency for the application access and checkpoint thread to be so slow that prefix trims can never be triggered.  

If there are streams being mutated at a very high rate, then the MultiCheckpointWriter snapshot can be stale (i.e. escapes thes client's address space cache) very quickly. As a consequence, the checkpointer thread competes with the application over the cache, which manifests in NoRollBackExceptions, causing expensive complete stream rebuilds every couple of accesses.

PR #1682, opened 2019-01-18
MultiVersion Checkpoints

Maithem:
## Overview
Instead of using the same snapshot to checkpoint all the maps
to be checkpointed, this patch makes the CheckpointWriter
use increasing checkpoint snapshots for every map and
therefore decreasing the probability of rolling back the stream.
Because multiple increasing snapshots are used, only the smallest
snapshot is used for trimming.

Why should this be merged: Improves the currently broken compaction design by a "bit"

Related issue(s) (if applicable): #1681

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
@@ -29,13 +29,6 @@ public class MultiCheckpointWriter<T extends Map> {
     @Getter
     private List<ICorfuSMR<T>> maps = new ArrayList<>();
 
-    @Getter
-    private List<Long> checkpointLogAddresses = new ArrayList<>();
-
-    @Setter
-    @Getter
-    boolean enablePutAll = false;
-
     /** Add a map to the list of maps to be checkpointed by this class. */
     @SuppressWarnings("unchecked")
     public void addMap(T map) {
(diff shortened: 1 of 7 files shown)


test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -292,6 +298,7 @@ public void checkpointWriterTest() throws Exception {
      *  to use a different position 'history' for our assertion
      *  check.
      */
+    /**
```
Maithem: Still need to restore this test. Not sure if its still applicable.

Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
```diff
@@ -50,75 +43,65 @@ public void addAllMaps(Collection<T> maps) {
...
                 log.trace("appendCheckpoints: checkpoint map {} begin",
                         Utils.toReadableId(map.getCorfuStreamID()));
-                List<Long> addresses = cpw.appendCheckpoint();
+                Token minCPSnapshot = cpw.appendCheckpoint();
+
+                if (minSnapshot == Token.UNINITIALIZED) {
+                    minSnapshot = minCPSnapshot;
+                } else if (minSnapshot.getEpoch() != minCPSnapshot.getEpoch()) {
+                    //TODO(Maithem): add epochs?
+                    throw new IllegalStateException("Epoch changed during GC cycle aborting");
```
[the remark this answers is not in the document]
Maithem: Yes, the checkpointer thread catches RuntimeException

PR #1684, opened 2019-01-18
Mulithreaded Checkpointer

Maithem:
## Overview
This patch introduces multithreading to CheckpointWriter in order
to emit checkpoints faster for very large tables.

Why should this be merged: for very large tables, a single writer can be very slow, so writing the checkpoint state in parallel can improve performance by reducing the checkpointing cycle. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -4,6 +4,24 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -217,72 +211,58 @@ private long nonCachedAppend(Object object, UUID ... streamIDs) {
...
                 }
 
                 CheckpointEntry cp = new CheckpointEntry(CheckpointEntry
                         .CheckpointEntryType.CONTINUATION,
                         author, checkpointId, streamId, mdkv, smrEntries);
-                long pos = nonCachedAppend(cp, checkpointStreamID);
-
-                postAppendFunc.accept(cp, pos);
-                continuationAddresses.add(pos);
+                futures.add(CompletableFuture.runAsync(() -> submitWrite(cp, checkpointStreamID),
```
lixinchengdu: Why not including the generation of checkpoint entry in the async task?

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
```diff
@@ -50,75 +43,65 @@ public void addAllMaps(Collection<T> maps) {
...
                         ((CorfuCompileProxy<Map>) map.getCorfuSMRProxy())
                                 .getSerializer();
                 cpw.setSerializer(serializer);
-                cpw.setPostAppendFunc(postAppendFunc);
                 log.trace("appendCheckpoints: checkpoint map {} begin",
                         Utils.toReadableId(map.getCorfuStreamID()));
-                List<Long> addresses = cpw.appendCheckpoint();
+                Token minCPSnapshot = cpw.appendCheckpoint();
+
+                if (minSnapshot == Token.UNINITIALIZED) {
```
annym: It would always come in here, right? we've never changed it since initialized to this value.

Maithem:
@zalokhan you can use this
https://github.com/CorfuDB/CorfuDB/compare/multiThreadCPWriter

Maithem:
No longer needed.

Issue #1688, opened 2019-01-23
Control Thread Usage In FastObjectLoader::resurrectCheckpoints


Maithem:
**Bug report.**
parallelStream is used with the common pool, Tasks spawned from resurrectCheckpoints depend on fetchLayout, but if fetchLayout can't be scheduled because the common pool is exhausted, then a deadlock happens. This usually will happen if invalidateLayout is called during fast loading.

Issue #1687, opened 2019-01-22
fetchLayout deadlock 


Maithem:
**Bug report.** 
The CorfuRuntime relies on the common forkjoin pool to submit fetchLayout jobs, if the pool is exhausted with threads waiting on operations that the fetchLayout to succeed then a deadlock can occur. Where fetchLayout needs a thread to unblock other threads in the forkjoin pool, but the threads in the forkjoin pool are waiting on the fetchLayout to succeed.

PR #1689, opened 2019-01-23
Common Pool Deadlock

Maithem:
## Overview
This patch fixes a deadlock situation where the FastObjectLoader
exhausts the common thread pool with parallelStream and then a
invalidateLayout is issued. Since the latter call is scheduled
on the same thread pool, the FastObjectLoader will not progress
because it relies on fetchLayout to complete, but it can't.

Why should this be merged: Fixes a deadlock created by the FastObjectLoader

Related issue(s) (if applicable): #1688 and #1687

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
@@ -18,7 +18,10 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -632,8 +643,12 @@ private void findCheckPointsInLogAddress(long address, ILogData logData) {
      *
      */
     private void resurrectCheckpoints() {
-        streamsMetaData.entrySet().parallelStream()
-                .forEach(entry -> {
+        ExecutorService executorService = Executors.newFixedThreadPool(numOfWorkers, new ThreadFactoryBuilder()
```
StrikingThirteen: Maybe add a bit comment to the method doc highlighting the behavior that it throws FastObjectException if submit fails.

Maithem: Why would submit fail? Tasks are not submitted after shutdown is called

Maithem:
I've update this patch. Please review it again.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -632,8 +645,13 @@ private void findCheckPointsInLogAddress(long address, ILogData logData) {
...
+                .setNameFormat("FastObjectLoaderResurrectCheckpointsThread-%d").build());
+        CompletableFuture[] cfs = new CompletableFuture[streamsMetaData.size()];
+        int i = 0;
+        try {
+            for (Map.Entry<UUID, StreamMetaData> entry : streamsMetaData.entrySet()) {
+                cfs[i++] = CompletableFuture.runAsync(() -> {
```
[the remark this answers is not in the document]
Maithem: Good catch, I dropped it from the last PR.
Addressed.

PR #1694, opened 2019-01-24
@log-unit: Deal gracefully with partial-writes

vjeko:
On a partial write, a log unit would refuse to start due to a detected
data corruption. This patch changes this behaviour by simply discarding
partial written (corrupted) data and leaving it up to the client to
instantiate the repair protocol.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
@@ -13,6 +13,7 @@ import org.corfudb.comm.ChannelImplementation;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuRuntime.CorfuRuntimeParameters;
 import org.corfudb.runtime.exceptions.WrongEpochException;
+import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.ConservativeFailureHandlerPolicy;
 import org.corfudb.runtime.view.IReconfigurationHandlerPolicy;
 import org.corfudb.runtime.view.Layout;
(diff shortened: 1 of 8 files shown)


apavan:
> per batch write):

Do we have any tool that can be leveraged to run the benchmarks? OR any piece of code that could serve as starting point to run benchmarks?

Maithem:
@apavan any microbenchmark is fine. You can use JMH, but even a unit test is fine. 

I'm just worried that we will introduce latency per write, remember that delta increase is multiplied by the number of nodes in the chain. So if this change adds an extra 50 milliseconds then transactions will start observing an extra 150ms per transaction on a 3-node chain.

vjeko:
@Maithem Pavan was asking if we had any benchmarking in place or are we running blind?

vjeko:
@apavan @Maithem According to my benchmarks, the overhead is only visible for small writes is less than 2 milliseconds -- this marks the upper bound. The amortized cost is going to be less and it depends on the batch size -- in our case this translates to about 50 microseconds.

Maithem:
Can you upload your benchmark code. We need to consider different configurations.

vjeko:
Here you go. Instrument the BatchWriter. On the client side, you can do either rangeWrite/write.

```java
            while (true) {
                BatchWriterOperation currOp;

                if (lastOp == null) {
                    currOp = operationsQueue.take();
                } else {
                    currOp = operationsQueue.poll();

                    if (currOp == null || processed > BATCH_SIZE
                            || currOp == BatchWriterOperation.SHUTDOWN) {
                        long startTime = System.nanoTime();
                        streamLog.sync(sync);
                        streamLog.persistCommittedAddress();

                        long endTime = System.nanoTime();
                        long duration = (endTime - startTime);
                        points.add(duration);
                        System.out.println("Time taken " + duration/1000000);
                        System.out.println(processed + " Average: " + points.stream()
                                .mapToDouble(a -> a)
                                .average().getAsDouble()/1000000);

                        log.trace("Sync'd {} writes", processed);

                        for (BatchWriterOperation operation : res) {
                            handleOperationResults(operation);
                        }
                        res.clear();
                        processed = 0;
                    }
                }
```

Maithem:
Driver code? I want to run the same test.

vjeko:
Just do a rangeWrite/write from a client...

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -440,10 +442,30 @@ public long getStartingAddress() {
...
+        Long address = dataStore.get(Long.class, PREFIX_LOGUNIT, ADDRESS_JOURNAL);
+        return address == null ? Address.NON_ADDRESS : address;
+    }
+
+    /**
+     * Set the last globally committed address for this server context.
+     *
+     * @param address the address to set
+     */
+    public synchronized void setLastCommittedAddress(long address) {
```
medhavidhawan: The file based datastore was not built for high throughput. At least I never put in the effort to do make it efficient. I am worried that this could have some unintended affects on performance if the datastore is not performant enough.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -440,10 +442,30 @@ public long getStartingAddress() {
...
     public void setStartingAddress(long startingAddress) {
         dataStore.put(Long.class, PREFIX_STARTING_ADDRESS, KEY_STARTING_ADDRESS, startingAddress);
     }
 
+    /**
+     * Fetch last globally committed address for this server context.
+     *
+     * @return global address
+     */
+    public long getLastCommittedAddress() {
```
medhavidhawan: What is the expectation when a null is returned ?

annym:


```diff

Issues
======
+ Solved 22
- Added 11
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/CorfuDB/CorfuDB/pullRequest?prid=4653454&bid=15584345)

annym:


```diff

Issues
======
+ Solved 22
- Added 11
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651482&bid=15577795)

PR #1695, opened 2019-01-24
[Clean Rebase] Get Cluster Status API

annym:
Simplify and remove dependency with failure detector.
Introduce DB_SYNCING as a cluster state.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
@@ -1,18 +1,15 @@
 package org.corfudb.runtime.view;
 
-import com.google.common.collect.Sets;
-
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -319,125 +238,313 @@ private NodeStatus getLogUnitNodeStatusInLayout(Layout layout, String server) {
...
-            });
+        return connectivityStatusMap;
+    }
+
+    private Map<String, NodeStatus> getNodeStatusMap(Layout layout) {
+        Map<String, NodeStatus> nodeStatusMap = new HashMap<>();
+        for (String endpoint: layout.getAllServers()) {
+            if (layout.getUnresponsiveServers().contains(endpoint)) {
+                nodeStatusMap.put(endpoint, NodeStatus.DOWN);
+            } else if (layout.getSegments().size() != layout.getSegmentsForEndpoint(endpoint).size()) {
```
zalokhan: This is assuming that all the nodes in the layout are LogUnit servers.
This should be fine for now. But can you add a comment noting this assumption here?

annym: @zalokhan I have some minor comments to add on other things, if it's ok I'll address this along those in a separate PR so we can merge this soon and generate a build.

annym:
This PR was also reviewed by other members as it was part of an old branch.

PR #1699, opened 2019-01-25
Change log level in ChainReplicationProtocol for recover and propagate

StrikingThirteen:
## Overview

Description:
Changing the log level of recover and propagate in ChainReplicationProtocol to decrease the amount of logging.
Also tightening the visibility of recover and propagate.

Why should this be merged: 
Too many logs being emitted 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
@@ -1,21 +1,21 @@
 package org.corfudb.runtime.view.replication;
 
 import com.google.common.collect.Range;
-
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import javax.annotation.Nullable;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.Token;
 import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.exceptions.RecoveryException;
-import org.corfudb.runtime.view.RuntimeLayout;
 import org.corfudb.runtime.view.Layout;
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -143,13 +143,14 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
      */
-    protected void propagate(RuntimeLayout runtimeLayout,
+    private void propagate(RuntimeLayout runtimeLayout,
                              long globalAddress,
                              @Nullable ILogData data) {
         int numUnits = runtimeLayout.getLayout().getSegmentLength(globalAddress);
 
         for (int i = 1; i < numUnits; i++) {
-            log.info("Propogate[{}]: chain {}/{}", Token.of(runtimeLayout.getLayout().getEpoch(), globalAddress),
+            log.trace("Propogate[{}]: chain {}/{}", Token.of(runtimeLayout.getLayout().getEpoch(),
```
annym: Now that we're on this, can you please correct the typo in 'propagate' here and also in the catch clause..

StrikingThirteen: Done

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -166,7 +167,7 @@ protected void propagate(RuntimeLayout runtimeLayout,
                             .fillHole(token), OverwriteException.class);
                 }
             } catch (OverwriteException oe) {
-                log.trace("Propogate[{}]: Completed by other writer", globalAddress);
+                log.trace("Propagate[{}]: Completed by other writer", globalAddress);
```
Maithem: This is a rare event. I think we should make it INFO

annym: I agree with this, this should be at info level.

PR #1700, opened 2019-01-25
Fix a correcness issue with bulkWrite

vjeko:
BatchWriter does not wait for the CompletableFuture after doing
operationsQueue.add(...) during bulkWrite(...).
In some circumstances this could lead to data-loss.

- [ ] Public API has Javadoc


vjeko:
@Maithem are you talking about this pull request? https://github.com/CorfuDB/CorfuDB/pull/1694

This one deals just deals with the CompletableFuture.

vjeko:
@Maithem Unfortunately the workflow engine is pretty limited, and there is no way (as far know) to define a different type of retry policy. So, the only practical solution is to implement address detection.

vjeko:
@Maithem Can you please review the patch?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -953,6 +955,20 @@ public void append(List<LogData> range) {
...
+                .map(entry -> entry.getGlobalAddress()).collect(Collectors.toSet());
+
+        // See if the provided range overlaps with any of the previously written entries.
+        Set<Long> segOneOverlap = Sets.intersection(pendingWrite,
+                firstSh.getKnownAddresses().keySet());
+        Set<Long> segTwoOverlap = Sets.intersection(pendingWrite,
+                lastSh.getKnownAddresses().keySet());
+        if (!segOneOverlap.isEmpty() || !segTwoOverlap.isEmpty()) {
+            log.error("Overlapping addresses detected: {}, {}", segOneOverlap, segTwoOverlap);
+            throw new OverwriteException(OverwriteCause.NONE);
```
Maithem: I don't think this makes sense:
/** Indicates there is no actual cause for the overwrite, hence no overwrite exception has occurred. */
NONE(4);

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -953,6 +955,20 @@ public void append(List<LogData> range) {
         SegmentHandle firstSh = getSegmentHandleForAddress(first.getGlobalAddress());
         SegmentHandle lastSh = getSegmentHandleForAddress(last.getGlobalAddress());
 
+        // Extract all addresses associated with the provided write range.
+        Set<Long> pendingWrite = range.stream()
+                .map(entry -> entry.getGlobalAddress()).collect(Collectors.toSet());
+
```
Maithem: This is converting a list to a set, a duplicate within the same list won't be detected. Its a non-sorted set anyways, why not just do the overlap check on the List ?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
vjeko: The problem here is that there is really no suitable "cause" for this overwrite, mostly because of the N-to-1 mapping. Do we specify an overwrite cause for the first conflicting entry? Last? In any case, this will not give us any useful information.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
vjeko: I will add a check for duplicates.

There are two reasons for converting it to a set:
1) Relying on the input to be sorted is fragile. If that assumption is ever violated/changed, then this piece of the code will not work correctly.
2) It is conceptually easier to understand/more readable.

vjeko:
@Maithem This is an optimization we can deal with in a separate patch.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: I'm not saying we should rely on the List to be sorted, I was just pointing out that a linear scan won't be more expensive because the current set intersection logic is as expensive.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: You can actually map it to either a duplicate, same write, or a hole. It won't give us more information, but for the sake of consistency. 

You can actually remove the OverwriteCause enum, Its used for an optimization, which I don't think is a reasonable one.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
vjeko: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
vjeko: Done.

vjeko:
Can we please merge this?

PR #1702, opened 2019-01-26
Failure detector tune timeouts

xnull:
## Overview

Description:

Why should this be merged: 
Tune failure detector timeouts

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
@@ -55,7 +55,7 @@ public class FailureDetector implements IDetector {
      */
     @Getter
     @Setter
-    private long maxPeriodDuration = 8_000L;
+    private long maxPeriodDuration = 5_000L;
 
     /**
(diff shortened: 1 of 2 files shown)


apavan:
LGTM

apavan:
> Expose these through the server configs #1703
> 
> Also, change idleConnectionTimeout to 5 seconds
> https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java#L255

Maithem,
We can push that change with Slava's link failure PR. For now we will go with this as I tested 2 times with these settings and it passed both times. Better to keep it minimal in this attempt.

apavan:
@Maithem can you kick off Injection Framwork tests?

Maithem:
@apavan You can make it configurable later, but you also need to change the corfu runtime parameter defaults. They actually don't make sense and will break for cold shutdown on VMs.

apavan:
> @apavan You can make it configurable later, but you also need to change the corfu runtime parameter defaults. They actually don't make sense and will break for cold shutdown on VMs.

Let me give these setting a try and get back to you.

PR #1705, opened 2019-01-30
Add Log for failed authentication on server

annym:
## Overview

Log the IP for the client that failed to authenticate on server side.

Why should this be merged: whenever a client fails to connect due to certificate validation errors, we need the means to report from the server the IP of this failing client.


infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -187,6 +191,13 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
                 log.debug("exceptionCaught: Handshake timeout " +
                         "checker: discarded (handshake OK)");
             }
+        } else if (cause instanceof SSLHandshakeException) {
+            // If an SslException is thrown by the inbound SslHandler it will be caught by this
+            // upstream handler (in the pipeline). For debugging purposes we log the address of
+            // the client that failed to authenticate.
+
+            log.error("Client authentication failed for remote address {}",
+                    ctx.channel().remoteAddress());
```
xnull: What if the channel is closed? Will we get an exception?

xnull: looks like it will return null in any case. Looks ok

PR #1706, opened 2019-01-30
Fix getClusterStatus bug

annym:
Whenever the cluster status is unavailable from quorum nodes,
we should return the list of servers from the available layout(s)
instead of runtime layout servers list, as this last one might not
reflect all servers but only the ones used for runtime init.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
@@ -2,6 +2,7 @@ package org.corfudb.runtime.view;
 
 import java.time.Duration;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -408,6 +414,7 @@ private ClusterStatusReport getUnavailableClusterStatusReport(List<String> layou
                     .collect(Collectors.toList());
 
             if (!discoveredLayoutServers.isEmpty()) {
+                log.debug("Get layout from discovered layout servers: {} ", discoveredLayoutServers);
```
xnull: Please change log level to trace

PR #1717, opened 2019-02-06
Refresh Runtime Layout Servers

annym:
## Overview

Differentiate bootstrap layout servers (initial connection string)
from layout servers that are dynamically discovered from the cluster.
Also, refresh this latter list on every fetch layout.


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -788,7 +799,7 @@ public CorfuRuntime parseConfigurationString(String configurationString) {
      * @return A CorfuRuntime, to support the builder pattern.
      */
     public CorfuRuntime addLayoutServer(String layoutServer) {
-        layoutServers.add(layoutServer);
+        bootstrapLayoutServers.add(layoutServer);
```
zalokhan: Add to `layoutServers` here too?

PR #1719, opened 2019-02-07
Remove synchronized for cache eviction handler

cjqhenry14:
No need to use synchronized keyword for cache eviction
handler in LogUnitServer. The evictionHandler function
is already thread safe, both impls of StreamLog.release
method are no-op now.

Remove synchronized will improve the performance as it
will be triggered frequently when cache entries get evicted.


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -320,7 +320,7 @@ private synchronized ILogData handleRetrieval(long address) {
         return entry;
     }
 
-    private synchronized void handleEviction(long address, ILogData entry, RemovalCause cause) {
+    private void handleEviction(long address, ILogData entry, RemovalCause cause) {
```
StrikingThirteen: What if the implementations of release changes from no op?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -320,7 +320,7 @@ private synchronized ILogData handleRetrieval(long address) {
...
         log.trace("Eviction[{}]: {}", address, cause);
         streamLog.release(address, (LogData) entry);
```
Maithem: Release was removed, please rebase.
Currently this is only used for logging.

zalokhan:
Addressed by #1857

PR #1728, opened 2019-02-12
Fix partial bootstraps during AddNode.

### LINKED ISSUE #1725: Partial Bootstrap during Add Node Workflow
## Overview

Bootstrap Action consists of 2 parts: Layout Server bootstrap and Management Server bootstrap.
In case of a partial bootstrap, the layout server is bootstrap and then the action fails and retries.
On the retry, the layout server is already bootstrapped and hence the action is considered completed. This is incorrect as the management server is left un-bootstrapped forever.

zalokhan:
## Overview

Description: Fixes partial bootstraps in case of already bootstrapped components.

Why should this be merged: The management server remains un-bootstrapped forever.

Related issue(s) (if applicable): Fixes #1725 

Tests affecting this:
testPartialBootstrapNodeSuccess
testPartialBootstrapNodeFailure

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
@@ -14,6 +14,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.protocols.wireprotocol.LayoutPrepareResponse;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.AlreadyBootstrappedException;
 import org.corfudb.runtime.exceptions.NetworkException;
 import org.corfudb.runtime.exceptions.OutrankedException;
 import org.corfudb.runtime.exceptions.QuorumUnreachableException;
@@ -324,5 +325,29 @@ public class LayoutView extends AbstractView {
 
             log.debug("committed: Successful responses={}, timeouts={}", responses, timeouts);
         }
+    }
+
+    /**
+     * Bootstraps the layout server of the specified node.
+     * If already bootstrapped, it completes silently.
+     *
+     * @param endpoint Endpoint to bootstrap.
+     * @param layout   Layout to bootstrap with.
+     * @return Completable Future which completes with True when the layout server is bootstrapped.
+     */
+    CompletableFuture<Boolean> bootstrapLayoutServer(@Nonnull String endpoint, @Nonnull Layout layout) {
+        return getRuntimeLayout(layout).getLayoutClient(endpoint).bootstrapLayout(layout)
+                .exceptionally(throwable -> {
+                    try {
+                        CFUtils.unwrap(throwable, AlreadyBootstrappedException.class);
+                    } catch (AlreadyBootstrappedException e) {
+                        log.info("bootstrapLayoutServer: Layout Server {} already bootstrapped.", endpoint);
+                    }
+                    return true;
+                })
+                .thenApply(result -> {
+                    log.info("bootstrapLayoutServer: Layout Server {} bootstrap successful", endpoint);
+                    return true;
+                });
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
@@ -22,10 +22,12 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.clients.IClientRouter;
 import org.corfudb.runtime.clients.LayoutClient;
+import org.corfudb.runtime.exceptions.AlreadyBootstrappedException;
 import org.corfudb.runtime.exceptions.NetworkException;
 import org.corfudb.runtime.exceptions.WorkflowException;
 import org.corfudb.runtime.exceptions.WorkflowResultUnknownException;
 import org.corfudb.runtime.exceptions.WrongEpochException;
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.runtime.view.ClusterStatusReport.ClusterStatus;
 import org.corfudb.runtime.view.ClusterStatusReport.ClusterStatusReliability;
 import org.corfudb.runtime.view.ClusterStatusReport.ConnectivityStatus;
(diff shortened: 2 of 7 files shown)


xnull:
Is this PR still relevant?

zalokhan:
@xnull yes

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -325,4 +326,22 @@ public void committed(long epoch, Layout layout, boolean force)
...
+    /**
+     * Bootstraps the layout server of the specified node.
+     * If already bootstrapped, it completes silently.
+     *
+     * @param endpoint Endpoint to bootstrap.
+     * @param layout   Layout to bootstrap with.
+     */
+    public void bootstrapLayoutServer(@Nonnull String endpoint, @Nonnull Layout layout) {
+        try {
+            CFUtils.getUninterruptibly(getRuntimeLayout(layout)
```
xnull: Can you please use async computation. Just return completable future. Then compose this future with a future provided by `bootstrapManagementServer()` method.
If you do that you don't need to have a boilerplate block of code `try, catch`, you just need to handle an exception in one place

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -584,4 +585,22 @@ private Layout getHighestEpochLayout(Map<String, Layout> layoutMap) {
...
+    /**
+     * Bootstraps the management server if not already bootstrapped.
+     * If already bootstrapped, it completes silently.
+     *
+     * @param endpoint Endpoint ot bootstrap.
+     * @param layout   Layout to bootstrap with.
+     */
+    public void bootstrapManagementServer(@Nonnull String endpoint, @Nonnull Layout layout) {
+        try {
+            CFUtils.getUninterruptibly(runtime.getLayoutView().getRuntimeLayout(layout)
```
xnull: Same as prev comment

zalokhan:
@xnull I don't think `compose` would be helpful here.
The purpose of this PR is to catch exceptions on these calls separately and continue to the next call if required.
We cannot club the exception handling of these calls together.

xnull:
if `runtime.getLayoutView().bootstrapLayoutServer(endpoint, layout);` failed, why you ignore that fact and execute next operation `runtime.getManagementView().bootstrapManagementServer(endpoint, layout);` ?

xnull:
There is no excuse for using `get()` on a future. The more you use `get()` the more you have to make everything synchronous. CFUtils is a completely wrong implementation of an async workflow.
Why you even use futures if you block your current thread waiting for an operation completion, you just waste resources, just use your current thread then.

Please take a look at this change https://github.com/CorfuDB/CorfuDB/commit/e25255e6f825c4e3aa7bdd1862fc788305d72f5b

We have to execute `get()` on AddNodeWorkflow only once. If we make `IWorkflow` async, we don't even need to do that we just return the future.

zalokhan:
This doesn't make the flow asynchronous. This calls to both bootstrap methods are still executed one after the other and it should be that way. I feel it's similar to the current implementation.
I just feel the current implementation feels cleaner and simpler to debug.

xnull:
1. Without `get()` You block your current thread as late as possible. Actually, our endpoints start/end from netty and netty is non-blocking. If you are able to provide non-blocking result for your operation you can pass the result asynchronously to a netty endpoint. Our implementations completely synchronous everywhere that's why we can't do that right now and have to use get() in some methods. But you try to use get() everywhere which prevents us use truly non-blocking approach. If we decide to migrate to non-blocking API it will be extremely hard to get rid of all synchronous `CFUtils` calls in code. Let's develop things as it supposed to be used. It's a very bad practice to transform async operations to sync ones. I would say it's worst that we can do.
Imagine this:
   1. wrapSync(async)->wrapSync(async)->wrapSync(async)->wrapSync(async)->asyncResponce->client
   2. async->async->async->async->async->async->async->asyncResponce->client
Second approcah is non blocking, first one is not.
 
2. As I said it's a waste of resources for blocking the current thread until operation finished. Next step in the same method - you block the current thread again. And you do it again and again.

3. CFUtils seems pretty ugly if you try to follow the async approach, also `getUninterruptibly` is overcomplicated with wrapping and unwrapping exceptions which is absolutely unnecessary. It's hard to understand, scary to change and very complicated.

4. We generate boilerplate code because we always use `try, catch` for all CFUtils methods, it pollutes code. Very often you need to catch an exception on a very beginning of execution stack, and so on...

CFUtils breaks async workflow and that's very bad practice, let's just lear how to use completablefutures correctly.

test/src/test/java/org/corfudb/runtime/view/AbstractViewTest.java
```diff
@@ -94,9 +95,11 @@
     public AbstractViewTest() {
         // Force all new CorfuRuntimes to override the getRouterFn
         CorfuRuntime.overrideGetRouterFunction = this::getRouterFunction;
+        final Duration holeFillTimeout = Duration.ofSeconds(5);
         runtime = CorfuRuntime.fromParameters(CorfuRuntimeParameters.builder()
-            .nettyEventLoop(NETTY_EVENT_LOOP)
-            .build());
+                .nettyEventLoop(NETTY_EVENT_LOOP)
+                .holeFillTimeout(holeFillTimeout)
```
WenbinZhu: Should this change be moved to #1763 ?

zalokhan: Yes, I'll move it.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -325,4 +326,30 @@ public void committed(long epoch, Layout layout, boolean force)
...
+     * @param layout   Layout to bootstrap with.
+     * @return Completable Future which completes with True when the layout server is bootstrapped.
+     */
+    CompletableFuture<Boolean> bootstrapLayoutServer(@Nonnull String endpoint, @Nonnull Layout layout) {
+        return getRuntimeLayout(layout)
+                .getLayoutClient(endpoint)
+                .bootstrapLayout(layout)
+                .exceptionally(throwable -> {
+                    try {
+                        CFUtils.unwrap(throwable, AlreadyBootstrappedException.class);
```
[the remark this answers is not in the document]
zalokhan: No, it is the only exception that needs to be ignored.
The rest of the exceptions we do not need to handle and are rethrown.

test/src/test/java/org/corfudb/runtime/view/AbstractViewTest.java
```diff
@@ -94,9 +95,11 @@
     public AbstractViewTest() {
         // Force all new CorfuRuntimes to override the getRouterFn
         CorfuRuntime.overrideGetRouterFunction = this::getRouterFunction;
+        final Duration holeFillTimeout = Duration.ofSeconds(5);
```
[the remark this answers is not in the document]
zalokhan: Moved and will address the comment in #1763

Issue #1729, opened 2019-02-12
FollowBackpointer Logging


annym:
## Overview

When followBackpointers downgrades to single step due to holes, in scale systems logs can roll over very fast as each single stepped entry is logged on warning level.

PR #1730, opened 2019-02-12
Fix followBackpointers Logging on Single Step

annym:
## Overview

Change log level for single step on followBackpointers and in addition add log to indicate only the start and end of a single step range. 

Why should this be merged: fixes log roll over

Related issue(s) (if applicable): #1729 

- [x ] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x ] Change is covered by automated tests
- [ x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
@@ -238,12 +238,14 @@ public class BackpointerStreamView extends AbstractQueuedStreamView {
                                       final long startAddress,
                                       final long stopAddress,
                                       final Function<ILogData, BackpointerOp> filter) {
-        log.trace("followBackPointers: stmreadId[{}], queue[{}], startAddress[{}], stopAddress[{}]," +
+        log.trace("followBackpointers: streamId[{}], queue[{}], startAddress[{}], stopAddress[{}]," +
                 "filter[{}]", streamId, queue, startAddress, stopAddress, filter);
         // Whether or not we added entries to the queue.
         boolean entryAdded = false;
         // The current address which we are reading from.
         long currentAddress = startAddress;
+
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -286,23 +288,32 @@ protected boolean followBackpointers(final UUID streamId,
...
 
             if (singleStep) {
+                if (startingSingleStep) {
+                    startingSingleStep = false;
+                    log.info("followBackpointers[{}]: Found hole at address {}. Starting single step downgrade.", this, currentAddress);
+                }
                 // backpointers failed, so we're
                 // downgrading to a linear scan
-                log.warn("followBackPointers[{}]: downgrading to single step, found hole at {}", this, currentAddress);
+                log.trace("followBackpointers[{}]: downgrading to single step, found hole at {}", this, currentAddress);
```
zalokhan: Do we need `Trace`? Or you think `Debug` would be fine.

annym: @zalokhan I was actually not sure on this one. I was trying to be consistent with the other logging level/messages we have in this same method. For instance, printing the backpointer for each entry is on trace level, which is kind of the same we're doing here but for entries not belonging to this stream. But if there is a strong reason to move it to debug, that is fine also, can you think of any? The only thing I can think of, is the fact that we'd typically want to debug this and not change to trace level which brings many more data. Let me know what you think is better in this case.

zalokhan: I agree. Let's keep it in Trace then.

PR #1733, opened 2019-02-12
Fix followBackpointers Logging on Single Step

annym:
## Overview

Description:

Change logging level for single step on followBackpointers and add a start and end logging message to ease debugging. (Rebased on corfu-0.2.2)

Why should this be merged: logs roll over in short period for scale tests.

Related issue(s) (if applicable): #1729

- [ x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
@@ -238,12 +238,14 @@ public class BackpointerStreamView extends AbstractQueuedStreamView {
                                       final long startAddress,
                                       final long stopAddress,
                                       final Function<ILogData, BackpointerOp> filter) {
-        log.trace("followBackPointers: stmreadId[{}], queue[{}], startAddress[{}], stopAddress[{}]," +
+        log.trace("followBackpointers: streamId[{}], queue[{}], startAddress[{}], stopAddress[{}]," +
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -286,23 +288,32 @@ protected boolean followBackpointers(final UUID streamId,
...
-                log.trace("followBackPointers: backpointer points to {}", tmp);
+                log.trace("followBackpointers: backpointer points to {}", tmp);
                 // if backpointer is a valid log address or Address.NON_EXIST
                 // (beginning of the stream), do not single step back on the log
                 if (Address.isAddress(tmp) || tmp == Address.NON_EXIST) {
                     currentAddress = tmp;
                     singleStep = false;
+                    if (!startingSingleStep) {
+                        // A started single step period finishes here, refresh flag for next cycle.
+                        log.info("followBackpointers[{}]: Found backpointer for this stream at address {}. Stop single step downgrade.");
```
zalokhan: In a good case. This method will start with `startingSingleStep` set to false.
Will we log this statement every time this method is run? Won't that cause a lot of logging?

annym: @zalokhan It is set to true at the beginning, initially, it will only be set to false once the singleStep is actually started, which it is no longer 'starting' cause it started already.

PR #1735, opened 2019-02-14
Fix Runtime GC Bug

annym:
## Overview

Description: Whenever there is an ongoing transaction which snapshot (version) falls in the space of addresses that has already been trimmed from the log and the runtime GC kicks in, we might have "temporal loss" of data, given that the resolvedQueues are trimmed even when we have pointers in this space, meaning that the object loses all updates from this position until the trim mark. This is recovered on resets when the stream is loaded from the checkpoint. 

Why should this be merged: 'apparent' data loss.

- [X ] There are no TODOs left in the code
- [X ] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [X ] Change is covered by automated tests
- [X ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractStreamContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractStreamContext.java
@@ -1,6 +1,11 @@
 package org.corfudb.runtime.view.stream;
 
 import java.util.UUID;
+
+import lombok.Data;
+import lombok.Getter;
+import lombok.Setter;
+import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.view.Address;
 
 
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractStreamContext.java
```diff
@@ -25,11 +30,31 @@
...
+
     /**
      * A pointer to the current global address, which is the
      * global address of the most recently added entry.
      */
-    long globalPointer;
+    @Getter
+    private long globalPointer;
+
+    public void setGlobalPointerCheckGCTrimMark(long globalPointer) {
```
zalokhan: These can be Access-level protected.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
```diff
@@ -360,4 +363,182 @@ public void addNodeWithTrim() throws Exception {
...
+        t.join();
+
+        boolean exceptionCaught = false;
+        try {
+            final int snapshotTxAddress = 4;
+            corfuRuntime.getObjectsView().TXBuild().type(TransactionType.SNAPSHOT)
+                    .snapshot(new Token(0, snapshotTxAddress))
+                    .build().begin();
+            table.get(0);
+            corfuRuntime.getObjectsView().TXEnd();
```
zalokhan: You can put a `fail()` here and get rid of the exceptionCaught assertion.
Or you can look at  `assertThrownBy` and assert that it throws a transaction aborted exception with cause trim.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractStreamContext.java
annym: Done

test/src/test/java/org/corfudb/integration/WorkflowIT.java
annym: Done.

test/src/test/java/org/corfudb/integration/WorkflowIT.java
```diff
@@ -360,4 +355,273 @@ public void addNodeWithTrim() throws Exception {
...
+     * 8. On T0, prefixTrim @7
+     * 9. On T0, trigger runtimeGC
+     * 10. Attempt to access table, we should not have any data loss.
+     * 11. Initiate a snapshot transaction in the space of trimmed addresses, should be aborted.
+     * 12. Attempt to access table, we should not have any data loss.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testRuntimeGCForActiveTransactionsInTrimRangeMultiThread() throws Exception {
```
zalokhan: Great test.

PR #1736, opened 2019-02-16
Runtime gc fix corfu 0.2.2

annym:
## Overview

Description: Whenever there is an ongoing transaction which snapshot (version) falls in the space of addresses that has already been trimmed from the log and the runtime GC kicks in, we might have "temporal loss" of data, given that the resolvedQueues are trimmed even when we have pointers in this space, meaning that the object loses all updates from this position until the trim mark. This is recovered on resets when the stream is loaded from the checkpoint.

Why should this be merged: data loss


annym:
Already approved in PR #1735 based on master.

PR #1754, opened 2019-02-21
Avoid stateTransfer when adjacent segments have same number of servers

WenbinZhu:
## Overview

Description:

In some cases we could have multiple segments but all segments have same
number of servers. The current code will still start stateTransfer in
this case and result in unnecessary reads of data.

Why should this be merged: 

Avoid unnecessary reads which causes performance overheads

- [x] Public API has Javadoc


zalokhan:
@WenbinZhu Can you write a unit test for this?

WenbinZhu:
> @WenbinZhu Can you write a unit test for this?
Done.

PR #1757, opened 2019-02-22
Avoid wait after holeFill in range read.

zalokhan:
## Overview

Description:
Wait for only one address in range before hole fills instead of all addresses.
Change hole fills timeout from 5 to 10 seconds.

Why should this be merged: Time spent after acquiring the token and before writing was not considered in hole fill timeout.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
@@ -81,32 +81,32 @@ public class ChainReplicationProtocol extends AbstractReplicationProtocol {
      */
     @Override
     public Map<Long, ILogData> readAll(RuntimeLayout runtimeLayout, List<Long> globalAddresses) {
-        long startAddress = globalAddresses.iterator().next();
-        int numUnits = runtimeLayout.getLayout().getSegmentLength(startAddress);
-        log.trace("readAll[{}]: chain {}/{}", globalAddresses, numUnits, numUnits);
-
-        Map<Long, LogData> logResult = CFUtils.getUninterruptibly(
-                runtimeLayout
-                        .getLogUnitClient(startAddress, numUnits - 1)
-                        .read(globalAddresses)).getAddresses();
-
-        //in case of a hole, do a normal read and use its hole fill policy
-        Map<Long, ILogData> returnResult = new TreeMap<>();
-        for (Map.Entry<Long, LogData> entry : logResult.entrySet()) {
-            ILogData value = entry.getValue();
-            if (value == null || value.isEmpty()) {
-                value = read(runtimeLayout, entry.getKey());
-            }
-
-            returnResult.put(entry.getKey(), value);
-        }
-
-        return returnResult;
+        Range<Long> range = Range.encloseAll(globalAddresses);
+        return readRange(runtimeLayout, range, true);
     }
+
+
 
     @Override
     public Map<Long, ILogData> readRange(RuntimeLayout runtimeLayout, Set<Long> globalAddresses) {
         Range<Long> range = Range.encloseAll(globalAddresses);
+        return readRange(runtimeLayout, range, true);
+    }
+
+    /**
+     * Reads a range of addresses from the Chain of log unit servers. This method optimizes for the time to wait to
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -81,32 +81,19 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
-        return returnResult;
-    }
 
     @Override
     public Map<Long, ILogData> readRange(RuntimeLayout runtimeLayout, Set<Long> globalAddresses) {
         Range<Long> range = Range.encloseAll(globalAddresses);
+        return readDataInRange(runtimeLayout, range, true);
+    }
+
+    public Map<Long, ILogData> readDataInRange(RuntimeLayout runtimeLayout, Range<Long> range, boolean waitForWrite) {
```
medhavidhawan: could you please add comments to this method

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -193,6 +197,29 @@ protected ILogData read(final long address) {
             throw te;
         }
     }
+    final long HOLE_FILL_THRESHOLD = 6000;
```
medhavidhawan: I think this can come from the hole fill timeout parameter

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -193,6 +197,29 @@ protected ILogData read(final long address) {
...
+
+    protected ILogData readData(final long address, long startTime) {
```
medhavidhawan: it needs a comment can be made private method.

private ILogData read(final long address, long readStartTime)

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -117,12 +104,28 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
                         .getLogUnitClient(startAddress, numUnits - 1)
                         .read(range)).getAddresses();
 
-        //in case of a hole, do a normal read and use its hole fill policy
+        // In case of holes, use the standard backoff policy for hole fill for
+        // the first entry in the list. All subsequent holes in the list can
+        // be hole filled without waiting as we have already waited for the first
+        // hole.
+        boolean waitedForWrite = !waitForWrite;
```
medhavidhawan: rename waitedForWrite to maybe just wait

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
medhavidhawan: rename to readRange

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -267,6 +294,7 @@ protected boolean followBackpointers(final UUID streamId,
                                       final Function<ILogData, BackpointerOp> filter) {
         log.trace("followBackpointers: streamId[{}], queue[{}], startAddress[{}], stopAddress[{}]," +
                 "filter[{}]", streamId, queue, startAddress, stopAddress, filter);
+        long startTime = System.currentTimeMillis();
```
medhavidhawan: rename readStartTime

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
medhavidhawan: Comment:
Reads data  from an address in the address space. It will give the writer a chance to complete based on the time when the reads of which this individual read is a step started. If the reads have been going on for longer than the grace period given for a writer to complete a write, the subsequent individual read calls will immediately fill the hole on absence of data at the given address.

PR #1759, opened 2019-02-23
StreamLog refactoring/simplification, exception handling

xnull:
## Overview

Description:
 - If can't create stream log directory then throw UnrecoverableException 
 - extract StreamLogDataStore 
 - get  rid of unnecessary variables

Why should this be merged: 
Improving StreamLogFiles by better exception handling, improving/extracting StreamLogDataStore.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
@@ -322,7 +322,7 @@ public class CorfuServer {
                     router,
                     (String) opts.get("--address"),
                     port).channel().closeFuture().syncUninterruptibly();
-        } catch (Exception e) {
+        } catch (Throwable e) {
             log.error("CorfuServer: Server exiting due to unrecoverable error: ", e);
             System.exit(EXIT_ERROR_CODE);
         }
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java
@@ -1,11 +1,21 @@
 package org.corfudb.infrastructure;
 
+import static org.corfudb.infrastructure.utils.Persistence.syncDirectory;
+
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.CacheWriter;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.RemovalCause;
+import com.google.common.hash.Hasher;
+import com.google.common.hash.Hashing;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.exceptions.DataCorruptionException;
+import org.corfudb.util.JsonUtils;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
@@ -16,20 +26,6 @@ import java.nio.file.StandardOpenOption;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.function.Consumer;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-
-import com.google.common.hash.Hasher;
-import com.google.common.hash.Hashing;
-
-import lombok.extern.slf4j.Slf4j;
-import lombok.Getter;
-
-import org.corfudb.runtime.exceptions.DataCorruptionException;
-import org.corfudb.util.JsonUtils;
-
-import static org.corfudb.infrastructure.utils.Persistence.syncDirectory;
 
 /**
  * Stores data as JSON.
@@ -53,7 +49,7 @@ import static org.corfudb.infrastructure.utils.Persistence.syncDirectory;
 @Slf4j
 public class DataStore implements IDataStore {
 
-    static String EXTENSION = ".ds";
+    static final String EXTENSION = ".ds";
 
     @Getter
     private final Cache<String, Object> cache;
@@ -69,14 +65,13 @@ public class DataStore implements IDataStore {
     /**
      * Return a new DataStore object.
      *
-     * @param opts map of option strings
+     * @param opts        map of option strings
      * @param cleanupTask method to cleanup DataStore files
      */
     public DataStore(@Nonnull Map<String, Object> opts,
                      @Nonnull Consumer<String> cleanupTask) {
 
-        if ((opts.get("--memory") != null && (Boolean) opts.get("--memory"))
-                || opts.get("--log-path") == null) {
+        if ((opts.containsKey("--memory") && (Boolean) opts.get("--memory")) || !opts.containsKey("--log-path")) {
             this.logDirPath = null;
             this.cleanupTask = fileName -> { };
             cache = buildMemoryDs();
@@ -169,7 +164,7 @@ public class DataStore implements IDataStore {
 
     @Override
     public synchronized <T> void put(Class<T> tclass, String prefix, String key, T value) {
-        cache.put(getKey(prefix, key), value);
+        put(new KvRecord<>(prefix, key, tclass), value);
     }
 
     private <T> T load(Class<T> tClass, String key) {
@@ -204,10 +199,25 @@ public class DataStore implements IDataStore {
 
     @Override
     public synchronized <T> T get(Class<T> tclass, String prefix, String key) {
-        String path = getKey(prefix, key);
+        return get(new KvRecord<>(prefix, key, tclass));
+    }
+
+    @Override
+    public synchronized <T> void delete(Class<T> tclass, String prefix, String key) {
+        delete(new KvRecord<>(prefix, key, tclass));
+    }
+
+    @Override
+    public synchronized <T> void put(KvRecord<T> key, T value) {
+        cache.put(key.getFullKeyName(), value);
+    }
+
+    @Override
+    public synchronized <T> T get(KvRecord<T> key) {
+        String path = key.getFullKeyName();
         Object val = cache.get(path, k -> {
             if (!inMem) {
-                T loadedVal = load(tclass, path);
+                T loadedVal = load(key.getDataType(), path);
                 if (loadedVal != null) {
                     return loadedVal;
                 }
@@ -222,11 +232,13 @@ public class DataStore implements IDataStore {
     }
 
     @Override
-    public synchronized <T> void delete(Class<T> tclass, String prefix, String key) {
-        cache.invalidate(getKey(prefix, key));
+    public <T> T get(KvRecord<T> key, T defaultValue) {
+        T value = get(key);
+        return value == null ? defaultValue : value;
     }
 
-    private String getKey(String prefix, String key) {
-        return prefix + "_" + key;
+    @Override
+    public synchronized <T> void delete(KvRecord<T> key) {
+        cache.invalidate(key.getFullKeyName());
     }
 }
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
@@ -1,6 +1,9 @@
 package org.corfudb.infrastructure;
 
 
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
 /**
  * Key Value data store abstraction that provides persistence for variables that need
  * retain values across node restarts or need to be accessed by multiple modules/threads.
@@ -11,32 +14,72 @@ package org.corfudb.infrastructure;
  * <p>Created by mdhawan on 7/27/16.
  */
 public interface IDataStore {
+
+    @Deprecated
+    <T> void put(Class<T> tclass, String prefix, String key, T value);
+
+    @Deprecated
+    <T> T get(Class<T> tclass, String prefix, String key);
+
+    @Deprecated
+    <T> void delete(Class<T> tclass, String prefix, String key);
+
     /**
      * Stores a value for a key under a prefix (namespace).
      *
-     * @param tclass the class of the object being stored
-     * @param prefix namespace
-     * @param key    key-value key to store into
+     * @param key record meta information
      * @param value  Immutable value (or a value that won't be changed)
      */
-    public <T> void put(Class<T> tclass, String prefix, String key, T value);
+    <T> void put(KvRecord<T> key, T value);
 
     /**
      * Retrieves the value for a key under a prefix.
      *
-     * @param tclass the class of the object being retrieved
-     * @param prefix namespace
-     * @param key    key-value key to look up
+     * @param key record meta information
      * @return value stored under key
      */
-    public <T> T get(Class<T> tclass, String prefix, String key);
+    <T> T get(KvRecord<T> key);
+
+    /**
+     * Retrieves the value for a key or a default value
+     *
+     * @param key key meta info
+     * @param defaultValue a default value
+     * @param <T>
+     * @return
+     */
+    <T> T get(KvRecord<T> key, T defaultValue);
 
     /**
      * Deletes the value for a key under a prefix.
      *
-     * @param tclass the class of the object being retrieved
-     * @param prefix namespace
-     * @param key    key-value key to delete
+     * @param key record meta information
      */
-    public <T> void delete(Class<T> tclass, String prefix, String key);
+    <T> void delete(KvRecord<T> key);
+
+    /**
+     * Key-value meta information class, provides all the information for saving and getting data from a data store
+     *
+     * @param <T> data type
+     */
+    @AllArgsConstructor
+    @Getter
+    class KvRecord<T> {
+        /**
+         * namespace prefix for a key
+         */
+        private final String prefix;
+        /**
+         * key in a data store
+         */
+        private final String key;
+        /**
+         * The class of the value in a data store
+         */
+        private final Class<T> dataType;
+
+        public String getFullKeyName() {
+            return prefix + "_" + key;
+        }
+    }
 }
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
@@ -54,6 +54,7 @@ import java.util.regex.Pattern;
  */
 @Slf4j
 public class ServerContext implements AutoCloseable {
+
     // Layout Server
     private static final String PREFIX_EPOCH = "SERVER_EPOCH";
     private static final String KEY_EPOCH = "CURRENT";
@@ -66,10 +67,6 @@ public class ServerContext implements AutoCloseable {
     private static final String PREFIX_LAYOUTS = "LAYOUTS";
 
     // Sequencer Server
-    private static final String PREFIX_TAIL_SEGMENT = "TAIL_SEGMENT";
-    private static final String KEY_TAIL_SEGMENT = "CURRENT";
-    private static final String PREFIX_STARTING_ADDRESS = "STARTING_ADDRESS";
-    private static final String KEY_STARTING_ADDRESS = "CURRENT";
     private static final String KEY_SEQUENCER = "SEQUENCER";
     private static final String PREFIX_SEQUENCER_EPOCH = "EPOCH";
 
@@ -447,30 +444,6 @@ public class ServerContext implements AutoCloseable {
 
     public void setLayoutInHistory(Layout layout) {
         dataStore.put(Layout.class, PREFIX_LAYOUTS, String.valueOf(layout.getEpoch()), layout);
-    }
-
-    public long getTailSegment() {
-        Long tailSegment = dataStore.get(Long.class, PREFIX_TAIL_SEGMENT, KEY_TAIL_SEGMENT);
-        return tailSegment == null ? 0 : tailSegment;
-    }
-
-    public void setTailSegment(long tailSegment) {
-        dataStore.put(Long.class, PREFIX_TAIL_SEGMENT, KEY_TAIL_SEGMENT, tailSegment);
-    }
-
-    /**
-     * Returns the dataStore starting address.
-     *
-     * @return the starting address
-     */
-    public long getStartingAddress() {
-        Long startingAddress = dataStore.get(Long.class, PREFIX_STARTING_ADDRESS,
-                KEY_STARTING_ADDRESS);
-        return startingAddress == null ? 0 : startingAddress;
-    }
-
-    public void setStartingAddress(long startingAddress) {
-        dataStore.put(Long.class, PREFIX_STARTING_ADDRESS, KEY_STARTING_ADDRESS, startingAddress);
     }
 
     /**
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
@@ -1,0 +1,119 @@
+package org.corfudb.infrastructure.log;
+
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.infrastructure.IDataStore;
+import org.corfudb.infrastructure.IDataStore.KvRecord;
+import org.corfudb.runtime.view.Address;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Data access layer for StreamLog.
+ * Keeps stream log related meta information: startingAddress and tailSegment.
+ * Provides access to the stream log related meta information.
+ */
+@Builder
+@Slf4j
+public class StreamLogDataStore {
+    private static final String TAIL_SEGMENT_PREFIX = "TAIL_SEGMENT";
+    public static final String TAIL_SEGMENT_KEY = "CURRENT";
+
+    public static final String STARTING_ADDRESS_PREFIX = "STARTING_ADDRESS";
+    public static final String STARTING_ADDRESS_KEY = "CURRENT";
+
+    public static final KvRecord<Long> TAIL_SEGMENT_RECORD = new KvRecord<>(
+            TAIL_SEGMENT_PREFIX, TAIL_SEGMENT_KEY, Long.class
+    );
+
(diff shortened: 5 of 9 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -380,7 +395,9 @@ private LogData getLogData(LogEntry entry) {
      * @throws IOException IO exception
      */
     private Metadata parseMetadata(FileChannel fileChannel) throws IOException {
-        if (fileChannel.size() - fileChannel.position() < METADATA_SIZE) {
+        long actualMetaDataSize = fileChannel.size() - fileChannel.position();
+        if (actualMetaDataSize < METADATA_SIZE) {
+            log.debug("Meta data has wrong size. Actual size: {}, expected: {}", actualMetaDataSize, METADATA_SIZE);
             return null;
```
vjeko: Do you not want to short-circuit in this case?

infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java
```diff
@@ -204,10 +204,25 @@ public synchronized void delete(@Nonnull String key,
 
     @Override
     public synchronized <T> T get(Class<T> tclass, String prefix, String key) {
-        String path = getKey(prefix, key);
+        return get(new KvRecord<>(prefix, key, tclass));
+    }
+
+    @Override
+    public synchronized <T> void delete(Class<T> tclass, String prefix, String key) {
```
StrikingThirteen: I'm not sure these methods can carry their weights and they add complexity.

infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
```diff
@@ -11,32 +14,62 @@
  * <p>Created by mdhawan on 7/27/16.
  */
 public interface IDataStore {
+
+    @Deprecated
```
StrikingThirteen: Please avoid adding interface which is already deprecated.

infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
```diff
@@ -11,32 +14,62 @@
...
      * Retrieves the value for a key under a prefix.
      *
-     * @param tclass the class of the object being retrieved
-     * @param prefix namespace
-     * @param key    key-value key to look up
+     * @param record record meta information
      * @return value stored under key
      */
-    public <T> T get(Class<T> tclass, String prefix, String key);
+    <T> T get(KvRecord<T> record);
```
StrikingThirteen: Maybe KeyValueRecord

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -119,6 +120,21 @@ public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
         }
     }
 
+    /**
+     * Create stream log directory if not exists
+     */
+    private void initStreamLogDirectory() {
+        if (logDir.toFile().exists()) {
```
StrikingThirteen: Maybe log the fact the it already existed.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -228,22 +244,21 @@ public void prefixTrim(long address) {
...
     }
 
     private void verifyLogs() {
         String[] extension = {"log"};
-        File dir = new File(logDir);
+        File dir = logDir.toFile();
 
         if (!dir.exists()) {
-            return;
+            throw new UnrecoverableCorfuError("Stream log data directory doesn't exists");
```
StrikingThirteen: This is changing the behavior of verify. why?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -448,7 +465,7 @@ private LogHeader parseHeader(FileChannel channel) throws IOException {
         }
 
         if (Checksum.getChecksum(buffer.array()) != metadata.getPayloadChecksum()) {
-            throw new DataCorruptionException();
+            throw new DataCorruptionException("Invalid metadata check sum");
```
StrikingThirteen: checksum

infrastructure/src/test/java/org/corfudb/infrastructure/log/StreamLogDataStoreTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+        final int tailSegment = 333;
+        streamLogDs.saveTailSegment(tailSegment);
+        assertEquals(tailSegment, streamLogDs.getTailSegment());
+
+        final int startingAddress = 555;
+        streamLogDs.saveStartingAddress(startingAddress);
+        assertEquals(startingAddress, streamLogDs.getStartingAddress());
+    }
+}
```
StrikingThirteen: new line

infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java
xnull: They are used in many places (actually everywhere), I created new methods (which use KvRecord) and made these methods - deprecated.

infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
xnull: These methods used before, I made them deprecated but in the PR it looks as a just created them.  In reality, I added some new methods and deprecated these ones

infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
xnull: I like KV) it's concise

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: Because previously we could go to an invalid state. If the log directory doesn't exist corfu can't start, in the previous version it could happen.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: done

infrastructure/src/test/java/org/corfudb/infrastructure/log/StreamLogDataStoreTest.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: yep, I want to make it more explicit, in the prev version it was a bit unclear what means `fileChannel.size() - fileChannel.position();`

vjeko: Previously a null value was being returned. Now, we just continue on... Why are we continuing when we know the metadata is corrupt?

xnull: In `parseHeader()` we check if metadata is null. It's a bit different behaviour when it's null or when we throw an exception. Right now I am not ready to fix/simplify that logic. In next PR-s I will do it

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -129,7 +146,7 @@ public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
      * @return LogMetadata constructed from all the addresses in the
      * consecutive segments from [startSegment, endSegment]
      */
-    private LogMetadata initializeLogMetadata(long startSegment, long endSegment) {
+    private LogMetadata initializeLogMetadata(long startSegment, long endSegment, long startingAddress) {
```
zalokhan: Why do you need the startingAddress here?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -82,8 +88,6 @@
     // a new instance of this class should be created after deleting
     // the files of the old instance
     private LogMetadata logMetadata;
-    private long lastSegment;
-    private volatile long startingAddress;
```
zalokhan: What is the advantage of removing this?
Now all methods accessing these fields need to go via the datastore. We might pay the cost of IO calls.
Even if we argue that this is cached, the get is synchronized. This lock is shared among all components.
This can cause a performance impact. We have seent the datastore being a performance bottleneck in the past.

xnull: It can't cause a performance issue, because we use it only when we create a streamLog, trim it or release it. It happens pretty rear so it just can't affect performance. You can also see that we read and save the variable same number of times in prev version and the current version. I don't see any reasons to overcomplicate logic and have those variables. 

The advantage is - it comes more consistent. You use only your datastore to get and update those variables, it's synchronized, it guarantees ordering and so on.  Also in prev version, you could have the variable is not equals to the value in the datastore, in current version we improved that.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: Because we already read it from the dataStore, we just pass it as a parameter we don't need to read it form the datastore second time

infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
StrikingThirteen: ok

infrastructure/src/main/java/org/corfudb/infrastructure/IDataStore.java
StrikingThirteen: yeah it looks like that. ok then.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
StrikingThirteen: even rare might have effects on the performance though.

xnull: Trimming happens every... how long? 15 minutes. It just can't

zalokhan: I don't agree with this.
This call would be made in `isTrimmed()`. This is called on every operation - read and append.

1. Do not introduce performance or behaviour changes in a cleanup PR.
2. If you still want to change this then create the same volatile variables in the StreamLogDatastore and cache them and update them there.

As I said before, since the synchronized lock is shared by all servers, there can be contention on the lock causing performance degradation.

xnull: oh, you are right, will return startingAddress variable then

xnull: startingAddress and tailSegment moved to StreamLogDataStore, added a a test

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
```diff
@@ -2,31 +2,65 @@
...
+    private final AtomicLong tailSegment = new AtomicLong(Address.NON_ADDRESS);
+
+    /**
+     * Return current cached tail segment or get the segment from the data store if not initialized
+     * @return tail segment
+     */
     public long getTailSegment() {
-        return Optional.ofNullable(dataStore.get(TAIL_SEGMENT_RECORD)).orElse(0L);
+        if (tailSegment.get() == Address.NON_ADDRESS) {
+            long dbTailSegment = Optional.ofNullable(dataStore.get(TAIL_SEGMENT_RECORD)).orElse(0L);
```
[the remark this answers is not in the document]
xnull: Actially no, NON_ADDRESS equals to -1, ZERO_ADDRESS equals to 0
This is correct.
If cached value is undefined (equals to NON_ADDRESS) we get the value from the datastore and update the value

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -322,7 +322,7 @@ public static void startServer(Map<String, Object> opts, boolean bindToAllInterf
                     router,
                     (String) opts.get("--address"),
                     port).channel().closeFuture().syncUninterruptibly();
-        } catch (Exception e) {
+        } catch (Throwable e) {
```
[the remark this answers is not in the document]
xnull: Because we close (System.exit()) the application after this catch, if you don't catch Throwable you miss it in logs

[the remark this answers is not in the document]
xnull: Also the application will be closed anyway
if you don't catch throwable you can miss some exceptions in the log

infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java
```diff
@@ -23,6 +23,7 @@
 import com.google.common.hash.Hasher;
 import com.google.common.hash.Hashing;
 
+import lombok.AllArgsConstructor;
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -6,7 +6,9 @@
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.netty.channel.EventLoopGroup;
+import lombok.AllArgsConstructor;
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -6,7 +6,9 @@
...
 import lombok.Getter;
+import lombok.NonNull;
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -31,6 +33,7 @@
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+
+        return startingAddress.get();
+    }
+
+    /**
+     * Update current starting address in the data store
+     * @param newStartingAddress updated starting address
+     */
+    public void updateStartingAddress(long newStartingAddress) {
+        dataStore.put(newStartingAddress, STARTING_ADDRESS_RECORD);
```
annym: at log entry at trace level

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Data access layer for StreamLog
+ */
+@Builder
+@Slf4j
+public class StreamLogDataStore {
+    public static final KvRecord<Long> TAIL_SEGMENT_RECORD = new KvRecord<>("TAIL_SEGMENT", "CURRENT", Long.class);
+
+    public static final KvRecord<Long> STARTING_ADDRESS_RECORD = new KvRecord<>(
+            "STARTING_ADDRESS", "CURRENT", Long.class
```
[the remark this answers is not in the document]
xnull: don't get your question

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -94,23 +98,20 @@
...
-        this.serverContext = serverContext;
+        this.dataStore = StreamLogDataStore.builder().dataStore(serverContext.getDataStore()).build();
+
+        initStreamLogDirectory();
         verifyLogs();
         // Starting address initialization should happen before
         // initializing the tail segment (i.e. initializeMaxGlobalAddress)
-        startingAddress = serverContext.getStartingAddress();
+        long startingAddress = dataStore.getStartingAddress();
         long firstSegment = startingAddress / RECORDS_PER_LOG_FILE;
```
Maithem: You can have a helper method to get a segment number for an address.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -94,23 +98,20 @@
...
         this.verify = !noVerify;
-        this.serverContext = serverContext;
+        this.dataStore = StreamLogDataStore.builder().dataStore(serverContext.getDataStore()).build();
+
+        initStreamLogDirectory();
         verifyLogs();
         // Starting address initialization should happen before
         // initializing the tail segment (i.e. initializeMaxGlobalAddress)
-        startingAddress = serverContext.getStartingAddress();
+        long startingAddress = dataStore.getStartingAddress();
```
Maithem: Why did you drop the volatile from startingAddress?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: If you are passing startingAddress, then you can derive startSegment from it as well. Maybe just have initializeLogMetadata(long startingAddress, long endSegment)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Returns the dataStore starting address.
+     *
+     * @return the starting address
+     */
+    public long getStartingAddress() {
+        if (startingAddress.get() == Address.NON_ADDRESS) {
+            long dbStartingAddress = Optional.ofNullable(dataStore.get(STARTING_ADDRESS_RECORD)).orElse(ZERO_ADDRESS);
```
Maithem: can you change this name, let's not use db, maybe use trimMark.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1108,10 +1124,8 @@ public void reset() {
             }
         });
 
-        serverContext.setStartingAddress(0L);
-        serverContext.setTailSegment(0L);
-        startingAddress = 0L;
-        lastSegment = 0L;
+        dataStore.resetStartingAddress();
```
Maithem: Why don't you just delete and recreate the log dir?

xnull: will do it in next PR

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.infrastructure.IDataStore;
+import org.corfudb.infrastructure.IDataStore.KvRecord;
+import org.corfudb.runtime.view.Address;
+
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Data access layer for StreamLog
```
Maithem: This class comment is not very accurate. Please add more info about how this only stores metadata about log markers. Also, why isn't this on the LogUnit level (i.e. LogUnitMetaDataStore)?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: I don't use `startingAddress` and `tailSegment` in StreamLogFiles, I moved them to StreamLogDataStore. They are volatile in StreamLogDataStore

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -119,6 +120,22 @@ public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
...
+     * Create stream log directory if not exists
+     */
+    private void initStreamLogDirectory() {
+        if (logDir.toFile().exists()) {
+            log.info("Log directory already exists: {}", logDir);
+            return;
+        }
+
+        try {
+            Files.createDirectories(logDir);
```
Maithem: I think you need to fsync the parent directory after you create the log directory (i.e use Persistence.syncDirectory) , or maybe just the directory itself since we "only" support ext4: 

https://github.com/torvalds/linux/blob/master/fs/ext4/fsync.c#L122

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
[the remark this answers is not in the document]
StrikingThirteen: I was referring to 0L:

 if (tailSegment.get() == Address.NON_ADDRESS) {
            long dbTailSegment = Optional.ofNullable(dataStore.get(TAIL_SEGMENT_RECORD)).orElse(**0L**);

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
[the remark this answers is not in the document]
StrikingThirteen: ok

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
xnull: Added comments.

P.S. this DS manages only stream log related variables, I believe it's not quite right to use for LogUnit purposes.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: Thanks, will apply it in the next PR

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+     * Update current tail segment in the data store
+     * @param newTailSegment updated tail segment
+     */
+    public void updateTailSegment(long newTailSegment) {
+        if (tailSegment.get() >= newTailSegment) {
+            log.debug("New tail segment equals to the old one: {}. Ignore", newTailSegment);
+            return;
+        }
+
+        dataStore.put(newTailSegment, TAIL_SEGMENT_RECORD);
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+        return tailSegment.get();
+    }
+
+    /**
+     * Update current tail segment in the data store
+     * @param newTailSegment updated tail segment
+     */
+    public void updateTailSegment(long newTailSegment) {
+        if (tailSegment.get() >= newTailSegment) {
+            log.debug("New tail segment equals to the old one: {}. Ignore", newTailSegment);
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogDataStore.java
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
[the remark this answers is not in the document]
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java
```diff
@@ -204,10 +204,25 @@ public synchronized void delete(@Nonnull String key,
...
+        return get(new KvRecord<>(prefix, key, tclass));
+    }
+
+    @Override
+    public synchronized <T> void delete(Class<T> tclass, String prefix, String key) {
+        delete(new KvRecord<>(prefix, key, tclass));
+    }
+
+    @Override
+    public synchronized <T> void put(T value, KvRecord<T> record) {
```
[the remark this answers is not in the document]
xnull: done

Issue #1743, opened 2019-02-19
Lots of "remove request" in NettyClientRouter


WenbinZhu:
## Overview

A lot of "remove request" are seen in server logs in some cases:

```
Remove request 164181 to tcp://10.115.76.124:9000/ due to timeout! Message:CorfuMsg(clientID=64acd3c9-4b65-431d-9e4b-eeed984a07bf, requestID=164181, epoch=0, buf=null, msgType=PING)
2019-02-15T06:46:50.842Z | [39mDEBUG[0;39m |                      client-12 |      o.c.r.c.NettyClientRouter | Remove request 2215 to tcp://10.115.76.123:9000/ due to timeout! Message:CorfuMsg(clientID=43457d69-74d5-4e59-b80d-3a9974877b8c, requestID=2215, epoch=0, buf=null, msgType=PING)
2019-02-15T06:46:50.844Z | [39mDEBUG[0;39m |                      client-19 |      o.c.r.c.NettyClientRouter | Remove request 1040 to tcp://10.115.76.124:9000/ due to timeout! Message:CorfuMsg(clientID=b62f1e7a-db0a-41c3-8ab9-432a452c377a, requestID=1040, epoch=0, buf=null, msgType=PING)
2019-02-15T06:46:50.844Z | [34mINFO [0;39m |                      client-20 |      o.c.r.c.NettyClientRouter | connectAsync[tcp://10.115.76.122:9000/]: Channel connected.
2019-02-15T06:46:50.844Z | [34mINFO [0;39m |                      client-20 | o.c.p.w.ClientHandshakeHandler | channelActive: Outgoing connection established to: /10.115.76.122:9000
2019-02-15T06:46:50.844Z | [34mINFO [0;39m |                      client-20 | o.c.p.w.ClientHandshakeHandler | channelActive: Initiate handshake. Send handshake message.
2019-02-15T06:46:50.844Z | [39mDEBUG[0;39m |                      client-20 | c.p.w.NettyCorfuMessageEncoder | encode: New max write buffer found 69
2019-02-15T06:46:50.844Z | [39mDEBUG[0;39m |                      client-20 | o.c.p.w.ClientHandshakeHandler | channelActive: Add readTimeoutHandler to channel pipeline.
[...]

PR #1762, opened 2019-02-27
Make Netty keep alive message ignore epoch, fix Future handling in NettyClientRouter

WenbinZhu:
## Overview

Description:

- Create a new KeepAlive message which is basically a ping but ignores epoch.
- Fix CompletableFuture exception handling in NettyClientRouter when the exception
is not due to timeout, which results in handling same exception twice.
- Removed client side SO_KEEPALIVE socket option since we already have keep alive mechanism in client side.

Why should this be merged: 

Fixed two issues:
- Keep alive message gets WrongEpochException
- Same CompletableFuture exception handled twice in NettyClientRouter

Related issue(s) (if applicable): #1743

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
@@ -50,8 +50,21 @@ public class BaseServer extends AbstractServer {
      * @param r     The server router.
      */
     @ServerHandler(type = CorfuMsgType.PING)
-    private static void ping(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+    private void ping(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
         r.sendResponse(ctx, msg, CorfuMsgType.PONG.msg());
+    }
+
+    /**
+     * Respond to a keep alive message.
+     * Note: this message ignores epoch.
+     *
+     * @param msg   The incoming message
+     * @param ctx   The channel context
+     * @param r     The server router.
+     */
+    @ServerHandler(type = CorfuMsgType.KEEP_ALIVE)
+    private void keepAlive(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+        r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
     }
 
     /**
@@ -84,7 +97,7 @@ public class BaseServer extends AbstractServer {
             long epoch = msg.getPayload();
             log.info("handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch {}", epoch);
             serverContext.setServerEpoch(epoch, r);
-            r.sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.ACK));
+            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
         } catch (WrongEpochException e) {
             log.debug("handleMessageSetEpoch: Rejected SET_EPOCH current={}, requested={}",
                     e.getCorrectEpoch(), msg.getPayload());
--- a/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
@@ -460,9 +460,15 @@ public class NettyClientRouter extends SimpleChannelInboundHandler<CorfuMsg>
         final CompletableFuture<T> cfTimeout =
             CFUtils.within(cfBenchmarked, Duration.ofMillis(timeoutResponse));
         cfTimeout.exceptionally(e -> {
-            outstandingRequests.remove(thisRequest);
-            log.debug("Remove request {} to {} due to timeout! Message:{}",
-                    thisRequest, node, message);
+            // CFUtils.within() can wrap different kinds of exceptions in
+            // CompletionException, just dealing with TimeoutException here since
+            // the router is not aware of it and this::completeExceptionally()
+            // takes care of others. This avoids handling same exception twice.
+            if (e.getCause() instanceof TimeoutException) {
+                outstandingRequests.remove(thisRequest);
+                log.debug("sendMessageAndGetCompletable: Remove request {} to {} due to timeout! Message:{}",
+                        thisRequest, node, message);
+            }
             return null;
         });
         return cfTimeout;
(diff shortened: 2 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -524,9 +530,8 @@ public void sendResponseToServer(ChannelHandlerContext ctx, CorfuMsg inMsg, Corf
      */
     public void completeExceptionally(long requestID, Throwable cause) {
         CompletableFuture cf;
-        if ((cf = outstandingRequests.get(requestID)) != null) {
+        if ((cf = outstandingRequests.remove(requestID)) != null) {
```
StrikingThirteen: If cause can be null, then this changes the behavior by removing request first before completeExceptionally. if cause can not be null then add @Nonnull annotation.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -585,11 +590,11 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
      */
     private void keepAlive() {
         if (!channel.isOpen()) {
-            log.warn("keepAlive: channel not open, skipping ping. ");
+            log.warn("keepAlive: channel not open, skipping sending keep alive.");
             return;
         }
-        sendMessageAndGetCompletable(null, new CorfuMsg(CorfuMsgType.PING));
-        log.trace("keepAlive: sending ping to {}", this.channel.remoteAddress());
+        sendMessageAndGetCompletable(null, CorfuMsgType.KEEP_ALIVE.msg());
```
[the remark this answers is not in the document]
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -50,10 +50,23 @@ public BaseServer(@Nonnull ServerContext context) {
...
 
+    /**
+     * Respond to a keep alive message.
+     * Note: this message ignores epoch.
+     *
+     * @param msg   The incoming message
+     * @param ctx   The channel context
+     * @param r     The server router.
+     */
+    @ServerHandler(type = CorfuMsgType.KEEP_ALIVE)
```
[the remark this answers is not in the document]
WenbinZhu: Keep alive seems more descriptive since this is the purpose of this message. Also we already have HEARTBEAT_REQUEST/HEARTBEAT_RESPONSE, use HEART_BEAT again brings confusion.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
WenbinZhu: The caller of this is simply catching exception and passing it to this method, so cause is not null. Added @Nonnull annotation.

xnull: `
if(outstandingRequests.contains(...)){
    ...
}
` ?

WenbinZhu: @xnull remove is simpler, if you use contains, you need to call remove again.

xnull: up to you but it looks awkward

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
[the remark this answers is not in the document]
Maithem: Why do you need to introduce a new message? Why not just use ping ?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -460,9 +460,15 @@ public void stop(boolean shutdown) {
...
-            outstandingRequests.remove(thisRequest);
-            log.debug("Remove request {} to {} due to timeout! Message:{}",
-                    thisRequest, node, message);
+            // CFUtils.within() can wrap different kinds of exceptions in
+            // CompletionException, just dealing with TimeoutException here since
+            // the router is not aware of it and this::completeExceptionally()
+            // takes care of others. This avoids handling same exception twice.
+            if (e.getCause() instanceof TimeoutException) {
+                outstandingRequests.remove(thisRequest);
+                log.debug("sendMessageAndGetCompletable: Remove request {} to {} due to timeout! Message:{}",
```
Maithem: Is it possible to have an exception that prevents the TimeoutException to happen, which results in leaking requests in outstandingRequests?

Maithem: Just log the message + exception (so that all exceptions are logged)

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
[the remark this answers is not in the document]
WenbinZhu: @Maithem Ping is epoch aware, and being used getOrchestrator() and getClusterStatus(). This new message ignores epoch since keep alive doesn't care epoch. Otherwise using ping for keep alive could result in WrongEpochException.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
WenbinZhu: In case of exceptions other than TimeoutException, `completeExceptionally()` will handle it and remove it from outstandingRequests. https://github.com/CorfuDB/CorfuDB/blob/1d6c4a1a70f91b5315fa6162a7cdce558d4bb339/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java#L530

WenbinZhu: Done in completeExceptionally().

WenbinZhu:
Removed client side SO_KEEPALIVE socket option since we already have keep alive mechanism in client side.

xnull:
@WenbinZhu why you don't merge your PR-s? It's already four

WenbinZhu:
> @WenbinZhu why you don't merge your PR-s? It's already four

Updated branch, will merge today after travis and injection framework are passed

PR #1766, opened 2019-03-01
Make NodeLocator equivalent to the old format host:port

xnull:
## Overview

Description:
Allows using `NodeLocator` for node names instead of strings.

Why should this be merged: 
Improves type safety for naming nodes. Enables us to use `clientId` and differentiate between Corfu runtime clients. Makes connectin string more standardized.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/util/NodeLocator.java
```diff
@@ -83,55 +109,29 @@ public static NodeLocator parseString(String toParse) {
...
+                    .options(options)
+                    .build();
 
         } catch (URISyntaxException m) {
             throw new IllegalArgumentException(m);
         }
     }
 
+    public String toEndpointUrl(){
+        return host + ":" + port;
```
Maithem: Why not create this during construction?

xnull: Don't want to create an extra variable

Maithem: I suppose this shouldn't generate too much garbage.

Issue #1760, opened 2019-02-27
Sequencer Concurrency Bug


Maithem:
**Bug report.**
The sequencer's cache can be accessed concurrently by the RPC thread and the cache's background thread (via common thread pool). This can introduce tricky correctness race conditions when resolving transactions.

PR #1770, opened 2019-03-07
Sequencer cache correction

xnull:
## Overview
Since caffeine cache has a thread (for maintenance purposes) and eviction works in parallel it can cause race condition issues, which can allow finishing transaction with conflict state.
To prevent that 

Description:
Fix inconsistency issue between caffeine cache and sequencer server. 
Caffeine - get rid of an extra thread, eviction works in the same thread as Sequencer server.
Extract SequencerServerCache, write a unit test. 

Why should this be merged: 
Fix inconsistency issue for conflict transactions

Related issue(s) (if applicable): #1760

- [x] Public API has Javadoc


ben-manes:
Quick question -

From my naive understanding, it looks like it was a race due to multiple `RemovalListener` tasks being executing asynchronously? By switching to the `CacheWriter`, you are ensuring only the thread that is evicting (under an internal cache lock) runs the delete operation. The problem isn't the executor, but that `maxConflictWildcard` was not being operated on atomically in a safe manner. It looks like that field can still be operated on unsafely, e.g. multiple `invalidate` calls may invoke `CacheWriter#delete` at the same time and `#delete` may race with `updateMaxConflictAddress`.

I very much don't know how things should work, but a quick glance leads me to suspect that you reduced the problem but still have a race. The executor exacerbated it, but wasn't the cause.

P.S. It would also be great if you upgrade to v2.7, since we made a lot of improvements since the one you use (v2.4).

xnull:
>>The problem isn't the executor, but that maxConflictWildcard was not being operated on atomically in a safe manner. 
@ben-manes yes, you are right. We don't need to update `maxConflictWildcard` and the cache at the same time, we just need to update `maxConflictWildcard` before we delete a value from the cache. 
So, it should work correctly.
Regarding `updateMaxConflictAddress` we don't need to update it during the eviction, we only need to execute `updateMaxConflictAddress` during reset_sequencer, it also works correctly.

Thanks, for your comment

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+    public SequencerServerCache(long cacheSize) {
+        CacheWriter<String, Long> writer = new CacheWriter<String, Long>() {
+            @Override
+            public void write(@Nonnull String key, @Nonnull Long value) {
+                //ignore
+            }
+
+            @Override
+            public void delete(@Nonnull String key, @Nullable Long globalAddress, @Nonnull RemovalCause cause) {
+                if (cause == RemovalCause.REPLACED) {
```
Maithem: throw an exception ?

xnull: Not sure. It was in the prev version, I wouldn't like to change the behaviour in this case

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+                if (globalAddress == null) {
+                    globalAddress = Address.NOT_FOUND;
+                }
+                maxConflictWildcard = Math.max(globalAddress, maxConflictWildcard);
+            }
+        };
+
+        this.conflictToGlobalTailCache = Caffeine.newBuilder()
+                .maximumSize(cacheSize)
+                .executor(Runnable::run)
```
Maithem: Can you add a comment about how this "trick" makes the cache use the accessing thread and not a background thread.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+        log.debug("Invalidate sequencer cache. Trim mark: {}", trimMark);
+
+        long entries = 0;
+        for (String key : conflictToGlobalTailCache.asMap().keySet()) {
+            Long currTrimMark = getIfPresent(key);
+            if (currTrimMark >= trimMark) {
+                continue;
+            }
+
+            conflictToGlobalTailCache.invalidate(key);
```
Maithem: The sequencer is single threaded, why not just use remove? I think invalidate schedules the removal for another time.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+                .writer(writer)
+                .recordStats()
+                .build();
+    }
+
+    public Long getIfPresent(String conflictKeyHash) {
+        return conflictToGlobalTailCache.getIfPresent(conflictKeyHash);
+    }
+
+    public void invalidate(long trimMark) {
```
Maithem: maybe rename this to invalidateUpToTrimMark and please add a comment.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: `Invalidate` is always synchronous

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+        return conflictToGlobalTailCache.estimatedSize();
+    }
+
+    public void put(String conflictHashCode, long newTail) {
+        conflictToGlobalTailCache.put(conflictHashCode, newTail);
+    }
+
+    public void invalidate() {
+        log.info("Invalidate sequencer server cache");
+        conflictToGlobalTailCache.invalidateAll();
```
Maithem: Why not just use map.clear() ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: invalidate == revome

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public void updateMaxConflictAddress(long newMaxConflictWildcard) {
+        log.info("updateMaxConflictAddress, new address: {}", newMaxConflictWildcard);
```
Maithem: I would make this debug level, once the cache fills, it will print this statement on every transaction commit.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: The cache provides you an API for that, which is `invalidateAll`.
Use `conflictToGlobalTailCache.asMap().clear()` seems strange to me

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: We update it only during `resetServer`

xnull: Which happens not so often

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: Why repeat same thing twice? We invalidate the cache, also we see trim mark in a parameter. Making method names more verbose trying to explain what they do, it seems not so good idea to me. Also, we can explain details in javadoc

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -385,8 +336,8 @@ public synchronized void trimCache(CorfuPayloadMsg<Long> msg,
     public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg> msg,
                                          ChannelHandlerContext ctx, IServerRouter r) {
         long initialToken = msg.getPayload().getGlobalTail();
-        final Map<UUID, Long> streamTails = msg.getPayload().getStreamTails();
-        final long bootstrapMsgEpoch = msg.getPayload().getSequencerEpoch();
+        Map<UUID, Long> streamTails = msg.getPayload().getStreamTails();
+        long bootstrapMsgEpoch = msg.getPayload().getSequencerEpoch();
```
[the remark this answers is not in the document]
xnull: I believe using `final` inside methods is overhead, it brings boilerplate code and doesn't give any advantages

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: done

> Slava, please follow the code guidelines! The line break is at 100 not 120.
> 
> Also, you need a lot more documentation, especially WRT concurrency. I assume the secret ingredient is ".executor(Runnable::run)".

Line length 120 - https://github.com/CorfuDB/CorfuDB/blob/5540d594eba3b7adfed1eb9eba505ce90cc8be40/checks.xml#L88

P.S. Will add some doc

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: a comment has added

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -145,41 +107,29 @@
      * Handler for this server.
      */
     @Getter
-    private final CorfuMsgHandler handler =
-            CorfuMsgHandler.generateHandler(MethodHandles.lookup(), this);
+    private final CorfuMsgHandler handler = CorfuMsgHandler.generateHandler(MethodHandles.lookup(), this);
```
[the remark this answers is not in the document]
xnull: it's 120, see my prev comment

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+                );
+
+                if (globalAddress == null) {
+                    globalAddress = Address.NOT_FOUND;
+                }
+                maxConflictWildcard = Math.max(globalAddress, maxConflictWildcard);
+            }
+        };
+
+        this.conflictToGlobalTailCache = Caffeine.newBuilder()
```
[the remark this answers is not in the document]
xnull: We don't. we create a new one in the constructor. What could confuse you is that we have two constructors in each one we create a new instance.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+                .maximumSize(cacheSize)
+                .executor(Runnable::run)
+                .writer(writer)
+                .recordStats()
+                .build();
+    }
+
+    public Long getIfPresent(String conflictKeyHash) {
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            conflictToGlobalTailCache.invalidate(key);
+            entries++;
+        }
+        log.info("Evicted entries: {}", entries);
+
+    }
+
+    public long size() {
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+        return conflictToGlobalTailCache.estimatedSize();
+    }
+
+    public void put(String conflictHashCode, long newTail) {
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+        conflictToGlobalTailCache.put(conflictHashCode, newTail);
+    }
+
+    public void invalidate() {
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,133 @@
...
+        log.info("Invalidate sequencer server cache");
+        conflictToGlobalTailCache.invalidateAll();
+    }
+
+    public void updateMaxConflictAddress(long newMaxConflictWildcard) {
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: After some discussion, we decided to throw IllegalStateException.
Done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
vjeko: Slava, please take a look at the official style guideline: https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines

The line wrap is at 100: "Limit line length to 100 characters."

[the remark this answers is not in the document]
xnull: That's obsolete information) it was never correct also. There was limitation 80 letters, now it's 120

xnull:
Guys, please leave your comments if you have anything. I fixed/answered to all review questions by now

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
vjeko: What do you mean it is outdated? It is you who changed it from 100 to 120 in this PR: https://github.com/CorfuDB/CorfuDB/pull/1460/files

Why was this change made?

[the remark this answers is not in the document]
xnull: Oh, sorry, my bad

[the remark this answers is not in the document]
xnull: I thought it was 80
We discuss it, to increase it, finally decided to increase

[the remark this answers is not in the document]
vjeko: Who discussed it? You and who else? If you want to make changes to code style-guidelines, you need to include the entire team.

But, you still have not answered my question -- why was this change made? The only thing I see on that PR is your comment:

> In my code I use 120, is it really important to extract it in another PR?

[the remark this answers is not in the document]
xnull: The entire team discussed it. The main statement was - since we have screens with a high resolution we can increase the line length, also IDEA has limit 120 letters by default.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,204 @@
...
+    public Long getIfPresent(ConflictTxStream conflictKey) {
+        return conflictCache.getIfPresent(conflictKey);
+    }
+
+    /**
+     * Invalidate all records up to a trim mark.
+     *
+     * @param trimMark trim mark
+     */
+    public void invalidate(long trimMark) {
```
zalokhan: I agree with Maithem. Please rename this method to invalidateUpTo. From the name it seems, it's invalidating only one entry. This method invokes `conflictCache.invalidate(key)` which invalidates only 1 key. It can get confusing.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -423,11 +367,9 @@ public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg>
         // It is necessary because we reset the sequencer.
         if (!bootstrapWithoutTailsUpdate) {
             // Evict all entries from the cache. This eviction triggers the callback modifying the maxConflictWildcard.
-            conflictToGlobalTailCache.cleanUp();
-
+            cache.invalidateAll();
```
zalokhan: Why is this changed from cleanUp to invalidateAll required?

xnull: because `cleanUp` doesn't delete all records it just triggers `maintenance` logic

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,204 @@
...
+                .executor(Runnable::run)
+                .writer(getDefaultCacheWriter())
+                .recordStats()
+                .build();
+    }
+
+    private CacheWriter<ConflictTxStream, Long> getDefaultCacheWriter() {
+        return new CacheWriter<ConflictTxStream, Long>() {
+            /**
+             * Let caffeine do its work
```
StrikingThirteen: Can you rephrase this generic comment?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -90,51 +84,20 @@
...
-     * which were evicted from the cache by the time this server is elected
-     * the primary sequencer. This means that any snapshot timestamp below this
-     * actual threshold would abort due to NEW_SEQUENCER cause.
-     */
-    private final Cache<String, Long> conflictToGlobalTailCache;
-
-    private long maxConflictWildcard = Address.NOT_FOUND;
-
-    private long maxConflictNewSequencer = Address.NOT_FOUND;
+    private final ConcurrentMap<UUID, Long> streamTailToGlobalTailMap = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
xnull: I didn't touch it, so I left it as is
also I believe that adding suffixes which equals to the data structure is not a good idea

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -193,54 +144,48 @@ public SequencerServer(ServerContext serverContext) {
...
+        this.cache = new SequencerServerCache(cacheSize);
 
         setUpTimerNameCache();
     }
 
+    @Override
+    public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
+        if ((sequencerEpoch != serverContext.getServerEpoch())
+                && (!msg.getMsgType().equals(CorfuMsgType.BOOTSTRAP_SEQUENCER))) {
+            log.warn("Rejecting msg at sequencer : sequencerStateEpoch:{}, serverEpoch:{}, "
```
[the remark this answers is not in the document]
xnull: wouldn't like to touch this code, out of scope. 
Also, I have no idea what should be the message. 
If you provide a better version of this method I can add it in this pr

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -398,18 +338,22 @@ public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg>
...
             r.sendResponse(ctx, msg, CorfuMsgType.NACK.msg());
             return;
         }
 
         // Stale bootstrap request should be discarded.
         if (serverContext.getSequencerEpoch() >= bootstrapMsgEpoch) {
-            log.info("Sequencer already bootstrapped at epoch {}. "
-                    + "Discarding bootstrap request with epoch {}", sequencerEpoch, bootstrapMsgEpoch);
+            log.info("Sequencer already bootstrapped at epoch {}. Discarding bootstrap request with epoch {}",
+                    sequencerEpoch, bootstrapMsgEpoch
```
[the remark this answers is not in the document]
xnull: I prefer to move it to the next line

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -595,37 +531,25 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
...
-                    .forEach(txEntry ->
-                            // and for each conflict param
-                            txEntry.getValue().stream().forEach(conflictParam ->
-                                    // insert an entry with the new timestamp
-                                    // using the hash code based on the param
-                                    // and the stream id.
-                                    conflictToGlobalTailCache.put(
-                                            getConflictHashCode(txEntry
-                                                    .getKey(), conflictParam), newTail - 1)));
+            // and for each conflict param
```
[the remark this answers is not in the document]
xnull: no, it was an original comment

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,204 @@
...
+import javax.annotation.Nonnull;
+import java.util.UUID;
+
+/**
+ * Sequencer server cache.
+ * Contains transaction conflict-resolution data structures.
+ *
+ * The SequencerServer use its own thread/s. To guarantee correct tx conflict-resolution,
+ * the {@link SequencerServerCache#conflictCache} must be updated
+ * along with {@link SequencerServerCache#maxConflictWildcard} at the same time (atomically) to prevent race condition
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,204 @@
...
+import org.corfudb.util.Utils;
+
+import javax.annotation.Nonnull;
+import java.util.UUID;
+
+/**
+ * Sequencer server cache.
+ * Contains transaction conflict-resolution data structures.
+ *
+ * The SequencerServer use its own thread/s. To guarantee correct tx conflict-resolution,
```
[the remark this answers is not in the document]
xnull: what is not clear in the comment? Could you suggest what to improve?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,204 @@
...
+ * the {@link SequencerServerCache#conflictCache} must be updated
+ * along with {@link SequencerServerCache#maxConflictWildcard} at the same time (atomically) to prevent race condition
+ * when the conflict stream is already evicted from the cache but `maxConflictWildcard` is not updated yet,
```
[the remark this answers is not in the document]
xnull: highlight?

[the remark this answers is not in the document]
xnull: This comment exists because of the problem)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,204 @@
...
+    private long maxConflictWildcard = Address.NOT_FOUND;
+
+    /**
+     * maxConflictNewSequencer represents the max update timestamp of all the conflict keys
+     * which were evicted from the cache by the time this server is elected
+     * the primary sequencer. This means that any snapshot timestamp below this
+     * actual threshold would abort due to NEW_SEQUENCER cause.
+     */
+    @Getter
+    private long maxConflictNewSequencer = Address.NOT_FOUND;
```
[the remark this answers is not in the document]
xnull: I like expressive names) but that's not my variable, so I left it as is)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: done.
Just FYI: Also, caffeine has very good documented methods, it should be clear from caffeine java doc

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * Invalidate all records up to a trim mark.
+     *
+     * @param trimMark trim mark
+     */
+    public void invalidateUpTo(long trimMark) {
+        log.debug("Invalidate sequencer cache. Trim mark: {}", trimMark);
+
+        long entries = 0;
+        for (ConflictTxStream key : conflictCache.asMap().keySet()) {
```
StrikingThirteen: Why not taking the value from the entry set instead of getIfpresent?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
StrikingThirteen: ok

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
StrikingThirteen: why? at least on github doesn't seem to align with anything

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: Because we want to use the API provided by the cache. Going underlying, get underlying structure and use it, that seems no so good idea to me. We get the underlying map and use it (`asMap()` method) only because there is no different ways to iterate through cache

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
StrikingThirteen: but before change it was from a couple lines below at the forEach statement.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
StrikingThirteen: I suggested a couple of points above and below.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
xnull: `class.method(
   variable,
   variable
   ... 
 )
`
1. You have separated and aligned blocks of code. Easier to read.
2. Sometimes useful for debug, imagine this situation:
`class.method(v1, v2,...v8)` gives you NPE, so try to figure out what variable is null? Is it method? v1 or v2 or v3 etc?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
StrikingThirteen: ok

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
StrikingThirteen: maybe something like "The problem is/happens when..."
also go -> go through

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
ben-manes: fyi the `asMap()` is safe to use however you prefer and is meant as an escape hatch for your more complex flows. There is no performance penalty or quirks to using it. The `Cache` interface is to be more idiomatic for the common case and better communicates intent. You can use `entrySet` or the convenience `forEach(k, v -> ...)` methods if desired (or not, of course).

xnull: Yep, I just try to avoid leaking abstractions. If I use some class I want to use its API. I prefer to avoid extracting underlying data structures and use them directly.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
xnull: Decided create a PR, and continue discussion in there

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -145,41 +108,29 @@
...
      * Returns a new SequencerServer.
+     *
      * @param serverContext context object providing parameters and objects
      */
     public SequencerServer(ServerContext serverContext) {
         this.serverContext = serverContext;
-        this.opts = serverContext.getServerConfig();
-        this.executor = Executors.newFixedThreadPool(serverContext.getSequencerThreadCount(),
-                new ServerThreadFactory("sequencer-", new ServerThreadFactory.ExceptionHandler()));
+        Map<String, Object> opts = serverContext.getServerConfig();
```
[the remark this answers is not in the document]
xnull: There is another PR for that - extracting SequencerServerConfig, will be fixed there

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
xnull: done. reverted

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
xnull: Actually, there is no problem. There was a problem before with prev implementation, but that's not relevant anymore and shouldn't go to documentation

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -145,41 +108,29 @@
...
+        this.executor = Executors.newFixedThreadPool(
+                serverContext.getSequencerThreadCount(),
```
Maithem: The design is based on a single threaded sequencer, having this configuration doesn't make sense really.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
Maithem: You can also install a new cache.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -367,14 +314,7 @@ public synchronized void trimCache(CorfuPayloadMsg<Long> msg,
...
 
-        long entries = 0;
-        for (Map.Entry<String, Long> entry : conflictToGlobalTailCache.asMap().entrySet()) {
-            if (entry.getValue() < trimMark) {
-                conflictToGlobalTailCache.invalidate(entry.getKey());
-                entries++;
-            }
-        }
-        log.info("trimCache: Evicted {} entries", entries);
+        cache.invalidateUpTo(trimMark);
```
Maithem: If the trim mark doesn't change why do you still iterate the cache ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -0,0 +1,210 @@
...
+            Long currTrimMark = getIfPresent(key);
```
Maithem: currTrimMark is actually not a trim mark, its transaction version. I think its an inappropriate name.

Maithem: This is really superfluous. Why don't you just iterate the entry set?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xnull: Right now it's possible to set up more than one thread for a sequencer. Let's discuss if we should delete that parameter. Anyways, that's out of scope for this PR

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: fixed

xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xnull: Recreate cache is a very expensive operation, I just measured it's about 5 times slower than invalidateCache

ben-manes: Yep. The cache does a few reflective calls on create (e.g. class load the optimal entry factory for minimal memory overhead). Some of that could be faster but unfortunately there was a JDK memory leak with Method Handles. Regardless things are optimized favoring the assumption that the cache will probably be long-lived.

xnull:
Performance testing.
Algorithm: 
 - start a corfu server
 - create 64 corfuRuntime's
 - execute: `rt.getSequencerView().query(...);` and `rt.getSequencerView().next(...)` asynchronously using 8 threads.

Results:
 - master: 7-8 sec for a million operations
 - pr:  7-8 sec for a million operations

Issue #1767, opened 2019-03-01
Do not abort transaction if same primary sequencer spans consecutive epochs


WenbinZhu:
## Overview

Currently any transaction that spans across epochs will be aborted with cause NEW_SEQUENCER, but this is not necessary. If the primary sequencer doesn't change during consecutive epoch changes, we should let the transactions commit, which could improve performance.

WenbinZhu:
Addressed in #1771

PR #1771, opened 2019-03-08
Allow transactions that span epochs to commit if primary sequencer does not change.

WenbinZhu:
## Overview

- Currently we abort any transaction that spans epochs. But this is not necessary. If the primary sequencer remains the same (i.e. it has consecutive epochs), we could allow the transaction to commit as the sequencer has all the information.

- This could improve performance in case of flakey nodes.

- A unit test that covers two cases: primary sequencer has / does have consecutive epochs.

- Removed sequencer thread configurations and all synchronization
    modifiers in SequencerServer since it is single thread by design.

Related issue(s) (if applicable): #1767 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -434,6 +445,11 @@ public synchronized void resetServer(CorfuPayloadMsg<SequencerTailsRecoveryMsg>
             streamTailToGlobalTailMap.putAll(streamTails);
         }
 
+        // Update firstConsecutiveEpoch if the bootstrap epoch is consecutive.
```
Maithem: This optimization is very specific, why don't you address the general case (i.e. track a range of epochs as opposed to just the last epoch change) ?

Every time the sequencers epoch changes, you can maintain a consecutive range [a, b] and then any epoch within [a, b] can be accepted.

WenbinZhu: @Maithem Sorry for the confusion, this comment is wrong,  missed a "not", should be "if the bootstrap epoch is not consecutive". I've made the correction. 

This patch does track a range of epochs: the `firstConsecutiveEpoch` is `a` in your example, and the current sequencer epoch is `b`.  Any time the sequencer epoch changes, we update `firstConsecutiveEpoch` if the new epoch is not consecutive. If it's consecutive, we don't update `firstConsecutiveEpoch`. So in this way, the range [firstConsecutiveEpoch, sequencer epoch] is a consecutive range. Any transaction falls in this range can be accepted for further resolution.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -248,10 +257,12 @@ private TxResolutionResponse txnCanCommit(TxResolutionInfo txInfo) {
...
         // A transaction can start with a timestamp issued from a previous
         // epoch, so we need to reject transactions that have a snapshot
-        // timestamp with a different epoch than the this sequencer's epoch.
-        if (txSnapshotTimestamp.getEpoch() != sequencerEpoch) {
-            log.debug("ABORT[{}] snapshot-ts[{}] current epoch[{}]", txInfo,
-                    txSnapshotTimestamp, sequencerEpoch);
+        // timestamp less than the firstConsecutiveEpoch since we are sure
+        // this sequencer is always the primary sequencer after this epoch.
+        long tokenEpoch = txSnapshotTimestamp.getEpoch();
+        if (tokenEpoch != sequencerEpoch && tokenEpoch < firstConsecutiveEpoch) {
```
vjeko: In what case will you end up tokenEpoch > firstConsecutiveEpoch? From what I understand, you will end up with either:

1) tokenEpoch < firstConsecutiveEpoch
2) tokenEpoch == firstConsecutiveEpoch

WenbinZhu: Every time system epoch changes, the primary sequencer will get a bootstrap request. In the bootstrap handling logic, firstConsecutiveEpoch will be updated if the bootstrap epoch is a consecutive epoch to the current sequencer epoch:https://github.com/CorfuDB/CorfuDB/blob/17f8d5314d0b415a9127f48333d1605580d3b79e/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java#L449

In this way, we keep a consecutive range: [firstConsecutiveEpoch, sequencerEpoch]. This sequencer is the primary sequencer for every epoch within this range. So when a transaction has a tokenEpoch within this range, it could be accepted. In this case, from the time when the transaction starts to the time when transaction wants to commit, the primary sequencer never changes (this is the case when we end up tokenEpoch >= firstConsecutiveEpoch).

zalokhan: Shouldn't this be ```if (tokenEpoch != sequencerEpoch || tokenEpoch < firstConsecutiveEpoch)```?

vjeko: I see. I though you were updating the `firstConsecutiveEpoch` when `bootstrapMsgEpoch == sequencerEpoch + 1`. I would be nice if you expressed this logic in terms of a sliding window/range:

`strictlyMonotonicEpochWindow = [firstConsecutiveEpoch, sequencerEpoch]`

.. and then do the boundary check in form of 'range.contains(tokenEpoch)'. You might want to take a look at: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Range.html

WenbinZhu: Actually my first thought was also using Range, but then I found the approach I'm currently using is more efficient since it only keeps a `long` variable and does a simple long comparison, instead of using `contains`, which is more costly.

vjeko: You really should not worry about performance in this case, since epoch changes happen very infrequently (relative to other events in the system). I would argue that this is a premature optimization.

Readability and understandability are way more important.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -148,11 +150,18 @@
...
     @Setter
     private volatile long sequencerEpoch = Layout.INVALID_EPOCH;
 
+    /**
+     * The first consecutive epoch of this sequencer as primary sequencer
+     * i.e. this sequencer has been primary sequencer for all the consecutive
+     * epochs from this epoch to {@link this#sequencerEpoch}
+     */
+    @Getter
+    private volatile long firstConsecutiveEpoch = Layout.INVALID_EPOCH;
```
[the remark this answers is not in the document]
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -248,10 +257,12 @@ private TxResolutionResponse txnCanCommit(TxResolutionInfo txInfo) {
 
         // A transaction can start with a timestamp issued from a previous
         // epoch, so we need to reject transactions that have a snapshot
-        // timestamp with a different epoch than the this sequencer's epoch.
-        if (txSnapshotTimestamp.getEpoch() != sequencerEpoch) {
-            log.debug("ABORT[{}] snapshot-ts[{}] current epoch[{}]", txInfo,
-                    txSnapshotTimestamp, sequencerEpoch);
+        // timestamp less than the firstConsecutiveEpoch since we are sure
```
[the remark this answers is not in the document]
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -85,6 +84,9 @@
 @Slf4j
 public class SequencerServer extends AbstractServer {
 
+    // Default cache size for transaction conflict resolution
+    private static final long DEFAULT_CACHE_SIZE = 250_000;
```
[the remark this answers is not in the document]
WenbinZhu: Done. Actually this default value is already exposed in CorfuServer configurations. Removed this variable since it's redundant. Moved get config logic to ServerContext.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -148,11 +147,19 @@
...
     private volatile long sequencerEpoch = Layout.INVALID_EPOCH;
 
+    /**
+     * The lower bound of the consecutive epoch range that this sequencer
+     * observes as the primary sequencer. i.e. this sequencer has been the
+     * primary sequencer for all the consecutive epochs from this epoch to
+     * {@link this#sequencerEpoch}
+     */
+    @Getter
+    private volatile long epochRangeLowerBound = Layout.INVALID_EPOCH;
```
xnull: Please use AtomicLong instead of volatile

xnull: And also why you need volatile if the sequencer is single threaded

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -468,6 +468,15 @@ public long getSequencerEpoch() {
         return epoch == null ? Layout.INVALID_EPOCH : epoch;
     }
 
+    /**
+     * Get the sequencer cache size.
+     *
+     * @return Sequencer cache size
+     */
+    public long getSequencerCacheSize() {
+        return Long.parseLong(getServerConfig(String.class, "--sequencer-cache-size"));
```
xnull: It colud be an NPE here, please add more checks

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
WenbinZhu: Yes I agree, will do.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
WenbinZhu: There won't be NPE here. See https://github.com/CorfuDB/CorfuDB/blob/master/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java#L134 
We provide a default value if user not specified.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xnull: Also, you could extract `[firstConsecutiveEpoch, sequencerEpoch]` into a class (it could be ConsequtiveEpoch?) also move all related logic to that class and cover the class by unit tests.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
xnull: What if tomorrow someone deletes that default value?
Your code must work in any case, it must be independent of some other ode somewhere, don't rely heavily on only current code, think about changes

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
WenbinZhu: Will do.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
WenbinZhu: @vkeko Add a helper method for range check as discussed.

WenbinZhu:
Updates:

Removed sequencer thread configurations and all synchronization
modifiers in SequencerServer (One unit test failing because the TestServerRouter is not using the server executors, which is not correct. Addressed in #1781, Once that fix is in, this PR can pass)

Please review again. Thanks.

Maithem:
Can you please rebase this? I think its time to check it in.

WenbinZhu:
@Maithem  rebased.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -128,23 +140,18 @@ public SequencerServer(ServerContext serverContext) {
         this.serverContext = serverContext;
         Map<String, Object> opts = serverContext.getServerConfig();
         this.executor = Executors.newFixedThreadPool(
```
Maithem: You can make this a single threaded executor and get rid of the variable (and merge the comments).

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -455,8 +483,10 @@ private Timer getTimer(byte reqType) {
     private void handleRawToken(CorfuPayloadMsg<TokenRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
         final TokenRequest req = msg.getPayload();
 
-        Token token = new Token(sequencerEpoch, globalLogTail.getAndAdd(req.getNumTokens()));
-        r.sendResponse(ctx, msg, CorfuMsgType.TOKEN_RES.payloadMsg(new TokenResponse(token, Collections.emptyMap())));
+        Token token = new Token(sequencerEpoch, globalLogTail);
```
Maithem: Maybe add a comment about how the tail points to an open slot and not the last written slot and therefore you return the current position and then update.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
WenbinZhu: Done.

PR #1775, opened 2019-03-12
Message epoch validation on prefixTrim.

zalokhan:
## Overview

Description: Avoid batch writer validating the token epoch on prefix trim. This is to prevent rejecting the trim request for an address with an old epoch. The address in its completeness is a valid address with the old epoch. As long as the address is generated by the CheckpointWriter after persisting the checkpoint, the trim on this address should be processed.

Why should this be merged: We invalidate a valid trim request on an older address.

- [x] Public API has Javadoc


zalokhan:
The `trimMark` returned by the `CheckpointWriter` is not the first address appended to but it is the snapshot address returned by the Sequencer. Since this address relies on the sequencer, this can still regress. Hence, the epoch still can't be separated from the token in the prefixTrim request.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -234,7 +234,11 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
                             IServerRouter r) {
         try {
             TrimRequest req = msg.getPayload();
-            batchWriter.prefixTrim(req.getAddress());
+            // Here the token is not used as is to perform the trim as the epoch at which the checkpoint was completed
```
xnull: why not to transform this doc to javadoc for this method?

xnull:
the test  `ensureMCWUsesRealTail` is deleted, can you provide some alternative to this?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
```diff
@@ -59,32 +59,6 @@ public void testCheckpointTrim() throws Exception {
                 .containsKeys("a", "b", "c");
     }
 
-    @Test
```
annym: Shouldn't we keep this test to show that this will not cause the trim to fail, and attempt to access the actual address to verify it was trimmed?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zalokhan: Done.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
zalokhan: Done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,8 +126,11 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
      */
     public Token appendCheckpoint() {
         long start = System.currentTimeMillis();
+        Token markerToken = rt.getSequencerView().next().getToken();
+        rt.getAddressSpaceView().write(markerToken, "marker".getBytes());
```
xnull: Please use a constant for `"marker".getBytes()`

zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -249,12 +249,19 @@ private void fillHole(CorfuPayloadMsg<FillHoleRequest> msg, ChannelHandlerContex
...
+     * trim is performed only on the token provided by the CheckpointWriter which ensures that the checkpoint
+     * was persisted. Using any other address to perform a trim can cause data loss.
+     */
     @ServerHandler(type = CorfuMsgType.PREFIX_TRIM)
     private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx,
                             IServerRouter r) {
         try {
             TrimRequest req = msg.getPayload();
-            batchWriter.prefixTrim(req.getAddress());
+            batchWriter.prefixTrim(req.getAddress(), msg.getEpoch());
```
xnull: Can you also check that msg.getEpoch() not more than epoch we have right now

zalokhan: You mean `msg.getEpoch()` == the seal epoch?
That check is already being made in the BatchWriter.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,8 +132,11 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
      */
     public Token appendCheckpoint() {
         long start = System.currentTimeMillis();
+        Token markerToken = rt.getSequencerView().next().getToken();
+        rt.getAddressSpaceView().write(markerToken, SNAPSHOT_MARKER);
```
xnull: can we bring some useful information instead of a mock word here?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: ok

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
```diff
@@ -78,11 +82,11 @@ public void ensureMCWUsesRealTail() throws Exception {
...
 
         MultiCheckpointWriter mcw = new MultiCheckpointWriter();
-        mcw.addMap((SMRMap) map);
+        mcw.addMap(map);
         Token trimAddress = mcw.appendCheckpoints(getRuntime(), "author");
         Token staleTrimAddress = new Token(trimAddress.getEpoch() - 1, trimAddress.getSequence());
-        assertThatThrownBy(() -> getRuntime().getAddressSpaceView().prefixTrim(staleTrimAddress))
-                .isInstanceOf(WrongEpochException.class);
+
+        getRuntime().getAddressSpaceView().prefixTrim(staleTrimAddress);
```
xnull: CAn you also check that LogUnitServer TRIM_MARK_REQUEST gives you correct/expected trim mark address

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -360,7 +377,7 @@ public void checkpointWriterInterleavedTest() throws Exception {
         // for the checkpoint. Since regular writes are being interleaved with
         // the checkpointer perfectly, the checkpointer will write on every other
         // address (i.e. odd offsets from the base, startAddress)
-        long startAddress = globalTail.getSequence() + 1;
+        long startAddress = globalTail.getSequence() + 2;
```
xnull: Please extract `2` as a local variable and provide the explanation why we have +2, what those 2 addresses used for

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -61,13 +61,13 @@ public void testEmptyMapCP() throws Exception {
 
         // Verify that a CP wasn't generated
         long address = mcw.appendCheckpoints(r, "A1").getSequence();
-        assertThat(address).isEqualTo(-1);
+        assertThat(address).isEqualTo(0);
```
xnull: 0 means? If it's possible and presented in `Address` class, please use Adress.SOME_CONSTANT for that

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
zalokhan: Done.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
zalokhan: Good idea. Done.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
zalokhan: 0 is not a constant. It is the marker entry placed at 0. I have made it a constant in the test and added comments for clarification.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -108,12 +108,13 @@ public void bulkWrite(List<LogData> entries, long epoch) {
      * Trim addresses from log up to a prefix.
      *
      * @param address prefix address to trim to (inclusive)
+     * @param epoch   epoch stamped on the trim request.
      */
-    public void prefixTrim(@Nonnull Token address) {
+    public void prefixTrim(@Nonnull Token address, long epoch) {
```
StrikingThirteen: Can we make the variable a more expressive one such as prefixEpoch or similar. It might be confusing to maintainers considering the address also has epoch.

StrikingThirteen: This is changing the public api. we are sure about the old usage? Might want to consider overloading this method.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
StrikingThirteen: Just to prevent the confusion maybe consider a prefix method with a long address (from Token) and epoch from the message.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -300,6 +303,20 @@ public void checkpointWriterTest() throws Exception {
      *  destroyed (logically, not physically) by the CP, so we have
      *  to use a different position 'history' for our assertion
      *  check.
+     *
+     * +---------------------------------------------------------------------+
```
StrikingThirteen: Clear example for test. Thanks.

Maithem:
Lets get this in, please rebase. 
Please re-review it @xnull @WenbinZhu

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -57,6 +57,12 @@
             "append-checkpoint";
     private Timer appendCheckpointTimer = metricRegistry.timer(CHECKPOINT_TIMER_NAME);
 
+    /**
+     * Marker to insert to start a snapshot transaction to bypass sequencer token regression on sequencer failover.
+     * Persisting this marker freezes all tokens dispatched before this with the current/old sequencer.
```
WenbinZhu: Maybe wrap the above two comment lines? They look a little too long compared to other lines in this file.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zalokhan: I feel that will create more confusion. We have only one prefixTrim. The epoch should only be the message epoch and address - trim sequence number.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,8 +132,11 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
      */
     public Token appendCheckpoint() {
         long start = System.currentTimeMillis();
```
Maithem: Actually let's get rid of the marker, its not needed you can just use the stream tail.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,8 +132,11 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
...
+        Token markerToken = rt.getSequencerView().next().getToken();
```
Maithem: Just use the tail directly, this is not needed.

`Token markerToken = rt.getSequencerView().query(streamId)`

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -107,13 +107,14 @@ public void bulkWrite(List<LogData> entries, long epoch) {
     /**
      * Trim addresses from log up to a prefix.
      *
-     * @param address prefix address to trim to (inclusive)
+     * @param address     prefix address to trim to (inclusive)
+     * @param prefixEpoch epoch stamped on the trim request.
      */
-    public void prefixTrim(@Nonnull Token address) {
+    public void prefixTrim(@Nonnull Token address, long prefixEpoch) {
         try {
```
Maithem: Redundant information. Just pass a token with an un-initialized epoch.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -260,15 +261,15 @@ private void batchWriteProcessor() {
                                 res.add(currOp);
                                 break;
                             case TAILS_QUERY:
-                                TailsResponse tails = streamLog.getTails();
+                                TailsResponse tails = streamLog.getTails(currOp.getEpoch());
```
Maithem: I don't understand this. Why pass the message epoch all the way down to StreamLogs and then return it in the response? It seems like only BatchWriterOperation needs to know this, no?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: Just use the token and document it.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
```diff
@@ -52,7 +52,7 @@
     /**
      * Get the global tail and stream tails.
      */
-    TailsResponse getTails();
+    TailsResponse getTails(long epoch);
```
Maithem: I don't see this as necessary, why make the StreamLog aware of message epochs ?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -57,6 +57,12 @@
             "append-checkpoint";
     private Timer appendCheckpointTimer = metricRegistry.timer(CHECKPOINT_TIMER_NAME);
 
+    /**
+     * Marker to insert to start a snapshot transaction to bypass sequencer token regression on sequencer failover.
+     * Persisting this marker freezes all tokens dispatched before this with the current/old sequencer.
+     */
+    private static final byte[] SNAPSHOT_MARKER = "marker".getBytes();
```
Maithem: Just use stream tails, this is not needed.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -472,16 +472,19 @@ public static void printLogAnatomy(CorfuRuntime runtime, ILogData logData) {
     static TailsResponse getTails(Set<TailsResponse> responses) {
         long globalTail = Address.NON_ADDRESS;
         Map<UUID, Long> globalStreamTails = new HashMap<>();
+        long minEpoch = Layout.INVALID_EPOCH;
 
         for (TailsResponse res : responses) {
+            // All epochs should be equal as all the tails are queried using a single runtime layout.
+            minEpoch = Math.min(minEpoch, res.getEpoch());
```
Maithem: If all tails are expected to be computed with the same epoch, then why do we have a minEpoch ?

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
zalokhan: I still need the epoch for the BatchWriter epoch validation. I will change it to carry the prefixEpoch in the token's epoch.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zalokhan: Cool. And will read on the stream tail so as to persist it by hole-filling if needed.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
zalokhan: I agree. Since the TailsResponse is immutable, I did that to keep the construction of the TailsResponse in a single place. I'll change it.

zalokhan:
@Maithem Addressed review comments.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TailsResponse.java
```diff
@@ -18,19 +21,24 @@
 
 @Data
 @RequiredArgsConstructor
+@AllArgsConstructor
 public class TailsResponse implements ICorfuPayload<TailsResponse> {
 
+    long epoch = Layout.INVALID_EPOCH;
+
```
Maithem: This is not needed anymore, right?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,8 +128,12 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
      */
     public Token appendCheckpoint() {
         long start = System.currentTimeMillis();
+        //TODO:Add comments.
+        long address = rt.getStreamsView().append(new MultiSMREntry(), null, streamId);
```
Maithem: Complete this comment. Just mention how this adds a no-op to the stream while maintaining the backpointers (i.e. different than a hole).

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -375,19 +375,11 @@ public void prefixTrim(final Token address) {
...
-                            "and retry, attempt: {}/{}", address, runtimeEpoch, serverEpoch, x+1, numRetries);
-                    runtime.invalidateLayout();
-                } else {
-                    // wrongEpochException corresponds to a stale trim address (prefix trim token on the wrong epoch)
-                    log.error("prefixTrim[{}]: stale prefix trim. Prefix trim on wrong epoch {}, " +
-                            "while server on epoch {}", address, address.getEpoch(), serverEpoch);
-                    throw wee;
-                }
+                long runtimeEpoch = runtime.getLayoutView().getLayout().getEpoch();
+                log.warn("prefixTrim[{}]: wrongEpochException, runtime is in epoch {}, while server is in epoch {}. "
```
Maithem: This can be a different epoch than what the layout helped executed with.

WenbinZhu: @Maithem I think they are the same epoch, as this runtime is the same runtime in layoutHelper (they are both using AbstractView's runtime), right?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TailsResponse.java
zalokhan: Still needed. To return from getTails from Utils.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: The log statement executes outside the lambda that is executed by the layout helper, so its possible that the layout changes in between.

WenbinZhu: I see, thanks. @Maithem @zalokhan Also I think inside layoutHelper, when we get an exception, we can log the name of the function that is executed together with the details, so that outside the lambda, we don't need to log again the exact details (for example in this case, we just need to log the exception name and  number of retries here, the actual value of correct epoch will be logged in layoutHelper together with the function name: prefixTrim).

zalokhan: @Maithem The second parameter in the layoutHelper is true - rethrow all exceptions.
This means there are no retries if WrongEpochException is encountered.

Although there exists a case where if we encounter a ServerNotReadyException first, then there can be a retry.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -249,12 +249,19 @@ private void fillHole(CorfuPayloadMsg<FillHoleRequest> msg, ChannelHandlerContex
         }
     }
 
+    /**
+     * Perform a prefix trim.
+     * Here the token is not used as is to perform the trim as the epoch at which the checkpoint was completed
```
annym: typo:  as is to -> to perform the trim, as...

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,8 +127,18 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
      */
```
annym: I know this is not relevant to your change, but can we update the return description for this method?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,8 +127,18 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
...
     public Token appendCheckpoint() {
         long start = System.currentTimeMillis();
+
+        // Queries the sequencer for the global log tail. We then read the global log tail to
+        // persist the entry on the logunit in turn preventing from a new sequencer from
+        // regressing tokens.
+        Token markerToken = rt.getSequencerView().query().getToken();
+        if (Address.nonAddress(markerToken.getSequence())) {
+            return markerToken;
```
annym: Is it correct to return this token? appendCheckpoint is expecting the Token for the checkpoint and it is not true that this is it, we did not even checkpoint, should we throw an error? or is this our contract?

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -427,10 +428,22 @@ public void reconfigureSequencerServers(Layout originalLayout, Layout newLayout,
                         || !originalLayout.getPrimarySequencer()
                         .equals(newLayout.getPrimarySequencer())) {
 
+                    // The tails query and the sequencer recovery must be performed on the same
+                    // epoch. If they are not, we lose the atomicity of bootstrapping the sequencer
+                    // which can lead to data operations across epochs leaking into the
+                    // Address space causing inconsistencies and data loss.
+
                     //TODO(Maithem) why isn't this getting the tails
```
annym: Now that you're cleaning this code, can you take care of this pending TODO, from what I see addressView is getting It from Utils but through the LayoutHelper... if we do it directly from Utils would we guarantee it throws a WrongEpochException if not on the same epoch as runtime? (assuming it is on newLayout, am I right?)

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -300,6 +300,19 @@ public void checkpointWriterTest() throws Exception {
...
+     * +-----------------------------------------------------------------+
+     * | F0 | F1 | F2 | S | C | M0 | C | M1 | C | M2 | E  | L0 | L1 | L2 |
+     * +-----------------------------------------------------------------+
+     * F : First batch of entries.
+     * S : Start of checkpoint.
+     * C : Continuation of checkpoints
+     * M : Middle batch of entries.
+     * E : End of checkpoints
+     * L : Last batch of entries.
+     *
```
annym: very useful :)

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TailsResponse.java
Maithem: This is serialized and being sent to the server, but the server doesn't use this value.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem: I would rather increment the stream tail to prevent regression because that's correct "by-construction", but this method assumes that no writer will be able to write between the stream tail and the marker, which is still error prone.  

Anyways, we already had this discussion just wanted to put it on the record.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: That still doesn't prevent other threads from invalidating the layout between the time the exception happens and this log statement.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -249,12 +249,19 @@ private void fillHole(CorfuPayloadMsg<FillHoleRequest> msg, ChannelHandlerContex
...
+     * trim is performed only on the token provided by the CheckpointWriter which ensures that the checkpoint
+     * was persisted. Using any other address to perform a trim can cause data loss.
+     */
     @ServerHandler(type = CorfuMsgType.PREFIX_TRIM)
     private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx,
                             IServerRouter r) {
         try {
             TrimRequest req = msg.getPayload();
-            batchWriter.prefixTrim(req.getAddress());
+            batchWriter.prefixTrim(new Token(msg.getEpoch(), req.getAddress().getSequence()));
```
Maithem: The token is not persisted and this new code doesn't use the epoch. I suggest you drop the Token usage instead of transforming it.

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
Maithem: +1 please try to do the clean up if you can

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zalokhan: Done.

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
zalokhan: Done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zalokhan: Done.

zalokhan:
Do we have any more review comments on this? If not can we get this in today?

it/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
xnull: cool! Thanks

PR #1776, opened 2019-03-12
Abstract server state

xnull:
## Overview

Description:
A corfu netty server could be in various states. Exposing those states make a server lifecycle unambiguous.  

Why should this be merged: 
Improving/clarify a server lifecycle.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
@@ -60,12 +60,18 @@ public abstract class AbstractServer {
         }
     }
 
-    protected void setState(ServerState state) {
-        if (getState() == ServerState.SHUTDOWN && state != ServerState.SHUTDOWN) {
-            throw new IllegalStateException("Server is in SHUTDOWN state. Can't be changed");
-        }
+    protected void setState(ServerState newState) {
+        boolean updated = false;
 
-        this.state.set(state);
+        while(!updated){
+            ServerState currState = getState();
+
+            if (currState == ServerState.SHUTDOWN && newState != ServerState.SHUTDOWN) {
+                throw new IllegalStateException("Server is in SHUTDOWN state. Can't be changed");
+            }
+
+            updated = state.compareAndSet(currState, newState);
+        }
     }
 
     public ServerState getState() {
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
@@ -7,6 +7,7 @@ import org.corfudb.protocols.wireprotocol.CorfuMsgType;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.UnaryOperator;
 
 /**
  * Created by mwei on 12/4/15.
@@ -61,17 +62,13 @@ public abstract class AbstractServer {
     }
 
     protected void setState(ServerState newState) {
-        boolean updated = false;
-
-        while(!updated){
-            ServerState currState = getState();
-
+        state.updateAndGet(currState -> {
             if (currState == ServerState.SHUTDOWN && newState != ServerState.SHUTDOWN) {
                 throw new IllegalStateException("Server is in SHUTDOWN state. Can't be changed");
             }
 
-            updated = state.compareAndSet(currState, newState);
-        }
+            return newState;
+        });
     }
 
     public ServerState getState() {
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
@@ -84,7 +84,7 @@ public abstract class AbstractServer {
 
     /**
      * The server state.
-     * Represents server in a particular state: READY, NOT_READY, SHUTDOWN etc.
+     * Represents server in a particular state: READY, NOT_READY, SHUTDOWN.
      */
     public enum ServerState {
         READY, NOT_READY, SHUTDOWN
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
@@ -60,11 +60,15 @@ public abstract class AbstractServer {
         }
     }
 
-    protected void setState(ServerState state){
+    protected void setState(ServerState state) {
+        if (getState() == ServerState.SHUTDOWN && state != ServerState.SHUTDOWN) {
+            throw new IllegalStateException("Server is in SHUTDOWN state. Can't be changed");
+        }
+
         this.state.set(state);
     }
 
-    public ServerState getState(){
+    public ServerState getState() {
         return state.get();
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
@@ -1,12 +1,12 @@
 package org.corfudb.infrastructure;
 
 import io.netty.channel.ChannelHandlerContext;
-import lombok.Getter;
-import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.CorfuMsg;
+import org.corfudb.protocols.wireprotocol.CorfuMsgType;
 
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Created by mwei on 12/4/15.
@@ -14,11 +14,10 @@ import java.util.concurrent.ExecutorService;
 @Slf4j
 public abstract class AbstractServer {
 
-    private volatile boolean shutdown;
-
-    public AbstractServer() {
-        shutdown = false;
-    }
+    /**
+     * Current server state
+     */
+    private final AtomicReference<ServerState> state = new AtomicReference<>(ServerState.READY);
 
     /**
      * Get the message handler for this instance.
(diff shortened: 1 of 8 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -49,24 +45,44 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
+        setState(ServerState.SHUTDOWN);
         getExecutor().shutdownNow();
     }
 
+    /**
+     * The server state.
+     * Represents server in a particular state: READY, NOT_READY, SHUTDOWN etc.
+     */
+    public enum ServerState {
+        READY, NOT_READY, SHUTDOWN
```
hisundar: Just curious when is NOT_READY state used since I don't see it being set?

xnull: It's not used yet, it will be used at least in LogUnitServer::resetLogUnit() (next pr-s)

hisundar: ok thanks

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -49,24 +45,48 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
      */
     public void shutdown() {
-        shutdown = true;
+        setState(ServerState.SHUTDOWN);
         getExecutor().shutdownNow();
     }
 
+    /**
+     * The server state.
+     * Represents server in a particular state: READY, NOT_READY, SHUTDOWN etc.
```
StrikingThirteen: pls remove etc.

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -49,24 +45,48 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
      * @param r   The router that took in the message.
      */
     public void handleMessage(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
-        if (shutdown) {
+        if (getState() == ServerState.SHUTDOWN) {
             log.warn("Server received {} but is already shutdown.", msg.getMsgType().toString());
             return;
         }
 
+        if (!isServerReadyToHandleMsg(msg)) {
```
StrikingThirteen: Why is this a better place for checking the Not_ready compared to the check was being done in the CorfuMsgHandler?

xnull: Because this is an internal state of AbstractServer. The server in charge of this check. If CorfuMsgHandler checks this we violate Single Responsibility principle

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -1,24 +1,23 @@
...
 
-    private volatile boolean shutdown;
-
-    public AbstractServer() {
-        shutdown = false;
-    }
+    /**
+     * Current server state
+     */
+    private final AtomicReference<ServerState> state = new AtomicReference<>(ServerState.READY);
```
WenbinZhu: Why you need AtomicReference here? You are not using any atomic operation, volatile would be enough in this case.

xnull: Because atomic always better that volatile

xnull: And I use atomic operations, see setState()

xnull: also https://pmd.github.io/latest/pmd_rules_java_multithreading.html#avoidusingvolatile

zalokhan: > Because atomic always better that volatile
Not a valid justification.
https://stackoverflow.com/questions/281132/java-volatile-reference-vs-atomicreference
In case you are using only get and set, volatile is sufficient.

But I see that you are planning to update the state variable atomically if not shutdown.

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -49,24 +45,48 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
+            r.sendResponse(ctx, msg, CorfuMsgType.NOT_READY.msg());
+            return;
+        }
+
         if (!getHandler().handle(msg, ctx, r)) {
             log.warn("Received unhandled message type {}", msg.getMsgType());
         }
     }
 
+    protected void setState(ServerState state) {
```
zalokhan: Isn't there a race possible here?
You need to compareAndSet to check it's not shutdown.

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
xnull: >Not a valid justification.
Your link doesn't explain your point. I still don't see where is my argument is not valid.

WenbinZhu: @xnull First AtomicReference is internally using volatile: https://github.com/bpupadhyaya/openjdk-8/blob/45af329463a45955ea2759b89cb0ebfe40570c3f/jdk/src/share/classes/java/util/concurrent/atomic/AtomicReference.java#L62

Second, your setState() is not atomic, get() and set() are just same as normal operation, unless you need sth. like getAndSet(), compareAndSet() etc.

xnull: >your setState() is not atomic,

that's not true

WenbinZhu: @xnull why is it atomic?  getState() calls state.get(),  and then you call state.set(), this is not atomic?

xnull: Now, I got your point. Will fix setState preventing going to incorrect state.
Regarding atomicref - it's just more powerful volatile. Don't see any reason to replace it by volatile

WenbinZhu: Another point is if you need atomic operations like getAndSet(), you need AtomicReference, but if you're only using get() and set(), please use volatile. It will be confusing and pointless if you use AtomicReference, but does not use atomic operations it provides.

xnull: get and set are not atomic operations? Using Atomic over volatile always better even if you use only get and set

WenbinZhu: @xnull  please check the implementation of get() and set(). They are just doing `value = newValue` and `return value` where value is a volatile. It's just a simple wrapper and exactly same as using volatile.

xnull: >>your setState() is not atomic

fixed

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
xnull: > It's just a simple wrapper and exactly same as using volatile.
Exactly! So
 1. using Atomic over volatile prevents you from accidentally deleting `volatile`. Whenever you use atomic you know exactly that the variable it's atomic.
 2.  Atomic requres you to use it accordingly, with volatile it's much easier to make a mistake.

WenbinZhu: I don't agree with your justification. This is abusing AtomicXXX classes. It is designed to give you ability to use atomic operations, not to replace volatile. Using AtomicReference means you need the atomic operations. Using atomic classes with no atomic operations only creates confusion.

xnull: It's your opinion) so, up to you. I believe to what I explained

infrastructure/src/main/java/org/corfudb/infrastructure/AbstractServer.java
```diff
@@ -49,24 +45,54 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
+        boolean updated = false;
+
+        while(!updated){
+            ServerState currState = getState();
+
+            if (currState == ServerState.SHUTDOWN && newState != ServerState.SHUTDOWN) {
+                throw new IllegalStateException("Server is in SHUTDOWN state. Can't be changed");
+            }
+
+            updated = state.compareAndSet(currState, newState);
```
zalokhan: You can look up `updateAndGet`. It does the same thing just removes boilerplate code.

xnull: updateAndGet doesn't work in my case, as you can see I make some extra checks.
Or if you have a good idea how it could be implemented, let me know

vjeko: Slava, you should be able to use updateAndGet:

```java
    public final V updateAndGet(UnaryOperator<V> updateFunction) {
        V prev, next;
        do {
            prev = get();
            next = updateFunction.apply(prev);
        } while (!compareAndSet(prev, next));
        return next;
    }
```

xnull: fixed

PR #1780, opened 2019-03-12
[WIP] Replace Backpointers with Stream Address Maps

annym:
## Overview

Description: to discover new updates to a stream we currently follow backpointers which in the worst case generates an RPC call for every entry in the range of updates. Furthermore, this mechanism leads to single stepping the log in the presence of holes, which can further impact affect performance. This changes proposes the use of bitmaps to contain the space of addresses of each stream. Therefore, a stream is discovered by accessing this space and reads can be done in batches.

Why should this be merged: improve performance.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
@@ -19,7 +19,9 @@ import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.infrastructure.BatchWriterOperation.Type;
 import org.corfudb.infrastructure.log.StreamLog;
+import org.corfudb.protocols.wireprotocol.LogAddressSpaceResponse;
 import org.corfudb.protocols.wireprotocol.LogData;
+import org.corfudb.protocols.wireprotocol.StreamsAddressResponse;
 import org.corfudb.protocols.wireprotocol.TailsResponse;
 import org.corfudb.protocols.wireprotocol.Token;
 import org.corfudb.runtime.exceptions.WrongEpochException;
@@ -163,10 +165,55 @@ public class BatchWriter<K, V> implements CacheWriter<K, V>, AutoCloseable {
         }
     }
 
+    public Long queryLogTail(long epoch) {
+        try {
+            CompletableFuture<Long> cf = new CompletableFuture<>();
+            operationsQueue.add(new BatchWriterOperation(Type.LOG_TAIL_QUERY, null,
+                    null, epoch, null, cf));
+            return cf.get();
+        } catch (Exception e) {
+            if (e.getCause() instanceof RuntimeException) {
+                throw (RuntimeException) e.getCause();
+            } else {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
     public TailsResponse queryTails(long epoch) {
         try {
             CompletableFuture<TailsResponse> cf = new CompletableFuture<>();
             operationsQueue.add(new BatchWriterOperation(Type.TAILS_QUERY, null,
+                    null, epoch, null, cf));
+            return cf.get();
+        } catch (Exception e) {
+            if (e.getCause() instanceof RuntimeException) {
+                throw (RuntimeException) e.getCause();
+            } else {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    public StreamsAddressResponse queryStreamsAddressSpace(long epoch) {
+        try {
+            CompletableFuture<StreamsAddressResponse> cf = new CompletableFuture<>();
+            operationsQueue.add(new BatchWriterOperation(Type.STREAMS_ADDRESS_QUERY, null,
+                    null, epoch, null, cf));
+            return cf.get();
+        } catch (Exception e) {
+            if (e.getCause() instanceof RuntimeException) {
+                throw (RuntimeException) e.getCause();
+            } else {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    public LogAddressSpaceResponse queryLogAddressSpace(long epoch) {
+        try {
+            CompletableFuture<LogAddressSpaceResponse> cf = new CompletableFuture<>();
+            operationsQueue.add(new BatchWriterOperation(Type.LOG_ADDRESS_SPACE_QUERY, null,
                     null, epoch, null, cf));
             return cf.get();
         } catch (Exception e) {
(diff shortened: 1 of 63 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -162,6 +164,21 @@ public void reset(@Nonnull long epoch) {
         }
     }
 
+    public Long queryLogTail(long epoch) {
```
lixinchengdu: Is it better to return something like TailResponse? All other similar methods returns *Response type.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -591,6 +626,22 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
                     return newTail - 1;
                 }
             });
+
+            // step 3. Add allocated addresses to each stream's address map (keep track of all addresses of this stream)
+            streamsAddressSpaceMap.compute(id, (k, v) -> {
+                if (v == null) {
+                    final Roaring64NavigableMap bitmap = new Roaring64NavigableMap();
```
lixinchengdu: bitmap seems not a good name here. It is still used as one NavigableMap.

infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
```diff
@@ -77,14 +81,37 @@ public synchronized void prefixTrim(long address) {
...
+    public synchronized StreamsAddressResponse getStreamsAddressSpace() {
+        return new StreamsAddressResponse(getCopyStreamsAddressSpace());
+    }
+
+    @Override
+    public synchronized LogAddressSpaceResponse getLogAddressSpace() {
+        return new LogAddressSpaceResponse(logMetadata.getGlobalTail(), getCopyStreamsAddressSpace());
+    }
+
+    private Map<UUID, StreamAddressSpace> getCopyStreamsAddressSpace() {
```
lixinchengdu: How about return a ImmutableMap

lixinchengdu: How about returning an ImmutableMap instance?

infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
```diff
@@ -46,8 +53,18 @@ public void update(LogData entry) {
         long entryAddress = entry.getGlobalAddress();
         updateGlobalTail(entryAddress);
         for (UUID streamId : entry.getStreams()) {
+            // Update stream tails
             long currentStreamTail = streamTails.getOrDefault(streamId, Address.NON_ADDRESS);
             streamTails.put(streamId, Math.max(currentStreamTail, entryAddress));
+
+            // Update stream address map
```
lixinchengdu: Does the recovery consider stripping of LogUnitServers?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/StreamAddressRange.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * This class represents a range of addresses for a stream.
+ *
+ * This is used to request the address map of a stream in
+ * a given boundary (limits given by start and end).
+ *
+ * Created by annym on 03/06/19
+ */
+@Data
+public class StreamAddressRange {
```
lixinchengdu: It would be better if the comment specify the end address is open or close.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/StreamAddressSpace.java
```diff
@@ -0,0 +1,134 @@
...
+    public synchronized Roaring64NavigableMap getAddressMap() {
+        return addressMap;
+    }
+
+    /**
+     * Copy this streams addresses to a set.
+     *
+     * @param queue
+     */
+    public synchronized void copyAddressesToSet(final NavigableSet<Long> queue) {
```
lixinchengdu: Why the parameter is marked as final?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/StreamAddressSpace.java
```diff
@@ -0,0 +1,134 @@
...
+                // Because our search is referenced to the stream's tail => [start > end)
+                if (address > end && address <= start) {
+                    addressesInRange.add(address);
+                }
+            });
+        }
+        return addressesInRange;
+    }
+
+    public void setTrimMark(long trimMark) {
```
lixinchengdu: Why the two methods setTrimMark and getTrimMark are not synchronized?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -260,6 +262,25 @@ public void write(IToken token, Object data) throws OverwriteException {
         return fetch(address);
     }
 
+    /**
+     * Read a range of addresses in a predictive way.
+     *
+     * If there is a cache miss, read a range of addresses, otherwise
+     * directly return the requested entry and read is defer.
+     */
+    public @Nonnull ILogData predictiveReadRange(Long address, List<Long> addresses) {
```
lixinchengdu: What is wrong with readCache? Why not rely on its internal mechanism to fetch logData?

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
```diff
@@ -88,6 +92,30 @@ public TokenResponse next(UUID ... streamIds) {
...
+        return getMultipleStreamsAddressMap(Arrays.asList(streamsAddressesRange)).get(streamsAddressesRange.getStreamID());
+    }
+
+    /**
+     * Retrieve multiple streams address space.
+     *
+     * @param streamsAddressesRange list of streams ranges to be requested.
+     * @return address space for each stream in request.
+     */
+    public Map<UUID, StreamAddressSpace> getMultipleStreamsAddressMap(List<StreamAddressRange> streamsAddressesRange) {
```
lixinchengdu: Is getStreamsAddressMap a more appreciate name? getMultipleStreamsAddressMap is not consistent with other parts.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -463,6 +466,69 @@ public static void printLogAnatomy(CorfuRuntime runtime, ILogData logData) {
...
+    /**
+     * Get global log tail.
+     *
+     * @param layout Latest layout to query log tail from Log Unit
+     * @param runtime Runtime
+     *
+     * @return Log global tail
+     */
+    public static Long getLogTail(Layout layout, CorfuRuntime runtime) {
+        Set<Long> luResponses = new HashSet<>();
```
lixinchengdu: luResponses seems not as readable as logUnitResponses.

Maithem:
Subsumed by #1818

Issue #1782, opened 2019-03-13
ManagementAgents continue to bootstrap a bootstrapped primary sequencer.


zalokhan:
## Overview

Setup: 2 nodes 9000 and 9001

9000:
```
019-03-13T18:34:31.230Z | INFO  |                       Thread-6 |           o.c.i.o.Orchestrator | run: finished action RestoreRedundancyAndMergeSegments for workflow fddd3407-0da5-4789-824a-32f899ebc923 in 32 ms
2019-03-13T18:34:31.230Z | INFO  |                       Thread-6 |           o.c.i.o.Orchestrator | run: Completed workflow fddd3407-0da5-4789-824a-32f899ebc923 in 32 ms
2019-03-13T18:34:32.125Z | INFO  |              ManagementService |            o.c.i.ServerContext | saveManagementLayout: Updated to new layout at epoch 3
2019-03-13T18:34:32.175Z | INFO  |              DetectionWorker-0 |      o.c.r.v.w.WorkflowRequest | waitForWorkflow: waiting for fddd3407-0da5-4789-824a-32f899ebc923 on attempt 0
2019-03-13T18:34:32.178Z | INFO  |              DetectionWorker-0 | RestoreRedundancyMergeSegments | verifyRequest: RestoreRedundancyMergeSegments localhost:9001 in Layout(layoutServers=[localhost:9000, localhost:9001], sequencers=[localhost:9000, localhost:9001], segments=[Layout.LayoutSegment(replicationMode=CHAIN_REPLICATION, start=0, end=-1, stripes=[Layout.LayoutStripe(logServers=[localhost:9000, localhost:9001])])], unresponsiveServers=[], epoch=3, clusterId=3a1a4aae-17f6-4bd3-bd9a-01dc4210b65c)
2019-03-13T18:34:32.182Z | INFO  |              DetectionWorker-0 |      o.c.r.v.w.WorkflowRequest | WorkflowRequest: Successfully completed RestoreRedundancyMergeSegments localhost:9001
2019-03-13T18:34:34.096Z | INFO  |                   baseServer-0 |  o.c.infrastructure.BaseServer | handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch 3
2019-03-13T18:34:46.066Z | INFO  |                    sequencer-0 |          o.c.i.SequencerServer | Sequencer already bootstrapped at epoch 3. Discarding bootstrap request with epoch 3
```

9001:
```
[...]

PR #1784, opened 2019-03-14
Skip bootstrapping an already bootstrapped sequencer

hisundar:
## Overview

Description:

If the cluster state is already ready that implies that the primary sequencer is ready too.
If so, avoid repetitive futile attempts to re-bootstrap the primary sequencer.

Why should this be merged: 

Related issue(s) (if applicable): #1782 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
@@ -16,6 +16,7 @@ import org.corfudb.infrastructure.management.PollReport;
 import org.corfudb.infrastructure.management.ReconfigurationEventHandler;
 import org.corfudb.infrastructure.management.failuredetector.ClusterGraph;
 import org.corfudb.protocols.wireprotocol.ClusterState;
+import org.corfudb.protocols.wireprotocol.NodeState;
 import org.corfudb.protocols.wireprotocol.SequencerMetrics;
 import org.corfudb.protocols.wireprotocol.SequencerMetrics.SequencerStatus;
 import org.corfudb.protocols.wireprotocol.failuredetector.FailureDetectorMetrics;
@@ -541,8 +542,10 @@ public class RemoteMonitoringService implements MonitoringService {
     private CompletableFuture<DetectorTask> handleSequencer(Layout layout) {
         log.trace("Handling sequencer failures");
 
-        if (localMonitoringService.getMetrics().join().getSequencerStatus() == SequencerStatus.READY) {
-            log.trace("Primary sequencer is ready. Nothing to do");
+        ClusterState clusterState = clusterContext.getClusterView();
+        Optional<NodeState> primarySequencer = clusterState.getNode(layout.getPrimarySequencer());
+        if (primarySequencer.isPresent() && primarySequencer.get().getSequencerMetrics() == SequencerMetrics.READY) {
+            log.trace("Primary sequencer is already ready at: {} in {}", primarySequencer.get(), clusterState);
             return DETECTOR_TASK_SKIPPED;
         }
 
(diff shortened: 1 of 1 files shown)


xnull:
Can you please add more reviewers, it always useful to bring more developers to discussion

hisundar:
Sure, I pushed an incorrect patch by mistake. I will add reviewers once I upload the correct patch meant for review.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClusterState.java
```diff
@@ -69,4 +69,22 @@ public boolean isReady(){
...
+     * See if cluster is ready. If so, check if the cluster has at least 1 primary sequencer that's READY.
+     *
+     * @return if primary sequencer is ready
+     */
+    public boolean isPrimarySequencerReady() {
+        if (nodes.isEmpty()) {
+            return false;
+        }
+        for (NodeState nodeState : nodes.values()) {
+            if (nodeState.getSequencerMetrics().getSequencerStatus() == SequencerMetrics.SequencerStatus.READY) {
```
zalokhan: At any given point in time, there can only be 1 primary sequencer. You need to pass the endpoint of the primary sequencer currently active in the current layout and only check it's status. The return type of this method should be SequencerStatus corresponding to the status of the primary sequencer.

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -563,8 +563,14 @@ private DetectorTask detectFailure(PollReport pollReport) {
             return DETECTOR_TASK_SKIPPED;
         }
 
+        ClusterState clusterState = clusterContext.getClusterView();
+        if (clusterState.isPrimarySequencerReady()) {
```
StrikingThirteen: Would it be possible to bring it before L544 to avoid other checks as well?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClusterState.java
hisundar: What is the best way to find out the endpoint of the primary sequencer from the nodes?

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
hisundar: The reason I left it here is because this check examines the cluster state map which I thought may be more expensive compared to the other checks which appear to be cheaper in execution time. Please correct me if I am mistaken? thanks

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClusterState.java
zalokhan: You can find it from the layout.
`Layout.getPrimarySequencer`

hisundar: Thank you!

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -563,8 +564,15 @@ private DetectorTask detectFailure(PollReport pollReport) {
             return DETECTOR_TASK_SKIPPED;
         }
 
+        ClusterState clusterState = clusterContext.getClusterView();
```
zalokhan: You would want to shift this check just after the first if condition:
```if (localMonitoringService.getMetrics().join().getSequencerStatus() == SequencerStatus.READY)```

This is because we do not want to increment the sequencerNotReadyCounter if the primary sequencer is already bootstrapped.

zalokhan: Since now you are already checking if the primary sequencer is bootstrapped for a particular epoch, you can also remove the first if condition check ```if (localMonitoringService.getMetrics().join().getSequencerStatus() == SequencerStatus.READY)```

hisundar: Thank you so much for the insights.

hisundar:
Rebased. Thanks

Issue #1788, opened 2019-03-15
Netty channel leak in BootstrapUtil and IT


WenbinZhu:
## Overview

The BootstrapUtil is creating NettyClientRouter on each retry and not correctly shutting down them on failure. In most cases, bootstrap servers will have fail 1 or 2 times because server channel might not be fully established yet, and if the NettyClientRouter is not correctly shutdown, then these channels are leaked. They are still connected to the servers, and when some server goes down, theses channels will constantly reconnect  every second. Also, the integration tests are not closing the channels when tests complete.

WenbinZhu:
closed via #1789

PR #1789, opened 2019-03-15
Prevent netty channel leak

WenbinZhu:
## Overview

- Properly shutdown the netty routers in BootstapUtil when exception
happens. This prevents channel leakage, which may cause an issue that
after the boostrap finishes and then some server goes down, the leaked
channels are constantly reconnecting every second.

- Integrations tests are not shutting down runtime after tests finish,
which is also leaking channels.

- Removed all NetworkExceptions catches when creating NettyClientRouter,
since it is needed only for legacy versions.

Related issue(s) (if applicable): #1788

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
+++ b/runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
@@ -107,19 +107,18 @@ public class BootstrapUtil {
                                  @NonNull Duration retryTimeout) {
         for (String server : layout.getAllServers()) {
             int retry = retries;
+
+            IClientRouter router = new NettyClientRouter(NodeLocator.parseString(server),
+                    corfuRuntimeParameters);
+            router.addClient(new LayoutHandler())
+                    .addClient(new ManagementHandler())
+                    .addClient(new BaseHandler());
+
             while (retry-- > 0) {
                 try {
                     log.info("Attempting to bootstrap node:{} with layout:{}", server, layout);
-                    IClientRouter router = new NettyClientRouter(NodeLocator.parseString(server),
-                            corfuRuntimeParameters);
-                    router.addClient(new LayoutHandler())
-                            .addClient(new ManagementHandler())
-                            .addClient(new BaseHandler());
-
                     bootstrapLayoutServer(router, layout);
                     bootstrapManagementServer(router, layout);
-
-                    router.stop();
                     break;
                 } catch (AlreadyBootstrappedException abe) {
                     log.error("Bootstrapping node: {} failed with exception:", server, abe);
@@ -134,7 +133,9 @@ public class BootstrapUtil {
                     Sleep.MILLISECONDS.sleepUninterruptibly(retryTimeout.toMillis());
                 }
             }
+
+            router.stop();
         }
-        log.info("Bootstrapping layout:{} successful.", layout);
+        log.info("Successfully Bootstrapped layout:{} .", layout);
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
@@ -211,8 +211,7 @@ public class NettyClientRouter extends SimpleChannelInboundHandler<CorfuMsg>
         b.channel(parameters.getSocketType().getChannelClass());
         parameters.getNettyChannelOptions().forEach(b::option);
         b.handler(getChannelInitializer());
-        b.option(ChannelOption.CONNECT_TIMEOUT_MILLIS,
-                (int) parameters.getConnectionTimeout().toMillis());
+        b.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, (int) timeoutConnect);
 
         // Asynchronously connect, retrying until shut down.
         // Once connected, connectionFuture will be completed.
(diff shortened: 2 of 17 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -340,16 +339,14 @@ private void addReconnectionOnCloseFuture(@Nonnull Channel channel,
      * @param bootstrap         The channel boostrap to use
      * @return                  A {@link ChannelFuture} which is c
      */
-    private ChannelFuture connectAsync(@Nonnull Bootstrap bootstrap) {
+    private void connectAsync(@Nonnull Bootstrap bootstrap) {
         // If shutdown, return a ChannelFuture that is exceptionally completed.
         if (shutdown) {
-            return new DefaultChannelPromise(channel, GlobalEventExecutor.INSTANCE)
```
zalokhan: Does this only return an exceptionally completed channel promise? It doesn't close the `channel` right?

test/src/test/java/org/corfudb/integration/MemoryFootprintIT.java
```diff
@@ -28,6 +29,15 @@
     private static int corfuStringNodePort;
     private static String singleNodeEndpoint;
 
+    private CorfuRuntime runtime;
```
zalokhan: Would it make more sense to move this repeated code to AbstractIT?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
WenbinZhu: No It does't close the channel.

test/src/test/java/org/corfudb/integration/MemoryFootprintIT.java
WenbinZhu: Yes, I'm trying to move it to the cleanUp() method in AbstractIT.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -216,6 +226,8 @@ public void transactionsSpanningReconfigurationTest() throws Exception {
         }
 
         assertThat(txnFailed).isTrue();
+
+        runtime.shutdown();
```
hisundar: Should runtime be set to null to prevent double shutdown during shutdownRuntime() ?

WenbinZhu: Good catch! will move to AbstractIT and address it.

test/src/test/java/org/corfudb/integration/MemoryFootprintIT.java
WenbinZhu: Done.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
WenbinZhu: Done, set to null in setUp() method which will be called before any test.

runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
```diff
@@ -107,19 +107,18 @@ public static void bootstrap(@NonNull Layout layout,
...
+            IClientRouter router = new NettyClientRouter(NodeLocator.parseString(server),
+                    corfuRuntimeParameters);
+            router.addClient(new LayoutHandler())
+                    .addClient(new ManagementHandler())
+                    .addClient(new BaseHandler());
+
             while (retry-- > 0) {
                 try {
                     log.info("Attempting to bootstrap node:{} with layout:{}", server, layout);
-                    IClientRouter router = new NettyClientRouter(NodeLocator.parseString(server),
```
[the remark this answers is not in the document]
WenbinZhu: It was previously recreating because the creating router method used to throw NetworkException.  Now it doesn't so we don't need to recreate router every retry.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -340,16 +339,14 @@ private void addReconnectionOnCloseFuture(@Nonnull Channel channel,
      * @param bootstrap         The channel boostrap to use
      * @return                  A {@link ChannelFuture} which is c
      */
-    private ChannelFuture connectAsync(@Nonnull Bootstrap bootstrap) {
+    private void connectAsync(@Nonnull Bootstrap bootstrap) {
```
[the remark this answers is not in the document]
WenbinZhu: Fixed comment.

test/src/test/java/org/corfudb/integration/CmdletIT.java
```diff
@@ -6,6 +6,7 @@
 import org.corfudb.runtime.collections.SMRMap;
 import org.corfudb.runtime.view.Layout;
 import org.corfudb.runtime.view.stream.IStreamView;
+import org.junit.After;
```
zalokhan: Not used.

test/src/test/java/org/corfudb/integration/CorfuMsgFilterIT.java
```diff
@@ -8,6 +8,7 @@
 import org.corfudb.protocols.wireprotocol.MsgHandlingFilter;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuRuntime.CorfuRuntimeParameters;
+import org.junit.After;
```
zalokhan: Not used.

test/src/test/java/org/corfudb/integration/LargeWriteIT.java
```diff
@@ -5,6 +5,7 @@
 import org.corfudb.runtime.exceptions.AbortCause;
 import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.exceptions.WriteSizeException;
+import org.junit.After;
```
zalokhan: Not used.

test/src/test/java/org/corfudb/integration/CmdletIT.java
WenbinZhu: Done.

test/src/test/java/org/corfudb/integration/CorfuMsgFilterIT.java
WenbinZhu: Done.

test/src/test/java/org/corfudb/integration/LargeWriteIT.java
WenbinZhu: Done. Cleared unused imports in other ITs.

zalokhan:
Looks like a UI issue. Travis has passed for this patch:
https://travis-ci.org/CorfuDB/CorfuDB/builds/507956581
I am proceeding to merge this patch.

PR #1792, opened 2019-03-18
Replace null with empty collections

StrikingThirteen:
## Overview

Description:
Replace returning null with empty collections
Make constants final
Combine nested if loops

Why should this be merged: 
Avoid returning nulls
Constants should be final

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenResponse.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenResponse.java
@@ -17,8 +17,8 @@ import lombok.Data;
 @AllArgsConstructor
 public class TokenResponse implements ICorfuPayload<TokenResponse>, IToken {
 
-    public static byte[] NO_CONFLICT_KEY = new byte[]{};
-    public static UUID NO_CONFLICT_STREAM = new UUID(0, 0);
+    public static final byte[] NO_CONFLICT_KEY = new byte[]{};
+    public static final UUID NO_CONFLICT_STREAM = new UUID(0, 0);
 
     /**
      * Constructor for TokenResponse.
--- a/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
@@ -7,7 +7,6 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import com.google.common.annotations.VisibleForTesting;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.ISMRConsumable;
 import org.corfudb.protocols.logprotocol.SMREntry;
@@ -55,14 +54,14 @@ public class StreamViewSMRAdapter implements ISMRStream {
             // This is a CHECKPOINT record.  Extract the SMREntries, if any.
             CheckpointEntry cp = (CheckpointEntry) logData.getPayload(runtime);
             if (cp.getSmrEntries() != null
-                    && cp.getSmrEntries().getUpdates().size() > 0) {
+                    && !cp.getSmrEntries().getUpdates().isEmpty()) {
                 cp.getSmrEntries().getUpdates().forEach(e -> {
                     e.setRuntime(runtime);
                     e.setEntry(logData);
                 });
                 return cp.getSmrEntries().getUpdates();
             } else {
-                return (List<SMREntry>) Collections.EMPTY_LIST;
+                return Collections.emptyList();
             }
         } else {
             return ((ISMRConsumable) logData.getPayload(runtime)).getSMRUpdates(streamView.getId());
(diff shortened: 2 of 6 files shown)


runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenResponse.java
```diff
@@ -17,8 +17,8 @@
 @AllArgsConstructor
 public class TokenResponse implements ICorfuPayload<TokenResponse>, IToken {
 
-    public static byte[] NO_CONFLICT_KEY = new byte[]{};
-    public static UUID NO_CONFLICT_STREAM = new UUID(0, 0);
+    public final static byte[] NO_CONFLICT_KEY = new byte[]{};
```
xnull: static final?

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -98,14 +97,12 @@ public void gc(long trimMark) {
...
-                return ((ISMRConsumable) data.getPayload(runtime))
-                        .getSMRUpdates(streamView.getId());
-            }
+        if (data != null && data.getType() == DataType.DATA
+                && data.getPayload(runtime) instanceof ISMRConsumable) {
+            return ((ISMRConsumable) data.getPayload(runtime))
+                    .getSMRUpdates(streamView.getId());
         }
-        return null;
+        return Collections.emptyList();
```
xnull: have you checked all the places executing this code? If they have checks for null, we should be very careful here, we can break some of them silently

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
```diff
@@ -92,8 +94,8 @@
     default @Nonnull
     Map<Long, ILogData> readRange(RuntimeLayout runtimeLayout, Set<Long> globalAddresses) {
         return globalAddresses.parallelStream()
-                .map(a -> new AbstractMap.SimpleImmutableEntry<>(a, read(runtimeLayout, a)))
-                .collect(Collectors.toMap(r -> r.getKey(), r -> r.getValue()));
+                .map(a -> new SimpleImmutableEntry<>(a, read(runtimeLayout, a)))
```
xnull: it seems to me you don't need `map` here

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
```diff
@@ -128,8 +130,8 @@
     default @Nonnull Map<Long, ILogData> peekAll(RuntimeLayout runtimeLayout,
                                                  Set<Long> globalAddresses) {
         return globalAddresses.parallelStream()
-                .map(a -> new AbstractMap.SimpleImmutableEntry<>(a, peek(runtimeLayout, a)))
-                .collect(Collectors.toMap(r -> r.getKey(), r -> r.getValue()));
+                .map(a -> new SimpleImmutableEntry<>(a, peek(runtimeLayout, a)))
```
xnull: seems, don't need map method

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenResponse.java
StrikingThirteen: done

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
StrikingThirteen: done

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
StrikingThirteen: done

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
StrikingThirteen: Yes, checked all usages. They are are in current and previous and the change not breaking them.

lixinchengdu: The comment should be updated too.

StrikingThirteen: done

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -580,7 +582,7 @@ protected void rollbackStreamUnsafe(ISMRStream stream, long rollbackVersion) {
 
         List<SMREntry> entries = stream.current();
 
-        while (entries != null) {
+        while (entries != null && !entries.isEmpty()) {
```
zalokhan: You are overriding the empty returned by the existing implementation of `current()` and `previous()`. We need to make sure it doesn't conflict with that.

Maithem: If you want to make this change, then change the lower layers to return empty. Using both null and predicates seem to go against what this PR proposes (i.e. use predicates instead of null)

Maithem:
Superseded by #2033

PR #1793, opened 2019-03-21
Remove unused exceptions and dead code

hisundar:
## Overview

Description: A few exceptions don't appear to be thrown anywhere in the repo.

Why should this be merged: Removing dead code with unused exceptions to avoid confusions.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/exceptions/GarbageCollectorException.java
```diff
@@ -1,15 +0,0 @@
-package org.corfudb.runtime.exceptions;
-
-/**
- *
- * An exception for any exception that is thrown during
- * garbage (on the runtime) collection.
- *
- * Created by maithem on 11/16/18.
```
StrikingThirteen: Considering this one is added relatively recently (end of 2018), can you check with @Maithem to confirm it is obsolete?

hisundar: Sure, actually @vjeko removed this exception as part of #1669, but will check with @Maithem too. thanks

test/src/test/java/org/corfudb/runtime/collections/FGMapTest.java
```diff
@@ -39,26 +39,6 @@ public void canReadWriteToSingle()
                 .isEqualTo("b");
     }
 
-    /*
```
zalokhan: I am not sure if FGMaps are used or plan to be supported in future. If we get an agreement lets remove FGMaps too.

hisundar: I thought about that too, but felt that would be too invasive, so chickened out to just removing commented code for now :)

hisundar: Looks like FGMaps are used by CorfuDB-Helloworld repo. If I remove that it will break that example.

xnull: maybe it's time to remove corfu-helloworld repo? :)

hisundar: Ok, if we are sure that FGMap is not required, I can remove it here, but I do not have the admin privileges for the corfu-helloworld repo. Also should we fix that repo with a better example given FGMap isn't supported?

xnull: Discuss with Maithem

Maithem: I wouldn't remove it. People can use it as an example of how to build a composable SMR object.

zalokhan:
@hisundar Can you also remove the migration module in this PR?
That was only required to upgrade to 0.2.0 I believe.

hisundar:
> @hisundar Can you also remove the migration module in this PR?
> That was only required to upgrade to 0.2.0 I believe.

Done. Could you please help merge the same? thanks

PR #1794, opened 2019-03-22
Cache-After-Write Race

Maithem:
## Overview

Why should this be merged: Fixes a cache race bug that causes
objects to resync (i.e. NoRollBackException)

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -61,6 +61,7 @@ public class AddressSpaceView extends AbstractView {
             .expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .expireAfterWrite(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .recordStats()
+            .concurrencyLevel(Runtime.getRuntime().availableProcessors() * 2)
             .build(new CacheLoader<Long, ILogData>() {
                 @Override
                 public ILogData load(Long value) throws Exception {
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -169,9 +170,19 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
 
             // Do the write
             try {
-                l.getReplicationMode(token.getSequence())
-                        .getReplicationProtocol(runtime)
-                        .write(e, ld);
+                // Cache the successful write
```
WenbinZhu: I think the comment here should have more details, sth. like atomically write to server and then cache to prevent race condition that can overwrite the undo record.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -61,6 +61,7 @@
             .expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .expireAfterWrite(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .recordStats()
+            .concurrencyLevel(Runtime.getRuntime().availableProcessors() * 2)
```
vjeko: Can you please explain how this is related to NoRollBackException fix?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
vjeko: Ditto.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: This addresses the side effect of locking. Internally, the cache partitions the address space into buckets and each bucket has a reader/writer lock. The size of these buckets (i.e. key-partition space size) affects concurrency on the global cache. For example, if there's one bucked then there's only serial access to the cache. Conversely, if there's as many locks as keys then the cache is fully parallel (with respect to key access). In essence, this is a trade-off between memory usage and concurrency and needs to be tuned, setting CPU * 2 is just a "reasonable guess".

Maithem:
To be addressed in a bigger change related to the stream layer. To track it, I created this issue #1798

Issue #1796, opened 2019-03-22
Nested transaction can have different type, which shouldn't be allowed


WenbinZhu:
## Overview

Currently when starting a nested transaction using  `runtime.getObjectsView().type("TransactionType.XXX").build().begin()` it can bypass the transaction type inherit logic, which ends up in nested transactions with different transaction types. This could violate transaction semantics, though in some cases, the type mismatch will be detected upon txn commit, but in other cases, inconsistency will occur without being noticed.

lixinchengdu:
What is the problem of having different transaction types? For example, it is possible to have a snapshot transaction nested in an optimistic transaction.

WenbinZhu:
@lixinchengdu In your example, yes, but you can't do it vice versa

Maithem:
I think for the sake of simplicity we should just allow transactions of the same type to nest.
Transaction nesting is kind of broken, so I'm guessing that this is just sanity check for now?

WenbinZhu:
@Maithem Yes, for simplicity, just a sanity check and throw an exception if type mismatch.

WenbinZhu:
Addressed in #1797

PR #1797, opened 2019-03-22
Check nested transaction type mismatch upon txn begin

WenbinZhu:
## Overview

- Currently when starting nested transactions, client can bypass the txn
type inheritance, and start nested transactions with different types,
which could violate txn semantics. This patch enforces a txn type check
upon transaction start.

Related issue(s) (if applicable): #1796 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/transactions/Transaction.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/transactions/Transaction.java
@@ -54,22 +54,32 @@ public class Transaction {
     }
 
     /**
-     * Verifies that this transaction  has a valid snapshot and context.
+     * Verifies that this transaction has a valid snapshot and context.
      */
     private void verify() {
         final AbstractTransactionalContext parentCtx = TransactionalContext.getCurrentContext();
         if (parentCtx != null) {
+            // In a nested transaction, the runtime should be the same as the parent's
             if (parentCtx.getTransaction().getRuntime() != this.runtime) {
-                throw new IllegalStateException("Attempting to nest transactions from different clients!");
+                throw new IllegalStateException(
+                        "Attempting to nest transactions from different clients!");
             }
-            // If we're in a nested transaction, the first read timestamp
-            // needs to come from the root.
+
+            // In a nested transaction, the transaction type should be the same as the parent's
+            if (type != parentCtx.getTransaction().getType()) {
+                String msg = String.format("Attempting to nest transactions with a " +
+                                "different transaction type, parent type=%s, nested type=%s",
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/object/transactions/Transaction.java
```diff
@@ -54,22 +54,30 @@ public boolean isLoggingEnabled() {
...
+                throw new IllegalStateException(
+                        "Attempting to nest transactions from different clients!");
             }
-            // If we're in a nested transaction, the first read timestamp
-            // needs to come from the root.
+
+            // In a nested transaction, the transaction type should be the same as the parent's
+            if (type != parentCtx.getTransaction().getType()) {
+                throw new IllegalArgumentException(
+                        "Attempting to nest transactions with a different transaction type!");
```
annym: should we add info on the types mismatched?

WenbinZhu: Done, added additional logging.

runtime/src/main/java/org/corfudb/runtime/object/transactions/Transaction.java
```diff
@@ -54,22 +54,32 @@ public boolean isLoggingEnabled() {
...
+                throw new IllegalStateException(
+                        "Attempting to nest transactions from different clients!");
             }
-            // If we're in a nested transaction, the first read timestamp
-            // needs to come from the root.
+
+            // In a nested transaction, the transaction type should be the same as the parent's
+            if (type != parentCtx.getTransaction().getType()) {
+                String msg = String.format("Attempting to nest transactions with a " +
+                                "different transaction type, parent type=%s, child type=%s",
```
StrikingThirteen: Maybe nested instead of child transaction?

WenbinZhu: Done.

WenbinZhu:
> Would it be possible that the criteria for a nested transaction type is stronger than the parent's and hence possible to allow such cases?

It could, but for simplicity, we just force it for now. Because anyway the current nested transaction is kind of broken in some other places, may need to remove the support of nested transaction in the future.

WenbinZhu:
> We used to have this check in ObjectsView, but some how go lost when the Transaction builder was refactored. This PR actually fixes a bug in the current code, where if you start transaction and call a SMRMap operation that implicitly starts a transactions, you can end up nesting transactions with different types.
> 
> Can you add a test to nest transactions with different types?

@Maithem Done, added a unit test.

Issue #1751, opened 2019-02-20
Premature client router shutdown


zalokhan:
## Overview

The runtime runs the `fetchLayout` whenever a layout is required. With this action, we also piggyback `pruneRemovedRouters`. 
`pruneRemovedRouters` detects all routers which have a connection to nodes which are not present in the layout. These routers are shut down and removed.

In order to access a node outside the layout (example: to bootstrap or reset) we might need to create a router to connect to them. Either the `pruneRemovedRouters` should take care of routers in use or these calls should not be made via the runtime.

annym:
This is the same Issue as #1822 which will be addressed by @WenbinZhu

PR #1804, opened 2019-03-29
Fix Netty router race and channel leak.

WenbinZhu:
## Overview

- Fix a race condition when pruning routers that are not in layout, someone may
still using it which causes a ShutdownException
- Fix a channel leak issue that channels are not closed when stopping routers.

Related issue(s) (if applicable): #1751, #1822 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -864,7 +864,9 @@ public class CorfuRuntime {
                         .contains(NodeLocator.getLegacyEndpoint(endpoint)))
                 .forEach(endpoint -> {
                     try {
-                        nodeRouterPool.getNodeRouters().get(endpoint).stop();
+                        // Simply remove the router from the pool and let finalize() method close
+                        // the channel when being GC'ed in order to prevent race conditions where
+                        // someone may still holding this router and planing to send messages.
                         nodeRouterPool.getNodeRouters().remove(endpoint);
                     } catch (Exception e) {
                         log.warn("fetchLayout: Exception in stopping and removing "
--- a/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
@@ -11,7 +11,6 @@ import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
-import io.netty.channel.DefaultChannelPromise;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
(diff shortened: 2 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -864,7 +864,9 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
                         .contains(NodeLocator.getLegacyEndpoint(endpoint)))
                 .forEach(endpoint -> {
                     try {
-                        nodeRouterPool.getNodeRouters().get(endpoint).stop();
+                        // Simply remove the router from the pool and let finalize() method close
```
xnull: but stop() method does some actions, do we lose those actions with this change?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -394,6 +395,15 @@ public void stop(boolean shutdown) {
         stop();
     }
 
+    /**
+     * {@inheritDoc
+     */
+    @Override
+    protected void finalize() throws Throwable {
```
xnull: don't use finalize, it's a bad practice

xnull: https://howtodoinjava.com/java/basics/why-not-to-use-finalize-method-in-java/

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
WenbinZhu: That's why I use finalize() to clean up. stop() is removed here because it can create a race where others are holding this router, and if it sends a message with this stoped router, it gets a ShutdownException.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
WenbinZhu: I'm fully aware of these caveats when I create this path. But this is the only approach for now to solve this race condition and meanwhile clean up the resources (i.e. channels) without relying on client invoking it.

The biggest argument in this link is that "finalize() execution is not guaranteed", which I already thought about. However this doesn't matter, even if it's not executed, the result is just some channels remain opening. Still it's better than not providing finalize(). 

My point is, providing finalize() is giving it another chance to clean up the resource, even it's not executed, just leave it. I'm willing to discuss any better ideas if you have, otherwise I think finalize() is ok to be used here.

WenbinZhu:
> Let's not use finalize

See my previous comment.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
xnull: I don't see any reasons to use finalize, I also don't see how finalize() can sole the "race condition" issue

xnull: Can you please find a better way, or/and maybe redesign this code, to solve the issue without finalize

WenbinZhu: If you want to know the reason to use finalize to solve the race condition, I can explain to you offline, I've already spent some time thinking about better approaches before I go with finalize, but looks like this is the best and cheapest way. Maybe you will have a better idea after we discuss.

WenbinZhu:
Please review, thanks.

WenbinZhu:
@xnull based on our offline discussion, there is no other easy way to fix the issue while cleaning up the resource. Please suggest other valid ideas, otherwise I think finalize is ok to use here as it's just a janitor.

xnull:
I won't allow this PR to be merged, that it. Anything else up to you guys

WenbinZhu:
Address in #1843, add a new utility for user to prevent race

Issue #1798, opened 2019-03-25
Cache-After-Write Race 


Maithem:
**Bug report.**
Reader and writer threads within the same client need to be synchronized to prevent expensive cache misses. For example, a writer thread will only cache a writer after it has been successfully replicated, this introduces a race where other read threads will discover the update and bring in into the cache before the original writer thread. This interleaving can cause undo records to be lost affecting other transactions: causing them to experience NoRollBackExceptions. 

One way to address this problem is to lock on global addresses forcing readers to wait on the writing thread for a particular address, resulting in a subsequent cache hit (for example #1794)

PR #1805, opened 2019-03-29
Cache Race Fix

Maithem:
## Overview

Why should this be merged: Fixes a cache race bug that causes
objects to resync (i.e. NoRollBackException)

Related issue(s) (if applicable): #1798

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -5,6 +5,7 @@ import static org.corfudb.util.Utils.getTails;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
@@ -35,8 +36,10 @@ import org.corfudb.util.Sleep;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.time.Duration;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -56,22 +59,12 @@ public class AddressSpaceView extends AbstractView {
     /**
      * A cache for read results.
      */
-    final LoadingCache<Long, ILogData> readCache = CacheBuilder.newBuilder()
+    final Cache<Long, ILogData> readCache = CacheBuilder.newBuilder()
             .maximumSize(runtime.getParameters().getNumCacheEntries())
             .expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .expireAfterWrite(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .recordStats()
-            .build(new CacheLoader<Long, ILogData>() {
-                @Override
-                public ILogData load(Long value) throws Exception {
-                    return cacheFetch(value);
-                }
-
-                @Override
-                public Map<Long, ILogData> loadAll(Iterable<? extends Long> keys) throws Exception {
-                    return cacheFetch((Iterable<Long>) keys);
-                }
-            });
+            .build();
 
     /**
      * Constructor for the Address Space View.
@@ -230,31 +223,42 @@ public class AddressSpaceView extends AbstractView {
      * @param address An address to read from.
      * @return A result, which be cached.
      */
-    public @Nonnull ILogData read(long address) {
+    public @Nonnull
+    ILogData read(long address) {
         if (!runtime.getParameters().isCacheDisabled()) {
-            ILogData data;
-            try {
-                data = readCache.get(address);
-            } catch (ExecutionException | UncheckedExecutionException e) {
-                // Guava wraps the exceptions thrown from the lower layers, therefore
-                // we need to unwrap them before throwing them to the upper layers that
-                // don't understand the guava exceptions
-                Throwable cause = e.getCause();
-                if (cause instanceof RuntimeException) {
-                    throw (RuntimeException) cause;
-                } else {
-                    throw new RuntimeException(cause);
-                }
+            // The VersionLockedObject and the Transaction layer will generate
+            // undoRecord(s) during a transaction commit, or object sync. These
+            // undo records are stored in transient fields and are not persisted.
+            // A missing undo record can cause a NoRollbackException, thus forcing
+            // a complete object rebuild that generates a "scanning" behavior
+            // which affects the LRU window. In essence, affecting other cache users
+            // and making the VersionLockedObject very sensitive to caching behavior.
+            // A concrete example of this would be unsynchronized readers/writes:
+            // 1. Thread A starts replicating write1
+            // 2. Thread B discovers the write (via stream tail query) and
+            //    tries to read write1
+            // 3. Thread B's read results in a cache miss and the reader thread
+            //    starts loading the value into the cache
+            // 4. Thread A completes its write and caches it with undo records
+            // 5. Thread B finishes loading and caches the loaded value replacing
+            //    the cached value from step 4 (i.e. loss of undo records computed
+            //    by thread A)
+            ILogData data = readCache.getIfPresent(address);
+            if (data == null) {
+                // Loading a value without the cache loader can result in
+                // redundant loading calls (i.e. multiple threads try to
+                // load the same value), but currently a redundant RPC
+                // is much cheaper than the cost of a NoRollBackException, therefore
+                // this trade-off is reasonable
+                final ILogData loadedVal = fetch(address);
+                data = readCache.asMap().computeIfAbsent(address, (k) -> loadedVal);
+                return data;
+            } else {
+                return data;
             }
-            if (data == null || data.getType() == DataType.EMPTY) {
-                throw new RuntimeException("Unexpected return of empty data at address "
-                        + address + " on read");
-            } else if (data.isTrimmed()) {
-                throw new TrimmedException();
-            }
-            return data;
+        } else {
+            return fetch(address);
         }
-        return fetch(address);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -56,22 +59,12 @@
     /**
      * A cache for read results.
      */
-    final LoadingCache<Long, ILogData> readCache = CacheBuilder.newBuilder()
+    final Cache<Long, ILogData> readCache = CacheBuilder.newBuilder()
```
Maithem: Now that loading happens in one step, add more concurrency.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -234,8 +227,9 @@ public void write(IToken token, Object data) throws OverwriteException {
         if (!runtime.getParameters().isCacheDisabled()) {
             ILogData data;
             try {
-                data = readCache.get(address);
-            } catch (ExecutionException | UncheckedExecutionException e) {
+                data = readCache.asMap().computeIfAbsent(address, this::fetch);
```
WenbinZhu: Maybe a comment here why using computeIfAbsent and how it helps with the race condition.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -56,22 +59,12 @@
...
-    final LoadingCache<Long, ILogData> readCache = CacheBuilder.newBuilder()
+    final Cache<Long, ILogData> readCache = CacheBuilder.newBuilder()
             .maximumSize(runtime.getParameters().getNumCacheEntries())
             .expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .expireAfterWrite(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .recordStats()
-            .build(new CacheLoader<Long, ILogData>() {
-                @Override
-                public ILogData load(Long value) throws Exception {
-                    return cacheFetch(value);
```
WenbinZhu: You can remove the implementation of this method as it's not used.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -264,11 +258,30 @@ public void write(IToken token, Object data) throws OverwriteException {
      * @return A result, which be cached.
      */
     public Map<Long, ILogData> read(Iterable<Long> addresses) {
-        Map<Long, ILogData> addressesMap;
+        Map<Long, ILogData> addressesMap = new HashMap<>();
         if (!runtime.getParameters().isCacheDisabled()) {
             try {
-                addressesMap = readCache.getAll(addresses);
-            } catch (ExecutionException | UncheckedExecutionException e) {
+                Set<Long> addressesToLoadFetch = new HashSet();
```
WenbinZhu: s/HashSet()/HashSet<>()

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -234,7 +228,11 @@ public void write(IToken token, Object data) throws OverwriteException {
         if (!runtime.getParameters().isCacheDisabled()) {
             ILogData data;
             try {
-                data = readCache.get(address);
+                data = readCache.getIfPresent(address);
+                if (data == null) {
+                    final ILogData loadedVal = fetch(address);
+                    data = readCache.get(address, () -> loadedVal);
```
annym: For future reference can you please add a comment why we check the read cache again upon return, and how it relates to NRE.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -56,22 +59,13 @@
     /**
      * A cache for read results.
      */
-    final LoadingCache<Long, ILogData> readCache = CacheBuilder.newBuilder()
+    final Cache<Long, ILogData> readCache = CacheBuilder.newBuilder()
             .maximumSize(runtime.getParameters().getNumCacheEntries())
             .expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .expireAfterWrite(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
+            .concurrencyLevel(Runtime.getRuntime().availableProcessors() * 2)
```
Maithem: Since no locks are acquired while loading, this is no longer required.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: cacheFetch actually had extra logic, which checks if the ILogData type is empty. I collapsed the two implementations, this adds more consistency.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -246,6 +246,7 @@ private long logUpdateInner(String smrUpdateFunction, final boolean keepUpcallRe
         // We need to add the acquired token into the pending upcall list.
         SMREntry smrEntry = new SMREntry(smrUpdateFunction, args, serializer);
         long address = underlyingObject.logUpdate(smrEntry, keepUpcallResult);
+        log.error("logUpdateInner: Non-transactional write at {}", smrEntry.getEntry());
```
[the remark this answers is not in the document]
Maithem: Removed.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Done. Also cleaned up this method.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -232,9 +225,35 @@ public void write(IToken token, Object data) throws OverwriteException {
...
+                //    the cached value from step 4 (i.e. loss of undo records computed
+                //    by thread A)
+                ILogData data = readCache.getIfPresent(address);
+                if (data == null) {
+                    // Loading a value without the cache loader can result in
+                    // redundant loading calls (i.e. multiple threads try to
+                    // load the same value), but currently a redundant RPC
+                    // is much cheaper than the cost of a NoRollBackException, therefore
+                    // this trade-off is reasonable
+                    final ILogData loadedVal = fetch(address);
```
vjeko: This is effectively your synchronization barrier? After fetch(...) returns, you are guaranteed that the value will be populated in the cache, assuming the writer is utilizing the same cache. Am I understanding this correctly?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -264,29 +276,35 @@ public void write(IToken token, Object data) throws OverwriteException {
...
+            // At this point we computed a subset of the addresses that
+            // resulted in a cache miss and need to be fetched
+            if (!addressesToFetch.isEmpty()) {
+                Map<Long, ILogData> fetchedAddresses = fetchAll(addressesToFetch);
+                for (Map.Entry<Long, ILogData> entry : fetchedAddresses.entrySet()) {
+                    // After fetching a value, we need to insert it in the cache.
+                    // Note that based on code inspection it seems like operations
+                    // on the cache's map view are reflected in the cache's statistics.
+                    result.put(entry.getKey(), readCache.asMap()
+                            .computeIfAbsent(entry.getKey(), (k) -> entry.getValue()));
```
xnull: The operation seems not atomic.
What if during your fetching all missing entries some entries were added? You will replace those entries by your fetched ones

xnull: Actually `computeIfAbsent` should prevent that.
Looks ok, no worries

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -463,20 +465,44 @@ public void invalidateClientCache() {
...
+     * returns null, Empty, or trimmed an exception will be
+     * thrown.
+     * @param address The address being checked
+     * @param logData the IlogData at the address being checked
+     */
+    private void checkLogData(long address, ILogData logData) {
+        if (logData == null || logData.getType() == DataType.EMPTY) {
+            throw new RuntimeException("Unexpected return of empty data at address "
+                    + address + " on read");
+        } else if (logData.isTrimmed()) {
```
xnull: extra `else` statement, please remove

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -232,9 +225,35 @@ public void write(IToken token, Object data) throws OverwriteException {
...
+                ILogData data = readCache.getIfPresent(address);
+                if (data == null) {
+                    // Loading a value without the cache loader can result in
+                    // redundant loading calls (i.e. multiple threads try to
+                    // load the same value), but currently a redundant RPC
+                    // is much cheaper than the cost of a NoRollBackException, therefore
+                    // this trade-off is reasonable
+                    final ILogData loadedVal = fetch(address);
+                    data = readCache.get(address, () -> loadedVal);
+                    return data;
```
annym: We are missing to return the data if it is present in the cache, right?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Not sure which line you commented on, but the synchronization happens here:
`data = readCache.get(address, () -> loadedVal);`
Internally, guava will lock and will return loadedVal only if the mapping doesn't exist at that time. If it does, then guava will not even consider loadedVal. In other words, the fetch in line 253 becomes a no-op.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: If the mapping exists in the cache guava will return the old value and loadedVal will be ignored.
Essentially, line 254 will be 
`data = existing mapping (not loadedVal)`

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: The fetch phase and cache insert are not atomic, but the remapping of the values in the cache are. As a consequence, there could be redundant fetches, but not redundant cache remapping.

xnull: can we use `readCache .get(...)` here?

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
```diff
@@ -98,6 +104,13 @@ public ILogData peek(RuntimeLayout runtimeLayout, long address) {
         }
     }
 
+    @Override
+    public Map<Long, ILogData> readRange(RuntimeLayout runtimeLayout, Set<Long> globalAddresses) {
+        return globalAddresses.parallelStream()
```
xnull: why parallel stream?

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
```diff
@@ -98,6 +104,13 @@ public ILogData peek(RuntimeLayout runtimeLayout, long address) {
...
+                .map(a -> new AbstractMap.SimpleImmutableEntry<>(a, read(runtimeLayout, a)))
+                .collect(Collectors.toMap(r -> r.getKey(), r -> r.getValue()));
+    }
```
xnull: globalAddresses.stream()
                .collect(Collectors.toMap(addr -> addr, addr -> read(runtimeLayout, addr)));

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: I was referring to line 246, if it is already present you can return it... otherwise It'll exercise the fetch and go to the log unit again.

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
Maithem: Just pushing code from the abstract class to the  concrete implementation. This used to be the abstract class' default implementation.

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
Maithem: The parallel stream is there for parallel writes.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Good catch!

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Using interfaces that use a loader forces you to handle execution exceptions, since we rethrow those exceptions anyways and we manage our own "loading" I don't see a strong reason to use it.

PR #1807, opened 2019-04-01
Reduce logs - corfu runtime

xnull:
## Overview

Description:
Reduces logs generated by corfu runtime. It's not needed to have connected/disconnected information for channels since it is pretty low level. Lowered log level to debug/trace.

The sh command helping find most frequent messages in the log:
`cut -d"|" -f4 ~/tests.log | sort | uniq -c | sort -nr | head -n 15`

Why should this be merged: 

- [x] Public API has Javadoc


xnull:
in case of network issues, there are more high-level messages for that. If one goes to logs he/she can see multiple messages protocoling the same event in logs, on different levels.
For instance:
    connectAsync node1: Channel connected.
    fetch layout...

Almost always information about the connection is redundant in logs.

Can you please provide at least one example when this information could be useful?

xnull:
Also this is a requirement, right now logs are very chatty, number of records must be reduced. There is a bug for that in bugzilla

WenbinZhu:
> in case of network issues, there are more high-level messages for that. If one goes to logs he/she can see multiple messages protocoling the same event in logs, on different levels.
> For instance:
> connectAsync node1: Channel connected.
> fetch layout...
> 
> Almost always information about the connection is redundant in logs.
> 
> Can you please provide at least one example when this information could be useful?

Channel connectivity is different from upper level logs. You can have flakey network with connect/reconnect logs, but the upper layer fault detector may not even notice.

WenbinZhu:
@xnull Just use debug instead of trace.

xnull:
@WenbinZhu  fixed

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -112,7 +112,7 @@ private void waitForWorkflow(@NonNull UUID workflow, @NonNull ManagementClient c
                 return;
             }
             Sleep.sleepUninterruptibly(pollPeriod);
-            log.info("waitForWorkflow: waiting for {} on attempt {}", workflow, x);
+            log.debug("waitForWorkflow: waiting for {} on attempt {}", workflow, x);
```
zalokhan: This has been important in debugging customer issues. Please do not remove this.

xnull: CAn you please briefly explain how it helps? Maybe there could be some other logs that can help in debugging?
I see this message pretty often in logs, it doesn't seem like an important message.
Anyways let's discuss maybe offline

xnull:
@WenbinZhu, as discussed offline I changed log level from trace to debug, are you ok with the changes?

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
zalokhan: As discussed changing this to debug should be fine.

PR #1812, opened 2019-04-03
FastObjectLoader fix nextRead offset

xnull:
## Overview

Description:
If FastObjectLoader face TRIMMED message, it starts loading trimmed data again.

Why should this be merged: 
updates logHead up to `trimMark` then the loading process continues from trim mark position

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
@@ -7,34 +7,14 @@ import static org.corfudb.recovery.RecoveryUtils.getLogData;
 import static org.corfudb.recovery.RecoveryUtils.getSnapShotAddressOfCheckPoint;
 import static org.corfudb.recovery.RecoveryUtils.getStartAddressOfCheckPoint;
 import static org.corfudb.recovery.RecoveryUtils.isCheckPointEntry;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
-import java.util.function.BiConsumer;
-
-import javax.annotation.Nonnull;
-
 import lombok.Data;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.LogEntry;
 import org.corfudb.protocols.logprotocol.MultiObjectSMREntry;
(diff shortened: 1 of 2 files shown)


zalokhan:
Please fix the tests. These seem to be broken by this PR.

test/src/test/java/org/corfudb/recovery/FastObjectLoaderTest.java
```diff
@@ -594,26 +594,6 @@ public void failBecauseOfTrimAnd1Attempt() throws Exception {
 
     }
 
-    @Test
-    public void failWhenTrimHappensWhileFastLoading() throws Exception{
```
zalokhan: Don't remove this test. Please replace it with this test:
https://github.com/CorfuDB/CorfuDB/pull/1819/files#diff-c19ccef54c8e5dfbb5cd99aa987eeee8L595

xnull: done

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -511,11 +509,12 @@ private void cleanUpForRetry() {
         runtime.getAddressSpaceView().invalidateClientCache();
         runtime.getObjectsView().getObjectCache().clear();
         runtime.getStreamsView().getStreamCache().clear();
-        nextRead = logHead;
-        resetAddressProcessed();
 
         // Re ask for the Head, if it changes while we were trying
         findAndSetLogHead();
+
```
annym: Can you please add a comment, so in the future it is clear that this ensures that the loading process continues from the trim mark.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -736,34 +734,35 @@ private void applyForEachAddress(BiConsumer<Long, ILogData> logDataProcessor) {
...
+                    }
+                    addressProcessed++;
 
-                if (logData.getType() == DataType.TRIMMED) {
-                    log.warn("applyForEachAddress[{}, start={}] address is trimmed", address, logHead);
-                    handleRetry();
-                    canProcessRange = false;
-                    break;
-                }
+                    if (logData.getType() == DataType.TRIMMED) {
```
annym: Will this ever happen? Trimmed addresses are detected by the AddressSpace layer through getLogData.

test/src/test/java/org/corfudb/recovery/FastObjectLoaderTest.java
```diff
@@ -558,29 +558,25 @@ public void ignoreStreamCheckpoint() throws Exception{
...
         fsm.loadMaps();
 
         assertThatMapsAreBuiltIgnore(rt2, "Map1");
-
-
     }
 
     @Test(expected = RuntimeException.class)
-    public void failBecauseOfTrimAnd1Attempt() throws Exception {
+    public void failBecauseOfTrimAndZeroAttempts() throws Exception {
```
annym: If the lower test is exercising attempts = 0, should we leave this as the original? one attempt test.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
xnull: You are right, it shouldn't happen, that's why we throw IllegalStateException. Since it was in prev version, decided not to get rid of this part

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
xnull: done

test/src/test/java/org/corfudb/recovery/FastObjectLoaderTest.java
xnull: Agree, zero attempts sounds strange. 
Done

xnull:
@annym any new questions?

PR #1813, opened 2019-04-03
NoRollBack Debug Information

Maithem:
## Overview
Added extra logging to show the cause of a NoRollBackException.

Why should this be merged: Currently its not possible to know why NoRollBackException(NRE)
happen, this patch introduces extra metadata tracking to also log the cause of the
NRE. The metadata is tracked using a cache, so the cause of the NRE can be determined
as long as the cause exists in the cache.  

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/exceptions/NoRollbackException.java
+++ b/runtime/src/main/java/org/corfudb/runtime/exceptions/NoRollbackException.java
@@ -2,6 +2,7 @@ package org.corfudb.runtime.exceptions;
 
 import java.util.Optional;
 
+import lombok.Getter;
 import org.corfudb.protocols.logprotocol.SMREntry;
 
 /**
@@ -9,15 +10,20 @@ import org.corfudb.protocols.logprotocol.SMREntry;
  */
 public class NoRollbackException extends RuntimeException {
 
+    @Getter
+    private final long undoableEntryVersion;
+
     public NoRollbackException(long rollbackVersion) {
         super("Can't roll back due to non-undoable exception "
                 + " but need "
                 + rollbackVersion + " so can't undo");
+        undoableEntryVersion = rollbackVersion;
     }
 
     public NoRollbackException(long address, long rollbackVersion) {
         super("Could only roll back to " + address + " but need "
                 + rollbackVersion + " so can't undo");
+        undoableEntryVersion = address;
     }
 
     public NoRollbackException(Optional<SMREntry> entry, long address, long rollbackVersion) {
@@ -28,5 +34,6 @@ public class NoRollbackException extends RuntimeException {
                 + address
                 + " but need "
                 + rollbackVersion + " so can't undo");
+        undoableEntryVersion = address;
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -9,6 +9,7 @@ import org.corfudb.runtime.exceptions.NoRollbackException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
 import org.corfudb.runtime.object.transactions.WriteSetSMRStream;
 import org.corfudb.runtime.view.Address;
+import org.corfudb.runtime.view.AddressSpaceView;
 import org.corfudb.util.CorfuComponent;
 import org.corfudb.util.MetricsUtils;
 import org.corfudb.util.Utils;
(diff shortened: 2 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/exceptions/NoRollbackException.java
```diff
@@ -2,22 +2,28 @@
...
 import org.corfudb.protocols.logprotocol.SMREntry;
 
 /**
  * Created by mwei on 11/21/16.
  */
 public class NoRollbackException extends RuntimeException {
 
+    @Getter
+    final long undoableEntryVersion;
+
```
xnull: private

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -324,7 +325,8 @@ public void syncObjectUnsafe(long timestamp) {
                     try {
                         rollbackObjectUnsafe(timestamp);
                     } catch (NoRollbackException nre) {
-                        log.warn("SyncObjectUnsafe[{}] to {} failed {}", this, timestamp, nre);
+                        log.warn("SyncObjectUnsafe[{}] to {} failed {}, cause {}", this, timestamp, nre,
+                                AddressSpaceView.removalCauseCache.getIfPresent(nre.getUndoableEntryVersion()));
```
Maithem: Accessing a static variable inside the AddressSpaceView from this context seems a bit sloppy, but it is necessary (to provide important logging information) with the current code structure. Let me know if you have better ways of achieving this.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -53,6 +57,18 @@
...
 
+    public static long removalCauseCacheSize = 500_000;
+
+    /**
+     * This cache maintains debug information that is required to
+     * determine the cause of a NoRollbackException. Using ehcache's
+     * deepSizeOf it shows that the memory foot print of the
+     * removalCauseCache is around ~50mb.
+     */
+    public static Cache<Long, RemovalCause>  removalCauseCache = CacheBuilder.newBuilder()
```
xnull: `public static final Cache<Long, RemovalCause>  REMOVAL_CAUSE_CACHE ...`
I really don't like it. Can we use maybe metrics for that somehow? @StrikingThirteen what do you think?

Maithem: Great, propose a better approach.

apavan:
Maithem, please open a PR on 0.2.2 branch

Maithem:
@xnull @StrikingThirteen @hisundar @apavan comments?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -56,6 +55,18 @@
 @Slf4j
 public class AddressSpaceView extends AbstractView {
 
+    public static long removalCauseCacheSize = 500_000;
+
+    /**
+     * This cache maintains debug information that is required to
+     * determine the cause of a NoRollbackException. Using ehcache's
```
hisundar: typo: Using the* cache's deepSizeOf shows that ..

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -64,6 +75,8 @@
             .expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .expireAfterWrite(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
             .recordStats()
+            .removalListener((RemovalListener<Long, ILogData>) notification ->
```
annym: Just curious, would the historical give us some interesting info? for instance realize that we're replacing an entry over and over again or is this not aligned with what you're trying to collect.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: ehcache is the library the provides deepSizeOf :)

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: The point of this PR is to provide the **cause** of NoRollBackException, of course if it exists in the history.

Maithem: There are 5 causes for entries to be evicted, but we usually can't report why because we don't track cache eviction causes. If the history is big enough to capture when NoRollBackException, then the VLO can log exactly why the NoRollBackException happened.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
hisundar: sorry newbie mistake :)

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: Yes, this I get. I was just wondering if we wanted to count the number of times we were replacing an entry in the cache, but now that it is not a loadingCache that actually shouldn't even be a removal cause in our case.

Maithem:
Any more suggestions?

runtime/src/main/java/org/corfudb/runtime/exceptions/NoRollbackException.java
Maithem: done.

PR #1817, opened 2019-04-06
Freeze segment head for epoch before electing new sequencer.

zalokhan:
## Overview

Description:
Introduces the address space freeze before a new sequencer is elected. This is different from a seal logically as it is intended to seal the head of the chain as not to allow any in-flight messages from the previous epoch or tokens dispatched by the previous sequencer.
Again, this is not combined with the seal protocol to keep both the algorithms semantically separate for what they are intended.

Why should this be merged: Causes a correctness issue.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/Layout.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/Layout.java
@@ -366,6 +366,12 @@ public class Layout {
     public enum ReplicationMode {
         CHAIN_REPLICATION {
             @Override
+            public void validateAddressSpaceSeal(LayoutSegment layoutSegment,
+                                                 Map<String, CompletableFuture<Boolean>> completableFutureMap) {
+                SealServersHelper.freezeChainSegment(layoutSegment, completableFutureMap);
+            }
+
+            @Override
             public void validateSegmentSeal(LayoutSegment layoutSegment,
                                             Map<String, CompletableFuture<Boolean>>
                                                     completableFutureMap)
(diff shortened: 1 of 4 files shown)


zalokhan:
> Consider the following example with a three node cluster A, B and C.
> 
> * epoch = 3, sequencer sees streamA and the global tail at 10
> * ThreadA acquires a token to write for streamA at position 11
> * ThreadB acquires tokens 12 to 20 and for streamB and writes at position 20
> * ThreadA only writes to the head logging unit and dies
> * ThreadC tries to sync stream A and sees a hole at position 11 and starts the propagation, it reads the
>   write at position 11 (still doesnt propagate it to B and C)
> * Now node A is marked dead and removed from the cluster
> * The sequencer moves from node A to node B
> * The sequencer restarts with epoch 4 and sees streamA at 10 and streamB at 20
> * The checkpointer writes a marker to position 21 and starts checkpointing streamA at tail 10
> * ThreadC starts propagating the write to node B and C, so now at global time stamp 21 streamA should
>   be at tail 11, but the checkpointer saw streamA at tail 10 (at global timestamp 21)
> * After threadC finishes propagating the write other threads observe that commit
> * When the checkpointer finishes, it trims at position 20, with all the checkpoints at addresses greater
>   than 21
> * A non epoch aware trim happens on address 20
> * StreamA's commit at position 11 is lost
> 
> Notice that streamA was checkpointed at tail 10 and trimmed at address 20, but position 11 was never checkpointed, but it was trimmed.

The concern is when new head (node B) is moved to a new epoch, some hole fill thread may be able to propagate the data from Node A to B and C, without new sequencer noticing. Actually this propagation will be rejected by the batch writer with WrongEpochException because the hole fill thread still has old epoch.

PR #1819, opened 2019-04-07
Cache-After-Write Race + Data Loss Bug (#1805)

Maithem:

Porting #1805 from master to corfu-0.2.2

Furthermore, this patch also fixes a correctness bug introduced
in 2ab39da845ba3e40a66e31ad550d54d59e3db103 where empty log
slots are ignored and not hole filled.

- [x] Public API has Javadoc


apavan:
@zeeshan
I do not understand the connection b/w this change and the fast loader change. Please clarify.

zalokhan:
@apavan This patch changes the place where TrimExceptions are exposed above. This has an impact on the FastObjectLoader as it would crash when trim is encountered. Ideally, it should catch this trim exception and just retry.
This patch fixes this issue:
https://github.com/CorfuDB/CorfuDB/pull/1812/files#diff-67b6f1a0dfb81f6c4b4d827360ff100aR763

Zeeshan:
Sorry. Am not sure about this project/ commit. Can u share more details

PR #1823, opened 2019-04-09
Stream Address Maps with Feature Switch (Corfu-0.2.2)

annym:
## Overview

Provide a way to configure the use of backpointers or streams maps as
the stream address discovery mechanism. This param is present in the CorfuRuntimeParameters.
- followBackpointersEnabled (true: backpointers, false: bitmaps)


annym:
No longer needed. Closing.

PR #1825, opened 2019-04-09
Log who changed the epoch

hisundar:
## Overview

Description:

Why should this be merged: Debugging patch to help diagnose epoch changes. Logging who changed an epoch saves time by pinpointing which cluster node to look at next.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
@@ -100,7 +100,8 @@ public class BaseServer extends AbstractServer {
                                                    @NonNull IServerRouter r) {
         try {
             long epoch = msg.getPayload();
-            log.info("handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch {}", epoch);
+            log.info("handleMessageSetEpoch: Received SET_EPOCH from {}, moving to new epoch {}",
+                    msg.getClientID(), epoch);
             serverContext.setServerEpoch(epoch, r);
             r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
         } catch (WrongEpochException e) {


infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -100,7 +100,8 @@ public synchronized void handleMessageSetEpoch(@NonNull CorfuPayloadMsg<Long> ms
                                                    @NonNull IServerRouter r) {
         try {
             long epoch = msg.getPayload();
-            log.info("handleMessageSetEpoch: Received SET_EPOCH, moving to new epoch {}", epoch);
+            log.info("handleMessageSetEpoch: Received SET_EPOCH from {}, moving to new epoch {}",
+                    msg.getClientID(), epoch);
```
WenbinZhu: This clientID is a UUID that generated when we create CorfuRuntime. Here we log it on server side, please double check if we log it on client side when using CorfuRuntime, because otherwise we can't tell which node is setting the epoch.

xnull: As I know we use clientId in many places. Even if we don't use it on the client side we should add it to the client then

hisundar: Good point, Wenbin. While the server's clientId does get logged in some messages, it would be nice if instead of clientId we can print the IP address of the node initiating this SET_EPOCH command? Do you know of a way to extract that info?

xnull: Don't do that, there could be many clients on one machine(ip address)
I believe clientId is a good thing to use

xnull: You can add clientId on client side in some logs in this PR, that is up to you

WenbinZhu: @hisundar I think you can do both, clientID + client Ip. For clientID, we can log it in CorfuRuntime when we first create it and/or when we issue a `setRemoteEpoch`. For logging client Ip address in `handleMessageSetEpoch`, you can use something like `((InetSocketAddress) ctx.channel().remoteAddress()).getAddress().getXXX()`.

hisundar: ok I will leave it as it is, since I see client id being logged in multiple messages of corfu server logs.

PR #1828, opened 2019-04-10
Message epoch validation on prefixTrim.

zalokhan:
## Overview

Description: Avoid batch writer validating the token epoch on prefix trim. This is to prevent rejecting the trim request for an address with an old epoch. The address in its completeness is a valid address with the old epoch. As long as the address is generated by the CheckpointWriter after persisting the checkpoint, the trim on this address should be processed.

Why should this be merged: We invalidate a valid trim request on an older address.

- [x] Public API has Javadoc


zalokhan:
Not required in 0.2.2.

PR #1831, opened 2019-04-11
Failure detector: aggregated cluster state

xnull:
## Overview

Description:
The failure detector uses the aggregated state to make a decision to mark a node as failed or

Why should this be merged: 
Improves failure detection logic, makes it more stable

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
```diff
@@ -0,0 +1,118 @@
...
+import org.corfudb.runtime.exceptions.WrongEpochException;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+@Builder
+@Slf4j
+public class ClusterStateCollector {
```
zalokhan: Comments.

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
```diff
@@ -0,0 +1,118 @@
...
+                .build();
+    }
+
+    public ImmutableMap<String, Long> collectWrongEpochs() {
+        Map<String, Long> wrongEpochs = new HashMap<>();
+
+        clusterState.forEach((server, state) -> {
+            try {
+                //Add a node to connectedNodes list if node responded with its NodeState
+                state.get(timeout, TimeUnit.MILLISECONDS);
```
zalokhan: Please remove this duplicate code and aggregate all lists in a single place. It gets difficult to understand the code.

There can also be the following scenario:
Since we have an RPC timeout of 5 seconds.
In this code, every method will timeout differently:
First, you would collect node states and timeout in 2 seconds.
Then you collect wrong epoch exceptions where your future could have completed and returned a wrongEpochException.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -187,15 +185,41 @@ private PollReport pollRound(
...
-                .failedNodes(ImmutableSet.copyOf(failedNodesAggregated))
+                .activeLayoutServers(layout.getActiveLayoutServers())
                 .wrongEpochs(ImmutableMap.copyOf(wrongEpochsAggregated))
-                .currentLayoutSlotUnFilled(isCurrentLayoutSlotUnFilled(layout, wrongEpochsAggregated.keySet()))
-                .clusterState(latestReport.getClusterState())
+                .clusterState(getAggregatedState(reports))
+                .build();
+    }
+
+    private ClusterState getAggregatedState(List<PollReport> reports) {
```
zalokhan: Add comments. Also, show a small example if possible.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -210,7 +234,7 @@ private long modifyIterationTimeouts(Map<String, IClientRouter> router, PollRepo
         long pollIterationStartTime = System.nanoTime();
```
zalokhan: Move this pollIterationStartTime to before the start of iteration in the pollRound.
This variable doesn't make sense here.
The `iterationElapsedTime` has to be the time it took to complete 1 poll iteration.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -283,103 +279,43 @@ private NodeState getLocalNodeState(Set<String> allServers, Set<String> allConne
...
+            SequencerMetrics sequencerMetrics, Layout layout) {
 
         log.trace("Poll iteration. Epoch: {}", epoch);
 
-        Map<String, CompletableFuture<NodeState>> stateFuture = pollAsync(allServers, clientRouters, epoch);
+        ClusterStateCollector clusterCollector = ClusterStateCollector.builder()
+                .localEndpoint(localEndpoint)
+                .timeout(period)
+                .clusterState(pollAsync(allServers, clientRouters, epoch))
+                .heartbeatCounter(heartbeatCounter)
```
zalokhan: The `heartbeatCounter` should be incremented every round and not every iteration.
The `heartbeatCounter` is sort of a version number to your cluster state which you update.

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
```diff
@@ -21,43 +23,78 @@
...
+    private final ClusterState clusterState;
+
     /**
-     * A map of nodes answered with {@link WrongEpochException}.
+     * List of responsive servers in a current layout
      */
-    @Default
-    private final ImmutableMap<String, Long> wrongEpochs = ImmutableMap.of();
+    @NonNull
+    private final List<String> activeLayoutServers;
```
zalokhan: Rename it to activeServers or activeNodes.
This can be confused with the layout server component in Corfu.

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
```diff
@@ -21,43 +23,78 @@
...
+                wrongEpochs, activeLayoutServers
+        );
+
+        // Check if all active layout servers are present in the outOfPhaseEpochNodes map.
+        boolean result = wrongEpochs.keySet().containsAll(activeLayoutServers);
+
+        if (result) {
+            String msg = "Current layout slot is empty. Filling slot with current layout." +
+                    " wrong epochs: {}, active layout servers: {}";
+            log.info(msg, wrongEpochs, activeLayoutServers);
```
zalokhan: Can you put the log message directly here?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClusterState.java
```diff
@@ -53,20 +64,41 @@ public int size() {
...
+            if (nodeState.getConnectivity().getType() == NodeConnectivityType.NOT_READY) {
                 return false;
             }
         }
 
         return true;
     }
+
+    public NodeConnectivity getLocalNodeConnectivity() {
+        Optional<NodeState> maybeNodeState = getNode(localEndpoint);
```
zalokhan: Rename this to localNodeState

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterStateCollectorTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class ClusterStateCollectorTest {
+
+    @Test
+    public void test(){
```
zalokhan: Better test name?

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
xnull: Actually, it gets simpler. Let me explain:
1. Single responsibility principle - why that one is one of the most important ones? This principle makes code much easier to understand. In the previous version of this algorithm, we had one `for cycle` where we update **3** collections, which brings next problems:
    - it increases cognitive complexity (it gets much harder to understand)
    - makes code error prone
    - prevents us from writing unit tests! This part is most important. The unit tests they are indicators, if you can write a unit test easily for your code, that means your code has good quality. You can see I covered ClusterStatusCollector by unit tests easily.

Now when all 3 parts of the `collecting` algorithm are independent, it's easier to understand, it's compostable, all parts are independent and covered by unit tests. 
There is a little bit more code, but it's worth it.

xnull: Regarding completable futures, you don't pay any extra cost if you navigated many times. 
Let me explain this part: when the completable future gets resolved you can get the result from the future immediately (you don't excecute CF many times, you execute it ones and then you reuse results as many times as you want)

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
xnull: I copy pasted this logic from the previous version. Not sure if we need to change it in this PR. Let's create another one especially for timeouts

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
xnull: I copy pasted this logic from the previous version. Not sure if we need to change it in this PR.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClusterState.java
xnull: the Optional value can be empty, that's why it's `maybe`, in the other case I would have this situation:
```
Optional<NodeState> nodeState = getNode(localEndpoint);
if(nodeState.isPresent()){
  NodeState ??? = nodeState.get();
} 
```

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
zalokhan: I don't feel this code is simplifying the current code. It makes it harder to read.
Another thing,
Going back to my previous observation, this is wrong because of the scenario I mentioned:
you wait on the same completable future twice for 2 seconds each. The completable future may not complete in the first 2 seconds but might complete in the 3rd second and return you a valid value.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
zalokhan: This is definitely a bug. Since you are anyways cleaning this code, you can clean this here.
We do not want to revisit this code again for some time.
Anyways, this is your call.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
zalokhan: Ok. But this needs cleanup too.
You can move the incrementAndGet heartbeatcounter here rather than in the aggregator.
That will be easier to understand.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
xnull: let's leave it for an another pr

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
xnull: let's do it in another pr

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
xnull: renamed to `responsiveServers`

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
xnull: Regarding timeouts you are right, it could happen in theory. 
Also, since we set up a client router to resolve the future after some timeout, we don't need to have the same timeout in the collector.  ClusterCollector shouldn't be in charge of timeouts. 
I have deleted all timeouts from the collector.

xnull: Regarding simplicity - if you can prove somehow that prev logic was better, I would agree. For now, on my side, I have objective arguments: best practices, and unit tests.

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
xnull: done

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterStateCollectorTest.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
xnull: done

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClusterState.java
```diff
@@ -53,20 +64,41 @@ public int size() {
...
+
+        if (!maybeNodeState.isPresent()) {
+            throw new IllegalArgumentException("Node not found: " + localEndpoint);
+        }
+
+        return maybeNodeState.get().getConnectivity();
```
vjeko: `maybeNodeState.map(this::getConnectivity).orElseThrow(...)`?

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
```diff
@@ -0,0 +1,134 @@
...
+     * @return map of endpoint:epoch entries
+     */
+    public ImmutableMap<String, Long> collectWrongEpochs() {
+        Map<String, Long> wrongEpochs = new HashMap<>();
+
+        clusterState.forEach((server, state) -> {
+            try {
+                //Add a node to connectedNodes list if node responded with its NodeState
+                state.get();
+            } catch (Exception e) {
```
vjeko: What happens if you get Thowable? Unless you are catching it somewhere it will cause your periodic task to stop.

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
```diff
@@ -21,43 +23,76 @@
...
+    }
+
     /**
-     * Current cluster state, collected by a failure detector
+     * Contains list of servers successfully connected with current node.
+     * It doesn't contain nodes answered with WrongEpochException.
      */
-    @NonNull
-    private final ClusterState clusterState;
+    public Set<String> getReachableNodes() {
```
vjeko: How about we rename this to getSameEpochNodes or getSealedNodes?

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
```diff
@@ -21,43 +23,76 @@
...
-     * @return
+     * Contains list of servers disconnected from this node. If the node A can't ping node B then node B will be added
+     * to failedNodes list.
      */
-    public ImmutableSet<String> getAllConnectedNodes() {
-        return Sets.union(connectedNodes, wrongEpochs.keySet()).immutableCopy();
+    public Set<String> getFailedNodes() {
+        Set<String> failedNodes = clusterState.getLocalNodeConnectivity().getFailedNodes();
+
+        return Sets.difference(failedNodes, wrongEpochs.keySet());
```
vjeko: Isn't the set difference always empty? When you get wrongEpoch does this mean the ping has not succeeded?

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -187,15 +184,61 @@ private PollReport pollRound(
...
+     * 4. reports = [CONNECTED, CONNECTED, CONNECTED], reports[2] will be used
+     * 5. reports = [UNAVAILABLE, UNAVAILABLE, NOT_READY], reports[2] will be used
+     *
+     * @param reports list of poll reports collected by failure detector
+     * @return aggregated cluster state
+     */
+    private ClusterState getAggregatedState(List<PollReport> reports) {
+        Map<String, NodeState> stateMap = new HashMap<>();
+
+        reports.forEach(report -> {
```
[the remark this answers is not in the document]
xnull: Thanks for the idea. Implemented in a bit different way using  java stream `reduce()`

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
```diff
@@ -0,0 +1,134 @@
...
+                if (e.getCause() instanceof WrongEpochException) {
+                    wrongEpochs.put(server, ((WrongEpochException) e.getCause()).getCorrectEpoch());
+                }
+            }
+        });
+
+        return  ImmutableMap.copyOf(wrongEpochs);
+    }
+
+    private Map<String, NodeState> collectNodeStates() {
```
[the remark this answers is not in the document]
xnull: done

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClusterState.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
xnull: We shouldn't catch Throwable right? If there is an error, you can't do much, you can only exit from your application

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
xnull: getSameEpochNodes - too `low` level name, too specific/implementation related
getSealedNodes - too specific also

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
xnull: the ping has succeeded https://github.com/CorfuDB/CorfuDB/pull/1831/files#diff-45a8e0bb947ac9cb433401fb4d2d24e9R113

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Collects information about cluster state.
+ * Provides list of servers in the cluster with higher epochs.
+ * Builds the {@link ClusterState} and local {@link NodeState}.
+ *
+ * The collectors navigates via list of requests represented as CompletableFuture-s
```
hisundar: nit: collector*

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
```diff
@@ -0,0 +1,134 @@
...
+        nodeStates.putAll(collectRemoteStates());
+
+        return ClusterState.builder()
+                .localEndpoint(localEndpoint)
+                .nodes(ImmutableMap.copyOf(nodeStates))
+                .build();
+    }
+
+    /**
+     * Provides list of servers contains different epochs to this server.
```
hisundar: nit: containing*

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
xnull: done

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterStateCollectorTest.java
```diff
@@ -0,0 +1,103 @@
...
+        NodeState nodeBState = clusterState.getNode("b").get();
+        NodeConnectivity nodeBConnectivity = nodeBState.getConnectivity();
+        assertThat(nodeBConnectivity.getConnectedNodes()).containsExactly("a", "b");
+
+        NodeState nodeCState = clusterState.getNode("c").get();
+        assertThat(nodeCState.isConnected()).isFalse();
+    }
+
+    @Test
+    public void testWrongEpochs(){
```
StrikingThirteen: Can you some comments on what the test is doing

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterStateAggregatorTest.java
```diff
@@ -0,0 +1,91 @@
...
+        assertThat(nodeBState.isConnected()).isTrue();
+        assertThat(nodeBState).isEqualTo(expectedNodeBState);
+
+        //check [UNAVAILABLE, UNAVAILABLE, NOT_READY]
+        NodeState expectedNodeCState = clusterState3.getNode("c").get();
+        NodeState nodeCState = aggregator.getAggregatedState().getNode("c").get();
+        assertThat(nodeCState.isConnected()).isFalse();
+        assertThat(nodeCState).isEqualTo(expectedNodeCState);
+    }
+}
```
StrikingThirteen: Add new line

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateAggregator.java
```diff
@@ -0,0 +1,75 @@
...
+            throw new IllegalStateException("Insufficient cluster state");
+        }
+
+        if (clusterStates.size() == 1) {
+            return clusterStates.get(0);
+        }
+
+        Map<String, NodeState> stateMap = clusterStates.stream()
+                .map(ClusterState::getNodes)
+                .reduce((prevNodes, currNodes) -> {
```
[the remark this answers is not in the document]
xnull: It doesn't affect performance in our case. 
To be able to hit the performance we need to have millions of nodes in the cluster, as I know we don't have any plans yet regarding that level of scalability.

At the same time `reduce()` makes the logic of aggregation as simple as possible (at least I don't know better ways)

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterStateAggregatorTest.java
xnull: done

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterStateCollectorTest.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/management/PollReport.java
```diff
@@ -21,43 +23,76 @@
...
+    private final ClusterState clusterState;
+
     /**
-     * A map of nodes answered with {@link WrongEpochException}.
+     * List of responsive servers in a current layout
      */
-    @Default
-    private final ImmutableMap<String, Long> wrongEpochs = ImmutableMap.of();
+    @NonNull
+    private final List<String> responsiveServers;
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterStateCollectorTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class ClusterStateCollectorTest {
+
+    @Test
+    public void testAggregatedState(){
```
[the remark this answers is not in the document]
xnull: done

xnull:
Fixed all questions, please review

infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
vjeko: We have been getting random InterruptedExceptions, so keep that in mind.

infrastructure/src/test/java/org/corfudb/infrastructure/management/PollReportTest.java
```diff
@@ -0,0 +1,107 @@
...
+                .wrongEpochs(ImmutableMap.of("a", epoch))
+                .clusterState(clusterState)
+                .build();
+
+        assertThat(pollReport.getReachableNodes()).isEmpty();
+        assertThat(pollReport.getAllReachableNodes()).containsExactly("a");
+    }
+
+    @Test
+    public void testConnectionStatuses() {
```
vjeko: Please put a description.

infrastructure/src/test/java/org/corfudb/infrastructure/management/PollReportTest.java
```diff
@@ -0,0 +1,107 @@
...
+                .clusterState(clusterState)
+                .build();
+
+        assertThat(pollReport.getReachableNodes()).containsExactly("a");
+        assertThat(pollReport.getAllReachableNodes()).containsExactly("a", "b");
+        assertThat(pollReport.getFailedNodes()).containsExactly("c");
+    }
+
+    @Test
+    public void testIsSlotUnfilled() {
```
vjeko: Description missing.

PR #1832, opened 2019-04-11
Failure detector aggregated state

xnull:
## Overview

Description:
The failure detector uses the aggregated state to make a decision to mark a node as failed

Why should this be merged:
Improves failure detection logic, makes it more stable

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
@@ -15,6 +15,7 @@ import org.corfudb.protocols.wireprotocol.NodeState.HeartbeatTimestamp;
 import org.corfudb.protocols.wireprotocol.SequencerMetrics;
 import org.corfudb.protocols.wireprotocol.failuredetector.NodeConnectivity;
 import org.corfudb.protocols.wireprotocol.failuredetector.NodeConnectivity.ConnectionStatus;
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -192,15 +193,49 @@ private PollReport pollRound(
...
+                String endpoint = nodeState.getConnectivity().getEndpoint();
+                NodeConnectivityType connType = nodeState.getConnectivity().getType();
+
+                if (!stateMap.containsKey(endpoint)) {
+                    stateMap.put(endpoint, nodeState);
+                    continue;
+                }
+
+                //Don't replace by a new state if the new state is not CONNECTED
+                if (stateMap.get(endpoint).getConnectivity().getType() == NodeConnectivityType.CONNECTED
```
zalokhan: This can just be: ```if (stateMap.get(endpoint).getConnectivity().getType() == NodeConnectivityType.CONNECTED)``` right?

xnull: no
you can replace CONNECTED node state only by another connected node state

zalokhan: Ok, that was my point. Why replace when you can even leave the first one.
But I see you would want to keep the latest node state.

xnull:
> LGTM. Are we adding tests?

Will add in #1831

PR #1834, opened 2019-04-11
Log sender's IP during epoch change, log clientID when epoch change is initiated

hisundar:
## Overview

Description:

Why should this be merged: More debugging info to diagnose epoch changes.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
@@ -1,5 +1,6 @@
 package org.corfudb.infrastructure;
 
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import lombok.Getter;
 import lombok.NonNull;
@@ -13,6 +14,7 @@ import org.corfudb.runtime.exceptions.WrongEpochException;
 
 import javax.annotation.Nonnull;
 import java.lang.invoke.MethodHandles;
+import java.net.InetSocketAddress;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
@@ -100,8 +102,14 @@ public class BaseServer extends AbstractServer {
                                                    @NonNull IServerRouter r) {
         try {
             long epoch = msg.getPayload();
-            log.info("handleMessageSetEpoch: Received SET_EPOCH from {}, moving to new epoch {}",
-                    msg.getClientID(), epoch);
+            String remoteHostAddress;
+            try {
+                remoteHostAddress = ((InetSocketAddress)ctx.channel().remoteAddress()).getAddress().getHostAddress();
+            } catch (NullPointerException e) {
+                remoteHostAddress = "unavailable";
+            }
+            log.info("handleMessageSetEpoch: Received SET_EPOCH from (clientId={}:{}), moving to new epoch {}",
+                    msg.getClientID(), remoteHostAddress, epoch);
             serverContext.setServerEpoch(epoch, r);
             r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
         } catch (WrongEpochException e) {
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -13,9 +14,12 @@
 
 import javax.annotation.Nonnull;
 import java.lang.invoke.MethodHandles;
+import java.net.InetSocketAddress;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
+import static com.sun.xml.internal.ws.policy.sourcemodel.wspolicy.XmlToken.Optional;
```
xnull: this is not that Optional that you want.
your optional is - java.util.Optional

or you just forget to delete this import

hisundar: sorry not used, will remove

hisundar: removed, added by error. thanks

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -100,8 +102,12 @@ public synchronized void handleMessageSetEpoch(@NonNull CorfuPayloadMsg<Long> ms
                                                    @NonNull IServerRouter r) {
         try {
             long epoch = msg.getPayload();
-            log.info("handleMessageSetEpoch: Received SET_EPOCH from {}, moving to new epoch {}",
-                    msg.getClientID(), epoch);
+            String remoteHostAddress = "0.0.0.0";
```
StrikingThirteen: why setting to 0.0.0.0 rather than null? remoteAddress() might return null anyways. If you want to differentiate between ctx.channel being null and ctx.channel being not connected then this is fine but not very expressive.

xnull: I would prefer seeing zeros instead of emptiness)

hisundar: > If you want to differentiate between ctx.channel being null and ctx.channel being not connected then this is fine but not very expressive
Yes, it was just some way of distinguishing the channel being null.

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -100,8 +102,12 @@ public synchronized void handleMessageSetEpoch(@NonNull CorfuPayloadMsg<Long> ms
...
+            if (ctx != null && ctx.channel() != null) {
+                remoteHostAddress = ((InetSocketAddress)ctx.channel().remoteAddress()).getAddress().getHostAddress();
```
WenbinZhu: Even if `ctx.channel()` is not null, `remoteAddress()` may also return null if the channel is not connected. You might have a NullPointerException here. 

Also address 0.0.0.0 has a meaning, I think it's better not to use it.

hisundar: Thanks, let me change it to a different string and handle the exception instead of checking for it.

hisundar: Done. could you please re-review? thanks

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
hisundar: I have removed 0.0.0.0 thanks

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
WenbinZhu: NIT: add a space after `(InetSocketAddress)`

Issue #1835, opened 2019-04-12
Serializer Issue in Client


annym:
## Overview
 
In the case a client is attempting to access a map for which it does not contain the serializer type specified in the data itself, we do not properly handle this exception and fail with a NullPointerException, which is not informative.

PR #1836, opened 2019-04-12
Handle Serializer not present

annym:
## Overview

Description:

If a client does not have the serializer required to access data from a given map, it would currently throw a NullPointerException and fail to open a map with no clear cause.

Why should this be merged: this change set properly handles this error and informs the client on the missing serializer.

Related issue(s) (if applicable): #1835

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
+++ b/runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
@@ -4,6 +4,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.exceptions.SerializerException;
 
 
 /**
@@ -38,10 +39,16 @@ public class Serializers {
      */
     public static ISerializer getSerializer(Byte type) {
         if (type <= SYSTEM_SERIALIZERS_COUNT) {
-            return serializersMap.get(type);
-        } else {
+            if (serializersMap.containsKey(type)) {
+                return serializersMap.get(type);
+            }
+        } else if (customSerializers.containsKey(type)) {
             return customSerializers.get(type);
         }
+
+        log.error("Serializer with type code {} not found. Please check custom serializers " +
+                "for this client.", type.intValue());
+        throw new SerializerException(type);
     }
 
     /**
--- a/runtime/src/main/java/org/corfudb/runtime/exceptions/SerializerException.java
+++ b/runtime/src/main/java/org/corfudb/runtime/exceptions/SerializerException.java
@@ -1,0 +1,16 @@
+package org.corfudb.runtime.exceptions;
+
+/**
+ * Created by annym on 04/12/19
+ */
+public class SerializerException extends RuntimeException {
+
+    /**
+     * Constructor for SerializerException specifying serializer type.
+     *
+     * @param type serializer type not found for client.
+     */
+    public SerializerException(Byte type) {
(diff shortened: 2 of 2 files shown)


runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
```diff
@@ -38,10 +38,19 @@
...
-            return customSerializers.get(type);
+            if (customSerializers.containsKey(type)) {
+                return customSerializers.get(type);
+            }
         }
+
+        String message = String.format("Serializer with type code %s not found. " +
+                "Please check custom serializers for this client.", type.intValue());
+        log.error(message);
+        throw new RuntimeException(message);
```
WenbinZhu: Maybe a more meaningful exception?

zalokhan: You can create a new Exception such as SerializerException.

runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
```diff
@@ -38,10 +38,19 @@
...
         } else {
-            return customSerializers.get(type);
+            if (customSerializers.containsKey(type)) {
+                return customSerializers.get(type);
+            }
         }
+
+        String message = String.format("Serializer with type code %s not found. " +
+                "Please check custom serializers for this client.", type.intValue());
+        log.error(message);
```
zalokhan: Can you put this message to log directly?

annym: Done.

runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
annym: I've created a SerializerException like suggested by Zeeshan. Please review.

WenbinZhu: Thanks, I’ve approved

runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
```diff
@@ -38,10 +39,18 @@
      */
     public static ISerializer getSerializer(Byte type) {
         if (type <= SYSTEM_SERIALIZERS_COUNT) {
-            return serializersMap.get(type);
+            if (serializersMap.containsKey(type)) {
+                return serializersMap.get(type);
+            }
         } else {
-            return customSerializers.get(type);
+            if (customSerializers.containsKey(type)) {
```
StrikingThirteen: Please combine and remove the nested if.

annym:
> I think this is more of a logging issue, the proxy was logging the toString (i.e. NullPointerException) and not the whole stack trace.
> https://github.com/CorfuDB/CorfuDB/pull/1813/files#diff-2bed8c5f4bb097f75e48cfd16015dfccR187

@Maithem The NullPointerException was coming from:

https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java#L115

Maithem:
@annym yes, but I was pointing out that this would have not been an issue (i.e. it would have been clear), if the proxy wasn't omitting the stacktrace via the toString()

annym:
@Maithem yes, totally agree. We should also push that in so we don't miss on the stack trace. But I also think we could throw a more specific exception in the absence of a serializer instead of failing with a NullPointerException.

Issue #1822, opened 2019-04-09
Tx End and Shutdown Exception


annym:
## Overview

AbortCause UNDEFINED has a non-deterministic behavior, the data could have been committed or not. 

TXEnd() in ObjectsView.java

catch (Exception e) {
            log.error("TXEnd[{}]: Unexpected exception", context, e);
            TxResolutionInfo txInfo = new TxResolutionInfo(context.getTransactionID(),
                    Token.UNINITIALIZED);
            TransactionAbortedException tae = new TransactionAbortedException(
                    txInfo, AbortCause.UNDEFINED, e, context);
            context.abortTransaction(tae);
            throw new UnrecoverableCorfuError("Unexpected exception during commit", e);
        }

Let's suppose the following scenario:

- A thread initiates an Optimistic TX 
- (performs different operations)
- It attempts to END the TX and actually commits to 2 server nodes successfully and fails on the 3rd node (as this node has been removed). 
- This exception causes to validate if the data was written or not (validateOfTheWrittenEntry) which actually attempts to read from the tail of the chain (node which has been removed from the configuration)... Furthermore a shutdown of that client router  is triggered when the node is detected to be out, which causes this thread to complete exceptionally (ShutdownException). 

This leads in the injection framework to correctness issues, because this thread actually dies, as it is only expecting TxAbortExceptions, and this UnrecoverableCorfuError bubbles up.

WenbinZhu:
I think currently we're not doing a sync after commit, if you mean `tryCommitAllProxies`, that part actually a no-op because no one is registering proxies.

annym:
Yes @WenbinZhu like we discussed offline it is not a sync it is the validateStateOfWrittenEntry for the last write in the chain. I've adjusted the text to properly describe the issue.

WenbinZhu:
@annym Cool, thanks for verifying it. Also this is probably related to a potential bug I just found today, we can discuss that tomorrow.

WenbinZhu:
Fixed in #1843

WenbinZhu:
Fixed in #1843

PR #1837, opened 2019-04-15
Make the LogUnit Async

Maithem:
## Overview
Now all LogUnit requests (except for reading and Reset) are processed
asynchrously. As a consequence, the number of working threads is
reduced to numCPU * 2.

Why should this be merged: Reduces the number of required threads and cleans up the implementation
of both the LogUnitServer and BatchWriter.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,69 +139,120 @@ public LogUnitServer(ServerContext serverContext) {
...
-        } catch (DataOutrankedException e) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_OUTRANKED.msg());
-        } catch (ValueAdoptedException e) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_VALUE_ADOPTED.payloadMsg(e
-                    .getReadResponse()));
-        }
+        LogData logData = (LogData) msg.getPayload().getData();
+
+        batchWriter.addTask(WRITE, msg)
+                .thenAccept(Void -> {
```
Maithem: thenAccept might block the main thread.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -227,72 +298,39 @@ private void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerC
...
      * - All operations after this operation but stamped with an older epoch will be failed.
      */
     @Override
     public void sealServerWithEpoch(long epoch) {
-        batchWriter.waitForSealComplete(epoch);
+        CorfuPayloadMsg<Long> msg = new CorfuPayloadMsg<>();
+        msg.setEpoch(epoch);
+        //TODO(Maithem): need to unwrap CompletionException for WrongEpochExceptions etc...
+        try {
+            batchWriter.addTask(SEAL, msg).join();
```
zalokhan: We need to wait on the future.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -215,7 +110,13 @@ private void batchWriteProcessor() {
                         log.trace("Sync'd {} writes", processed);
 
                         for (BatchWriterOperation operation : res) {
-                            handleOperationResults(operation);
+                            if (!operation.getFutureResult().isCompletedExceptionally()
+                            && !operation.getFutureResult().isCancelled()) {
+                                //TODO(Maithem): Fail the whole batch on any failure.
```
xnull: don't leave TODO's in your code

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriterOperation.java
```diff
@@ -1,37 +1,32 @@
...
     private final Type type;
-    private final Long address;
-    private final LogData logData;
-    private final Long epoch;
-    private final List<LogData> entries;
-    private final CompletableFuture future;
-    private Exception exception;
-
+    private final CorfuPayloadMsg msg;
+    private Object retVal;
```
xnull: Why object? Can we add a generic parameter which specifies retval type?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,69 +141,130 @@ public LogUnitServer(ServerContext serverContext) {
...
-        TailsResponse tails = batchWriter.queryTails(msg.getEpoch());
-        r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tails));
+        log.debug("handleTailRequest: received a tail request {}", msg);
+        CorfuPayloadMsg<Void> tailsReqMsg = new CorfuPayloadMsg<>();
+        tailsReqMsg.copyBaseFields(msg);
+        batchWriter.<TailsResponse>addTask(TAILS_QUERY, tailsReqMsg)
+                .thenAccept(tailsResp -> r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp)))
+                .exceptionally(ex -> {
+                    handleException(ex, ctx, tailsReqMsg, r);
+                    return null;
```
xnull: I don't like that we have to return null. Can you please replace `exceptionally()` by `handle(...)` or `whenCompleted(...)`

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,69 +141,130 @@ public LogUnitServer(ServerContext serverContext) {
...
     }
 
     /**
      * Service an incoming request to retrieve the starting address of this logging unit.
      */
     @ServerHandler(type = CorfuMsgType.TRIM_MARK_REQUEST)
-    public void handleHeadRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+    public void handleTrimMarkRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+        log.debug("handleTrimMarkRequest: received a trim mark request {}", msg);
+        // TODO(Maithem): Should go through the batchwriter?
```
xnull: Get rid of todos please

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,69 +141,130 @@ public LogUnitServer(ServerContext serverContext) {
...
         r.sendResponse(ctx, msg, CorfuMsgType.TRIM_MARK_RESPONSE.payloadMsg(streamLog.getTrimMark()));
     }
 
+    // TODO(Maithem): converting exceptions to message types is really superfluous,
```
xnull: Get rid of totos please

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: you can use `thenRun(..)` instead of `thenAccept()` it fits better

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,69 +141,130 @@ public LogUnitServer(ServerContext serverContext) {
...
-        }
+        LogData logData = (LogData) msg.getPayload().getData();
+
+        batchWriter.addTask(WRITE, msg)
+                .thenAccept(Void -> {
+                    dataCache.put(msg.getPayload().getGlobalAddress(), logData);
+                    r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg());
+                }).exceptionally(ex -> {
+                    handleException(ex, ctx, msg, r);
+                    return null;
```
xnull: I don't like that we have to return null. Can you please replace `exceptionally()` by `handle(...)` or `whenCompleted(...)`

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,69 +141,130 @@ public LogUnitServer(ServerContext serverContext) {
...
+                           ChannelHandlerContext ctx, IServerRouter r) {
+        List<LogData> range = msg.getPayload().getEntries();
+        log.debug("rangeWrite: Writing {} entries [{}-{}]", range.size(),
+                range.get(0).getGlobalAddress(), range.get(range.size() - 1).getGlobalAddress());
+
+        batchWriter.addTask(RANGE_WRITE, msg)
+                .thenAccept(Void -> r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg()))
+                .exceptionally(ex -> {
+                    handleException(ex, ctx, msg, r);
+                    return null;
```
xnull: I don't like that we have to return null. Can you please replace exceptionally() by handle(...) or whenCompleted(...)

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,69 +141,130 @@ public LogUnitServer(ServerContext serverContext) {
...
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_VALUE_ADOPTED.payloadMsg(e
-                    .getReadResponse()));
-        }
+                           ChannelHandlerContext ctx, IServerRouter r) {
+        List<LogData> range = msg.getPayload().getEntries();
+        log.debug("rangeWrite: Writing {} entries [{}-{}]", range.size(),
+                range.get(0).getGlobalAddress(), range.get(range.size() - 1).getGlobalAddress());
+
+        batchWriter.addTask(RANGE_WRITE, msg)
+                .thenAccept(Void -> r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg()))
```
xnull: please use `thenRun`

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -227,72 +298,39 @@ private void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerC
...
-            streamLog.compact();
-            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
-        } catch (RuntimeException ex) {
-            log.error("Internal Error");
-            //TODO(Maithem) Need an internal error return type
-            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
-        }
+    private void handleCompactRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+        log.debug("handleCompactRequest: received a compact request {}", msg);
+        streamLog.compact();
```
xnull: why if compact() throws an exception?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -227,72 +298,39 @@ private void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerC
...
-     *   in the queue before this operation.
+     * in the queue before this operation.
      * - All operations after this operation but stamped with an older epoch will be failed.
      */
     @Override
     public void sealServerWithEpoch(long epoch) {
-        batchWriter.waitForSealComplete(epoch);
+        CorfuPayloadMsg<Long> msg = new CorfuPayloadMsg<>();
+        msg.setEpoch(epoch);
+        //TODO(Maithem): need to unwrap CompletionException for WrongEpochExceptions etc...
```
xnull: get rid of todo

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -304,27 +342,29 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
                                            ChannelHandlerContext ctx, IServerRouter r) {
+
         // Check if the reset request is with an epoch greater than the last reset epoch seen to
         // prevent multiple reset in the same epoch. and should be equal to the current router
         // epoch to prevent stale reset requests from wiping out the data.
         if (msg.getPayload() > serverContext.getLogUnitEpochWaterMark()
                 && msg.getPayload() == serverContext.getServerEpoch()) {
             serverContext.setLogUnitEpochWaterMark(msg.getPayload());
-            batchWriter.reset(msg.getPayload());
+            batchWriter.addTask(RESET, msg).join();
```
xnull: Please use the advantages of the non-blocking approach:
```
batchWriter.addTask(RESET, msg).thenRun(() -> {
  dataCache.invalidateAll();
  r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
});
```

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -304,27 +342,29 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
             dataCache.invalidateAll();
             log.info("LogUnit Server Reset.");
         } else {
             log.info("LogUnit Server Reset request received but reset already done.");
         }
+        // TODO(Maithem): send a nack if a resent can't get executed
```
xnull: please finish todo

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -215,7 +110,13 @@ private void batchWriteProcessor() {
...
+                                // At this point we need to complete the requests
+                                // that completed successfully (i.e. haven't failed)
+                                operation.getFutureResult().complete(operation.getRetVal());
```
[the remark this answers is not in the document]
Maithem: No, because for writes we can only ack the request after a successful sync.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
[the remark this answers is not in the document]
Maithem: Also, unlike the stream tails, the underlying implementation is thread-safe, so it doesn't need to go through the batch processor.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: `CompletableFuture::join`

Is a blocking call.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriterOperation.java
Maithem: We can, but I see it unnecessary at the moment because its just a placeholder. In other words, we don't have code that casts it.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
```diff
@@ -228,40 +129,40 @@ private void batchWriteProcessor() {
                     log.trace("Shutting down the write processor");
                     streamLog.sync(true);
                     break;
-                } else if (currOp.getType() == Type.SEAL && currOp.getEpoch() >= sealEpoch) {
-                    sealEpoch = currOp.getEpoch();
+                } else if (currOp.getType() == Type.SEAL && currOp.getMsg().getEpoch() >= sealEpoch) {
+                    log.info("batchWriteProcessor: updateding from {} to {}", sealEpoch, currOp.getMsg().getEpoch());
```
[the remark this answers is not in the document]
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriter.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: Nice. Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: Done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: Its not expected to.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
+
+/**
+ * This class manages access for operations that need ordering while executing against
+ * the backing storage.
+ */
+@Slf4j
+public class BatchProcessor implements AutoCloseable {
+
+    static final int BATCH_SIZE = 50;
```
xnull: private

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+
+    final boolean sync;
```
xnull: private

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,217 @@
...
+
+    private StreamLog streamLog;
+
+    private BlockingQueue<BatchWriterOperation> operationsQueue;
```
xnull: final?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: I agree, I don't like it too, but I believe that's a shortcoming of the `CompletableFuture` API, not the usage. I see the application of `exceptionally` appropriate here because its used as the last stage. Furthermore, I don't see how `handle` and `whenCompleted` are a cleaner alternatives, because using them will also require "null checking".

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriterOperation.java
xnull: Even if we don't have casts I still vote for a prev version. `Object` confuses when reading code

Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+     */
+    private long sealEpoch;
+
+    /**
+     * Returns a new BatchProcessor for a stream log.
+     *
+     * @param streamLog      stream log for writes (can be in memory or file)
+     * @param sealEpoch All operations stamped with epoch less than the epochWaterMark are
+     *                       discarded.
+     * @param streamLog stream log for writes (can be in memory or file)
```
annym: streamLog is repeated twice.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+    public <T> CompletableFuture <T> addTask(@Nonnull Type type, @Nonnull CorfuPayloadMsg msg) {
+        BatchWriterOperation<T> operation = new BatchWriterOperation<>(type, msg);
+        operationsQueue.add(operation);
+        return operation.getFutureResult();
+    }
+
+    private void processor() {
+
+        if (!sync) {
+            log.warn("batchWriteProcessor: writes configured to not sync with secondary storage");
```
annym: Should we add some info on the implications, so it is more informative to users. Could incur in data loss?

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+
+                if (lastOp == null) {
+                    currOp = operationsQueue.take();
+                } else {
+                    currOp = operationsQueue.poll();
+
+                    if (currOp == null || processed == BATCH_SIZE
+                            || currOp == BatchWriterOperation.SHUTDOWN) {
+                        streamLog.sync(sync);
+                        log.trace("Sync'd {} writes", processed);
```
annym: If sync is set to false we are still printing that it has been synced, right?

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+                        }
+                        res.clear();
+                        processed = 0;
+                    }
+                }
+
+                if (currOp == null) {
+                    lastOp = null;
+                } else if (currOp == BatchWriterOperation.SHUTDOWN) {
+                    log.trace("Shutting down the write processor");
```
annym: Should this be warning level? so it can hint on why operations are not being processed in case of any issues.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+                                break;
+                            case RESET:
+                                streamLog.reset();
+                                break;
+                            case TAILS_QUERY:
+                                TailsResponse tails = streamLog.getTails();
+                                currOp.setRetVal(tails);
+                                break;
+                            default:
+                                log.warn("Unknown BatchWriterOperation {}", currOp);
```
annym: currOp.getType?

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: done

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: I think it would be redundant. Its kind of implied with "configured to not sync with secondary storage"

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: Changed it to 
`log.trace("Completed {} operations", processed);`

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: Added a ToString annotation.

Maithem:
Addressed all comments .

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+                } else {
+                    currOp = operationsQueue.poll();
+
+                    if (currOp == null || processed == BATCH_SIZE
+                            || currOp == BatchWriterOperation.SHUTDOWN) {
+                        streamLog.sync(sync);
+                        log.trace("Completed {} operations", processed);
+
+                        for (BatchWriterOperation operation : res) {
+                            if (!operation.getFutureResult().isCompletedExceptionally()
```
WenbinZhu: I think we don't need to check `isCompletedExceptionally` or `isCancelled()` here. If the future is already completed exceptionally  or cancelled, completing it with any value will not overwrite the exception (i.e. the exception will still be caught in `thenRun()`, `.exceptionally()` or `get()`, etc)

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,60 +142,126 @@ public LogUnitServer(ServerContext serverContext) {
      */
     @ServerHandler(type = CorfuMsgType.TAIL_REQUEST)
     public void handleTailRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
-        TailsResponse tails = batchWriter.queryTails(msg.getEpoch());
-        r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tails));
+        log.debug("handleTailRequest: received a tail request {}", msg);
+        CorfuPayloadMsg<Void> tailsReqMsg = new CorfuPayloadMsg<>();
+        tailsReqMsg.copyBaseFields(msg);
+        batchWriter.<TailsResponse>addTask(TAILS_QUERY, tailsReqMsg)
```
WenbinZhu: Question: why tails query needs to go to batch writer?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -131,60 +142,126 @@ public LogUnitServer(ServerContext serverContext) {
...
-        } catch (DataOutrankedException e) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_OUTRANKED.msg());
-        } catch (ValueAdoptedException e) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_VALUE_ADOPTED.payloadMsg(e
-                    .getReadResponse()));
-        }
+        LogData logData = (LogData) msg.getPayload().getData();
+
+        batchWriter.addTask(WRITE, msg)
+                .thenRun(() -> {
```
WenbinZhu: I think using `thenRun()` here has a performance issue. `thenRun()` will use the same thread as `addTask()`, which means if you print the thread name here, you will see it is using the `BatchProcessor` thread to do the following operations (dataCache.put and sendResponse), which is not intended. So I believe the right way to do this is to use `thenRunAsync()` and pass this log unit server's `executor` as the second parameter for `thenRunAsync()`. What do you think?

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+import org.corfudb.protocols.wireprotocol.WriteRequest;
+import org.corfudb.runtime.exceptions.WrongEpochException;
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
+
+/**
+ * This class manages access for operations that need ordering while executing against
+ * the backing storage.
+ */
+@Slf4j
+public class BatchProcessor implements AutoCloseable {
```
WenbinZhu: I have a question on the design of BatchProcessor, it is processing requests one by one, but commits (i.e. completes future) and sends RPC responses after a batch, which means there will be a cpu/network IO spike after each batch. Is this intended? Since now the log unit is fully asynchronous, maybe we don't need to batch it?

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: At this point, the uncompleted tasks need to be completed with their corresponding results. These are the tasks that have succeeded. L175 exceptionally completes the failed tasks. In the future, we will need to fail the whole batch if any write task failed within the batch.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: The `StreamLog::getTails` is not thread safe, so it is "made" thread safe by executing it sequentially with respect to other writes.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: The idea is to be non-blocking. 

Based on my understanding, the LogUnit thread can complete the task (i.e. executes `CompletableFuture::thenRun`) if the future completed "right away", otherwise the LogUnit thread will not block to wait for completion. In the case where the future is not completed "right away", then the completing thread will execute thenRun (i.e. the BatchProcessor thread). Since the post-task work is **minimal** and also **non-blocking** then I think its fine for the BatchProcessor to do it.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
Maithem: The initial design of the BatchWriter (now called BatchProcessor) is to execute a batch of writes, then fsync the batch. Notice that we can only acknowledge the writes after the fsync completes. That's why the futures are completed only after a successful fsync. I think naturally it is a "spiky" workload: multiple producers and a single consumer. 

Right now, I don't believe this is an issue, but in general it is. We still need to address topic of throttling, but that will come when performance improves.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: More info. 

Ran a test where some tasks execute right away. You can see two of the tasks were accepted by the LogUnit threads (i.e. completed right away) and the two others dont. 

o.c.i.LogUnitServer | write: handler thread LogUnit-1
**o.c.i.LogUnitServer | write: Accepting thread LogUnit-1**
 o.c.i.LogUnitServer | write: handler thread LogUnit-4
**o.c.i.LogUnitServer | write: Accepting thread LogUnit-4**
o.c.i.LogUnitServer | write: handler thread LogUnit-7
o.c.i.LogUnitServer | write: Accepting thread LogUnit-BatchProcessor-0
o.c.i.LogUnitServer | write: handler thread LogUnit-2
o.c.i.LogUnitServer | write: Accepting thread LogUnit-BatchProcessor-0
o.c.i.LogUnitServer | write: handler thread LogUnit-5


In another test, I force latency in the writes and you see that all requests are completed by the BatchProcessor. 

o.c.i.LogUnitServer | write: handler thread LogUnit-1
o.c.i.LogUnitServer | write: handler thread LogUnit-2
o.c.i.LogUnitServer | write: handler thread LogUnit-3
o.c.i.LogUnitServer | write: handler thread LogUnit-4
o.c.i.LogUnitServer | write: Accepting thread LogUnit-BatchProcessor-0
o.c.i.LogUnitServer | write: Accepting thread LogUnit-BatchProcessor-0
o.c.i.LogUnitServer | write: Accepting thread LogUnit-BatchProcessor-0
o.c.i.LogUnitServer | write: Accepting thread LogUnit-BatchProcessor-0

Maithem: This might be an issue under high cache contention, or on many cache misses where the BatchProcessor thread competes with the LogUnit server threads.

WenbinZhu: Yes, the  post-task work includes cache and Netty operations. Cache operation may have lock contention and Netty operations may have serialization costs. My point here is batchProcessor thread should be focused on handling log operations, and let the server's executor do all the extra works, which makes more sense since the server executor is meant to do such work, otherwise the single threaded batchProcessor will be overloaded.

Maithem: The netty operations are non-blocking and serialization happens via the netty worker thread, not the LogUnit handler threads, or the BatchProcessor.

At the moment its not an obvious choice to me. Note that the cache operations are "non-blocking" (i.e. since I removed the Writer). But there might be an issue when many of the reads result in cache misses and the loading mechanism starts locking many of the buckets, which can block the BatchProcessor thread. Well if that happens and the LogUnit threads are busy competing on "reads", then whose adding work to the BatchProcessor? 

In the case of no-contention,  why force the overhead of dispatching the work to another thread pool, if its not necessary. That trade-off is not obvious to me. Actually, that's one of the reasons why CompletableFuture tries to run on the same thread if the operation completes right away (i.e. it can be an optimization).

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
lixinchengdu: > I think we don't need to check `isCompletedExceptionally` or `isCancelled()` here. If the future is already completed exceptionally or cancelled, completing it with any value will not overwrite the exception (i.e. the exception will still be caught in `thenRun()`, `.exceptionally()` or `get()`, etc)

I think it is a kind of defensive programming and makes the code more readable.

Maithem: Alternatively, we can just call `CompletableFuture::isDone()` , but I think in this context, its more readable to check for a failure.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchWriterOperation.java
```diff
@@ -1,37 +1,34 @@
...
     private final Type type;
-    private final Long address;
-    private final LogData logData;
-    private final Long epoch;
-    private final List<LogData> entries;
-    private final CompletableFuture future;
-    private Exception exception;
-
+    private final CorfuPayloadMsg msg;
+    private T retVal;
```
lixinchengdu: retVal is better replaced with resultValue

Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -375,6 +420,25 @@ long getMaxCacheSize() {
...
 
+    @VisibleForTesting
+    BatchProcessor getBatchWriter() {
+        return batchWriter;
+    }
+
+    // The following methods should only be used for unit tests, ideally the executor should be
+    // final, but this "hack" is needed to now when a task as completed
+    @VisibleForTesting
+    void startHandler() {
```
Maithem: Java's Executor/ExecutorService don't have an API to track pending tasks without invoking shutdown or being the caller, so when a requests spans multiple executors to complete there's no way of figuring out when all the executors are "done", so by shutting down the executor and installing a new one, we implicitly wait on the tasks to complete.

Is there a better way to do this in java? 

When a request spans multiple executors

WenbinZhu: ExecutorCompletionService has an api take() which returns a completed task. I'm not sure if this is useful in your case.

annym:
Confirmed that the correctness issue seen in this PR is another occurrence of #1822 (@WenbinZhu any updates on this?)... I will retrigger correctness tests for this PR.

annym:
Analyzed second injection framework failure also due to known bug #1822 (we need to address this soon)

14:47:55.991 ^[[1;31mERROR^[[0;39m [pool-6-thread-9] ^[[36mo.c.r.v.ObjectsView^[[0;39m - TXEnd[TX[479f]]: Unexpected exception
 java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.corfudb.runtime.exceptions.NetworkException: Router stopped [endpoint=tcp://192.168.0.3:9000/]
        at org.corfudb.util.CFUtils.getUninterruptibly(CFUtils.java:65)
        at org.corfudb.util.CFUtils.getUninterruptibly(CFUtils.java:99)
        at org.corfudb.runtime.view.replication.ChainReplicationProtocol.peek(ChainReplicationProtocol.java:71)
Caused by: java.util.concurrent.ExecutionException: org.corfudb.runtime.exceptions.NetworkException: Router stopped [endpoint=tcp://192.168.0.3:9000/]
        at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
        at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)
        at org.corfudb.runtime.clients.NettyClientRouter.sendMessageAndGetCompletable(NettyClientRouter.java:414)
Caused by: org.corfudb.runtime.exceptions.NetworkException: Router stopped [endpoint=tcp://192.168.0.3:9000/]
        at org.corfudb.runtime.clients.NettyClientRouter.stop(NettyClientRouter.java:381)
        at org.corfudb.runtime.CorfuRuntime.lambda$pruneRemovedRouters$3(CorfuRuntime.java:870)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)

zalokhan:
@annym The ShutdownException was changed to NetworkException.
This was done primarily to fix the transaction handling on shutdown exception. This should have fixed the injection framework. We will need to revisit this and see why the transactions are not completed gracefully.

WenbinZhu:
@annym @zalokhan I think it's probably because NetworkException is wrapped in ExecutionException and RuntimeException, while TxEnd() is trying to catch a raw NetworkException.

annym:
> @annym @zalokhan I think it's probably because NetworkException is wrapped in ExecutionException and RuntimeException, while TxEnd() is trying to catch a raw NetworkException.

@zalokhan @WenbinZhu yes, It's related to this.. it's actually the same issue but with the NetworkException instead. On the path to validate the written entry, the router is disconnected and the generated NetworkException is wrapped in a RuntimeException which bubbles up and does not cause a TAE. Therefore this thread terminates with no notification of success or failure to the correctness engine (and actually It is successful because the third node is removed, is this correct?).. should we continue to validate the state of the written entry in the new tail and then take a decision?

zalokhan:
@annym I think the validation is required.
Both instances of NetworkExceptions - via a true NetworkException / via router shutdown Network Exception are treated differently since the latter is wrapped.
I am addressing this in #1861

PR #1839, opened 2019-04-17
Delete junk doc files

xnull:
## Overview

Description:
Delete junk/empty/not used documentation files.

Why should this be merged: 

- [x] Public API has Javadoc


Maithem:
@hisundar I think we should keep one of the two (not both rst and md). If we still want to keep these sections then I would prefer markdown.

xnull:
I don't want to create empty files. Let's just use `docs` dir for docs, anyone can write his documentation and leave it in this directory. I am going to write `failure detector` doc in the next pr.

@hisundar I agree with Maithem, let's use only markdown

PR #1843, opened 2019-04-19
Add new reboot utilty to prevent router race, fix netty channel leak.

WenbinZhu:
## Overview

- Add a new reboot utility for use including reset and restart which can
prevent a race when someone holds a router in the router pool but it's
pruned by another thread by invoking invalidateLayout.

- Fix the channel leak where channels are not closed when shutting
down CorfuRuntime.

- Remove ShutdownException, use NetworkException instead.

Why should this be merged:  Fix router race

Related issue(s) (if applicable): #1751, #1822

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -864,8 +863,10 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
                         .contains(NodeLocator.getLegacyEndpoint(endpoint)))
                 .forEach(endpoint -> {
                     try {
-                        nodeRouterPool.getNodeRouters().get(endpoint).stop();
-                        nodeRouterPool.getNodeRouters().remove(endpoint);
+                        IClientRouter router = nodeRouterPool.getNodeRouters().remove(endpoint);
```
xnull: cool!

runtime/src/main/java/org/corfudb/runtime/RebootUtil.java
```diff
@@ -0,0 +1,131 @@
...
+import java.time.Duration;
+
+/**
+ * Utility to Reboot a server which includes reset or restart
+ * Reset wipes out all existing Corfu data while restart does not.
+ *
+ * <p>Created by WenbinZhu on 4/18/19.
+ */
+@Slf4j
+public class RebootUtil {
```
xnull: If only unit tests use RebootUtil, why it seats in runtime?

WenbinZhu: This is for user to do reset/restart

runtime/src/main/java/org/corfudb/runtime/RebootUtil.java
```diff
@@ -0,0 +1,131 @@
...
+                    CFUtils.getUninterruptibly(baseClient.reset());
+                } else {
+                    log.info("Attempting to restart server: {}", endpoint);
+                    CFUtils.getUninterruptibly(baseClient.restart());
+                }
+                break;
+            } catch (Exception e) {
+                log.error("Reboot node: {} failed with exception:", endpoint, e);
+                if (retries == 0) {
+                    throw new RuntimeException(e);
```
xnull: please replace RuntimeException by a more specific exception

WenbinZhu: Done. Changed BootstrapUtil as well.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -332,6 +331,39 @@ public void restartTest() throws Exception {
...
+        runtime = createDefaultRuntime();
+        Layout layout = incrementClusterEpoch(runtime);
+        RebootUtil.restart(SERVER_0, runtime.getParameters(), retries, PARAMETERS.TIMEOUT_LONG);
+
+        waitForEpochChange(epoch -> epoch >= layout.getEpoch() + 1, runtime);
+        assertThat(shutdownCorfuServer(corfuServer)).isTrue();
+        runtime.shutdown();
+
+        // Start testing reset utility
+        corfuServer = runPersistentServer(corfuSingleNodeHost, PORT_0, true);
```
zalokhan: Just a test optimization: You can reuse the same server from restart and check if the epoch returns to 0.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -864,8 +863,10 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
...
+                        if (router != null) {
+                            router.stop();
```
zalokhan: Just documenting our discussion here.
If the stop fails, there could be a possible leak of the router.
Reference to the router is lost and as experimented, the router does not get garbage collected and the channel remains open.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
WenbinZhu: Done. I was not able to reuse it because after restart I assert `assertThat(shutdownCorfuServer(corfuServer)).isTrue();`. But I think it's not necessary, so I removed it and now it can be reused.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -381,7 +378,10 @@ private void channelConnectionFutureHandler(@Nonnull ChannelFuture future,
     public void stop() {
         log.debug("stop: Shutting down router for {}", node);
         shutdown = true;
-        connectionFuture.completeExceptionally(new ShutdownException());
+        connectionFuture.completeExceptionally(new NetworkException("Router stopped", node));
+        if (channel != null && channel.isOpen()) {
+            channel.close();
```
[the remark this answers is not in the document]
WenbinZhu: I think we don't need to wait on the close, just let it close asynchronously. Also all the other places in this class that close the channel are doing in the same way.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
WenbinZhu: Done.

PR #1851, opened 2019-04-24
Reduce corfu server logs

xnull:
## Overview

Description:
Reduce corfu server logs

Why should this be merged:
Reduces the number of unnecessary log messages. 
Prevents polluting server logs with exceptions added to the log by NettyClientRouter, but the stack trace can't be used to find a potential issue and the exception will be thrown anyway by the other part.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -148,7 +148,7 @@ public class LogUnitServer extends AbstractServer {
      */
     @ServerHandler(type = CorfuMsgType.WRITE)
     public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.debug("log write: global: {}, streams: {}", msg.getPayload().getToken(),
+        log.trace("log write: global: {}, streams: {}", msg.getPayload().getToken(),
                 msg.getPayload().getData().getBackpointerMap());
 
         try {
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
@@ -107,7 +107,7 @@ public class ServerHandshakeHandler extends ChannelDuplexHandler {
         // Validate handshake, but first verify if node identifier is set to default (all 0's)
         // which indicates node id matching is not required.
         if (serverId.equals(UUID.fromString("00000000-0000-0000-0000-000000000000"))) {
-            log.info("channelRead: node id matching is not requested by client.");
+            log.trace("channelRead: node id matching is not requested by client.");
         } else if (!serverId.equals(this.nodeId)) {
             log.error("channelRead: Invalid handshake: this is {} and client is trying to connect to {}",
                     this.nodeId, serverId);
@@ -118,7 +118,7 @@ public class ServerHandshakeHandler extends ChannelDuplexHandler {
         // Store clientID as a channel attribute.
         ctx.channel().attr(clientIdAttrKey).set(clientId);
         log.info("channelRead: Handshake validated by Server.");
-        log.debug("channelRead: Sending handshake response: Node Id: {} Corfu Version: {}",
+        log.trace("channelRead: Sending handshake response: Node Id: {} Corfu Version: {}",
                 this.nodeId, this.corfuVersion);
 
         CorfuMsg handshakeResponse = CorfuMsgType.HANDSHAKE_RESPONSE
@@ -126,13 +126,13 @@ public class ServerHandshakeHandler extends ChannelDuplexHandler {
         ctx.writeAndFlush(handshakeResponse);
 
         // Flush messages in queue
-        log.debug("channelRead: There are [{}] messages in queue to be flushed.", this.messages.size());
+        log.trace("channelRead: There are [{}] messages in queue to be flushed.", this.messages.size());
         while (!messages.isEmpty()) {
             ctx.writeAndFlush(messages.poll());
         }
 
         // Remove this handler from the pipeline; handshake is completed.
-        log.info("channelRead: Removing handshake handler from pipeline.");
+        log.trace("channelRead: Removing handshake handler from pipeline.");
         ctx.pipeline().remove(this);
         this.fireHandshakeSucceeded();
     }
@@ -159,7 +159,7 @@ public class ServerHandshakeHandler extends ChannelDuplexHandler {
      */
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws Exception {
-        log.debug("channelInactive: Channel closed.");
+        log.trace("channelInactive: Channel closed.");
         if (!this.state.completed()) {
             this.fireHandshakeFailed(ctx);
         }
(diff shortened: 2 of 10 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -148,7 +148,7 @@ public void handleHeadRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRo
      */
     @ServerHandler(type = CorfuMsgType.WRITE)
     public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.debug("log write: global: {}, streams: {}", msg.getPayload().getToken(),
+        log.trace("log write: global: {}, streams: {}", msg.getPayload().getToken(),
```
zalokhan: We need this enabled for debugging.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -107,7 +107,7 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
         // Validate handshake, but first verify if node identifier is set to default (all 0's)
         // which indicates node id matching is not required.
         if (serverId.equals(UUID.fromString("00000000-0000-0000-0000-000000000000"))) {
-            log.info("channelRead: node id matching is not requested by client.");
+            log.trace("channelRead: node id matching is not requested by client.");
```
zalokhan: Do we see this log message often?
If not, there is no point in reducing the level.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -118,21 +118,21 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
...
         // Flush messages in queue
-        log.debug("channelRead: There are [{}] messages in queue to be flushed.", this.messages.size());
+        log.trace("channelRead: There are [{}] messages in queue to be flushed.", this.messages.size());
         while (!messages.isEmpty()) {
             ctx.writeAndFlush(messages.poll());
         }
 
         // Remove this handler from the pipeline; handshake is completed.
-        log.info("channelRead: Removing handshake handler from pipeline.");
+        log.trace("channelRead: Removing handshake handler from pipeline.");
```
zalokhan: You can add 1 info/debug log saying handshake succeeded. This is anyways a rare event - when a node/client reconnects.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -156,9 +156,9 @@ public void channelActive(ChannelHandlerContext ctx)
         CorfuMsg handshake = CorfuMsgType.HANDSHAKE_INITIATE
             .payloadMsg(new HandshakeMsg(this.clientId, this.nodeId));
 
-        log.debug("channelActive: Initiate handshake. Send handshake message.");
+        log.trace("channelActive: Initiate handshake. Send handshake message.");
         ctx.writeAndFlush(handshake);
-        log.debug("channelActive: Add {} to channel pipeline.", READ_TIMEOUT_HANDLER);
+        log.trace("channelActive: Add {} to channel pipeline.", READ_TIMEOUT_HANDLER);
```
zalokhan: You can have 1 logline indicating the handshake was successful.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -819,8 +820,12 @@ public synchronized void invalidateLayout() {
             // Don't create a new request for a layout if there is one pending.
             return;
         }
-        layout = fetchLayout(latestLayout == null
-                ? bootstrapLayoutServers : latestLayout.getLayoutServers());
+
```
zalokhan: Create a separate PR for code cleanup.
The subject of this PR says "reduce server logs".

runtime/src/main/java/org/corfudb/runtime/exceptions/NetworkException.java
```diff
@@ -9,16 +9,28 @@
...
 
-    public NetworkException(String message, NodeLocator node, Throwable cause) {
+    public NetworkException(Type exceptionType, String message, NodeLocator node, Throwable cause) {
         super(message + " [endpoint=" + node.toString() + "]", cause);
+        this.exceptionType = exceptionType;
         this.node = node;
     }
+
+    public enum Type {
+        DISCONNECTED
```
zalokhan: What are the other types you think we can have?
Doesn't NetworkException imply disconnected?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -885,82 +890,94 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
...
+            Sleep.sleepUninterruptibly(parameters.connectionRetryRate);
+            if (isShutdown) {
+                return null;
+            }
+        }
+    }
 
-                        // Prune away removed node routers from the nodeRouterPool.
-                        pruneRemovedRouters(l);
+    private Optional<Layout> fetLayoutServer(String server) {
```
zalokhan: fetLayoutServer -> fetchLayout
It's not fetching a layout server.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
xnull: Creating a PR, then running it several times then asking many times to review changes is a pretty heavy loaded and time-consuming operation. I don't see any reasons for extracting this small piece of code into another pr. Also, since I changed the related method (fetchLayout) I believe it's ok to bring this small change in this pr.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: This record actually is pretty useless, it writes a message in the log indicating that the operation has happened but in reality, the entry goes to the batch writer and can stay there (in a queue) for a relatively long time. 
I would prefer using StreamLogFiles to write the message to the log after the fact when actual writing to disk has happened.
Let me double check if I did it in this pr or not

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
xnull: All log levels I changed I see pretty often in the log. The number of those messages growing pretty fast in any connection issue happens, there are many duplicating messages, the log level some of them can be lowered. There are 3 classes which generate most of  the messages in a log: ServerHadshakeHandler, NettyClientRouter, NettyServerRouter

xnull: Also since we don't use `serverId` I don't see any reasons to print this message

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: > This record actually is pretty useless
False, I found this log very useful for debugging many issues.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: Put a note in the git commit.

Maithem:
@xnull if you still want this PR, then please address comments and rebase.

Maithem:
superseded by  #2043 #2047

Issue #1841, opened 2019-04-18
Don't cache state transfer read/write 


WenbinZhu:
## Overview

State transfer nowadays is no difference as normal read/write, which means  caches all read/write data in log unit server. This could pollute the cache. We can add an option to read/write messages used in state transfer to prevent these entries to be cached.

WenbinZhu:
Addressed in #1901

Issue #1795, opened 2019-03-22
ReadRange in ChainReplication can read from incorrect log units.


zalokhan:
## Overview

https://github.com/CorfuDB/CorfuDB/blob/2ab39da845ba3e40a66e31ad550d54d59e3db103/runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java#L109-L118

From the above code we can see that the chain replication protocol gets the logunit client based on only the start address.
This is incorrect as the start address can lie in a different stripe or even a different segment.
Since different log units would be responsible for serving reads in different chains, this read-range would lead to an incorrect read such as missing data.

zalokhan:
Test proving this:

```

    @Test
    public void xyz() {

        addServer(SERVERS.PORT_0);
        addServer(SERVERS.PORT_1);
        addServer(SERVERS.PORT_2);

        TestRule blockAll = new TestRule().drop().always();

        Layout l = new TestLayoutBuilder()
                .addLayoutServer(SERVERS.PORT_0)
                .addLayoutServer(SERVERS.PORT_1)
                .addLayoutServer(SERVERS.PORT_2)
                .addSequencer(SERVERS.PORT_0)
                .addSequencer(SERVERS.PORT_1)
                .addSequencer(SERVERS.PORT_2)
                .buildSegment()
                .setStart(0L)
                .setEnd(1L)
                .setReplicationMode(Layout.ReplicationMode.CHAIN_REPLICATION)
                .buildStripe()
                .addLogUnit(SERVERS.PORT_0)
                .addToSegment()
                .addToLayout()
                .buildSegment()
                .setStart(1L)
                .setEnd(-1L)
                .setReplicationMode(Layout.ReplicationMode.CHAIN_REPLICATION)
                .buildStripe()
                .addLogUnit(SERVERS.PORT_0)
                .addLogUnit(SERVERS.PORT_1)
                .addToSegment()
                .addToLayout()
                .build();
        bootstrapAllServers(l);
        addClientRule(getManagementServer(SERVERS.PORT_0).getManagementAgent().getCorfuRuntime(), blockAll);
        addClientRule(getManagementServer(SERVERS.PORT_1).getManagementAgent().getCorfuRuntime(), blockAll);
        addClientRule(getManagementServer(SERVERS.PORT_2).getManagementAgent().getCorfuRuntime(), blockAll);

[...]

PR #1855, opened 2019-04-26
Revamp read/write API for better performance, stability & readibility.

WenbinZhu:
## Overview

Description:

- Revamp read/write APIs in ChainReplicationProtocol, AddressSpaceView
and LogUnitClient, including code clean up, stabilization (better error
handling) and improve performance of some APIs by reducing number of
calls and leveraging cache.
- Fix a previous bug which readAll/readRange ignores ranges that span segments,
which causes loss of data read.

This would be a precedence PR for a following patch to remove state transfer 
caching on log unit server.

Why should this be merged: 

Related issue(s) (if applicable): #1841, #1795 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -417,68 +425,72 @@ public void invalidateServerCaches() {
...
+     * @param addresses closed range of addresses to read from.
+     * @return a map of read addresses.
      */
-    public @Nonnull
-    Map<Long, ILogData> fetchRange(Set<Long> addresses) {
-        Map<Long, ILogData> result = layoutHelper(e -> e.getLayout().getReplicationMode(addresses.iterator().next())
+    @Nonnull
+    public Map<Long, ILogData> fetchRange(Range<Long> addresses, boolean waitForWrite) {
+        Map<Long, ILogData> result = layoutHelper(e -> e.getLayout()
+                .getReplicationMode(addresses.lowerEndpoint())
```
Maithem: Btw this can be incorrect if the read spans two segments that don't have the same replication protocol.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WriteRequest.java
```diff
@@ -1,52 +1,35 @@
...
 import lombok.Getter;
 
+import java.util.EnumMap;
+
 /**
  * Created by mwei on 8/9/16.
  */
 @Builder
-@AllArgsConstructor
 public class WriteRequest implements ICorfuPayload<WriteRequest>, IMetadata {
```
Maithem: Can you re-add `@AllArgsConstructor`

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -739,9 +739,9 @@ private void applyForEachAddress(BiConsumer<Long, ILogData> logDataProcessor) {
         while (nextRead <= logTail) {
             try {
                 final long start = nextRead;
-                final long stopNotIncluded = Math.min(start + batchReadSize, logTail + 1);
-                nextRead = stopNotIncluded;
-                final Map<Long, ILogData> range = getLogData(runtime, start, stopNotIncluded);
+                final long stopIncluded = Math.min(start + batchReadSize - 1, logTail);
```
Maithem: Maybe its better to use interval terminology here, perhaps LeftEndpoint and RightEndpoint?

runtime/src/main/java/org/corfudb/recovery/RecoveryUtils.java
```diff
@@ -82,41 +81,33 @@ static ILogData getLogData(CorfuRuntime runtime, boolean loadInCache, long addre
...
-     * (e.g [0, 5) == (0,1,2,3,4))
+     * <p>It uses a closed range: [start, end]
+     * (e.g [0, 5] == [0,1,2,3,4])
      *
      * @param start start address for the bulk read
-     * @param end end address for the bulk read
+     * @param end   end address for the bulk read
      * @return logData map ordered by addresses (increasing)
      */
     static Map<Long, ILogData> getLogData(CorfuRuntime runtime, long start, long end) {
```
Maithem: I don't think this is necessary. Can you inline it?

The `AddressSpaceView` should expose such API directly.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -2,6 +2,7 @@
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.google.common.annotations.VisibleForTesting;
```
Maithem: Not needed.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -95,34 +93,23 @@ public Integer getPort() {
...
-     * @param type    The data type
-     * @param streams The streams, if any, that this write belongs to.
-     * @param rank    The rank of this write]
+     * @param address the address to write to.
+     * @param type    the data type
+     * @param rank    the rank of this write
+     * @return a completable future which returns true on success.
      */
-    public CompletableFuture<Boolean> writeEmptyData(long address, DataType type, Set<UUID> streams,
+    public CompletableFuture<Boolean> writeEmptyData(long address, DataType type,
```
Maithem: You can remove this.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -268,6 +264,17 @@ ILogData read(long address) {
...
+
+    /**
+     * Read the given object from a range of addresses.
+     *
+     * @param addresses An iterable with addresses to read from
+     * @param waitForWrite Flag whether wait for write is required or hole fill directly.
+     * @return A map of read addresses, which will be cached if caching is enabled
+     */
+    public Map<Long, ILogData> read(Iterable<Long> addresses, boolean waitForWrite) {
         if (!runtime.getParameters().isCacheDisabled()) {
```
Maithem: `waitForWrite` is a hack, don't expose it as a part of the `AddressSpaceView` API. We should also clean that up,

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -417,68 +425,72 @@ public void invalidateServerCaches() {
...
-     * @param addresses collection of addresses to read from.
-     * @return A result to be cached
+     * @param addresses closed range of addresses to read from.
+     * @return a map of read addresses.
      */
-    public @Nonnull
-    Map<Long, ILogData> fetchRange(Set<Long> addresses) {
-        Map<Long, ILogData> result = layoutHelper(e -> e.getLayout().getReplicationMode(addresses.iterator().next())
+    @Nonnull
+    public Map<Long, ILogData> fetchRange(Range<Long> addresses, boolean waitForWrite) {
```
Maithem: I don't think we should expose a Guava collection as a part of the API. Maybe pass the endpoints in the method signature and internally construct a `Range?`

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -73,96 +73,140 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
-        Range<Long> range = Range.encloseAll(globalAddresses);
-        return readRange(runtimeLayout, range, true);
-    }
-
+    @Nonnull
+    public Map<Long, ILogData> readAll(RuntimeLayout runtimeLayout,
+                                       List<Long> addresses,
+                                       boolean waitForWrite) {
+        long startAddress = Collections.min(addresses);
+        long endAddress = Collections.max(addresses);
```
Maithem: This can lose data if the addresses span a segment.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -73,96 +73,140 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
+        int numUnits = runtimeLayout.getLayout().getSegmentLength(startAddress);
+        log.trace("readAll[{}-{}]: chain {}/{}", startAddress, endAddress, numUnits, numUnits);
 
+        Map<Long, LogData> logResult = CFUtils.getUninterruptibly(runtimeLayout
```
Maithem: Maybe `readResult` is a better variable name

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -73,96 +73,140 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
 
-        Map<Long, LogData> logResult = CFUtils.getUninterruptibly(
-                runtimeLayout
-                        .getLogUnitClient(startAddress, numUnits - 1)
-                        .read(range)).getAddresses();
+        Map<Long, LogData> logResult = CFUtils.getUninterruptibly(runtimeLayout
+                .getLogUnitClient(startAddress, numUnits - 1)
+                .readRange(addresses)).getAddresses();
+
+        return readOrHoleFill(runtimeLayout, logResult, startAddress, numUnits, waitForWrite);
```
Maithem: Factor out the fill hole logic, to take disjoint sets and ranges. Let's get rid of readOrHoleFill. It should be read Vs peak

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -73,96 +73,140 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
-        Map<Long, ILogData> returnResult = new TreeMap<>();
+        // the first entry in the list. All subsequent holes in the list can be
+        // hole filled without waiting as we have already waited for the first hole.
+        boolean wait = true;
         for (Map.Entry<Long, LogData> entry : logResult.entrySet()) {
             long address = entry.getKey();
             ILogData value = entry.getValue();
-            if (value == null || value.isEmpty()) {
-                if (!wait) {
+            if (!isLogDataPresent(value)) {
```
Maithem: Just use `ILogData::isEmpty`, lower layers shouldn't use null to signal empty positions.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -73,96 +73,140 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
             }
-
-            returnResult.put(entry.getKey(), value);
         }
+
         return returnResult;
     }
 
+
+    private boolean isLogDataPresent(ILogData logData) {
```
Maithem: I dont think we need this. Just remove nulls from the lower layers.

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
```diff
@@ -63,49 +69,62 @@
...
+     * @return a map of addresses to data commit at these address, hole filling if necessary.
      */
-    Map<Long, ILogData> readRange(RuntimeLayout runtimeLayout, Set<Long> globalAddresses);
-    /** Peek data from a given address.
+    default @Nonnull
+    Map<Long, ILogData> readRange(RuntimeLayout runtimeLayout,
+                                Range<Long> addresses,
+                                boolean waitForWrite) {
+        return ContiguousSet.create(addresses, DiscreteDomain.longs())
+                .parallelStream()
```
Maithem: Remove this default implementation. This usage of parallelStream can be problematic.

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
```diff
@@ -114,16 +133,16 @@
...
+     * @param runtimeLayout   the RuntimeLayout stamped with layout to use for the peekAll.
+     * @param globalAddresses a set of addresses to read from.
+     * @return a map of addresses to uncommitted addresses, without hole filling.
      */
-    default @Nonnull Map<Long, ILogData> peekAll(RuntimeLayout runtimeLayout,
-                                                 Set<Long> globalAddresses) {
+    default @Nonnull
+    Map<Long, ILogData> peekAll(RuntimeLayout runtimeLayout,
+                                Set<Long> globalAddresses) {
         return globalAddresses.parallelStream()
```
Maithem: Remove the default implementation.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -2,11 +2,7 @@
 
 import com.google.common.annotations.VisibleForTesting;
 
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.NavigableSet;
-import java.util.UUID;
+import java.util.*;
```
Maithem: Expand.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
WenbinZhu: Yes this is an existing issue, and there is a comment say we don't handle this case. Will copy that comment here as well.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: This is also an existing issue. Let's discuss if we need to fix it in this patch

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
zalokhan: Let's fix this issue if we are refactoring this code right now.
Issue: #1795

WenbinZhu: #1795 Addressed.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WriteRequest.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/recovery/RecoveryUtils.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
WenbinZhu: As discussed, leaving this for now.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Addressed this existing issue #1795

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
WenbinZhu: Removed this method.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
WenbinZhu: Done.

WenbinZhu:
@Maithem @zalokhan Comments addressed, fix for #1795 is also addressed in this patch

WenbinZhu:
Will add a unit test for the fix of #1795

WenbinZhu:
Added a unit test for cross segment reads

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -417,68 +421,71 @@ public void invalidateServerCaches() {
...
-     * @return A result to be cached
+     * @param lower lower endpoint of the range, included.
+     * @param upper upper endpoint of the range, included.
+     * @return a map of read addresses.
      */
-    public @Nonnull
-    Map<Long, ILogData> fetchRange(Set<Long> addresses) {
-        Map<Long, ILogData> result = layoutHelper(e -> e.getLayout().getReplicationMode(addresses.iterator().next())
+    @Nonnull
+    public Map<Long, ILogData> fetchRange(long lower, long upper, boolean waitForWrite) {
```
lixinchengdu: Why there is no limit on the bulk size for fetchRange() as for fetchAll()?

WenbinZhu: To keep order of result map. FastObjectLoader expects an ordered result.

WenbinZhu: Anyway I'm removing this API to only do sorting in FastObjectLoader, which makes it cleaner.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
```diff
@@ -16,24 +12,20 @@
...
      *
      * @param buf The buffer to deserialize
      */
     public ReadRequest(ByteBuf buf) {
-        range = ICorfuPayload.rangeFromBuffer(buf, Long.class);
-    }
-
-    public ReadRequest(Long address) {
-        range = Range.singleton(address);
+        address = ICorfuPayload.fromBuffer(buf, Long.class);
```
Maithem: fyi `ICorfuPayload.fromBuffer` can block. You can just serialize the message directly. 
Example:
address = buf.readLong();
buf.writeLong(address)

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -37,13 +32,7 @@
...
 
 import javax.annotation.Nonnull;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
Maithem: expand.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -53,6 +42,8 @@
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 
+import static org.corfudb.recovery.RecoveryUtils.*;
```
Maithem: expand

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -728,20 +719,21 @@ public void loadMaps() {
...
-                final long stopNotIncluded = Math.min(start + batchReadSize, logTail + 1);
-                nextRead = stopNotIncluded;
-                final Map<Long, ILogData> range = getLogData(runtime, start, stopNotIncluded);
+                final long lower = nextRead;
+                final long upper = Math.min(lower + batchReadSize - 1, logTail);
+                nextRead = upper + 1;
+                Map<Long, ILogData> range =
+                        runtime.getAddressSpaceView().fetchAll(ContiguousSet.create(
+                                Range.closed(lower, upper), DiscreteDomain.longs()), true);
+                range = new TreeMap<>(range);
```
Maithem: hmm, don't reuse the variable?
I suggest you created a new variable SortedRange

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -1,16 +1,14 @@
...
-
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
+import com.google.common.collect.ContiguousSet;
+import com.google.common.collect.DiscreteDomain;
```
Maithem: unused imports.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -36,17 +34,11 @@
...
 import javax.annotation.Nullable;
 import java.time.Duration;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
Maithem: expand.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -181,6 +172,7 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
             } catch (WriteSizeException we) {
                 throw we;
             } catch (RuntimeException re) {
+                log.error("write: got exception during replication protocol write", re);
```
Maithem: include the token too ?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -268,6 +260,17 @@ ILogData read(long address) {
...
+    }
+
+    /**
+     * Read the given object from a range of addresses.
+     *
+     * @param addresses An iterable with addresses to read from
+     * @param waitForWrite Flag whether wait for write is required or hole fill directly.
+     * @return A map of read addresses, which will be cached if caching is enabled
+     */
+    public Map<Long, ILogData> read(Iterable<Long> addresses, boolean waitForWrite) {
```
Maithem: This is a public API now, can you document how the wait happens. Does it just wait for the greatest address in addresses and then fills the prefix without waiting ?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -417,68 +421,51 @@ public void invalidateServerCaches() {
...
 
         Iterable<List<Long>> batches = Iterables.partition(addresses,
-            runtime.getParameters().getBulkReadSize());
+                runtime.getParameters().getBulkReadSize());
 
         for (List<Long> batch : batches) {
             try {
-                //doesn't handle the case where some address have a different replication mode
+                // Doesn't handle the case where some address have a different replication mode
                 result.putAll(layoutHelper(e -> e.getLayout()
```
Maithem: Should the API guarantee, that the set of addresses to be read is exactly the same set of addresses that were read?

What would happen if the client requests [1,2,3] and the server only returns [1, 3] ?

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -68,83 +71,107 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
     }
 
     /**
-     * {@inheritDoc}
+     * Reads a list of global addresses from the chain of log unit servers.
+     *
+     * - This method optimizes for the time to wait to hole fill in case an empty address
+     * is encountered.
+     * - If the waitForWrite flag is set to true, when an empty address is encountered,
+     * it waits for the hole to be filled. All subsequent empty addresses are hole filled
```
Maithem: > All subsequent empty addresses are hole filled
within the collection, not all subsequent addresses.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -68,83 +71,107 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
+                                       boolean waitForWrite) {
+
+        // A map of log unit server endpoint to addresses it's responsible for
+        Map<String, List<Long>> serverAddressMap = new HashMap<>();
 
+        for (Long address : addresses) {
+            List<String> logServers = runtimeLayout.getLayout().getStripe(address).getLogServers();
+            String logServer = logServers.get(logServers.size() - 1);
+            serverAddressMap.compute(logServer, (server, list) -> {
+                if (list == null) {
```
Maithem: I feel this is clearer:
List<Long> luServer = serverAddressMap.getOrDefault(logServer, new ArrayList());
luServer.add(address);

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -68,83 +71,107 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
+    @Nonnull
+    public Map<Long, ILogData> readAll(RuntimeLayout runtimeLayout,
+                                       List<Long> addresses,
+                                       boolean waitForWrite) {
+
+        // A map of log unit server endpoint to addresses it's responsible for
+        Map<String, List<Long>> serverAddressMap = new HashMap<>();
 
+        for (Long address : addresses) {
+            List<String> logServers = runtimeLayout.getLayout().getStripe(address).getLogServers();
```
Maithem: If you can, I suggest you also use `public LayoutStripe getStripe(long globalAddress)` to implement the "segment finding logic" in `getStripe`

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -1,41 +1,16 @@
...
 import org.corfudb.runtime.CorfuRuntime.CorfuRuntimeParameters;
 import org.corfudb.runtime.RebootUtil;
-import org.corfudb.runtime.clients.BaseHandler;
-import org.corfudb.runtime.clients.IClientRouter;
-import org.corfudb.runtime.clients.LayoutClient;
-import org.corfudb.runtime.clients.LayoutHandler;
-import org.corfudb.runtime.clients.ManagementClient;
-import org.corfudb.runtime.clients.ManagementHandler;
-import org.corfudb.runtime.clients.NettyClientRouter;
+import org.corfudb.runtime.clients.*;
```
Maithem: expand

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
```diff
@@ -7,10 +7,8 @@
 package org.corfudb.runtime.view.replication;
 
 import java.time.Duration;
-import java.util.AbstractMap;
-import java.util.Comparator;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
Maithem: expand. 

Also: https://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports

runtime/src/main/java/org/corfudb/runtime/view/replication/IReplicationProtocol.java
```diff
@@ -63,67 +68,26 @@
...
-     * <p>An implementation may optimize for this type of
-     * bulk request, but the default implementation
-     * just performs multiple peeks (possibly in parallel).
-     *
-     * @param runtimeLayout         The RuntimeLayout stamped with layout to use for the peekAll.
-     * @param globalAddresses       A set of addresses to read from.
-     * @return                      A map of addresses to uncommitted
-     *                              addresses, without hole filling.
-     */
-    default @Nonnull Map<Long, ILogData> peekAll(RuntimeLayout runtimeLayout,
```
Maithem: Hmm, after all it seems like this is not necessary, because we can just hole fill the prefix based on only peeking the greatest address to be read.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -68,83 +71,107 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
-     */
-    public Map<Long, ILogData> readRange(RuntimeLayout runtimeLayout, Range<Long> range, boolean waitForWrite) {
-        long startAddress = range.lowerEndpoint();
-        long endAddress = range.upperEndpoint();
-        int numUnits = runtimeLayout.getLayout().getSegmentLength(startAddress);
-        log.trace("readRange[{}-{}]: chain {}/{}", startAddress, endAddress, numUnits, numUnits);
+    private Map<Long, ILogData> waitOrHoleFill(RuntimeLayout runtimeLayout,
+                                               Map<Long, LogData> readResult,
+                                               boolean waitForWrite) {
+        Map<Long, ILogData> returnResult = new TreeMap<>(readResult);
```
Maithem: Is there any instance where we need to do a multi read, but we don't want order over the addresses? Can't we just use A TreeMap instead of a Map since it seems like we always need order.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -68,83 +71,107 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
-        // the first entry in the list. All subsequent holes in the list can
-        // be hole filled without waiting as we have already waited for the first
-        // hole.
-        boolean wait = !waitForWrite;
-        Map<Long, ILogData> returnResult = new TreeMap<>();
-        for (Map.Entry<Long, LogData> entry : logResult.entrySet()) {
+        // the first entry in the list. All subsequent holes in the list can be
+        // hole filled without waiting as we have already waited for the first hole.
+        boolean wait = true;
+        for (Map.Entry<Long, LogData> entry : readResult.entrySet()) {
```
Maithem: Shouldn't this actually iterate on returnResult?
Please refactor this, changing status values inline is not clear.

Maybe we can do this:
1. Get largest address from the sorted read result
2. Apply read (with hole fill policy) on address from #1 
3. Iterate the sorted read result and try to fill hole from high to low

Also, instead of peek, then hole fill, you can hole fill directly and ignore OverWrite exceptions.

test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
```diff
@@ -169,30 +166,88 @@ public void ensureStripingReadAllWorks()
...
+        final String testString = "hello world ";
+
+        final long numAddresses = 10L;
+        for (long i = 0L; i < numAddresses; i++) {
+            TokenResponse token = rt.getSequencerView().next();
+            rt.getAddressSpaceView().write(token, (testString + i).getBytes());
+        }
+
+        Map<Long, ILogData> readResult = rt.getAddressSpaceView().read(
+                ContiguousSet.create(Range.closed(0L, numAddresses - 1), DiscreteDomain.longs()));
```
Maithem: Disable caching on the client and also test fetchAll. Also, `waitForWrite` option doesn't seem to be directly tested.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
WenbinZhu: Removed. sorting provided in AddressSpaceView

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
WenbinZhu: Done

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
WenbinZhu: Done. Added a sanity check.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Done. Should use `serverAddressMap.computeIfAbsent(logServer, s -> new ArrayList<>());` instead.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Done. Cleaned up getStripe() method.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
WenbinZhu: Done. Thanks for this tip

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: Composed ordering in AddressSpaceView.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: As discussed, for range read, waiting on last address or an arbitrary one are conceptually and effectively same.

test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
WenbinZhu: Done.

WenbinZhu:
@Maithem Thanks for the review. All comments are addressed. For the batch holeFill optimization I proposed, will probably address in another path.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -468,68 +486,59 @@ public void invalidateServerCaches() {
...
-                //doesn't handle the case where some address have a different replication mode
-                result.putAll(layoutHelper(e -> e.getLayout()
+                // Doesn't handle the case where some address have a different replication mode
+                Map<Long, ILogData> batchResult = layoutHelper(e -> e.getLayout()
                         .getReplicationMode(batch.iterator().next())
                         .getReplicationProtocol(runtime)
-                        .readAll(e, new ArrayList<>(batch))));
+                        .readAll(e, batch, waitForWrite));
+                // Sanity check for returned addresses
+                if (batchResult.size() != batch.size()) {
```
Maithem: In your next PRs, please check the addresses as well.

Issue #1846, opened 2019-04-21
Failure detector improvements


xnull:
- Add current epoch in `NodeState`. A `ClusterState` considered as invalid if it contains NodeStates from different epochs. Since the cluster state is obsolete and could be incorrect, the failure detector has to skip the current iteration.

- If a link failure between two nodes is found but the first node is already in the unresponsive list, then the second node can't be added to the list. This current failed node can be skipped and the algorithm continues searching recursively for a next failed node. **[DONE](https://github.com/CorfuDB/CorfuDB/pull/1856)**

- Prevent a situation when a cluster of three nodes `[a, b, c]` contains two unresponsive nodes. By definition, it's impossible to have the majority of nodes in the unresponsive list because the new layout must be accepted by the majority number of nodes.

- Add `LogUnit failures` metric.

- Don't update a layout many times: there are not necessary updates on the layout when we heal/fail node

- Refactor Network stretching out of FailureDetector for better code quality. **[DONE](https://github.com/CorfuDB/CorfuDB/pull/1854)**

- separate `correct wrong epochs` and `filling slot` logic from failure detection.

PR #1856, opened 2019-04-29
Validate epoch in FailureDetector

hisundar:
For proper validation of cluster state in failure detector, ensure that the epoch is uniform

Co-authored-by: xnull xrw.null@gmail.com

## Overview

Description:

Why should this be merged: 

Related issue(s) (if applicable): #1846 

- [x] Public API has Javadoc


zalokhan:
@hisundar Why is this needed? 
The epoch inconsistencies are exposed as WrongEpochExceptions.
The failure detector would not be triggered if the epochs are not in sync.

xnull:
@zalokhan  there can be a race when the layout already updated but one of the node states is still from the previous epoch. Epoch added to prevent that kind of situations and also for a double check and correctness checking which makes failure detector more resilient and consistent.

xnull:
@zalokhan we check epochs of all nodes and if there is an inconsistency then we break current iteration of failure detection. Actually, this change should prevent that situation you described

Issue #1859, opened 2019-04-30
LogUnitServer does not fail if file system is read only


hisundar:
## Overview

What kind of issue is this?
Cluster Availability

If LogUnitServer is opened on a read-only file system, then it comes up correctly only to fail log writes later.

## Deliverables

- [ ] LogUnit Server must not even start up if it the underlying filesystem/directory permissions are incorrect.

PR #1860, opened 2019-04-30
Fail LogUnitServer startup if log directory is read-only

hisundar:
LogUnitServer should not come up on a read-only directory, because
it would later fail log writes. This could result in clustering
and availability issues.

Co-authored-by: xnull <xrw.null@gmail.com>
Issue: #1859

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.FileStore;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -129,6 +130,28 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
      */
     private void initStreamLogDirectory() {
         if (logDir.toFile().exists()) {
+            String corfuDir = logDir.getParent().toString();
+            // If FileSystem is mounted as read-only, Corfu server cannot function.
+            try {
+                FileStore fs = Files.getFileStore(Paths.get(corfuDir));
+                if (fs.isReadOnly()) {
+                    throw new UnrecoverableCorfuError("Cannot start Corfu on a read-only filesystem:"
+                            +corfuDir);
+                }
+            } catch (IOException e) {
+                throw new UnrecoverableCorfuError("Unable to retrieve Corfu Filesystem permissions"+
+                        corfuDir);
+            }
+
+            // corfu dir in the filesystem must be writable for writing configuration files.
+            File corfuDirFile = new File(corfuDir);
+            if (!corfuDirFile.canWrite()) {
+                throw new UnrecoverableCorfuError("Corfu directory is not writable "+corfuDir);
+            }
+            File logDirectory = new File(logDir.toString());
+            if (!logDirectory.canWrite()) {
+                throw new UnrecoverableCorfuError("Stream log directory not writable in "+corfuDir);
+            }
             log.info("Log directory already exists: {}", logDir);
             return;
         }
--- a/test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
+++ b/test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
@@ -2,10 +2,12 @@ package org.corfudb.infrastructure;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import org.apache.commons.io.FileUtils;
 import org.assertj.core.api.Assertions;
 import org.corfudb.infrastructure.log.StreamLogFiles;
 import org.corfudb.protocols.wireprotocol.*;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.util.serializer.Serializers;
 import org.junit.Test;
@@ -13,6 +15,7 @@ import org.junit.Test;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Random;
 import java.util.HashMap;
@@ -21,6 +24,7 @@ import java.util.UUID;
 
 import static org.corfudb.infrastructure.LogUnitServerAssertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
 
 /**
  * Created by mwei on 2/4/16.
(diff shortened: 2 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -268,7 +268,12 @@ private void verifyLogs() {
         for (File file : files) {
             LogHeader header;
 
-            try (FileChannel fileChannel = FileChannel.open(file.toPath())) {
+            // StreamLog will need to be writable, otherwise it isn't considered usable.
```
xnull: *needs to be

hisundar: fixed.

hisundar:
> This assumes that there are *.log files when the server starts, which is a false assumption. In other words, this check is not sufficient.
> 
> This patch tries to infer the file system mount options by doing an operation, but that's not necessary.
> The mount options can be directly inspected using the `FileSystem` API. Please consider using that.

Looks like that case is already handled - Corfu Server fails to come up because it cannot generate a node id. So I think checking for FileSystem again may be an overkill.
java.lang.RuntimeException: java.nio.file.NoSuchFileException: /rodir/corfu/_NODE_ID.ds.tmp
	at org.corfudb.infrastructure.DataStore$1.write(DataStore.java:145)
	at org.corfudb.infrastructure.DataStore$1.write(DataStore.java:116)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.put(BoundedLocalCache.java:1567)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.put(BoundedLocalCache.java:1491)
Please correct me if I am wrong?

Maithem:
> > This assumes that there are *.log files when the server starts, which is a false assumption. In other words, this check is not sufficient.
> > This patch tries to infer the file system mount options by doing an operation, but that's not necessary.
> > The mount options can be directly inspected using the `FileSystem` API. Please consider using that.
> 
> Looks like that case is already handled - Corfu Server fails to come up because it cannot generate a node id. So I think checking for FileSystem to be read-only once again may be an overkill.
> java.lang.RuntimeException: java.nio.file.NoSuchFileException: /rodir/corfu/_NODE_ID.ds.tmp
> at org.corfudb.infrastructure.DataStore$1.write(DataStore.java:145)
> at org.corfudb.infrastructure.DataStore$1.write(DataStore.java:116)
> at com.github.benmanes.caffeine.cache.BoundedLocalCache.put(BoundedLocalCache.java:1567)
> at com.github.benmanes.caffeine.cache.BoundedLocalCache.put(BoundedLocalCache.java:1491)
> Please correct me if I am wrong?

The node id is generated once, not on every boot.

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -129,6 +131,61 @@ public void checkThatWritesArePersisted()
...
+                assertThat("Should not hit security exception"+e.toString());
+            }
+        }
+
+        boolean readOnlyFailed = false;
+        try {
+            LogUnitServer s2 = new LogUnitServer(new ServerContextBuilder()
+                    .setLogPath(serviceDir)
+                    .setMemory(false)
+                    .build());
```
zalokhan: Similarly, use `fail()` if log unit starts successfully.

hisundar:
> > > This assumes that there are *.log files when the server starts, which is a false assumption. In other words, this check is not sufficient.
> > > This patch tries to infer the file system mount options by doing an operation, but that's not necessary.
> > > The mount options can be directly inspected using the `FileSystem` API. Please consider using that.
> > 
> > 
> > Looks like that case is already handled - Corfu Server fails to come up because it cannot generate a node id. So I think checking for FileSystem to be read-only once again may be an overkill.
> > java.lang.RuntimeException: java.nio.file.NoSuchFileException: /rodir/corfu/_NODE_ID.ds.tmp
> > at org.corfudb.infrastructure.DataStore$1.write(DataStore.java:145)
> > at org.corfudb.infrastructure.DataStore$1.write(DataStore.java:116)
> > at com.github.benmanes.caffeine.cache.BoundedLocalCache.put(BoundedLocalCache.java:1567)
> > at com.github.benmanes.caffeine.cache.BoundedLocalCache.put(BoundedLocalCache.java:1491)
> > Please correct me if I am wrong?
> 
> The node id is generated once, not on every boot.

Got it. Fixed. Could you please help re-review the additional 2 lines? Thanks.

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -129,6 +131,61 @@ public void checkThatWritesArePersisted()
                 .matchesDataAtAddress(HIGH_ADDRESS, high_payload.getBytes());
     }
 
+    @Test
```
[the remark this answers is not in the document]
hisundar: Done.

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -129,6 +131,61 @@ public void checkThatWritesArePersisted()
...
+        String[] extension = {"log"};
+        File serviceDirectory = new File(serviceDir);
+        Collection<File> files = FileUtils.listFiles(serviceDirectory, extension, true);
+
+        for (File file : files) {
+            try {
+                boolean ok = file.setReadable(false);
+                assertThat(ok).isEqualTo(true);
+            } catch (SecurityException e) {
+                assertThat("Should not hit security exception"+e.toString());
```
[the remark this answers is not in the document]
hisundar: Done. Thanks

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
hisundar: Done. thanks

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -128,6 +129,10 @@ private long getStartingSegment() {
      * Create stream log directory if not exists
      */
     private void initStreamLogDirectory() {
+        File corfuDir = new File(logDir.getParent().toString());
+        if (!corfuDir.canWrite()) {
```
Maithem: Do you want to include the path as well?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -288,7 +293,12 @@ private void verifyLogs() {
         for (File file : files) {
             LogHeader header;
 
-            try (FileChannel fileChannel = FileChannel.open(file.toPath())) {
+            // StreamLog needs to be writable, otherwise it isn't considered usable.
+            EnumSet<StandardOpenOption> options = EnumSet.of(
+                    StandardOpenOption.READ, StandardOpenOption.WRITE,
```
Maithem: This is a read only operation, please remove StandardOpenOption.WRITE

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: I believe this only check file permisions, but doesn't reflect the filesystem mount option. 

Please use:
File f = new File("/home/maithem/CorfuDB/n1");
FileStore fs = Files.getFileStore(Paths.get("/home/maithem/CorfuDB/n1"));
fs.isReadOnly();

hisundar: Fixed. Thanks

hisundar: Added this check in addition to corfu directory and log directory checks.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -128,7 +130,26 @@ private long getStartingSegment() {
      * Create stream log directory if not exists
      */
     private void initStreamLogDirectory() {
+        // If FileSystem is mounted as read-only, Corfu server cannot function.
+        try {
+            FileStore fs = Files.getFileStore(Paths.get(logDir.getParent().toString()));
```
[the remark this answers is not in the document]
hisundar: Even if the directory does not exist, it should be able to retrieve the filesystem that it is mounted on which can then be used to test for permissions. Please correct me if I am mistaken.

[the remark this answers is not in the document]
Maithem: It will actually fail with an `java.nio.file.NoSuchFileException` , but throw `UnrecoverableCorfuError("Unable to retrieve Corfu Filesystem permissions")` 

Also, on failure, please print the `logDir` its more useful.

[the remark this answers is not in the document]
hisundar: done. fixed.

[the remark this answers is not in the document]
Maithem: Actually, it doesn't matter, because both would fail the initialization.

xnull:
@hisundar please make unit tests green https://travis-ci.org/CorfuDB/CorfuDB/jobs/528942907

hisundar:
> @hisundar please make unit tests green https://travis-ci.org/CorfuDB/CorfuDB/jobs/528942907

I am really not sure why on the Travis runs, the corfu directory is marked as read-only. On Mac and linux, I don't see this issue and these failing tests always pass.

PR #1870, opened 2019-05-02
Fix Log Unit Handler Test

annym:
## Overview

Move Log Unit test to infrastructure module so we can access the BatchProcessor for a test that intermittently fails because of async executions on the processor.

### DIFF
--- a/test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
+++ b/test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
@@ -6,6 +6,10 @@ import org.assertj.core.api.Assertions;
 import org.corfudb.infrastructure.log.StreamLogFiles;
 import org.corfudb.protocols.wireprotocol.*;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.clients.BaseHandler;
+import org.corfudb.runtime.clients.LogUnitClient;
+import org.corfudb.runtime.clients.LogUnitHandler;
+import org.corfudb.runtime.clients.TestClientRouter;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.util.serializer.Serializers;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


annym:
> #1869 fixed the intermittent log unit test failure, is this still required?

@WenbinZhu this is a different test, actually a test that is part of the log unit handler tests, which needs to be moved to this other module so we can wait on the batchProcessor before verifying a given condition.

test/src/test/java/org/corfudb/runtime/clients/LogUnitHandlerTest.java
```diff
@@ -180,35 +180,7 @@ public void writeRange() throws Exception {
...
-        client.write(address1, Collections.<UUID>emptySet(), null, testString, Collections.emptyMap()).get();
-        LogData r = client.read(address0).get().getAddresses().get(0L);
-        assertThat(r.getType())
-                .isEqualTo(DataType.DATA);
-        r = client.read(address1).get().getAddresses().get(1L);
-        assertThat(r.getType())
-                .isEqualTo(DataType.DATA);
-
-        client.prefixTrim(new Token(0L, address0));
-        client.compact();
```
Maithem: Tests using the client directly should not be affected by the async processing in the server.
I can't reproduce this, but it might be failing because the client doesn't wait on prefix trim and compact. Maybe you can do:
client.prefixTrim(new Token(0L, address0)).get();
client.compact().get();

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -439,6 +444,62 @@ public void CheckCacheSizeIsCorrectRatio() throws Exception {
         assertThat(s1).hasCorrectCacheSize(randomCacheRatio);
     }
 
+    @Test
+    public void readingTrimmedAddress() throws Exception {
+        LogUnitServer server1 = getLogUnitServer();
+        LogUnitServer server2 = null;
```
lixinchengdu: The declaration of servers should be closed to its usage.

Maithem:
Do we still want this?

Maithem:
@annym  lets close this, please address my comments and rebase.

annym:
We can close this, I haven't been able to reproduce it either after all other changes.

PR #1877, opened 2019-05-04
Fix async prefix trim issue

annym:
## Overview

Description:

Because trim's are async (between the sequencer and the log units), the checkpoint's address space--as viewed by the sequencer, might not reflect the actual state of valid addresses, when addresses have already been trimmed from the log unit but prefix trim message has not reached the sequencer. 

For this reason--in the context of address maps, we should not be reading the complete map of checkpoint addresses in a batch, as (1) not all addresses are required (the last checkpoint subsumes all other checkpoints) and (2) part of this space might be trimmed continuously causing trim exceptions. 

Why should this be merged: this corrects the address build behavior and corrects intermittent failure of the preiodicCkpointTrimTest (which fails because checkpoint is performing faster so between reading a checkpoint and updates to the regular stream, a new checkpoint and trim point have been generated which. invalidates some address of the regular stream).


annym:
@WenbinZhu I'll leave the result of our offline conversation here for others following this thread:

1. We do agree that the intermittent test failure is solved by admitting TrimmedExceptions for that test, as it is possible that if we are checkpointing and trimming continuously (depending on how fast this goes) we might hit trimmed positions on two continuous cycles (given the nature of the test). 

2. However, there are disadvantages of reading all checkpoint entries in a single batch:
   (a) There might be more than one checkpoint and not all of them might be required.
         The latest checkpoint subsumes all other checkpoints (we could be reading
         more data than necessary).
   (b) Because trims are async (between sequencer and log unit), part of these
         addresses (for instance for the first checkpoint) might have been already
         trimmed from the log, but stream maps (sequencer) still do not reflect it
         (causing unnecessary TrimmedExceptions--as this checkpoint is not even
         needed in the first place). Even in the case if all trim messages are lost or delayed 
         for the sequencer, we should be able to recover, for this reason we should be able to discard 
        trimmed entries still reflected in the address maps---if not required.

**Proposed Solution** 
Since reading one entry at a time is slower, we adopt an intermediate solution, which is actually consistent with the fact that we need to limit our read size to avoid bottlenecks. We'll read checkpoints in batches of a predefined size, and in the case the stop condition is met we stop reading (avoids reading more data than the necessary)... also if we hit a trimmedException, we will default to read this last batch one at a time, if stop condition is met we continue, if it is not met and we do hit the trimmed exception again we bubble this to upper layers as expected.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -193,36 +193,79 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
-            BackpointerOp op = filter.apply(data);
-            if (op == BackpointerOp.INCLUDE || op == BackpointerOp.INCLUDE_STOP) {
-                log.trace("getStreamAddressMap[{}]: Adding checkpoint address[{}] to queue",
-                        this, data.getGlobalAddress());
-                queue.add(data.getGlobalAddress());
-                // Check if we need to stop
-                if (op == BackpointerOp.INCLUDE_STOP) {
-                    break;
-                }
+    private boolean filterCheckpointEntry(ILogData data, Function<ILogData, BackpointerOp> filter,
```
zalokhan: Add java doc.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -193,36 +193,79 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
+                        break;
+                    }
+                }
+            } catch (TrimmedException te) {
+                // Read one entry at a time for the last failed batch, this way we might load
+                // the required checkpoint entries until the stop condition is fulfilled
+                // without hitting a trimmed position. If we do hit a trim it is an actual
+                // TrimmedException
+                long lastReadAddress = Address.NON_ADDRESS;
+                try {
```
zalokhan: Can we refactor this block?
It seems off throwing a trimmed exception inside a catch block of a trimmed exception.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -193,36 +193,79 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
+                    for (long address : batch) {
```
zalokhan: This code looks repeated. Can we refactor this to a method with a flag for single-step or batch.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -400,9 +400,8 @@ protected boolean fillReadQueue(final long maxGlobal,
                     return true;
                 }
             } catch (TrimmedException te) {
-                // If we reached a trim and didn't hit a checkpoint, this might be okay,
-                // if the stream was created recently and no checkpoint exists yet.
-                log.warn("Fill_Read_Queue[{}] Trim encountered and no checkpoint detected.", this);
+                log.warn("Fill_Read_Queue[{}] Trim encountered.", this);
+                throw te;
```
WenbinZhu: Can we add a comment here how throwing a TrimmedException here will later reset the object and fetch the latest checkpoint? Also if you want, maybe not in this patch, you can add a testcase later to test this situation to see if after TrimmedException, the latest checkpoint is correctly retrieved.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -193,36 +193,109 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
+                ILogData data = read(address);
+                checkpointResolved = filterCheckpointEntry(data, filter, queue);
+                if (checkpointResolved) {
+                    // Return if checkpoint has already been resolved (reached stop).
+                    return true;
+                }
+            }
+        } catch (TrimmedException ste) {
+            // The valid checkpoint has been trimmed.
             if (options.ignoreTrimmed) {
```
WenbinZhu: I think we should put this ignoreTrimmed option check into `processTrimmedException`, for two reasons: 1) cleaner code organization 2) `processTrimmedException` is basically checking if this TrimmedException is retryable. If we ignore trim, there is no point doing this retryable check, right?

Issue #1878, opened 2019-05-07
Guava Cache Throwing Assertion Errors


Maithem:
**Bug report.**
This failure is not expected to happen. 

`java.lang.AssertionError
        at com.google.common.cache.LocalCache$Segment.getNextEvictable(LocalCache.java:2653)
        at com.google.common.cache.LocalCache$Segment.evictEntries(LocalCache.java:2637)
        at com.google.common.cache.LocalCache$Segment.storeLoadedValue(LocalCache.java:3158)
        at com.google.common.cache.LocalCache$Segment.getAndRecordStats(LocalCache.java:2316)
        at com.google.common.cache.LocalCache$Segment.compute(LocalCache.java:2250)
        at com.google.common.cache.LocalCache.compute(LocalCache.java:4196)
        at com.google.common.cache.LocalCache.computeIfAbsent(LocalCache.java:4203)
        at org.corfudb.runtime.view.AddressSpaceView.read(AddressSpaceView.java:254)
        at org.corfudb.runtime.view.stream.BackpointerStreamView.read(BackpointerStreamView.java:214)
        at org.corfudb.runtime.view.stream.BackpointerStreamView.followBackpointers(BackpointerStreamView.java:322)
        at org.corfudb.runtime.view.stream.BackpointerStreamView.fillReadQueue(BackpointerStreamView.java:532)
        at org.corfudb.runtime.view.stream.AbstractQueuedStreamView.getNextEntries(AbstractQueuedStreamView.java:136)
        at org.corfudb.runtime.view.stream.AbstractQueuedStreamView.getNextEntries(AbstractQueuedStreamView.java:40)
        at org.corfudb.runtime.view.stream.AbstractContextStreamView.remainingUpTo(AbstractContextStreamView.java:164)
        at org.corfudb.runtime.view.stream.ThreadSafeStreamView.remainingUpTo(ThreadSafeStreamView.java:89)
        at org.corfudb.runtime.view.stream.IStreamView.remaining(IStreamView.java:170)`

WenbinZhu:
Maybe we can open an issue in Guava's  repo and ask the maintainers about this AssertionError to save some time?

Maithem:
@WenbinZhu I will do that, first I'm trying to reproduce it.

Maithem:
Already opened an issue on their repo
https://github.com/google/guava/issues/3464

ben-manes:
[...]

PR #1879, opened 2019-05-07
Reintroducing the Loading Cache

Maithem:
## Overview
It seems like once we switched to asMap().computeIfAbsent we
started seeing spurious AssertionError. Although the problem
is not completely understood this change is being reverted
until it is investigated further.

Why should this be merged: 

Related issue(s) (if applicable): #1878

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -5,7 +5,10 @@ import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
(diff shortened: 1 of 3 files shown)


apavan:
Please do the needful to merge and start a new Corfu build

Maithem:
> Should we add the `checkLogData()` back? Because patches after the cache fix patch assumes the empty data check is done in addressSpaceView, right?

fetch and fetchAll do call `checkLogData` and the cache loader methods call those two methods. 
Do you see any path where the log entries are not checked post retrieval ?

WenbinZhu:
> > Should we add the `checkLogData()` back? Because patches after the cache fix patch assumes the empty data check is done in addressSpaceView, right?
> 
> fetch and fetchAll do call `checkLogData` and the cache loader methods call those two methods.
> Do you see any path where the log entries are not checked post retrieval ?

Oh ok my mistake, I forgot the check was put to fetch/fetchAll

Issue #1880, opened 2019-05-08
Get rid of LogData deserialization dependency on CorfuRuntime 


WenbinZhu:
## Overview

Seems like we used to serialize the CorfuObject to the global log data, which introduced a dependency on CorfuRuntime when deserializing the payload from log data. We need to remove this dependency as we're not serialize the whole CorfuObject now.

Maithem:
@WenbinZhu please merge the revert and close this.

WenbinZhu:
@Maithem This requires changing production code, but our current production release is no longer accepting changes other than bug fixes. I will do this for next production release branch once it's created.

PR #1881, opened 2019-05-08
Serializer: get rid of Runtime dependency on ILogData::getPayload.

WenbinZhu:
## Overview

Description:

For some historical reasons, the deserialization of LogData's payload
depends on CorfuRuntime. This patch gets rid of this dependency.

Related issue(s) (if applicable): #1880 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
@@ -7,7 +7,6 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import com.google.common.annotations.VisibleForTesting;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.ISMRConsumable;
 import org.corfudb.protocols.logprotocol.SMREntry;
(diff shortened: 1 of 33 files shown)


runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -37,35 +36,29 @@
...
                 cp.getSmrEntries().getUpdates().forEach(e -> {
-                    e.setRuntime(runtime);
                     e.setEntry(logData);
                 });
                 return cp.getSmrEntries().getUpdates();
-            } else {
-                return (List<SMREntry>) Collections.EMPTY_LIST;
             }
+            return Collections.emptyList();
+        } else if (payload instanceof ISMRConsumable) {
```
lixinchengdu: Should payload being of ISMRConsumable type a pre-condition of this function as indicated by the name?

runtime/src/main/java/org/corfudb/util/serializer/JsonSerializer.java
```diff
@@ -39,31 +38,17 @@ public byte getType() {
...
     @Override
-    public Object deserialize(ByteBuf b, CorfuRuntime rt) {
+    public Object deserialize(ByteBuf b) {
         int classNameLength = b.readShort();
         byte[] classNameBytes = new byte[classNameLength];
         b.readBytes(classNameBytes, 0, classNameLength);
         String className = new String(classNameBytes);
         if (className.equals("null")) {
             return null;
-        } else if (className.equals("CorfuObject")) {
```
lixinchengdu: Is CorfuObject deprecated?

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
WenbinZhu: ISMRConsumable is the normal use case, the check is like a sanity check

runtime/src/main/java/org/corfudb/util/serializer/JsonSerializer.java
WenbinZhu: Yes

PR #1886, opened 2019-05-09
Remove load all

Maithem:
## Overview
The purpose of this patch is to load values into the cache without
overwriting existing values (if they are already cached). Overwriting
cache entries can result in a loss of undo records; this patch
attempts to address that issue.

Why should this be merged: hopefully addresses #1878 and improves the bitset implementation of the stream layer

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -6,6 +6,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Iterables;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.netty.handler.timeout.TimeoutException;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.DataType;
@@ -41,6 +42,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 
@@ -249,12 +251,10 @@ public class AddressSpaceView extends AbstractView {
                 // is much cheaper than the cost of a NoRollBackException, therefore
                 // this trade-off is reasonable
                 final ILogData loadedVal = fetch(address);
-                return readCache.asMap().computeIfAbsent(address, (k) -> loadedVal);
+                return cacheLoadAndGet(readCache, address, loadedVal);
             }
-
-            return data;
-        }
-
+        return data;
+    }
         return fetch(address);
     }
 
@@ -301,11 +301,30 @@ public class AddressSpaceView extends AbstractView {
     }
 
     /**
+     * Attempts to insert a loaded value into the cache and return the cached value for a particular key.
+     */
+    private ILogData cacheLoadAndGet(Cache<Long, ILogData> cache, long address, ILogData loadedValue) {
+        try {
+            return cache.get(address, () -> loadedValue);
+        } catch (ExecutionException | UncheckedExecutionException e) {
+            // Guava wraps the exceptions thrown from the lower layers, therefore
+            // we need to unwrap them before throwing them to the upper layers that
+            // don't understand the guava exceptions
+            Throwable cause = e.getCause();
+            if (cause instanceof RuntimeException) {
+                throw (RuntimeException) cause;
+            } else {
+                throw new RuntimeException(cause);
+            }
+        }
+    }
+
+    /**
      * Read the given object from a range of addresses.
      *
      * @param addresses An iterable with addresses to read from
      * @param waitForWrite Flag whether wait for write is required or hole fill directly.
-     * @return A map of read addresses, which will be cached if caching is enabled
+     * @return A map of addresses read, which will be cached if caching is enabled
      */
     public Map<Long, ILogData> read(Iterable<Long> addresses, boolean waitForWrite) {
         Map<Long, ILogData> result = new HashMap<>();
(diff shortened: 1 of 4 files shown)


Maithem:
> Can you please add a test that ensures that regression #1878 does not happen in the future.

I can't reproduce the issue, so I can't really test it. I'm just using a different interface that we didn't have this issue with. Although the java doc says that `asMap` implementation is thread-safe, the code path that's failing the assertion indicates that there's a race issue.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -238,8 +226,25 @@ public void write(IToken token, Object data) throws OverwriteException {
...
+            //    tries to read write1
+            // 3. Thread B's read results in a cache miss and the reader thread
+            //    starts loading the value into the cache
+            // 4. Thread A completes its write and caches it with undo records
+            // 5. Thread B finishes loading and caches the loaded value replacing
+            //    the cached value from step 4 (i.e. loss of undo records computed
+            //    by thread A)
             try {
-                return readCache.get(address);
+                readCache.get(address, () -> fetch(address));
```
WenbinZhu: Putting `fetch` inside `get()` might not be good. If this api blocks writer, writer can starve because `fetch` can really take a long time due to hole fills. If this api doesn't block writer, then the NoRollBackException issue will still occur quite often as this reader may override the writer's value with no undo record.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -294,23 +300,66 @@ ILogData read(long address) {
...
+            }
+
+            // At this point we computed a subset of the addresses that
+            // resulted in a cache miss and need to be fetched
+            if (!addressesToFetch.isEmpty()) {
+                Map<Long, ILogData> fetchedAddresses = fetchAll(addressesToFetch, waitForWrite);
+                for (Map.Entry<Long, ILogData> entry : fetchedAddresses.entrySet()) {
+                    // After fetching a value, we need to insert it in the cache.
+                    // Note that based on code inspection it seems like operations
+                    // on the cache's map view are reflected in the cache's statistics.
```
WenbinZhu: Fix this comment? It seems to be obsolete comment from that old patch.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: I have the same question, wouldn't we still be overwriting the writers value? shouldn't it also go though the cacheLoadAndGet?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -294,23 +300,66 @@ ILogData read(long address) {
...
+            }
+        }
+    }
+
+    /**
+     * Read the given object from a range of addresses.
+     *
+     * @param addresses An iterable with addresses to read from
+     * @param waitForWrite Flag whether wait for write is required or hole fill directly.
+     * @return A map of read addresses, which will be cached if caching is enabled
```
hisundar: micro nit: @return A map of addresses read, which will be cached if caching is enabled
seems to be clearer in stating that the returned values are addresses that have already been read?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Based on code inspection the puts and get/load are protected by a segment lock, but there is another issue related to concurrency. I'll remove the fetch from the load.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: It is clearer. Changed. Thanks for the suggestion.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: done.

Maithem:
> Why can't it be reproduced?
I don't know. I wrote a driver that simulates the poller + read and write and I can't see the issue.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Actually, one of the tests was failing because of this.

Maithem:
Also, since the error was being thrown by the cache itself, I don't see a strong reason to write a test against  Guava's interface. Its not a regression in our code per-se, its just unexpected behavior from the cache interface.

apavan:
> The issue was being observed very consistently. Why can't it be reproduced?

We started observing the issue after changing loadall to map interface. Exception stack trace clearly shows the origin of the issue. Now we are reverting back to old way with a minor tweak. If we are not going to use the map interface, i don't think we should be spending any time on this effort.

apavan:
> I am sorry, but "hopefully addresses" does not cut it. We need to be able to trigger this issue.
> 
> If you are having trouble reproducing the bug, you can do two things:
> 
> 1. Take a look at the failed Colibri tests and instrument the code to the point where you understand what the problem is and how to reproduce it.
> 2. Take a look at nsx-platform KV Store tests which seemed to have caught the regression.

We started observing the issue after changing loadall to map interface. Exception stack trace clearly shows the origin of the issue. Now we are reverting back to old way with a minor tweak. If we are not going to use the map interface, i don't think we should be spending any time on this effort.

test/src/test/java/org/corfudb/integration/TransactionStreamIT.java
```diff
@@ -0,0 +1,148 @@
...
+
+        producers.shutdown();
+        consumer.shutdown();
+
+        Map<UUID, Integer> counters = consumerState.get(timeout, TimeUnit.SECONDS);
+        assertThat(counters).hasSize(numWriters);
+        // Since we know the size of the maps and the written ranges, we can verify that
+        // all updates have been received on the pollers end by computing the sum of all
+        // the writes. The total sum of writes between [1, n] is n * (n + 1) / 2.
+        int sumOfWritesPerTable = (numWritesPerThread * (numWritesPerThread + 1)) / 2;
```
hisundar: Should this be a long in case someone wants to try large values?

PR #1888, opened 2019-05-11
Fix PrefixTrim Bug on Address Maps

annym:
## Overview

Description: when attempting to trim the bitmaps on a negative value, the rankLong function will return the cardinality of infinity, i.e., the complete address map, thus trimming all addresses from the sequencer (when non should've been trimmed as trim mark was -1). This can lead to potential issues in which a stream cannot be rebuilt as all addresses appear to be trimmed and not covered by a checkpoint. This PR fixes this issue.

Why should this be merged: fix a potential bug detected on an intermittent test failure.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
@@ -5,6 +5,8 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NavigableSet;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.UUID;
 import java.util.function.Function;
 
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -194,8 +194,7 @@ private void processCheckpoint(StreamAddressSpace streamAddressSpace, Function<I
                                    NavigableSet<Long> queue) {
         List<Long> checkpointAddresses = new ArrayList<>();
         streamAddressSpace.getAddressMap().forEach(checkpointAddresses::add);
-        checkpointAddresses.sort(null);
-        Collections.reverse(checkpointAddresses);
+        checkpointAddresses.sort(Collections.reverseOrder());
```
xnull: can we replace List by SortedSet?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -108,21 +109,23 @@ public void removeAddresses(List<Long> addresses) {
      * @param trimMark upper limit of addresses to trim
      */
     public void trim(Long trimMark) {
-        long numAddressesToTrim = addressMap.rankLong(trimMark);
+        if (Address.isAddress(trimMark)) {
```
xnull: please invert if condition

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
annym: Done.

PR #1892, opened 2019-05-17
Don't add fully connected node into unresponsive list if there is failed node

xnull:
## Overview

Description:
Failed connection(s) between unresponsive and fully connected node(s) can't be used to determine if a fully connected node is failed. 

If a failed node already added in the unresponsive list then a fully connected node can't be added into the unresponsive list because we cannot believe to the cluster state, based on information about the connection between the failed node and fully connected node. 
Which means the failed node is already in the unresponsive list, we don't need to add the fully connected node also into the unresponsive list.  

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/management/failuredetector/ClusterGraph.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/management/failuredetector/ClusterGraph.java
@@ -109,8 +109,8 @@ public class ClusterGraph {
 
                         //If current node is not the local node and another node is unavailable we don't change
                         // the adjacent node connectivity matrix, we leave it as is
-                        if (adjNode.getType() == NodeConnectivityType.UNAVAILABLE){
-                            if (!isLocalNode(node)){
+                        if (adjNode.getType() == NodeConnectivityType.UNAVAILABLE) {
+                            if (!isLocalNode(node)) {
                                 newConnectivity.put(adjNodeName, node.getConnectionStatus(adjNodeName));
                                 return;
                             }
@@ -124,7 +124,7 @@ public class ClusterGraph {
                         //Symmetric failure - connection successful only if both nodes connected status is true
                         //in the other case - make the failure symmetric
                         ConnectionStatus status = ConnectionStatus.OK;
-                        if (EnumSet.of(nodeConnection, oppositeNodeConnection).contains(ConnectionStatus.FAILED)){
+                        if (EnumSet.of(nodeConnection, oppositeNodeConnection).contains(ConnectionStatus.FAILED)) {
                             status = ConnectionStatus.FAILED;
                         }
                         newConnectivity.put(adjNodeName, status);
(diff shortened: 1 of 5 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/management/failuredetector/ClusterGraph.java
```diff
@@ -212,6 +212,21 @@ private boolean isLocalNode(NodeConnectivity node) {
             return Optional.empty();
         }
 
+        Optional<NodeRank> fullyConnected = findFullyConnectedNode(
```
PavelZaytsev: Leave a comment with an example as for why this step is needed.

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterGraphTest.java
```diff
@@ -30,142 +34,146 @@ public void testTransformation() {
         NodeState a = NodeState.builder()
                 .sequencerMetrics(SequencerMetrics.READY)
                 .heartbeat(new HeartbeatTimestamp(0, 0))
-                .connectivity(connectivity("a", ImmutableMap.of("a", OK, "b", OK, "c", FAILED)))
+                .connectivity(connectivity(A, ImmutableMap.of(A, OK, B, OK, C, FAILED)))
```
lixinchengdu: Why change the test naming convention?

xnull: don't get your question

infrastructure/src/main/java/org/corfudb/infrastructure/management/failuredetector/ClusterGraph.java
xnull: done

infrastructure/src/test/java/org/corfudb/infrastructure/management/ClusterGraphTest.java
lixinchengdu: I am sorry for the confusion. I mean why change "a" to A? Other tests seems still use "a" instead of A.

xnull: 'A' is just a constant in `NodeStateTestUtil`. The node name is still lower case.
Don't want to repeat string names everywhere, so I decided to bring constants. Also, I can't replace old string names everywhere at one shot, will be doing that gradually

xnull:
If:
1. the node is not fully connected
2. has the smallest number of sccessful connections

then we mark the node as a failed node

infrastructure/src/main/java/org/corfudb/infrastructure/management/failuredetector/ClusterGraph.java
```diff
@@ -198,7 +198,7 @@ private boolean isLocalNode(NodeConnectivity node) {
      *
      * @return failed node
      */
-    public Optional<NodeRank> findFailedNode() {
+    public Optional<NodeRank> findFailedNode(List<String> unresponsiveServers) {
```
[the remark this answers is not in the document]
xnull: done

PR #1897, opened 2019-05-17
Log stacktrace on CorfuCompileProxy failures. Access/Update only log …

epalta:
…the exception name omitting the stack trace. We should log both.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -184,7 +184,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
                 return TransactionalContext.getCurrentContext()
                         .access(this, accessMethod, conflictObject);
             } catch (Exception e) {
-                log.error("Access[{}] Exception: {}", this, e);
+                log.error("Access[{}]", this, e);
                 this.abortTransaction(e);
             }
         }
@@ -237,7 +237,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
                 return TransactionalContext.getCurrentContext()
                         .logUpdate(this, entry, conflictObject);
             } catch (Exception e) {
-                log.warn("Update[{}] Exception: {}", this, e);
+                log.warn("Update[{}]", this, e);
                 this.abortTransaction(e);
             }
         }


runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -184,7 +184,7 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
                 return TransactionalContext.getCurrentContext()
                         .access(this, accessMethod, conflictObject);
             } catch (Exception e) {
-                log.error("Access[{}] Exception: {}", this, e);
+                log.error("Access[{}]", this, e);
```
Maithem: please fix the error logs for logUpdateInner as well.

lixinchengdu: Can I know why previous one fails? Why there is only one parameter placeholder but there are another two arguments?

hisundar: This was super confusing to me as well. It seems the last parameter to log.error() should be an Exception, so that parameter is eaten up by the error() method, and the only parameter printed is "this". I think this interface is utterly confusing and warrants a comment wherever it is used.

WenbinZhu: @lixinchengdu The previous one prints out `this` and `e.toString()` (without stacktrace), while the new one is another overloaded api, which treats `e` as a `Throwable` and prints the stacktrace

lixinchengdu: > @lixinchengdu The previous one prints out `this` and `e.toString()` (without stacktrace), while the new one is another overloaded api, which treats `e` as a `Throwable` and prints the stacktrace

I didn't find any match function in slf4j Logger.

xnull: @WenbinZhu right, his change is correct, prev version was wrong

WenbinZhu: Actually I need to take back my comment on the previous implementation. I think it was correctly printing out the stacktrace, just a minor issue: the second placeholder is printed as it is: `{}`. This seems because a Throwable at the end is always treated as a Throwable and the stacktrace is printed. The mismatch of the number of placeHolders will be neglected. So both implementations are actually correct, but the new one is cleaner. It can be easily verified.

Maithem: The format string will call `toString()` on the  `{}` place holders, which will result printing the exception name and not the stack trace. But slf4j gives you the ability to pass a throwable and it will print its stack trace (no place holder required).

`void error(String var1, Throwable var2);`

hisundar: I am not saying that the interface or the usage is incorrect. I am just voicing my opinion that having such an overloaded interface in slf4j is really confusing and not obvious. Its almost like a secret trove that is only known to a privileged few who go digging there. That's why I felt adding a comment maybe helpful.

PR #1901, opened 2019-05-20
Introducing non-cacheable reads on log unit server.

WenbinZhu:
## Overview

- For batch read scenarios like state transfer and fast loader, caching
the read results on log unit server would pollute the cache. This patch
adds the capability of not caching the read results in these situations.

- Separates log unit cache into a new class for more readability and
    testability.

Why should this be merged: 
Improve performance on log unit server by preventing cache pollution.

Related issue(s) (if applicable): #1841 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -1,27 +1,10 @@
 package org.corfudb.infrastructure;
 
-import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.LoadingCache;
-import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.google.common.annotations.VisibleForTesting;
-
 import io.netty.channel.ChannelHandlerContext;
-
-import java.lang.invoke.MethodHandles;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-
 import lombok.Builder;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.infrastructure.log.InMemoryStreamLog;
 import org.corfudb.infrastructure.log.StreamLog;
 import org.corfudb.infrastructure.log.StreamLogCompaction;
@@ -52,6 +35,17 @@ import org.corfudb.runtime.exceptions.ValueAdoptedException;
 import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.view.stream.StreamAddressSpace;
 import org.corfudb.util.Utils;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import static org.corfudb.infrastructure.BatchWriterOperation.Type.LOG_ADDRESS_SPACE_QUERY;
 import static org.corfudb.infrastructure.BatchWriterOperation.Type.PREFIX_TRIM;
@@ -101,7 +95,7 @@ public class LogUnitServer extends AbstractServer {
      * it is not backed by anything, but in a disk implementation it is backed by persistent
      * storage.
      */
-    private final LoadingCache<Long, ILogData> dataCache;
+    private final LogUnitServerCache dataCache;
     private final StreamLog streamLog;
     private final StreamLogCompaction logCleaner;
     private final BatchProcessor batchWriter;
@@ -141,13 +135,8 @@ public class LogUnitServer extends AbstractServer {
             streamLog = new StreamLogFiles(serverContext, config.isNoVerify());
         }
 
+        dataCache = new LogUnitServerCache(config, streamLog);
         batchWriter = new BatchProcessor(streamLog, serverContext.getServerEpoch(), !config.isNoSync());
-
-        dataCache = Caffeine.newBuilder()
-                .<Long, ILogData>weigher((k, v) -> ((LogData) v).getData() == null ? 1 : ((LogData) v).getData().length)
-                .maximumWeight(config.getMaxCacheSize())
-                .removalListener(this::handleEviction)
-                .build(this::handleRetrieval);
 
         logCleaner = new StreamLogCompaction(streamLog, 10, 45, TimeUnit.MINUTES, ServerContext.SHUTDOWN_TIMER);
     }
@@ -297,16 +286,18 @@ public class LogUnitServer extends AbstractServer {
     }
 
     @ServerHandler(type = CorfuMsgType.READ_REQUEST)
-    private void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
+    public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
         long address = msg.getPayload().getAddress();
-        log.trace("read: {}", msg.getPayload().getAddress());
+        boolean cacheable = msg.getPayload().isCacheReadResult();
+        log.trace("read: {}, cacheable: {}", msg.getPayload().getAddress(), cacheable);
+
         ReadResponse rr = new ReadResponse();
         try {
-            ILogData e = dataCache.get(address);
-            if (e == null) {
+            ILogData logData = dataCache.get(address, cacheable);
+            if (logData == null) {
                 rr.put(address, LogData.getEmpty(address));
             } else {
-                rr.put(address, (LogData) e);
+                rr.put(address, (LogData) logData);
             }
             r.sendResponse(ctx, msg, CorfuMsgType.READ_RESPONSE.payloadMsg(rr));
         } catch (DataCorruptionException e) {
@@ -315,17 +306,18 @@ public class LogUnitServer extends AbstractServer {
     }
 
     @ServerHandler(type = CorfuMsgType.MULTIPLE_READ_REQUEST)
-    private void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.trace("multiRead: {}", msg.getPayload().getAddresses());
+    public void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
+        boolean cacheable = msg.getPayload().isCacheReadResult();
+        log.trace("multiRead: {}, cacheable: {}", msg.getPayload().getAddresses(), cacheable);
 
         ReadResponse rr = new ReadResponse();
         try {
-            for (Long l : msg.getPayload().getAddresses()) {
-                ILogData e = dataCache.get(l);
-                if (e == null) {
-                    rr.put(l, LogData.getEmpty(l));
+            for (Long address : msg.getPayload().getAddresses()) {
+                ILogData logData = dataCache.get(address, cacheable);
+                if (logData == null) {
+                    rr.put(address, LogData.getEmpty(address));
                 } else {
-                    rr.put(l, (LogData) e);
+                    rr.put(address, (LogData) logData);
                 }
             }
             r.sendResponse(ctx, msg, CorfuMsgType.READ_RESPONSE.payloadMsg(rr));
@@ -429,25 +421,7 @@ public class LogUnitServer extends AbstractServer {
         }
     }
 
-    /**
-     * Retrieve the LogUnitEntry from disk, given an address.
-     *
-     * @param address The address to retrieve the entry from.
-     * @return The log unit entry to retrieve into the cache.
-     * <p>
-     * This function should not care about trimmed addresses, as that is handled in
-     * the read() and append(). Any address that cannot be retrieved should be returned as
-     * unwritten (null).
-     */
-    private ILogData handleRetrieval(long address) {
-        LogData entry = streamLog.read(address);
-        log.trace("Retrieved[{} : {}]", address, entry);
-        return entry;
-    }
 
-    private void handleEviction(long address, ILogData entry, RemovalCause cause) {
-        log.trace("Eviction[{}]: {}", address, cause);
-    }
 
     /**
      * Shutdown the server.
@@ -460,7 +434,7 @@ public class LogUnitServer extends AbstractServer {
     }
 
     @VisibleForTesting
-    public LoadingCache<Long, ILogData> getDataCache() {
+    public LogUnitServerCache getDataCache() {
         return dataCache;
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
@@ -197,8 +197,8 @@ public class StateTransfer {
 
         long ts1 = System.currentTimeMillis();
 
-        Map<Long, ILogData> dataMap = runtime.getAddressSpaceView()
-                .fetchAll(chunk, true);
+        // Don't cache the read results on server for state transfer.
+        Map<Long, ILogData> dataMap = runtime.getAddressSpaceView().nonCacheFetchAll(chunk, true);
 
         long ts2 = System.currentTimeMillis();
 
@@ -215,8 +215,7 @@ public class StateTransfer {
         }
 
         try {
-
-            // Write segment chunk to the new logunit
+            // Write segment chunk to the new log unit
             ts1 = System.currentTimeMillis();
             boolean transferSuccess = CFUtils.getUninterruptibly(runtime.getLayoutView()
                     .getRuntimeLayout(layout)
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
@@ -5,7 +5,6 @@ import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.Getter;
 
-import java.util.Collections;
 import java.util.List;
 
 /**
@@ -13,11 +12,15 @@ import java.util.List;
  *
  * Created by maithem on 7/28/17.
  */
-@Data
+@Getter
 @AllArgsConstructor
 public class MultipleReadRequest implements ICorfuPayload<MultipleReadRequest> {
-    @Getter
-    final List<Long> addresses;
+
+    // List of requested addresses to read.
+    private final List<Long> addresses;
+
+    // Whether the read results should be cached on server.
+    private final boolean cacheReadResult;
 
     /**
      * Deserialization Constructor from ByteBuf to ReadRequest.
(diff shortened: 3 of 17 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -286,10 +286,12 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
...
     private void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
         long address = msg.getPayload().getAddress();
-        log.trace("read: {}", msg.getPayload().getAddress());
+        boolean cacheable = msg.getPayload().isCacheable();
+        log.trace("read: {}, cacheable: {}", msg.getPayload().getAddress(), cacheable);
+
         ReadResponse rr = new ReadResponse();
         try {
-            ILogData e = dataCache.get(address);
+            ILogData e = cacheable ? dataCache.get(address) : handleRetrieval(address);
```
hisundar: Suppose the value to be read is already in the cache, then this will actually result in a slowdown because handleRetrieval will go to the fileChannel anyway right? Instead instead of directly going to handleRetrieval, can we do that only if dataCache.getIfPresent() returns null?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
```diff
@@ -16,24 +16,24 @@
 @Data
 @AllArgsConstructor
 public class MultipleReadRequest implements ICorfuPayload<MultipleReadRequest> {
-    @Getter
+
     final List<Long> addresses;
 
+    final boolean cacheable;
```
zalokhan: Can you rename it to cacheableOnServer or something which specifies the server.
It can be confused with the client side cache.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -29,10 +15,19 @@
 import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
 import org.corfudb.runtime.view.Address;
-import org.corfudb.runtime.view.RuntimeLayout;
-import org.corfudb.runtime.view.replication.ChainReplicationProtocol;
 import org.corfudb.util.Utils;
 
+import javax.annotation.Nonnull;
```
zalokhan: Lexicographically order imports.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
```diff
@@ -12,20 +12,26 @@
 @AllArgsConstructor
 public class ReadRequest implements ICorfuPayload<ReadRequest> {
 
+    // Requested address to read.
     final long address;
 
+    // Whether the read result should be cached on server.
+    final boolean cacheable;
```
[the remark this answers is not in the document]
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
WenbinZhu: This is Intellij's auto import optimization. As you can see other files are also placing `import javax.*`  before `import java.*`

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -286,14 +286,16 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
     @ServerHandler(type = CorfuMsgType.READ_REQUEST)
     private void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
         long address = msg.getPayload().getAddress();
-        log.trace("read: {}", msg.getPayload().getAddress());
+        boolean cacheable = msg.getPayload().isCacheOnServer();
```
annym: isCacheOnServer, suggests if there is a cache on the server. Instead I believe that you want to refer to the fact that if the read should be cachedOnServer? am I right? if so can we properly rename.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -127,31 +127,56 @@ public Integer getPort() {
...
+
     /**
      * Read data from the log unit server for a list of addresses.
+     * Read results are <b>NOT</b> cached at log unit server.
      *
      * @param list list of global addresses.
      * @return a completableFuture which returns a ReadResponse on completion.
      */
     public CompletableFuture<ReadResponse> readAll(List<Long> list) {
+       return readAll(list, false);
```
annym: can we rename list to addresses or something more verbose.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -127,31 +127,56 @@ public Integer getPort() {
...
+    }
+
+    /**
+     * Read data from the log unit server for a list of addresses.
+     *
+     * @param list list of global addresses.
+     * @param cacheable Whether the read results should be cached on log unit server.
+     * @return a completableFuture which returns a ReadResponse on completion.
+     */
+    public CompletableFuture<ReadResponse> readAll(List<Long> list, boolean cacheable) {
```
annym: same here. list > addresses

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -321,6 +324,15 @@ private void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerC
         }
     }
 
+    private ILogData readSingle(long address, boolean cacheable) {
+        if (!cacheable) {
+            ILogData ld = dataCache.getIfPresent(address);
+            return ld != null ? ld : handleRetrieval(address);
```
annym: Could we add a test to validate, that reads are not cached and read again?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: Well, the Lombok generated Getter() will add a "is" prefix, which makes the name "cacheOnServer" confusing. I have renamed it to "cacheableOnSever", which makes more sense.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: Actually I think it would not  be easy to test this part. This needs to separate the cache, and unit test on it.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -286,14 +286,16 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
     @ServerHandler(type = CorfuMsgType.READ_REQUEST)
     private void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
         long address = msg.getPayload().getAddress();
-        log.trace("read: {}", msg.getPayload().getAddress());
+        boolean cacheable = msg.getPayload().isCacheableOnServer();
```
Maithem: Why not use `CacheOption` ?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: I had the same thought as @annym . 

I think we should wrap the Caffeine cache and only access the LogUnitCache. Also, you can have a method with like `cache.read(key, CacheOptions)`

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -86,8 +86,9 @@ public static void transfer(Layout layout,
 
             long ts1 = System.currentTimeMillis();
 
+            // Don't cache the read results on server for state transfer.
             Map<Long, ILogData> dataMap = runtime.getAddressSpaceView()
```
Maithem: Just pass `CacheOptions`

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: Actually my first version was using `CacehOption`, but I found this server option needs to be passed to client in the the message (in AddressSpaceView), and this creates confusion on client side.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: @annym @Maithem Done. Separated data cache and added a test.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
```diff
@@ -16,24 +14,26 @@
 @Data
 @AllArgsConstructor
 public class MultipleReadRequest implements ICorfuPayload<MultipleReadRequest> {
-    @Getter
+
+    // List of requested addresses to read.
     final List<Long> addresses;
 
+    // Whether the read results should be cached on server.
+    final boolean cacheableOnServer;
```
Maithem: Consider changing this veriable name to `cacheResponse ` or `cacheResult`

WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -284,16 +282,18 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
...
         long address = msg.getPayload().getAddress();
-        log.trace("read: {}", msg.getPayload().getAddress());
+        boolean cacheable = msg.getPayload().isCacheReadResult();
+        log.trace("read: {}, cacheable: {}", msg.getPayload().getAddress(), cacheable);
+
         ReadResponse rr = new ReadResponse();
         try {
-            ILogData e = dataCache.get(address);
-            if (e == null) {
+            ILogData ld = dataCache.get(address, cacheable);
```
xnull: can you please rename to logData

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -284,16 +282,18 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
...
+            if (ld == null) {
```
xnull: you can hide this `if else` block inside dataCache.get(key, putFunction)`

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
```diff
@@ -16,24 +14,26 @@
 @Data
 @AllArgsConstructor
 public class MultipleReadRequest implements ICorfuPayload<MultipleReadRequest> {
-    @Getter
+
+    // List of requested addresses to read.
```
xnull: please add `private`

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
```diff
@@ -16,24 +14,26 @@
...
     final List<Long> addresses;
 
+    // Whether the read results should be cached on server.
+    final boolean cacheReadResult;
```
xnull: please add `private`

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
```diff
@@ -16,24 +14,26 @@
 @Data
```
xnull: Would be also greate to use `@Getter` instead of `@Data`

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
```diff
@@ -12,20 +12,26 @@
 @AllArgsConstructor
 public class ReadRequest implements ICorfuPayload<ReadRequest> {
 
+    // Requested address to read.
     final long address;
 
+    // Whether the read result should be cached on server.
+    final boolean cacheReadResult;
```
xnull: please add `private`

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
```diff
@@ -12,20 +12,26 @@
 @AllArgsConstructor
 public class ReadRequest implements ICorfuPayload<ReadRequest> {
 
+    // Requested address to read.
```
xnull: please add `private`

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -743,8 +742,10 @@ private void applyForEachAddress(BiConsumer<Long, ILogData> logDataProcessor) {
                 final long lower = nextRead;
                 final long upper = Math.min(lower + batchReadSize - 1, logTail);
                 nextRead = upper + 1;
+
+                // Don't cache the read results on server for fast loader
                 Map<Long, ILogData> range =
-                        runtime.getAddressSpaceView().fetchAll(ContiguousSet.create(
+                        runtime.getAddressSpaceView().nonCacheFetchAll(ContiguousSet.create(
```
xnull: please reformat this piece of code, it is pretty inconvenient to read
```
                ContiguousSet<Long> addresses = ContiguousSet.create(
                        Range.closed(lower, upper),
                        DiscreteDomain.longs()
                );
                Map<Long, ILogData> range = runtime.getAddressSpaceView().fetchAll(addresses, true);
```

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -557,6 +558,21 @@ public void invalidateClientCache() {
...
+    /**
+     * Fetch a collection of addresses for insertion into the cache.
+     * The fetch results are <b>NOT</b> cached on log unit server.
+     * The result map returned is ordered by address.
+     *
+     * @param addresses    collection of addresses to read from.
+     * @param waitForWrite flag whether wait for write is required or hole fill directly.
+     * @return an ordered map of read addresses.
+     */
+    @Nonnull
```
xnull: looks like you don't need this annotation here

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -557,6 +558,21 @@ public void invalidateClientCache() {
...
+    public Map<Long, ILogData> nonCacheFetchAll(Iterable<Long> addresses,
```
xnull: Why Iterable? Why not sorted set or a list?

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.corfudb.infrastructure.LogUnitServerAssertions.assertThat;
+import static org.corfudb.infrastructure.LogUnitServerTest.waitForLogUnit;
+
+/**
+ * Created by WenbinZhu on 5/30/19.
+ */
+public class LogUnitCacheTest extends AbstractServerTest {
+
+    private static final double minHeapRatio = 0.1;
```
xnull: MIN_HEAP_RATIO

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
```diff
@@ -0,0 +1,102 @@
...
+    private static final double maxHeapRatio = 0.9;
```
xnull: MAX_HEAP_RATIO

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
```diff
@@ -0,0 +1,102 @@
...
+                .build());
+    }
+
+    /**
+     * Test non-cacheable reads on log unit sever will not affect server cache.
+     */
+    @Test
+    public void checkNonCacheableReads() throws Exception {
+        final int size = 10;
+        final long start = 0L, end = start + size;
```
xnull:
```
long start = 0L;
long end = start + size;
```

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MultipleReadRequest.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: This if else logic that returns LogData.getEmpty() is part of the log unit server's logic, not the cache's responsibility and capability, so I think it's better to put outside the cache.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadRequest.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
WenbinZhu: 1) keep consistent with other fetch api in this class 2) the caller of this is not only using a set or list.

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
WenbinZhu: Done.

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
WenbinZhu: Done.

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
WenbinZhu: Done.

PR #1904, opened 2019-05-21
Corfu as a platform platform interfaces.

xnull:
## Overview

Description:
Corfu platform core API

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/platform/core/src/main/java/org/corfudb/platform/core/Disposable.java
+++ b/platform/core/src/main/java/org/corfudb/platform/core/Disposable.java
@@ -1,0 +1,30 @@
+package org.corfudb.platform.core;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Denote an entity or resource that may be cancelled or disposed via {@link #dispose()}, which
+ * should be an idempotent operation.
+ *
+ * @author jameschang
+ * @since 2018-07-25
+ */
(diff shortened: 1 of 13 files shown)


platform/core/src/main/java/org/corfudb/platform/core/Disposable.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @author jameschang
+ * @since 2018-07-25
+ */
+@FunctionalInterface
+public interface Disposable extends AutoCloseable {
+
+    @Override
+    default void close() {
+        dispose().join();
```
lixinchengdu: What if the return value of dispose() is false?

platform/core/src/main/java/org/corfudb/platform/core/Result.java
```diff
@@ -0,0 +1,113 @@
...
+        return value;
+    }
+
+    /**
+     * Returns the encapsulated error if the {@link Result} instance is an error.
+     *
+     * @return the encapsulated {@link #error}.
+     * @throws NoSuchElementException if there is no error present.
+     */
+    public E getError() {
```
lixinchengdu: It would be better if there is another function to test whether there is any error by return true/false.

PR #1906, opened 2019-05-22
Huge performance impact. Avoid reading entire segment while building address space index

xnull:
## Overview

Description:
When corfu starts it goes through all the segments and builds the address space index for each segment. To build the index it loads all entire database (all the data from disk into memory). 
Also when any segment needs to be loaded we build the index it also loads the entire segment into memory just to build the index. 

Why should this be merged: 
Building index - loads all the data, it dramatically affects performance. 
Adding globalAddress into the index will reduce the data load by 95% when building the index. 

- [x] Public API has Javadoc

### DIFF
--- a/format/proto/types.proto
+++ b/format/proto/types.proto
@@ -52,6 +52,7 @@ message Metadata {
     required sfixed32 payload_checksum = 1;
     required sfixed32 length_checksum = 2;
     required sfixed32 length = 3;
+    optional sfixed64 global_address = 4;
 }
 
 message TrimEntry {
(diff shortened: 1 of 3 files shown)


format/proto/types.proto
```diff
@@ -52,6 +52,7 @@ message Metadata {
     required sfixed32 payload_checksum = 1;
     required sfixed32 length_checksum = 2;
     required sfixed32 length = 3;
+    optional sfixed64 global_address = 4;
```
Maithem: Good change.

The sequence number needs to be checksum'd too.

medhavidhawan:
If we are doing this we should also explore other serialization formats. We might not need to deserialize to extract data check out flatbuffers if that helps in this.

Maithem:
Can you address the comments?

xnull:
Need to check that we don't break checksum if we add a new field,
so in progress

Maithem:
@xnull in either case, data migration is required, so the checksums need to be recomputed anyways. 
Please, also try to pull out the `LogUnitMetadataType` portion too, other wise this performance enhancement won't be useful for logunit restarts (i.e. since the global address alone is not sufficient to build the stream tails).

PR #1908, opened 2019-05-23
Hex to table name fix in vlo

PavelZaytsev:
## Overview

Description:
When VersionLockedObject's functions throw exceptions, it is not clear looking at the logs during which table the exception is seen. Original stream name should be preserved in that case. 

Why should this be merged: 
It is easier to link the VLO failures to a particular table.

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -40,7 +40,7 @@ import java.util.function.Supplier;
  * are backed by a stream.
  * <p>
  * <p>This class implements the methods that an in-memory corfu-object proxy carries
- * in order to by in sync with a stream.
+ * in order to be in sync with a stream.
  * <p>
  * <p>We refer to the program's object as the -corfu object-,
  * and to the internal object implementation as the -proxy-.
@@ -75,11 +75,15 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
     CorfuRuntime rt;
 
     /**
-     * The ID of the stream of the log.
+     * The StreamId of the stream of the log.
      */
-    @Deprecated // TODO: Add replacement method that conforms to style
-    @SuppressWarnings("checkstyle:abbreviation") // Due to deprecation
-            UUID streamID;
+    private final StreamId streamId;
+
+    /**
+     * The name of the stream of the log.
+     */
+    @Getter
+    String streamName;
 
     /**
      * The type of the underlying object. We use this to instantiate
@@ -122,7 +126,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
      * Creates a CorfuCompileProxy object on a particular stream.
      *
      * @param rt                  Connected CorfuRuntime instance.
-     * @param streamID            StreamID of the log.
+     * @param streamId            StreamID of the log.
      * @param type                Type of underlying object to instantiate a new instance.
      * @param args                Arguments to create this proxy.
      * @param serializer          Serializer used by the SMR entries to serialize the arguments.
@@ -133,7 +137,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
      */
     @Deprecated // TODO: Add replacement method that conforms to style
     @SuppressWarnings("checkstyle:abbreviation") // Due to deprecation
-    public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[] args,
+    public CorfuCompileProxy(CorfuRuntime rt, StreamId streamId, Class<T> type, Object[] args,
                              ISerializer serializer,
                              Map<String, ICorfuSMRUpcallTarget<T>> upcallTargetMap,
                              Map<String, IUndoFunction<T>> undoTargetMap,
@@ -141,7 +145,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
                              Set<String> resetSet
     ) {
         this.rt = rt;
-        this.streamID = streamID;
+        this.streamId = streamId;
         this.type = type;
         this.args = args;
         this.serializer = serializer;
@@ -149,7 +153,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
         // Since the VLO is thread safe we don't need to use a thread safe stream implementation
         // because the VLO will control access to the stream
         underlyingObject = new VersionLockedObject<T>(this::getNewInstance,
-                new StreamViewSMRAdapter(rt, rt.getStreamsView().getUnsafe(streamID)),
+                new StreamViewSMRAdapter(rt, rt.getStreamsView().getUnsafe(streamId.getId()), streamId),
                 upcallTargetMap, undoRecordTargetMap,
                 undoTargetMap, resetSet);
 
@@ -193,7 +197,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
         for (int x = 0; x < rt.getParameters().getTrimRetry(); x++) {
             // Linearize this read against a timestamp
             final long timestamp = rt.getSequencerView()
-                            .query(getStreamID()).getToken().getSequence();
+                    .query(getStreamID()).getToken().getSequence();
             log.debug("Access[{}] conflictObj={} version={}", this, conflictObject, timestamp);
 
             try {
@@ -343,7 +347,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
      */
     @Override
     public UUID getStreamID() {
-        return streamID;
+        return streamId.getId();
     }
 
     /**
@@ -477,7 +481,7 @@ public class CorfuCompileProxy<T> implements ICorfuSMRProxyInternal<T> {
 
     @Override
     public String toString() {
-        return type.getSimpleName() + "[" + Utils.toReadableId(streamID) + "]";
+        return type.getSimpleName() + "[" + Utils.toReadableId(streamId.getId()) + "]";
     }
 
     private void abortTransaction(Exception e) {
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileWrapperBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileWrapperBuilder.java
@@ -18,7 +18,7 @@ public class CorfuCompileWrapperBuilder {
      *
      * @param type       Type of SMR object.
      * @param rt         Connected instance of the CorfuRuntime.
-     * @param streamID   StreamID of the SMR Object.
+     * @param streamId   StreamID of the SMR Object.
      * @param args       Arguments passed to instantiate the object.
      * @param serializer Serializer to be used to serialize the object arguments.
      * @param <T>        Type
@@ -27,10 +27,9 @@ public class CorfuCompileWrapperBuilder {
      * @throws IllegalAccessException Illegal Access to the Object.
      * @throws InstantiationException Cannot instantiate the object using the arguments and class.
      */
-    @Deprecated // TODO: Add replacement method that conforms to style
-    @SuppressWarnings("checkstyle:abbreviation") // Due to deprecation
+    @SuppressWarnings("checkstyle:abbreviation")
     public static <T> T getWrapper(Class<T> type, CorfuRuntime rt,
-                                   UUID streamID, Object[] args,
+                                   StreamId streamId, Object[] args,
                                    ISerializer serializer)
             throws ClassNotFoundException, IllegalAccessException,
             InstantiationException {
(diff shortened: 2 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -81,6 +81,12 @@
     @SuppressWarnings("checkstyle:abbreviation") // Due to deprecation
             UUID streamID;
 
+    /**
+     * The name of the stream of the log.
+     */
+    @Getter
+    String streamName;
```
xnull: private final

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -133,7 +139,7 @@
      */
     @Deprecated // TODO: Add replacement method that conforms to style
     @SuppressWarnings("checkstyle:abbreviation") // Due to deprecation
-    public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[] args,
+    public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, String streamName, Class<T> type, Object[] args,
```
xnull: what the difference between `streamID` and `streamName` ?
Isn't it a string representation of UUID by the chance?

xnull: Ok, `streamName` not equals to `streamID`

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -81,7 +81,7 @@
     /**
      * The stream view this object is backed by.
      */
-    private final ISMRStream smrStream;
+    private final StreamViewSMRAdapter smrStream;
```
lixinchengdu: Why change from interface to concrete implementation?

PavelZaytsev: Because I am using StreamViewSMRAdapter's streamName getter to get the underlying stream.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
Maithem: I don't think this is an appropriate place to store the name.
The stream name, which is also the object's name can be exposed through `IStreamView`.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
Maithem: Its cleaner to pass down the stream name and then derive the stream id. Passing down redundant pieces of information that can be derived is error-prone.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -142,14 +148,15 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
...
+        this.streamName = streamName;
         this.type = type;
         this.args = args;
         this.serializer = serializer;
 
         // Since the VLO is thread safe we don't need to use a thread safe stream implementation
         // because the VLO will control access to the stream
         underlyingObject = new VersionLockedObject<T>(this::getNewInstance,
-                new StreamViewSMRAdapter(rt, rt.getStreamsView().getUnsafe(streamID)),
+                new StreamViewSMRAdapter(rt, rt.getStreamsView().getUnsafe(streamID), streamName),
```
Maithem: Why don't you store the stream reference (i.e. `rt.getStreamsView().getUnsafe(streamID)`) and then just use `stream.getName()`

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileWrapperBuilder.java
```diff
@@ -30,7 +30,7 @@
     @Deprecated // TODO: Add replacement method that conforms to style
     @SuppressWarnings("checkstyle:abbreviation") // Due to deprecation
     public static <T> T getWrapper(Class<T> type, CorfuRuntime rt,
-                                   UUID streamID, Object[] args,
+                                   UUID streamID, String streamName, Object[] args,
```
Maithem: I think its fine to drop the UUID and just pass the stream name.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -44,10 +45,18 @@
      */
     final CorfuRuntime runtime;
 
+    /**
+     * Name of the wrapped stream.
+     */
+    @Getter
+    private final String streamName;
```
Maithem: This is not needed. Since you have access to the underlaying stream (i.e. passed through the constructor), then you can get the name via `streamView.getName()`

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
Maithem: This is not needed if you expose `IStream::getName`.

I think we should just use the `ISMRStream` interface.

runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
```diff
@@ -101,13 +101,13 @@ public T open() {
         try {
```
Maithem: If both the name and uuid are specified in the builder, then we need to check that the uuid can be derived from the name.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
PavelZaytsev: I can't change IStreamView to expose a name because it is used in many stream views, whose methods are used by Layout  and Layout is used heavily by NSX clients

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
PavelZaytsev: See above

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
PavelZaytsev: Sounds good

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileWrapperBuilder.java
PavelZaytsev: Sounds good

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
PavelZaytsev: See above

runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
PavelZaytsev: Sounds good

PavelZaytsev:
@Maithem 
I applied some of the changes, plz take a look

PavelZaytsev:
Actually looking at the recovery and fast object loader logic, it seems like we should support either of the name or the udid.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -133,23 +137,23 @@
      */
     @Deprecated // TODO: Add replacement method that conforms to style
     @SuppressWarnings("checkstyle:abbreviation") // Due to deprecation
-    public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[] args,
+    public CorfuCompileProxy(CorfuRuntime rt, StreamId streamId, Class<T> type, Object[] args,
```
xnull: maybe add `@NonNull` StreamId streamId?

runtime/src/main/java/org/corfudb/runtime/object/StreamId.java
```diff
@@ -0,0 +1,35 @@
...
+
+import javax.annotation.Nullable;
+import java.util.UUID;
+
+@EqualsAndHashCode
+@Getter
+@ToString
+public class StreamId {
+    @Nullable
+    private String name;
```
xnull: let's transform it to Optional, as you suggested earlier

PavelZaytsev: I'll just go ahead and ignore the warking then

runtime/src/main/java/org/corfudb/runtime/object/StreamId.java
```diff
@@ -0,0 +1,56 @@
...
+    public static StreamId build(Optional<String> streamName, Optional<UUID> uuid) {
+        StreamId streamId;
+
+        if (streamName.isPresent() && uuid.isPresent()) {
+            streamId = new StreamId(uuid.get(), streamName);
+        } else if (streamName.isPresent()) {
+            UUID id = CorfuRuntime.getStreamID(streamName.get());
+            streamId = new StreamId(id, streamName);
+        } else if (uuid.isPresent()) {
+            streamId = new StreamId(uuid.get(), streamName);
```
hisundar: If streamName is not present, why pass it in? Do we want a null stream name displayed?

PavelZaytsev: We display a readable udid if streamName is not present.

hisundar: Got it. Thanks

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
Maithem: Not sure I follow. A new method is added to the interface, how could it break other code if its not consumed anywhere?

runtime/src/main/java/org/corfudb/runtime/object/StreamId.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Either of the variables should be defined to create a streamId.
+     *
+     * @param streamName name of a stream
+     * @param uuid       id of a stream
+     * @return streamId object
+     * @throws IllegalStateException exception if none are defined
+     */
+    public static StreamId build(Optional<String> streamName, Optional<UUID> uuid) {
+        StreamId streamId;
```
Maithem: You can use a builder here.
Also, I think that overloading is more appropriate than passing optional here, since a stream can be identified in two ways. Either by name, or by UUID.

runtime/src/main/java/org/corfudb/runtime/object/StreamId.java
```diff
@@ -0,0 +1,56 @@
...
+
+        if (streamName.isPresent() && uuid.isPresent()) {
```
Maithem: An `IllegalArgumentException` should be thrown if the UUID cannot be derived from the streamName

runtime/src/main/java/org/corfudb/runtime/object/StreamId.java
```diff
@@ -0,0 +1,56 @@
...
+        } else if (uuid.isPresent()) {
+            streamId = new StreamId(uuid.get(), streamName);
+        } else {
+            throw new IllegalArgumentException("Either uuid or streamName should be defined");
+        }
+
+        return streamId;
+    }
+
+    public String getName() {
```
Maithem: Why not just overload `toString` and include both pieces of information ?

Maithem:
@PavelZaytsev @WenbinZhu The table name can be passed through the `StreamOptions` object, its much cleaner.

annym:


```diff

Issues
======
+ Solved 1
- Added 1
           

Complexity increasing per file
==============================
- runtime/src/main/java/org/corfudb/runtime/object/StreamId.java  6
- runtime/src/test/java/org/corfudb/runtime/object/StreamIdTest.java  1
         

Complexity decreasing per file
==============================
+ runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/CorfuDB/CorfuDB/pullRequest?prid=4653447&bid=15584338)

runtime/src/test/java/org/corfudb/runtime/object/StreamIdTest.java
```diff
@@ -0,0 +1,37 @@
+package org.corfudb.runtime.object;
+
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.util.Utils;
+import org.junit.Test;
+
+import java.util.Optional;
+import java.util.UUID;
+
+import static org.junit.Assert.*;
```
annym: Codacy Issue found: [Avoid unused imports such as 'org.junit.Assert'](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651475)

annym:


```diff

Issues
======
+ Solved 1
- Added 1
           

Complexity increasing per file
==============================
- runtime/src/test/java/org/corfudb/runtime/object/StreamIdTest.java  1
- runtime/src/main/java/org/corfudb/runtime/object/StreamId.java  6
         

Complexity decreasing per file
==============================
+ runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651475&bid=15577788)

PR #1909, opened 2019-05-23
SnowflakeGuidFactory: Generates unique Ids with global ordering

hisundar:
+ OrderedGUIDGenerator
|-> SnowflakeGuidGenerator which does a best effort generation of
globally unique and ordered ids.

## Overview

Description:

Why should this be merged: This is a prerequisite for building collections like CorfuQueue.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -34,6 +34,8 @@ import org.corfudb.runtime.view.LayoutManagementView;
 import org.corfudb.runtime.view.LayoutView;
 import org.corfudb.runtime.view.ManagementView;
 import org.corfudb.runtime.view.ObjectsView;
+import org.corfudb.runtime.view.OrderedGuidGenerator;
+import org.corfudb.runtime.view.SnowflakeGuidGenerator;
 import org.corfudb.runtime.view.SequencerView;
 import org.corfudb.runtime.view.StreamsView;
 import org.corfudb.util.CFUtils;
@@ -534,6 +536,9 @@ public class CorfuRuntime {
     public volatile UUID clusterId;
 
     @Getter
+    private volatile OrderedGuidGenerator guidGenerator;
+
+    @Getter
     final ViewsGarbageCollector garbageCollector = new ViewsGarbageCollector(this);
 
     /**
@@ -658,6 +663,9 @@ public class CorfuRuntime {
 
         // Set the initial cluster Id
         clusterId = parameters.getClusterId();
+
+        // Initialize the Guid Generator
+        guidGenerator = new SnowflakeGuidGenerator(System.identityHashCode(this));
 
         // Generate or set the NettyEventLoop
         nettyEventLoop = parameters.nettyEventLoop == null ? getNewEventLoopGroup()
--- a/test/src/test/java/org/corfudb/runtime/view/SnowflakeGuidGeneratorTest.java
+++ b/test/src/test/java/org/corfudb/runtime/view/SnowflakeGuidGeneratorTest.java
@@ -1,0 +1,47 @@
+package org.corfudb.runtime.view;
+
+import org.corfudb.runtime.CorfuRuntime;
+import org.junit.Test;
+
+import java.util.HashSet;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Created by Sundar Sridharan on May 23, 2019.
+ */
+public class SnowflakeGuidGeneratorTest extends AbstractViewTest {
+
+    @Test
+    public void areUniqueAndOrderedLong() {
+        final int iterations = 1000;
+        CorfuRuntime r = getDefaultRuntime();
+        OrderedGuidGenerator guidGenerator = r.getGuidGenerator();
+        long lastValue = 0;
+        HashSet<Long> uniq = new HashSet<>(iterations);
+        for (int i = 0; i < iterations; i++) {
+            long current = guidGenerator.nextLong();
+            assertThat(current).isGreaterThan(lastValue);
+            lastValue = current;
+            assertThat(uniq).doesNotContain(current);
+            uniq.add(current);
+        }
+    }
+
+    @Test
+    public void areUniqueAndOrderedUUID() {
+        final int iterations = 1000;
+        CorfuRuntime r = getDefaultRuntime();
+        OrderedGuidGenerator guidGenerator = r.getGuidGenerator();
+        UUID lastValue = new UUID(0,0);
+        HashSet<UUID> uniq = new HashSet<>(iterations);
+        for (int i = 0; i < iterations; i++) {
+            UUID current = guidGenerator.nextUUID();
+            assertThat(current).isGreaterThan(lastValue);
+            lastValue = current;
+            assertThat(uniq).doesNotContain(current);
+            uniq.add(current);
+        }
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
@@ -14,7 +14,6 @@ import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.runtime.CorfuRuntime;
-import org.corfudb.runtime.collections.CorfuTable;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
 import org.corfudb.runtime.object.CorfuCompileProxy;
 import org.corfudb.runtime.object.CorfuCompileWrapperBuilder;
@@ -79,11 +78,6 @@ public class ObjectBuilder<T> implements IObjectBuilder<T> {
     }
 
     public ObjectBuilder<T> setArguments(Object... arguments) {
-        this.arguments = arguments;
-        return this;
-    }
-
-    public ObjectBuilder<T> setArgumentsArray(Object[] arguments) {
         this.arguments = arguments;
         return this;
     }
--- a/runtime/src/main/java/org/corfudb/runtime/view/OrderedGuidGenerator.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/OrderedGuidGenerator.java
@@ -1,0 +1,21 @@
+package org.corfudb.runtime.view;
+
+import java.util.UUID;
+
+/**
+ * Globally Unique Identity (GUID) generator that returns GUIDs having a notion of
+ * a comparable global ordering.
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+public interface OrderedGuidGenerator {
+    /**
+     * @return a compact but low resolution ordered guid
+     */
+    long nextLong();
+
+    /**
+     * @return a high resolution ordered guid
+     */
+    UUID nextUUID();
+}
--- a/runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidGenerator.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidGenerator.java
@@ -1,0 +1,66 @@
+package org.corfudb.runtime.view;
+
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Fast Globally Unique Identity (GUID) generator that returns compact ids
+ * having a notion of a universal comparable ordering.
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+public class SnowflakeGuidGenerator implements OrderedGuidGenerator {
+    private static final short NODE_MAX_VAL = 1023;
+    private static final int NODE_SHIFT_LONG = 10;
+    private static final int NODE_SHIFT_UUID = 53;
+    private static final short SEQ_MAX_VAL = 4095;
+    private static final int SEQ_SHIFT = 12;
+
+    private final short node;
+    private final AtomicInteger sequence;
+
+    /**
+     * Please pass something unique to this node.
+     * For example:
+     *      UUID nodeId = UUID.fromString("mac_address_of_machine");
+     *      SnowflakeGuidGenerator guidGenerator = new SnowflakeGuidGenerator(nodeId.node());
+     * @param node - Simple integer representing the current node.
+     */
+    public SnowflakeGuidGenerator(long node) {
+        this.node = (short)(node & NODE_MAX_VAL);
+        sequence = new AtomicInteger(0);
+    }
+
+    /**
+     * Fast best effort generation scheme for compact snowflake generation.
+     *    ______________________________________________
+     *   | timestamp lower 42 bits | Node Id | Sequence |
+     *    ------------------------- --------- ----------
+     *    <------ 42 bits ---------><-10bits-><-12bits-->
+     * @return The next 64-bit integer.
+     */
+    @Override
+    public long nextLong() {
+        long currentTime = System.currentTimeMillis();
+        long counter = this.sequence.incrementAndGet() & SEQ_MAX_VAL;
+        return currentTime << NODE_SHIFT_LONG << SEQ_SHIFT
+                | node << SEQ_SHIFT
+                | counter;
+    }
+
+    /**
+     * Fast best effort generation scheme for snowflake generation.
+     *    ____________________________________________________
+     *   | timestamp 8 bytes      | Node Id  | Sequence       |
+     *    ------------------------- --------------------------
+     *    <------ 64 bits --------><-10bits--><----54 bits---->
+     * @return a higher resolution unique identifier.
+     */
+    @Override
+    public UUID nextUUID() {
+        long currentTime = System.currentTimeMillis();
+        long counter = this.sequence.incrementAndGet();
+        counter |= node << NODE_SHIFT_UUID;
+        return new UUID(currentTime, counter);
+    }
+}


runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
```diff
@@ -0,0 +1,18 @@
...
+ * Why not use Java UUIDs?
+ *   JAVA UUIDs lack cross node comparability or global ordering in the absence of time sync.
+ *   They are also 36 bytes long which can be an overkill for simple use cases.
+ *
+ * What about schemes like Twitter Snowflake?
+ *    Requires time synchronization between cluster nodes for ordering.
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+public interface AbstractOrderedGUIDFactory {
```
xnull: Don't use a sequence of more than two capital letters. Please rename to AbstractOrderedGuidFactory

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
```diff
@@ -0,0 +1,18 @@
...
+    long nextGUID();
```
xnull: you can rename to`next`

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
WenbinZhu: The name Abstract* should be used for abstract class, right? But this is an interface.

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+public class SnowflakeGuidFactory implements AbstractOrderedGUIDFactory {
+    public static final short NODE_MAX_VAL = 1023;
+    public static final int NODE_BIT_POS = 10;
+    public static final short SEQ_MAX_VAL = 4095;
+    public static final int SEQ_MSB = 12;
+
+    private short node;
```
xnull: private final

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    private AtomicInteger sequence;
```
xnull: private final

test/src/test/java/org/corfudb/runtime/view/SnowflakeGuidFactoryTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Created by Sundar Sridharan on May 23, 2019.
+ */
+public class SnowflakeGuidFactoryTest extends AbstractViewTest {
+
+    @Getter
+    final String defaultConfigurationString = getDefaultEndpoint();
```
xnull: private

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Use this node id to scope the generator to a namespace.
+     * For example:
+     *      UUID nodeId = UUID.fromString("mac_address_of_machine");
+     *      SnowflakeGuidFactory guidFactory = new SnowflakeGuidFactory(nodeId.node());
+     * @param node
+     */
+    public SnowflakeGuidFactory(long node) {
```
WenbinZhu: Should this class be a singleton class and `nexGUID()` be a static method? Also if the node() is a machine specific mac address, why not let this class generate it instead of pass it here?

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
hisundar: Done.

hisundar: Any suggestions then?
(I wanted to avoid use of the I prefix even though it is used widely in the code today.)

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
hisundar: done

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: done.

test/src/test/java/org/corfudb/runtime/view/SnowflakeGuidFactoryTest.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: I did not know of an easy way to generate it. But I think you are right, let me attempt your suggestion instead.

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
```diff
@@ -0,0 +1,18 @@
+package org.corfudb.runtime.view;
+
+/**
+ * Globally Unique Identity (GUID) generator that returns compact GUIDs having a notion of
+ * a comparable global ordering.
+ *
+ * Why not use Java UUIDs?
+ *   JAVA UUIDs lack cross node comparability or global ordering in the absence of time sync.
+ *   They are also 36 bytes long which can be an overkill for simple use cases.
```
Maithem: hmm, are they really?

UUID constructor uses two longs. In java a long is only 8 bytes. 

```java
public UUID(long mostSigBits, long leastSigBits) {
        this.mostSigBits = mostSigBits;
        this.leastSigBits = leastSigBits;
    }
```

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
```diff
@@ -0,0 +1,49 @@
...
+ * having a notion of a universal comparable ordering.
+ *
+ *    ______________________________________________
+ *   | timestamp lower 42 bits | Node Id | Sequence |
+ *    ------------------------- --------- ----------
+ *    <------ 42 bits ---------><-10bits-><-12bits-->
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+public class SnowflakeGuidFactory implements AbstractOrderedGUIDFactory {
```
Maithem: Should we really keep this name?

Its a little bit confusing, the java doc in `AbstractOrderedGUIDFactory.java` gives a reason of why we're not implementing "snowflake-style" ids, but then this is called a Snowflake factory.

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
hisundar: Good point. Only the string representation of the UUID is 36 bytes. They do lack global ordering though. I will update the comment. Thanks!

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Fast Globally Unique Identity (GUID) generator that returns compact ids
+ * having a notion of a universal comparable ordering.
+ *
+ *    ______________________________________________
+ *   | timestamp lower 42 bits | Node Id | Sequence |
+ *    ------------------------- --------- ----------
+ *    <------ 42 bits ---------><-10bits-><-12bits-->
+ *
```
Maithem: What's the difference between a timestamp and a sequence?

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
xnull: I don't agree, never use static, always try to avoid using it

xnull: Or if you really want to use static method in this case then rename SnowflakeGuidUtil and don't se interface

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
WenbinZhu: Maybe just remove Abstract to be OrderedGuidFactory?

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: Just the resolution. Timestamp is coarse grained and the sequence helps distinguish several consecutive ids within that same time quantum.

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
xnull: Just use `OrderedGuidFactory`
also, I don't see reasons for using interfaces for Factory classes

why you need an interface here? Maybe it's better to get rid of it and just use Util class with a static method

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
```diff
@@ -0,0 +1,49 @@
...
+        this.node = (short)(node & NODE_MAX_VAL);
+        sequence = new AtomicInteger(0);
+    }
+
+    /**
+     * Fast best effort generation scheme for snowflakes generation.
+     * @return The next 64-bit integer.
+     */
+    @Override
+    public long nextGUID() {
```
epalta: Is this safe to be accessed by multiple threads?

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: Yes, sorry for the confusion, I was in the middle of creating 2 id generators - one with just the snowflake style generation scheme and the other with corfu's token and forgot to update the comment.

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: It should be. The only local variable accessed is the sequence and that is updated Atomically.

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: Sorry for the confusion, I started with the corfu's token based guid factory and forgot to update this comment.

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
hisundar: Actually I can't make this a singleton because it is meant to be scoped to a corfu runtime. If we ever support multiple runtimes in the same jvm then they should have their own copies of the guid generators. If not, then it will be a singleton anyway.

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
hisundar: Actually I cannot make it a singleton because it is meant to be scoped to a corfu runtime. If we ever support multiple runtime in the same process, then they should have their own namespaces. Otherwise it is already singleton.

runtime/src/main/java/org/corfudb/runtime/view/AbstractOrderedGUIDFactory.java
hisundar: Need for interface is to have multiple types of ordered guid factories - one that generates snowflake style IDs and other which uses corfu's token based approach.

runtime/src/main/java/org/corfudb/runtime/view/SnowflakeGuidFactory.java
WenbinZhu: > Actually I cannot make it a singleton because it is meant to be scoped to a corfu runtime. If we ever support multiple runtime in the same process, then they should have their own namespaces. Otherwise it is already singleton.

Then I think you should instantiate this  factory class when the runtime is created, instead of instantiating it every time you need to generate a GUID, right? Or you can make `nexGUID()` a static method, to make this class sth. like util class.

hisundar: Done. I have included the instantiation into the runtime and modified the tests accordingly.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -533,6 +526,9 @@
     @Getter
     public volatile UUID clusterId;
 
+    @Getter
+    public volatile OrderedGUIDGenerator guidGenerator;
```
xnull: you don't need public if you use `@Getter`

xnull: it can be private final

hisundar: good point. fixed. thanks!

runtime/src/main/java/org/corfudb/runtime/view/OrderedGUIDGenerator.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.UUID;
+
+/**
+ * Globally Unique Identity (GUID) generator that returns GUIDs having a notion of
+ * a comparable global ordering.
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+public interface OrderedGUIDGenerator {
```
xnull: Please rename to `OrderedGuidGenerator` in the other case checkstyle will show you a warning

runtime/src/main/java/org/corfudb/runtime/view/OrderedGUIDGenerator.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * @return a compact but low resolution ordered guid
+     */
+    long nextLong();
+
+    /**
+     * @return a high resolution ordered guid
+     */
+    UUID nextUUID();
```
xnull: `nextGuid` ?

test/src/test/java/org/corfudb/runtime/view/SnowflakeGuidGeneratorTest.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.HashSet;
+import java.util.Random;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Created by Sundar Sridharan on May 23, 2019.
+ */
+public class SnowflakeGuidGeneratorTest extends AbstractViewTest {
```
xnull: do you really need to extend AbstractViewTest?

xnull: oh,  yes, you need, ok

test/src/test/java/org/corfudb/runtime/view/SnowflakeGuidGeneratorTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Test
+    public void areUniqueAndOrderedLong() {
+        final int iterations = 1000;
+        CorfuRuntime r = getDefaultRuntime();
```
xnull: Actually you can create your own instance of CorfuRuntime. Then you don't need to extend AbstractViewTest

runtime/src/main/java/org/corfudb/runtime/view/OrderedGUIDGenerator.java
hisundar: Done. Sorry I missed that from the previous round!

runtime/src/main/java/org/corfudb/runtime/view/OrderedGUIDGenerator.java
hisundar: well, they are both Guids (globally unique identifiers) I used nextUUID() and nextLong() to denote the return type (similar to Java's Random)

xnull: ok, up to you

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -659,6 +655,9 @@ private CorfuRuntime(@Nonnull CorfuRuntimeParameters parameters) {
         // Set the initial cluster Id
         clusterId = parameters.getClusterId();
 
+        // Initialize the Guid Generator
+        guidGenerator = new SnowflakeGuidGenerator(this.hashCode());
```
xnull: maybe `System.identityHashCode` ?

hisundar: done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -28,14 +28,7 @@
...
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
-import org.corfudb.runtime.view.AddressSpaceView;
-import org.corfudb.runtime.view.Layout;
-import org.corfudb.runtime.view.LayoutManagementView;
-import org.corfudb.runtime.view.LayoutView;
-import org.corfudb.runtime.view.ManagementView;
-import org.corfudb.runtime.view.ObjectsView;
-import org.corfudb.runtime.view.SequencerView;
-import org.corfudb.runtime.view.StreamsView;
+import org.corfudb.runtime.view.*;
```
annym: Can you remove the wildcard?

hisundar: That is coming from the IDE not me. If the wildcard should not exist then I have to find a way to trick Intellij. But curious to know why do we want to remove the wildcard, the runtime anyway imports anything in the view right?

hisundar: Done. Got rid of the wildcard anyway.

xnull: https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

PR #1911, opened 2019-05-24
Add SMRRecordLocator

lixinchengdu:
## Overview

Description:

SMRRecordLocator is used to denote the location of a SMRRecord in the global SpaceAddressView.

Why should this be merged: 

SMRRecordLocator will be used by GC to denote the where the garbage is in the global SpaceAddressView. It could also be used by runtime to retrieve data from LogUnit.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecord.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecord.java
@@ -10,12 +10,15 @@ import lombok.Getter;
 import lombok.Setter;
 import lombok.NonNull;
 
+
+import java.util.Arrays;
+
+
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
 
-import java.util.Arrays;
 
 /**
  * Created by mwei on 1/8/16.
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecordLocator.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecordLocator.java
@@ -1,0 +1,59 @@
+package org.corfudb.protocols.logprotocol;
+
+import io.netty.buffer.ByteBuf;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.NonNull;
+
+import org.corfudb.util.serializer.ICorfuSerializable;
+
+import java.util.UUID;
+
+
+@ToString
+@AllArgsConstructor
+@NoArgsConstructor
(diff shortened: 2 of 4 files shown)


runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -128,4 +127,51 @@ public void setEntry(ILogData entry) {
...
+        @Override
+        public int compareTo(@NonNull ILocalSMREntryLocator other) {
+            if (other instanceof LocalMultiObjectSMREntryLocator) {
+                if (!streamId.equals(((LocalMultiObjectSMREntryLocator) other).getStreamId())) {
+                    throw new RuntimeException("Can't compare LocalMultiObjectSMREntryLocator of different streams");
+                }
+
+                return Integer.compare(index, ((LocalMultiObjectSMREntryLocator) other).getIndex());
+            } else {
+                throw new RuntimeException("Can't compare Local SMREntryLocators of different types");
```
lixinchengdu: IIlegalArgumentException

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiSMREntry.java
```diff
@@ -7,8 +7,7 @@
 import java.util.List;
 import java.util.UUID;
 
-import lombok.Getter;
-import lombok.ToString;
+import lombok.*;
```
lixinchengdu: do not use wildcard

annym:
@lixinchengdu why are there some many changes in this PR? Isn't this only aimed to introduce SMREntryLocator?

lixinchengdu:
> @lixinchengdu why are there some many changes in this PR? Isn't this only aimed to introduce SMREntryLocator?

I will try to rebase it in order make the change clearer.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecord.java
```diff
@@ -10,12 +10,15 @@
 import lombok.Setter;
 import lombok.NonNull;
 
+
+import java.util.Arrays;
```
WenbinZhu: Not used.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecordLocator.java
```diff
@@ -0,0 +1,51 @@
+package org.corfudb.protocols.logprotocol;
+
+import io.netty.buffer.ByteBuf;
+import lombok.*;
```
WenbinZhu: Expand.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecord.java
lixinchengdu: It is used in line 181

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecordLocator.java
lixinchengdu: done.

PR #1912, opened 2019-05-24
ICorfuPayload supports nested containers

lixinchengdu:
## Overview

The existing ICorfuPlayload cannot serialize/deserialize nested containers. This patch enable this feature.

Description:

Why should this be merged: 

In the future, GC would write a RPC message to inform LogUnits about garbage collection decision. This message contains nested containers.  

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/DetectorMsg.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/DetectorMsg.java
@@ -1,5 +1,6 @@
 package org.corfudb.protocols.wireprotocol;
 
+import com.google.common.reflect.TypeToken;
 import io.netty.buffer.ByteBuf;
 
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


lixinchengdu:
> I think probably we don't need to support nested containers. Using TypeToken makes code less cleaner, and TypeToken is heavily uses reflections to resolve types, which is not performant. In the current codebase, we usually create another ICorfuPayload class for the inner data structure to avoid nested containers. That would be cleaner and more performant. I think for the GC part, we can similarly create another class for the inner data structure.

There will no performance loss if the client does not use nested containers. The purpose to have ICorfuPayload is to ease the effort to define the serializer for every wired data structure. I agree that if the performance difference is significant, we could design a dedicate serializer. Nevertheless, for functions not on critical path, it could use nested containers if they want. After all, the client does not hurt if they do not use nested containers.

WenbinZhu:
> There will no performance loss if the client does not use nested conthttps://guides.github.com/features/mastering-markdown/ainers. The purpose to have ICorfuPayload is to ease the effort to define the serializer for every wired data structure. I agree that if the performance difference is significant, we could design a dedicate serializer. Nevertheless, for functions not on critical path, it could use nested containers if they want. After all, the client does not hurt if they do not use nested containers.

Actually for GC, sending garbage information will frequent, so performance does matter. If we can come up with some data structure for GC that does not need nested containers, it would be better. Let's see if we can come up with any.

PR #1882, opened 2019-05-08
CorfuQueue: Persisted Queue over State Machine Replication

hisundar:
## Overview

Description:
A persisted queue over CorfuDB using state machine replication.
String append(Object) : Generates a non-conflicting UUID and possibly inserts into a CorfuTable
Object remove(id): Removes the object using the id
List<Map.Entry<String, Object>> entryList() returns a list of all the entries along with their ids, sorted by their realized stream addresses (thus logically a FIFO queue).

Why should this be merged:
An interface that now supports ordered insertion and retrieval without any fundamental alterations.

Related issue(s) (if applicable): #

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [ ] Change is covered by automated tests
- [x] Public API has Javadoc

PR #1913, opened 2019-05-29
CorfuGuidGenerator: Use SequencerServer to make unique ordered tokens

hisundar:
A simple counter within the Sequence server that issues
ordered guids (OGUIDs) tokens by simply incrementing a counter
and shipping it with the epoch.
If the Sequence server restarts or regresses, the epoch will tick up
to guarantee ordering and global uniqueness.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
@@ -1,0 +1,26 @@
+package org.corfudb.runtime.collections;
+
+import org.corfudb.annotations.CorfuObject;
+import org.corfudb.annotations.DontInstrument;
+import org.corfudb.annotations.MutatorAccessor;
+
+@CorfuObject
+public class CorfuCounter {
+    private long counter;
+
+    public CorfuCounter() {
+        counter = 0L;
+    }
+
+    @MutatorAccessor(name="next", undoFunction = "prev")
+    public Long next() {
+        counter = counter+1;
+        return counter;
+    }
+
+    @DontInstrument
+    protected Long prev(CorfuCounter corfuCounter) {
+        corfuCounter.counter = corfuCounter.counter-1;
+        return corfuCounter.counter;
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
@@ -1,0 +1,130 @@
+package org.corfudb.runtime.view;
+
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.SMRMap;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
+import org.corfudb.runtime.object.transactions.TransactionType;
+
+import java.util.UUID;
+
+/**
+ * Globally Unique Identity generator that returns ids
+ * with weak comparable ordering with minimal distributed state sync.
+ * On startup increment a globally unique id via Corfu Object.
+ * Use this id with locally incrementing values to return unique ids.
+ * Higher bits are made of an UTC timestamp while lower bits carry the
+ * counter.
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+@Slf4j
+public class CorfuGuidGenerator implements OrderedGuidGenerator {
+    private final static int TIMESTAMP_SHIFT            = 24;
+    private final static int INSTANCE_ID_SHIFT          = 8;
+    private final static int MAX_TIMESTAMP_CORRECTION   = 0xFF;
+    private final static int MAX_INSTANCE_ID            = 0xFFff;
+
+    private final String GUID_STREAM_NAME = "CORFU_GUID_COUNTER_STREAM";
+    private final Integer GUID_STREAM_KEY = 0xdeadbeef;
+
+    private final SMRMap<Integer, Long> distributedCounter;
+    private final CorfuRuntime runtime;
+
+    /**
+     * Initialize timestamp & correction to MAX to force an update on the instance id on first call.
+     * This prevents the instance id from bumping up if there are no calls made to this class.
+     */
+    private long timestampCorrection = MAX_TIMESTAMP_CORRECTION;
+    private long previousTimestamp = Long.MAX_VALUE;
+    private long instanceId = 0L;
+
+    public CorfuGuidGenerator(CorfuRuntime rt) {
+        runtime = rt;
+        distributedCounter = rt.getObjectsView().build()
+                               .setType(SMRMap.class)
+                               .setStreamName(GUID_STREAM_NAME)
+                               .open();
+    }
+
+    private long updateInstanceId() {
+        long nextInstanceId = 0L;
+        while(true) {
+            try {
+                runtime.getObjectsView().TXBuild()
+                        .type(TransactionType.OPTIMISTIC)
+                        .build()
+                        .begin();
+
+                nextInstanceId = distributedCounter.getOrDefault(GUID_STREAM_KEY, 1L) + 1;
+                distributedCounter.put(GUID_STREAM_KEY, nextInstanceId);
+
+                runtime.getObjectsView().TXEnd();
+                break;
+            } catch (TransactionAbortedException e) {
+                log.error("updateInstanceId: Transaction aborted while updating GUID counter", e);
+            }
+        }
+
+        instanceId = nextInstanceId;
+        return nextInstanceId;
+    }
+
+    /**
+     * +----------------+--------------------+------------+
+     * |  UTC Timestamp | Unique Instance ID | Correction |
+     * +----------------+--------------------+------------+
+     * <----5 bytes----><------2 bytes------><---1 byte--->
+     *
+     * @return a globally unique id with minimal distributed sync up
+     */
+    @Override
+    public long nextLong() {
+        long currentTimestamp = getCorrectedTimestamp();
+        return (currentTimestamp << TIMESTAMP_SHIFT) |
+                ((instanceId & MAX_INSTANCE_ID) << INSTANCE_ID_SHIFT) |
+                (timestampCorrection & MAX_TIMESTAMP_CORRECTION);
+    }
+
+    /**
+     * +-------------------++--------------+------------+
+     * | UTC Timestamp     || Instance Id  | Correction |
+     * +-------------------++--------------+------------+
+     * <------8 byte-------><---7 byte------><--1 byte-->
+     *
+     * @return a higher resolution UUID with a timestamp built in.
+     */
+    @Override
+    public UUID nextUUID() {
+        long currentTimestamp = getCorrectedTimestamp();
+        return new UUID(currentTimestamp,
+                (instanceId << INSTANCE_ID_SHIFT) | (timestampCorrection & MAX_TIMESTAMP_CORRECTION));
+
+    }
+
+    /**
+     * Ensure that timestamp is monotonically increasing and if not,
+     * 1. First bump up the correction
+     * 2. If correction has also rolled over, bump up instance id
+     *
+     * @return currentTimestamp but with side-effects to object state
+     */
+    private synchronized long getCorrectedTimestamp() {
+        long currentTimestamp = System.currentTimeMillis();
+        if (currentTimestamp <= previousTimestamp) {
+            timestampCorrection++;
+            if (timestampCorrection > MAX_TIMESTAMP_CORRECTION) {
+                timestampCorrection = 0;
+                log.info("updateInstanceId: CorfuGuidGenerator corrected timestamp "+
+                                MAX_TIMESTAMP_CORRECTION+
+                                " times!"+
+                                " timestamp={} previousTimestamp={} correction={} instanceId={}",
+                        currentTimestamp, previousTimestamp, timestampCorrection, instanceId);
+                updateInstanceId();
+            }
+        }
+        previousTimestamp = currentTimestamp;
+        return currentTimestamp;
+    }
+
+}
--- a/test/src/test/java/org/corfudb/runtime/view/CorfuGuidGeneratorTest.java
+++ b/test/src/test/java/org/corfudb/runtime/view/CorfuGuidGeneratorTest.java
@@ -8,37 +8,41 @@ import java.util.UUID;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
+ * Validate that Corfu token based Guid Generator satisfies the following two criteria:
+ * 1. The guids are unique.
+ * 2. The guids are monotonically increasing.
  * Created by Sundar Sridharan on May 23, 2019.
  */
-public class SnowflakeGuidGeneratorTest {
+public class CorfuGuidGeneratorTest extends AbstractViewTest {
 
     @Test
     public void areUniqueAndOrderedLong() {
-        final int iterations = 1000;
-        OrderedGuidGenerator guidGenerator = new SnowflakeGuidGenerator(System.identityHashCode(this));
+        final int iterations = PARAMETERS.NUM_ITERATIONS_MODERATE;
+        OrderedGuidGenerator guidGenerator = new CorfuGuidGenerator(getDefaultRuntime().connect());
         long lastValue = 0;
         HashSet<Long> uniq = new HashSet<>(iterations);
         for (int i = 0; i < iterations; i++) {
             long current = guidGenerator.nextLong();
+            assertThat(uniq).doesNotContain(current);
             assertThat(current).isGreaterThan(lastValue);
             lastValue = current;
-            assertThat(uniq).doesNotContain(current);
             uniq.add(current);
         }
     }
 
     @Test
     public void areUniqueAndOrderedUUID() {
-        final int iterations = 1000;
-        OrderedGuidGenerator guidGenerator = new SnowflakeGuidGenerator(System.identityHashCode(this));
+        final int iterations = PARAMETERS.NUM_ITERATIONS_MODERATE;
+        OrderedGuidGenerator guidGenerator = new CorfuGuidGenerator(getDefaultRuntime().connect());
         UUID lastValue = new UUID(0,0);
         HashSet<UUID> uniq = new HashSet<>(iterations);
         for (int i = 0; i < iterations; i++) {
             UUID current = guidGenerator.nextUUID();
+            assertThat(uniq).doesNotContain(current);
             assertThat(current).isGreaterThan(lastValue);
             lastValue = current;
-            assertThat(uniq).doesNotContain(current);
             uniq.add(current);
         }
     }
 }
+


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -96,6 +96,12 @@
     @Getter
     private long globalLogTail = Address.getMinAddress();
 
+    /**
+     * - {@link SequencerServer::globalSequence}: A simple counter that goes up on every request.
+     */
+    @Getter
+    private Long globalSequence = Address.getMinAddress() + 1;
```
xnull: why +1?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -552,6 +562,29 @@ private void handleTxToken(CorfuPayloadMsg<TokenRequest> msg, ChannelHandlerCont
...
+     * @param msg corfu message containing allocation
+     * @param ctx netty ChannelHandlerContext
+     * @param r   server router
+     */
+    private void handleOrderedGuidToken(CorfuPayloadMsg<TokenRequest> msg,
+                                        ChannelHandlerContext ctx, IServerRouter r) {
+        final TokenRequest req = msg.getPayload();
+
+        Token token;
+        synchronized (globalSequence) {
```
xnull: Sequencer is a single threaded, so you don't need `synchronized`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -552,6 +562,29 @@ private void handleTxToken(CorfuPayloadMsg<TokenRequest> msg, ChannelHandlerCont
...
+            token = new Token(sequencerEpoch, globalSequence);
+            globalSequence += req.getNumTokens();
+        }
+        r.sendResponse(ctx, msg, CorfuMsgType.TOKEN_RES.payloadMsg(
```
xnull: what if token is null?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
```diff
@@ -78,6 +81,17 @@ public TokenRequest(Long numTokens, List<UUID> streams) {
...
 
+    /**
+     * Constructor for generating an Ordered Globally Unique Id TokenRequest.
+     * {@link TokenRequest::TK_OGUID}
+     *
+     */
+    public TokenRequest(Long numTokens) {
+        this.reqType = TK_OGUID;
+        this.numTokens = numTokens;
+        this.streams = null;
```
xnull: is it possible to avoid using null-s here?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
```diff
@@ -123,16 +138,26 @@ public TokenRequest(ByteBuf buf) {
...
+                break;
+
+            case TK_TX:
+                ICorfuPayload.serialize(buf, numTokens);
+                ICorfuPayload.serialize(buf, streams);
+                ICorfuPayload.serialize(buf, txnResolution);
+                break;
         }
     }
 }
```
xnull: don't forget to add an `enter`

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Created by Sundar Sridharan on 5/22/19.
+ */
+public class CorfuGuidGenerator implements OrderedGuidGenerator {
+    // 3 bytes for epoch ~= 16.7 million epoch changes before rollover.
+    private final static int MAX_EPOCH = 0xFFffFF;
+    private final static int SHIFT_EPOCH = 40;
+    // 5 bytes for sequence ~= 1 trillion sequence changes per epoch before rollover.
+    private final static long MAX_SEQUENCE = 0xFFffFFffFFL;
+    CorfuRuntime rt;
```
xnull: private final?
Also can you please avoid sing runtime here if possible?

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+    public CorfuGuidGenerator(CorfuRuntime rt) {
+        this.rt = rt;
+    }
+
+    @Override
+    public long nextLong() {
+        TokenResponse token = rt.getSequencerView().nextOrderedGuid();
```
xnull: this looks strange. Can you please explain why  `nextOrderedGuid ` situated inside SeqencerView?

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
```diff
@@ -92,6 +92,18 @@ public TokenResponse next(UUID ... streamIds) {
         }
     }
 
+    /**
+     * Return the next ordered globally unique sequence number from the sequencer.
+     *
+     * @return A token that can serve as the globally ordered unique sequence number.
+     */
+    public TokenResponse nextOrderedGuid() {
```
xnull: Try to avoid using CFUtils, you should use CompletableFutures instead. I can help to figure out how to change this piece of code

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
```diff
@@ -92,6 +92,18 @@ public TokenResponse next(UUID ... streamIds) {
...
+        try (Timer.Context context = MetricsUtils.getConditionalContext(sequencerNextOneStream)){
+            return layoutHelper(e -> CFUtils.getUninterruptibly(e.getPrimarySequencerClient().
+                    nextOrderedGuid(1)));
```
xnull: magic number

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
hisundar: that explains a lot. thank you!

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
hisundar: Multiple reasons:
1. If the min address is set to a value that was already returned, this avoids returning a counter value that was already returned before. 
2. Avoids returning 0 as a counter value (since 0 is the min address). This helps distinguish from initial state
3. easier for testing since we can use compare operator starting at 0.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
hisundar: thanks! I will modify it as you suggest.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xnull: can you add this to the java doc?

xnull: maybe even add this number into Address class?

hisundar: done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
hisundar: It's required for the serializer.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: Done. Added private final, also will just add the SequencerView as you suggest.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: it is a wrapper around the SequencerView's nextOrderedGuid()

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
hisundar: Its copy pasted code :( I may need to refactor entire RPC layer

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
hisundar: copy pasted code from nextToken.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -552,6 +562,29 @@ private void handleTxToken(CorfuPayloadMsg<TokenRequest> msg, ChannelHandlerCont
         handleAllocation(msg, ctx, r);
     }
 
+    /**
+     * this method serves a monotonically increasing counter as an Ordered GUID.
+     * It simply returns a counter along with the current epoch value.
+     * If the sequencer restarts the epoch ticks up while the counter resets, thereby
+     * allowing a globally unique monotonically increasing sequence to be served.
+     *
+     * @param msg corfu message containing allocation
```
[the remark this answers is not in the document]
hisundar: yes, thanks I will update the comment.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
xnull: You still can use Optional, let's try to implement that tomorrow

runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
```diff
@@ -73,6 +73,17 @@ public SequencerClient(IClientRouter router, long epoch) {
                 new TokenRequest(numTokens, streamIDs, conflictInfo)));
     }
 
+    /**
+     * Fetches the next counter value from the sequencer as an ordered globally unique id.
+     *
+     * @param numTokens Number of tokens to be reserved.
+     * @return A completable future with the token response from the sequencer.
+     */
+    public CompletableFuture<TokenResponse> nextOrderedGuid(long numTokens) {
```
WenbinZhu: If we are always passing `1` for `numTokens`, why we need this parameter?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -552,6 +562,30 @@ private void handleTxToken(CorfuPayloadMsg<TokenRequest> msg, ChannelHandlerCont
...
+     * @param r   server router
+     */
+    private void handleOrderedGuidToken(CorfuPayloadMsg<TokenRequest> msg,
+                                        ChannelHandlerContext ctx, IServerRouter r) {
+        final TokenRequest req = msg.getPayload();
+
+        // GlobalSequence points to the next available value, so return it first...
+        Token token = new Token(sequencerEpoch, globalSequence);
+        // ... then increment it by the number of such tokens requested.
+        globalSequence += req.getNumTokens();
```
WenbinZhu: It seems not making sense to increase the sequence by `numTokens`  and return only one sequence. Besides the requested `numTokens` in SequencerView is always 1. This seems weird, right?

runtime/src/main/java/org/corfudb/runtime/view/Address.java
```diff
@@ -38,6 +38,12 @@ public static boolean isAddress(long addr) {
     @Getter
     private static long minAddress = 0L;
 
+    /**
+     * Return the first address higher than the minAddress so sequence is never 0.
```
lixinchengdu: Why sequence should be greater than 0?

runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
hisundar: 2 reasons: 
1. I am following same patter as next() Token, so it is copy pasted convention in a way.
2. There was a discussion that we will soon need ability to request multiple tokens/ids in one request atomically. So this just provisions for that.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
hisundar: 2 reasons:
I am following same patter as next() Token, so it is copy pasted convention in a way.
There was a discussion that we will soon need ability to request multiple tokens/ids in one request atomically. So this just provisions for that.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
hisundar: Added comment instead as we discussed.

runtime/src/main/java/org/corfudb/runtime/view/Address.java
hisundar: It's really helpful in sanity checking the sequence and also for testing. I added a comment explaining this.

hisundar:
> See if we need to reset the globalPointer to 0 on a sequencer reset.

While it will probably work even if we don't, I think it would be better to explicitly reset it to 0 on sequencer reset to validate that only the epoch ticks up. Thanks I will make this change.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -567,6 +580,30 @@ private void handleTxToken(CorfuPayloadMsg<TokenRequest> msg, ChannelHandlerCont
...
+     * If the sequencer restarts the epoch ticks up while the counter resets, thereby
+     * allowing a globally unique monotonically increasing sequence to be served.
+     *
+     * Not Thread Safe: Only OK when SequencerServer is single threaded.
+     *
+     * @param msg corfu message containing guid request
+     * @param ctx netty ChannelHandlerContext
+     * @param r   server router
+     */
+    private void handleOrderedGuidToken(CorfuPayloadMsg<TokenRequest> msg,
```
Maithem: Since ordering is only determined post commit, this id generation scheme should only address the "uniqueness" issue, where ids don't collide, ordering is not required at this point.

You acquire a token once (during initializing), not per enqueue and then use that to generate unique tokens within that client.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
```diff
@@ -78,6 +81,18 @@ public TokenRequest(Long numTokens, List<UUID> streams) {
         txnResolution = null;
     }
 
+    /**
+     * Constructor for generating an Ordered Globally Unique Id TokenRequest.
+     * {@link TokenRequest::TK_OGUID}
```
Maithem: We can use a write token for uniqueness, please don't add this new API to the sequencer.

Maithem: Also, note that after you acquire the token, you need to write to it, to prevent the sequence regression.

hisundar: I thought the whole idea of the new api was to avoid the write and the hole fills etc?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
hisundar: Sure, but here the ordering comes for free, so it's just advertised as such. Any issue with advertising the ordering?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
Maithem: The hole-issue is only a problem when we bump the sequencer per enqueue, but if you do it once every time a client boots, then it wont be a problem, since clients are not expected to restart at a really high rate.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
Maithem: True, but now we don't even need the sequencer to generate the GUID because its being generated on the client via the regular token acquire.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
hisundar: Sure, but we are not making an assumption on the usage, right? Also what is the real issue with the new API? Its just another switch case in an existing call right?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
hisundar: Got it. Will change this.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenRequest.java
Maithem: Its not about code complexity, but since its not really necessary why add a new API?

hisundar: New API avoids a write why add a write when we don't need to? Moreover this was your idea to begin with :)

hisundar: Will remove the api and use the token approach you suggest instead. thanks

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+    private final static long TIMESTAMP_CORRECTION_MASK = 0xFF00000000000000L;
+    private final static int TIMESTAMP_CORRECTION_SHIFT = 56;
+    private final static long TIMESTAMP_MASK = 0x00FFffFFffFF0000L;
+    private final static int TIMESTAMP_SHIFT = 16;
+    private final static int MAX_TIMESTAMP_CORRECTION = 0xFF;
+    private final static int MAX_INSTANCE_ID = 0xFFff;
+
+    private final String GUID_STREAM_NAME = "CORFU_GUID_COUNTER_STREAM";
+
+    private CorfuCounter distributedCounter;
```
Maithem: An a SMR counter is not needed to obtain a unique number. Its cleaner to just use:

```java
        TokenResponse resp = rt.getSequencerView().next();
        ILogData ld = rt.getAddressSpaceView().read(resp.getSequence());
        instanceId = resp.getSequence();
```

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * +------------------------------+------------------------------+
+     * | Correction | Timestamp       |      Instance Id             |
+     * +------------------------------+------------------------------+
+     * <----1 byte--><---7 bytes-----><-------- 8 bytes ------------->
+     *
+     * @return non lossy UUID carrying both epoch and sequence
+     */
+    @Override
+    public UUID nextUUID() {
```
Maithem: From what i understand, all this machinery is needed to generate unique keys, but is this all needed?
Why isn't it just <instanceId><counter>, where instanceId is a sequencer number obtained from the sequencer and counter is a local atomic long that gets incremented on every generation.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * +----------------------+---------------+--------------------+
+     * | Timestamp Correction | UTC Timestamp | Unique Instance ID |
+     * +----------------------+---------------+--------------------+
+     * <------1 byte ---------><--5 bytes----><------2 bytes------->
+     *
+     * @return a globally unique id with minimal distributed sync up
+     */
+    @Override
+    public long nextLong() {
```
Maithem: btw the java doc says that this is a GUID (but it isn't)

> a compact but low resolution ordered guid

runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
```diff
@@ -0,0 +1,26 @@
+package org.corfudb.runtime.collections;
```
Maithem: Doesn't seem to be required. Please see comment below.

test/src/test/java/org/corfudb/runtime/view/CorfuGuidGeneratorTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Test
+    public void areUniqueAndOrderedLong() {
+        final int iterations = PARAMETERS.NUM_ITERATIONS_MODERATE;
+        OrderedGuidGenerator guidGenerator = new CorfuGuidGenerator(getDefaultRuntime());
+        long lastValue = 0;
+        HashSet<Long> uniq = new HashSet<>(iterations);
+        for (int i = 0; i < iterations; i++) {
+            long current = guidGenerator.nextLong();
+            lastValue = current;
```
Maithem: `lastValue` is actually not being used, also the test seems to be missing the second check (i.e. order of longs)

test/src/test/java/org/corfudb/runtime/view/CorfuGuidGeneratorTest.java
```diff
@@ -0,0 +1,47 @@
...
+            assertThat(uniq).doesNotContain(current);
+            uniq.add(current);
+        }
+    }
+
+    @Test
+    public void areUniqueAndOrderedUUID() {
+        final int iterations = PARAMETERS.NUM_ITERATIONS_MODERATE;
+        OrderedGuidGenerator guidGenerator = new CorfuGuidGenerator(getDefaultRuntime());
+        UUID lastValue = new UUID(0,0);
```
Maithem: `lastValue` last value is not being used, and order is not checked.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: The main reason for having a separate counter instead of hole filling is to allow for a densely packed instance id, which, when packed in smaller fields will be much harder to collide.
If the instance id is 2 bytes long and we use the the token, it is quite possible, that between 2 calls, there are over 65K mutations, which means 2 instance ids will be the same. However if they are stored in a separate counter, we will need 65K clients to come up before we hit any collision, which is practically impossible.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: fixed.

test/src/test/java/org/corfudb/runtime/view/CorfuGuidGeneratorTest.java
hisundar: thanks, fixed.

test/src/test/java/org/corfudb/runtime/view/CorfuGuidGeneratorTest.java
hisundar: thanks, fixed.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
hisundar: Please see response below.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.corfudb.annotations.CorfuObject;
+import org.corfudb.annotations.DontInstrument;
+import org.corfudb.annotations.MutatorAccessor;
+
+@CorfuObject
+public class CorfuCounter {
```
Maithem: This needs to be checkpointed. Why not just use a map that's already supported?

Also, from what I understand multiple clients will be using this shared counter, so it seems like transactions are needed (i.e. need conflict parameters).

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,139 @@
...
+     * This prevents the instance id from bumping up if there are no calls made to this class.
+     */
+    private long timestampCorrection = MAX_TIMESTAMP_CORRECTION;
+    private long previousTimestamp = Long.MAX_VALUE;
+    private long instanceId = 0L;
+
+    public CorfuGuidGenerator(CorfuRuntime rt) {
+        runtime = rt;
+        try {
+            distributedCounter = (SMRMap<Integer, Long>)
```
Maithem: Why try/catch? just let it fail.

Also, the cast is not necessary.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,139 @@
...
+                            .open());
+        } catch (Exception e) {
+            log.error("Failed to initialize CorfuGuidGenerator", e);
+            throw e;
+        }
+    }
+
+    private long updateInstanceId() {
+        long nextInstanceId = 0L;
+        boolean transactionAborted = false;
```
Maithem: Instead of tracking an extra variable, you can while(true) and use break.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,139 @@
...
+                        .build()
+                        .begin();
+
+                nextInstanceId = distributedCounter.getOrDefault(GUID_STREAM_KEY, 1L) + 1;
+                distributedCounter.put(GUID_STREAM_KEY, nextInstanceId);
+
+                runtime.getObjectsView().TXEnd();
+                transactionAborted = false;
+            } catch (TransactionAbortedException e) {
+                log.error("Transaction aborted while trying to update GUID counter", e);
```
Maithem: prefix the error message with the method name `updateInstanceId`

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,139 @@
...
+                transactionAborted = true;
+            }
+        } while (transactionAborted);
+
+        instanceId = nextInstanceId;
```
Maithem: Why track the instanceId if you can just access it from the map?

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,139 @@
...
+                        .type(TransactionType.OPTIMISTIC)
+                        .build()
+                        .begin();
+
+                nextInstanceId = distributedCounter.getOrDefault(GUID_STREAM_KEY, 1L) + 1;
+                distributedCounter.put(GUID_STREAM_KEY, nextInstanceId);
+
+                runtime.getObjectsView().TXEnd();
+                transactionAborted = false;
+            } catch (TransactionAbortedException e) {
```
Maithem: `TransactionAbortedException` has an `AbortCause` do you want to retry on all causes?

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -0,0 +1,139 @@
+package org.corfudb.runtime.view;
+
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.SMRMap;
+import org.corfudb.runtime.collections.SMRMap$CORFUSMR;
```
Maithem: Remove this import.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
Maithem: I see. This class seems to be obsolete. Please remove it.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: fixed.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: just to avoid another lookup and a lock?

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: 1. With timestamp baked in, it allows debugging since the key carries the time when it was generated.
2. Given that typical deployments have NTP, this also brings in global ordering across nodes.
I thought these are good properties to have?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
hisundar: Sorry looks like this class got re-added when I did a rebase... It is not used anywhere and will be removed in https://github.com/CorfuDB/CorfuDB/pull/1882 which depends on this PR

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: yes, for now that seemed like the simplest thing to do.
we can enhance this to handle specific failures later?

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: done. already fixed.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuCounter.java
hisundar: this class will be removed in https://github.com/CorfuDB/CorfuDB/pull/1882

PR #1918, opened 2019-06-02
Transfer delta in state transfer.

### LINKED ISSUE #1910: Add node fails if partial state transfer fails.
## Overview

These are the following events:
- The first Join is successful.
- The second join starts:
- The Orchestrator starts syncing the data to the newly joined node.
- State transfer continues in chunks of 10 addresses until a certain chunk which is too large:

Thread-8 |  o.c.i.o.actions.StateTransfer | stateTransfer: read 24655-24754 in 6 ms
Thread-8 |  o.c.i.o.actions.StateTransfer | stateTransfer: Transferred address chunk [24655, 24754] to 10.172.177.55:9000 in 170 ms

Large chunk. The read takes 4.7 secs. The Corfu timeout is 5 secs.
Thread-8 |  o.c.i.o.actions.StateTransfer | stateTransfer: read 24755-24854 in 4739 ms

The write timesout on the client side.
Thread-8 |      o.c.i.orchestrator.Action | execute: Error executing action RestoreRedundancy on retry 0. Invalidating layout.
java.util.concurrent.ExecutionException: java.util.concurrent.TimeoutException
However, the data was written on the server.
We fail on attempting to re-write this chunk with an overwrite exception.
The Orchestrator needs to check if the data to be copied is the same and ignore the overwrite exception.

We do not see this in Heal Node workflow as it wipes out the data before retrying.

zalokhan:
## Overview

Description:
Transfer delta in state transfer.

Why should this be merged: 
If transfer range overlaps with previous transfer, overwrite exceptions are thrown and state transfer fails.

Related issue(s) (if applicable): Fixes #1910 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -4,10 +4,24 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.google.common.annotations.VisibleForTesting;
+
 import io.netty.channel.ChannelHandlerContext;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
 import lombok.Builder;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+
 import org.corfudb.infrastructure.log.InMemoryStreamLog;
 import org.corfudb.infrastructure.log.StreamLog;
 import org.corfudb.infrastructure.log.StreamLogCompaction;
@@ -17,6 +31,7 @@ import org.corfudb.protocols.wireprotocol.CorfuMsgType;
 import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
 import org.corfudb.protocols.wireprotocol.FillHoleRequest;
 import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.protocols.wireprotocol.KnownAddressRequest;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.MultipleReadRequest;
 import org.corfudb.protocols.wireprotocol.RangeWriteMsg;
@@ -38,17 +53,13 @@ import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.view.stream.StreamAddressSpace;
 import org.corfudb.util.Utils;
 
-import java.lang.invoke.MethodHandles;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.*;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.LOG_ADDRESS_SPACE_QUERY;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.PREFIX_TRIM;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.RANGE_WRITE;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.RESET;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.SEAL;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.TAILS_QUERY;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.WRITE;
 
 
 /**
@@ -320,6 +331,25 @@ public class LogUnitServer extends AbstractServer {
             r.sendResponse(ctx, msg, CorfuMsgType.READ_RESPONSE.payloadMsg(rr));
         } catch (DataCorruptionException e) {
             r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_CORRUPTION.msg());
+        }
+    }
+
+    /**
+     * Handles requests for known entries in specified range.
+     * This is used by state transfer to catch up only the remainder of the segment.
+     */
+    @ServerHandler(type = CorfuMsgType.KNOWN_ADDRESS_REQUEST)
+    private void getKnownAddressesInRange(CorfuPayloadMsg<KnownAddressRequest> msg,
+                                          ChannelHandlerContext ctx, IServerRouter r) {
+
+        KnownAddressRequest request = msg.getPayload();
+        try {
+            Set<Long> knownAddresses = streamLog
+                    .getKnownAddressesInRange(request.getStartRange(), request.getEndRange());
+            r.sendResponse(ctx, msg,
+                    CorfuMsgType.KNOWN_ADDRESS_RESPONSE.payloadMsg(knownAddresses));
+        } catch (Exception e) {
+            handleException(e, ctx, msg, r);
         }
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
@@ -1,6 +1,7 @@
 package org.corfudb.infrastructure.log;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -122,6 +123,26 @@ public class InMemoryStreamLog implements StreamLog, StreamLogWithRankedAddressS
             // the method below might throw DataOutrankedException or ValueAdoptedException
             assertAppendPermittedUnsafe(address, entry);
         }
+    }
+
+    /**
+     * Returns the known addresses in this Log Unit in the specified consecutive
+     * range of addresses.
+     *
+     * @param rangeStart Start address of range.
+     * @param rangeEnd   End address of range.
+     * @return Set of known addresses.
+     */
+    @Override
+    public Set<Long> getKnownAddressesInRange(long rangeStart, long rangeEnd) {
+
+        Set<Long> result = new HashSet<>();
+        for (long address = rangeStart; address <= rangeEnd; address++) {
+            if (logCache.containsKey(address)) {
+                result.add(address);
+            }
+        }
+        return result;
     }
 
     @Override
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
@@ -2,10 +2,10 @@ package org.corfudb.infrastructure.log;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 
 import org.corfudb.protocols.wireprotocol.LogData;
-import org.corfudb.protocols.wireprotocol.StreamAddressRange;
 import org.corfudb.protocols.wireprotocol.StreamsAddressResponse;
 import org.corfudb.protocols.wireprotocol.TailsResponse;
 import org.corfudb.runtime.exceptions.OverwriteCause;
@@ -76,6 +76,16 @@ public interface StreamLog {
      * Get the first untrimmed address in the address space.
      */
     long getTrimMark();
+
+    /**
+     * Returns the known addresses in this Log Unit in the specified consecutive
+     * range of addresses.
+     *
+     * @param rangeStart Start address of range.
+     * @param rangeEnd   End address of range.
+     * @return Set of known addresses.
+     */
+    Set<Long> getKnownAddressesInRange(long rangeStart, long rangeEnd);
 
     /**
      * Sync the stream log file to secondary storage.
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -9,9 +9,12 @@ import com.google.common.hash.Hashing;
 import com.google.protobuf.AbstractMessage;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
+
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+
 import lombok.extern.slf4j.Slf4j;
+
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.corfudb.format.Types;
@@ -28,11 +31,7 @@ import org.corfudb.runtime.exceptions.DataCorruptionException;
 import org.corfudb.runtime.exceptions.OverwriteCause;
 import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
-import org.corfudb.runtime.view.Address;
-import org.corfudb.runtime.view.stream.StreamAddressSpace;
-import org.corfudb.util.Utils;
 
-import javax.annotation.Nullable;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.FileNotFoundException;
@@ -60,6 +59,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Collectors;
 
+import javax.annotation.Nullable;
 
 /**
  * This class implements the StreamLog by persisting the stream log as records in multiple files.
@@ -1019,6 +1019,26 @@ public class StreamLogFiles implements StreamLog, StreamLogWithRankedAddressSpac
         long endSegment = lastAddress / RECORDS_PER_LOG_FILE;
 
         return endSegment - firstSegment <= 1;
+    }
+
+    /**
+     * This method requests for known addresses in this Log Unit in the specified consecutive
+     * range of addresses.
+     *
+     * @param rangeStart Start address of range.
+     * @param rangeEnd   End address of range.
+     * @return Set of known addresses.
+     */
+    @Override
+    public Set<Long> getKnownAddressesInRange(long rangeStart, long rangeEnd) {
+
+        Set<Long> result = new HashSet<>();
+        for (long address = rangeStart; address <= rangeEnd; address++) {
+            if (getSegmentHandleForAddress(address).getKnownAddresses().containsKey(address)) {
+                result.add(address);
+            }
+        }
+        return result;
     }
 
     @Override
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
@@ -3,13 +3,11 @@ package org.corfudb.infrastructure.orchestrator.actions;
 import com.google.common.collect.Sets;
 
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
 
 import javax.annotation.Nonnull;
 
 import org.corfudb.infrastructure.orchestrator.Action;
 import org.corfudb.runtime.CorfuRuntime;
-import org.corfudb.runtime.exceptions.OutrankedException;
 import org.corfudb.runtime.view.Layout;
 
 /**
@@ -44,9 +42,7 @@ public class RestoreRedundancyMergeSegments extends Action {
     }
 
     @Override
-    public void impl(@Nonnull CorfuRuntime runtime) throws OutrankedException,
-            ExecutionException,
-            InterruptedException {
+    public void impl(@Nonnull CorfuRuntime runtime) throws Exception {
 
         // Refresh layout.
         runtime.invalidateLayout();
@@ -64,7 +60,9 @@ public class RestoreRedundancyMergeSegments extends Action {
             // Currently the state is transferred for the complete segment.
             // TODO: Add stripe specific transfer granularity for optimization.
             // Transfer the replicated segment to the difference set calculated above.
-            StateTransfer.transfer(layout, lowRedundancyServers, runtime, layout.getFirstSegment());
+            for (String lowRedundancyServer : lowRedundancyServers) {
+                StateTransfer.transfer(layout, lowRedundancyServer, runtime, layout.getFirstSegment());
+            }
 
             // Merge the 2 segments.
             runtime.getLayoutManagementView().mergeSegments(new Layout(layout));
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
@@ -1,23 +1,26 @@
 package org.corfudb.infrastructure.orchestrator.actions;
 
-import com.google.common.collect.ContiguousSet;
-import com.google.common.collect.DiscreteDomain;
-import com.google.common.collect.Range;
-
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
 
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.Token;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.OverwriteException;
+import org.corfudb.runtime.exceptions.RetryExhaustedException;
 import org.corfudb.runtime.view.Layout;
 import org.corfudb.util.CFUtils;
+import org.corfudb.util.retry.ExponentialBackoffRetry;
+import org.corfudb.util.retry.IRetry;
+import org.corfudb.util.retry.RetryNeededException;
 
 /**
  * State transfer utility.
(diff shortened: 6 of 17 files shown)


test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1499,6 +1501,113 @@ public void verifyStateTransferNotHappenButMergeSucceeds() throws Exception {
...
+                    .addToLayout()
+                    .build();
+            bootstrapAllServers(layout);
+
+            corfuRuntime = getNewRuntime(getDefaultNode()).connect();
+
+            setAggressiveTimeouts(layout, corfuRuntime);
+
+            IStreamView testStream = corfuRuntime.getStreamsView().get(CorfuRuntime.getStreamID("test"));
+            // Write to address spaces 0 to 11_000 (inclusive) to SERVER 0 only.
```
hisundar: nit: Write to address spaces 0 to 11_000 (inclusive) go to SERVER 0 only.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1499,6 +1501,113 @@ public void verifyStateTransferNotHappenButMergeSucceeds() throws Exception {
...
+            // Write to address spaces 11_000 to 12_000 (inclusive) to SERVERS 0 & 1.
```
hisundar: nit:Write to address spaces 11_000 to 12_000 (inclusive) go to SERVERS 0 & 1.
?

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -26,6 +25,8 @@
 @Slf4j
 public class StateTransfer {
 
+    private static final int RECORDS_PER_LOG_FILE = 10_000;
```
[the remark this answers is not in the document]
zalokhan: Yes, I tried that. 
It has to be the other way round as runtime does not have any dependency on infrastructure.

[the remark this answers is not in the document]
zalokhan: Sorry, my bad. The class is under infrastructure and I can import the field directly. Will do that. Thanks.

zalokhan:
If 2 state transfer jobs are running in parallel, both of them will preempt each other since they both try to transfer the addresses not present at the destination.
I am adding an exponential backoff mechanism to this in case of an overwrite exception.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +39,206 @@ private StateTransfer() {
...
-                                Set<String> endpoints,
+                                @NonNull String endpoint,
                                 CorfuRuntime runtime,
-                                Layout.LayoutSegment segment) throws ExecutionException, InterruptedException {
+                                Layout.LayoutSegment segment) throws Exception {
 
-        if (endpoints.isEmpty()) {
-            log.debug("stateTransfer: No server needs to transfer for segment [{} - {}], " +
-                    "skipping state transfer for this segment.", segment.getStart(), segment.getEnd());
+        if (endpoint.isEmpty()) {
```
WenbinZhu: This block should be removed as endpoints is no longer a `Set`.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
```diff
@@ -64,7 +60,9 @@ public void impl(@Nonnull CorfuRuntime runtime) throws OutrankedException,
             // Currently the state is transferred for the complete segment.
             // TODO: Add stripe specific transfer granularity for optimization.
             // Transfer the replicated segment to the difference set calculated above.
-            StateTransfer.transfer(layout, lowRedundancyServers, runtime, layout.getFirstSegment());
+            for (String lowRedundancyServer : lowRedundancyServers) {
```
WenbinZhu: This whole merge logic assumes no more than two segments exists. It inspects the first two segments, do transfer, but merge the last two segment, which is wrong if there are multiple segments (could happen). This wrong logic has been there for a while, can we fix it this time?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -89,6 +89,8 @@
     /** The node Id, stored as a base64 string. */
     private static final String NODE_ID = "NODE_ID";
 
+    public static final int RECORDS_PER_LOG_FILE = 10_000;
```
Maithem: The way the LogUnit partitions its data is implementation detail and its not meant to be exposed.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
```diff
@@ -77,6 +79,15 @@
...
 
+    /**
+     * Returns the missing addresses in this Log Unit in the specified consecutive
+     * range of addresses. The range cannot be larger than RECORDS_PER_LOG_FILE.
+     *
+     * @param range Range of addresses to compare.
+     * @return Set of missing addresses.
+     */
+    Set<Long> getMissingAddresses(Range<Long> range);
+
```
Maithem: I think this is more appropriate `List<Long> queryAddressesInRange(long lowerEndpoint, UpperEndpoint)`

Also, since the request size is not validated, I don't feel like it makes sense to mention RECORDS_PER_LOG_FILE. I think this limit should be set by the client anyways.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1012,6 +1018,35 @@ private boolean verify(List<LogData> range) {
...
 
+    /**
+     * This method requests for missing addresses in this Log Unit in the specified consecutive
+     * range of addresses. The range cannot be larger than RECORDS_PER_LOG_FILE.
+     *
+     * @param range Range of addresses to compare.
+     * @return Set of missing addresses.
+     */
+    @Override
+    public Set<Long> getMissingAddresses(Range<Long> range) {
```
Maithem: please simplify

```java

    public List<Long> rangeQuery(Range<Long> range) {
        List<Long> ret = new ArrayList<>(1000);
        for (long a = range.lowerEndpoint(); a <= range.upperEndpoint(); a++) {
            if (getSegmentHandleForAddress(a).getKnownAddresses().containsKey(a)) {
                ret.add(a);
            }
        }
        return ret;
    }
```

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +39,206 @@ private StateTransfer() {
...
+            log.debug("stateTransfer: No server needs to transfer for segment [{} - {}], "
+                            + "skipping state transfer for this segment.",
+                    segment.getStart(), segment.getEnd());
             return;
         }
 
-        int batchSize = runtime.getParameters().getBulkReadSize();
+        int chunkSize = runtime.getParameters().getBulkReadSize();
+        final AtomicInteger overwriteRetries = new AtomicInteger();
```
Maithem: Why is this an AtomicInteger ?

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
Maithem: Also, why call transfer, if the caller can just not call it?

zalokhan: @WenbinZhu Removed the call.
@Maithem The layout could have changed on a retry which causes the segments to have merged.
Fixed.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +39,206 @@ private StateTransfer() {
...
+                                @NonNull String endpoint,
                                 CorfuRuntime runtime,
-                                Layout.LayoutSegment segment) throws ExecutionException, InterruptedException {
+                                Layout.LayoutSegment segment) throws Exception {
 
-        if (endpoints.isEmpty()) {
-            log.debug("stateTransfer: No server needs to transfer for segment [{} - {}], " +
-                    "skipping state transfer for this segment.", segment.getStart(), segment.getEnd());
+        if (endpoint.isEmpty()) {
+            log.debug("stateTransfer: No server needs to transfer for segment [{} - {}], "
```
annym: Also this message should change right, as it is not all servers.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +39,206 @@ private StateTransfer() {
...
+                if (trimMark > segment.getEnd()) {
+                    log.info("stateTransfer: Nothing to transfer, trimMark {}"
+                                    + "greater than end of segment {}",
+                            trimMark, segment.getEnd());
+                    return true;
+                }
 
-            long ts2 = System.currentTimeMillis();
+                // State transfer should start from segment start address or trim mark
+                // whichever is lower.
```
annym: lower or higher?

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
zalokhan: That's not true. That would have been a serious data loss issue.
We merge only the segment which we transfer.
Let's discuss this offline if I am not understanding this right.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +39,206 @@ private StateTransfer() {
...
+                final long segmentStart = Math.max(trimMark, segment.getStart());
```
annym: couldn't the trim mark be out of this segment? in which case it does not correspond to the given segment end?

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1499,6 +1501,113 @@ public void verifyStateTransferNotHappenButMergeSucceeds() throws Exception {
         waitForLayoutChange(l -> l.getSegments().size() == 1, rt);
     }
 
+    /**
```
annym: Could we add a test for competing state transfers? that certainly they do not block each other.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
zalokhan: Will make it another configuration parameter.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
zalokhan: It's used in the lambda which requires a final variable.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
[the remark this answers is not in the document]
zalokhan: No, these should not be linked. I have reverted this change.
The StreamLogFiles will have its `RECORDS_PER_LOG_FILE` and the State transfer will use its own parameter.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -321,6 +333,22 @@ private void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerC
         }
     }
 
+    /**
+     * Handles requests for missing entries in specified range.
+     * This is used by state transfer to catchup only the remainder of the segment.
```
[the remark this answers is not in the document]
zalokhan: Fixed.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
zalokhan: This would acquire a lock, every time you call `getSegmentHandleForAddress`.
I can work on simplifying this.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -184,6 +187,15 @@ private void processor() {
                                 // Retrieve the address space for every stream in the log.
                                 currOp.setResultValue(streamLog.getStreamsAddressSpace());
                                 break;
+                            case MISSING_ADDRESSES_IN_RANGE:
```
Maithem: This doesn't have to be part of the batch writer, please remove it.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/MissingAddressResponse.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Response for missing addresses in the log unit server for a specified range.
+ * Created by zlokhandwala on 2019-06-01.
+ */
+@Data
+@AllArgsConstructor
+public class MissingAddressResponse implements ICorfuPayload<MissingAddressResponse> {
+
+    private final Set<Long> missingAddresses;
```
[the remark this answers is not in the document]
zalokhan: The payload is not that large. So I haven't considered this.
10,000 addresses will correspond to 80Kbytes.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1012,6 +1018,35 @@ private boolean verify(List<LogData> range) {
...
+
+        if (range.upperEndpoint() - range.lowerEndpoint() > RECORDS_PER_LOG_FILE) {
+            String msg = String.format("Range too large: [%d-%d]. Number of addresses: %d",
+                    range.lowerEndpoint(), range.upperEndpoint(),
+                    (range.upperEndpoint() - range.lowerEndpoint()));
+            throw new IllegalArgumentException(msg);
+        }
+
+        SegmentHandle firstSh = getSegmentHandleForAddress(range.lowerEndpoint());
```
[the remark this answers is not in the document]
zalokhan: They could. I changed the implementation.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
zalokhan: Fixed.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
zalokhan: higher. Fixed.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
zalokhan: Fixed this.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
zalokhan: In that case, there is nothing to transfer.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1012,6 +1017,26 @@ private boolean verify(List<LogData> range) {
...
+     * range of addresses.
+     *
+     * @param range Range of addresses to compare.
+     * @return List of missing addresses.
+     */
+    @Override
+    public List<Long> getMissingAddresses(Range<Long> range) {
+
+        List<Long> result = new ArrayList<>(10_000);
+        for (long address = range.lowerEndpoint(); address <= range.upperEndpoint(); address++) {
```
Maithem: I think the converse is more natural. Given a set, we have contains method and not doesntContain.

infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
```diff
@@ -121,6 +126,20 @@ private void throwLogUnitExceptionsIfNecessary(long address, LogData entry) {
...
 
+    /**
+     * Returns the missing addresses in this Log Unit in the specified consecutive
+     * range of addresses.
+     *
+     * @param range Range of addresses to compare.
+     * @return Set of missing addresses.
+     */
+    @Override
+    public Set<Long> getMissingAddresses(Range<Long> range) {
```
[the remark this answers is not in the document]
zalokhan: Changed this. Removed the limit now.
The range is specified by the client. A bigger range could lead to longer processing times on the server and thus should be considered while setting the parameter.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1499,6 +1501,113 @@ public void verifyStateTransferNotHappenButMergeSucceeds() throws Exception {
...
+                    .setEnd(writtenAddressesBatch1)
+                    .buildStripe()
+                    .addLogUnit(SERVERS.PORT_0)
+                    .addToSegment()
+                    .addToLayout()
+                    .buildSegment()
+                    .setStart(writtenAddressesBatch1)
+                    .setEnd(-1)
+                    .buildStripe()
+                    .addLogUnit(SERVERS.PORT_0)
```
epalta: Already did addLogUnit(SERVERS.PORT_0) at line 1546

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1499,6 +1501,113 @@ public void verifyStateTransferNotHappenButMergeSucceeds() throws Exception {
...
+                    .addSequencer(SERVERS.PORT_0)
+                    .addSequencer(SERVERS.PORT_1)
+                    .buildSegment()
+                    .setStart(0L)
+                    .setEnd(writtenAddressesBatch1)
+                    .buildStripe()
+                    .addLogUnit(SERVERS.PORT_0)
+                    .addToSegment()
+                    .addToLayout()
+                    .buildSegment()
```
epalta: Dup of line 1542

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: These are 2 different segments which are being created.
Your layout needs to look like this:
```
{
  "layoutServers": [
    "localhost:9000",
    "localhost:9001"
  ],
  "sequencers": [
    "localhost:9000",
    "localhost:9001"
  ],
  "segments": [
    {
      "replicationMode": "CHAIN_REPLICATION",
      "start": 0,
      "end": 11000,
      "stripes": [
        {
          "logServers": [
            "localhost:9000"
          ]
        }
      ]
    },
    {
      "replicationMode": "CHAIN_REPLICATION",
      "start": 11000,
      "end": -1,
      "stripes": [
        {
          "logServers": [
            "localhost:9000",
            "localhost:9001"
          ]
        }
      ]
    }
  ],
  "epoch": 1
}
```

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: Please see the other comment. This is the layout builder and both these statements are required.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +38,199 @@ private StateTransfer() {
...
+     * @param endpoint destination node
+     * @param runtime  The runtime to read the segment from
+     * @param segment  segment to transfer
      */
     public static void transfer(Layout layout,
-                                Set<String> endpoints,
+                                @NonNull String endpoint,
                                 CorfuRuntime runtime,
-                                Layout.LayoutSegment segment) throws ExecutionException, InterruptedException {
+                                Layout.LayoutSegment segment) throws Exception {
```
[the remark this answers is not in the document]
zalokhan: Done.

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1499,6 +1501,113 @@ public void verifyStateTransferNotHappenButMergeSucceeds() throws Exception {
         waitForLayoutChange(l -> l.getSegments().size() == 1, rt);
     }
 
+    /**
+     * This test verifies that partial state transfers when retried only transfer the delta
+     * of the address space.
+     * Setup: Layout - Write 11_000 entries to SERVER 0 and 1_000 entries to SERVERS 0 & 1.
+     * We then trigger the state transfer and force the transfer to fail on the last range write.
+     * Another transfer is then triggered and verified that only the remaining data is transferred.
+     */
```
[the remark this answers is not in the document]
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +38,204 @@ private StateTransfer() {
...
+                        transferChunk(layout, runtime, endpoint, chunk);
+                    }
+                }
+            } catch (OverwriteException oe) {
 
-        for (long chunkStart = segmentStart; chunkStart < segment.getEnd()
-                ; chunkStart = chunkStart + batchSize) {
-            long chunkEnd = Math.min((chunkStart + batchSize - 1), segment.getEnd() - 1);
+                // If Overwrite Exception is due to Different Data, log the error and abort.
+                if (!oe.getOverWriteCause().equals(OverwriteCause.SAME_DATA)) {
```
Maithem: This doesn't do what it seems like its doing. Look here #1931

Also, even if `OverwriteCause.SAME_DATA` does a byte-wise comparison its not sufficient because two different writes can skip each others writes without even observing the `OverwriteException`. If you really want to be paranoid then multiple writes should be reading each other's writes and verifying them against theirs.

zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
```diff
@@ -124,6 +125,25 @@ private void throwLogUnitExceptionsIfNecessary(long address, LogData entry) {
...
 
+    /**
+     * Returns the known addresses in this Log Unit in the specified consecutive
+     * range of addresses.
+     *
+     * @param range Range of addresses to compare.
+     * @return Set of known addresses.
+     */
+    @Override
+    public Set<Long> getKnownAddressesInRange(Range<Long> range) {
```
Maithem: This API is actually not very clear, because the Range can be any combination of open/close endpoints, but the for loop on like 139 assumes that the range is closed/closed. Essentially making an assumption about a generic. For example, [1,2,3) would can return 1,2 and 3, but based on the interface it should return 1 and 2

Maithem: Maybe just use the endpoints directly, no need to wrap them in a Range ?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1021,6 +1022,25 @@ private boolean verify(List<LogData> range) {
...
+    /**
+     * This method requests for known addresses in this Log Unit in the specified consecutive
+     * range of addresses.
+     *
+     * @param range Range of addresses to compare.
+     * @return Set of known addresses.
+     */
+    @Override
+    public Set<Long> getKnownAddressesInRange(Range<Long> range) {
+
```
Maithem: See comments above.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +33,197 @@ private StateTransfer() {
...
+                        ; chunkStart += chunkSize) {
 
-            long ts2 = System.currentTimeMillis();
+                    long chunkEnd = Math.min(segmentEnd, chunkStart + chunkSize - 1);
 
-            log.info("stateTransfer: read {}-{} in {} ms", chunkStart, chunkEnd, (ts2 - ts1));
+                    // Fetch all missing entries in this range [chunkStart - chunkEnd].
+                    List<Long> chunk = getMissingEntriesChunk(layout, runtime, endpoint,
+                            chunkStart, chunkEnd);
+                    if (chunk.isEmpty()) {
```
Maithem: Please log that chunk is being omitted because the orchestrator didn't detect it as missing

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +33,197 @@ private StateTransfer() {
...
-            for (long x = chunkStart; x <= chunkEnd; x++) {
-                if (!dataMap.containsKey(x)) {
-                    log.error("Missing address {} in range {}-{}", x, chunkStart, chunkEnd);
-                    throw new IllegalStateException("Missing address");
+                    // Read and write in chunks of chunkSize.
+                    transferChunk(layout, runtime, endpoint, chunk);
                 }
-                entries.add((LogData) dataMap.get(x));
-            }
+            } catch (OverwriteException oe) {
```
Maithem: log on OverWriteException

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +33,197 @@ private StateTransfer() {
...
+     *
+     * @param layout   Current layout.
+     * @param runtime  Corfu runtime instance.
+     * @param endpoint Endpoint ot transfer data to.
+     * @return Trim Address.
+     */
+    private static long setTrimOnNewLogUnit(Layout layout, CorfuRuntime runtime,
+                                            String endpoint) {
+
+        long trimMark = runtime.getAddressSpaceView().getTrimMark().getSequence();
```
Maithem: This can execute on a different layout, but subsequent calls don't. Why not call this with a runtime layout?

Also, what happens if there are other intermittent failures other than WrongEpochException ?

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
Maithem: Since `transferChunk` accesses the first and last index of the list, it assumes that the list is not empty.
It would make sense to move this check in side  the `transferChunk` method.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -30,96 +33,197 @@ private StateTransfer() {
...
+     * @param chunk    List of addresses to be read and transferred.
+     */
+    private static void transferChunk(Layout layout,
+                                      CorfuRuntime runtime,
+                                      String endpoint,
+                                      List<Long> chunk) {
+        long ts1 = System.currentTimeMillis();
+
+        Map<Long, ILogData> dataMap = runtime.getAddressSpaceView()
+                .fetchAll(chunk, true);
```
Maithem: @WenbinZhu please make sure you change this.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
zalokhan: Moved the check to transferChunk.
That is already being logged in getMissingEntries.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java
```diff
@@ -213,4 +214,18 @@ private static Object handleTrimMarkResponse(CorfuPayloadMsg<Long> msg,
...
+    /**
+     * Handle a KNOWN_ADDRESS_RESPONSE message.
+     *
+     * @param msg Incoming Message
+     * @param ctx Context
+     * @param r   Router
+     * @return KnownAddressResponse payload with the missing addresses set.
+     */
+    @ClientHandler(type = CorfuMsgType.KNOWN_ADDRESS_RESPONSE)
+    private static Object handleMissingAddressesResponse(CorfuPayloadMsg<KnownAddressResponse> msg,
```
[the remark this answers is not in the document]
zalokhan: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
zalokhan: The reads are on the AddressSpaceView layer to abstract the replication layer and also the exceptions are taken care of.
This exposes a TrimException on which the action fails and is retried with the new trim mark.
WrongEpochExceptions should be just retried. Similarly, NetworkExceptions, Timeouts.
The writes do not use this layoutHelper method and hence are aborted if any exception is hit.
I haven't modified this functionality in this PR. I will address that in the next PR with appropriate testing.

PR #1920, opened 2019-06-03
Make ping epoch unaware.

### LINKED ISSUE #1917: Force remove fails due to sealed live node
## Overview

ForceRemoveNode attempts to invoke the Corfu orchestrator after the Layout has been sealed.
There are no available orchestrators then.
2 nodes - A & B are already unresponsive and they do not respond to the ping.
C has been sealed and rejects the ping with a WrongEpochException.
This Hand of God Workflow must reach the node regardless of the epoch.

zalokhan:
## Overview

Description: Make ping epoch unaware.
A wrongEpochException means the node is reachable.

Why should this be merged: ForceRemoveNode is not able to find an orchestrator if the only remaining live node is sealed. Making the ping epoch unaware makes it possible to choose an orchestrator and hence trigger the workflow.

Related issue(s) (if applicable): Fixes #1917 

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
+++ b/test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
@@ -3,6 +3,7 @@ package org.corfudb.integration;
 import com.google.common.collect.ContiguousSet;
 import com.google.common.collect.DiscreteDomain;
 import com.google.common.collect.Range;
+
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.ReadResponse;
 import org.corfudb.protocols.wireprotocol.TokenResponse;
@@ -20,6 +21,7 @@ import org.corfudb.runtime.clients.NettyClientRouter;
 import org.corfudb.runtime.collections.CorfuTable;
 import org.corfudb.runtime.exceptions.AlreadyBootstrappedException;
 import org.corfudb.runtime.exceptions.TransactionAbortedException;
+import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.view.Layout;
 import org.corfudb.runtime.view.stream.IStreamView;
 import org.corfudb.util.CFUtils;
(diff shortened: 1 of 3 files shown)


zalokhan:
Ping by definition gives us the information whether a node is reachable. Overloading this with information about being on the same epoch does not make sense to me. As you said, for epoch information we should use another message such as `NodeStatusRequest`?
Let me know if this makes sense.

WenbinZhu:
Actually I was always having this question: why PING message is epoch aware because like you said it's just a  reachability check, but since I'm not sure if there is any consideration to keep the epoch aware PING, so I suggested overloading it. But actually my first thought was to keep only the epoch ignored PING. Do we really need it to be epoch aware?

zalokhan:
Epoch awareness was introduced when we started using pings as heartbeat requests in the Fault detector. Since we now use a different message for fault detection we can revert the functionality of the ping to be just a reachability check.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -1010,4 +1012,83 @@ public void reconnectDisconnectedClient() throws Exception {
...
+        Process corfuServer_2 = runPersistentServer(corfuSingleNodeHost, PORT_1, false);
+        Process corfuServer_3 = runPersistentServer(corfuSingleNodeHost, PORT_2, false);
+        List<Process> corfuServers = Arrays.asList(corfuServer_1, corfuServer_2, corfuServer_3);
+        final Layout layout = getLayout(3);
+        final int retries = 3;
+        Sleep.SECONDS.sleepUninterruptibly(1);
+        BootstrapUtil.bootstrap(layout, retries, PARAMETERS.TIMEOUT_SHORT);
+
+        final int systemDownHandlerLimit = 10;
+        runtime = CorfuRuntime.fromParameters(CorfuRuntimeParameters.builder()
```
epalta: Just for my understanding.
How does runtime know here about corfuServers? I don't see it being passed here.

zalokhan: It doesn't need to know about all the servers. I've passed one of the nodes as a connection endpoint `.layoutServer(NodeLocator.parseString(DEFAULT_ENDPOINT))`. Once the runtime connects to this node, it can fetch the layout and get info about all the servers in the cluster.

WenbinZhu:
> Epoch awareness was introduced when we started using pings as heartbeat requests in the Fault detector. Since we now use a different message for fault detection we can revert the functionality of the ping to be just a reachability check.

@zalokhan Can we address this in this PR? Because I feel making pingSync() epoch ignored while keeping ping() epoch aware is not clean, right?

zalokhan:
@WenbinZhu Addressed your comments.
Please take a look again.

WenbinZhu:
> @WenbinZhu Addressed your comments.
> Please take a look again.

Approved. Thanks for addressing it, please also change PR title, description and commit message: s/make pingSync/make ping

Issue #1922, opened 2019-06-04
testCorfuTablePutMemoryFootprint is failing on travis master intermittently


WenbinZhu:
## Overview

Following failure is seen quite often for most of the PRs.

```
[ERROR] testCorfuTablePutMemoryFootprint(org.corfudb.integration.MemoryFootprintIT)  Time elapsed: 15.988 s  <<< FAILURE!
java.lang.AssertionError: 
Expecting:
 <9938512L>
to be greater than or equal to:
 <10000000L> 
	at org.corfudb.integration.MemoryFootprintIT.testCorfuTablePutMemoryFootprint(MemoryFootprintIT.java:92)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.corfudb.AbstractCorfuTest$1$1.lambda$evaluate$0(AbstractCorfuTest.java:108)
	at java.lang.Thread.run(Thread.java:748)
```

PR #1924, opened 2019-06-04
Make MemoryFootprintIT more reliable by forcing GC first

hisundar:
## Overview
Sometimes GC kicks in the middle of the loading in the test causing the after load memory use to show a smaller value than before.
Forcing GC before the test can prevent this discrepancy.
Description:

Why should this be merged: 

Related issue(s) (if applicable): #1922 

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/MemoryFootprintIT.java
+++ b/test/src/test/java/org/corfudb/integration/MemoryFootprintIT.java
@@ -73,6 +73,9 @@ public class MemoryFootprintIT extends AbstractIT {
                 .setStreamName("volbeat")
                 .open();
 
+        // Force GC first to prevent it from interfering with the size estimates after.
+        System.gc();
+
         // Register memory footprint tracking
         final Gauge<Long> corfuTableSizeGauge = MetricsUtils.addMemoryMeasurerFor(
                 CorfuRuntime.getDefaultMetrics(),


test/src/test/java/org/corfudb/integration/MemoryFootprintIT.java
```diff
@@ -73,6 +73,9 @@ public void testCorfuTablePutMemoryFootprint() throws Exception {
                 .setStreamName("volbeat")
                 .open();
 
+        // Force GC first to prevent it from interfering with the size estimates after.
+        System.gc();
```
Maithem: Btw this is just a hint. I think sizeOf is not reliable.
I think this will just make intermittent.

hisundar: do you think we should get rid of the test itself?

Maithem: Hmm, I triggered CI a couple of times and it seems to be working. Maybe its a GC timing issue?

I remember this didnt work when we used oraclejdk, maybe openjdk handles the gc hints more reliably?

hisundar: Yes this fix did seem to make MemoryFootprintIT a bit more reliable, so can we please merge this fix so other PRs have fewer failures from this test?

PR #1930, opened 2019-06-06
Migrate to a new version of Junit (Junit5)

xnull:
## Overview
Junit5 can provide additional powerful features https://developer.ibm.com/dwblog/2017/top-five-reasons-to-use-junit-5-java/

At least https://www.baeldung.com/junit-5-repeated-test

Also, we have some unit tests written for junit5, they will be added into Corfu very soon 

Description:

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,9 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <logback.classic.version>1.2.3</logback.classic.version>
+
+        <!-- Dependency versions -->
+        <junit.jupiter.version>5.4.2</junit.jupiter.version>
     </properties>
 
     <scm>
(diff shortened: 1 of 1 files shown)


lixinchengdu:
LGTM.

pom.xml
```diff
@@ -93,9 +96,27 @@
 
     <dependencies>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <groupId>org.junit.jupiter</groupId>
```
Maithem: Why does this need to be in the top level pom and not in the test module?

xnull: Because using `test` module is a bad practice, unit tests should be in their modules. Right now we have a number of unit tests outside of `test` module

Maithem: I see. Why is it bad practice ?

xnull: Unit tests inside their modules - is a recommended project structure for any standard Java project. Any build tool uses that agreement by default. 

Modularity suffers

You can't run tests for a particular module, like `./mvnw -pl infrastructure test` you can only run unit tests for all projects.

Unit tests and Integration tests sitting in the same project, if you want to run them separately you need to have custom commands for that (see our .travis.yml)

PR #1943, opened 2019-06-12
Add Log Info on Ignored TrimmedExceptions

annym:
## Overview

Adding log information for debugging.


runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -22,6 +22,7 @@
 import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.StreamOptions;
+import org.corfudb.util.Utils;
```
zalokhan: Do you need this?

annym: Removed.

PR #1921, opened 2019-06-03
LogUnit server GC phase 1: refactoring, params and add skeleton codes

WenbinZhu:
## Overview

Pull request on the feature branch:

LogUnit server GC phase 1: refactoring, extract/add parameters and add add skeleton codes for compaction.

## Checklist (Definition of Done):

- [ ] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #1945, opened 2019-06-14
StreamLogFile fixes and improvements.

WenbinZhu:
## Overview

Fixed some defects and did some minor performance improvements.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1064,13 +1061,7 @@ public void append(List<LogData> range) {
 
         // Extract all addresses associated with the provided write range.
         Set<Long> pendingWrites = range.stream()
-                .map(entry -> entry.getGlobalAddress()).collect(Collectors.toSet());
```
lixinchengdu: Why remove the check?

WenbinZhu: Because the verify() method already did the check for uniqueness.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -366,7 +363,7 @@ private void trimPrefix() {
         // (i.e. trim only complete segments)
         long endSegment = dataStore.getStartingAddress() / RECORDS_PER_LOG_FILE - 1;
```
Maithem: Can you drop the -1 and just use less than, instead of less than or equal. I think its more clear that way.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1064,13 +1061,7 @@ public void append(List<LogData> range) {
 
         // Extract all addresses associated with the provided write range.
         Set<Long> pendingWrites = range.stream()
-                .map(entry -> entry.getGlobalAddress()).collect(Collectors.toSet());
-
-        // Input validation.
-        if (pendingWrites.size() != range.size()) {
-            log.error("Input validation failed! Received entries are not unique.");
-            throw new OverwriteException(OverwriteCause.SAME_DATA);
```
Maithem: Client side also does this validation

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1195,6 +1186,7 @@ private void closeSegmentHandlers(long endSegment) {
                         sh.getSegment(), sh.getRefCount()
                 );
             }
+            channelsToSync.remove(sh.getWriteChannel());
```
Maithem: Why is this needed? `sh.close();` will flush internal channels.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
```diff
@@ -450,25 +449,24 @@ public void testGetGlobalTail() {
...
+        assertThat((long) logs.list().length).isEqualTo(numSegments);
 
         final long endSegment = 25;
         long trimAddress = endSegment * StreamLogFiles.RECORDS_PER_LOG_FILE + 1;
 
         // Get references to the segments that will be trimmed
-        Set<SegmentHandle> trimmedHandles = new HashSet();
-        for (SegmentHandle sh : ((StreamLogFiles)log).getSegmentHandles()) {
+        Set<SegmentHandle> trimmedHandles = new HashSet<>();
+        for (SegmentHandle sh : log.getSegmentHandles()) {
```
Maithem: Maybe rename this to `OpenSegments` ?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: You can refactor this into a `getEndSegment` method, it seems like its being used in multiple places

Maithem: I think its better to use getStartingSegment and keep the - 1, also make sure that you don't close the first active segment in `closeSegmentHandlers` cause an unnecessary reload.

WenbinZhu: Done, changed to getStartingSegment

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
WenbinZhu: This is to ensure that next time when sync() is called by batchProcessor, there are no closed channels in the channelsToSync map.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
WenbinZhu: changed to use getStartingSegment. In my new GC patch #1921 , I extracted the linear addressing method: address / RECORDS_PER_LOG_FILE  and cleaned up all usages of it.

PR #1946, opened 2019-06-14
Bug Fix Stream Layer Ignore Trimmed Exceptions

annym:
## Overview

Stream Layer for backpointers and address maps has a path where trimmed exceptions are thrown despite a flag set to ignore trimmed exceptions.


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -257,12 +268,12 @@ ILogData read(long address) {
...
                 // is much cheaper than the cost of a NoRollBackException, therefore
                 // this trade-off is reasonable
-                final ILogData loadedVal = fetch(address);
+                final ILogData loadedVal = fetch(address, ignoreTrimmed);
                 return cacheLoadAndGet(readCache, address, loadedVal);
             }
         return data;
     }
-        return fetch(address);
+        return fetch(address, ignoreTrimmed);
```
zalokhan: Fix the indentation here.

annym: Done.

runtime/src/main/java/org/corfudb/recovery/RecoveryUtils.java
```diff
@@ -70,7 +70,7 @@ static ILogData getLogData(CorfuRuntime runtime, boolean loadInCache, long addre
         if (loadInCache) {
             return runtime.getAddressSpaceView().read(address);
         } else {
-            return runtime.getAddressSpaceView().fetch(address);
+            return runtime.getAddressSpaceView().fetch(address, false);
```
Maithem: Why does caching determine the behavior of the fetch? 
Why not handle the trimmed exception on the higher level. 
This seems like overloaded semantics to me.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -55,6 +59,8 @@
 @Slf4j
 public class AddressSpaceView extends AbstractView {
 
+    private static final int COUNTER_INIT_VALUE = 0;
```
Maithem: Just inline this? it doesn't seem to be used outside the  `AddressSpaceView.`

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -257,12 +268,13 @@ ILogData read(long address) {
                 // load the same value), but currently a redundant RPC
                 // is much cheaper than the cost of a NoRollBackException, therefore
                 // this trade-off is reasonable
-                final ILogData loadedVal = fetch(address);
+                final ILogData loadedVal = fetch(address, ignoreTrimmed);
```
Maithem: If ignoreTrimmed is true, then this will pollute the cache with trimmed entries, right?

runtime/src/main/java/org/corfudb/recovery/RecoveryUtils.java
annym: We can't handle the trimmed exceptions in the stream layer cause we want to still read the entries (they won't be returned as a TrimmedException was thrown from the address space layer).

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: done.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: yes, you're right I'll remove the trimmed entries!

runtime/src/main/java/org/corfudb/recovery/RecoveryUtils.java
annym: Actually for the single address fetch we can handle it in upper layers, but for the readRange we can't. I'll modify it on the single read path but don't see an alternative for multi reads.

Maithem: Yes, the batch read semantics can be different,  where only the subset of addresses can be returned and the upper layers can determine if they response batch contains all the addresses in the request.

annym:
@Maithem @zalokhan I've addressed the comments, can you please revisit this PR.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -259,8 +264,13 @@ protected ILogData read(final long address, long readStartTime) {
     protected List<ILogData> readAll(@Nonnull List<Long> addresses) {
         try {
             Map<Long, ILogData> dataMap =
-                    runtime.getAddressSpaceView().read(addresses);
-            return addresses.stream().map(dataMap::get).collect(Collectors.toList());
+                    runtime.getAddressSpaceView().read(addresses, options.ignoreTrimmed);
+            // If trimmed exceptions are ignored, the data retrieved by the read API might not correspond
+            // to all requested addresses, for this reason we must filter out data entries not included (null).
+            // Also, we need to preserver ordering for checkpoint logic.
```
hisundar: typo: preserve*

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -523,12 +563,22 @@ public void invalidateClientCache() {
...
      * @return an ordered map of read addresses.
      */
     @Nonnull
-    public Map<Long, ILogData> fetchAll(Iterable<Long> addresses,
-                                        boolean waitForWrite,
-                                        boolean cacheOnServer) {
+    private Map<Long, ILogData> fetchAll(Iterable<Long> addresses,
+                                         boolean waitForWrite,
+                                         boolean cacheOnServer,
+                                         boolean cacheOnClient,
```
WenbinZhu: fetchAll() by definition should not deal with caching on client, it's the caller of it, i.e. read(), doing this, we shouldn't mix them up, right?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -554,7 +604,46 @@ public void invalidateClientCache() {
...
+
+        for (Map.Entry<Long, ILogData> entry : result.entrySet()) {
+            // Add trimmed addresses to list
+            if (!checkLogData(entry.getKey(), entry.getValue(), false)) {
+                trimmedAddresses.add(entry.getKey());
+            } else {
+                if (cacheOnClient) {
+                    // After fetching a value, we need to insert it in the cache.
+                    // Even if trimmed exceptions are thrown, we cache the valid data.
+                    result.put(entry.getKey(), cacheLoadAndGet(readCache, entry.getKey(), entry.getValue()));
```
WenbinZhu: Same as above, I think it's not good to use cacheLoadAndGet here, read() does this.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -247,7 +252,7 @@ protected ILogData read(final long address, long readStartTime) {
                 return runtime.getAddressSpaceView().read(address);
             }
             return runtime.getAddressSpaceView()
-                    .read(Collections.singleton(address), false)
+                    .read(Collections.singleton(address))
```
WenbinZhu: Why is this behavior changed? It used to be NOT waitForWrite, but now it waits instead. Is that true?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -259,8 +264,13 @@ protected ILogData read(final long address, long readStartTime) {
...
         try {
             Map<Long, ILogData> dataMap =
-                    runtime.getAddressSpaceView().read(addresses);
-            return addresses.stream().map(dataMap::get).collect(Collectors.toList());
+                    runtime.getAddressSpaceView().read(addresses, options.ignoreTrimmed);
+            // If trimmed exceptions are ignored, the data retrieved by the read API might not correspond
+            // to all requested addresses, for this reason we must filter out data entries not included (null).
+            // Also, we need to preserver ordering for checkpoint logic.
+            return  addresses.stream().map(dataMap::get)
+                    .filter(data -> data != null)
```
WenbinZhu: I think a cleaner way to do this is to change read() API,  the returned map of read() an ordered map, then you can just return the ordered value set of the `dataMap`

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -83,10 +72,30 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
...
+
+            try {
+                // The read range has a special behaviour, it attempts to read the address of interest,
+                // and in the case it is not in the cache, it reads a batch of addresses
+                // that will be further required, returning only the data for the address of interest.
+
+                // Because addresses are requested in increasing order and system relies on prefix trim,
+                // we force trimmedExceptions to be thrown by lower layers, and handle ignoreTrimmed at this layer.
+                // Note that lower layers will cache the valid entries, to optimize read performance.
+                ld = readRange(thisRead, queue, false);
```
WenbinZhu: If readRange is always used with ignoreTrimmed to be false, why not remove this parameter?
Also readRange doesn't sound like a good name, as it returns a single value.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -83,10 +72,30 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
...
+                    throw te;
+                }
+
+                log.warn("removeFromQueue[{}]: ignoring trimmed addresses {}", this, te.getTrimmedAddresses());
+                // Ignore trimmed address, remove trimmed addresses and get next from queue
+                te.getTrimmedAddresses().forEach(address -> queue.remove(address));
+                return removeFromQueue(queue);
+            }
+
+            if (queue == getCurrentContext().readQueue && ld != null) {
```
WenbinZhu: readRange return value is marked @Nonnull, `ld != null` is a redundant check.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
annym: Done

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
annym: Leftover from the previous change. Good catch, fixed it.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: I pushed it down to the fetch all, as the checkLogData (check for trimmed addresses) is the one throwing the TrimmedException and for multiple reads it is convenient to cache the valid reads before throwing back the TrimmedException (in the case the flag is not set),  or we will be going again to the log units for these set of valid addresses. I wanted to avoid requiring fetchAll + checkLogData (2 calls) and confine them through the fetchAll only (as it is also used by the nonCacheFetchAll) 

If you consider the fetchAll should not deal with caching, I can take the checkLogData out of this call and call it after the fetchAll... I. was avoiding the combination of these two...

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
annym: Fixed.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: actually this annotation is only informational, it's the caller's responsibility to check it is not null.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Yes, you're right I've changed the name and signature to be more suggestive.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
WenbinZhu: I think taking the checkLogData out of fetchAll() might not be good because there are someone else consuming fetchAll() directly like state transfer/fast loader, then you have to add the check to consumer's logic, which isn't good. Maybe we can populate TrimmedException up to the read() API and let read() catch and handle it as well?

annym: That is precisely why I left it within the fetchAll... the problem with bubbling it up to the read api is that if you throw a trimmedException you lose the valid LogData entries, unless we push them in the TrimmedException but I thought this is not really clean. Any thoughts?

WenbinZhu: In checkForTrimmedAddresses(), you will put trimmedAddress into TrimmedException, so in read() when you get the exception, you know the trimmedAddress, so then you can cache the addresses that are not trimmed, right?

PR #1957, opened 2019-06-20
Quota Check

Maithem:
## Overview
Enforce size quota on the logunit

Why should this be merged: 

- [ ] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/ResourceQuota.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A quota for a resource that can be modeled with a long.
+ *
+ * Created by Maithem on 6/20/19.
+ */
+public class ResourceQuota {
+
+    private final String name;
+    volatile long used;
```
xnull: private?

infrastructure/src/main/java/org/corfudb/infrastructure/ResourceQuota.java
```diff
@@ -0,0 +1,39 @@
...
+    private final long limit;
+
+    public ResourceQuota(String name, long limit) {
```
xnull: limit in bytes? Maybe it's better to create A class like InformationUnit for that? Something like this https://github.com/JakeWharton/byteunits
similar to TimeUnit

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1297,4 +1300,34 @@ public static int getChecksum(int num) {
             return hasher.putInt(num).hash().asInt();
         }
     }
+
+    /**
+     * Estimate the size (in bytes) of a directory.
+     * From https://stackoverflow.com/a/19869323
+     */
+    long estimateSize(Path directoryPath) {
```
zalokhan: Why not use Files.size() ?

Maithem: `Files::Size` doesn't actually return what I want (i.e. total size of all the logs in /log). We have to traverse the directory and sum the log file sizes.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1258,6 +1260,7 @@ public void reset() {
         dataStore.resetTailSegment();
         logMetadata = new LogMetadata();
         writeChannels.clear();
+        logSizeQuota.release(Long.MIN_VALUE);
```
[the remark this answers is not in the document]
Maithem: Good point. Just created a new instance.

infrastructure/src/main/java/org/corfudb/infrastructure/ResourceQuota.java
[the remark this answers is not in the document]
Maithem: Done.

True, synchronized forces a memory barrier.

Maithem: done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html

`sizeOf` ?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -897,35 +912,23 @@ private LogEntry getLogEntry(long address, LogData entry) {
...
-    private static void safeWrite(FileChannel channel, ByteBuffer buf) throws IOException {
-        long prev = channel.position();
-        try {
-            do {
-                channel.write(buf);
-            } while (buf.hasRemaining());
+    private void writeByteBuffer(FileChannel channel, ByteBuffer buf) throws IOException {
+        int bufSize = buf.remaining();
 
-        } catch (IOException e) {
```
Maithem: We have an outstanding issue where on io exceptions we need to reinitialize the logunit.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: `sizeOf` calls `sizeOfDirectory` the reason I didn't use the latter is because it can throw an exception 
if the directory is concurrently modified. https://stackoverflow.com/a/11285406

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -897,35 +912,23 @@ private LogEntry getLogEntry(long address, LogData entry) {
...
-            // any bytes that were written. This is required to avoid an ambiguous case
-            // where a subsequent write (after a failed write) succeeds but writes less
-            // bytes than the partially written buffer. In that case, the log unit can't
-            // determine if the bytes correspond to a partially written buffer that needs
-            // to be ignored, or if the bytes correspond to a corrupted metadata field.
-            channel.position(prev);
-            channel.truncate(prev);
-            channel.force(true);
-            throw e;
+        logSizeQuota.updateAndGet(quota -> {
```
xnull: I believe this step should be after `while (buf.hasRemaining()) {`

There could be a case when you updated resourceQuota but the write is failed

Maithem: If its moved after the write, then writes will only fail after they have been written. 
Logically it should work like this:
reserve quota
do operation
if operation fails return the reserved quota

But in our case, IOExceptions are bubbled all the way up, so after such exception we will need to reinitialize the LogUnit which will reinitialize the quota.

Maithem:
> This PR fails all writes to the logUnit with a QuotaExceededException. I did not see any way to bypass this. Did I miss something or is it part of another PR?

This is a hard limit, the priority and quota bypass is in the second pr.

infrastructure/src/main/java/org/corfudb/infrastructure/ResourceQuota.java
```diff
@@ -0,0 +1,43 @@
...
+    public ResourceQuota acquire(long amount) {
+        return new ResourceQuota(name, used + amount, limit);
+    }
+
+    /**
+     * Checks if the resource quota has been exceeded.
+     */
+    public void check() {
+        if (used > limit) {
+            String msg = String.format("Resource Exhausted[%s]: limit=%d, exceededSize=%d",
```
annym: exceeded size suggests it's how much size was exceeded, used - limit? maybe renaming to be more clear.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -106,12 +119,20 @@
      */
     public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
         logDir = Paths.get(serverContext.getServerConfig().get("--log-path").toString(), "log");
+        initStreamLogDirectory();
+
+        String logSizeQuotaParam = serverContext.getServerConfig().get("--log-size-quota").toString();
+        logSizeLimit = Long.parseLong(logSizeQuotaParam);
+        long initialLogSize = estimateSize(logDir);
+        logSizeQuota = new AtomicReference<>(new ResourceQuota("LogSizeQuota",
+                initialLogSize, logSizeLimit));
```
annym: On startup, if we come up in the state of exceeded quota, should we unless log that? cause maybe clients are already backing off, to understand why writes are not coming into this server.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -54,6 +54,7 @@
                     + "\tcorfu_server (-l <path>|-m) [-nsN] [-a <address>|-q <interface-name>] "
                     + "[-t <token>] [-c <ratio>] [-d <level>] [-p <seconds>] "
                     + "[--layout-server-threads=<layout_server_threads>] [--base-server-threads=<base_server_threads>] "
+                    + "[--log-size-quota=<max_log_size>]"
```
zalokhan: Duplicate. Remove this.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -96,6 +102,13 @@
     // the files of the old instance
     private LogMetadata logMetadata;
 
+    final AtomicReference<ResourceQuota> logSizeQuota;
```
zalokhan: Do we need an atomic reference if we have a single-threaded writer?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -897,35 +919,23 @@ private LogEntry getLogEntry(long address, LogData entry) {
...
-            // where a subsequent write (after a failed write) succeeds but writes less
-            // bytes than the partially written buffer. In that case, the log unit can't
-            // determine if the bytes correspond to a partially written buffer that needs
-            // to be ignored, or if the bytes correspond to a corrupted metadata field.
-            channel.position(prev);
-            channel.truncate(prev);
-            channel.force(true);
-            throw e;
+        logSizeQuota.updateAndGet(quota -> {
+            ResourceQuota newQuota = quota.acquire(bufSize);
```
zalokhan: A new object is created on every write. Would that be a concern?

Maithem:
superseded by #1974

PR #1963, opened 2019-06-24
High priority writes

Maithem:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -239,7 +240,7 @@ public void writeHeader(FileChannel fileChannel, int version, boolean verify) th
                 .build();
 
         ByteBuffer buf = getByteBufferWithMetaData(header);
-        writeByteBuffer(fileChannel, buf);
+        writeByteBuffer(fileChannel, buf, false);
```
Maithem: The priority needs to be inherited from the write operation.

runtime/src/main/java/org/corfudb/runtime/view/Priority.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ *
+ * Indicates the priority of LogUnit tasks.
+ *
+ * Created by Maithem on 6/24/19.
+ */
+@AllArgsConstructor
+public enum Priority {
+    LOW(0),
```
hisundar: Do we need 3 levels because each level can multiply the cases that we might need to handle. Can it just be 2 for now? NORMAL and HIGH?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -907,7 +908,9 @@ private LogEntry getLogEntry(long address, LogData entry) {
             }
 
             allRecordsBuf.flip();
-            writeByteBuffer(segment.getWriteChannel(), allRecordsBuf);
+            // This API is called by stateTransfer, is that considered a high priority
+            // operation
+            writeByteBuffer(segment.getWriteChannel(), allRecordsBuf, true);
```
hisundar: The comment says if checkQuota is false then it is a high priority request, but here true is passed instead?
To avoid confusion, should it just pass Priority itself instead of binding it to yet another decision variable?

runtime/src/main/java/org/corfudb/runtime/view/Priority.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: Good suggestion. Done.

test/src/test/java/org/corfudb/integration/LogSizeQuotaIT.java
```diff
@@ -0,0 +1,102 @@
...
+                quotaReached = true;
+                break;
+            }
+        }
+
+        assertThat(quotaReached).isTrue();
+
+        // Start/Commit a high priority transaction
+        rt.getObjectsView().TXBuild().priority(Priority.HIGH).build().begin();
+        map.put("k3", "k3");
```
Maithem: If an object sync encounters a hole, the hole filling mechanism won;t tag its write with high priority.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -885,7 +908,9 @@ private LogEntry getLogEntry(long address, LogData entry) {
             }
 
             allRecordsBuf.flip();
-            safeWrite(segment.getWriteChannel(), allRecordsBuf);
+            // This API is called by stateTransfer, is that considered a high priority
```
lixinchengdu: is that -> that is

lixinchengdu: Is it a high priority or normal priority? The comment and code are inconsistent.

Maithem:
superseded by #1974

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: I moved the priority to the msg level. We should take all requests with a client's priority level.

runtime/src/main/java/org/corfudb/runtime/view/Priority.java
Maithem: > Here we are coupling the priority of the writes to the quota check directly.
> Can we add a policy as a handler of this priority? The default policy could be to check the quota.

I moved the priority level to the msg level, its much cleaner now. You can see #1974

PR #1964, opened 2019-06-24
Print Server Arguments

Maithem:
## Overview

Description:
Print the server arguments when the server starts.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -216,6 +216,7 @@ public static void main(String[] args) {
             opts.put("--bind-to-all-interfaces", false);
         }
 
+        log.info("Server arguments: {}", opts);
```
hisundar: Do you want to remove the debug statement above?

Maithem: Didn't actually see that log statement. Moved it just before the server starts listening on the port because some of the opts are changed after they are parsed.

Maithem:
We already print these args.

PR #1967, opened 2019-06-25
Compaction mark in read response

lixinchengdu:
## Overview

This patch add a new field in the ReadResponse such that the runtime could get the newest compaction mark from LogUnit servers when reading data.

Description:

Why should this be merged: 

Compaction mark is the hint to the runtime about when the versioning is lost in the LogUnit Servers. With that information, the runtime could decide to abort a TX/access if the snapshot time is smaller than the compaction mark. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadResponse.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadResponse.java
@@ -4,10 +4,12 @@ import io.netty.buffer.ByteBuf;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.Getter;
+import lombok.Setter;
 
 /**
  * Created by mwei on 8/15/16.
@@ -19,12 +21,18 @@ public class ReadResponse implements ICorfuPayload<ReadResponse> {
     @Getter
     Map<Long, LogData> addresses;
 
+    @Setter
+    @Getter
+    Map<UUID, Long> compactionMarks;
+
     public ReadResponse(ByteBuf buf) {
         addresses = ICorfuPayload.mapFromBuffer(buf, Long.class, LogData.class);
+        compactionMarks = ICorfuPayload.mapFromBuffer(buf, UUID.class, Long.class);
     }
 
     public ReadResponse() {
-        addresses = new HashMap<Long, LogData>();
+        addresses = new HashMap<>();
+        compactionMarks = new HashMap<>();
     }
 
     public void put(Long address, LogData data) {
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -43,6 +44,13 @@
      */
     final CorfuRuntime runtime;
 
+    /**
+     * The compaction mark of this stream.
+     */
+    @Getter
+    @Setter
+    volatile long compactionMark = Address.NON_ADDRESS;
```
WenbinZhu: I think setCompactionMark should check if it is greater than currentCompactionMark, i.e. Math.max(current, new). Because log unit failure could regress the compactionMark returned by log unit.

runtime/src/main/java/org/corfudb/runtime/view/stream/ThreadSafeStreamView.java
```diff
@@ -112,6 +112,16 @@ public long getTotalUpdates() {
         return stream.getTotalUpdates();
     }
 
+    @Override
+    public long getCompactionMark() {
+        return stream.getCompactionMark();
+    }
+
+    @Override
+    public void setCompactionMark(long compactionMark) {
```
WenbinZhu: same here, we need to compare the current compactionMark.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
lixinchengdu: That is a good point. I have fixed it in next patch.

WenbinZhu: I think it's better be Address.MAX

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadResponse.java
```diff
@@ -19,20 +20,30 @@
...
     public ReadResponse() {
         addresses = new HashMap<Long, LogData>();
+        streamIdToCompactionMark = new HashMap<UUID, Long>();
     }
 
     public void put(Long address, LogData data) {
         addresses.put(address, data);
     }
 
+    public void putCompactionMark(UUID streamId, Long compactionMark) {
```
WenbinZhu: It's easier to consume if the parameter is  Map<UUID, Long>, not separate.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadResponse.java
```diff
@@ -19,20 +20,30 @@
     @Getter
     Map<Long, LogData> addresses;
 
+    @Getter
+    Map<UUID, Long> streamIdToCompactionMark;
```
lixinchengdu: shorter name

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ReadResponse.java
lixinchengdu: add setter

PR #1968, opened 2019-06-25
CleanUp RemoteMonitoringService and ClusterGraph

PavelZaytsev:
## Overview

Description:
Cluster state graph improvements. Unresponsive list of servers moved to the constructor.
Do not read management layout all the time in Failure Detector code.

Why should this be merged: 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -727,12 +740,9 @@ private void correctWrongEpochs(PollReport pollReport) {
      *
      * @param layoutCompletableFutureMap Map of layout server endpoints to their layout requests.
      */
-    private void updateTrailingLayoutServers(Map<String, CompletableFuture<Layout>> layoutCompletableFutureMap) {
+    private void updateTrailingLayoutServers(
+            Layout latestLayout, Map<String, CompletableFuture<Layout>> layoutCompletableFutureMap) {
 
-        // We should utilize only the unmodified management layout as it has already been committed to the layout
```
zalokhan: Please take a note of the comments added here.
We have introduced this bug earlier too and we should be extra cautious.
This layout is being modified by the sealing method in `correctWrongEpochs` and cannot be used to update the layout servers directly. We need an unmodified copy.

xnull: This code is identical to the previous version, it uses the same instance of a management layout like in the previous version. 
If we want to change this logic and use an unmodified copy of the layout then let's create a pr for that.
Not sure if we want to fix the bug in this pr.

zalokhan: The layout we are referring to is being modified in correctWrongEpoch. This is using the same layout.
We already fixed this bug in this PR #1755

xnull: I can't get your question, do you believe there is a bug in this pr or not?

xnull: So, I double checked your PR and this change, don't see any issues. Please explain if you see any issues here.

zalokhan: I am saying that **this PR** is introducing the same bug which was fixed by 1755.
The correctWrongEpochs is modifying the layout and that should not be used to update the layout servers.

xnull: If you take a look you can see that we send a modified version of the management layout, we don't use the old version of the management layout. So, the changes didn't bring new bugs and it's the same as a previous version. We just pass the layout instead of getting it from serverContext

xnull: And also it seems to me that your #1755 doesn't fix anything. Correct me if I am wrong

xnull: Ok, now I see what you are trying to say. It's really dangerous and completely unclear part of the code.

Fixed.

xnull: @zalokhan if you are ok with the changes, let's have progress, or leave your comments

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -340,11 +340,11 @@ private synchronized void runDetectionTasks() {
             // Corrects out of phase epoch issues if present in the report. This method
             // performs re-sealing of all nodes if required and catchup of a layout server to
             // the current state.
-            correctWrongEpochs(pollReport);
+            Layout updatedLayout = correctWrongEpochs(pollReport, layout);
 
             Result<DetectorTask, RuntimeException> failure = Result.of(() -> {
                 if (pollReport.isCurrentLayoutSlotUnFilled()) {
-                    detectFailure(Collections.emptySet(), pollReport).join();
+                    detectFailure(updatedLayout, Collections.emptySet(), pollReport).join();
```
zalokhan: The term updatedLayout is a bit misleading here. It might not have been updated if the seal fails.
Can we rename this? If not can we add a comment her explaining this?

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -496,11 +504,9 @@ private DetectorTask detectHealing(PollReport pollReport, Layout layout) {
      * @param pollReport Poll report obtained from failure detection policy.
      * @return boolean result if failure was handled. False if there is no failure
      */
-    private DetectorTask detectFailure(PollReport pollReport) {
+    private DetectorTask detectFailure(PollReport pollReport, Layout layout) {
```
zalokhan: There are 2 methods with the same name. Please change the names to a more intuitive name.
For example, you can name the other one as handleFailure as it's sending a message to handle the detected failures.

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -666,30 +672,43 @@ private void correctWrongEpochs(PollReport pollReport) {
...
+            return newManagementLayout;
 
         } catch (QuorumUnreachableException e) {
             log.error("Error in correcting server epochs", e);
         }
+
+        return serverContext.copyManagementLayout();
+    }
+
+    private void sealMinServerSet(Map<String, Long> wrongEpochs, Layout managementLayout) {
```
zalokhan: Can we come up with a better name?

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -304,9 +304,8 @@ private synchronized void runDetectionTasks() {
                         });
     }
 
-    private CompletableFuture<PollReport> pollReport(SequencerMetrics sequencerMetrics) {
+    private CompletableFuture<PollReport> pollReport(Layout layout, SequencerMetrics sequencerMetrics) {
         return CompletableFuture.supplyAsync(() -> {
-            Layout layout = serverContext.copyManagementLayout();
```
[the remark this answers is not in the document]
xnull: There is no difference between them. See lines 269-270

image

But anyway, added an extra step - get another copy of management layout from datastore (see line 271):
`layout = serverContext.copyManagementLayout();`

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
xnull: what name do you prefer?

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
xnull: Last time you asked exactly a vise-versa thing) 
if you remember you asked to rename that method from `handleFailure` to `detectFailure` because we have `handleFailure`:
```
return getCorfuRuntime()
                .getLayoutView()
                .getRuntimeLayout(layout)
                .getManagementClient(serverContext.getLocalEndpoint())
                .handleFailure(pollReport.getPollEpoch(), failedNodes)
                .thenApply(DetectorTask::fromBool);
```

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
xnull: I am ok with this name. Please, any Ideas, happy to rename

xnull:
@zalokhan addressed all comments and I renamed the PR title

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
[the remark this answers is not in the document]
zalokhan: There is a difference. You can see that in saveManagementLayout.

Cool. Will take a look.

xnull:
@zalokhan all your comments have addressed, lets have some progress on this change

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -666,30 +672,43 @@ private void correctWrongEpochs(PollReport pollReport) {
...
-            layout = serverContext.copyManagementLayout();
-
-            // In case of a partial seal, a set of servers can be sealed with a higher epoch.
-            // We should be able to detect this and bring the rest of the servers to this epoch.
-            long maxOutOfPhaseEpoch = Collections.max(wrongEpochs.values());
-            if (maxOutOfPhaseEpoch > layout.getEpoch()) {
-                layout.setEpoch(maxOutOfPhaseEpoch);
-            }
+            Layout newManagementLayout = latestLayout.get();
+            serverContext.saveManagementLayout(newManagementLayout);
```
zalokhan: This does not guarantee that newManagementLayout is the latest layout.
Consider the following.
- We fetch the quorum layout - `latestLayout`. But we did not receive a response from the layout servers with the latest committed layout. Hence, `latestLayout` can be old.
- We save the management layout. This does not save this layout as it has already observed a newer layout.

This statement should be followed by something like this:
`newManagementLayout = serverContext.copyManagementLayout()`

This is an optimization to not waste the current detection round.
If we do not get the latest layout, the sealing will fail which will cause this poll round to fail.

Also, we are passing this to updateTrailingServers. Although the layout server has checks to discard commits with an old epoch, we do not want to attempt to commit an old layout.

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
zalokhan: How about `sealWithLatestLayout` or `resealWithLatestKnownLayout`?

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
xnull: Let's discuss this offline and check that we don't break anything

PR #1969, opened 2019-06-26
Fix to Concurrent Checkpointers

annym:
Introduce a fix on how to resolve the checkpoint for a given stream
based on the fact that multiple checkpointers could be running at
the same time.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -2,6 +2,7 @@ package org.corfudb.runtime;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import lombok.Getter;
@@ -123,38 +124,60 @@ public class CheckpointWriter<T extends Map> {
     }
 
     /**
+     * Append a checkpoint, i.e., coalesce the state of the log. The snapshot
+     * of this checkpoint will be internally computed as to the current tail of the log.
+     *
      * @return Token at which the snapshot for this checkpoint was taken.
      */
     public Token appendCheckpoint() {
+        // Queries the sequencer for the global log tail. We then read the global log tail to
+        // persist the entry on the log unit in turn preventing from a new sequencer from
+        // regressing tokens.
+        Token snapshot = rt.getSequencerView().query().getToken();
+        if (Address.isAddress(snapshot.getSequence())) {
+            appendCheckpointOnSnapshot(snapshot);
+        }
+        return snapshot;
+    }
+
+    /**
+     * Append checkpoint at a specific snapshot. This is visible only for testing.
+     *
+     * @param snapshotTimestamp snapshot at which to take a checkpoint.
+     * @return snapshot time for this checkpoint.
+     */
+    @VisibleForTesting
+    public Token appendCheckpoint(Token snapshotTimestamp) {
+        appendCheckpointOnSnapshot(snapshotTimestamp);
+        return snapshotTimestamp;
+    }
+
+    /**
+     * Write a checkpoint which reflects the state at snapshot.
+     *
+     * @param snapshotTimestamp snapshot at which the checkpoint is taken.
+     */
+    private void appendCheckpointOnSnapshot(Token snapshotTimestamp) {
         long start = System.currentTimeMillis();
 
-        // Queries the sequencer for the global log tail. We then read the global log tail to
-        // persist the entry on the logunit in turn preventing from a new sequencer from
-        // regressing tokens.
-        Token markerToken = rt.getSequencerView().query().getToken();
-        if (Address.nonAddress(markerToken.getSequence())) {
-            return markerToken;
-        }
-        rt.getAddressSpaceView().read(markerToken.getSequence());
+        rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
         rt.getObjectsView().TXBuild()
                 .type(TransactionType.SNAPSHOT)
-                .snapshot(markerToken)
+                .snapshot(snapshotTimestamp)
                 .build()
                 .begin();
         try (Timer.Context context = MetricsUtils.getConditionalContext(appendCheckpointTimer)) {
-            Token snapshot = TransactionalContext.getCurrentContext().getSnapshotTimestamp();
             // A checkpoint writer will do two accesses one to obtain the object
             // vlo version and to get a shallow copy of the entry set
-            log.info("appendCheckpoint: Started checkpoint for {} at snapshot {}", streamId, snapshot);
+            log.info("appendCheckpoint: Started checkpoint for {} at snapshot {}", streamId, snapshotTimestamp);
             ICorfuSMR<T> corfuObject = (ICorfuSMR<T>) this.map;
             Set<Map.Entry> entries = this.map.entrySet();
             long vloVersion = corfuObject.getCorfuSMRProxy().getVersion();
-            startCheckpoint(snapshot, vloVersion);
+            startCheckpoint(snapshotTimestamp, vloVersion);
             appendObjectState(entries);
             finishCheckpoint();
             log.info("appendCheckpoint: completed checkpoint for {}, num of entries {} at snapshot {} in {} ms",
-                    streamId, entries.size(), snapshot, System.currentTimeMillis() - start);
-            return snapshot;
+                    streamId, entries.size(), snapshotTimestamp, System.currentTimeMillis() - start);
         } finally {
             rt.getObjectsView().TXEnd();
         }
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -1,6 +1,7 @@
 package org.corfudb.runtime.view.stream;
 
 import com.google.common.annotations.VisibleForTesting;
+import lombok.Data;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
@@ -19,6 +20,7 @@ import org.corfudb.runtime.view.StreamOptions;
 import org.corfudb.util.Utils;
 
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -74,7 +76,6 @@ public abstract class AbstractQueuedStreamView extends
                                       long globalAddress,
                                       ILogData ld) {
         context.resolvedQueue.add(globalAddress);
-        context.resolvedEstBytes += ld.getSizeEstimate();
 
         if (context.maxResolution < globalAddress) {
             context.maxResolution = globalAddress;
@@ -96,7 +97,7 @@ public abstract class AbstractQueuedStreamView extends
 
         // If maxGlobal is before the checkpoint position, throw a
         // trimmed exception
-        if (maxGlobal < context.checkpointSuccessStartAddr) {
+        if (maxGlobal < context.checkpoint.startAddress) {
             throw new TrimmedException();
         }
 
@@ -106,7 +107,7 @@ public abstract class AbstractQueuedStreamView extends
             getFrom = context.readCpQueue;
             // Note: this is a checkpoint, we do not need to verify it is before the trim mark, it actually should be
             // cause this is the last address of the trimmed range.
-            context.setGlobalPointer(context.checkpointSuccessStartAddr);
+            context.setGlobalPointer(context.checkpoint.startAddress);
         } else {
             getFrom = context.readQueue;
         }
@@ -280,7 +281,7 @@ public abstract class AbstractQueuedStreamView extends
     private void processTrimmedException(TrimmedException te) {
         if (TransactionalContext.getCurrentContext() != null
                 && TransactionalContext.getCurrentContext().getSnapshotTimestamp().getSequence()
-                < getCurrentContext().checkpointSnapshotAddress) {
+                < getCurrentContext().checkpoint.snapshot) {
             te.setRetriable(false);
         }
     }
@@ -303,7 +304,7 @@ public abstract class AbstractQueuedStreamView extends
 
         // If maxGlobal is before the checkpoint position, throw a
         // trimmed exception
-        if (maxGlobal < context.checkpointSuccessStartAddr) {
+        if (maxGlobal < context.checkpoint.startAddress) {
             throw new TrimmedException();
         }
 
@@ -393,7 +394,7 @@ public abstract class AbstractQueuedStreamView extends
         // any checkpoint entries, we should consult
         // a checkpoint first.
         if (context.getGlobalPointer() == Address.NEVER_READ &&
-                context.checkpointSuccessId == null) {
+                context.checkpoint.id == null) {
             // The checkpoint stream ID is the UUID appended with CP
             final UUID checkpointId = CorfuRuntime
                     .getCheckpointStreamIdFromId(context.id);
@@ -402,7 +403,7 @@ public abstract class AbstractQueuedStreamView extends
                 if (discoverAddressSpace(checkpointId, context.readCpQueue,
                         runtime.getSequencerView()
                                 .query(checkpointId).getToken().getSequence(),
-                        Address.NEVER_READ, d -> resolveCheckpoint(context, d, maxGlobal),
+                        Address.NEVER_READ, d -> scanCheckpointStream(context, d, maxGlobal),
                         true, maxGlobal)) {
                     log.trace("Fill_Read_Queue[{}] Get Stream Address Map using checkpoint with {} entries",
                             this, context.readCpQueue.size());
@@ -473,7 +474,7 @@ public abstract class AbstractQueuedStreamView extends
             return fillFromResolved(latestTokenValue, context);
         }
 
-        long stopAddress = Long.max(context.globalPointer, context.checkpointSnapshotAddress);
+        long stopAddress = Long.max(context.globalPointer, context.checkpoint.snapshot);
 
         // We check if we can fill partially from the resolved queue
         // This is a requirement for the getStreamAddressMaps as it considers the content of the resolved queue
@@ -492,7 +493,7 @@ public abstract class AbstractQueuedStreamView extends
         discoverAddressSpace(context.id, context.readQueue,
                 latestTokenValue,
                 stopAddress,
-                d -> BackpointerOp.INCLUDE, false, maxGlobal);
+                d -> true, false, maxGlobal);
 
         return !context.readCpQueue.isEmpty() || !context.readQueue.isEmpty();
     }
@@ -519,7 +520,7 @@ public abstract class AbstractQueuedStreamView extends
                                                     final NavigableSet<Long> queue,
                                                     final long startAddress,
                                                     final long stopAddress,
-                                                    final Function<ILogData, BackpointerOp> filter,
+                                                    final Function<ILogData, Boolean> filter,
                                                     final boolean checkpoint,
                                                     final long maxGlobal);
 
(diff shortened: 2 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +629,150 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
         return result == null ? Address.NOT_FOUND : result;
     }
 
-    protected BackpointerOp resolveCheckpoint(final QueuedStreamContext context, ILogData data,
-                                              long maxGlobal) {
+    private Map<UUID, StreamCheckpoint> checkpointCandidates = new HashMap<>();
```
zalokhan: Add comments.

annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -220,6 +217,10 @@ private void processCheckpoint(StreamAddressSpace streamAddressSpace, Function<I
                 checkpointResolved = processCheckpointBatchByEntry(batch, filter, queue);
             }
         }
+
+        // Select correct checkpoint - Highest
+        List<Long> checkpointEntries = selectCheckpoint(getCurrentContext());
```
medhavidhawan: what is this supposed to do ? Is it supposed to provide a list of just the highest checkpoint ?

test/src/test/java/org/corfudb/integration/ServerRestartIT.java
```diff
@@ -810,4 +811,77 @@ public void testCorfuTableMultiIndexReconstruction() throws Exception {
         runtime2.shutdown();
         runtime3.shutdown();
     }
+
+    /**
+     * This test verifies that a stream is rebuilt from a checkpoint, whenever two valid checkpoints exist, but the
```
medhavidhawan: I think this comment should say that the stream is rebuilt from the latest checkpoint (marked latest based on the snapshot it covers) even though there is an older checkpoint that appears later in the stream.

test/src/test/java/org/corfudb/integration/ServerRestartIT.java
```diff
@@ -810,4 +811,77 @@ public void testCorfuTableMultiIndexReconstruction() throws Exception {
...
+     * It also verifies that behaviour is kept the same after the node is restarted and both checkpoints
+     * still exist.
+     *
+     * 1. Write 25 entries to stream A.
+     * 2. Start a checkpoint (CP2) at snapshot 15, complete it.
+     * 3. Start a checkpoint (CP1) at snapshot 10, complete it.
+     * 4. Trim on token for CP2 (snapshot = 15).
+     * 5. New runtime instantiate stream A (do a mutation to force to load from checkpoint).
+     * 6. Restart the server
+     * 7. Instantiate map again.
```
medhavidhawan: Please also write the expectation (assertion criteria) from the test.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,19 +127,28 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
...
-        rt.getAddressSpaceView().read(markerToken.getSequence());
+        return appendCheckpointOnSnapshot(markerToken);
+    }
+
+    @VisibleForTesting
+    public Token appendCheckpoint(Token snapshot) {
+        return appendCheckpointOnSnapshot(snapshot);
+    }
+
+    private Token appendCheckpointOnSnapshot(Token snapshotTs) {
```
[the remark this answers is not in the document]

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +629,153 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
+        }
+
+        public void reset() {
+            checkpointId = null;
+            checkpointedStartAddress = Address.NEVER_READ;
+        }
+    }
+
+    /**
+     * Resolve all potential checkpoints for the given max global.
```
epalta: Do we need to resolve all the checkpoints with end address lesser than given max global? Can we return as soon as we find start address of a single checkpoint while traversing from the end?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -126,19 +127,28 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
...
         Token markerToken = rt.getSequencerView().query().getToken();
         if (Address.nonAddress(markerToken.getSequence())) {
             return markerToken;
         }
-        rt.getAddressSpaceView().read(markerToken.getSequence());
+        return appendCheckpointOnSnapshot(markerToken);
+    }
+
+    @VisibleForTesting
+    public Token appendCheckpoint(Token snapshot) {
```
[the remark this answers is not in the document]

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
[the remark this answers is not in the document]

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +629,153 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
+            if (context.checkpoint.checkpointSuccessId == null &&
                     cpEntry.getCpType() == CheckpointEntry.CheckpointEntryType.END
                     && Long.decode(cpEntry.getDict()
                     .get(CheckpointEntry.CheckpointDictKey.START_LOG_ADDRESS)) <= maxGlobal) {
                 log.trace("Checkpoint[{}] END found at address {} type {} id {} author {}",
                         this, data.getGlobalAddress(), cpEntry.getCpType(),
                         Utils.toReadableId(cpEntry.getCheckpointId()),
                         cpEntry.getCheckpointAuthorId());
-                context.checkpointSuccessId = cpEntry.getCheckpointId();
+                boolean foundHighest = currentCheckpoint.validateHighest(cpEntry.getCheckpointId(),
```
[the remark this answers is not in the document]
annym: I've added comments and renamed variables to make it clear enough.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -887,4 +966,50 @@ synchronized void seek(long globalAddress) {
...
 
+    /**
+     * Represents a contained form of a stream's checkpoint.
+     * Only with relevant information for the stream view.
+     */
+    @Data
+    static class StreamCheckpoint {
+
+        UUID checkpointSuccessId = null;
+        long checkpointSuccessStartAddr = Address.NEVER_READ;
```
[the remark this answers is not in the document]
annym: I ported old code, but I agree it's confusing. I'll rename now that we're moving it. Adding comments as well.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +629,153 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
+
+    /**
+     * From list of potential (valid) checkpoints, select the max (i.e.,
+     * the checkpoint with the highest start address).
+     *
+     * @param context this stream's current context.
+     *
+     * @return list of addresses for this checkpoint.
+     */
+    public List<Long> selectCheckpoint(final QueuedStreamContext context) {
```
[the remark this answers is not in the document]
annym: I got rid of the checkpointCandidates, and keep only the highest valid around.. so this method has changed.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +629,153 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
+            UUID successCheckpointId = currentCheckpoint.getCheckpointId();
+            log.trace("selectCheckpoint[{}]: selecting checkpoint {} with start address {}", this,
+                    successCheckpointId, currentCheckpoint.getCheckpointedStartAddress());
+            context.checkpoint = checkpointCandidates.get(successCheckpointId);
+            checkpointAddresses.addAll(checkpointCandidates.get(successCheckpointId).getCheckpointAddresses());
+        }
+
+        // Reset valid checkpoint candidates for this maxGlobal
+        checkpointCandidates = new HashMap<>();
+        currentCheckpoint.reset();
```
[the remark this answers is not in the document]
annym: this changed, removed actually this object.

test/src/test/java/org/corfudb/integration/ServerRestartIT.java
annym: Done.

test/src/test/java/org/corfudb/integration/ServerRestartIT.java
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
annym: No, we actually want to traverse all valid checkpoints, because an older checkpoint might appear later in the stream. Therefore, even this is the first we find it is not the highest one.

annym:
@medhavidhawan  I've addressed the comments already.
@zalokhan Can you please take a look after latest changes.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +628,119 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
+                        Utils.toReadableId(cpEntry.getCheckpointId()), cpEntry.getCheckpointAuthorId());
+
+                // Because checkpoint ordering is not guaranteed, i.e., a checkpoint for a lower snapshot
+                // could appear in the log after a checkpoint for a higher snapshot (case of multiple
+                // checkpointers running in parallel). We need to inspect all checkpoints and keep the one
+                // with the highest VLO version.
+                UUID checkpointId = cpEntry.getCheckpointId();
+                long checkpointVLOVersion = Long.decode(cpEntry.getDict()
+                        .get(CheckpointEntry.CheckpointDictKey.START_LOG_ADDRESS));
+                boolean isCheckpointForHighestVLOVersion = latestValidCheckpoint.validateHighest(checkpointId,
```
WenbinZhu: In name `isCheckpointForHighestVLOVersion` and `validateHighest`, `Highest` is confusing here, I think it should be `Higher` right?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +628,119 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
+                    log.trace("Checkpoint[{}] START found at address {} type {} id {} author {}",
+                            this, data.getGlobalAddress(), cpEntry.getCpType(),
                             Utils.toReadableId(cpEntry.getCheckpointId()),
                             cpEntry.getCheckpointAuthorId());
-                    return BackpointerOp.INCLUDE_STOP;
+
+                    if (cpEntry.getDict().get(CheckpointEntry.CheckpointDictKey
+                                        .SNAPSHOT_ADDRESS) != null) {
+                        latestValidCheckpoint.setCheckpointSnapshotAddress(Long.decode(cpEntry.getDict()
+                                .get(CheckpointEntry.CheckpointDictKey.SNAPSHOT_ADDRESS)));
```
WenbinZhu: On Friday we checked each checkpoint entry contains the checkpoint metadata SNAPSHOT_ADDRESS, so we can setCheckpointSnapshotAddress  when reading the END checkpoint entry and remove this block, right?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +628,119 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
-                return BackpointerOp.EXCLUDE;
             }
         }
-        return BackpointerOp.INCLUDE;
+
+        return false;
+    }
+
+    /**
+     *
```
WenbinZhu: Complete the javadoc.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -119,16 +118,11 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
-                        || op == BackpointerOp.INCLUDE_STOP) {
+                filter.apply(d);
+
+                if (!checkpoint) {
                     log.trace("followBackpointers: Adding backpointer to address[{}] to queue", currentAddress);
                     queue.add(currentAddress);
-                    entryAdded = true;
-                    // Check if we need to stop
-                    if (op == BackpointerOp.INCLUDE_STOP) {
-                        return entryAdded;
```
WenbinZhu: You can remove this variable now, it always false.

annym: Done. Actually returning the state of a queue cause entries could have been added before hitting a trimmed exception in back pointer implementation.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
annym: Right! Removed.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
annym: Sure, if it's more intuitive. Also I added comments to the method.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -887,4 +931,73 @@ synchronized void seek(long globalAddress) {
...
+    /**
+     * Represents a contained form of a stream's checkpoint.
+     * Only with relevant information for the stream view.
+     */
+    @Data
+    static class StreamCheckpoint {
+
+        UUID checkpointId = null;
+        // Represents the VLO version at the time of checkpoint
+        long checkpointedStartAddress = Address.NEVER_READ;
```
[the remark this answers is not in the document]
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -887,4 +931,73 @@ synchronized void seek(long globalAddress) {
...
+        // Represents the address of the END record of this checkpoint
+        long checkpointEndAddress = Address.NEVER_READ;
+        // Total number of entries in this checkpoint
+        long checkpointNumEntries = 0L;
+        // Total number of Bytes in this checkpoint
+        long checkpointBytes = 0L;
+
+        long resolvedEstBytes = 0L;
```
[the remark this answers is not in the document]
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -887,4 +931,73 @@ synchronized void seek(long globalAddress) {
...
+
+        public StreamCheckpoint() {
+        }
+
+        public void addAddress(long address) {
+            checkpointAddresses.add(address);
+        }
+
+        public boolean validateHighest(UUID id, long vloVersion) {
+            if (Address.isAddress(vloVersion) && vloVersion > checkpointedStartAddress) {
```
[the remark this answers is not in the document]
annym: That was my initial idea, but snapshots are not serialized in checkpoints, and this would require data migration.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -123,38 +124,60 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
...
+        appendCheckpointOnSnapshot(snapshotTimestamp);
+        return snapshotTimestamp;
+    }
+
+    /**
+     * Write a checkpoint which reflects the state at snapshot.
+     *
+     * @param snapshotTimestamp snapshot at which the checkpoint is taken.
+     */
+    private void appendCheckpointOnSnapshot(Token snapshotTimestamp) {
```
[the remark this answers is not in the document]
annym: The reason to change it is that the value being returned was the actual value being passed, so it had no special meaning nor it was changing. 

Yes, I agree. This is old code not sure why there is no error handling. We'd need to agree either if throwing an exception or returning an empty token (-1).

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -626,52 +628,119 @@ public synchronized long find(long globalAddress, SearchDirection direction) {
...
-                        cpEntry.getCheckpointAuthorId());
-                context.checkpointSuccessId = cpEntry.getCheckpointId();
+                        Utils.toReadableId(cpEntry.getCheckpointId()), cpEntry.getCheckpointAuthorId());
+
+                // Because checkpoint ordering is not guaranteed, i.e., a checkpoint for a lower snapshot
+                // could appear in the log after a checkpoint for a higher snapshot (case of multiple
+                // checkpointers running in parallel). We need to inspect all checkpoints and keep the one
+                // with the highest VLO version.
+                UUID checkpointId = cpEntry.getCheckpointId();
+                long checkpointVLOVersion = Long.decode(cpEntry.getDict()
```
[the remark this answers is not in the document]
annym: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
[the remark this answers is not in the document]
annym: I think we should find a more suitable naming or some other prefix that suggests a difference. endAddress refers to the end address of a checkpoint stream, while 'startAddress' is not the checkpoint start address but instead refers to the  last update observed on the stream being checkpointed at the time of snapshot (I know it's confusing but it was already under those names)

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -887,4 +931,73 @@ synchronized void seek(long globalAddress) {
...
+     * Only with relevant information for the stream view.
+     */
+    @Data
+    static class StreamCheckpoint {
+
+        UUID checkpointId = null;
+        // Represents the VLO version at the time of checkpoint
+        long checkpointedStartAddress = Address.NEVER_READ;
+        // Represents the address of the END record of this checkpoint
+        long checkpointEndAddress = Address.NEVER_READ;
```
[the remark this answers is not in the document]
annym: adding to the previous comment, we can actually get rid of this field. It is not being used and just generates confusion. In that case, I'll rename to startAddress and clearly state in the comments it's meaning.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -887,4 +931,73 @@ synchronized void seek(long globalAddress) {
...
+        long checkpointNumEntries = 0L;
+        // Total number of Bytes in this checkpoint
+        long checkpointBytes = 0L;
+
+        long resolvedEstBytes = 0L;
+        /** The address the current checkpoint snapshot was taken at.
+         *  The checkpoint guarantees for this stream there are no entries
+         *  between checkpointedStartAddress and checkpointSnapshotAddress.
+         */
+        long checkpointSnapshotAddress = Address.NEVER_READ;
```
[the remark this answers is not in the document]
annym: done.

PR #1970, opened 2019-06-26
Unify Fill Hole Write Path

Maithem:
## Overview
Hole fill writes are no different than the regular writes, so they
should use the same write path.

Why should this be merged: Removes duplicate code

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -211,9 +210,9 @@ void handleException(Throwable ex, ChannelHandlerContext ctx, CorfuPayloadMsg ms
      */
     @ServerHandler(type = CorfuMsgType.WRITE)
     public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.debug("log write: global: {}, streams: {}", msg.getPayload().getToken(),
-                msg.getPayload().getData().getBackpointerMap());
         LogData logData = (LogData) msg.getPayload().getData();
+        log.debug("log write: type: {}, address: {}, streams: {}", logData.getType(),
```
xnull: Are you sure?
It seems to me that writing to log the fact that the server just gets a message is not so good idea. 
Maybe log the fact that the message has written into the log is more useful, but for me is still questionable, because we are going to have millions of useless lines of log messages.

Maithem: Im not adding this line, just editing this log statement.

hisundar:
@annym need help in merging this to 0.2.3 branch. thanks

PR #1971, opened 2019-06-26
[Do not review] Bug Fix Stream Address Space Rebuilt Log Unit

annym:
On server restart, stream's address spaces are rebuilt (stream addresses and trim mark). A stream's trim mark was being initialized with the value NON_EXIST (-6) if it appeared before an actual checkpoint entry, and computed as the min of all backpointers. Because it was initialized with -6, this would always be the minimum value. This would make stream trim marks hold the wrong value for cases in which regular entries appear before a checkpoint, until a next trim comes in. This commit corrects this bug, setting the proper trim mark and a test is included.

This change also considers the case of multiple checkpoints running at the same time for address space rebuilt.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
@@ -49,45 +49,23 @@ public class LogMetadata {
     }
 
     public void update(LogData entry) {
-        update(entry, false, Address.NON_ADDRESS);
-    }
-
-    public void update(LogData entry, boolean initialize, long globalTrimMark) {
         long entryAddress = entry.getGlobalAddress();
         updateGlobalTail(entryAddress);
         for (UUID streamId : entry.getStreams()) {
             // Update stream tails
             long currentStreamTail = streamTails.getOrDefault(streamId, Address.NON_ADDRESS);
             streamTails.put(streamId, Math.max(currentStreamTail, entryAddress));
-
             // Update stream address map (used for sequencer recovery)
             // Since entries might have been written in random order
             // We update the trim mark to be the min of all backpointer addresses.
-            streamsAddressSpaceMap.computeIfAbsent(streamId, k ->
-                    new StreamAddressSpace(Address.NON_EXIST, new Roaring64NavigableMap()));
-
             streamsAddressSpaceMap.compute(streamId, (id, addressSpace) -> {
                 // If restarting the log unit, i.e., scanning all records in the log for initialization,
-                // update the stream trim mark as we read (as entries might not be ordered).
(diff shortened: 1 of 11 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
```diff
@@ -74,10 +71,10 @@ public void update(LogData entry, boolean initialize, long globalTrimMark) {
                 // states of the actual address map. In the case of log updates, the trim mark will be set
                 // as prefix trims are performed.
                 if (addressSpace == null) {
-                    Long streamTrimMark = Address.NON_EXIST;
+                    long streamTrimMark = Address.NON_EXIST;
```
WenbinZhu: If a stream is opened first time with a log write (not initialize), would it be better to use globalTrimMark for its streamTrimMark instead of -1? (Maybe less trimmedException here: https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java#L160 ? Not sure...)

test/src/test/java/org/corfudb/integration/ServerRestartIT.java
```diff
@@ -810,4 +812,64 @@ public void testCorfuTableMultiIndexReconstruction() throws Exception {
         runtime2.shutdown();
         runtime3.shutdown();
     }
+
+    /**
+     * Check streamAddressSpace rebuilt from log unit
+     * when stream has been previously checkpointed and trimmed.
+     **/
+    @Test
+    public void checkStreamAddressSpaceRebuiltWithTrim() throws Exception {
```
WenbinZhu: This test could pass in current master branch. When we introduce a new testcase for a bug fix, we should come up with a testcase that fails with current implementation, but could succeed with the fix, otherwise we cannot be sure if the fix really works, right?

annym: Yes, I had written a test on top of my PR 1969 and thought this would replicate it, but we need entries to be placed in between the snapshot time and the actual occurrence of the checkpoint entries. So, I've rebased on top of 1969 to have the API for option. Still I'll leave this test to catch any regression.

infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
annym: That would actually lead to more trimmedExceptions, many of them false. Think of these scenarios:

1. That stream didn't exist by the time the trim happened, hence there are no checkpoints for it. We would think the trim mark is global trim mark and fail to find a cp that covers that space.
2. Because we don't persist snapshot address in our checkpoint entries we would also fail.

annym:
Superseded by #1979

PR #1972, opened 2019-06-26
Add a CorfuRuntimeParameter write priority

hisundar:
## Overview

Description:
This allows clients to identify themselves as a "high" priority client so that its writes are not affected by quota limitations placed on the corfu server. Sort of a QoS.

Why should this be merged: 

- [x] Public API has Javadoc


test/src/test/java/org/corfudb/integration/LogSizeQuotaIT.java
```diff
@@ -99,4 +99,58 @@ public void testLogSizeQuota() throws Exception {
...
+
+        // Create a map
+        map.put("k1", "v1");
+        map.put("k2", "v2");
+
+        // Fill the log with data till the quota is reached
+        IStreamView sv = rt.getStreamsView().get(UUID.randomUUID());
+
+        boolean quotaReached = false;
+        for (int x = 0; x < numWrites; x++) {
```
annym: Since it gets right to the limit of the quota, should you afterwards also exceed the quota to make the test a bit more complete?

hisundar: it is part of testLogSizeQuota() but good point let me make that change. thanks

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -183,8 +184,8 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
                     // Large writes are also rejected right away.
                     throw ex;
                 }
-            } catch (WriteSizeException we) {
-                throw we;
+            } catch (WriteSizeException | QuotaExceededException ie) {
+                throw ie;
```
Maithem: Sleep here?

hisundar: Interesting suggestion I was thinking of sleeping in the retry logic up in the caller, but I think we can add a sleep here too.

hisundar:
superseded by https://github.com/CorfuDB/CorfuDB/pull/1974

PR #1977, opened 2019-06-27
Change Log Level for Trim messages in Stream View

annym:
## Overview

Modify log levels for trimmed messages that are not really critical until they lead to a TAE which is already logged as an error. It is common for trim message "warnings" to happen once before a valid checkpoint is loaded.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
@@ -161,11 +161,12 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
                     String message = String.format("getStreamAddressMap[{%s}] stream has been " +
                                     "trimmed at address %s and this space is not covered by the " +
                                     "loaded checkpoint with start address %s, while accessing the " +
-                                    "stream at version %s", this, streamAddressSpace.getTrimMark(),
+                                    "stream at version %s. Looking for a new checkpoint.", this,
+                            streamAddressSpace.getTrimMark(),
                             getCurrentContext().checkpoint.startAddress, maxGlobal);
-                    log.warn(message);
+                    log.info(message);
                     if (options.ignoreTrimmed) {
-                        log.warn("getStreamAddressMap[{}]: Ignoring trimmed exception for address[{}].",
+                        log.debug("getStreamAddressMap[{}]: Ignoring trimmed exception for address[{}].",
                                 this, streamAddressSpace.getTrimMark());
                     } else {
                         throw new TrimmedException(message);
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -161,11 +161,12 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
                                     "loaded checkpoint with start address %s, while accessing the " +
-                                    "stream at version %s", this, streamAddressSpace.getTrimMark(),
+                                    "stream at version %s. Looking for a new checkpoint.", this,
+                            streamAddressSpace.getTrimMark(),
                             getCurrentContext().checkpointSuccessStartAddr, maxGlobal);
-                    log.warn(message);
+                    log.info(message);
                     if (options.ignoreTrimmed) {
-                        log.warn("getStreamAddressMap[{}]: Ignoring trimmed exception for address[{}].",
+                        log.debug("getStreamAddressMap[{}]: Ignoring trimmed exception for address[{}].",
```
epalta: Did you figure out why we were seeing the huge number at the end of the message ?
2019-06-27T03:26:49.390Z WARN CorfuDB-DCN-Publisher-0 AddressMapStreamView - getStreamAddressMap[{ae2a@448367}] stream has been trimmed at address 407912 and this space is not covered by the loaded checkpoint with start address -1, while accessing the stream at version 9223372036854775807

annym: Yes, that is when we do remaining() which goes up to Long.MAX

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -250,7 +251,7 @@ private boolean processCheckpointBatchByEntry(List<Long> batch,
         } catch (TrimmedException ste) {
             // The valid checkpoint has been trimmed.
             if (options.ignoreTrimmed) {
-                log.warn("processCheckpoint[{}]: Ignoring trimmed exception for address[{}]," +
+                log.debug("processCheckpoint[{}]: Ignoring trimmed exception for address[{}]," +
```
medhavidhawan: In the case of ignoreTrimmed is this misleading ?

annym: That's why we are changing the log level. As it only ignores it if it was explicitly set to be ignored. Shouldn't then be something to worry about or notify. If ignored trimmed flag is set,  this stream is never reset to load from a checkpoint which means that this message is likely to appear a lot of times.

PR #1978, opened 2019-06-28
Add Checkpoint Metadata

annym:
Include snapshot address and start address as part of the END
record metadata. This information will be used to ease stream's
address space rebuilt on log units.

### DIFF
--- a/format/proto/types.proto
+++ b/format/proto/types.proto
@@ -35,6 +35,8 @@ message LogEntry {
     optional int64 checkpointedStreamId_least_significant = 14;
     //  Tail of the stream at the time of taking the checkpoint snapshot.
     optional int64 checkpointedStreamStartLogAddress = 15;
+    optional int64 checkpointStartAddress = 19;
+    optional int64 checkpointSnapshotAddress = 20;
 
     // ClientId is the Corfu runtime id that created this LogEntry
     optional int64 clientId_least_significant = 16;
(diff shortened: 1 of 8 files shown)


format/proto/types.proto
```diff
@@ -35,6 +35,8 @@ message LogEntry {
     optional int64 checkpointedStreamId_least_significant = 14;
     //  Tail of the stream at the time of taking the checkpoint snapshot.
     optional int64 checkpointedStreamStartLogAddress = 15;
+    optional int64 checkpointStartAddress = 19;
```
medhavidhawan: could you please add comments as to what these mean.

runtime/src/main/java/org/corfudb/protocols/logprotocol/CheckpointEntry.java
```diff
@@ -47,10 +47,11 @@ public byte asByte() {
     public enum CheckpointDictKey {
         START_TIME(0),
         END_TIME(1),
-        START_LOG_ADDRESS(2),
+        VLO_VERSION(2),
```
medhavidhawan: If possible could you please add comments for these.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -511,6 +514,9 @@ private void writeCheckpointRecords(UUID streamId, String checkpointAuthor, UUID
         if (write3) {
             CheckpointEntry cp3 = new CheckpointEntry(CheckpointEntry.CheckpointEntryType.END,
                     checkpointAuthor, checkpointId, streamId, mdKV, null);
+            mdKV.put(CheckpointEntry.CheckpointDictKey.VLO_VERSION, Long.toString(addr1 + 1));
+            mdKV.put(CheckpointEntry.CheckpointDictKey.SNAPSHOT_ADDRESS, Long.toString(addr1 + 1));
```
medhavidhawan: how do we check that these values got written.

annym:
This approach will require a strategy for data migration, which is not a path to explore at this moment.

Issue #1980, opened 2019-06-28
Size Base Eviction


Maithem:
**Feature request.**
The `AddressSpaceView` cache evicts based on total number of entries. This means that the cache needs to be tuned via profiling and estimating in-memory objects, its a tedious and fragile process.  We should provide an option to specify eviction based on total memory used.

PR #1981, opened 2019-06-28
Memory Size Based Eviction

Maithem:
## Overview
Added weight(i.e. in-memory size) based eviction for the
AddressSpaceView cache.

Why should this be merged: Makes it easier to tune the cache for some workloads 

Related issue(s) (if applicable): #1980

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -152,10 +152,16 @@ public class CorfuRuntime {
         boolean cacheDisabled = false;
 
         /**
-         * The maximum size of the cache, in bytes.
+         * The maximum number of entries in the cache.
          */
         @Default
-        long numCacheEntries = 5000;
+        long maxCacheEntries;
+
+        /**
+         * The max in-memory size of the cache in bytes
+         */
+        @Default
+        long maxCacheWeight;
 
         /**
          * Sets expireAfterAccess and expireAfterWrite in seconds.
@@ -1178,19 +1184,6 @@ public class CorfuRuntime {
     public CorfuRuntime setHoleFillingDisabled(boolean disable) {
         log.warn("setHoleFillingDisabled: Deprecated, please set parameters instead");
         parameters.setHoleFillingDisabled(disable);
-        return this;
-    }
-
-    /**
-     * Set the number of cache entries.
-     *
-     * @param numCacheEntries The number of cache entries.
-     * @deprecated Deprecated, set using {@link CorfuRuntimeParameters} instead.
-     */
-    @Deprecated
-    public CorfuRuntime setNumCacheEntries(long numCacheEntries) {
-        log.warn("setNumCacheEntries: Deprecated, please set parameters instead");
-        parameters.setNumCacheEntries(numCacheEntries);
         return this;
     }
 
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -31,6 +31,7 @@ import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
 import org.corfudb.util.CFUtils;
 import org.corfudb.util.CorfuComponent;
+import org.corfudb.util.MetricsUtils;
 import org.corfudb.util.Sleep;
 import org.corfudb.util.Utils;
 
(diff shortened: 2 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -62,19 +63,32 @@
...
+
+        CacheBuilder cacheBuilder = CacheBuilder.newBuilder();
+
+        if (runtime.getParameters().getMaxCacheEntries() != 0) {
+            cacheBuilder.maximumSize(runtime.getParameters().getMaxCacheEntries());
+        }
+
+        if (runtime.getParameters().getMaxCacheWeight() != 0) {
+            cacheBuilder.maximumWeight(runtime.getParameters().getMaxCacheWeight());
+            cacheBuilder.weigher((k, v) -> (int) (MetricsUtils.sizeOf.deepSizeOf(k)
```
Maithem: Probably can hardcode the size of the key, it would be cheaper.

xnull:
Can we have a microbenchmark test for that?
Also, let's investigate what's the performance: findbugs SizeOf vs native Java instrumentation https://www.baeldung.com/java-size-of-object

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -62,19 +63,32 @@
...
      * A cache for read results.
      */
-    final Cache<Long, ILogData> readCache = CacheBuilder.newBuilder()
-            .maximumSize(runtime.getParameters().getNumCacheEntries())
-            .expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
-            .expireAfterWrite(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
-            .removalListener(this::handleEviction)
-            .recordStats()
-            .build();
+    final Cache<Long, ILogData> readCache;
```
lixinchengdu: should readCache be private?

Maithem:
> Can we have a microbenchmark test for that?
> Also, let's investigate what's the performance: findbugs SizeOf vs native Java instrumentation https://www.baeldung.com/java-size-of-object

From the link

> This approach only supports size estimation of the considered object itself and not the sizes of objects it references. To estimate a total size of the object, we would need a code that would go over those references and calculate the estimated size.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Maithem: Its the default qualifier.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -1175,8 +1181,8 @@ public CorfuRuntime setHoleFillingDisabled(boolean disable) {
      */
     @Deprecated
     public CorfuRuntime setNumCacheEntries(long numCacheEntries) {
```
epalta: Is it possible to change name of set method as well? setMaxCacheEntries?

test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
```diff
@@ -43,6 +45,52 @@ private void setupNodes() {
...
+        CorfuRuntime rt = CorfuRuntime.fromParameters(params)
+                .parseConfigurationString(getDefaultConfigurationString())
+                .connect();
+
+        assertThatThrownBy(() -> rt.getAddressSpaceView())
+                .isInstanceOf(IllegalStateException.class);
+    }
+
+    @Test
+    public void testAddressSpaceCache() throws Exception {
```
medhavidhawan: Please add comments.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: This is deprecated, removed it.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: Can we do this?

Maithem: Done.

test/src/test/java/org/corfudb/runtime/view/AddressSpaceViewTest.java
Maithem: done.

PR #1988, opened 2019-06-28
MultiCheckpointWriter Bug Fix + More Logging

Maithem:
## Overview
The multicheckpointwriter was ending a transaction without starting
one, which this patch removes. Also, added extra stats(i.e. in-memory
size, CP size in bytes)  logging for the checkpoint writer.

Why should this be merged: Bug fix + more stat logging

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -10,7 +10,9 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.MultiSMREntry;
 import org.corfudb.protocols.logprotocol.SMREntry;
+import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.runtime.object.ICorfuSMR;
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -131,14 +133,14 @@ public Token appendCheckpoint() {
         // Queries the sequencer for the global log tail. We then read the global log tail to
         // persist the entry on the logunit in turn preventing from a new sequencer from
         // regressing tokens.
-        Token markerToken = rt.getSequencerView().query().getToken();
-        if (Address.nonAddress(markerToken.getSequence())) {
-            return markerToken;
+        TokenResponse tokenResponse = rt.getSequencerView().next();
```
lixinchengdu: Introduce a hole?

zalokhan: Some tests might be failing due to this change too. Tests assume no token is reserved for trimming.
Is there a specific reason for this change?

Maithem: Nope, its not because of that. Tests are failing because of the `MultiCheckpointWriter.java` change

I think when I did this change, we were just take the current tail (not a new token), so it was possible to generate a hole on a stream, which is not good.

Maithem: yes. Will change it to stream tail.

PR #1994, opened 2019-07-02
Revamp secondary index logic

vjeko:
This patch simplifies the logic associated with secondary index
extraction. It also makes failures that arise due to improper index
function implementation more explicit.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -28,7 +28,6 @@ import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
 
-import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.annotations.Accessor;
@@ -98,43 +97,29 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
      */
     public static class Index<K, V, I extends Comparable<?>> {
         private final CorfuTable.IndexName name;
-        private final CorfuTable.IndexFunction<K, V, I> indexFunction;
-        private final CorfuTable.MultiValueIndexFunction<K, V, I> multiValueIndexFunction;
+        private final CorfuTable.MultiValueIndexFunction<K, V, I> indexFunction;
 
-        // A flag representing whether a single indexer or a multi indexer
-        private final boolean monoIndex;
 
         public Index(CorfuTable.IndexName name, CorfuTable.IndexFunction<K, V, I> indexFunction) {
             this.name = name;
-            this.indexFunction = indexFunction;
-            this.multiValueIndexFunction =
+            this.indexFunction =
                     (k, v) -> Collections.singletonList(indexFunction.apply(k, v));
-            monoIndex = true;
         }
 
         public Index(CorfuTable.IndexName name,
                      CorfuTable.MultiValueIndexFunction<K, V, I> indexFunction) {
             this.name = name;
-            this.indexFunction = (k, v) -> indexFunction.apply(k, v).iterator().next();
-            this.multiValueIndexFunction = indexFunction;
-            monoIndex = false;
+            this.indexFunction = indexFunction;
         }
 
         public CorfuTable.IndexName getName() {
             return name;
         }
 
-        public CorfuTable.IndexFunction<K, V, I> getIndexFunction() {
+        public CorfuTable.MultiValueIndexFunction<K, V, I> getMultiValueIndexFunction() {
             return indexFunction;
         }
 
-        public CorfuTable.MultiValueIndexFunction<K, V, I> getMultiValueIndexFunction() {
-            return multiValueIndexFunction;
-        }
-
-        private boolean isMonoIndex() {
-            return monoIndex;
-        }
 
         @Override
         public boolean equals(Object o) {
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -761,36 +751,31 @@ public V merge(K key, V value, BiFunction<? super V, ? super V, ? extends V>
      */
     @SuppressWarnings("unchecked")
     protected void unmapSecondaryIndexes(K key, V value) {
+        if (value == null) {
+            return;
```
xnull: Please add a log message

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -802,33 +787,39 @@ protected void unmapSecondaryIndexes(K key, V value) {
      */
     @SuppressWarnings("unchecked")
     protected void mapSecondaryIndexes(K key, V value) {
+        if (value == null) {
+            return;
```
xnull: please add a log message

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -802,33 +787,39 @@ protected void unmapSecondaryIndexes(K key, V value) {
...
-                        slot.put(key, value);
-                    } else {
-                        for (Comparable<?> indexKey
-                                : index.getMultiValueIndexFunction().apply(key, value)) {
-                            Map<K, V> slot = secondaryIndex
-                                    .computeIfAbsent(indexKey, k -> new HashMap<>());
-                            slot.put(key, value);
-                        }
-                    }
+            // Map entry into secondary indexes
```
xnull: this method and the prev. one looks like a copy-paste. Is it possible to get rid of copy paste here?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -802,33 +787,39 @@ protected void unmapSecondaryIndexes(K key, V value) {
...
+            for (Index<K, V, ? extends Comparable> index : indexSpec) {
+                String indexName = index.getName().get();
+                Map<Comparable, Map<K, V>> secondaryIndex = secondaryIndexes.get(indexName);
+                for (Comparable indexKey : index.getMultiValueIndexFunction().apply(key, value)) {
+                    Map<K, V> slot = secondaryIndex.computeIfAbsent(indexKey, k -> new HashMap<>());
```
xnull: Is it secure to use hashmap here?

vjeko: Not clear. secondaryIndex and mainMap might suffer from the same concurrency problem, but before I do anything about it, I really have to be able to prove it.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -98,43 +98,33 @@
...
+            this.indexFunction = indexFunction;
         }
 
         public CorfuTable.IndexName getName() {
             return name;
         }
 
         public CorfuTable.IndexFunction<K, V, I> getIndexFunction() {
-            return indexFunction;
+            return (k, v) -> indexFunction.apply(k, v).iterator().next();
```
xnull: Is it possible that `iterator` would be empty? If yes then you could get NPE

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -761,36 +751,32 @@ public V merge(K key, V value, BiFunction<? super V, ? super V, ? extends V>
...
             }
         } catch (Exception e) {
-            indexSpec.clear();
-            secondaryIndexes.clear();
-            indexGenerationFailed = true;
-            log.error("unmapSecondaryIndexes: Exception unmapping {}, {},"
-                            + " UNMAPPING ALL INDEXES, indexing is disabled",
-                    key, value, e);
+            log.error("Received an exception while computing the index. " +
+                    "This is most likely an issue with the client's indexing function. {}", e);
```
xnull: Please delete `{}`
image

xnull: You can install `sonar` plugin into IDEA, it will highlight all that kind of issues

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -802,33 +788,40 @@ protected void unmapSecondaryIndexes(K key, V value) {
...
         }
     }
 
+    /**
+     *  Disable all seconday indecies for this table. Only used during error-recovery.
+     */
+    protected void clearIndex() {
+        indexSpec.clear();
+        secondaryIndexes.clear();
+        indexGenerationFailed = true;
```
xnull: should `indexGenerationFailed` be volatile or AtomicBoolean?

xnull: Also what about `indexSpec` and `secondaryIndexes` ConcurrentSet and ConcurrentMap respectively?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -761,36 +751,32 @@ public V merge(K key, V value, BiFunction<? super V, ? super V, ? extends V>
      */
     @SuppressWarnings("unchecked")
     protected void unmapSecondaryIndexes(K key, V value) {
+        if (value == null) {
+            log.trace("Attempting to build an index with a null value, skipping.");
```
zalokhan: I think these can be warn level logs.
Trace is disabled by default on the clients.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -802,33 +788,40 @@ protected void unmapSecondaryIndexes(K key, V value) {
      */
     @SuppressWarnings("unchecked")
     protected void mapSecondaryIndexes(K key, V value) {
+        if (value == null) {
+            log.trace("Attempting to build an index with a null value, skipping.");
```
zalokhan: Warn level?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Oh, cool.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Good point.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: indexSpec and secondaryIndexes are actually fine, since their mutation is tied to CorfuTable construction.

I don't think that indexGenerationFailed is used at all. I'll remove it.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Yep.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
xnull: ok

vjeko:
Also, removed some of the dead code. I made sure it is not being used upstream.

PR #2005, opened 2019-07-04
Add timestamp into layout

xnull:
## Overview

Description:
It's very useful for having a timestamp in a layout. Helps in debugging a lot.

Why should this be merged: 

- [x] Public API has Javadoc


hisundar:
Thanks @xnull, just curious can this break upgrades/downgrade where the file does/doesn't have the timestamp field?

xnull:
@hisundar great question, thanks. I discovered that how we serialize/deserialize layouts is a pretty bizarre approach.
So, I made some cleanups and add a test to check that `timestamp` doesn't break anything

zalokhan:
@xnull I don't think adding wall clock time into the layout is a good idea. There are some drawbacks to this which I can think of. If there is clock skew between the nodes (It is not uncommon to have a clock skew between nodes), there is no reliability of this timestamp. Also, the commit time of this layout could be different than the proposed or creation time of this layout.

xnull:
@zalokhan 
1. having a timestamp in a layout doesn't require you to do anything, it's for information.
2. We can say that almost always the time is the same on different nodes, even if not, the timestamp in a layout doesn't affect logic. And it's even better, having different layouts with different timestamps will help figure out that time is not synchronized on different nodes, which is a pretty big issue and must be fixed.
3. We want to know when the particular layout was created, it's ok if a node saves a new layout with some latency (and that latency usually very small - milliseconds).

Now, disadvantages when we don't have a timestamp in the layout:
You always have to figure out when the layout was created, and it's a very annoying and manual process.

medhavidhawan:
I too do not see the advantages. I do see that wall clock time is not
useful in our protocol why don’t you make a tool to read the layouts and
creation time. Why put it inside a layout where it has no significance and
can be misconstrued over time

xnull:
@medhavidhawan 
>>You always have to figure out when the layout was created, and it's a very annoying and manual process.

it's such a big pain, anyone who ever worked with layouts experienced that pain. You always waste your time on tracing - when the layout was updated. It's a very strong reason for having a timestamp in a layout. It brings much more clearness

apavan:
I see arguments both ways for and against this change. 
Just curious, could we not use the timestamp from the corfu server log that has the info about the healed layout for any tools that we are building to make debugging easier? 
Also, if you need to make manual debugging easier then add this info to the file name as LAYOUTS_1_<timestamp>.ds . This way the files can always be ordered with our numbering as well as has the UTC timestamp for convenience.

medhavidhawan:
What is the problem with making a separate tool. Please do not pollute the
layout just for some minor convenience that can be achieved via a tool

PR #2008, opened 2019-07-08
Convert Recursion to Iteration

annym:
Convert removeFromQueue from recursive to iterative to avoid StackOverflow
exceptions in cases where holes are predominant in a stream.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
@@ -62,57 +62,62 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
 
     @Override
     protected ILogData removeFromQueue(NavigableSet<Long> queue) {
-        // Because the discovery mechanism implemented by this class
-        // does not require to read log entries in advance (requests
-        // full address map in a single call), entries can be read in
-        // batches whenever we have a cache miss. This allows next reads
-        // to be serviced immediately, rather than reading one entry at a time.
-        Long thisRead = queue.pollFirst();
+        boolean readNext;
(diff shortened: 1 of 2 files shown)


hisundar:
Thanks Anny, should we also consider removing the recursion from BackpointerStreamView.java where we seem to run into a similar loop in removeFromQueue() in case of a trimmed exception?

annym:
> Thanks Anny, should we also consider removing the recursion from BackpointerStreamView.java where we seem to run into a similar loop in removeFromQueue() in case of a trimmed exception?

I thought of it, but Backpointers was only kept if anything went wrong in the transition to address maps. It is cleaner if we do t in case the flag needs to be switched at any point. I'll do the change as well!

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -62,57 +62,61 @@ public AddressMapStreamView(final CorfuRuntime runtime,
...
+                    // Validate that the data entry belongs to this stream, otherwise, skip.
+                    // This verification protects from sequencer regression (tokens assigned in an older epoch
+                    // that were never written to, and reassigned on a newer epoch)
+                    if (ld.containsStream(this.id)) {
+                        addToResolvedQueue(getCurrentContext(), currentRead, ld);
+                        readNext = false;
+                    } else {
+                        log.trace("getNextEntry[{}]: the data for address {} does not belong to this stream. Skip.",
+                                this, currentRead);
+                        readNext = true;
```
hisundar: Just curious, where do we remove entries that do not belong to the current stream from the queue? So if we hit an entry that does not belong to the queue, won't we still loop forever?

PR #2013, opened 2019-07-09
Cache Concurrency

Maithem:
## Overview
Added a parameter to configure the cache's concurrency level.

Why should this be merged: Ability to tune the cache's performance

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -165,6 +165,15 @@ public class CorfuRuntime {
         long maxCacheWeight;
 
         /**
+         * This is a hint to size the AddressSpaceView cache, a higher concurrency
+         * level allows for less lock contention at the cost of more memory overhead.
+         * The default value of zero will result in using the cache's internal default
+         * concurrency level (i.e. 4). 
+         */
+        @Default
+        int cacheConcurrencyLevel = 0;
+
+        /**
          * Sets expireAfterAccess and expireAfterWrite in seconds.
          */
         @Default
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -164,6 +164,12 @@
         @Default
         long maxCacheWeight;
 
+        /**
+         * A parameter to tune memory usage vs cache concurrency
+         */
+        @Default
+        int cacheConcurrency = 4;
```
Maithem: Hmm, maybe set this to # cpu ?

Maithem:
This is a straightforward change, let's get some reviews here. Thanks.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
WenbinZhu: Can we do this?

Maithem: This is the default for guava. I might just leave it as uninitialized and use the library's defaults instead.

Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -164,6 +164,12 @@
         @Default
         long maxCacheWeight;
 
+        /**
+         * A parameter to tune memory usage vs cache concurrency
+         */
+        @Default
+        int cacheConcurrency = 0;
```
lixinchengdu: It would be better if more comments about what cacheConcurrency means.

hisundar: I agree, without some comments on what this means it is highly possible consumers can end up misusing it.

annym: +1 can we add some more info. Perhaps some hint on how to set? number of threads that could potentially access the cache at one time? not sure.

Maithem: Done.

PR #2018, opened 2019-07-11
Add Log for Fast Object Loader Failure

annym:
If the FoL fails we have no logs that provide data on the actual error.

### DIFF
--- a/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
+++ b/runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
@@ -657,17 +657,23 @@ public class FastObjectLoader {
             for (Map.Entry<UUID, StreamMetaData> entry : streamsMetaData.entrySet()) {
                 cfs[i++] = CompletableFuture.runAsync(() -> {
                     CheckPoint checkPoint = entry.getValue().getLatestCheckPoint();
-                    if (checkPoint == null) {
-                        log.info("resurrectCheckpoints[{}]: Truncated checkpoint for this stream",
-                                Utils.toReadableId(entry.getKey()));
-                        return;
-                    }
 
-                    // For now one by one read and apply
-                    for (long address : checkPoint.getAddresses()) {
-                        updateCorfuObject(getLogData(runtime, loadInCache, address));
-                    }
-                }, executorService);
+                    try {
+                        if (checkPoint == null) {
+                            log.info("resurrectCheckpoints[{}]: Truncated checkpoint for this stream",
+                                    Utils.toReadableId(entry.getKey()));
+                            return;
+                        }
+
+                        // For now one by one read and apply
+                        for (long address : checkPoint.getAddresses()) {
+                            updateCorfuObject(getLogData(runtime, loadInCache, address));
+                        }
+                    } catch (Throwable t) {
+                        log.error("resurrectCheckpoints[{}]: error on addresses {}", checkPoint.getCheckPointId(),
+                                checkPoint.getAddresses(), t);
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -656,18 +656,26 @@ private void resurrectCheckpoints() {
...
+                            log.info("resurrectCheckpoints[{}]: Truncated checkpoint for this stream",
+                                    Utils.toReadableId(entry.getKey()));
+                            return;
+                        }
+
+                        // For now one by one read and apply
+                        for (long address : checkPoint.getAddresses()) {
+                            updateCorfuObject(getLogData(runtime, loadInCache, address));
+                        }
+                    } catch (Throwable t) {
```
PavelZaytsev: Throwable is very broad, maybe we can be more specific?

annym: The thing here is that if anything goes wrong at this level (exception, error, whatever) we want to log it before this completes exceptionally. For instance we were seeing StackOverflowError which wouldn't have been captured if we caught any more specific exception.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -656,18 +656,26 @@ private void resurrectCheckpoints() {
...
+                        if (checkPoint != null) {
```
WenbinZhu: I think `checkPoint` will not be null here: `CheckPoint checkPoint = entry.getValue().getLatestCheckPoint();` This is a simple operation which should not throw any exception, then the above code checks `if (checkPoint == null)`, which will return. So if we get a Throwable here, it means checkpoint is not null. So I think you can remove this null check.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -656,18 +656,26 @@ private void resurrectCheckpoints() {
...
-                    // For now one by one read and apply
-                    for (long address : checkPoint.getAddresses()) {
-                        updateCorfuObject(getLogData(runtime, loadInCache, address));
-                    }
-                }, executorService);
+                    CheckPoint checkPoint = null;
+                    try {
+                        checkPoint = entry.getValue().getLatestCheckPoint();
+                        if (checkPoint == null) {
+                            log.info("resurrectCheckpoints[{}]: Truncated checkpoint for this stream",
```
hisundar: would logging the UUID also help here?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -656,18 +656,26 @@ private void resurrectCheckpoints() {
...
+
+                        // For now one by one read and apply
+                        for (long address : checkPoint.getAddresses()) {
+                            updateCorfuObject(getLogData(runtime, loadInCache, address));
+                        }
+                    } catch (Throwable t) {
+                        if (checkPoint != null) {
+                            log.error("resurrectCheckpoints: error on addresses {}", checkPoint.getAddresses());
+                        }
+                        log.error("resurrectCheckpoints: ", t);
```
hisundar: will logging UUID help here?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
annym: Done.

PR #2019, opened 2019-07-12
Add Log for Fast Object Loader Failure

annym:
Port #2018 to corfu-0.2.3


runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -657,17 +657,23 @@ private void resurrectCheckpoints() {
...
+                            log.info("resurrectCheckpoints[{}]: Truncated checkpoint for this stream",
+                                    Utils.toReadableId(entry.getKey()));
+                            return;
+                        }
+
+                        // For now one by one read and apply
+                        for (long address : checkPoint.getAddresses()) {
+                            updateCorfuObject(getLogData(runtime, loadInCache, address));
+                        }
+                    } catch (Throwable t) {
```
Maithem: Is the thread just unexpectedly dying?

PR #2020, opened 2019-07-12
Invalidate layout returns CompletableFuture 

xnull:
## Overview

Description:
The way how we invalidate (update) the layout right now is pretty inconvenient and completely synchronous (if you want to get updated layout).
This change enables us to invalidate layout asynchronously.
P.S. the old way also still working

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -53,6 +53,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -845,14 +846,17 @@ public class CorfuRuntime {
      * If the layout has been previously invalidated and a new layout has not yet been retrieved,
      * this function does nothing.
      */
-    public synchronized void invalidateLayout() {
+    public synchronized CompletableFuture<Layout> invalidateLayout() {
         // Is there a pending request to retrieve the layout?
-        if (!layout.isDone()) {
-            // Don't create a new request for a layout if there is one pending.
-            return;
+        if (layout.isDone()) {
+            List<String> servers = Optional.ofNullable(latestLayout)
+                    .map(Layout::getLayoutServers)
+                    .orElse(bootstrapLayoutServers);
+
+            layout = fetchLayout(servers);
         }
-        layout = fetchLayout(latestLayout == null
-                ? bootstrapLayoutServers : latestLayout.getLayoutServers());
+
+        return layout;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -845,14 +846,17 @@ public IClientRouter getRouter(String address) {
      * If the layout has been previously invalidated and a new layout has not yet been retrieved,
      * this function does nothing.
      */
-    public synchronized void invalidateLayout() {
+    public synchronized CompletableFuture<Layout> invalidateLayout() {
```
StrikingThirteen: changing the the public api. have you checked the implications?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -845,14 +846,17 @@ public IClientRouter getRouter(String address) {
...
         // Is there a pending request to retrieve the layout?
-        if (!layout.isDone()) {
-            // Don't create a new request for a layout if there is one pending.
-            return;
+        if (layout.isDone()) {
```
StrikingThirteen: Change the question comment above to an statement.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
Maithem: > pretty inconvenient and completely synchronous 
This is false. This call doesn't block until the layout is fetched. It might block on synchronized, but this method will just spawn an async task. The future generated from this method can be accessed by `CorfuRuntime::layout`

Maithem:
> If we change the public API to async, we need to account for this behavior in many different places or perform the get calls on the completable future everywhere else to preserve a synchronous behavior.

This change just exposes the same future that is used by getLayout (blocking call). This change shouldn't change much really.

PavelZaytsev:
Ok, sounds good

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
xnull: yep, it doesn't affect anything

xnull: >completely synchronous (if you want to get updated layout)

So, it is not convenient to execute two methods instead of one. 
Also, make `layout` variable public is a bad practice

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -851,14 +852,17 @@ public IClientRouter getRouter(String address) {
      * If the layout has been previously invalidated and a new layout has not yet been retrieved,
```
zalokhan: Please update the comments.

xnull: Why do I need to change this comment?

PR #2033, opened 2019-07-17
VLO clean up: return an empty list instead of null

xnull:
## Overview

Description:
VLO clean up: return an empty list instead of null

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
@@ -1,13 +1,5 @@
 package org.corfudb.runtime.object;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.UUID;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import com.google.common.annotations.VisibleForTesting;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.ISMRConsumable;
 import org.corfudb.protocols.logprotocol.SMREntry;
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -97,14 +97,13 @@ public void gc(long trimMark) {
...
-        if (data != null) {
-            if (data.getType() == DataType.DATA
-                    && data.getPayload(runtime) instanceof ISMRConsumable) {
-                return ((ISMRConsumable) data.getPayload(runtime))
-                        .getSMRUpdates(streamView.getId());
-            }
+        if (data == null
+                || data.getType() != DataType.DATA
+                || !(data.getPayload(runtime) instanceof ISMRConsumable)) {
+            return new ArrayList<>();
```
WenbinZhu: Can you use `Collections.emptyList()` instead?

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -124,7 +123,8 @@ public void gc(long trimMark) {
             }
             data = streamView.previous();
         }
-        return null;
+
+        return new ArrayList<>();
```
WenbinZhu: Can you use `Collections.emptyList()` instead?

xnull: done

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
xnull: done

Issue #2038, opened 2019-07-17
CorfuGuidGenerator does not work if nested in a different transaction type


hisundar:
java.lang.IllegalArgumentException: Attempting to nest transactions with a different transaction type, parent type=WRITE_AFTER_WRITE, nested type=OPTIMISTIC

	at org.corfudb.runtime.object.transactions.Transaction.verify(Transaction.java:73)
	at org.corfudb.runtime.object.transactions.Transaction.begin(Transaction.java:48)
	at org.corfudb.runtime.view.CorfuGuidGenerator.updateInstanceId(CorfuGuidGenerator.java:57)

This is because our transaction nesting is pretty strict and fails if nested transaction differs from the parent transaction type.

PR #2039, opened 2019-07-17
CorfuGuidGenerator: Use Parent TransactionType if nested

hisundar:
We don't allow nesting transactions of different types.
To get around this,
+ use parent transaction type if in one.
+ Separate test cases to validate different transaction types.
+ In CorfuQueue, remove comment that says getEntryId() and getOrdering() should only be used for testing since they are valid uses of the API.

Why should this be merged: 
Related issue(s) (if applicable): #2038 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
@@ -5,6 +5,7 @@ import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.collections.SMRMap;
 import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.object.transactions.TransactionType;
+import org.corfudb.runtime.object.transactions.TransactionalContext;
 
 import java.util.UUID;
 
@@ -48,11 +49,19 @@ public class CorfuGuidGenerator implements OrderedGuidGenerator {
     }
 
     private long updateInstanceId() {
-        long nextInstanceId = 0L;
+        long nextInstanceId;
         while(true) {
             try {
+                TransactionType txnType;
+                if (TransactionalContext.isInTransaction()) {
+                    txnType = TransactionalContext.getCurrentContext()
+                            .getTransaction().getType();
+                } else { // Pick default type as OPTIMISTIC
+                    txnType = TransactionType.OPTIMISTIC;
+
+                }
                 runtime.getObjectsView().TXBuild()
-                        .type(TransactionType.OPTIMISTIC)
+                        .type(txnType)
                         .build()
                         .begin();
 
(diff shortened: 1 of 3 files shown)


test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
```diff
@@ -24,7 +25,22 @@
...
+    public void TXBegin(TransactionType type) {
+        switch (type){
+            case WRITE_AFTER_WRITE:
+                WWTXBegin();
+                return;
+            case OPTIMISTIC:
+                OptimisticTXBegin();
+                return;
+            default:
+                throw new IllegalArgumentException("Unsupported TXN type");
```
epalta: Do we want to include input "type" as parameter in exception for debugging?

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -48,11 +49,19 @@ public CorfuGuidGenerator(CorfuRuntime rt) {
     }
 
     private long updateInstanceId() {
-        long nextInstanceId = 0L;
+        long nextInstanceId;
         while(true) {
             try {
+                TransactionType txnType;
+                if (TransactionalContext.isInTransaction()) {
+                    txnType = TransactionalContext.getCurrentContext()
```
xnull: currentContext can be null, right?

hisundar: not if it is in a transaction

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
hisundar: I was thinking of this, but this is really used for this test case in this file only, so skipped it for simplicity.

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
```diff
@@ -24,7 +25,22 @@
 @Slf4j
 public class CorfuQueueTxTest extends AbstractTransactionsTest {
     @Override
-    public void TXBegin() { OptimisticTXBegin(); }
+    public void TXBegin() {
+        OptimisticTXBegin();
```
lixinchengdu: How about TXBegin() calling TXBegin(TransactionType.OPTIMISTIC)? Maybe it is more intuitive that way to implement default parameters.

hisundar: Sounds good. Done.

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
hisundar: but let me make the change as you suggest. thanks

PR #2041, opened 2019-07-17
Console logging compression

xnull:
## Overview

Description:
Average compression ration for our logs is 1:20.
Change max file size to make average gzipped logs size equals 35mb.
Math:
When log file comes 700mb it gets compressed to ~35-40mb
We can have up to 30 compressed log files
Max total disk space consumed by gzipped logs: 40*30 = 1.2GB
And one, not compressed log file: 700mb.
Total number of uncompressed logs: 700 * 30 + 700 = 22GB.

Previous parameters for a total number of uncompressed logs: 100*30 + 100 = 3GB which is ~ 9 hours in a prod environment.

The diff: 22GB - 3GB = 18GB which is ~ 54 hours in a prod environment.

New settings, total time: ~70 hours. 

P.S. if we get rid of log messages:
1. `Trying connection to layout server...`
2. `log write...`

It should reduce logs about 60-80%.
Then we can count on ~ 300-500 hours of logs.

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/resources/logback.prod.xml
+++ b/infrastructure/src/main/resources/logback.prod.xml
@@ -11,7 +11,7 @@
             <maxIndex>30</maxIndex>
         </rollingPolicy>
         <triggeringPolicy class="ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy">
-            <maxFileSize>100MB</maxFileSize>
+            <maxFileSize>700MB</maxFileSize>
         </triggeringPolicy>
         <encoder>
             <pattern>


zalokhan:
Your calculation is not accounting for the compaction and the wrapper logs which share the same quota.

xnull:
>wrapper logs
pretty small, if I am not wrong we limit it by 100mb. Usually, it's 50mb

xnull:
> compaction logs
outside of Corfu. But again if compaction logs consume more than 100mb, it means something wrong with that logs

zalokhan:
>But again if compaction logs consume more than 100mb, it means something wrong with that logs

@xnull That is totally relative. 100MB might be too small if it can't even store 2-3 compaction runs.
You might want to check the compactor log xml configurations to know the maximum size the compactor logs can take.

xnull:
> Is there any performance tests? Will compression significantly affect the logging performance?

I don't see how this question can be relevant. So, I am pretty sure there is no performance impact

xnull:
> @xnull That is totally relative. 100MB might be too small if it can't even store 2-3 compaction runs.

Yes, will do, but it must not affect this change anyway

xnull:
Are we ok with the change?

PR #2045, opened 2019-07-18
Change default NO_CONFLICT_KEY constant

vjeko:
Change it to a more sensible value byte[]{ 0 }.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/protocols/wireprotocol/TokenResponse.java
```diff
@@ -17,7 +17,7 @@
 @AllArgsConstructor
 public class TokenResponse implements ICorfuPayload<TokenResponse>, IToken {
 
-    public static byte[] NO_CONFLICT_KEY = new byte[]{};
+    public static byte[] NO_CONFLICT_KEY = new byte[]{ 0 };
```
Maithem: It seems like this change is needed only because there's code that consumes `TransactionAbortedException::getConflictKey` but it doesn't handle an empty array, which makes this change seem like a hack.

vjeko:
Done.

Issue #2042, opened 2019-07-17
StreamsView::append uses conflictKey/conflictStream from a stale tokenResponse


WenbinZhu:
## Overview

On OverwriteException, StreamsView::append will grab a new token and retry, however the current implementation uses the conflict info from the old tokenResponse. In case of transaction, if we encounter an OverwriteException, that means the there was no conflict transactions, and the conflictKey/conflictStream returned from sequencer is empty, however when we grab a new token, there could be a conflict, so using the conflictKey/conflictStream could result in something like this:

```
org.corfudb.runtime.exceptions.TransactionAbortedException: TX ABORT  | Snapshot Time = Token(epoch=388, sequence=1681679) | Failed Transaction ID = 905a3651-2377-4676-8794-XXXXX | Offending Address = 1681732 | Conflict Key =  | Conflict Stream = 00000000-0000-0000-0000-000000000000 | Cause = CONFLICT | Time = 7543 ms
```

The cause  is CONFLICT, but Conflict Key and Conflict Stream are empty, which is because we used the information from a stale response.

Maithem:
Addressed.

PR #2046, opened 2019-07-18
Do not use stale conflictKey/conflictStream on overwritten for TXNs.

WenbinZhu:
## Overview

- On OverwriteException, StreamsView::append will grab a new token and
retry, but it was still using the conflictKey/conflictStream info from
the stale tokeResponse. In case of transactions and there is a conflict
when we try to re-requst the token, the stale conflictKey/conflictStream
is empty. This patch directly uses the new token and retry.

- Some clean up in StreamsView and TokenType.

Related issue(s) (if applicable): #2042 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
@@ -1,24 +1,14 @@
 package org.corfudb.runtime.view;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.UUID;
-import java.util.stream.Collectors;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.protocols.wireprotocol.DataType;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.TokenResponse;
-import org.corfudb.protocols.wireprotocol.TokenType;
 import org.corfudb.protocols.wireprotocol.TxResolutionInfo;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.AbortCause;
@@ -29,6 +19,12 @@ import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
 import org.corfudb.runtime.view.stream.IStreamView;
 import org.corfudb.util.Utils;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Arrays;
+import java.util.UUID;
+import java.util.stream.Collectors;
 
 /**
  * Created by mwei on 12/11/15.
@@ -95,7 +91,6 @@ public class StreamsView extends AbstractView {
      * unsafe streams will be excluded (because its unsafe for the garbage
      * collector thread to operate on them while being used by a different
      * thread).
-     *
      */
     public void gc(long trimMark) {
         for (IStreamView streamView : getStreamCache().values()) {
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -133,31 +128,28 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
                 : runtime.getSequencerView().next(conflictInfo, streamIDs); // Token w/ conflict info
```
Maithem: I think its cleaner to push Line 126 inside the for-loop and get rid of the other sequencer calls.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -167,37 +159,25 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
...
                 if (conflictInfo == null) {
                     // Token w/o conflict info
-                    temp = runtime.getSequencerView().next(streamIDs);
+                    tokenResponse = runtime.getSequencerView().next(streamIDs);
                 } else {
-
-                    // On retry, check for conflicts only from the previous
-                    // attempt position
+                    // On retry, check for conflicts only from the previous attempt position,
+                    // otherwise the transaction will always conflict with itself.
```
Maithem: This is very subtle, can you write a unit test for it.

WenbinZhu: @Maithem  Done, added a test.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
WenbinZhu: Done.

test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
```diff
@@ -842,4 +867,35 @@ public void txnSpansEpochsCanCommitIfPrimarySequencerSpansConsecutiveEpochs() th
...
+        rt.connect();
+
+        OptimisticTXBegin();
+
+        put("k1", "v1");
+        put("k2", "v2");
+
+        // Add a client rule that does a hole fill to create an OverwriteException
+        // after the token is issued with no conflict and before writing to log unit.
+        AtomicInteger counter = new AtomicInteger(0);
```
Maithem: A cleaner way is to just fill a hole directly without going to the sequencer such that the next token handed out will cause the writer to fail and retry. Its much cleaner than this type of synchronization

WenbinZhu: Done.

test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
```diff
@@ -842,4 +867,29 @@ public void txnSpansEpochsCanCommitIfPrimarySequencerSpansConsecutiveEpochs() th
...
+
+        put("k1", "v1");
+        put("k2", "v2");
+
+        // Hole fill the address that the transaction is about to commit, to
+        // create an OverwriteException and retry. The retry should succeed.
+        long currentTail = rt.getSequencerView().query().getToken().getSequence();
+        rt.getAddressSpaceView().read(currentTail + 1);
+
+        assertThatCode(this::TXEnd).doesNotThrowAnyException();
```
Maithem: Can you check that the returned address is currentTail + 2 and that the data is there?

WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -116,82 +111,63 @@ public void gc(long trimMark) {
...
         ld.checkMaxWriteSize(runtime.getParameters().getMaxWriteSize());
 
-        // Go to the sequencer, grab an initial token.
-        TokenResponse tokenResponse = conflictInfo == null
-                ? runtime.getSequencerView().next(streamIDs) // Token w/o conflict info
-                : runtime.getSequencerView().next(conflictInfo, streamIDs); // Token w/ conflict info
-
+        TokenResponse tokenResponse = null;
         for (int x = 0; x < runtime.getParameters().getWriteRetry(); x++) {
+            // Go to the sequencer, grab an token to write.
```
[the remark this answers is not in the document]
WenbinZhu: Done.

PR #2047, opened 2019-07-18
Decrease log level for OverwriteException

xnull:
## Overview

Description:
Descrease log level for OverwriteException

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
@@ -26,6 +26,7 @@ import org.corfudb.protocols.wireprotocol.TailsRequest;
 import org.corfudb.protocols.wireprotocol.TailsResponse;
 import org.corfudb.protocols.wireprotocol.TrimRequest;
 import org.corfudb.protocols.wireprotocol.WriteRequest;
+import org.corfudb.runtime.exceptions.OverwriteException;
 import org.corfudb.runtime.exceptions.QuotaExceededException;
 import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
@@ -193,9 +194,14 @@ public class BatchProcessor implements AutoCloseable {
                             default:
                                 log.warn("Unknown BatchWriterOperation {}", currOp);
                         }
-                    } catch (Exception e) {
-                        log.error("Stream log error. Batch [queue size={}]. StreamLog: [trim mark: {}, tails: {}].",
-                                operationsQueue.size(), streamLog.getTrimMark(), streamLog.getAllTails(), e);
+                    } catch (OverwriteException ex) {
+                        log.trace("OverwriteException. Batch [queue size={}]. StreamLog: [trim mark: {}].",
+                                operationsQueue.size(), streamLog.getTrimMark(), ex);
+                        currOp.getFutureResult().completeExceptionally(ex);
+                    }
+                    catch (Exception e) {
+                        log.error("Stream log error. Batch [queue size={}]. StreamLog: [trim mark: {}].",
+                                operationsQueue.size(), streamLog.getTrimMark(), e);
                         currOp.getFutureResult().completeExceptionally(e);
                     }
                     res.add(currOp);


infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -194,7 +194,7 @@ private void processor() {
                                 log.warn("Unknown BatchWriterOperation {}", currOp);
                         }
                     } catch (Exception e) {
-                        log.error("Stream log error. Batch [queue size={}]. StreamLog: [trim mark: {}, tails: {}].",
```
hisundar: Wont this swallow other exceptions as well?

xnull: no

xnull:
@WenbinZhu please see the mail chain

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
lixinchengdu: It looks strange for an exception to use trace level logging.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -194,7 +194,7 @@ private void processor() {
...
+                        log.trace("Stream log error. Batch [queue size={}]. StreamLog: [trim mark: {}, tails: {}].",
```
zalokhan: Apart from reducing the log level, can we add one more log line which does not prin the streamLog.getAllTails.
This will keep just one line for debugging.

xnull:
changes:
 - log level is debug
 - don't print tails

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -194,8 +194,8 @@ private void processor() {
                                 log.warn("Unknown BatchWriterOperation {}", currOp);
                         }
                     } catch (Exception e) {
-                        log.error("Stream log error. Batch [queue size={}]. StreamLog: [trim mark: {}, tails: {}].",
-                                operationsQueue.size(), streamLog.getTrimMark(), streamLog.getAllTails(), e);
+                        log.debug("Stream log error. Batch [queue size={}]. StreamLog: [trim mark: {}].",
```
WenbinZhu: The title of this patch suggests we want to reduce the occurrence of OverwriteException in logs, but here you reduced log level for all kinds of exceptions. It swallows important exceptions like IOException if we are in production which may use higher log level. Why don't we check the exception type and process overwriteException differently?

xnull: I don't see any reasons to do that, what's the motivation?
 anyway we are going to write this message in the log, just without stream tails

xnull: And also the exception is not swallowed, it goes to the CompletableFuture and can (will?) be handled on next steps

WenbinZhu: In production, we may use INFO level logging. Then since you make it debug level, we will miss  possibly some really critical but unexpected exception.

xnull: we don't use it. If we migrate to INFO level, then yes. But the migration is a huge change. When it happens, this also will be changed. Now I vote for - keep it consistent with the current implementation. And also this must be the debug level anyway because we handle the exception on next staeps

WenbinZhu: Why is this related to migration? It's just a deployment configuration right?

xnull: I don't understand your question, sorry

xnull: Changed, according to offline discussion

WenbinZhu: Cool.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -193,9 +194,14 @@ private void processor() {
                             default:
                                 log.warn("Unknown BatchWriterOperation {}", currOp);
                         }
-                    } catch (Exception e) {
-                        log.error("Stream log error. Batch [queue size={}]. StreamLog: [trim mark: {}, tails: {}].",
-                                operationsQueue.size(), streamLog.getTrimMark(), streamLog.getAllTails(), e);
+                    } catch (OverwriteException ex) {
+                        log.trace("OverwriteException. Batch [queue size={}]. StreamLog: [trim mark: {}].",
```
zalokhan: @xnull We do want to print a one-liner for OverwriteExceptions too. With this, it's difficult to know there were overwrite exceptions.

xnull: fixed

PR #2049, opened 2019-07-18
Logback: migrate to an asynchronous appender

xnull:
## Overview

Description:
Asynchronous appender is very helpful in case of a slow disk (critical for us) or other cases (see https://logback.qos.ch/manual/appenders.html#AsyncAppender).

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/resources/logback.prod.xml
+++ b/infrastructure/src/main/resources/logback.prod.xml
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration scan="true" scanPeriod="30 seconds">
 
+    <!-- https://logback.qos.ch/manual/configuration.html#stopContext -->
+    <shutdownHook/>
+
     <property name="LOG_DIRECTORY" value="/var/log/corfu" />
 
     <appender name="file" class="ch.qos.logback.core.rolling.RollingFileAppender">
@@ -20,7 +23,13 @@
         </encoder>
     </appender>
 
+    <!-- https://logback.qos.ch/manual/appenders.html#AsyncAppender -->
+    <appender name="async_file" class="ch.qos.logback.classic.AsyncAppender">
+        <appender-ref ref="file" />
+        <queueSize>1024</queueSize>
+    </appender>
+
     <root level="info">
-        <appender-ref ref="file"/>
+        <appender-ref ref="async_file"/>
     </root>
 </configuration>


infrastructure/src/main/resources/logback.prod.xml
```diff
@@ -20,7 +23,13 @@
         </encoder>
     </appender>
 
+    <!-- https://logback.qos.ch/manual/appenders.html#AsyncAppender -->
```
Maithem: Why don't you also use `ImmediateFlush=false` and `BufferedIO`

xnull: Since we have AsyncAppender what's the point for immediateFlush=false?
Didn't see BufferedIO in the documentation, can't say anything

infrastructure/src/main/resources/logback.prod.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration scan="true" scanPeriod="30 seconds">
 
+    <!-- https://logback.qos.ch/manual/configuration.html#stopContext -->
+    <shutdownHook/>
```
WenbinZhu: Could you explain this line?

infrastructure/src/main/resources/logback.prod.xml
```diff
@@ -20,7 +23,13 @@
         </encoder>
     </appender>
 
+    <!-- https://logback.qos.ch/manual/appenders.html#AsyncAppender -->
+    <appender name="async_file" class="ch.qos.logback.classic.AsyncAppender">
+        <appender-ref ref="file" />
+        <queueSize>1024</queueSize>
```
WenbinZhu: Is there a way to tune this parameter?

infrastructure/src/main/resources/logback.prod.xml
vjeko: "In standalone Java applications, adding a <shutdownHook/> directive to your configuration file is an easy way to ensure that any ongoing compression tasks are allowed to finish before JVM exit."

infrastructure/src/main/resources/logback.prod.xml
xnull: You define this parameter it's up to you. Don't get your question

WenbinZhu: I mean this is hardcoded now, is there any way to change this parameter when you deploy corfu server without having to manually change the file, like this: https://github.com/CorfuDB/CorfuDB/blob/master/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java#L115

xnull: You have direct access to `logback-prod.xml` and can/should make changes in that file if needed. I strongly believe this parameter shall not be configured via command line

WenbinZhu: Changing the source file on every deployment is not a good idea. I believe it's better to have this parameter configurable as it shall change in terms of production machine's memory, IO bandwidth, etc. But anyway it's up to you.

xnull: It must be in an xml config file. There are many proper ways to manage these parameters in/for different environments. Changing it via command-line definitely a wrong way. 
If we would need to make this parameter more dynamic we will figure out that. For now, this is not needed

PR #2015, opened 2019-07-10
Non-Cachable Streams

Maithem:
## Overview
Allow certain streams to not be cached.

Why should this be merged: Some access patterns are pure consumers, once a stream entry is read it is no longer needed, if its cached it can pollute the AddressSpace cache and compete with transaction workloads ( that need the cache).   

Related issue(s) (if applicable): #<number>

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #2055, opened 2019-07-19
Non cacheable streams port

Maithem:
## Overview
Porting:

Allow certain streams to not be cached.
nonCacheableStreams (#2015)


Maithem:
Let's get some more reviews here./

PR #2057, opened 2019-07-20
Reduce stream interface

Maithem:
## Overview
Removed unnecessary APIs.

Why should this be merged: less code, reduced complexity. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -601,39 +601,7 @@ public abstract class AbstractQueuedStreamView extends
      */
     @Override
     public void close() {}
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public synchronized long find(long globalAddress, SearchDirection direction) {
-        final QueuedStreamContext context = getCurrentContext();
-        // First, check if we have resolved up to the given address
-        if (context.maxResolution < globalAddress) {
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -43,64 +41,36 @@
...
      * @param id                    The id of the stream.
-     * @param contextFactory        A function which generates a context,
-     *                              given the stream id and a maximum global
-     *                              address.
      */
     public AbstractContextStreamView(final CorfuRuntime runtime,
-                                     final UUID id,
-                                     final BiFunction<UUID, Long, T>
-                                             contextFactory) {
+                                     final UUID id) {
```
lixinchengdu: streamId is a more informative name compared to id

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamContext.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public StreamContext(UUID id, long maxGlobalAddress) {
+        super(id, maxGlobalAddress);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    void reset() {
```
lixinchengdu: This function would better be synchronized. Since all streamview operations are wrapped in ThreadSafeStreamView, it is also possible to not use synchronization in this function but synchronization should also be removed at seek().

PR #2059, opened 2019-07-23
Reduced The Sequencer's Memory Overhead

Maithem:
## Overview
Instead of converting the conflictParams to strings, its better to
keep them as byte arrays, its more compact.

Why should this be merged: Reduces the sequencer's memory footprint by ~4x

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
@@ -196,11 +196,11 @@ public class SequencerServerCache {
     @EqualsAndHashCode
     public static class ConflictTxStream {
         private final UUID streamId;
-        private final String conflictParam;
+        private final byte[] conflictParam;
 
         public ConflictTxStream(UUID streamId, byte[] conflictParam) {
             this.streamId = streamId;
-            this.conflictParam = Utils.bytesToHex(conflictParam);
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -196,11 +196,11 @@ public void updateMaxConflictAddress(long newMaxConflictWildcard) {
     @EqualsAndHashCode
     public static class ConflictTxStream {
         private final UUID streamId;
-        private final String conflictParam;
+        private final byte[] conflictParam;
```
Maithem: Verified that lombok does the right thing for the equality check (i.e. calls java.util.Arrays.equals). Here's the delomboked class

```java
public boolean equals(final Object o) {
            if (o == this) return true;
            if (!(o instanceof ConflictTxStream)) return false;
            final ConflictTxStream other = (ConflictTxStream) o;
            if (!other.canEqual((Object) this)) return false;
            final Object this$streamId = this.streamId;
            final Object other$streamId = other.streamId;
            if (this$streamId == null ? other$streamId != null : !this$streamId.equals(other$streamId)) return false;
            if (!java.util.Arrays.equals(this.conflictParam, other.conflictParam)) return false;
            return true;
        }
```

PR #2062, opened 2019-07-24
Dont Instrument unmapSecondaryIndexes

Maithem:
## Overview
The unmapSecondaryIndexes is not an SMRAcessor, so the annotation
processor shouldn't generator an accessor method for it.

Why should this be merged: prevents the annotation processor of generating an accessor in CorfuTable$CORFUSMR

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -465,12 +465,9 @@ public class CorfuTable<K ,V>
     @Override
     @Mutator(name = "put", noUpcall = true)
     public void insert(@ConflictParameter K key, V value) {
-        V previous = mainMap.put(key, value);
-        // If we have index functions, update the secondary indexes.
-        if (!secondaryIndexes.isEmpty()) {
-            unmapSecondaryIndexes(key, previous);
-            mapSecondaryIndexes(key, value);
-        }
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -711,6 +711,7 @@ public V merge(K key, V value, BiFunction<? super V, ? super V, ? extends V>
      * @param key   The primary key (index) for the mapping.
      * @param value The value to unmap.
      */
+    @DontInstrument
```
Maithem: Hmm, probably need to mark `clearIndex`  as well

Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -748,6 +749,7 @@ protected void unmapSecondaryIndexes(K key, V value) {
      * @param key   the primary key associated with the indexing.
      * @param value the value to map.
      */
+    @DontInstrument
     @SuppressWarnings("unchecked")
     protected void mapSecondaryIndexes(K key, V value) {
```
Maithem: `CorfuTable::insert` used to sync on maoSecondaryIndexes, but the method is a pure mutator

PR #2063, opened 2019-07-24
Set log-size-quota in percentage of total disk instead of absolute bytes

hisundar:
## Overview
Setting a quota in absolute bytes has problems when different filesystems have different volume capabilities.

Description:

Why should this be merged: 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -117,15 +118,15 @@ public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
...
 
-        logSizeLimit = Long.parseLong(logSizeQuotaParam);
+        logSizeLimitPercentage = Double.parseDouble(logSizeLimitPercParam);
 
         initStreamLogDirectory();
 
         long initialLogSize = estimateSize(logDir);
         log.info("StreamLogFiles: {} size is {} bytes, limit {}", logDir, initialLogSize, logSizeLimit);
-        logSizeQuota = new ResourceQuota("LogSizeQuota", logSizeLimit == -1 ? Long.MAX_VALUE : logSizeLimit);
+        logSizeQuota = new ResourceQuota("LogSizeQuota", logSizeLimit);
```
WenbinZhu: Is there a way to explicitly initialize `logSizeLimit` in this constructor? I see the initialization of `logSizeLimit` is inside `initStreamLogDirectory()`, which seems confusing, right?

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
```diff
@@ -38,7 +38,7 @@
     String address = "test";
     int port = 9000;
     String seqCache = "1000";
-    String logSizeQuota = "-1";
+    String logSizeLimitPercentage = "100";
```
WenbinZhu: "100.0"?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -117,15 +118,15 @@ public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
         channelsToSync = new HashSet<>();
         this.verify = !noVerify;
         this.dataStore = StreamLogDataStore.builder().dataStore(serverContext.getDataStore()).build();
-        String logSizeQuotaParam = (String) serverContext.getServerConfig().get("--log-size-quota-bytes");
+        String logSizeLimitPercParam = (String) serverContext.getServerConfig().get("--log-size-quota-percentage");
 
-        logSizeLimit = Long.parseLong(logSizeQuotaParam);
+        logSizeLimitPercentage = Double.parseDouble(logSizeLimitPercParam);
```
WenbinZhu: Is is better to make `logSizeLimitPercentage` a local parameter and pass it into initStreamLogDirectory() ?

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
hisundar: done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
hisundar: If we move it to the constructor we have to duplicate the IOException handling logic again, which might be uglier right?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
hisundar: done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -99,7 +99,7 @@
     // the files of the old instance
     private LogMetadata logMetadata;
 
-    private final long logSizeLimit;
+    private long logSizeLimit;
```
Maithem: Can we keep this as final ? maybe `initStreamLogDirectory` can return this number instead of assigning it inside the method

PR #2064, opened 2019-07-24
Change how we deal with the transaction stream

vjeko:
The transaction stream it not checkpoint aware, thus, we need to ensure
that address space discovery mechanism takes into account.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
@@ -18,6 +18,7 @@ import org.corfudb.protocols.wireprotocol.StreamAddressRange;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.view.Address;
+import org.corfudb.runtime.view.ObjectsView;
 import org.corfudb.runtime.view.StreamOptions;
 import org.roaringbitmap.longlong.Roaring64NavigableMap;
 
@@ -159,16 +160,32 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
                 queue.addAll(streamAddressSpace.copyAddressesToSet(maxGlobal));
 
                 long trimMark = streamAddressSpace.getTrimMark();
+
+                // In case we are dealing with a stream that does not have the checkpoint
+                // capability, check to see if we are trying to access an address that has been
+                // previously trimmed.
+                if (!isCheckpointCapable()
+                        && Address.isAddress(trimMark)
+                        && trimMark > stopAddress) {
+                    String message = String.format("getStreamAddressMap[{%s}] stream has been " +
+                                    "trimmed at address %s and we are trying to access the " +
+                                    "stream starting at address %s. This stream does not have " +
+                                    "the checkpoint capability.", this, trimMark, stopAddress);
+                    log.info(message);
+                    throw new TrimmedException(message);
+                }
+
                 // Address maps might have been trimmed, hence not reflecting all updates to the stream
                 // For this reason, in the case of a valid trim mark, we must be sure this space is
                 // already resolved or loaded by a checkpoint.
-                if (Address.isAddress(trimMark) && !isTrimCoveredByCheckpointOrLocalView(trimMark)) {
+                if (isCheckpointCapable()
+                        && Address.isAddress(trimMark)
+                        && !isTrimCoveredByCheckpointOrLocalView(trimMark)) {
                     String message = String.format("getStreamAddressMap[{%s}] stream has been " +
                                     "trimmed at address %s and this space is not covered by the " +
                                     "loaded checkpoint with start address %s, while accessing the " +
-                                    "stream at version %s. Looking for a new checkpoint.", this,
-                            streamAddressSpace.getTrimMark(),
-                            getCurrentContext().checkpoint.startAddress, maxGlobal);
+                                    "stream at version %s. Looking for a new checkpoint.",this,
+                            trimMark, getCurrentContext().checkpoint.startAddress, maxGlobal);
                     log.info(message);
                     if (getReadOptions().isIgnoreTrim()) {
                         log.debug("getStreamAddressMap[{}]: Ignoring trimmed exception for address[{}].",
@@ -181,7 +198,6 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
         }
 
         addressCount += queue.size();
-
         return !queue.isEmpty();
     }
 
@@ -346,6 +362,15 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
     private boolean isTrimCoveredByCheckpoint(long trimMark) {
         return getCurrentContext().checkpoint.id != null &&
                 getCurrentContext().checkpoint.startAddress >= trimMark;
+    }
+
+    /**
+     * Check to see if the current stream is checkpoint capable.
+     *
+     * @return whether this stream is capable of being checkpointed
+     */
+    private boolean isCheckpointCapable() {
+        return !getId().equals(ObjectsView.TRANSACTION_STREAM_ID);
     }
 
     @Override
--- a/test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
+++ b/test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
@@ -1,5 +1,6 @@
 package org.corfudb.runtime.view;
 
+import com.google.common.reflect.TypeToken;
 import lombok.Getter;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.Token;
@@ -15,11 +16,12 @@ import org.corfudb.runtime.view.stream.IStreamView;
 import org.junit.Before;
 import org.junit.Test;
 
-import java.io.IOException;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.stream.IntStream;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -153,8 +155,13 @@ public class StreamViewTest extends AbstractViewTest {
         // Open the stream with a new client
         CorfuRuntime rt2 = getNewRuntime(getDefaultNode())
                                         .connect();
-        txStream = rt2.getStreamsView().get(ObjectsView.TRANSACTION_STREAM_ID, options);
-        entries = txStream.remainingUpTo(Long.MAX_VALUE);
+        IStreamView txStream2 = rt2.getStreamsView()
+                 .get(ObjectsView.TRANSACTION_STREAM_ID, options);
+        assertThatThrownBy(() -> txStream2.remaining())
+                .isInstanceOf(TrimmedException.class);
+
+        txStream2.seek(firstIter / 2);
+        entries = txStream2.remainingUpTo(Long.MAX_VALUE);
         assertThat(entries.size()).isEqualTo((firstIter / 2));
     }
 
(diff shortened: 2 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -159,16 +160,31 @@ protected boolean discoverAddressSpace(final UUID streamId,
                 queue.addAll(streamAddressSpace.copyAddressesToSet(maxGlobal));
 
                 long trimMark = streamAddressSpace.getTrimMark();
+
+                // In case we are dealing with a stream that does not have the checkpoint
+                // capability, check to see if we are trying to access an address that has been
+                // previously trimmed.
+                if (!checkpointCapableStream()
```
lixinchengdu: What if the stream is transactional stream but trimMark >= stopAddress?

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -348,6 +363,10 @@ private boolean isTrimCoveredByCheckpoint(long trimMark) {
                 getCurrentContext().checkpoint.startAddress >= trimMark;
     }
 
+    private boolean checkpointCapableStream() {
```
lixinchengdu: It seems that checkpointCapable is a more appropriate name.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
vjeko: This is exactly what this if clause is checking: if (transactionStream && trimMark >= stopAddress).

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -159,16 +160,31 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
+                // In case we are dealing with a stream that does not have the checkpoint
+                // capability, check to see if we are trying to access an address that has been
+                // previously trimmed.
+                if (!isCheckpointCapable()
+                        && Address.isAddress(trimMark)
+                        && trimMark >= stopAddress) {
+                    String message = String.format("getStreamAddressMap[{%s}] stream has been " +
+                                    "trimmed at address %s and we are trying to access the " +
+                                    "stream starting at address %s.", this, trimMark, stopAddress);
+                    log.info(message);
```
annym: But, how will we recover? throwing a TrimmedException, makes upper layers attempt to reset the stream to load from a checkpoint... failing to find one (as the stream is not checkpoint capable)... hence we don't move the pointer and it is always pointing to the last read position. 

Or will this be left to the client to seek up to the new tail?

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -159,16 +160,31 @@ protected boolean discoverAddressSpace(final UUID streamId,
                 queue.addAll(streamAddressSpace.copyAddressesToSet(maxGlobal));
 
                 long trimMark = streamAddressSpace.getTrimMark();
+
+                // In case we are dealing with a stream that does not have the checkpoint
+                // capability, check to see if we are trying to access an address that has been
+                // previously trimmed.
+                if (!isCheckpointCapable()
+                        && Address.isAddress(trimMark)
+                        && trimMark >= stopAddress) {
```
annym: Since stop address is exclusive it should be >, right?

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -464,4 +466,69 @@ public void testPreviousWithStreamCheckpoint() throws Exception {
...
+            localRuntime.getObjectsView().TXEnd();
+        });
+
+        // Force a prefix trim beyond our current pointer.
+        localRuntime.getAddressSpaceView().prefixTrim(new Token(epoch,
+                PARAMETERS.NUM_ITERATIONS_LOW + PARAMETERS.NUM_ITERATIONS_LOW/2));
+
+        // Wait until log is trimmed.
+        TimeUnit.SECONDS.sleep(5);
+        assertThatThrownBy(() -> txStream.remaining()).isInstanceOf(TrimmedException.class);
```
annym: Can you complete this test with the case I described earlier, seeing that retrying will fail again.. (cause the pointer does not move) and seek again to the end to see we get out of this failing loop.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
vjeko: That is correct -- the client has to heal manually by seeking.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
vjeko: OK great. I was not sure about that one.

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
vjeko: Sounds good.

vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Just one last thought, we should be OK for now as this stream is not consumed through the object layer. But for nonCheckpointCapable streams accessed through the Object Layer, this TrimmedException would cause a retry->reset->load from checkpoint (which would incur in unnecessary sequencer calls) as there is no checkpoint to load from. Just leaving this note as we might want to improve this later on. For now, it's OK as it will be directly caught by the client.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -348,6 +363,10 @@ private boolean isTrimCoveredByCheckpoint(long trimMark) {
                 getCurrentContext().checkpoint.startAddress >= trimMark;
     }
 
+    private boolean isCheckpointCapable() {
```
WenbinZhu: Please add comments for this method.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
WenbinZhu: +1 we may need to define a new exception that will bypass the reset and load.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -159,16 +160,31 @@ protected boolean discoverAddressSpace(final UUID streamId,
                 queue.addAll(streamAddressSpace.copyAddressesToSet(maxGlobal));
 
                 long trimMark = streamAddressSpace.getTrimMark();
+
+                // In case we are dealing with a stream that does not have the checkpoint
+                // capability, check to see if we are trying to access an address that has been
+                // previously trimmed.
+                if (!isCheckpointCapable()
+                        && Address.isAddress(trimMark)
+                        && trimMark > stopAddress) {
```
WenbinZhu: Question here: Why we need to add a new `if` block here? I think the `isTrimCoveredByCheckpointOrLocalView(trimMark)` in the next `if` block covers the case of a non checkpoint capable stream, because it checks if `trimMark` is in the resolved queue, if it's not, it will throw a TrimmedException. The only difference I see is `options.ignoreTrimmed`, because in case of non checkpoint capable stream, we need to ignore this option and always throw TrimmedException. Then I think we don't need to add this `if` block, just add a isCheckpointCapable in the next `if` block when we check the option:

```java
if (isCheckpointCapable() && options.ignoreTrimmed) {              
    log.debug("getStreamAddressMap[{}]: Ignoring trimmed exception for address[{}].",
                                this, streamAddressSpace.getTrimMark());
 } else {
    log.warn(.....)
    throw new TrimmedException(message);
}
```

Is that true?

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -464,4 +471,75 @@ public void testPreviousWithStreamCheckpoint() throws Exception {
...
+            localRuntime.getObjectsView().TXEnd();
+        });
+
+        // Force a prefix trim.
+        final long epoch = 0L;
+        localRuntime.getAddressSpaceView().prefixTrim(
+                new Token(epoch, PARAMETERS.NUM_ITERATIONS_LOW/2/2));
+
+        // Wait until log is trimmed.
+        Thread.sleep(sleepDuration.toMillis());
```
zalokhan: Putting sleeps in the test is a bad idea and makes these tests flaky. Travis environments are slow and might behave differently. Can you query the trimMark and busy-wait on that?

vjeko: I mostly copied the test from somewhere else and made the necessary changes.

But yea -- will do.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
vjeko: Done.

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -159,16 +160,31 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
+                    String message = String.format("getStreamAddressMap[{%s}] stream has been " +
```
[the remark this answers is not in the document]
vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
vjeko: Option ignoreTrimmed needs to go away since it was introduced to deal with trimmed exceptions at the beginning of the stream consumption life-cycle.

vjeko:
BTW, I do not think that the streaming layer is behaving the way clients are expecting. 

For example, KVStore consumes the stream layer directly and assumes that all entries that it receives will be in the exact order they were produced (written). This whole check=pointing business makes sense from the object layer, but if consumed directly we might end up seeing the check-pointed entries which is something that client is not expecting.

Maithem:
> BTW, I do not think that the streaming layer is behaving the way clients are expecting.
> 
> For example, KVStore consumes the stream layer directly and assumes that all entries that it receives will be in the exact order they were produced (written). This whole check=pointing business makes sense from the object layer, but if consumed directly we might end up seeing the check-pointed entries which is something that client is not expecting.

That's because the object layer is tightly coupled with the stream layer (even for purposes of MVCC), which I think we should move away from in the future. Because of the tight coupling, the stream layer implementation exposes a single stream abstraction over two streams (main stream and checkpoint stream). This should be pushed up to an upper layer abstraction (i.e. a checkpointableStream) that the objects would consume (instead of consuming the raw stream).

PR #2070, opened 2019-07-26
Log Stream name + ID on startup to help debugging

hisundar:
## Overview
Without it, locating tables that do not print table names is impossible.


runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
```diff
@@ -90,6 +90,7 @@ public T open() {
 
         if (streamName != null) {
             streamID = CorfuRuntime.getStreamID(streamName);
+            log.info("CorfuObject: creating stream {} id {}", streamName, streamID);
```
WenbinZhu: CorfuObject -> ObjectBuilder?

hisundar: Since this can appear in any vertical, having Corfu in the name I thought helps.

WenbinZhu: Maybe we can put Corfu after this？ something like `"ObjectBuilder: creating Corfu stream ..."`

runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
```diff
@@ -90,6 +90,9 @@ public T open() {
 
         if (streamName != null) {
             streamID = CorfuRuntime.getStreamID(streamName);
+            log.info("CorfuObject: creating stream {} id {}", streamName, streamID.toString());
```
Maithem: I think you can consolidate the log lines, printing null is fine.

runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/ObjectBuilder.java
hisundar: done.

PR #2072, opened 2019-07-26
MultiCheckpointWriter Bug Fix

annym:
Whenever the last address of a stream is a hole (right before the checkpointer is triggered)
We build the checkpoint start log address based on the wrong start of the stream,
this is because we discard holes and hence set the start to the previous DATA entry.
This will diverge from the state kept by the sequencer for this stream on trims (trim mark).

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -11,10 +11,12 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.MultiSMREntry;
 import org.corfudb.protocols.logprotocol.SMREntry;
+import org.corfudb.protocols.wireprotocol.DataType;
+import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.object.ICorfuSMR;
 import org.corfudb.runtime.object.transactions.TransactionType;
-import org.corfudb.runtime.object.transactions.TransactionalContext;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.CacheOption;
 import org.corfudb.runtime.view.StreamsView;
@@ -130,37 +132,41 @@ public class CheckpointWriter<T extends Map> {
      * @return Token at which the snapshot for this checkpoint was taken.
      */
     public Token appendCheckpoint() {
-        // Queries the sequencer for the global log tail. We then read the global log tail to
-        // persist the entry on the log unit in turn preventing from a new sequencer from
+        // Queries the sequencer for the stream tail. We then read the stream tail to
+        // persist the entry on the log unit in turn preventing a new sequencer from
         // regressing tokens.
-        Token snapshot = rt.getSequencerView().query().getToken();
-        if (Address.isAddress(snapshot.getSequence())) {
-            appendCheckpointOnSnapshot(snapshot);
-        }
-        return snapshot;
-    }
-
-    /**
-     * Append checkpoint at a specific snapshot. This is visible only for testing.
-     *
-     * @param snapshotTimestamp snapshot at which to take a checkpoint.
-     * @return snapshot time for this checkpoint.
-     */
-    @VisibleForTesting
-    public Token appendCheckpoint(Token snapshotTimestamp) {
-        appendCheckpointOnSnapshot(snapshotTimestamp);
-        return snapshotTimestamp;
+        Token snapshot = rt.getSequencerView().query(streamId).getToken();
+        return appendCheckpoint(snapshot);
     }
 
     /**
      * Write a checkpoint which reflects the state at snapshot.
      *
-     * @param snapshotTimestamp snapshot at which the checkpoint is taken.
+     * This API should not be directly invoked.
+     *
+     *  @param snapshotTimestamp snapshot at which the checkpoint is taken.
      */
-    private void appendCheckpointOnSnapshot(Token snapshotTimestamp) {
+    @VisibleForTesting
+    public Token appendCheckpoint(Token snapshotTimestamp) {
         long start = System.currentTimeMillis();
 
-        rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
+        // If snapshotTimestamp (stream tail) is a negative value
+        // we need to enforce a hole to ensure the start log address
+        // is not negative (as MultiCheckpointWriter will select the min of all checkpoints)
+        // Having a negative value will prevent trimming from happening.
+        if (!Address.isAddress(snapshotTimestamp.getSequence())) {
+            snapshotTimestamp = forceHole();
+        }
+
+        try {
+            rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
+        } catch (TrimmedException te) {
+            // Stream Tail was trimmed, we can just ignore this trim. This means
+            // no other update has been done to this stream since trim.
+            log.info("appendCheckpointOnSnapshot: stream tail at {} has been trimmed. Checkpoint on same state" +
+                    "as previous checkpoint.", snapshotTimestamp);
+        }
+
         rt.getObjectsView().TXBuild()
                 .type(TransactionType.SNAPSHOT)
                 .snapshot(snapshotTimestamp)
@@ -172,7 +178,11 @@ public class CheckpointWriter<T extends Map> {
             log.info("appendCheckpoint: Started checkpoint for {} at snapshot {}", streamId, snapshotTimestamp);
             ICorfuSMR<T> corfuObject = (ICorfuSMR<T>) this.map;
             Set<Map.Entry> entries = this.map.entrySet();
-            long vloVersion = corfuObject.getCorfuSMRProxy().getVersion();
+            // The vloVersion which will determine the checkpoint START_LOG_ADDRESS (last observed update for this
+            // stream by the time of checkpointing) is defined by the stream's tail instead of the stream's version,
+            // as the latter discards holes for resolution, hence if last address is a hole it would diverge
+            // from the stream address space maintained by the sequencer.
+            long vloVersion = snapshotTimestamp.getSequence();
             startCheckpoint(snapshotTimestamp, vloVersion);
             appendObjectState(entries);
             finishCheckpoint();
@@ -184,6 +194,16 @@ public class CheckpointWriter<T extends Map> {
         } finally {
             rt.getObjectsView().TXEnd();
         }
+
+        return snapshotTimestamp;
+    }
+
+    private Token forceHole() {
+        Token writeToken = rt.getSequencerView().next(streamId).getToken();
+        LogData logData = new LogData(DataType.HOLE);
+        logData.useToken(writeToken);
+        rt.getAddressSpaceView().write(writeToken, logData, CacheOption.WRITE_AROUND);
+        return writeToken;
     }
 
     /** Append a checkpoint START record to this object's stream.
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -184,6 +200,15 @@ private void appendCheckpointOnSnapshot(Token snapshotTimestamp) {
         } finally {
             rt.getObjectsView().TXEnd();
         }
+
+        return snapshotTimestamp;
+    }
+
+    private Token forceHole() {
+        Token writeToken = rt.getSequencerView().next(streamId).getToken();
+        LogData hole = LogData.getHole(writeToken);
```
Maithem: Its better to use `LogData::useToken` instead of `LogData.getHole`

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -172,7 +188,7 @@ private void appendCheckpointOnSnapshot(Token snapshotTimestamp) {
             log.info("appendCheckpoint: Started checkpoint for {} at snapshot {}", streamId, snapshotTimestamp);
             ICorfuSMR<T> corfuObject = (ICorfuSMR<T>) this.map;
             Set<Map.Entry> entries = this.map.entrySet();
-            long vloVersion = corfuObject.getCorfuSMRProxy().getVersion();
+            long vloVersion = snapshotTimestamp.getSequence();
```
Maithem: This needs a comment about why its reasonable to "change" the vlo version

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
```diff
@@ -1002,4 +1002,78 @@ public void testStreamRebuiltWithHoleAsFirstEntryAfterTrimNoCP() throws Exceptio
...
+        // Write 10 Entries to streamA
+        for (int i = 0; i < numEntries; i++) {
+            mA.put(String.valueOf(i), i);
+        }
+
+        // Force a hole as the last update to streamA
+        Token token = runtime.getSequencerView().next(CorfuRuntime.getStreamID(streamA)).getToken();
+        LogData hole = LogData.getHole(token);
+        runtime.getLayoutView().getRuntimeLayout()
+                .getLogUnitClient("tcp://localhost:9000")
```
Maithem: You can use a NodeLocator here.

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
```diff
@@ -1002,4 +1002,78 @@ public void testStreamRebuiltWithHoleAsFirstEntryAfterTrimNoCP() throws Exceptio
...
+        CheckpointWriter cpwB = new CheckpointWriter(runtime, CorfuRuntime.getStreamID(streamB),
+                "checkpointer-Test", mB);
+        cpwB.appendCheckpoint();
+
+        // Add an update to streamA after checkpoint
+        mA.put(String.valueOf(numEntries), numEntries);
+
+        // Trim the log at B's CPToken
+        runtime.getAddressSpaceView().prefixTrim(cpTokenA);
+
```
Maithem: Restart, or flush the server caches after the trim

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
```diff
@@ -1002,4 +1002,78 @@ public void testStreamRebuiltWithHoleAsFirstEntryAfterTrimNoCP() throws Exceptio
...
+        // Instantiate streamA from new Runtime, so stream is rebuilt
+        CorfuRuntime rt2 = new CorfuRuntime(DEFAULT_ENDPOINT).connect();
+        Map<String, Integer> mA2 = createMap(rt2, streamA);
+
+        // By accessing the map, we ensure we are able to load from the checkpoint.
+        try {
+            rt2.getObjectsView().TXBuild()
+                    .type(TransactionType.OPTIMISTIC)
```
Maithem: Can you also check the data ?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -567,7 +577,7 @@ public void MultiCheckpointWriterTestInner(boolean consecutiveCkpoints) throws E
         mcw2.addMap((SMRMap) mA);
         mcw2.addMap((SMRMap) mB);
         long firstGlobalAddress2 = mcw2.appendCheckpoints(r, author).getSequence();
-        assertThat(firstGlobalAddress2).isGreaterThan(firstGlobalAddress1);
+        assertThat(firstGlobalAddress2).isGreaterThanOrEqualTo(firstGlobalAddress1);
```
Maithem: Why isn't this a more strict check ? (i.e. just equality)

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
```diff
@@ -1002,4 +1002,78 @@ public void testStreamRebuiltWithHoleAsFirstEntryAfterTrimNoCP() throws Exceptio
...
+    /**
+     * Test rebuilding a stream from a new runtime, whenever the last address before the checkpoint
+     * was a hole.
+     *
+     * This case is interesting to test as addresses that become holes are discarded by the streamView,
+     * while sequencer's are agnostic of this info, hence, take it into account for trim mark computation.
+     */
+    @Test
+    public void testCheckpointWhenLastAddressIsHole() throws Exception {
+        final int numEntries = 10;
```
Maithem: Do we need a test that creates an empty checkpoint and restarts the server to verify that the sequencer bootstrap state is correct?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -148,19 +146,37 @@ public Token appendCheckpoint() {
...
+        // Note: we could simply not checkpoint empty streams with the introduction
+        // of address maps, but for now, we'll keep it consistent with old backpointer
+        // strategy, which required checkpoints for empty streams to avoid TrimmedExceptions
+        if (!Address.isAddress(snapshotTimestamp.getSequence())) {
+            snapshotTimestamp = forceHole();
+        }
+
+        try {
+            rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
+        } catch (TrimmedException te) {
```
Maithem: Do you really need this? The idea of writing a whole is to mark a slot in the address space to prevent the sequencer regressing, but if its trimmed, then we can't regress from a trimmed address, right?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -148,19 +146,37 @@ public Token appendCheckpoint() {
...
+        // If snapshotTimestamp (stream tail) is a negative value
+        // we need to enforce a hole to ensure the start log address
+        // is not negative (as MultiCheckpointWriter will select the min of all checkpoints)
+        // Having a negative value will prevent trimming from happening.
+
+        // Note: we could simply not checkpoint empty streams with the introduction
+        // of address maps, but for now, we'll keep it consistent with old backpointer
+        // strategy, which required checkpoints for empty streams to avoid TrimmedExceptions
+        if (!Address.isAddress(snapshotTimestamp.getSequence())) {
+            snapshotTimestamp = forceHole();
```
lixinchengdu: Is it better to define another type of LogData to represent dummy data but has backpointers? If we still want checkpoint to be compatible with back-pointer mechanism, we should avoid holes if possible.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
annym: Done.

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
annym: Done.

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
annym: Done.

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
annym: Sure, this is a good idea, cause the test we have does not exercise reboot. Adding it.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
annym: Because this method serves the purpose of two different tests (L.571) so in the case of consecutive checkpoints it will be greater cause the stream tail will change, if not, it will be equal cause the stream tail won't change.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: We do need it, because consider the case where the stream has not been updated for several checkpoint cycles, therefore the stream tail has not changed, but after a couple of cycles let's say it is trimmed and entirely subsumed by a checkpoint. This read will not serve the purpose of hole fill in that case and will throw a TrimmedException, cause it is no longer present.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: I don't really think we need that, anyways our goal is to completely move away from back pointers. And address maps does not have this issue.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: I misunderstood this comment, thinking it was related to catching the TrimmedException but I believe it was actually related to forcing a hole in these cases. Which I agree with, there is no case in filling a hole as there cannot be regression in this case. Removing this!

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -148,19 +147,37 @@ public Token appendCheckpoint() {
      */
     @VisibleForTesting
     public Token appendCheckpoint(Token snapshotTimestamp) {
-        appendCheckpointOnSnapshot(snapshotTimestamp);
-        return snapshotTimestamp;
+        return appendCheckpointOnSnapshot(snapshotTimestamp);
```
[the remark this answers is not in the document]
annym: appendCheckpoint is public and only visible for testing.. this other is private.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -130,14 +132,11 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
      * @return Token at which the snapshot for this checkpoint was taken.
      */
     public Token appendCheckpoint() {
-        // Queries the sequencer for the global log tail. We then read the global log tail to
+        // Queries the sequencer for the stream tail. We then read the stream tail to
         // persist the entry on the log unit in turn preventing from a new sequencer from
```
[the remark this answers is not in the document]
annym: done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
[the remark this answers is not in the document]
WenbinZhu: @annym  We used to have two different methods because they have different return type, but now they are exactly same.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem: Instead of reassigning the vloVersion variable, you can check if long vloVersion == corfuObject.getCorfuSMRProxy().getVersion(); and maybe log something ? do we even care?

PR #2074, opened 2019-07-30
Add Detailed Message to IOException

annym:
## Overview
 
Improve debugging.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ICorfuPayload.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ICorfuPayload.java
@@ -93,7 +93,7 @@ public interface ICorfuPayload<T> {
                             return new StreamAddressSpace(trimMark, map);
                         } catch (IOException ioe) {
                             throw new SerializerException("Exception when attempting to " +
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/protocols/wireprotocol/ICorfuPayload.java
```diff
@@ -93,7 +93,7 @@
                             return new StreamAddressSpace(trimMark, map);
                         } catch (IOException ioe) {
                             throw new SerializerException("Exception when attempting to " +
-                                    "deserialize stream address space.");
+                                    "deserialize stream address space: " + ioe.getMessage());
```
Maithem: Just wrap the ieo with SerializerException, otherwise you lose the stack trace. 

`throw new SerializerException(ioe)`

Maithem: Probably need to overload 

```
public SerializerException(Throwable t) {
        super(t);
    }
```

annym: Sure, it's cleaner. Done.

PR #2075, opened 2019-07-30
Do not depend on mutable state for Paxos

vjeko:
Within Paxos, avoid relying on mutable state (server epoch to be more
precise), since this can lead to wrong data being persisted and agreed
upon.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
@@ -1,5 +1,6 @@
 package org.corfudb.infrastructure;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.netty.channel.ChannelHandlerContext;
 import lombok.Getter;
 import lombok.NonNull;
@@ -22,6 +23,7 @@ import javax.annotation.Nonnull;
 import java.lang.invoke.MethodHandles;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
@@ -137,17 +139,19 @@ public class LayoutServer extends AbstractServer {
         if (!isBootstrapped(msg, ctx, r)) {
             return;
         }
-        long epoch = msg.getPayload();
-        if (epoch <= serverContext.getServerEpoch()) {
+
+        final long msgEpoch = msg.getPayload();
+        final long serverEpoch = getServerEpoch();
+
+        if (msgEpoch <= serverEpoch) {
             r.sendResponse(ctx, msg, new LayoutMsg(getCurrentLayout(), CorfuMsgType
                     .LAYOUT_RESPONSE));
         } else {
             // else the client is somehow ahead of the server.
             //TODO figure out a strategy to deal with this situation
-            long serverEpoch = serverContext.getServerEpoch();
             r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH, serverEpoch));
             log.warn("handleMessageLayoutRequest: Message Epoch {} ahead of Server epoch {}",
-                    epoch, serverEpoch);
+                    msgEpoch, serverEpoch);
         }
     }
 
@@ -198,12 +202,14 @@ public class LayoutServer extends AbstractServer {
         if (!isBootstrapped(msg, ctx, r)) {
             return;
         }
-        Rank prepareRank = new Rank(msg.getPayload().getRank(), msg.getClientID());
-        Rank phase1Rank = getPhase1Rank();
-        Layout proposedLayout = getProposedLayout();
 
-        long serverEpoch = getServerEpoch();
-        if (msg.getPayload().getEpoch() != serverEpoch) {
+        final long payloadEpoch = msg.getPayload().getEpoch();
+        final long serverEpoch = getServerEpoch();
+
+        final Rank prepareRank = new Rank(msg.getPayload().getRank(), msg.getClientID());
+        final Rank phase1Rank = getPhase1Rank(payloadEpoch);
+
+        if (payloadEpoch != serverEpoch) {
             r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH, serverEpoch));
             log.trace("handleMessageLayoutPrepare: Incoming message with wrong epoch, got {}, "
                             + "expected {}, message was: {}",
@@ -211,6 +217,7 @@ public class LayoutServer extends AbstractServer {
             return;
         }
 
+        Layout proposedLayout = getProposedLayout(payloadEpoch);
 
         // This is a prepare. If the rank is less than or equal to the phase 1 rank, reject.
         if (phase1Rank != null && prepareRank.lessThanEqualTo(phase1Rank)) {
@@ -221,9 +228,9 @@ public class LayoutServer extends AbstractServer {
         } else {
             // Return the layout with the highest rank proposed before.
             Rank highestProposedRank = proposedLayout == null ? new Rank(-1L, msg.getClientID())
-                    : getPhase2Rank();
-            setPhase1Rank(prepareRank);
-            log.debug("handleMessageLayoutPrepare: New phase 1 rank={}", getPhase1Rank());
+                    : getPhase2Rank(payloadEpoch);
+            setPhase1Rank(prepareRank, payloadEpoch);
+            log.debug("handleMessageLayoutPrepare: New phase 1 rank={}", prepareRank);
             r.sendResponse(ctx, msg, CorfuMsgType.LAYOUT_PREPARE_ACK.payloadMsg(new
                     LayoutPrepareResponse(highestProposedRank.getRank(), proposedLayout)));
         }
@@ -246,17 +253,18 @@ public class LayoutServer extends AbstractServer {
         if (!isBootstrapped(msg, ctx, r)) {
             return;
         }
-        Rank proposeRank = new Rank(msg.getPayload().getRank(), msg.getClientID());
-        Rank phase1Rank = getPhase1Rank();
 
-        long serverEpoch = getServerEpoch();
+        final long payloadEpoch = msg.getPayload().getEpoch();
+        final long serverEpoch = getServerEpoch();
+
+        final Rank proposeRank = new Rank(msg.getPayload().getRank(), msg.getClientID());
+        final Rank phase1Rank = getPhase1Rank(payloadEpoch);
 
         // Check if the propose is for the correct epoch
-        if (msg.getPayload().getEpoch() != serverEpoch) {
+        if (payloadEpoch != serverEpoch) {
             r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH, serverEpoch));
             log.trace("handleMessageLayoutPropose: Incoming message with wrong epoch, got {}, "
-                            + "expected {}, message was: {}",
-                    msg.getPayload().getEpoch(), serverEpoch, msg);
+                            + "expected {}, message was: {}", payloadEpoch, serverEpoch, msg);
             return;
         }
         // This is a propose. If no prepare, reject.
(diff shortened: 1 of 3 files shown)


vjeko:
This is an early preview. Test(s) are coming.

vjeko:
Added some tests that trigger the race condition.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -198,19 +202,22 @@ public synchronized void handleMessageLayoutPrepare(
...
 
-        long serverEpoch = getServerEpoch();
-        if (msg.getPayload().getEpoch() != serverEpoch) {
+        final long msgEpoch = msg.getPayload().getEpoch();
+        final long serverEpoch = getServerEpoch();
+
+        final Rank prepareRank = new Rank(msg.getPayload().getRank(), msg.getClientID());
+        final Rank phase1Rank = getPhase1Rank(msgEpoch);
+
+        if (msgEpoch != serverEpoch) {
```
Maithem: Are these checks really necessary ? 

Its possible that this condition is false, at the time of the check, but becomes true immediately after the evaluation.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -221,9 +228,9 @@ public synchronized void handleMessageLayoutPrepare(
         } else {
             // Return the layout with the highest rank proposed before.
             Rank highestProposedRank = proposedLayout == null ? new Rank(-1L, msg.getClientID())
-                    : getPhase2Rank();
-            setPhase1Rank(prepareRank);
-            log.debug("handleMessageLayoutPrepare: New phase 1 rank={}", getPhase1Rank());
+                    : getPhase2Rank(msgEpoch);
```
Maithem: I think the wrappers around the PaxosDataStore methods are superfluous. I thought one of the reasons we factored out those operations was to just consume the PaxosDataStore directly.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -384,18 +402,18 @@ public void setCurrentLayout(Layout layout) {
...
-    public void setPhase1Rank(Rank rank) {
-        paxosDataStore.setPhase1Rank(rank, getServerEpoch());
+    public void setPhase1Rank(Rank rank, long epoch) {
+        paxosDataStore.setPhase1Rank(rank, epoch);
     }
 
-    public void setPhase2Data(Phase2Data phase2Data) {
-        paxosDataStore.setPhase2Data(phase2Data, getServerEpoch());
+    public void setPhase2Data(Phase2Data phase2Data, long epoch) {
+        paxosDataStore.setPhase2Data(phase2Data, epoch);
```
Maithem: I think its worthwhile to add some sanity checks here, for example `assert phase2Data.getLayout().getEpoch() == epoch` . Actually, if we had this check this the problem wouldn't have happened.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
vjeko: We will. I have a bigger patch that addresses that. There are other issues with this code that do not necessarily relate to the bug at hand.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
Maithem: Its cleaner to do these checks here:

```java
    LAYOUT_PREPARE(12, new TypeToken<CorfuPayloadMsg<LayoutPrepareRequest>>(){}, true),
    LAYOUT_PROPOSE(14, new TypeToken<CorfuPayloadMsg<LayoutProposeRequest>>(){}, true),
```

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
vjeko: Whether you do `assert phase2Data.getLayout().getEpoch() == epoch`or `if (msgEpoch != serverEpoch) return NACK` does not really matter since both of them are doing the same thing -- ensuring that epoch does not change between the time we do the validation and and the time we persist the accepted data.

Ideally all Paxos phases should base their decision on immutable state. So enforcing immutability at the beginning of each round seems cleaner to me.

Maithem: Its not about immutability on the server side. If there's a race condition on the client side -- where the message is being stamped with an epoch that is different than the layout's epoch (inside the message), then pre-check doesn't matter (i.e. it won't be detected).

vjeko: If you look at `LayoutView`, you will notice that within `updateLayout(Layout layout, long rank)` we do:

``` java
        long epoch = layout.getEpoch();
        ...
        //phase 1: prepare with a given rank.
        Layout alreadyProposedLayout = prepare(epoch, rank);
        ...
        //phase 2: propose the new layout.
        propose(epoch, rank, layoutToPropose);
        //phase 3: committed
        committed(epoch, layoutToPropose);
```

Which means that you can carry either:

1) Your layout to phase II. We know that these cannot be different since we extract the epoch directly from the layout: `long epoch = layout.getEpoch()`.
2) Already accepted layout to phase II. This one also cannot happen since we make sure to return the layout associated with the provided epoch, which cannot be different because of the previous proposition.

``` java
            // Return the layout with the highest rank proposed before.
            Rank highestProposedRank = proposedLayout == null ? new Rank(-1L, msg.getClientID())
                    : getPhase2Rank(msgEpoch);
```

Which means there is no way for the client for send us a layout with a different epoch. However, I do agree that we should probably check for that invariant as well. It will shield us  from future regressions.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -291,7 +301,7 @@ public synchronized void handleMessageLayoutPropose(
 
         log.debug("handleMessageLayoutPropose: New phase 2 rank={},  layout={}",
                 proposeRank, proposeLayout);
-        setPhase2Data(new Phase2Data(proposeRank, proposeLayout));
+        setPhase2Data(new Phase2Data(proposeRank, proposeLayout), msgEpoch);
```
WenbinZhu: Is this correct? The `msgEpoch` can still be different from the epoch in `proposeLayout`, then we will still have the inconsistent epoch issue, right?

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -347,19 +362,22 @@ public synchronized void handleMessageLayoutCommit(
...
-        Layout commitLayout = msg.getPayload().getLayout();
         if (!isBootstrapped(msg, ctx, r)) {
             return;
         }
-        long serverEpoch = getServerEpoch();
-        if (msg.getPayload().getEpoch() < serverEpoch) {
+
+        final long msgEpoch = msg.getPayload().getEpoch();
+        final long serverEpoch = getServerEpoch();
+        final Layout msgLayout = msg.getPayload().getLayout();
```
WenbinZhu: I think the original name `commitLayout` is more descriptive, could you change it back?

test/src/test/java/org/corfudb/infrastructure/LayoutServerTest.java
```diff
@@ -464,6 +522,53 @@ public void testReboot() throws Exception {
...
+                .build();
+
+        // set up base server
+        BaseServer bs = new BaseServer(sc);
+
+        IServerRouter routerMock = Mockito.mock(IServerRouter.class);
+        ChannelHandlerContext chcMock = Mockito.mock(ChannelHandlerContext.class);
+
+        // set up layout server
+        LayoutServer ls = new LayoutServer(sc);
```
vjeko: Pavel, just one note -- having more descriptive variable names would be great. Using two letter variable names makes the code quite unreadable IMO.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -221,9 +228,9 @@ public synchronized void handleMessageLayoutPrepare(
         } else {
             // Return the layout with the highest rank proposed before.
             Rank highestProposedRank = proposedLayout == null ? new Rank(-1L, msg.getClientID())
-                    : getPhase2Rank();
-            setPhase1Rank(prepareRank);
-            log.debug("handleMessageLayoutPrepare: New phase 1 rank={}", getPhase1Rank());
+                    : getPhase2Rank(msgEpoch);
+            setPhase1Rank(prepareRank, msgEpoch);
+            log.debug("handleMessageLayoutPrepare: New phase 1 rank={}", getPhase1Rank(msgEpoch));
```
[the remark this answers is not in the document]
vjeko: I tried no to do code restructuring, but sure, I'll change it.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
vjeko: There is already a comment that describes why this will not happen: https://github.com/CorfuDB/CorfuDB/pull/2075/files#r308951272

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
vjeko: I changed it because it was ambiguous to begin with -- commitLayout implies that the layout has been committed, which is not the case. The other alternative is to use layoutToBeCommited, but the name was too long.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
vjeko: I see where the confusion is coming from. There are three (!) epoch in that message:

1) The layout epoch.
2) The payload epoch.
3) The message epoch.

We are not relying on the message epoch. (1) and (2) should probably be collapsed into one, but that is future work.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
vjeko: As discussed, I changed the name to resolve the ambiguity.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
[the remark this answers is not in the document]
vjeko: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
vjeko: Let me address this whole epoch mess in the next patch. What you are referring to is the message and not LayoutPrepareRequest/LayoutProposeRequest epoch.

test/src/test/java/org/corfudb/infrastructure/LayoutServerTest.java
PavelZaytsev: Fixed

vjeko: Thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -276,8 +284,19 @@ public synchronized void handleMessageLayoutPropose(
...
         }
-        Rank phase2Rank = getPhase2Rank();
-        Layout proposeLayout = msg.getPayload().getLayout();
+
+        final Rank phase2Rank = getPhase2Rank(payloadEpoch);
+        final Layout proposeLayout = msg.getPayload().getLayout();
+
+        // Make sure that the layout epoch is the same as the LayoutProposeRequest epoch.
+        if (proposeLayout.getEpoch() != payloadEpoch) {
+            log.debug("Phase II error: layout {} and payload {} epoch should be the same",
```
zalokhan: Prepend log message with `handleMessageLayoutPropose` to keep it consistent with the rest of the method.

vjeko: I will be changing it in the next patch. There are a lot of formatting issues in this file.

test/src/test/java/org/corfudb/infrastructure/LayoutServerTest.java
```diff
@@ -464,6 +464,73 @@ public void testReboot() throws Exception {
...
+        spyLayoutServer.handleMessageLayoutPropose(layoutProposeRequestMsg, context, router);
+
+        // Make sure the epoch has actually changed.
+        Assertions.assertThat(newSealEpoch)
+                .isEqualTo(spyLayoutServer.getServerContext().getServerEpoch());
+
+        // Make sure no data has been accepted for the new epoch.
+        Assertions.assertThat(spyLayoutServer.getPhase2Data(newSealEpoch).isPresent())
+                .isFalse();
+        // Make sure no data has been accepted for the correct epoch.
```
zalokhan: Make sure **data has been** accepted

PR #2078, opened 2019-07-30
Remove backpointersDisabled

Maithem:
## Overview
Only one variable is needed to toggle backpointers, so I removed
backpointersDisabled for followBackpointersEnabled.

Why should this be merged: Removes a redundant and confusing option. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -86,17 +86,6 @@ public class CorfuRuntime {
         private final long nettyShutdownTimeout = 300;
 
         // region Object Layer Parameters
-        /**
-         * True, if undo logging is disabled.
-         */
-        @Default
-        boolean undoDisabled = false;
-
-        /**
-         * True, if optimistic undo logging is disabled.
-         */
-        @Default
-        boolean optimisticUndoDisabled = false;
 
         /**
          * Max size for a write request.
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -309,7 +309,7 @@ private boolean isAddressToBackpointerResolved(long startAddress, UUID streamId)
             }
         }
 
-        if (!runtime.getParameters().isBackpointersDisabled() && d.hasBackpointer(streamId)) {
```
Maithem: @annym why do we even care about backpointers here?

annym: This is an optimization to avoid going to the sequencer if the remaining space is locally resolved already, but the check on that flag was just to be consistent with the fact that it indicated that backpointers might not exist, given that it is always the case we don't need it.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -187,12 +187,6 @@
         @Default
         boolean followBackpointersEnabled = false;
 
-        /**
-         * Whether or not to disable backpointers.
```
Maithem: remove undoDisabled and optimisticUndoDisabled

Maithem:
Let's get some reviews here.

PR #2079, opened 2019-07-31
MultiObjectSMREntry Lazy Evaluation

Maithem:
## Overview
Since getPayload will deserialize all streams in MultiObjectSMREntry, the client needs to register all the corresponding stream's serializers even the ones that it doesn't care about, which can be problematic. This PR, lazily deserializes the MultiObjectSMREntry and therefore a client only needs to register serializers for streams that it opened.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
@@ -2,12 +2,14 @@ package org.corfudb.protocols.logprotocol;
 
 import io.netty.buffer.ByteBuf;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
+import io.netty.buffer.Unpooled;
 import lombok.Getter;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
@@ -27,28 +29,14 @@ import org.corfudb.util.serializer.Serializers;
 public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
 
     // map from stream-ID to a list of updates encapsulated as MultiSMREntry
+    //TODO(Maithem): we should hide entryMap
     @Getter
-    public Map<UUID, MultiSMREntry> entryMap = Collections.synchronizedMap(new HashMap<>());
+    public Map<UUID, List<SMREntry>> entryMap = Collections.synchronizedMap(new HashMap<>());
+
+    private Map<UUID, byte[]> streamBuffers = new HashMap<>();
 
     public MultiObjectSMREntry() {
         this.type = LogEntryType.MULTIOBJSMR;
-    }
-
-    public MultiObjectSMREntry(Map<UUID, MultiSMREntry> entryMap) {
-        this.type = LogEntryType.MULTIOBJSMR;
-        this.entryMap = entryMap;
-    }
-
-    /** Extract a particular stream's entry from this object.
-     *
-     * @param streamID StreamID
-     * @return the MultiSMREntry corresponding to streamId
-     */
-    protected MultiSMREntry getStreamEntry(UUID streamID) {
-        return getEntryMap().computeIfAbsent(streamID, u -> {
-                    return new MultiSMREntry();
-                }
-        );
     }
 
     /**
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -83,13 +75,14 @@ public void mergeInto(MultiObjectSMREntry other) {
...
-        for (int i = 0; i < numUpdates; i++) {
-            entryMap.put(
-                    new UUID(b.readLong(), b.readLong()),
-                    ((MultiSMREntry) Serializers.CORFU.deserialize(b, rt)));
+        int numStreams = b.readInt();
+        for (int i = 0; i < numStreams; i++) {
+            UUID streamId = new UUID(b.readLong(), b.readLong());
+
+            int updateLength = b.readInt();
+            byte[] streamUpdates = new byte[updateLength];
```
Maithem: This adds an extra copy, but also reduces the memory overhead when not all stream updates are needed. 
If we know all open streams, then we can omit the copy and directly deserialize only the opened streams.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -111,10 +112,20 @@ public void serialize(ByteBuf b) {
      * @return an empty list if object has no updates; a list of updates if exists
      */
     @Override
-    public List<SMREntry> getSMRUpdates(UUID id) {
-        MultiSMREntry entry = entryMap.get(id);
-        return entryMap.get(id) == null ? Collections.emptyList() :
-                entry.getUpdates();
+    public synchronized List<SMREntry> getSMRUpdates(UUID id) {
```
Maithem: Multiple VLOs can access this object at the same time, but only after it has been committed, so this should be thread-safe.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -57,7 +45,9 @@ protected MultiSMREntry getStreamEntry(UUID streamID) {
      * @param updateEntry SMREntry to add
      */
     public void addTo(UUID streamID, SMREntry updateEntry) {
-        getStreamEntry(streamID).addTo(updateEntry);
+        List<SMREntry> updates = entryMap.getOrDefault(streamID, new ArrayList<>());
```
Maithem: Before this is committed, it will be owned by a transaction (i.e. contained in thread local storage), so adding to this collection should be thread safe.

hisundar: These read as pretty useful comments to me, any reason why they cannot be comments in the code itself?

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
hisundar: Seems like this should be a comment in the code and not just in GitHub?

Maithem: This was a reminder, will add the comments to the source.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -27,28 +29,14 @@
 public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
 
     // map from stream-ID to a list of updates encapsulated as MultiSMREntry
+    //TODO(Maithem): we should hide entryMap
     @Getter
-    public Map<UUID, MultiSMREntry> entryMap = Collections.synchronizedMap(new HashMap<>());
+    public Map<UUID, List<SMREntry>> entryMap = Collections.synchronizedMap(new HashMap<>());
```
Maithem: @WenbinZhu is this what you meant when you talked about collapsing MultiSMREntry and MultiObjectSMREntry ?

WenbinZhu: Not really, the collapsing I was referring to is to use `MultiObjectSMREntry` for both transaction and non-transaction, instead of the current `MultiObjectSMREntry` for transaction and, `SMREntry` for non-transaction and `MultiSMREntry` for checkpoint. But yes, one result of this kind of collapsing is `MultiSMREntry` will be completely removed.

Maithem:
A cleaner solution is to implement direct stream reads.

zalokhan:
Leaving this untouched for history keeping. Re-opening this as a separate rebased PR.

PR #2081, opened 2019-08-01
State Transfer Stream Address Space Bug Fix

annym:
## Overview

Description:

Correctly build the stream address space of a stream that
is completely checkpointed-- on state transfer
(i.e., a stream built from the checkpoint).

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
@@ -45,7 +45,11 @@ public class LogMetadata {
 
     public void update(List<LogData> entries) {
         for (LogData entry : entries) {
-            update(entry, false);
+            // This API is only used on range writes for state transfer
+            // On state transfer we need to inspect checkpoint streams (initialize = true),
+            // to avoid losing data for streams that are completely checkpointed (no actual
+            // log entry for the regular stream is present in the log).
+            update(entry, true);
         }
     }
 
--- a/test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
+++ b/test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
@@ -4,12 +4,16 @@ import com.google.common.collect.ContiguousSet;
 import com.google.common.collect.DiscreteDomain;
 import com.google.common.collect.Range;
 
+import com.google.common.reflect.TypeToken;
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.ReadResponse;
+import org.corfudb.protocols.wireprotocol.Token;
 import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.runtime.BootstrapUtil;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuRuntime.CorfuRuntimeParameters;
+import org.corfudb.runtime.MultiCheckpointWriter;
 import org.corfudb.runtime.RebootUtil;
 import org.corfudb.runtime.clients.BaseHandler;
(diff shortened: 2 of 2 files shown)


test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -1092,4 +1102,104 @@ public void forceRemoveAfterSeal() throws Exception {
             shutdownCorfuServer(corfuServer);
         }
     }
+
+    /**
+     * Verify that a stream can be correctly loaded after a reconfiguration and
+     * all the updates on this stream are subsumed by a checkpoint and trimmed.
+     */
+    @Test
+    public void testFailoverWithStreamSubsumedByCheckpoint() throws Exception {
```
Maithem: Should also verify that the response for LOG_ADDRESS_SPACE_REQUEST on all servers is equal.

I think we need a more general test case, but this might not be the best place to write it.

infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
```diff
@@ -45,7 +45,11 @@ public LogMetadata() {
 
     public void update(List<LogData> entries) {
         for (LogData entry : entries) {
-            update(entry, false);
+            // This API is only used on range writes for state transfer
+            // On state transfer we need to inspect checkpoint streams (initialize = true),
+            // to avoid losing data for streams that are completely checkpointed (no actual
+            // log entry for the regular stream is present in the log).
+            update(entry, true);
```
Maithem: Can we try to remove the special handling for initialization vs post-initialization ?

annym: As we discussed yesterday, we can do that, but that would require changes in the stream layer to not re-sync when data is still available because it has not actually been trimmed yet. Actually maybe it's the correct way to solve it, cause if state transfer happens before the trim, we could run into this issue as well. I'll do this change accordingly.

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -1092,4 +1102,104 @@ public void forceRemoveAfterSeal() throws Exception {
...
+
+        CorfuTable<String, String> table2 = runtime2.getObjectsView().build()
+                .setTypeToken(new TypeToken<CorfuTable<String, String>>() {})
+                .setStreamID(streamId)
+                .open();
+
+        // Verify sequencer has correct address map for this stream (addresses and trim mark)
+        StreamAddressSpace addressSpace = runtime2.getAddressSpaceView()
+                .getLogAddressSpace()
+                .getAddressMap().get(CorfuRuntime.getStreamID(streamName));
```
WenbinZhu: @annym You can directly use the `streamId` variable here.

annym: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/LogMetadata.java
annym: I was attempting to do this, but actually It won't work cause we could have data loss (if I try to optimize for these cases in the stream layer). So I believe it is better to pay the cost of a reset for state transfer scenarios only (when streams are subsumed by a checkpoint), instead of paying it for all cases (not only subsumed streams) when removing the initialize path which would enforce stream resets on regular operation (appends), state transfer and sequencer reset.

medhavidhawan:
I think we should get rid of special casing +1

annym:
> Having special handling for initialization/post-initialization is error prone.This PR seems to be fixing an issue. But let's have a discussion offline, maybe we can get rid of this special case.

@Maithem @medhavidhawan sure, we can discuss this offline and get rid of the special case. My. only concern is that we'll pay the cost of stream resets on every access if we don't do something else, maybe there's another way to approach this.

PR #2084, opened 2019-08-02
TrimmedException log level message - 0.2.3

annym:
Change log level from WARN to INFO and do not print stack trace, as this exception
is properly handled by upper layers and accordingly printed if leading to an actual exception.

Port to corfu-0.2.3


hisundar:
Lgtm

PR #2087, opened 2019-08-02
Rename SET_EPOCH to SEAL

Maithem:
## Overview
The set epoch operation is in fact a seal operation. Set epoch
might imply arbitrarily setting an epoch, which is not the case.

Why should this be merged: Improves the nomenclature  

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/clients/BaseClient.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/BaseClient.java
@@ -15,7 +15,7 @@ import org.corfudb.protocols.wireprotocol.VersionInfo;
  * This is a base client which sends basic messages.
  * It mainly sends PINGs, as well as the ACK/NACKs defined by
  * the Corfu protocol.
- * This is also responsible to send SET_EPOCH messages used to seal the servers with an epoch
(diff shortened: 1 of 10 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/BaseClient.java
```diff
@@ -57,9 +57,9 @@ public boolean pingSync() {
      * @param newEpoch New Epoch to be set
      * @return Completable future which returns true on successful epoch set.
      */
-    public CompletableFuture<Boolean> setRemoteEpoch(long newEpoch) {
-        CorfuMsg msg = new CorfuPayloadMsg<>(CorfuMsgType.SET_EPOCH, newEpoch).setEpoch(epoch);
-        log.info("setRemoteEpoch: send SET_EPOCH from me(clientId={}) to new epoch {}",
+    public CompletableFuture<Boolean> sealRemote(long newEpoch) {
```
WenbinZhu: Maybe sealServer or sealRemoteServer is more appropriate?

Maithem: Done.

Issue #2080, opened 2019-08-01
Circular dependency between LogEntry and LogData.


WenbinZhu:
## Overview

LogEntry is the payload of LogData, however LogEntry also contains a reference to LogData. This is unnecessary because the only reason to have this backward dependency is LogEntry needs to know about the global address inside LogData. We should remove this dependency as it harms understandability and error prone.

PR #1983, opened 2019-06-28
Get rid of Runtime dependency in payload deserialization.

WenbinZhu:
# Overview

For some historical reasons, the deserialization of LogData's payload
depends on CorfuRuntime. This patch gets rid of this dependency.

Related issue(s) (if applicable): #1880

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #2090, opened 2019-08-02
ILogData: Get rid of backward dependency on LogEntry.

WenbinZhu:
## Overview

- LogEntry is the payload of LogData, however LogEntry also contains a
reference to LogData, which undermines readability and is error prone.
This is also unnecessary because the only reason to have this backward
dependency is LogEntry needs to know about the global address inside
LogData, which is consumed by VLO. This patch removed this dependency
by setting globalAddress directly.

- This is also a prerequisite a following patch that unifies SMREntry, MultiSMREntry
and MultiObjectSMREntry.

Related issue(s) (if applicable): #2080 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiSMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiSMREntry.java
@@ -73,9 +73,9 @@ public class MultiSMREntry extends LogEntry implements ISMRConsumable {
     }
 
     @Override
-    public void setEntry(ILogData entry) {
-        super.setEntry(entry);
-        this.getUpdates().forEach(x -> x.setEntry(entry));
+    public void setGlobalAddress(long address) {
+        super.setGlobalAddress(address);
+        this.getUpdates().forEach(x -> x.setGlobalAddress(address));
     }
 
     @Override
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
@@ -47,7 +47,7 @@ public interface IMetadata {
      */
     @SuppressWarnings("unchecked")
     default Set<UUID> getStreams() {
-        return (Set<UUID>) ((Map<UUID, Long>)getMetadataMap().getOrDefault(
+        return ((Map<UUID, Long>) getMetadataMap().getOrDefault(
                 LogUnitMetadataType.BACKPOINTER_MAP, Collections.emptyMap())).keySet();
     }
 
@@ -131,12 +131,8 @@ public interface IMetadata {
      */
     @SuppressWarnings("unchecked")
     default Long getGlobalAddress() {
-        if (getMetadataMap() == null
-                || getMetadataMap().get(LogUnitMetadataType.GLOBAL_ADDRESS) == null) {
-            return -1L;
-        }
         return Optional.ofNullable((Long) getMetadataMap()
-                .get(LogUnitMetadataType.GLOBAL_ADDRESS)).orElse((long) -1);
+                .get(LogUnitMetadataType.GLOBAL_ADDRESS)).orElse(Address.NON_ADDRESS);
     }
 
     /**
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -76,9 +76,8 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
                         ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
                         final Object actualValue =
                                 Serializers.CORFU.deserialize(copyBuf, runtime);
-                        // TODO: Remove circular dependency on logEntry.
                         if (actualValue instanceof LogEntry) {
-                            ((LogEntry) actualValue).setEntry(this);
+                            ((LogEntry) actualValue).setGlobalAddress(getGlobalAddress());
                             ((LogEntry) actualValue).setRuntime(runtime);
                         }
                         value = actualValue == null ? this.payload : actualValue;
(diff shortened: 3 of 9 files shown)


runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -174,9 +173,6 @@ public LogData(DataType type, final Object object) {
             this.type = type;
             this.data = null;
             this.payload.set(object);
-            if (object instanceof LogEntry) {
-                ((LogEntry) object).setEntry(this);
```
WenbinZhu: With this patch, we cannot setGlobalAddress here because by the time LogData is constructed, the address (token) is not yet assigned.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiSMREntry.java
```diff
@@ -73,9 +73,9 @@ public void serialize(ByteBuf b) {
     }
 
     @Override
-    public void setEntry(ILogData entry) {
-        super.setEntry(entry);
-        this.getUpdates().forEach(x -> x.setEntry(entry));
+    public void setGlobalAddress(long address) {
+        super.setGlobalAddress(address);
```
Maithem: Is it a big change to make globalAddress a final ?

WenbinZhu: I guess so. Not sure if we can do this because by the time a log entry is constructed, the global address is unknown.

WenbinZhu:
Can we have some reviews here? Another GC change is dependent on this PR. Thanks.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
```diff
@@ -131,12 +131,12 @@ default Long getThreadId() {
      */
     @SuppressWarnings("unchecked")
     default Long getGlobalAddress() {
-        if (getMetadataMap() == null
-                || getMetadataMap().get(LogUnitMetadataType.GLOBAL_ADDRESS) == null) {
-            return -1L;
+        if (getMetadataMap() == null) {
+            return Address.NON_ADDRESS;
```
Maithem: I don't think you need this, Its either empyt, or it doesn't have the key.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -76,9 +76,8 @@ public Object getPayload(CorfuRuntime runtime) {
                         ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
                         final Object actualValue =
                                 Serializers.CORFU.deserialize(copyBuf, runtime);
-                        // TODO: Remove circular dependency on logEntry.
                         if (actualValue instanceof LogEntry) {
-                            ((LogEntry) actualValue).setEntry(this);
+                            ((LogEntry) actualValue).setGlobalAddress(getGlobalAddress());
                             ((LogEntry) actualValue).setRuntime(runtime);
```
Maithem: I think Its time to remove this too  `.setRuntime(runtime)`
#1983

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -174,9 +173,6 @@ public LogData(DataType type, final Object object) {
             this.type = type;
             this.data = null;
             this.payload.set(object);
```
Maithem: When this path is invoked (during a transaction write), the global address is not set, so when sync is subsequently called, it will see that the payload is already deserialized (since its in the cache), but it won't call setGlobalAddress. Is it possible that the VLO will see this as uniitalized and thefore thing its an optimistic SMR entry ?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -190,6 +186,23 @@ public LogData(DataType type, final Object object) {
...
+     */
+    @Override
+    public void useToken(IToken token) {
+        setGlobalAddress(token.getSequence());
+        setEpoch(token.getEpoch());
+        if (token.getBackpointerMap().size() > 0) {
+            setBackpointerMap(token.getBackpointerMap());
+        }
+        if (payload.get() instanceof LogEntry) {
+            ((LogEntry) payload.get()).setGlobalAddress(token.getSequence());
```
Maithem: Oh, I see, this will fix the issue i mentioned above.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Yes you're right.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Sure, will do after this PR.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
lixinchengdu: Why useToken is not ILogData default method anymore？

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -477,28 +477,26 @@ public String toString() {
 
 
     /**
-     * Given a SMR entry with an undo record, undo the update.
+     * Given a SMR entry with an undo entry, undo the update.
```
lixinchengdu: Why change the name of record to entry?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Because I need to access `payload`, but `payload` is not in `ILogData` interface.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
WenbinZhu: I renamed because this undoRecord is an SMREntry. In my following patch that unifies all three entry types, `SMREntry` will be rename into `SMRRecord`, then this will be renamed again.

PR #2095, opened 2019-08-06
Use A Reverse Iterator

Maithem:
## Overview
As more entries are added to a stream, the stream tail query (i.e
getAddressesInRange) operation becomes slower and slower, because
the whole bit set has to be iterated. This happens despite the fact
that clients have most of the StreamSpace resolved. This patch
improves the performance by iterating backwards and not iterate
more addresses than needed.

Why should this be merged: Fixes a serious performance issue, which is, clients become slower and slower over time. 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -149,15 +149,17 @@ public void trim(Long trimMark) {
...
-                    addressesInRange.add(address);
-                }
-            });
+        LongIterator iterator = addressMap.getReverseLongIterator();
+
+        while (iterator.hasNext()) {
+            long val = iterator.next();
+            if (val > range.getEnd() && val <= range.getStart()) {
+                addressesInRange.add(val);
+            } else if (val < range.getEnd()) {
```
Maithem: This can be changed to if-else

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -149,15 +149,17 @@ public void trim(Long trimMark) {
...
     public Roaring64NavigableMap getAddressesInRange(StreamAddressRange range) {
         Roaring64NavigableMap addressesInRange = new Roaring64NavigableMap();
-        if (range.getStart() > range.getEnd()) {
-            addressMap.forEach(address -> {
-                // Because our search is referenced to the stream's tail => (end < start]
-                if (address > range.getEnd() && address <= range.getStart()) {
-                    addressesInRange.add(address);
-                }
-            });
+        LongIterator iterator = addressMap.getReverseLongIterator();
```
vjeko: Please put a comment here explaining why reverse iterator is needed.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
annym: Don't change for if-else, otherwise we could break while iterating over the range (end, start] ----BREAK FOR THIS SPACE--- and miss the range...

PR #2099, opened 2019-08-06
Correct CheckpointWriter return Token

annym:
## Overview

Description:

Return snapshot address at the time of checkpoint instead of stream tail address.

Why should this be merged: The checkpointing task will limit the trim task for those cases in which a stream has not observed updates for a while. In other words, the checkpointWriter token could remain the same for multiple checkpoint cycles if the tail has not progressed, preventing trim from happening. In this change, we correctly retrieve the actual snapshot as the checkpoint token and set the start log address as the stream's latest observed tail. 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -314,23 +314,19 @@ public class SequencerServer extends AbstractServer {
                                   ChannelHandlerContext ctx, IServerRouter r) {
         TokenRequest req = msg.getPayload();
         List<UUID> streams = req.getStreams();
-        List<Long> streamTails;
+        Map<UUID, Long> streamTails;
         Token token;
         if (req.getStreams().isEmpty()) {
             // Global tail query
             token = new Token(sequencerEpoch, globalLogTail - 1);
-            streamTails = Collections.emptyList();
-        } else if (req.getStreams().size() == 1) {
-            // single stream query
-            token = new Token(sequencerEpoch, streamTailToGlobalTailMap.getOrDefault(streams.get(0), Address.NON_EXIST));
-            streamTails = Collections.emptyList();
+            streamTails = Collections.emptyMap();
         } else {
-            // multiple stream query, the token is populated with the global tail and the tail queries are stored in
-            // streamTails
+            // multiple or single stream query, the token is populated with the global tail
+            // and the tail queries are stored in streamTails
             token = new Token(sequencerEpoch, globalLogTail - 1);
-            streamTails = new ArrayList<>(streams.size());
+            streamTails = new HashMap<>(streams.size());
             for (UUID stream : streams) {
-                streamTails.add(streamTailToGlobalTailMap.getOrDefault(stream, Address.NON_EXIST));
+                streamTails.put(stream, streamTailToGlobalTailMap.getOrDefault(stream, Address.NON_EXIST));
             }
         }
 
@@ -559,7 +555,7 @@ public class SequencerServer extends AbstractServer {
                     txResolutionResponse.getTokenType(),
                     txResolutionResponse.getConflictingKey(),
                     txResolutionResponse.getConflictingStream(),
-                    newToken, Collections.emptyMap(), Collections.emptyList())));
+                    newToken, Collections.emptyMap(), Collections.emptyMap())));
             return;
         }
 
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -14,7 +14,9 @@ import org.corfudb.protocols.logprotocol.SMREntry;
 import org.corfudb.protocols.wireprotocol.DataType;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.runtime.exceptions.TrimmedException;
+import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.object.ICorfuSMR;
 import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.runtime.view.Address;
@@ -26,6 +28,8 @@ import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
 
 import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -132,11 +136,17 @@ public class CheckpointWriter<T extends Map> {
      * @return Token at which the snapshot for this checkpoint was taken.
      */
     public Token appendCheckpoint() {
-        // Queries the sequencer for the stream tail. We then read the stream tail to
-        // persist the entry on the log unit in turn preventing a new sequencer from
-        // regressing tokens.
-        Token snapshot = rt.getSequencerView().query(streamId).getToken();
-        return appendCheckpoint(snapshot);
+        // Queries the sequencer for the stream tail and global log tail.
+        // The global log tail will provide the snapshot for this checkpoint,
+        // while the stream tail will provide the last observed update to this stream at
+        // the given snapshot.
+        TokenResponse response = rt.getSequencerView().query(new UUID[]{streamId});
+        // The checkpoint snapshot is given by the Global Tail
+        Token snapshot = response.getToken();
+        // The checkpoint start log address is given by the stream's tail at snapshot time
+        Long streamTail = response.getStreamTail(streamId);
+
+        return appendCheckpoint(snapshot, streamTail);
     }
 
     /**
@@ -145,26 +155,31 @@ public class CheckpointWriter<T extends Map> {
      * This API should not be directly invoked.
      *
      *  @param snapshotTimestamp snapshot at which the checkpoint is taken.
+     *  @param streamTail tail of the stream to checkpoint at snapshot time.
      */
     @VisibleForTesting
-    public Token appendCheckpoint(Token snapshotTimestamp) {
+    public Token appendCheckpoint(Token snapshotTimestamp, Long streamTail) {
         long start = System.currentTimeMillis();
 
-        // If snapshotTimestamp (stream tail) is a negative value
-        // we need to enforce a hole to ensure the start log address
+        // If stream tail is a negative value, we need to enforce a hole to ensure the start log address
         // is not negative (as MultiCheckpointWriter will select the min of all checkpoints)
         // Having a negative value will prevent trimming from happening.
-        if (!Address.isAddress(snapshotTimestamp.getSequence())) {
+        if (!Address.isAddress(streamTail)) {
             snapshotTimestamp = forceHole();
+            streamTail = snapshotTimestamp.getSequence();
         }
 
-        try {
-            rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
-        } catch (TrimmedException te) {
-            // Stream Tail was trimmed, we can just ignore this trim. This means
-            // no other update has been done to this stream since trim.
-            log.info("appendCheckpointOnSnapshot: stream tail at {} has been trimmed. Checkpoint on same state" +
-                    "as previous checkpoint.", snapshotTimestamp);
+        // This read attempts to ensure no future sequencer regression by materializing the global tail, i.e.,
+        // preventing reassignment of tokens on sequencer failover (for the non-materialized space).
+        rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
+
+        // However, we still need to verify there has not been a sequencer regression
+        // from the time the tuple (snapshotTimestamp, streamTail) was acquired and the actual read was performed.
+        // To this end, we query the sequencer for the last observed epoch. This validation
+        // is pessimistic as we might observe the change after the read, but it is safe to retry to avoid data loss.
+        long currentEpoch = rt.getSequencerView().query().getEpoch();
+        if (currentEpoch != snapshotTimestamp.getEpoch()) {
+            throw new WrongEpochException(currentEpoch);
         }
 
         rt.getObjectsView().TXBuild()
@@ -176,14 +191,12 @@ public class CheckpointWriter<T extends Map> {
             // A checkpoint writer will do two accesses one to obtain the object
             // vlo version and to get a shallow copy of the entry set
             log.info("appendCheckpoint: Started checkpoint for {} at snapshot {}", streamId, snapshotTimestamp);
-            ICorfuSMR<T> corfuObject = (ICorfuSMR<T>) this.map;
             Set<Map.Entry> entries = this.map.entrySet();
             // The vloVersion which will determine the checkpoint START_LOG_ADDRESS (last observed update for this
             // stream by the time of checkpointing) is defined by the stream's tail instead of the stream's version,
             // as the latter discards holes for resolution, hence if last address is a hole it would diverge
             // from the stream address space maintained by the sequencer.
-            long vloVersion = snapshotTimestamp.getSequence();
-            startCheckpoint(snapshotTimestamp, vloVersion);
+            startCheckpoint(snapshotTimestamp, streamTail);
             appendObjectState(entries);
             finishCheckpoint();
             long cpDuration = System.currentTimeMillis() - start;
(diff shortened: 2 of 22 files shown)


runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -135,8 +136,13 @@ public Token appendCheckpoint() {
         // Queries the sequencer for the stream tail. We then read the stream tail to
```
WenbinZhu: Can you change the comment accordingly? Because now we also need global tail. And also it would be better to explain this query api guarantees no updates between stream tail and global tail.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -147,15 +154,16 @@ public Token appendCheckpoint() {
...
         // we need to enforce a hole to ensure the start log address
         // is not negative (as MultiCheckpointWriter will select the min of all checkpoints)
         // Having a negative value will prevent trimming from happening.
-        if (!Address.isAddress(snapshotTimestamp.getSequence())) {
+        if (!Address.isAddress(streamTail)) {
             snapshotTimestamp = forceHole();
+            streamTail = snapshotTimestamp.getSequence();
         }
 
         try {
```
WenbinZhu: (Seems I cannot select the next line for comment, so just using this line.)
```java
rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
```
Is this correct? snapshotTimestamp.getSequence() is the global tail, I think you need the stream tail in order to prevent the issue we used to address in #2072 right?

WenbinZhu: My bad, this seems not related with issue #2072. And actually we have to read global tail to prevent sequencer regress.

annym: Exactly, because snapshotTimestamp (global tail) will always be greater or equal to the stream tail, we only need to read the global tail to prevent sequencer regression, i.e., reassignment of sequence number in a different epoch.

WenbinZhu: @annym There are other issues with sequencer regress in current approach and I just talked with @Maithem . Let’s have a discussions tomorrow.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -147,15 +154,16 @@ public Token appendCheckpoint() {
...
+    public Token appendCheckpoint(Token snapshotTimestamp, Long streamTail) {
         long start = System.currentTimeMillis();
 
-        // If snapshotTimestamp (stream tail) is a negative value
+        // If stream tail is a negative value
         // we need to enforce a hole to ensure the start log address
         // is not negative (as MultiCheckpointWriter will select the min of all checkpoints)
         // Having a negative value will prevent trimming from happening.
-        if (!Address.isAddress(snapshotTimestamp.getSequence())) {
+        if (!Address.isAddress(streamTail)) {
```
Maithem: The stream tail query and the read have to happen on the same epoch, otherwise the sequencer can regress in between and the purpose of the read is defeated. Imagine this scenario:
1. Tail query
2. read token from step 1
3. write checkpoint

The point of step #2 is to prevent the sequencer from regressing beyond the value returned in step 1, but if the sequencer actually regresses between step 1 and step 2, then step 2 becoems a point-read and can result writing a checkpoint before that address. As a consequence, data loss can happen.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -147,19 +155,29 @@ public Token appendCheckpoint() {
...
             snapshotTimestamp = forceHole();
+            streamTail = snapshotTimestamp.getSequence();
         }
 
         try {
+            // This read attempts to ensure no future sequencer regression by materializing the log until this boundary.
```
WenbinZhu: by materializing the **global tail**

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -147,19 +155,29 @@ public Token appendCheckpoint() {
...
+            // However, we still need to verify there has not been a sequencer regression
+            // from the time the tuple (snapshotTimestamp, streamTail) was acquired and the actual read was performed.
+            // To this end, we query the sequencer for the last observed epoch. This validation
+            // is pesimistic as we might observe the change after the read, but it is safe to retry.
+            long currentEpoch = rt.getSequencerView().query().getEpoch();
+            if (currentEpoch != snapshotTimestamp.getEpoch()) {
+                throw new WrongEpochException(currentEpoch);
+            }
         } catch (TrimmedException te) {
             // Stream Tail was trimmed, we can just ignore this trim. This means
```
WenbinZhu: This is not a stream tail any more, it's a global tail. Can we change the comment and the log message after this?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -205,6 +205,13 @@
         @Default
         int trimRetry = 2;
 
+        /**
+         * The total number of retries the checkpointer will attempt on sequencer regressions (epoch change). This
```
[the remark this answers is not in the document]
annym: Done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: You know what, actually we no longer need this. This was in place because on previous design the snapshot was given by the stream's tail and this. could be trimmed, but now the snapshot is given by the global tail, for which a trimmed exception is not admitted. Removing this.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -320,13 +320,9 @@ private void handleTokenQuery(CorfuPayloadMsg<TokenRequest> msg,
...
-        } else if (req.getStreams().size() == 1) {
-            // single stream query
-            token = new Token(sequencerEpoch, streamTailToGlobalTailMap.getOrDefault(streams.get(0), Address.NON_EXIST));
-            streamTails = Collections.emptyList();
         } else {
-            // multiple stream query, the token is populated with the global tail and the tail queries are stored in
-            // streamTails
+            // multiple or single stream query, the token is populated with the global tail
+            // and the tail queries are stored in streamTails
             token = new Token(sequencerEpoch, globalLogTail - 1);
```
Maithem: I think a map is a more appropriate data structure. That way the consumer doesn't have to maintain a uuid -> index

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -145,26 +153,32 @@ public Token appendCheckpoint() {
...
 
-        // If snapshotTimestamp (stream tail) is a negative value
+        // If stream tail is a negative value
         // we need to enforce a hole to ensure the start log address
         // is not negative (as MultiCheckpointWriter will select the min of all checkpoints)
         // Having a negative value will prevent trimming from happening.
-        if (!Address.isAddress(snapshotTimestamp.getSequence())) {
+        if (!Address.isAddress(streamTail)) {
             snapshotTimestamp = forceHole();
+            streamTail = snapshotTimestamp.getSequence();
```
Maithem: This is inconsistent. If you look at `CheckpointWriter::forceHole` it can return a different token than the one passed to `AddressSpaceView::write` because internally it uses `layoutHelper` which can reset the token.

Maithem: Since its a hole, why not just use `AddressSpace::read` directly and let if fill the hole?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -145,26 +153,32 @@ public Token appendCheckpoint() {
...
-        try {
-            rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
-        } catch (TrimmedException te) {
-            // Stream Tail was trimmed, we can just ignore this trim. This means
-            // no other update has been done to this stream since trim.
-            log.info("appendCheckpointOnSnapshot: stream tail at {} has been trimmed. Checkpoint on same state" +
-                    "as previous checkpoint.", snapshotTimestamp);
+        // This read attempts to ensure no future sequencer regression by materializing the global tail, i.e.,
+        // preventing reassignment of tokens on sequencer failover (for the non-materialized space).
+        rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
```
Maithem: I think it will be cleaner if the stream marking code is refactored into a seperate method. Maybe introduce a method called `markStream` which returns a "stable" token.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -145,26 +153,32 @@ public Token appendCheckpoint() {
...
         }
 
-        try {
-            rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());
-        } catch (TrimmedException te) {
-            // Stream Tail was trimmed, we can just ignore this trim. This means
-            // no other update has been done to this stream since trim.
-            log.info("appendCheckpointOnSnapshot: stream tail at {} has been trimmed. Checkpoint on same state" +
-                    "as previous checkpoint.", snapshotTimestamp);
+        // This read attempts to ensure no future sequencer regression by materializing the global tail, i.e.,
```
Maithem: Also `ICorfuSMR<T> corfuObject = (ICorfuSMR<T>) this.map;` is not being used.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
```diff
@@ -71,14 +72,27 @@ public Token appendCheckpoints(CorfuRuntime rt, String author) {
                                 .getSerializer();
                 cpw.setSerializer(serializer);
 
-                Token minCPSnapshot = cpw.appendCheckpoint();
+                Token minCPSnapshot = Token.UNINITIALIZED;
```
Maithem: Since the snapshot returned `appendCheckpoint` is a global snapshot that shouldn't regress, shouldn't we just select the first checkpoints snapshot and make sure subsequent checkpoints all return greater snapshots (instead of just finding a min of all snapshots).

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
```diff
@@ -71,14 +72,27 @@ public Token appendCheckpoints(CorfuRuntime rt, String author) {
...
+                    }
+                }
 
                 if (minSnapshot == Token.UNINITIALIZED) {
                     minSnapshot = minCPSnapshot;
-                } else if (minSnapshot.getEpoch() != minCPSnapshot.getEpoch()) {
-                    String msg = String.format("Epoch changed during GC cycle from %s to %s", minSnapshot,
-                            minCPSnapshot);
-                    throw new IllegalStateException(msg);
                 } else if (Token.min(minCPSnapshot, minSnapshot) == minCPSnapshot) {
```
Maithem: This min tracking logic can be cleaned up/removed if you decide to do the suggestion above.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -193,7 +193,7 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
         for (int x = 0; x < rt.getParameters().getTrimRetry(); x++) {
             // Linearize this read against a timestamp
             final long timestamp = rt.getSequencerView()
-                            .query(getStreamID()).getToken().getSequence();
+                            .query(getStreamID()).getStreamTails().get(0);
```
Maithem: Maybe hide this implementation in the SequencerView -- leaking the list "de-referencing  logic" doesn't seem clean.

Maithem: Maybe introduce `SequencerView::query(UUID streamId)`, then the external consumers don't have to access the list directly.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem: I think this is still possible. From the time forceHole is called to the time `rt.getAddressSpaceView().read(snapshotTimestamp.getSequence());` is called, a trim can happened. Its just an unhanded exception not sure if its a serious issue.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
WenbinZhu: +1

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
WenbinZhu: +1

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem: I think my first comment is incorrect. The addressspace write doesn't change the token, so `forceHole` should only fail with a OverwriteException/Trimmed Exception

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: But if a trimmed is called in between we should actually fail right? or you suggest to include this case as part of our retry logic?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
annym: Yeah, this is old code. I just took what was there but this is more convenient. Will do.

annym: Done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: Good point! Done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: But what if there is a sequencer failover (between acquiring the token for the hole and writing)... and this same position ends up being a hole but for a different stream, then the validateStateOfWrittenEntry succeeds cause it's a hole, that would be an issue, right? At the end we always have a problem because token request and write/read for hole filling is not performed atomically. We can always have regressions that lead to data loss.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
annym: Sure, then I'll throw an exception if by chance a snapshot is lower cause that would mean there was some kind of regression.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
annym: done.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
annym: You mean for the return type to be a long? ... for cases like the checkpointer we'd still need to go through the other API cause we want the global tail as well.

annym: Done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: Never mind, the epoch validation that we do afterwards ensures this won't happen.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
annym: Done.

annym:
@Maithem @WenbinZhu I've addressed all the comments, please do a final review.

PR #2102, opened 2019-08-07
GC: Unify LogEntry types: SMREntry, MultiSMREntry, MultiObjectSMREntry.

WenbinZhu:
## Overview

We have three SMR LogEntry types, namely SMREntry for a non-transactional
update, MultiObjectSMREntry for transactional updates and MultiSMREntry
for checkpoint entry. This creates a lot of confusion, boilerplates and
difficulties when handling log updates. With this patch, all transactional,
non-transactional updates and checkpoint entries are composed of a single
log entry type: SMRLogEntry, which is similar to MultiObjectSMREntry. And
each SMR update is turned into SMRRecord, which is similar to SMREntry,
but is not inherited from LogEntry any more, because now the only LogEntry
type persisted on log unit server would be SMRLogEntry.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRLogEntry.java
```diff
@@ -0,0 +1,129 @@
...
+        this.entryMap = entryMap;
+    }
+
+    /**
+     * Extract a particular stream's entry from this object.
+     *
+     * @param streamID stream ID
+     * @return the SMR Record list corresponding to streamId
+     */
+    private List<SMRRecord> getStreamEntry(UUID streamID) {
```
lixinchengdu: check null

WenbinZhu: This is private API, I think it's not that necessary since the public APIs are not even doing this.

WenbinZhu:
@lixinchengdu Merged, you can start the garbage info changes.

PR #2108, opened 2019-08-09
Persistent back-end for CorfuTable

vjeko:
## Overview

Description:

By default, CorfuTable builds its view in-memory, but in some
circumstances the data set is either too large or it causes
unnecessary memory bloat. This patch addresses this issue by
introducing a feature that lets client offload the data from
heap to disk.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -1,6 +1,5 @@
 package org.corfudb.runtime.collections;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.reflect.TypeToken;
@@ -37,7 +36,6 @@ import org.corfudb.annotations.DontInstrument;
 import org.corfudb.annotations.Mutator;
 import org.corfudb.annotations.MutatorAccessor;
 import org.corfudb.annotations.TransactionalMethod;
-import org.corfudb.util.ImmuableListSetWrapper;
 
 /** The CorfuTable implements a simple key-value store.
  *
@@ -223,39 +221,58 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
     }
 
     // The "main" map which contains the primary key-value mappings.
-    private final Map<K,V> mainMap;
+    private final StreamingMap<K,V> mainMap;
     private final Set<Index<K, V, ? extends Comparable>> indexSpec = new HashSet<>();
     private final Map<String, Map<Comparable, Map<K, V>>> secondaryIndexes = new HashMap<>();
 
     /**
-     * Generate a table with a given implementation for the mainMap
+     * Generate a table with a given implementation for the {@link StreamingMap}.
      */
-    public CorfuTable(IndexRegistry<K, V> indices, Map<K, V> mapImpl) {
+    public CorfuTable(IndexRegistry<K, V> indices, StreamingMap<K, V> streamingMap) {
         indices.forEach(index -> {
             secondaryIndexes.put(index.getName().get(), new HashMap<>());
             indexSpec.add(index);
         });
         log.info("CorfuTable: creating CorfuTable with the following indexes: {}",
                 secondaryIndexes.keySet());
-        mainMap = mapImpl;
+        mainMap = streamingMap;
+    }
+
+    /**
+     * Generate a table with a given implementation for the {@link Map}.
+     */
+    public CorfuTable(IndexRegistry<K, V> indices, Map<K, V> mapImpl) {
+        this(indices, new StreamingMapDecorator<>(mapImpl));
     }
 
     /**
      * Generate a table with the given set of indexes.
      */
     public CorfuTable(IndexRegistry<K, V> indices) {
-        this(indices, new HashMap<>());
+        this(indices, new StreamingMapDecorator<>(new HashMap<>()));
+    }
+
+    /**
+     * Generates a table with the given {@link Map} implementation and
+     * without any secondary indexes.
+     */
+    public CorfuTable(Map<K, V> mapImpl) {
+        this(IndexRegistry.empty(), mapImpl);
+    }
+
+    /**
+     * Generates a table with the given {@link StreamingMap} implementation
+     * and without any secondary indexes.
+     */
+    public CorfuTable(StreamingMap<K, V> streamingMap) {
+        this(IndexRegistry.empty(), streamingMap);
     }
 
     /**
      * Default constructor. Generates a table without any secondary indexes.
      */
     public CorfuTable() {
-        this(IndexRegistry.empty(), new HashMap<>());
-    }
-
-    public CorfuTable(Map<K, V> underlying) {
-        this(IndexRegistry.empty(), underlying);
+        this(IndexRegistry.empty(), new StreamingMapDecorator<>(new HashMap<>()));
     }
 
     /** {@inheritDoc} */
@@ -457,24 +474,24 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
      * This method has a memory/CPU advantage over the map iterators as no deep copy
      * is actually performed.
      *
-     * @param p java predicate (function to evaluate)
+     * @param valuePredicate java predicate (function to evaluate)
      * @return a view of the values contained in this map meeting the predicate condition.
      */
     @Accessor
-    public List<V> scanAndFilter(Predicate<? super V> p) {
-        return mainMap.values().parallelStream()
-                                    .filter(p)
-                                    .collect(Collectors.toCollection(ArrayList::new));
+    public List<V> scanAndFilter(Predicate<? super V> valuePredicate) {
+        return mainMap.entryStream()
+                .map(Entry::getValue).filter(valuePredicate)
+                .collect(Collectors.toCollection(ArrayList::new));
     }
 
     /** {@inheritDoc} */
     @Override
     @Accessor
-    public Collection<Map.Entry<K, V>> scanAndFilterByEntry(Predicate<? super Map.Entry<K, V>>
-                                                                    entryPredicate) {
-        return mainMap.entrySet().parallelStream()
-                                    .filter(entryPredicate)
-                                    .collect(Collectors.toCollection(ArrayList::new));
+    public Collection<Map.Entry<K, V>> scanAndFilterByEntry(
+            Predicate<? super Map.Entry<K, V>> entryPredicate) {
+        return mainMap.entryStream()
+                .filter(entryPredicate)
+                .collect(Collectors.toCollection(ArrayList::new));
     }
 
     /** {@inheritDoc} */
@@ -549,14 +566,16 @@ public class CorfuTable<K ,V> implements ICorfuMap<K, V> {
     @Override
     @Accessor
     public @Nonnull Set<K> keySet() {
-        return ImmutableSet.copyOf(mainMap.keySet());
+        return mainMap.entryStream().map(Entry::getKey)
+                .collect(ImmutableSet.toImmutableSet());
     }
 
     /** {@inheritDoc} */
     @Override
     @Accessor
     public @Nonnull Collection<V> values() {
-        return ImmutableList.copyOf(mainMap.values());
+        return mainMap.entryStream().map(Entry::getValue)
+                .collect(ImmutableSet.toImmutableSet());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


vjeko:
@hisundar as we discussed in the meeting, multiple CorfuRuntimes consuming the same table on the same physical node is not not a use-case that is needed by our clients. As such, there is no point of tackling this use case at the moment, since it would be adding significant complexity in form of IPC while giving us very little benefit. Furthermore, it is unclear if that approach would even work or if it is a good approach.

At this point we need to fix all issues surrounding this approach, which are related other components in CorfuRuntime (such as VLO).

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbStreamingMap.java
```diff
@@ -0,0 +1,273 @@
...
+                next = null;
+                return current;
+            } else {
+                throw new NoSuchElementException();
+            }
+
+        }
+    }
+
+    public static byte[] byteArrayFromBuf(final ByteBuf buf) {
```
medhavidhawan: comments

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbStreamingMap.java
```diff
@@ -0,0 +1,273 @@
...
+    public boolean isEmpty() {
+        return dataSetSize.get() == 0;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean containsKey(@NonNull Object key) {
+        ByteBuf payload = Unpooled.buffer();
```
medhavidhawan: Do we need to clean up this buffer ?

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.LongStream;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class DiskBackedCorfuClientTest extends AbstractViewTest {
```
medhavidhawan: comments

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
```diff
@@ -0,0 +1,171 @@
...
+        final Options options =
+                new Options().setCreateIfMissing(true)
+                        .setSstFileManager(sstFileManager)
+                        // The size is checked either during flush or compaction.
+                        .setWriteBufferSize(FileUtils.ONE_KB);
+
+        CorfuTable<String, String>
+                table = getDefaultRuntime().getObjectsView().build()
+                .setTypeToken(new TypeToken<CorfuTable<String, String>>() {})
+                .setArguments(new RocksDbStreamingMap<String, String>(
```
medhavidhawan: I think we should move the naming away from RocksDB. If we change the technology or have a different implementation in the future, all clients will have to change this. Maybe PersistedStreamingMap or something like that.

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
```diff
@@ -0,0 +1,171 @@
...
+                    String value = RandomStringUtils.random(ENTITY_CHAR_SIZE, true, true);
+                    table.put(key, value);
+                    String persistedValue = table.get(key);
+                    Assertions.assertEquals(value, persistedValue);
+                })).isInstanceOf(UnrecoverableCorfuError.class)
+                .hasCauseInstanceOf(RocksDBException.class);
+    }
+
+    @Test
+    public void scanAndFilter() throws RocksDBException {
```
medhavidhawan: please have high level comments with expectations from test.

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
medhavidhawan: 1. Do we need tests around any kind of failure conditions ?
2. Should we think of having some sort of disk size limitation ?
3. Can we add tests where transactions involve diskbased and memorybased tables ?

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbStreamingMap.java
```diff
@@ -0,0 +1,273 @@
...
+         */
+        @Override
+        public boolean hasNext() {
+            // If we do not have the next element pipelined, go ahead and fetch it.
+            if (next == null) {
+                // If the iterator is valid, this means that the next entry exists.
+                if (iterator.isValid()) {
+                    // Go ahead and cache that entry.
+                    next = new AbstractMap.SimpleEntry(
+                            serializer.deserialize(Unpooled.wrappedBuffer(iterator.key()), corfuRuntime),
```
medhavidhawan: Can the rocksdb iterator fail or throw runtime exceptions ? If so what should be corfu behavior in that case ?

xnull:
@xiaoqin2012 
>If the database is small that can be easily fit into memory, using rocksdb is kind of overdo.
I believe it's a false statement. But really doesn't matter right now

>It will be nice to have two options for corfutable: memory based and disk based.
yes, we have that

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMap.java
```diff
@@ -0,0 +1,18 @@
+package org.corfudb.runtime.collections;
+
+import java.util.Map;
+import java.util.stream.Stream;
+
+/**
+ *
+ * This interface provides additional functionality not provides by the standard {@link Map}.
```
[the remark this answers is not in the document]
vjeko: Fixed.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMap.java
```diff
@@ -0,0 +1,18 @@
...
+ * In cases when the actual data is no being backed by the heap, {@link Map#values()},
```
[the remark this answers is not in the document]
vjeko: Fixed

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -453,24 +474,24 @@ public void insert(@ConflictParameter K key, V value) {
...
     @Override
     @Accessor
-    public Collection<Map.Entry<K, V>> scanAndFilterByEntry(Predicate<? super Map.Entry<K, V>>
-                                                                    entryPredicate) {
-        return mainMap.entrySet().parallelStream()
-                                    .filter(entryPredicate)
-                                    .collect(Collectors.toCollection(ArrayList::new));
+    public Collection<Map.Entry<K, V>> scanAndFilterByEntry(
+            Predicate<? super Map.Entry<K, V>> entryPredicate) {
+        return mainMap.entryStream().parallel()
```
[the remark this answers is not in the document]
vjeko: That parallel stream had to be moved to the underlying StreamingMap implementation, since most embedded databases do not support multi-threaded access to the same iterator.

But to answer your question -- no there is no difference:
```java
    default Stream<E> stream() {
        return StreamSupport.stream(spliterator(), false);
    }

    default Stream<E> parallelStream() {
        return StreamSupport.stream(spliterator(), true);
    }

```
Calling parallel() will set the flag from `false` to `true`:

```java
    @Override
    public final S parallel() {
        sourceStage.parallel = true;
        return (S) this;
    }
```

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -545,14 +566,16 @@ public void clear() {
     @Override
     @Accessor
     public @Nonnull Set<K> keySet() {
-        return ImmutableSet.copyOf(mainMap.keySet());
+        return mainMap.entryStream().map(Entry::getKey)
```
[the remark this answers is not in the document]
vjeko: Mostly idiomatic. All bulk-level APIs have to consume the data via entryStream(), since it is not guaranteed that keySet() valueSet() and entrySet() are actually implemented.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbStreamingMap.java
```diff
@@ -0,0 +1,273 @@
...
+    public void putAll(@NonNull Map<? extends K, ? extends V> map) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void clear() {
+        dataSetSize.set(0);
```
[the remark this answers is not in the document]
vjeko: Keeping track of the size has to be done independently via AtomicInteger dataSetSize. So this is effectively resting the counter to zero.

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
vjeko: Done.

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
vjeko: Done.

Remember, Corfu clients will not be working with the `CorfuTable` directly, but rather via the `CorfuStore`, so this implementation detail will not leak outside of Corfu.

I thought about abstracting `RocksDbStreamingMap` away, but the problem is that each of the options is very specific to the embedded database we end up using (take a look at the `org.rocksdb.Options` parameter).

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
vjeko: Done.

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
vjeko: There is a test for (2). I will address (1) and (3).

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbStreamingMap.java
vjeko: Yes, it can fail, and in all of those cases there is no way to gracefully recover. I added a method `checkInvariants()` that deals with this aspect. That being said, if there is an issue with the iterator, then we have a bug in our code.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,294 @@
...
+ */
+@Slf4j
+public class PersistedStreamingMap<K, V> implements StreamingMap<K, V> {
+
+    private final AtomicInteger dataSetSize = new AtomicInteger();
+    private final CorfuRuntime corfuRuntime;
+    private final ISerializer serializer;
+    private final RocksDB rocksDb;
+
+    public PersistedStreamingMap(@NonNull File dataPath,
```
xnull: You can do this:
```
@Builder
public class PersistedStreamingMap ... {
    private final AtomicInteger dataSetSize = new AtomicInteger();
    
    @NonNull
    private final CorfuRuntime corfuRuntime;
    @NonNull
    private final ISerializer serializer;
    @NonNull
    private final RocksDB rocksDb;

    //public PersistedStreamingMap don't need it anymore
}
```

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,294 @@
...
+            }
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public boolean hasNext() {
+            // If we do not have the next element pipelined, go ahead and fetch it.
+            if (next == null) {
```
xnull: please invert `if` condition

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,294 @@
...
+        }
+        this.serializer = serializer;
+        this.corfuRuntime = corfuRuntime;
+    }
+
+    /**
+     * A Java compatible {@link RocksIterator} implementation
+     */
+    public class RocksDbIterator implements Iterator<Entry<K, V>> {
+        private RocksIterator iterator;
```
xnull: should these fileds be atomic or volatile?

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: It is not a simple data-class, and we need the initialization logic. We could move it to init(), but then we run into risk of forgetting to call it.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: `RocksIterator` does not support multi-threaded access.

xnull: why not? if rocks provides that, why not to support that

vjeko: `RocksIterator` is RocksDB implementaion. RocksDB does not support it. Not us.

vjeko:
@xiaoqin2012 Data that is stored in the `mainMap` is ephemeral. It is rebuilt each time `CorfuTable` is opened.

vjeko:
Superseded by: https://github.com/CorfuDB/CorfuDB/pull/2198

PR #2135, opened 2019-08-29
Streams Tails Rebuilt from Log Unit Fix

annym:
## Overview

Description:

- Fix checkpointer 'hole' writes to carry backpointerMap info so address maps are generated.
- Build address maps for checkpointed streams based on reader holes, to support sequencer reconfiguration
from log units (not miss tail info).

Why should this be merged: fixes an issue with sequencer reconfiguration which can lead to false trimmedExceptions preventing the system to recover.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -136,15 +130,14 @@ public CheckpointWriter(CorfuRuntime rt, UUID streamId, String author, T map) {
...
      */
     public Token appendCheckpoint() {
-        // Queries the sequencer for the stream tail and global log tail.
-        // The global log tail will provide the snapshot for this checkpoint,
-        // while the stream tail will provide the last observed update to this stream at
-        // the given snapshot.
-        TokenResponse response = rt.getSequencerView().query(new UUID[]{streamId});
-        // The checkpoint snapshot is given by the Global Tail
-        Token snapshot = response.getToken();
+        // We enforce a NO_OP entry for every checkpoint, i.e., a hole with backpointer map info,
```
WenbinZhu: Can we use LogData.EMPTY instead of a hole because I think a hole by definition should not have back pointers.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -87,7 +88,7 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
                     // Validate that the data entry belongs to this stream, otherwise, skip.
                     // This verification protects from sequencer regression (tokens assigned in an older epoch
                     // that were never written to, and reassigned on a newer epoch)
-                    if (ld.containsStream(this.id)) {
+                    if (ld.containsStream(this.id) && ld.getType() == DataType.DATA) {
```
Maithem: Should you just left the stream adapter to filter this out ?

test/src/test/java/org/corfudb/recovery/Helpers.java
```diff
@@ -72,7 +72,11 @@ static void assertThatMapIsBuilt(CorfuRuntime rt1, CorfuRuntime rt2, String stre
         // Assert that UnderlyingObjects are at the same version
         // If they are at the same version, a sync on the object will
         // be a no op for the new runtime.
-        assertThat(vo1Prime.getVersionUnsafe()).isEqualTo(vo1.getVersionUnsafe());
+
+        // This utility is used by many checkpoint tests and because new checkpoint
+        // inserts a NO_OP entry, VLO version will differ from original (before checkpoint) to
```
Maithem: uncomment test?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem: That might fail the replication layer code, because it checks for empty and tries to fill those slots.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -212,11 +212,10 @@ public Token appendCheckpoint(Token snapshotTimestamp, Long streamTail) {
     }
 
     private Token forceHole() {
-        Token writeToken = rt.getSequencerView().next(streamId).getToken();
+        TokenResponse writeToken = rt.getSequencerView().next(streamId);
         LogData logData = new LogData(DataType.HOLE);
-        logData.useToken(writeToken);
         rt.getAddressSpaceView().write(writeToken, logData, CacheOption.WRITE_AROUND);
-        return writeToken;
+        return writeToken.getToken();
```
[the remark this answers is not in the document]
Maithem: The write will set the token for the logData

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Actually, I came to this because a unit test was failing. If I'm not wrong it was a raw access to the stream and it just got stuck infinitely if we do not filter on the data type at this layer.

test/src/test/java/org/corfudb/recovery/Helpers.java
annym: There were too many. I'll later write another utility to separate the ones that can use this comparison from the ones that can't.

PR #2157, opened 2019-09-12
Adding Protobufs for CorfuStore [PART 1]

zalokhan:
## Overview

Description:
Introducing the relevant protobufs required to initialize the CorfuStore.

Why should this be merged: 
Introducing CorfuStore.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/proto/corfu_options.proto
+++ b/runtime/proto/corfu_options.proto
@@ -1,0 +1,20 @@
+syntax = "proto3";
+
+package org.corfudb.runtime;
+option java_package = "org.corfudb.runtime";
+
+import "google/protobuf/descriptor.proto";
+
+// Option tags to be used by the CorfuStore consumers to tag special fields to be detected by Corfu.
+message SchemaOptions {
+    // Secondary keys to be indexed.
+    bool secondary_key = 1;
+    // Version number in metadata field.
+    bool version = 2;
+}
+
+// Field options to be extended in the user's protobuf fields.
+extend google.protobuf.FieldOptions {
+    // 1036 is in the extendable range in the descriptor.proto.
+    SchemaOptions schema = 1036;
+}
--- a/runtime/proto/corfu_store_metadata.proto
+++ b/runtime/proto/corfu_store_metadata.proto
@@ -1,0 +1,40 @@
+syntax = "proto3";
+
+package org.corfudb.runtime;
+option java_package = "org.corfudb.runtime";
+
+import "google/protobuf/descriptor.proto";
+import "google/protobuf/any.proto";
+
+// Logical Timestamp to perform queries and transactions at a particular snapshot.
+message Timestamp {
+    int64 epoch = 1;
+    int64 sequence = 2;
+}
+
+// Table name.
+message TableName {
+    string namespace = 1;
+    string table_name = 2;
+}
+
+// Descriptor to provide self describing schemas for the records.
+message TableDescriptors {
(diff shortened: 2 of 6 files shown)


runtime/proto/corfu_options.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto2";
```
pankti-m: why proto2 and not 3?

test/src/test/resources/proto/test_schema.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto2";
+
```
xiaoqin2012: Same here, why someplace use proto3 some place uses proto2?

zalokhan:
@pankti-m @xiaoqin2012 Addressed your comments.
Changed all Corfu options and metadata to use proto3.
The test protobuf classes will continue to use proto2 to test the backward compatibility.

runtime/proto/corfu_options.proto
```diff
@@ -0,0 +1,18 @@
...
+// Option tags to be used by the CorfuStore consumers to tag special fields to be detected by Corfu.
+message SchemaOptions {
+    // Secondary keys to be indexed.
+    bool secondary_key = 1;
+    // Version number in metadata field.
+    bool version = 2;
+}
+
+extend google.protobuf.FieldOptions {
+    SchemaOptions schema = 1036;
```
lixinchengdu: Can someone explain to me what the magic number 1036 means here?

test/src/test/resources/proto/test_schema.proto
```diff
@@ -0,0 +1,24 @@
...
+package org.corfudb.test;
+option java_package = "org.corfudb.test";
+
+import "corfu_options.proto";
+
+message Uuid {
+    optional uint64 msb = 1;
```
lixinchengdu: Can I know why the both msb and lsb are optional? Does it means that if one field is missing, the value of that filed is 0?

runtime/proto/corfu_options.proto
hisundar: It is google protobuf's FieldOptions field number that does not collide with an existing number.

test/src/test/resources/proto/test_schema.proto
hisundar: "optional" is the default specifier in proto3 
In this case it should really be "required" but this test is highlighting the possibility of defining protobufs in this manner.

runtime/proto/corfu_options.proto
medhavidhawan: please add a comment

runtime/proto/corfu_store_metadata.proto
```diff
@@ -0,0 +1,38 @@
...
+message TableDescriptors {
+    google.protobuf.DescriptorProto keyDescriptor = 1;
+    google.protobuf.DescriptorProto valueDescriptor = 2;
+}
+
+// Metadata.
+message TableMetadata {
+    TableName table_name = 1;
+    bool disk_based = 2;
+    bool cache = 3;
```
medhavidhawan: please add comment for this

runtime/proto/corfu_options.proto
zalokhan: Done.

runtime/proto/corfu_store_metadata.proto
zalokhan: Done.

PR #2158, opened 2019-09-13
Clean up Trimmedexceptions

lixinchengdu:
## Overview

This PR clean up trimmed exception handlers at the stream layer. New GC mechanism does not throw trimmed exceptions due to prefixTrim.


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -736,40 +722,27 @@ public void invalidateClientCache() {
...
         }
 
-        if (!trimmedAddresses.isEmpty()) {
-            if (!options.isIgnoreTrim()) {
-                throw new TrimmedException(trimmedAddresses);
-            }
-
-            log.warn("read: ignoring trimmed addresses {}", trimmedAddresses);
-        }
+        result.keySet().removeAll(compactedAddressses);
```
annym: should we add a trace log to print the compacted addresses?

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
```diff
@@ -292,9 +291,7 @@ private ReadResponse getAdoptedValueWithHighestRankIfPresent(
         public int compare(ReadResponse o1, ReadResponse o2) {
             LogData ld1 = o1.getAddresses().get(logPosition);
             LogData ld2 = o2.getAddresses().get(logPosition);
-            if(ld1.isTrimmed() || ld2.isTrimmed()) {
-                throw new TrimmedException();
-            }
+
```
annym: I'm not very sure of the logic here but based on what we removed, shouldn't we check ifs either log data has been compacted? or will the rank information still be available even if it was compacted?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -711,20 +703,14 @@ public void invalidateClientCache() {
...
-
+    private Map<Long, ILogData> checkForCompactedAddresses(Map<Long, ILogData> result, ReadOptions options) {
+        Collection<Long> compactedAddressses = new ArrayList<>();
         for (Map.Entry<Long, ILogData> entry : result.entrySet()) {
-            // Add trimmed addresses to list
-            if (!checkLogData(entry.getKey(), entry.getValue(), false)) {
-                trimmedAddresses.add(entry.getKey());
+            // Add compacted addresses to list
+            if (!checkLogData(entry.getKey(), entry.getValue())) {
+                compactedAddressses.add(entry.getKey());
```
annym: Once we detect compacted addresses, shouldn't we remove them from the resolvedQueue? that way we clean up (do the gc functionality that we used to do on these queues before)

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -772,26 +560,13 @@ public synchronized ILogData previous() {
...
-            // Note: this is a checkpoint, we do not need to verify it is before the trim mark, it actually should be
-            // cause this is the last address of the trimmed range.
-            context.setGlobalPointer(context.checkpoint.startAddress);
-            return null;
-        }
-
-        throw new IllegalStateException("The stream pointer seems to be corrupted!");
+        // The stream hasn't been checkpointed and we need to
+        // move the stream pointer to an address before the first
+        // entry
```
annym: we should change the comment as it does not correspond to the new changes.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -53,16 +52,8 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
...
-                ld = read(thisRead);
-            } catch (TrimmedException te) {
-                if (!getReadOptions().isIgnoreTrim()) {
-                    throw te;
-                }
 
-                readNext = true;
-                continue;
-            }
+            ld = read(thisRead);
```
annym: shouldn't you check if the read entry is compacted? in case it is compacted set readNext to true and go to next log entry, as current one is not valid.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -109,18 +99,10 @@ protected boolean discoverAddressSpace(final UUID streamId,
...
-                    log.warn("followBackpointers: Ignoring trimmed exception for address[{}]," +
-                            " stream[{}]", currentAddress, id);
-                    return !queue.isEmpty();
-                } else {
-                    throw e;
-                }
-            }
+
+            log.trace("followBackpointers: readAddress[{}]", currentAddress);
+            d = read(currentAddress, readStartTime);
```
annym: I'm not sure if pure back pointer implementation is compatible with sparse trim, is it? cause a compacted address loses backpointer information, am I right? or it is always kept around?

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -76,46 +69,28 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
...
-                    }
-                } else {
+            // Because the discovery mechanism implemented by this class
+            // does not require to read log entries in advance (only requests
+            // the stream's full address map, without reading the actual data), entries can be read in
+            // batches whenever we have a cache miss. This allows next reads
+            // to be serviced immediately, rather than reading one entry at a time.
+            ld = read(currentRead, queue);
+
+            if (queue == getCurrentContext().readQueue && ld != null) {
```
annym: we no longer need the first condition, right? cause we only have the readQueue... the readCpQueue was removed.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -76,46 +69,28 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
...
+                // Validate that the data entry belongs to this stream, otherwise, skip.
+                // This verification protects from sequencer regression (tokens assigned in an older epoch
+                // that were never written to, and reassigned on a newer epoch)
+                if (ld.containsStream(this.id) && ld.getType() == DataType.DATA) {
+                    addToResolvedQueue(getCurrentContext(), currentRead, ld);
                     readNext = false;
+                } else {
+                    log.trace("getNextEntry[{}]: the data for address {} does not belong to this stream. Skip.",
+                            this, currentRead);
+                    // Invalid entry (does not belong to this stream). Read next.
```
annym: We should add (does not belong to this stream or is compacted). Read next.

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamView.java
```diff
@@ -37,12 +37,6 @@
      */
     void reset();
 
-    /**
-     * Garbage collect all the trimmed entries on this stream
```
annym: I'm still not sure I understand how we GC all the compacted addresses of this stream over time, once this is part of the old state of a stream. Can you point me out where we take care of this?

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -531,7 +528,7 @@ private void applyForEachAddress(BiConsumer<Long, ILogData> logDataProcessor) {
 
                 invokeNecromancer(range, logDataProcessor);
 
-            } catch (TrimmedException ex) {
+            } catch (Exception ex) {
```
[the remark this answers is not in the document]
lixinchengdu: I don't remember why I change TrimmedException to Exception. I revert it back.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -307,38 +306,30 @@ public void sync() {
...
+                // The version is already ahead, but we don't have the result.
+                // The only way to get the correct result
+                // of the upcall would be to rollback. For now, we throw an exception
+                // since this is generally not expected. --- and probably a bug if it happens.
+                throw new RuntimeException("Attempted to get the result "
+                        + "of an upcall@" + timestamp + " but we are @"
+                        + underlyingObject.getVersionUnsafe()
+                        + " and we don't have a copy");
+            });
+        } catch (TrimmedException ex) {
```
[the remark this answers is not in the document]
lixinchengdu: I added some comment here.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -181,23 +181,6 @@ public VersionLockedObject(Supplier<T> newObjectFn,
...
-    /**
-     * Run gc on this object. Since the stream that backs this object is not thread-safe:
-     * synchronization between gc and external object access is needed.
-     */
-    public void gc(long trimMark) {
-        long ts = 0;
-
-        try (Timer.Context vloGcDuration = VloMetricsHelper.getVloGcContext()) {
-            ts = lock.writeLock();
-            pendingUpcalls.removeIf(e -> e < trimMark);
```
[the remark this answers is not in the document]
lixinchengdu: I added the code back to more recent patch to address this issue.

runtime/src/main/java/org/corfudb/runtime/object/ISMRStream.java
```diff
@@ -38,8 +38,6 @@
 
     void seek(long globalAddress);
```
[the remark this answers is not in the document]
lixinchengdu: That is a very good question. I went over the codebase of Corfu and found that there is only one invocation of seek and that moreover that invocation is not necessary. I think we could delete the seek API because it does not make sense in the context of in-place GC.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
lixinchengdu: The removed addresses have been tracked inside another function checkLogData.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
lixinchengdu: I also think we should eliminate backpointer address discover in the new GC branch.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
lixinchengdu: Yes. This is a issue! I plan to eliminate the backpointerStreamView in a dedicated patch.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
lixinchengdu: Yes, exactly.

runtime/src/main/java/org/corfudb/runtime/view/stream/IStreamView.java
lixinchengdu: This is taken care of in the most recent patch where addresses before compaction mark is trimmed in stream view.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
lixinchengdu: I don't understand why the compacted or irrelevant addresses into the resolved queue.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
WenbinZhu: You should also check compacted for bitmap implementation.

lixinchengdu: I've checked that read() is compatible with bitmap implementation.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -9,6 +9,7 @@
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.netty.handler.timeout.TimeoutException;
+import lombok.Data;
```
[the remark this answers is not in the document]
lixinchengdu: deleted.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
WenbinZhu: I mean check if the address is compacted for bitmap stream layer implementation.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -530,14 +522,14 @@ private ILogData cacheLoadAndGet(Cache<Long, ILogData> cache, long address, ILog
             // resulted in a cache miss and need to be fetched
             if (!addressesToFetch.isEmpty()) {
                 Map<Long, ILogData> fetchedAddresses = fetchAll(addressesToFetch, options);
-                result.putAll(checkForTrimmedAddresses(fetchedAddresses, options));
+                result.putAll(checkForCompactedAddresses(fetchedAddresses, options));
```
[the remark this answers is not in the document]
lixinchengdu: It is a good suggestion. I have changed it to the name you suggest.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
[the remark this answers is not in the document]
lixinchengdu: Initially, every logEntry read from the logUnit server is maintained in the result map, no matter whether it is compacted or not. This function iterates the `result` map and adds the compacted addresses into `compactedAddressses `. Therefore, the last set removal is necessary.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -711,20 +703,14 @@ public void invalidateClientCache() {
...
-     * @param result list of read data
-     *
-     * @return map with valid data (not trimmed) or TrimmedException is thrown is data is trimmed and
-     * flag is not set.
+     * Cache list of returned data with given cache options.
      */
-    private Map<Long, ILogData> checkForTrimmedAddresses(Map<Long, ILogData> result, ReadOptions options) {
-        List<Long> trimmedAddresses = new ArrayList<>();
-
+    private Map<Long, ILogData> checkForCompactedAddresses(Map<Long, ILogData> result, ReadOptions options) {
```
lixinchengdu: Delete unnecesssary put

lixinchengdu: done.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
lixinchengdu: Yes. Once one address is detected as compacted, it should be not saved in the stream layer as well as address space cache.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -785,7 +758,7 @@ ILogData fetch(final long address) {
                 .read(e, address)
         );
 
-        checkLogDataThrowException(address, result);
+        checkLogData(address, result);
```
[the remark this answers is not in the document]
lixinchengdu: This is a good question! This should be addressed in future.

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
lixinchengdu: I also noticed this problem and addressed in the most recent PR.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
lixinchengdu: If this is a issue, I will address it in the branch called compactionMarkInRuntime.

PR #2159, opened 2019-09-13
Benchmark module. Large corfu table benchmarks: RocksDb vs EhCache

xnull:
## Overview

Benchmarks for Large/Persistent Corfu table backed by EhCache vs RocksDb.

Description:
Benchmarking corf table backed by EhCache vs RocksDb. 
Provides performance characteristics over the implementations that proves advantages/disadvantages one approach over another one.
Documentation https://github.com/CorfuDB/CorfuDB/blob/large_table_benchmarks/docs/benchmarks/corfu-table.md

The prototypes of the persistent CorfTable (EhCache/RocksDb) can/will be used to implement a performant production-ready Corfu table implementation.

Why should this be merged: 

- [ ] Public API has Javadoc


docs/benchmarks/corfu-table.md
```diff
@@ -0,0 +1,73 @@
...
+ 
+![Get. Four Threads performance benchmark, 1 million records](img/four-threads-get-1-mil.jpg)
+
+EhCache can't manage 10 million records data set, it throws GC error.
+
+RocksDb 50-100% faster than EhCache for single threaded application.
+RocksDb provides better performance for multithreaded applications.
+
+Conclusion:
+  - RocksDb more stable and solid solution then EhCache
```
hisundar: Would be necessary to compare the resource consumptions of the two engines compared for a fair comparison before making conclusions right?
(RocksDb is not designed or optimized to be an ephemeral datastore - it is designed to be a fast persistent datastore.)

xnull: I compare persistency features, so according to the numbers, it works better for us in this particular case.
Agree with you, maybe my statement is too strong.
Any ideas about how the sentence can be modified?

benchmarks/src/jmh/java/org/corfudb/benchmarks/util/DataGeneratorBenchmark.java
```diff
@@ -0,0 +1,126 @@
...
+    @Fork(3)
+    public void multiThreadedStringGenerator(Blackhole blackhole, DataGeneratorState state) {
+        String value = DataGenerator.generateDataString(state.getDataSize());
+        blackhole.consume(value);
+    }
+
+    /**
+     * Java random generator benchmark
+     *
+     * @param blackhole kmh blackhole
```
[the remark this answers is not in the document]
xnull: fixed

benchmarks/src/jmh/java/org/corfudb/benchmarks/runtime/collections/CorfuTableBenchmark.java
```diff
@@ -0,0 +1,175 @@
...
+     */
+    @Benchmark
+    public void hashMapGet(HashMapStateForGet state, Blackhole blackhole) {
+        CorfuTableBenchmarkHelper helper = state.getHelper();
+        String value = helper.getTable().get(helper.generate());
+        blackhole.consume(value);
+    }
+
+    /**
+     * Get/Put 50x50 load for HashMap
```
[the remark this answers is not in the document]
xnull: renamed

benchmarks/src/jmh/java/org/corfudb/benchmarks/runtime/collections/CorfuTableBenchmark.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public static void main(String[] args) throws RunnerException {
+
+        String benchmarkName = CorfuTableBenchmark.class.getSimpleName();
+
+        Options opt = new OptionsBuilder()
+                .include(benchmarkName)
+
+                .mode(Mode.Throughput)
+                .timeUnit(TimeUnit.SECONDS)
```
PavelZaytsev: Perhaps pass mode, timeUnit, warmupIterations, warmupTime, measurementIterations, measurementTime, threads, forks as a configurable parameters?

benchmarks/src/jmh/java/org/corfudb/benchmarks/runtime/collections/state/RocksDbStateForPut.java
```diff
@@ -0,0 +1,74 @@
...
+                .dataSize(dataSize)
+                .tableSize(tableSize)
+                .build()
+                .check();
+    }
+
+    @TearDown
+    public void tearDown() throws RocksDBException, IOException {
+        RocksDbMap<Integer, String> rocksDbMap = helper.getUnderlyingMap();
+        //log.info(rocksDbMap.getStats());
```
PavelZaytsev: Remove instead of comment out

xnull: For `production` code yes. Here I wanted to show how we can print statistics, if I remove it, it will be unclear how to use it

benchmarks/src/jmh/java/org/corfudb/benchmarks/runtime/collections/CorfuTableBenchmark.java
xnull: done

benchmarks/src/main/java/org/corfudb/benchmarks/runtime/collections/experiment/rocksdb/RocksDbMap.java
```diff
@@ -0,0 +1,259 @@
...
+    @NonNull
+    @Default
+    private final Options options = new Options()
+            .setCreateIfMissing(true)
+            .setWriteBufferSize((long) 256 * (long) 1024 * (long) 1024)
+            .setMaxWriteBufferNumber(4)
+            .setIncreaseParallelism(8)
+            //.setBloomLocality(10)
+
+            //.setOptimizeFiltersForHits(true)
```
zalokhan: Remove commented code.

zalokhan:
This PR is rather large.
Please break this down into 2 PRs:
Benchmark framework
Benchmark tests.

This looks like a POC to get a comparison. We can publish numbers for both but have tests only for ehcache as we are implementing only that. It will be tough to maintain the APIs for both in the future and the tests will get outdated.

xnull:
@zalokhan I don't agree, breaking down into 2 pr-s is a bunch of useless work. 
 1. What's the profit if that 2 pr-s will go one after another immediately?
 2. Benchmark framework is nothing but a JMH dependency and maven module. It's not worth creating a module with a dependency that doesn't contain any code.
 3. Next time if you need to compare Ehcache and rocks again, are you going to implement those benchmark tests one more time? Benchmarks are needed to show/prove the performance of a particular code/library, even if that library is not used in code, also the benchmark proves why a particular library/code used instead of another one.
 4.  It won't be tough to maintain the APIs because if your API changed dramatically then you have to prove that there is no critical performance degradation in your code which means you have to update tests and prove that it shows expected performance or you can delete tests if they not relevant anymore. 
We have the same number of arguments like in case of unit tests - unit tests can get obsolete for example.  That's true but that the reason why we have unit tests - check that nothing is broken. Same works for benchmark tests.
 5. I believe that any critical code must go through benchmarks and only then can be added into code-base. Because of this benchmark tests, we realized that EhCache doesn't work for us and we won't implement it in code and going to migrate to rocksDb.
6. When we have an implementation of CorfuTable backed by rocks or Ehcache in code we will modify benchmark tests accordingly. Nothing can get broken or get obsolete.

benchmarks/src/main/java/org/corfudb/benchmarks/runtime/collections/experiment/rocksdb/RocksDbMap.java
xnull: Not sure if we should remove it. 
1. Those parameters can be used to adjust rocksDb performance. They can be used for a quick check of how the DB behaves with different settings.
2. Since it's not production code I believe it's ok to have those settings commented.
3. There is no rich documentation for rocksdb_jni library and it's not immediately obvious how to set up those parameters when you read documentation and see that settings for c++ version. The commented code can help/clarify that.

benchmarks/src/jmh/java/org/corfudb/benchmarks/runtime/collections/CorfuTableBenchmark.java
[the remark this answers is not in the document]
xiaoqin2012: It will be nice to add updates and deletes too for complete fair comparison.

docs/benchmarks/corfu-table.md
hisundar: Maybe you can just qualify it saying "RocksDb is more performant than EhCache" ?

benchmarks/src/jmh/java/org/corfudb/benchmarks/runtime/collections/CorfuTableBenchmark.java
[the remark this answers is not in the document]
xnull: 1. Not sure how it can be measured without affecting the benchmark by `put` operation, you need to put some data and only then remove. The `update` operation is the same as `put`
2. What could be nice to measure is the speed of `compaction` algorithm, but it's for next pr-s

docs/benchmarks/corfu-table.md
xnull: done

xnull:
> like google or https://imgbb.com
wouldn't like to do that, don't want to be dependent on an external resource.
The images are very small
image

PR #2163, opened 2019-09-17
CorfuStore: TxBuilder, ProtobufSerializer, Indexer [PART 2]

hisundar:
Adding CorfuStore with its TxBuilder
Tests with sample schemas


runtime/proto/corfu_options.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto2";
```
xnull: so, can you tell me again what's the problem with proto3?

runtime/proto/corfu_store_metadata.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
```
xnull: in a file above it's proto2

runtime/proto/corfu_options.proto
```diff
@@ -0,0 +1,18 @@
...
+
+package org.corfudb.runtime;
+option java_package = "org.corfudb.runtime";
+
+import "google/protobuf/descriptor.proto";
+
+// Option tags to be used by the CorfuStore consumers to tag special fields to be detected by Corfu.
+message SchemaOptions {
+    // Secondary keys to be indexed.
+    optional bool secondary_key = 1;
```
xnull: if you can use proto3 please get rid of `optional` here

runtime/proto/corfu_store_metadata.proto
```diff
@@ -0,0 +1,38 @@
...
+// Descriptor to provide self describing schemas for the records.
+message TableDescriptors {
+    google.protobuf.DescriptorProto keyDescriptor = 1;
+    google.protobuf.DescriptorProto valueDescriptor = 2;
+}
+
+// Metadata.
+message TableMetadata {
+    TableName table_name = 1;
+    bool disk_based = 2;
```
xnull: looks like disk_based and cache should be an enum
```
enum Persistent {
  disk_based, cache
}
```

runtime/proto/corfu_store_metadata.proto
```diff
@@ -0,0 +1,38 @@
...
+    bool cache = 3;
+}
+
+// Record to be persisted containing the value and the metadata.
+message Record {
+    google.protobuf.Any payload = 1;
+    google.protobuf.Any metadata = 2;
```
xnull: not asking why metadata is any, hope you have a reason for that

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
```diff
@@ -0,0 +1,14 @@
+package org.corfudb.runtime.collections;
+
+import com.google.protobuf.Message;
+
+import lombok.Data;
+
+/**
+ * Created by zlokhandwala on 2019-08-15.
+ */
+@Data
```
xnull: Please never use `@data` and `@value` annotations, it's overcomplicated. Use not compound annotations like `@Getter`, `@ToString` and so on

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
```diff
@@ -0,0 +1,14 @@
...
+public class CorfuRecord<V extends Message, M extends Message> {
+    private final V payload;
```
xnull: why payload is `V` not `P` at least?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata.TableName;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+
+/**
+ * Created by zlokhandwala on 2019-08-02.
+ */
+public class CorfuStore {
+
+    private final CorfuRuntime runtime;
```
xnull: can you pass `SequencerView` and `TableRegistry` as parameters instead of runtime?
There are at least 3 disadvantages of using rt:
 - rt is a huge object, it's overweighted GOD object
 - rt brings circular dependencies between every single class using rt
 - harder to write tests if rt is used in code

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param <K>          Key type.
+     * @param <V>          Value type.
+     * @return Table instance.
+     * @throws NoSuchMethodException     Thrown if key/value class are not protobuf classes.
+     * @throws InvocationTargetException Thrown if key/value class are not protobuf classes.
+     * @throws IllegalAccessException    Thrown if key/value class are not protobuf classes.
+     */
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message>
+    Table<K, V, M> createTable(@Nonnull final String namespace,
```
xnull: you can use `TableName` instead of `namespace` and `tableName` right?

xnull: So many parameters in the method signature. Is it possible to reduce it?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata.TableName;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+
+/**
+ * Created by zlokhandwala on 2019-08-02.
+ */
+public class CorfuStore {
```
xnull: Why we even have this class? It's just a wrapper for TableRegistry, it doesn't contain any valuable logic. 
I don't see any reason to have this class in the code

runtime/src/main/java/org/corfudb/runtime/collections/MetadataOptions.java
```diff
@@ -0,0 +1,21 @@
...
+import com.google.protobuf.Message;
+import lombok.Builder;
+import lombok.Data;
+
+import static lombok.Builder.Default;
+
+/**
+ * Created by zlokhandwala on 2019-09-04.
+ */
+@Data
```
xnull: Don't use `@Data` see my comment above

runtime/src/main/java/org/corfudb/runtime/collections/MetadataOptions.java
```diff
@@ -0,0 +1,21 @@
...
+@Builder
+class MetadataOptions {
+
+    @Default
+    private final boolean metadataEnabled = false;
+
+    @Default
+    private final Message defaultMetadataInstance = null;
```
xnull: null? Maybe use Optional then

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Iterator;
+import java.util.Optional;
+
+/**
+ * Created by hisundar on 2019-08-12.
+ */
+public class ProtobufIndexer implements CorfuTable.IndexRegistry<Message, CorfuRecord<Message, Message>> {
+
+    private HashMap<String,
+            CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>>
```
xnull: looks really overcomplicated

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Optional;
+
+/**
+ * Created by hisundar on 2019-08-12.
+ */
+public class ProtobufIndexer implements CorfuTable.IndexRegistry<Message, CorfuRecord<Message, Message>> {
+
+    private HashMap<String,
```
xnull: final?

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,65 @@
...
+            CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>>
+            indices = new HashMap<>();
+
+    ProtobufIndexer(Message payloadSchema) {
+        payloadSchema.getDescriptorForType().getFields().forEach(this::registerIndices);
+    }
+
+    private <T extends Comparable<T>> CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>
```
xnull: Can you please create a particular (domain) interface (that will extend Comparable<T>) and the interface name can explain what is that. Because right now it's completely unclear why Comparable used here

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,65 @@
...
+    getIndex(String indexName, FieldDescriptor fieldDescriptor) {
+
+        return new Index<>(
+                () -> indexName,
+                (CorfuTable.IndexFunction<Message, CorfuRecord<Message, Message>, T>)
+                        (key, val) -> (T) val.getPayload().getField(fieldDescriptor));
```
xnull: please use `ClassUtil.cast`

xnull: Wat if filed is not present?

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    void registerIndices(final Descriptors.FieldDescriptor fieldDescriptor) {
+
+        if (fieldDescriptor.getOptions().getExtension(CorfuOptions.schema).getSecondaryKey()) {
+            final String indexName = fieldDescriptor.getName();
+            if (fieldDescriptor.getType() == FieldDescriptor.Type.GROUP) {
+                throw new IllegalArgumentException("group is a deprecated, unsupported type");
```
xnull: why not return `Result<Exception, List<Index>>` instead of throwing runtime exception?

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,65 @@
...
+            }
+            indices.put(indexName, getIndex(indexName, fieldDescriptor));
+        }
+    }
+
+    @Override
+    public Optional<CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>> get(
+            CorfuTable.IndexName name) {
+
+        String indexName = (name != null) ? name.get() : null;
```
xnull: please use Optional instead of null

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,65 @@
...
+        CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>> index = indices.get(indexName);
+        if (index != null) {
```
xnull: this block can be replaced by `Optional.ofNullable(index);`

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+     * Returns a Table instance backed by a CorfuTable.
+     *
+     * @param namespace               Namespace of the table.
+     * @param fullyQualifiedTableName Fully qualified table name.
+     * @param valueSchema             Value schema to identify secondary keys.
+     * @param corfuRuntime            Connected instance of the Corfu Runtime.
+     * @param serializer              Protobuf Serializer.
+     */
+    @Nonnull
+    public Table(@Nonnull final String namespace,
```
xnull: should be used TableName?

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+                 @Nonnull final String fullyQualifiedTableName,
+                 @Nonnull final V valueSchema,
+                 @Nullable final M metadataSchema,
+                 @Nonnull final CorfuRuntime corfuRuntime,
+                 @Nonnull final ISerializer serializer) {
+
+        this.corfuRuntime = corfuRuntime;
+        this.namespace = namespace;
+        this.fullyQualifiedTableName = fullyQualifiedTableName;
+        if (metadataSchema != null) {
```
xnull: please use Optional

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+            this.metadataOptions = MetadataOptions.builder()
+                    .metadataEnabled(true)
+                    .defaultMetadataInstance(metadataSchema)
+                    .build();
+        } else {
+            this.metadataOptions = MetadataOptions.<M>builder().build();
+        }
+        this.corfuTable = corfuRuntime.getObjectsView().build()
+                .setTypeToken(new TypeToken<CorfuTable<K, CorfuRecord<V, M>>>() {
```
xnull: Please use `CorfuTable.getTableType()`

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+                .setSerializer(serializer)
+                .setArguments(new ProtobufIndexer(valueSchema))
+                .open();
+    }
+
+    /**
+     * Begins a write after write transaction.
+     */
+    private boolean TxBegin() {
+        if (!TransactionalContext.isInTransaction()) {
```
xnull: invert `if` condition

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+            corfuRuntime.getObjectsView()
+                    .TXBuild()
+                    .type(TransactionType.OPTIMISTIC)
+                    .build().begin();
```
xnull: andd enter

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Ends an ongoing transaction.
+     */
+    private void TxEnd() {
+        corfuRuntime.getObjectsView().TXEnd();
```
xnull: how table can manage transactions?
It violates logic of transactions

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+     * @param metadata Record metadata.
+     * @return Previously stored record if any.
+     */
+    @Nullable
+    CorfuRecord<V, M> create(@Nonnull final K key,
+                             @Nullable final V value,
+                             @Nullable final M metadata) {
+        boolean beganNewTxn = false;
+        try {
+            beganNewTxn = TxBegin();
```
xnull: *begin

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
xnull: I don't like the Idea of managing transactions inside table class

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Fetch the value for a key.
+     *
+     * @param key Key.
+     * @return Corfu Record for key.
+     */
+    @Nullable
+    public CorfuRecord<V, M> get(@Nonnull final K key) {
+        return corfuTable.get(key);
```
xnull: check if corfutable returns null

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+     * @param key      Key.
+     * @param value    Value.
+     * @param metadata Metadata.
+     * @return Previously stored value for the provided key.
+     */
+    @Nullable
+    CorfuRecord<V, M> update(@Nonnull final K key,
+                             @Nonnull final V value,
+                             @Nullable final M metadata) {
+        boolean beganNewTxn = false;
```
xnull: *begin

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /**
+     * Scan and filter.
+     *
+     * @param p Predicate to filter the values.
+     * @return Collection of filtered values.
+     */
+    @Nonnull
+    Collection<CorfuRecord<V, M>> scanAndFilter(@Nonnull final Predicate<CorfuRecord<V, M>> p) {
```
xnull: Why collection? Why not List or Set?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+              @Nonnull final String namespace) {
+        this.objectsView = objectsView;
+        this.tableRegistry = tableRegistry;
+        this.namespace = namespace;
+        this.operations = new ArrayList<>();
+    }
+
+    private void validateNamespace(Table table) {
+        if (table.getNamespace().equals(this.namespace)) {
+            return;
```
xnull: add a log message

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+        }
+        throw new IllegalArgumentException(
```
xnull: can be used `Result` instead of throwing runtime exception

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+                String.format("Transaction namespace: %s. Attempted transaction on table in namespace: %s",
+                        this.namespace, table.getNamespace()));
+    }
+
+    private <M extends Message> void validateVersion(@Nullable M previousMetadata,
+                                                     @Nullable M userMetadata) {
+        if (userMetadata == null) {
```
xnull: please use Optional instead of checking for null

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+            return;
```
xnull: add a log message

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+        }
+        for (Descriptors.FieldDescriptor fieldDescriptor : userMetadata.getDescriptorForType().getFields()) {
+            if (fieldDescriptor.getOptions().getExtension(CorfuOptions.schema).getVersion()) {
```
xnull: invert if condition

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+                if (!versionTypes.contains(fieldDescriptor.getType())) {
+                    throw new IllegalArgumentException("Version field needs to be an Integer or Long type.");
+                }
+                long validatingVersion = (long) userMetadata.getField(fieldDescriptor);
+                long previousVersion = Optional.ofNullable(previousMetadata)
+                        .map(m -> (Long) m.getField(fieldDescriptor))
+                        .orElse(-1L);
+                if (validatingVersion != previousVersion) {
+                    throw new RuntimeException("Stale object Exception");
```
xnull: don't throw RuntimeException it's too generic

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+     * @param tableName Table name to perform the create on.
+     * @param key       Key.
+     * @param value     Value to create.
+     * @param <K>       Type of Key.
+     * @param <V>       Type of Value.
+     * @return TxBuilder instance.
+     */
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message>
+    TxBuilder create(@Nonnull final String tableName,
```
xnull: should be class TableName?

xnull: in the other case we don't know what namespace we should use

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+                     @Nonnull final K key,
+                     @Nonnull final V value,
+                     @Nullable final M metadata) {
+        Table<K, V, M> table = getTable(tableName);
+        validateNamespace(table);
+        operations.add(() -> {
+            CorfuRecord<V, M> previous = table.get(key);
+            if (previous != null) {
+                throw new RuntimeException("Cannot create a record on existing key.");
```
xnull: RuntimeException too generic

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+            }
+            M newMetadata = null;
```
xnull: optional

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+     * @param tableName Table name to perform the update on.
+     * @param key       Key.
+     * @param value     Value to update.
+     * @param <K>       Type of Key.
+     * @param <V>       Type of Value.
+     * @return TxBuilder instance.
+     */
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message>
+    TxBuilder update(@Nonnull final String tableName,
```
xnull: TableName?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+            M previousMetadata = Optional.ofNullable(previous)
+                    .map(CorfuRecord::getMetadata)
+                    .orElseGet(() -> {
+                        if (table.getMetadataOptions().isMetadataEnabled()) {
+                            return (M) table.getMetadataOptions().getDefaultMetadataInstance();
+                        }
+                        return null;
+                    });
+            M newMetadata = null;
+            if (previousMetadata != null) {
```
xnull: optional

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+     * @return TxBuilder instance.
+     */
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message>
+    TxBuilder touch(@Nonnull final String tableName,
+                    @Nonnull final K key) {
+        Table<K, V, M> table = getTable(tableName);
+        validateNamespace(table);
+        operations.add(() -> {
+            //TODO: Validate the get is executed.
```
xnull: get rid of TODO

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+     * Commit the transaction.
+     * The commit call begins a Corfu transaction at the latest timestamp, applies all the updates and then
+     * ends the Corfu transaction.
+     * The commit returns successfully if the transaction was committed.
+     * Otherwise this throws a TransactionAbortedException.
+     */
+    public void commit() {
+        commit(null);
```
xnull: null?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+     * The commit returns successfully if the transaction was committed.
+     * Otherwise this throws a TransactionAbortedException.
+     *
+     * @param timestamp Timestamp to commit the transaction on.
+     */
+    public void commit(final Timestamp timestamp) {
+        this.timestamp = timestamp;
+        try {
+            txBegin();
+            operations.forEach(Runnable::run);
```
xnull: it's not async then what's the point for Runnable? Please provide your own interface for that.
Would be great if your `Runnable` could return `Result` class, then you can get rid of `finally` section (I mean you can simplify it)

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Created by zlokhandwala on 2019-08-10.
+ */
+public class TableRegistry {
+
+    private static final String CORFU_SYSTEM_NAMESPACE = "CorfuSystem";
+    private static final String REGISTRY_TABLE_NAME = "RegistryTable";
+
+    private final CorfuRuntime runtime;
```
xnull: we pass CorfuRuntime everywhere, it's a good idea

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private final Map<String, Class<? extends Message>> classMap;
```
xnull: ConcurrentMap

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private final Map<String, Table<Message, Message, Message>> tableMap;
```
xnull: ConcurrentMap

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private final ISerializer protobufSerializer;
+
+    private final CorfuTable<TableName, CorfuRecord<TableDescriptors, Message>> registryTable;
+
+    public TableRegistry(CorfuRuntime runtime) {
+        this.runtime = runtime;
+        this.classMap = new ConcurrentHashMap<>();
+        this.tableMap = new ConcurrentHashMap<>();
+        this.protobufSerializer = new ProtobufSerializer((byte) 25, classMap);
```
xnull: magic number

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+                .build();
+        TableDescriptors tableDescriptors = TableDescriptors.newBuilder()
+                .setKeyDescriptor(keyDescriptor)
+                .setValueDescriptor(valueDescriptor)
+                .build();
+
+        try {
+            this.runtime.getObjectsView().TXBuild().type(TransactionType.OPTIMISTIC).build().begin();
+            this.registryTable.putIfAbsent(tableNameKey,
+                    new CorfuRecord<>(tableDescriptors, null));
```
xnull: is it possible to avoid using null?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+    private <T extends Message> void addTypeToClassMap(T msg) {
+        String typeUrl = getTypeUrl(msg.getDescriptorForType());
+        // Register the schemas to schema table.
+        if (!classMap.containsKey(typeUrl)) {
+            classMap.put(typeUrl, msg.getClass());
+        }
+    }
+
+    public <K extends Message, V extends Message, M extends Message>
+    Table<K, V, M> openTable(@Nonnull final String namespace,
```
xnull: TableName

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+                             @Nonnull final String tableName,
+                             @Nonnull final Class<K> kClass,
+                             @Nonnull final Class<V> vClass,
+                             @Nullable final Class<M> mClass,
+                             @Nonnull final TableOptions tableOptions)
+            throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+
+        // Register the schemas to schema table.
+        K defaultKeyMessage = (K) kClass.getMethod("getDefaultInstance").invoke(null);
```
xnull: why just not to pass a default instance?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+        return (Table<K, V, M>) tableMap.get(fullyQualifiedTableName);
+    }
+
+    public void deleteTable(String namespace, String tableName) {
+    }
+
+    public Collection<TableName> listTables(final String namespace) {
+        return registryTable.keySet()
+                .stream()
+                .filter(tableName -> namespace == null || tableName.getNamespace().equals(namespace))
```
xnull: don't use null

runtime/proto/corfu_options.proto
zalokhan: This was just to test whether it has an impact on the consumption of corfu_options when the application protobuf is proto 2 or 3. This was initially to allow backward compatibility such that the user can have a proto2 environment and still be able to consume our files. This, however, did not work.
I can change this back to proto3.

runtime/proto/corfu_store_metadata.proto
zalokhan: These are 2 different options.
Disk based is for backing the Corfu tables by a disk.
Cache is for non-cacheable streams.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
zalokhan: Will make the change.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
zalokhan: V was to distinguish from M. V and M stand for Value and Metadata.
This also keeps it uniform with the naming conventions used in the rest of the CorfuStore.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
zalokhan: CorfuStore is a single entity which the client instantiates and uses to create tables, perform CRUD, queries, etc. The client should not have visibility into the different parts of the runtime to instantiate the Corfu Store.
The corfuStore also needs the ObjectsView to handle transactions.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
zalokhan: Done.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
zalokhan: Not needed. `@Nullable` clarifies to the user to expect null values.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: It is literally the CorfuTable's Indexer. No added complexity outside of it.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: done

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: But won't that mean that I have to change CorfuTable's Indexer too - that's complicated as it is.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Just want to reduce the footprint of the create. We do need the actual type for the table, now we will need to ask the default instance of it as well. It makes the caller a lot more verbose.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: namespace comes in via the public api and is stored as such. @Nullable is used to clarify to caller the same.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Creates and registers a table.
+     * A table needs to be registered before it is used.
+     *
+     * @param namespace    Namespace of the table.
+     * @param tableName    Table name.
+     * @param kClass       Class of the Key Model.
+     * @param vClass       Class of the Value Model.
```
pankti-m: nit - update header to include meta class

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Fetches an existing table. This table should have been registered with this instance
+     * of the Corfu runtime.
+     *
+     * @param namespace Namespace of the table.
+     * @param tableName Table name.
+     * @param <K>       Key type.
+     * @param <V>       Value type.
```
pankti-m: nit - update header to include metadata

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: The means the client will have to agree to prefix all table names with the namespace prefix which we cannot expect them to do. That is why there are 2 separate asks - namespace and table name where same table name can exist in multiple namespaces.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
vjeko: You should have some sort of a TxManager for this logic.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,230 @@
...
+        return new ArrayList<>(corfuTable.scanAndFilter(p));
```
vjeko: Scan and filter should return a Stream.

runtime/src/main/java/org/corfudb/runtime/collections/TableOptions.java
```diff
@@ -0,0 +1,13 @@
+package org.corfudb.runtime.collections;
+
+import org.corfudb.runtime.collections.CorfuTable.IndexRegistry;
+import lombok.Builder;
+
+/**
+ * Created by zlokhandwala on 2019-08-09.
+ */
+@Builder
+public class TableOptions<K, V> {
```
vjeko: Comments.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+import org.corfudb.runtime.object.transactions.Transaction;
+import org.corfudb.runtime.object.transactions.TransactionType;
+import org.corfudb.runtime.view.ObjectsView;
+import org.corfudb.runtime.view.TableRegistry;
+
+/**
+ * Created by zlokhandwala on 2019-08-05.
+ */
+public class TxBuilder {
```
vjeko: Comments!

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+              @Nonnull final TableRegistry tableRegistry,
+              @Nonnull final String namespace) {
+        this.objectsView = objectsView;
+        this.tableRegistry = tableRegistry;
+        this.namespace = namespace;
+        this.operations = new ArrayList<>();
+    }
+
+    private void validateNamespace(Table table) {
+        if (table.getNamespace().equals(this.namespace)) {
```
vjeko: I don't understand why this logic is not negated. It would make it simpler.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
vjeko: Either make it Optional as Slava said or NonNull.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+                String.format("Transaction namespace: %s. Attempted transaction on table in namespace: %s",
+                        this.namespace, table.getNamespace()));
+    }
+
+    private <M extends Message> void validateVersion(@Nullable M previousMetadata,
+                                                     @Nullable M userMetadata) {
+        if (userMetadata == null) {
+            return;
+        }
+        for (Descriptors.FieldDescriptor fieldDescriptor : userMetadata.getDescriptorForType().getFields()) {
```
vjeko: Like break please. 100 is max.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
vjeko: Does .getVersion() return boolean? I don't get this line.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+                                                 @Nullable M userMetadata) {
+        M.Builder builder = previousMetadata.toBuilder();
+        for (Descriptors.FieldDescriptor fieldDescriptor : previousMetadata.getDescriptorForType().getFields()) {
+            if (fieldDescriptor.getOptions().getExtension(CorfuOptions.schema).getVersion()) {
+                builder.setField(
+                        fieldDescriptor,
+                        Optional.ofNullable(previousMetadata.getField(fieldDescriptor))
+                                .map(previousVersion -> ((Long) previousVersion) + 1)
+                                .orElse(0L));
+            } else if (userMetadata != null){
```
vjeko: space

vjeko: Why is this else-if? I am not following the logic.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message>
+    TxBuilder create(@Nonnull final String tableName,
+                     @Nonnull final K key,
+                     @Nonnull final V value,
+                     @Nullable final M metadata) {
+        Table<K, V, M> table = getTable(tableName);
+        validateNamespace(table);
+        operations.add(() -> {
+            CorfuRecord<V, M> previous = table.get(key);
```
vjeko: Why just not call table.create()? Doesn't that throw the same exception. If not, why is there a discrepancy?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+    public <K extends Message, V extends Message, M extends Message>
+    TxBuilder update(@Nonnull final String tableName,
+                     @Nonnull final K key,
+                     @Nonnull final V value,
+                     @Nullable final M metadata) {
+        Table<K, V, M> table = getTable(tableName);
+        validateNamespace(table);
+        operations.add(() -> {
+            CorfuRecord<V, M> previous = table.get(key);
+            M previousMetadata = Optional.ofNullable(previous)
```
vjeko: What are the semantic of update(). Is create() required?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
vjeko: Make it Optional.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Commit the transaction.
+     * The commit call begins a Corfu transaction at the specified snapshot or at the latest snapshot
+     * if no snapshot is specified, applies all the updates and then ends the Corfu transaction.
+     * The commit returns successfully if the transaction was committed.
+     * Otherwise this throws a TransactionAbortedException.
+     *
+     * @param timestamp Timestamp to commit the transaction on.
+     */
+    public void commit(final Timestamp timestamp) {
```
vjeko: I don't get it. Why is timeStamp required here? Is it even being used?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,306 @@
...
+     * @throws InvocationTargetException If this is not a protobuf message.
+     * @throws IllegalAccessException    If this is not a protobuf message.
+     */
+    public <K extends Message, V extends Message, M extends Message>
+    Table<K, V, M> openTable(@Nonnull final String namespace,
+                             @Nonnull final String tableName,
+                             @Nonnull final Class<K> kClass,
+                             @Nonnull final Class<V> vClass,
+                             @Nullable final Class<M> mClass,
+                             @Nonnull final TableOptions tableOptions)
```
vjeko: Same point here. Why don't we move this to tableOptions.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,306 @@
...
+        return (Table<K, V, M>) tableMap.get(fullyQualifiedTableName);
+    }
+
+    /**
+     * Deletes a table.
+     *
+     * @param namespace Namespace of the table.
+     * @param tableName Name of the table.
+     */
+    public void deleteTable(String namespace, String tableName) {
```
vjeko: Thow UnsupportedEx.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
vjeko: namespace is effectively a constant. Isn't this going to return all-or-nothing?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * Gets the table descriptors for a particular table name.
+     * This is used for reconstructing a message when the schema is not available.
+     *
+     * @param tableName Namespace and name of the table.
+     * @return Table Descriptor.
+     */
+    @Nullable
+    public TableDescriptors getTableDescriptor(@Nonnull TableName tableName) {
```
vjeko: Why is this not returning Optional?

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata.Record;
+import org.corfudb.runtime.collections.CorfuRecord;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class ProtobufSerializer implements ISerializer {
```
vjeko: This is a pretty important component. Please put detailed comments.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: this is from client's api, if not null then we will force them to pass either an Optional or a dummy metadata, both of which can be more complex than null which they have asked for right now.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: done

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -30,14 +30,7 @@
...
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
-import org.corfudb.runtime.view.AddressSpaceView;
-import org.corfudb.runtime.view.Layout;
-import org.corfudb.runtime.view.LayoutManagementView;
-import org.corfudb.runtime.view.LayoutView;
-import org.corfudb.runtime.view.ManagementView;
-import org.corfudb.runtime.view.ObjectsView;
-import org.corfudb.runtime.view.SequencerView;
-import org.corfudb.runtime.view.StreamsView;
+import org.corfudb.runtime.view.*;
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Fetch the value for a key.
+     *
+     * @param key Key.
+     * @return Corfu Record for key.
+     */
+    @Nullable
+    public CorfuRecord<V, M> get(@Nonnull final K key) {
```
WenbinZhu: Do we support snapshot read? Is that introduced in part 3 patch?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+     * @return TxBuilder instance.
+     */
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message>
+    TxBuilder create(@Nonnull final String tableName,
+                     @Nonnull final K key,
+                     @Nonnull final V value,
+                     @Nullable final M metadata) {
+        Table<K, V, M> table = getTable(tableName);
+        validateNamespace(table);
```
WenbinZhu: Why `validateNamespace` is  needed here? When calling getTable, we use the `namespace` in this class, and find the table registered by this namespace, unless there is a bug in tableRegistry, right?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+        operations.add(() -> {
+            CorfuRecord<V, M> previous = table.get(key);
+            if (previous != null) {
```
WenbinZhu: I think this check should be in table.create() method itself, not here. Currently table.create() is exactly same as table.update(). Also if table.create() is used outside TxBuilder, we won't check existence, which is inconsistent with the TxBuilder.create() semantics.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,230 @@
...
+     * @param namespace               Namespace of the table.
+     * @param fullyQualifiedTableName Fully qualified table name.
+     * @param valueSchema             Value schema to identify secondary keys.
+     * @param corfuRuntime            Connected instance of the Corfu Runtime.
+     * @param serializer              Protobuf Serializer.
+     */
+    @Nonnull
+    public Table(@Nonnull final String namespace,
+                 @Nonnull final String fullyQualifiedTableName,
+                 @Nonnull final V valueSchema,
```
WenbinZhu: Should this be named as payloadSchema?

runtime/src/main/java/org/corfudb/runtime/collections/MetadataOptions.java
```diff
@@ -0,0 +1,21 @@
...
+import lombok.Data;
+
+import static lombok.Builder.Default;
+
+/**
+ * Created by zlokhandwala on 2019-09-04.
+ */
+@Data
+@Builder
+class MetadataOptions {
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
xiaoqin2012: The key type and value type are in the comments but are not part of the function arguments. Should be removed from the comments?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
xiaoqin2012: Same here: The key type and value type are in the comments but are not part of the function arguments. Should be removed from the comments?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,311 @@
...
+     */
+    public Collection<TableName> listTables(@Nullable final String namespace) {
+        return registryTable.keySet()
+                .stream()
+                .filter(tableName -> namespace == null || tableName.getNamespace().equals(namespace))
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Gets the table descriptors for a particular table name.
```
annym: for a particular fully qualified table name.

As we are using namespace and table name for its parts and not the composed.

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
```diff
@@ -0,0 +1,143 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * The Protobuf serializer is the main component that allows CorfuStore to use Protobufs to
+ * convert a language specific (Java here) object into an identifiable byte buffer that
+ * can be then converted back to a language specific object.
+ *
+ * To achieve this, this serializer requires a map of all seen class types or a classMap
+ * and Google Protobuf 3's Any.
+ * Any type carries with it a typeUrl which help identify the class uniquely.
```
annym: typo ---> helps

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
```diff
@@ -0,0 +1,143 @@
...
+     * Deserialize an object from a given byte buffer.
+     *
+     * @param b The bytebuf to deserialize.
+     * @return The deserialized object.
+     */
+    @Override
+    public Object deserialize(ByteBuf b, CorfuRuntime rt) {
+
+        try (ByteBufInputStream bbis = new ByteBufInputStream(b)) {
+            MessageType type = MessageType.valueOf(bbis.readInt());
```
annym: type could be null, right?

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
zalokhan: Yes.

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
```diff
@@ -60,4 +66,34 @@ public void basicQueueOrder() {
         assertThat(records.get(0).compareTo(records2.get(1))).isLessThan(0);
         assertThat(records.get(0).getRecordId().compareTo(records2.get(1).getRecordId())).isLessThan(0);
     }
+
+    @Test
```
annym: Add comments

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
```diff
@@ -0,0 +1,310 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.corfudb.test.SampleSchema.ManagedResources;
+
+/**
+ * Created by zlokhandwala on 2019-08-12.
+ */
+public class CorfuStoreTest extends AbstractViewTest {
+
+    @Test
```
annym: Can we add descriptions to all these tests.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
```diff
@@ -0,0 +1,310 @@
...
+                .update(tableName,
+                        key1,
+                        EventInfo.newBuilder().setName("cde").build(),
+                        null)
+                .commit();
+        assertThat(corfuStore.getTable(nsxManager, tableName).get(key1).getMetadata())
+                .isNull();
+    }
+
+//    @Test
```
annym: Why is this test commented?

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
```diff
@@ -0,0 +1,14 @@
...
+    private final M metadata;
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata.TableName;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+
+/**
+ * Created by zlokhandwala on 2019-08-02.
```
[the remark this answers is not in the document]
hisundar: done. Sorry I think my push and your comment crossed each other at about the same time.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,230 @@
...
+    /**
+     * Begins a write after write transaction.
+     */
+    private boolean TxBegin() {
+        if (TransactionalContext.isInTransaction()) {
+            return false;
+        }
+        corfuRuntime.getObjectsView()
+                .TXBuild()
+                .type(TransactionType.OPTIMISTIC)
```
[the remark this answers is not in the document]
hisundar: OPTIMISTIC are necessary for read conflict detection via the touch api, so I will fix up the doc to remove WRITE_AFTER_WRITE
thanks!

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,230 @@
...
+                .open();
+    }
+
+    /**
+     * Returns the underlying CorfuTable object.
+     * NOTE: For internal use only.
+     *
+     * @return CorfuTable instance.
+     */
+    public CorfuTable<K, CorfuRecord<V, M>> getUnderlyingObject() {
```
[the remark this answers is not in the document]
hisundar: it is/will be used in the checkpointer code on the application side.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,230 @@
...
+    @Getter
+    private final MetadataOptions metadataOptions;
+
+    /**
+     * Returns a Table instance backed by a CorfuTable.
+     *
+     * @param namespace               Namespace of the table.
+     * @param fullyQualifiedTableName Fully qualified table name.
+     * @param valueSchema             Value schema to identify secondary keys.
+     * @param corfuRuntime            Connected instance of the Corfu Runtime.
```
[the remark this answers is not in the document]
hisundar: fixed. thanks

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/**
+ * Created by hisundar on 2019-08-12.
+ */
+public class ProtobufIndexer implements CorfuTable.IndexRegistry<Message, CorfuRecord<Message, Message>> {
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: It is a tradeoff between having multiple non-null parameters in the constructor which will catch null at compile time vs having a small footprint constructor which takes in null but then fails it only at runtime.
Seems like a compile time check is better than a runtime failure right?

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
hisundar: Thanks Anny, I think there is value in adding this test which demonstrates how to use DynamicMessage. Let me uncomment and update it.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
hisundar: done.

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
hisundar: done.

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
hisundar: should not be if data was written by one of Corfu's layers.

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
hisundar: thanks

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: thanks and yes.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * Gets the type Url of the protobuf descriptor. Used to identify the message during serialization.
+     * Note: This is same as used in Any.proto.
+     *
+     * @param descriptor Descriptor of the protobuf.
+     * @return Type url string.
+     */
+    private String getTypeUrl(Descriptor descriptor) {
+        return "type.googleapis.com/" + descriptor.getFullName();
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,311 @@
...
+    private final CorfuRuntime runtime;
+
+    /**
+     * Stores the schemas of the Key, Value and Metadata.
+     * A reference of this map is held by the {@link ProtobufSerializer} to serialize and deserialize the objects.
+     */
+    private final ConcurrentMap<String, Class<? extends Message>> classMap;
+
+    /**
+     * Cache of tables allowing the user to fetch a table by the namespace and table name without the other options.
```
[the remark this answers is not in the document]
hisundar: yes, thanks!

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -0,0 +1,311 @@
...
+import org.corfudb.runtime.object.transactions.TransactionType;
+import org.corfudb.util.serializer.ISerializer;
+import org.corfudb.util.serializer.ProtobufSerializer;
+import org.corfudb.util.serializer.Serializers;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Table Registry manages the table lifecycle.
```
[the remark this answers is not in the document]
hisundar: good point. let me rephrase. thanks

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -0,0 +1,270 @@
...
+
+    private <M extends Message> void validateVersion(@Nullable M previousMetadata,
+                                                     @Nullable M userMetadata) {
+        if (userMetadata == null) {
+            return;
+        }
+        for (Descriptors.FieldDescriptor fieldDescriptor : userMetadata.getDescriptorForType().getFields()) {
+            if (fieldDescriptor.getOptions().getExtension(CorfuOptions.schema).getVersion()) {
+                if (!versionTypes.contains(fieldDescriptor.getType())) {
+                    throw new IllegalArgumentException("Version field needs to be an Integer or Long type.");
```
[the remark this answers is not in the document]
hisundar: yes it is, thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * @param runtime Connected instance of the Corfu Runtime.
+     */
+    @Nonnull
+    public CorfuStore(@Nonnull final CorfuRuntime runtime) {
+        this.runtime = runtime;
+    }
+
+    /**
+     * Fetches the latest logical timestamp in Corfu's distributed log.
```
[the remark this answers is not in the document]
hisundar: yes, updated comment.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
hisundar: done. switched to payload

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
hisundar: Good point. Thanks I will move it as you suggest.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
hisundar: validateNamespace is used in case someone tries to start a transaction on one namespace but adds in a table which belongs to a different namespace.
It's to catch a user error.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -0,0 +1,130 @@
...
+        return runtime.getTableRegistry().getTable(namespace, tableName);
+    }
+
+    /**
+     * Deletes a table instance. [NOT SUPPORTED.]
+     *
+     * @param namespace Namespace of the table.
+     * @param tableName Table name.
+     */
+    private void deleteTable(String namespace, String tableName) {
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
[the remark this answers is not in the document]
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -0,0 +1,230 @@
...
+     * @param serializer              Protobuf Serializer.
+     */
+    @Nonnull
+    public Table(@Nonnull final String namespace,
+                 @Nonnull final String fullyQualifiedTableName,
```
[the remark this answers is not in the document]
hisundar: no its just a small optimization to prevent generating a fullyQualifiedTableName every time.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.google.protobuf.Message;
+
+import lombok.Data;
+
+/**
+ * Created by zlokhandwala on 2019-08-15.
+ */
+@Data
+public class CorfuRecord<V extends Message, M extends Message> {
```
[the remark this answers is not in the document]
hisundar: done.

PR #2164, opened 2019-09-17
CorfuStore: Add Query, QueryOptions, QueryResult [PART 3]

hisundar:
Adds the Query API

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
@@ -2,9 +2,8 @@ package org.corfudb.runtime.collections;
 
 import com.google.protobuf.Message;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
-
-import java.util.Objects;
 
 /**
  * Encapsulates the payload/value and metadata into one object.
@@ -15,28 +14,23 @@ import java.util.Objects;
  * modifications.
  *
  */
+@EqualsAndHashCode
 public class CorfuRecord<V extends Message, M extends Message> {
+    /**
+     * V encapsulates the user's Value payload - this is the main protobuf message that defines the schema
+     */
     @Getter
     private final V payload;
+
+    /**
+     * M encapsulates the user's metadata - this can be something like auto-incrementing versions, other
+     * services provided by the database for metadata fields.
+     */
     @Getter
     private final M metadata;
 
     public CorfuRecord(V payload, M metadata) {
         this.payload = payload;
         this.metadata = metadata;
-    }
-
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        CorfuRecord<?, ?> that = (CorfuRecord<?, ?>) o;
-        return Objects.equals(getPayload(), that.getPayload()) &&
-                Objects.equals(getMetadata(), that.getMetadata());
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(getPayload(), getMetadata());
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
@@ -7,11 +7,11 @@ import com.google.protobuf.Message;
 import org.corfudb.common.util.ClassUtils;
 import org.corfudb.runtime.CorfuOptions;
 import org.corfudb.runtime.collections.CorfuTable.Index;
+import org.corfudb.runtime.collections.CorfuTable.IndexName;
 
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Optional;
-import java.util.function.Supplier;
 
 /**
  * This layer implements the
@@ -23,15 +23,15 @@ import java.util.function.Supplier;
  */
 public class ProtobufIndexer implements CorfuTable.IndexRegistry<Message, CorfuRecord<Message, Message>> {
 
-    private HashMap<String,
-            CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>>
+    private final HashMap<String,
+            Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>>
             indices = new HashMap<>();
 
     ProtobufIndexer(Message payloadSchema) {
         payloadSchema.getDescriptorForType().getFields().forEach(this::registerIndices);
     }
 
-    private <T extends Comparable<T>> CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>
+    private <T extends Comparable<T>> Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>
     getIndex(String indexName, FieldDescriptor fieldDescriptor) {
 
         return new Index<>(
@@ -52,18 +52,8 @@ public class ProtobufIndexer implements CorfuTable.IndexRegistry<Message, CorfuR
     }
 
     @Override
-    public Optional<CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>> get(
-            CorfuTable.IndexName name) {
-
-        String indexName = Optional.ofNullable(name)
-                .map(Supplier::get)
-                .orElse(null);
-        CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>> index = indices.get(indexName);
-        if (index != null) {
-            return Optional.of(index);
-        } else {
-            return Optional.empty();
-        }
+    public Optional<Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>> get(IndexName name) {
+        return Optional.ofNullable(name).map(indexName -> indices.get(indexName));
     }
 
     @Override
--- a/runtime/src/main/java/org/corfudb/runtime/collections/Query.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/Query.java
@@ -1,0 +1,431 @@
+package org.corfudb.runtime.collections;
+
+import static org.corfudb.runtime.collections.QueryOptions.DEFAULT_OPTIONS;
+
+import com.google.protobuf.Message;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.TreeSet;
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+import org.corfudb.runtime.object.transactions.Transaction.TransactionBuilder;
+import org.corfudb.runtime.object.transactions.TransactionType;
+import org.corfudb.runtime.view.ObjectsView;
+import org.corfudb.runtime.view.TableRegistry;
+
+/**
+ * Query class provides methods to query the CorfuStore tables.
+ * It has several methods to query by secondary index, to perform predicate joins and apply merge filters.
+ *
+ * Created by zlokhandwala on 2019-08-09.
+ */
+public class Query {
+
+    private final TableRegistry tableRegistry;
+
+    private final ObjectsView objectsView;
+
+    private final String namespace;
+
+    /**
+     * Creates a Query interface.
+     *
+     * @param tableRegistry Table registry from the corfu client.
+     * @param objectsView   Objects View from the corfu client.
+     * @param namespace     Namespace to perform the queries within.
+     */
+    public Query(final TableRegistry tableRegistry, final ObjectsView objectsView, final String namespace) {
+        this.tableRegistry = tableRegistry;
+        this.objectsView = objectsView;
+        this.namespace = namespace;
+    }
+
+    private Token getToken(Timestamp timestamp) {
+        return new Token(timestamp.getEpoch(), timestamp.getSequence());
+    }
+
+    private void txBegin(Timestamp timestamp) {
+        TransactionBuilder transactionBuilder = objectsView.TXBuild()
+                .type(TransactionType.SNAPSHOT);
+
+        if (timestamp != null) {
+            transactionBuilder.snapshot(getToken(timestamp));
+        }
+
+        transactionBuilder.build().begin();
+    }
+
+    private void txEnd() {
+        objectsView.TXEnd();
+    }
+
+    private <K extends Message, V extends Message, M extends Message>
+    Table<K, V, M> getTable(@Nonnull final String tableName) {
+        return tableRegistry.getTable(this.namespace, tableName);
+    }
+
+    /**
+     * Fetch the CorfuRecord for the specified key at the LATEST timestamp (default).
+     *
+     * @param tableName Table name.
+     * @param key       Key.
+     * @param <K>       Type of key.
+     * @param <V>       Type of the payload/value.
+     * @param <M>       Type of the metadata.
+     * @return CorfuRecord encapsulating the payload and the metadata.
+     */
+    @Nullable
+    public <K extends Message, V extends Message, M extends Message>
+    CorfuRecord<V, M> getRecord(@Nonnull final String tableName,
+                                @Nonnull final K key) {
+        return getRecord(tableName, null, key);
+    }
+
+    /**
+     * Fetch the Value payload for the specified key at the specified snapshot/timestamp.
+     *
+     * @param tableName Table name.
+     * @param timestamp Timestamp to perform the query.
+     * @param key       Key.
+     * @param <K>       Type of key.
+     * @param <V>       Type of value/payload.
+     * @param <M>       Type of metadata.
+     * @return Value.
+     */
+    @Nullable
+    public <K extends Message, V extends Message, M extends Message>
+    CorfuRecord<V, M> getRecord(@Nonnull final String tableName,
+                          @Nullable final Timestamp timestamp,
+                          @Nonnull final K key) {
+        if (tableName == null) {
+            throw new IllegalArgumentException("Query::getRecord needs a valid tableName");
+        }
+        try {
+            txBegin(timestamp);
+            Table<K, V, M> table = getTable(tableName);
+            return table.get(key);
+        } finally {
+            txEnd();
+        }
+    }
+
+    /**
+     * Gets the count of records in the table.
+     *
+     * @param tableName Table name.
+     * @return Count of records.
+     */
+    public int count(@Nonnull final String tableName) {
+        return count(tableName, null);
+    }
+
+    /**
+     * Gets the count of records in the table at a particular timestamp.
+     *
+     * @param tableName Table name.
+     * @param timestamp Timestamp to perform the query on.
+     * @return Count of records.
+     */
+    public int count(@Nonnull final String tableName,
+                     @Nullable final Timestamp timestamp) {
+        try {
+            txBegin(timestamp);
+            return getTable(tableName).count();
+        } finally {
+            txEnd();
+        }
+    }
+
+    @Nonnull
+    private <K extends Message, V extends Message, M extends Message>
+    Collection<CorfuRecord<V, M>> scanAndFilter(@Nonnull final String tableName,
+                                                @Nullable Timestamp timestamp,
+                                                @Nonnull final Predicate<CorfuRecord<V, M>> p) {
+        try {
+            txBegin(timestamp);
+            return ((Table<K, V, M>) getTable(tableName)).scanAndFilter(p);
+        } finally {
+            txEnd();
+        }
+    }
+
+    /**
+     * Query by a secondary index.
+     *
+     * @param tableName Table name.
+     * @param indexName Index name. In case of protobuf-defined secondary index it is the field name.
+     * @param indexKey  Key to query.
+     * @param <K>       Type of Key.
+     * @param <V>       Type of Value.
+     * @param <I>       Type of index/secondary key.
+     * @return Result of the query.
+     */
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message, I extends Comparable<I>>
+    QueryResult<Entry<K, V>> getByIndex(@Nonnull final String tableName,
+                                        @Nonnull final String indexName,
+                                        @Nonnull final I indexKey) {
+        return new QueryResult<>(((Table<K, V, M>) getTable(tableName)).getByIndex(indexName, indexKey));
+    }
+
+    private <V extends Message, M extends Message, R>
+    Collection<R> initializeResultCollection(QueryOptions<V, M, R> queryOptions) {
+        if (!queryOptions.isDistinct()) {
+            return new ArrayList<>();
+        }
+        if (queryOptions.getComparator() != null) {
+            return new TreeSet<>(queryOptions.getComparator());
+        }
+        return new HashSet<>();
+    }
+
+    private <V extends Message, M extends Message, R>
+    Collection<R> transform(Collection<CorfuRecord<V, M>> queryResult,
+                            Collection<R> resultCollection,
+                            Function<CorfuRecord<V, M>, R> projection) {
+        return queryResult.stream()
+                .map(v -> Optional.ofNullable(projection)
+                        .map(function -> function.apply(v))
+                        .orElse((R) v))
+                .collect(Collectors.toCollection(() -> resultCollection));
+    }
+
+    /**
+     * Execute a scan and filter query.
+     *
+     * @param tableName Table name.
+     * @param query     Predicate to filter the values.
+     * @param <V>       Type of Value.
+     * @param <R>       Type of returned projected values.
+     * @return Result of the query.
+     */
+    @Nonnull
+    public <V extends Message, M extends Message, R>
+    QueryResult<R> executeQuery(@Nonnull final String tableName,
+                                 @Nonnull final Predicate<CorfuRecord<V, M>> query) {
+        return executeQuery(tableName, query, DEFAULT_OPTIONS);
+    }
+
+    /**
+     * Execute a scan and filter query.
+     * The query options enables to define:
+     * a projection function,
+     * timestamp at which the table needs to be queried,
+     * Flag to return only distinct results.
+     *
+     * @param tableName    Table name.
+     * @param query        Predicate to filter the values.
+     * @param queryOptions Query options.
+     * @param <V>          Type of Value.
+     * @param <R>          Type of returned projected values.
+     * @return Result of the query.
+     */
+    @Nonnull
+    public <V extends Message, M extends Message, R>
+    QueryResult<R> executeQuery(@Nonnull final String tableName,
+                                 @Nonnull final Predicate<CorfuRecord<V, M>> query,
+                                 @Nonnull final QueryOptions<V, M, R> queryOptions) {
+
+        Collection<CorfuRecord<V, M>> filterResult = scanAndFilter(tableName, queryOptions.getTimestamp(), query);
+        Collection<R> result = initializeResultCollection(queryOptions);
+        return new QueryResult<>(transform(filterResult, result, queryOptions.getProjection()));
+    }
+
+    /**
+     * Execute a join of 2 tables.
+     *
+     * @param tableName1     Table name 1.
+     * @param tableName2     Table name 2.
+     * @param query1         Predicate to filter entries in table 1.
+     * @param query2         Predicate to filter entries in table 2.
+     * @param joinPredicate  Predicate to filter entries during the join.
+     * @param joinFunction   Function to merge entries.
+     * @param joinProjection Project the merged entries.
+     * @param <Val1>         Type of Value in table 1.
+     * @param <Val2>         Type of Value in table 2.
+     * @param <T>            Type of resultant value after merging type V and type W.
+     * @param <U>            Type of value projected from T.
+     * @return Result of query.
+     */
+    @Nonnull
+    public <Val1 extends Message, Val2 extends Message, Meta1 extends Message, Meta2 extends Message, T, U>
+    QueryResult<U> executeJoinQuery(
+            @Nonnull final String tableName1,
+            @Nonnull final String tableName2,
+            @Nonnull final Predicate<CorfuRecord<Val1, Meta1>> query1,
+            @Nonnull final Predicate<CorfuRecord<Val2, Meta2>> query2,
+            @Nonnull final BiPredicate<Val1, Val2> joinPredicate,
+            @Nonnull final BiFunction<Val1, Val2, T> joinFunction,
+            final Function<T, U> joinProjection) {
+        return executeJoinQuery(
+                tableName1,
+                tableName2,
+                query1,
+                query2,
+                DEFAULT_OPTIONS,
+                DEFAULT_OPTIONS,
+                joinPredicate,
+                joinFunction,
+                joinProjection);
+    }
+
+    /**
+     * Execute a join of 2 tables.
+     *
+     * @param tableName1     Table name 1.
+     * @param tableName2     Table name 2.
+     * @param query1         Predicate to filter entries in table 1.
+     * @param query2         Predicate to filter entries in table 2.
+     * @param queryOptions1  Query options to transform table 1 filtered values.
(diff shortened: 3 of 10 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+     *
+     * @param tableName Table name.
+     * @param key       Key.
+     * @param <K>       Type of key.
+     * @param <V>       Type of value.
+     * @return Value.
+     */
+    @Nullable
+    public <K extends Message, V extends Message, M extends Message>
+    CorfuRecord<V, M> get(@Nonnull final String tableName,
```
WenbinZhu: Should this return value instead of CorfuRecord? This currently looks exactly same as getRecord().

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+     * @param key       Key.
+     * @param <K>       Type of key.
+     * @param <V>       Type of value.
+     * @return Corfu Record.
+     */
+    @Nullable
+    public <K extends Message, V extends Message, M extends Message>
+    CorfuRecord<V, M> getRecord(@Nonnull final String tableName,
+                                @Nonnull final K key) {
+        return get(tableName, null, key);
```
WenbinZhu: Should this be `return getRecord(tableName, null, key)`?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+    QueryResult<Entry<K, V>> getByIndex(@Nonnull final String tableName,
+                                        @Nonnull final String indexName,
+                                        @Nonnull final I indexKey) {
+        return new QueryResult<>(((Table<K, V, M>) getTable(tableName)).getByIndex(indexName, indexKey));
+    }
+
+    private <V extends Message, M extends Message, R>
+    Collection<R> initializeResultCollection(QueryOptions<V, M, R> queryOptions) {
+        if (!queryOptions.isDistinct()) {
+            return new ArrayList<>();
```
WenbinZhu: Here we return immediately, does that mean if the queryOption is set to "Not distinct" AND "Need order (comparator != null)", then the ordering is ignored?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+     *
+     * @param tableName Table name.
+     * @param query     Predicate to filter the values.
+     * @param <V>       Type of Value.
+     * @param <R>       Type of returned projected values.
+     * @return Result of the query.
+     */
+    @Nonnull
+    public <V extends Message, M extends Message, R>
+    QueryResult<R> exectuteQuery(@Nonnull final String tableName,
```
WenbinZhu: typo: exectute -> execute

WenbinZhu: If executeQuery is only for scanAndFilter, why not just name it scanAndFilter to be consistent with other APIs like get() and getByIndex()?  Naming it executeQuery  sounds like JDBC way of  doing query, which seems weird in this context since it's not consistent with other APIs.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+     * @param tableName    Table name.
+     * @param query        Predicate to filter the values.
+     * @param queryOptions Query options.
+     * @param <V>          Type of Value.
+     * @param <R>          Type of returned projected values.
+     * @return Result of the query.
+     */
+    @Nonnull
+    public <V extends Message, M extends Message, R>
+    QueryResult<R> exectuteQuery(@Nonnull final String tableName,
```
WenbinZhu: typo: exectute -> execute

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+        }
+        return new HashSet<>();
+    }
+
+    private <V extends Message, M extends Message, R>
+    Collection<R> transform(Collection<CorfuRecord<V, M>> queryResult,
+                            Collection<R> resultCollection,
+                            Function<CorfuRecord<V, M>, R> projection) {
+        return queryResult.stream()
+                .map(v -> Optional.ofNullable(projection)
```
WenbinZhu: Can we check projection being null in the beginning for just once, instead of checking every time for each record?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+     * @param <Val2>         Type of Value in table 2.
+     * @param <R>            Type of projected values from table 1 from type V.
+     * @param <S>            Type of projected values from table 2 from type W.
+     * @param <T>            Type of resultant value after merging type R and type S.
+     * @param <U>            Type of value projected from T.
+     * @return Result of query.
+     */
+    @Nonnull
+    public <Val1 extends Message, Val2 extends Message, Meta1 extends Message, Meta2 extends Message, R, S, T, U>
+    QueryResult<U> executeJoinQuery(
```
WenbinZhu: We can just call it join to be consistent with other APIs.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+
+        Collection<CorfuRecord<Val2, Meta2>> filterResult2 = scanAndFilter(tableName2, queryOptions2.getTimestamp(), query2);
+        Collection<S> queryResult2 = transform(
+                filterResult2,
+                initializeResultCollection(queryOptions2),
+                queryOptions2.getProjection());
+
+        Collection<T> joinResult = new ArrayList<>();
+
+        for (R value1 : queryResult1) {
```
WenbinZhu: As discussed, the ordering will be lost after join, right now it's okay as we don't have the requirement, later we can add a joinOption if needed. Just bookkeeping the discussion here.

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
```diff
@@ -0,0 +1,74 @@
...
+                         boolean distinct,
+                         Comparator<R> comparator,
+                         Function<CorfuRecord<V, M>, R> projection) {
+        this.timestamp = timestamp;
+        this.distinct = distinct;
+        this.comparator = comparator;
+        this.projection = projection;
+    }
+
+    public static class QueryOptionsBuilder<V extends Message, M extends Message, R> {
```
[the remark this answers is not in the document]
zalokhan: The lombok builder had issues with generics and I couldn't get it to work.

test/src/test/java/org/corfudb/integration/CorfuStorePerfIT.java
```diff
@@ -0,0 +1,145 @@
...
+        Map<UUID, Event> map = rt.getObjectsView()
+            .build()
+            .setStreamName("s1")
+            .setType(CorfuTable.class)
+            .open();
+
+        Event event;
+        for(int i=0; i<count; i++) {
+            event = new Event(i, "TestEvent"+i, 8000,
+                20000, 10
```
LynneD: maybe use variable or pass in as parameters instead of using magic numbers.

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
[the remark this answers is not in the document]
annym: Isn't it more consistent to call it Builder so we keep it in the form: QueryOptions.Builder()
.....
.build()??

test/src/test/java/org/corfudb/integration/CorfuStorePerfIT.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Test
+    public void corfuStorePerfComparisonTest() throws Exception {
+        long maxLogSize = FileUtils.ONE_MB / 2;
+        Process server1 = runServerWithQuota(DEFAULT_PORT, maxLogSize,
+            true);
+
+        CorfuRuntime rt = new CorfuRuntime(DEFAULT_ENDPOINT).connect();
+        final int count = 100;
+        addObjectsToTable(rt, count);
```
annym: Is the intention to compare these two values? put some assertion in place.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
```diff
@@ -99,11 +102,104 @@ public void basicTest() throws Exception {
...
+                        .setTableName(tableName)
+                        .build());
+
         table.update(lookupKey, EventInfo.getDefaultInstance(), null);
         assertThat(table.get(lookupKey))
                 .isEqualTo(new CorfuRecord<>(EventInfo.getDefaultInstance(), null));
     }
 
+    @Test
+    public void comparisonTest() throws Exception {
```
annym: Can we have a test description.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
```diff
@@ -99,11 +102,104 @@ public void basicTest() throws Exception {
                 .setEventTime(fifty)
                 .build();
 
+        assertThat((Message) q.get(tableName, timestamp, lookupKey)).isNull();
```
annym: do we have any test that exercises a non existing timestamp?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: sorry for the confusion but the purpose of overloading this get() method is to have one with timestamp specified and one without. Note below there is a get() where caller can explicitly specify a timestamp to run the query on. 
I have updated the comment to reflect the same. 
thanks

hisundar: got rid of these extra get() and just retained getRecord

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: yes, thanks!

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: thanks!

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: yes, we plan to add better support for range iteration and query later.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: You are right but this was an explicit ask to be consistent with the way the Query module is in use by our current consumers in different verticals.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: thanks!

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+     * @param <V>       Type of value.
+     * @return Value.
+     */
+    @Nullable
+    public <K extends Message, V extends Message, M extends Message>
+    CorfuRecord<V, M> get(@Nonnull final String tableName,
+                          @Nullable final Timestamp timestamp,
+                          @Nonnull final K key) {
+        try {
+            txBegin(timestamp);
```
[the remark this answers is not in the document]
hisundar: I suppose the table will sync to it? It is the same as how corfu's snapshot transaction would handle it since this is just a wrapper around the same.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: good suggestion. done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: sure. thanks

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+
+        return new QueryResult<>(joinResult.stream()
+                .map(v -> Optional.ofNullable(joinProjection)
+                        .map(function -> function.apply(v))
+                        .orElse((U) v))
+                .collect(Collectors.toList()));
+    }
+
+    /**
+     * Merge Function which
```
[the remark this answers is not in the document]
hisundar: fixed. thanks

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+         */
+        R merge(List<Object> arguments);
+    }
+
+    private <R> List<R> merge(@Nonnull List<Collection<?>> list,
+                              @Nonnull List<Object> mergePayload,
+                              @Nonnull MergeFunction<R> func,
+                              int depth) {
+        Collection<?> collection = list.get(depth);
+        if (list.size() - 1 == depth) {
```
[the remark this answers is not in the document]
hisundar: done.

test/src/test/java/org/corfudb/integration/CorfuStorePerfIT.java
hisundar: done. thanks

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
hisundar: will be coming up. thanks

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
```diff
@@ -0,0 +1,14 @@
+package org.corfudb.runtime.collections;
+
+import com.google.protobuf.Message;
+
+import lombok.Data;
+
+/**
+ * Created by zlokhandwala on 2019-08-15.
+ */
```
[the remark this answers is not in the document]
xnull: Please add what V, M means - it's value and metadata if I'm not wrong

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
```diff
@@ -0,0 +1,42 @@
...
+    @Getter
+    private final M metadata;
+
+    public CorfuRecord(V payload, M metadata) {
+        this.payload = payload;
+        this.metadata = metadata;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
xnull: you can use `@EqualsAndHashcode`

runtime/src/main/java/org/corfudb/runtime/collections/MetadataOptions.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * MetadataOptions extracts user's metadata schema, examines it for FieldOptions
+ * that need to be processed by the CorfuStore (example: version that increments on every update)
+ * and maintains simple flags on what was processed.
+ * It is an optimization to avoid having to parse the protobuf every time to determine if
+ * a metadata operation needs to be performed.
+ *
+ * Created by zlokhandwala on 2019-09-04.
+ */
+@Data
```
xnull: Get rid of `@Data` please

runtime/src/main/java/org/corfudb/runtime/collections/MetadataOptions.java
```diff
@@ -0,0 +1,27 @@
...
+@Builder
+class MetadataOptions {
+
+    @Default
+    private final boolean metadataEnabled = false;
+
+    @Default
+    private final Message defaultMetadataInstance = null;
```
xnull: remove null, it's not needed

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,74 @@
...
+ * This layer implements the
+ * ProtobufIndexer uses the special FieldOptions that the application can place on its
+ * protobuf definitions (like secondary_key) and create secondary indexes callbacks over CorfuTable
+ * based on that.
+ *
+ * Created by hisundar on 2019-08-12.
+ */
+public class ProtobufIndexer implements CorfuTable.IndexRegistry<Message, CorfuRecord<Message, Message>> {
+
+    private HashMap<String,
```
xnull: please add final

xnull: I'm worried about race condition here

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,74 @@
...
+            final String indexName = fieldDescriptor.getName();
+            if (fieldDescriptor.getType() == FieldDescriptor.Type.GROUP) {
+                throw new IllegalArgumentException("group is a deprecated, unsupported type");
+            }
+            indices.put(indexName, getIndex(indexName, fieldDescriptor));
+        }
+    }
+
+    @Override
+    public Optional<CorfuTable.Index<Message, CorfuRecord<Message, Message>, ? extends Comparable<?>>> get(
```
xnull: please add static import for Index

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -0,0 +1,74 @@
...
+            CorfuTable.IndexName name) {
+
+        String indexName = Optional.ofNullable(name)
```
xnull: why you transform Optional to null? Optional is used to get rid of null :)

xnull: you can transform it to:
```
return Optional.ofNullable(name).map(indexName -> indices.get(indexName));
```

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: done. thanks.

runtime/src/main/java/org/corfudb/runtime/collections/MetadataOptions.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/MetadataOptions.java
hisundar: already done in previous commit. thanks

test/src/test/java/org/corfudb/integration/CorfuStoreIT.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.Test;
+
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CorfuStoreIT extends AbstractIT {
+
+    @Test
+    public void testTx() throws Exception {
```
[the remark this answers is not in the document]
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: indexes are created only at the time of table creation and only read afterwards. so there is no concurrent mutations that can result in a race condition here right?

xnull: yep

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
hisundar: thanks Slava!

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
hisundar: done. thanks

test/src/test/java/org/corfudb/integration/CorfuStorePerfIT.java
hisundar: no its just to run these two for verifying that both methods work and for very basic performance tests.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+import org.corfudb.runtime.object.transactions.Transaction.TransactionBuilder;
+import org.corfudb.runtime.object.transactions.TransactionType;
+import org.corfudb.runtime.view.ObjectsView;
+import org.corfudb.runtime.view.TableRegistry;
+
+/**
+ * Created by zlokhandwala on 2019-08-09.
```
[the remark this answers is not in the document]
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuRecord.java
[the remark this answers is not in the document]
hisundar: done thanks!

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: want to keep same name as what the clients are familiar with if that's ok?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,449 @@
...
+    private Token getToken(Timestamp timestamp) {
+        return new Token(timestamp.getEpoch(), timestamp.getSequence());
+    }
+
+    private void txBegin(Timestamp timestamp) {
+        TransactionBuilder transactionBuilder = objectsView.TXBuild()
+                .type(TransactionType.SNAPSHOT);
+
+        if (timestamp != null) {
+            transactionBuilder.snapshot(getToken(timestamp));
```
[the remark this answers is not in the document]
hisundar: some exception from beginTransaction of Corfu

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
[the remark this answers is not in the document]
hisundar: want to keep same name as what the clients use currently if that's ok?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,424 @@
...
+     * @return Value.
+     */
+    @Nullable
+    public <K extends Message, V extends Message, M extends Message>
+    CorfuRecord<V, M> getRecord(@Nonnull final String tableName,
+                          @Nullable final Timestamp timestamp,
+                          @Nonnull final K key) {
+        try {
+            txBegin(timestamp);
+            return ((Table<K, V, M>) getTable(tableName)).get(key);
```
xnull: Looks like NPE cold happen here

xnull: Also you can use ClassUtil.cast(...) instead of direct casting

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,424 @@
...
+            return new ArrayList<>();
+        }
+        if (queryOptions.getComparator() != null) {
+            return new TreeSet<>(queryOptions.getComparator());
+        }
+        return new HashSet<>();
+    }
+
+    private <V extends Message, M extends Message, R>
+    Collection<R> transform(Collection<CorfuRecord<V, M>> queryResult,
```
xnull: Can we use List or Set here or anything else? As we discussed using Collection is almost a bad idea

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: Good catch! thanks Slava!

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,424 @@
...
+        Collection<CorfuRecord<Val2, Meta2>> filterResult2 = scanAndFilter(tableName2, queryOptions2.getTimestamp(), query2);
+        Collection<S> queryResult2 = transform(
+                filterResult2,
+                initializeResultCollection(queryOptions2),
+                queryOptions2.getProjection());
+
+        Collection<T> joinResult = new ArrayList<>();
+
+        for (R value1 : queryResult1) {
+            for (S value2 : queryResult2) {
```
xnull: You could use:
```
queryResult2.stream().filter(...).map(...).toList()
```

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,424 @@
...
+    @FunctionalInterface
+    public interface MergeFunction<R> {
+
+        /**
+         * Merges the specified argument list.
+         *
+         * @param arguments Objects across tables to be merged.
+         * @return Resultant object.
+         */
+        R merge(List<Object> arguments);
```
xnull: Can we replace `Object` by `R`?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,424 @@
...
+                }
+            }
+            return result;
+        }
+
+        List<R> result = new ArrayList<>();
+        for (Object o : collection) {
+            List<Object> mergeList = new ArrayList<>(mergePayload);
+            mergeList.add(o);
+            result.addAll(merge(list, mergeList, func, depth + 1));
```
xnull: not sure if a recursive call a good idea

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -0,0 +1,424 @@
...
+        List<Collection<? extends Object>> values = new ArrayList<>();
+        for (String tableName : tableNames) {
+            Collection<Message> messages = scanAndFilter(tableName, null, corfuRecord -> true)
+                    .stream()
+                    .map(corfuRecord -> corfuRecord.getPayload())
+                    .collect(Collectors.toList());
+            values.add(messages);
+        }
+
+        Collection<R> mergedResults = merge(values, new ArrayList<>(), joinFunction, 0);
```
xnull: what does 0 mean?

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: refactored this code based on your previous comment. thanks

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
```diff
@@ -0,0 +1,74 @@
...
+
+        private Timestamp timestamp = null;
```
xnull: delete null

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: In fact let me add a test case for this !

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
```diff
@@ -0,0 +1,74 @@
...
+        private boolean distinct = false;
+        private Comparator<R> comparator = null;
```
xnull: delete null

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
```diff
@@ -0,0 +1,74 @@
...
+        private Function<CorfuRecord<V, M>, R> projection = null;
```
xnull: delete nll, use Optional if needed

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
```diff
@@ -0,0 +1,74 @@
...
+
+        public static <VALUE extends Message, META extends Message, S>
+        QueryOptionsBuilder<VALUE, META, S> newBuilder() {
+            return new QueryOptionsBuilder<>();
+        }
+
+        public QueryOptionsBuilder<V, M, R> setTimestamp(Timestamp timestamp) {
```
xnull: for all these methods use can use `@Setter(chain=true)` (don't remember exact syntax)

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
```diff
@@ -0,0 +1,74 @@
...
+            this.comparator = comparator;
+            return this;
+        }
+
+        public QueryOptionsBuilder<V, M, R> setProjection(Function<CorfuRecord<V, M>, R> projection) {
+            this.projection = projection;
+            return this;
+        }
+
+        public QueryOptions<V, M, R> build() {
```
xnull: `@Builder` can be used

runtime/src/main/java/org/corfudb/runtime/collections/QueryResult.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.Collection;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+/**
+ * Created by zlokhandwala on 2019-08-10.
+ */
+@Data
```
xnull: get rid of `@Data`

runtime/src/main/java/org/corfudb/runtime/collections/QueryResult.java
```diff
@@ -0,0 +1,16 @@
...
+@AllArgsConstructor
+public class QueryResult<E> {
+
+    private final Collection<E> result;
```
xnull: I vote for using List or Set

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: The last parameter is the depth of the multi table join. In the first cut we are only going to support merging 2 tables at a time, perhaps in the future we can support multi table joins, but for now, we are just passing in 0.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
hisundar: again this is just a first cut. on top of this commit, there will be performance tests and correctness tests where we will likely revisit these changes.

runtime/src/main/java/org/corfudb/runtime/collections/QueryResult.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
hisundar: facing test issues with @Setter will try to do this in the subsequent patch

runtime/src/main/java/org/corfudb/runtime/collections/QueryOptions.java
hisundar: facing test issues with this will try to do this in the subsequent patch

PR #2166, opened 2019-09-18
Change how we initialize wrapper objects

vjeko:
Previously, we were initializing a wrapper object with any constructor
that would match the given arguments. This approach is not sufficient
when constructor type parameters are polymorphic, since the wrong
constructor might get called.

Please take a look at Java Specification 15.12.2.5. for more
information.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/ReflectionUtils.java
+++ b/runtime/src/main/java/org/corfudb/util/ReflectionUtils.java
@@ -1,13 +1,21 @@
 package org.corfudb.util;
 
 import com.google.common.collect.ImmutableMap;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.ClassUtils;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


vjeko:
@Maithem if you take a look at the source code:


```java
        // search through all methods 
        int paramSize = parameterTypes.length;
        Constructor[] ctors = clazz.getConstructors();
        for (int i = 0, size = ctors.length; i < size; i++) {
            // compare parameters
            Class[] ctorParams = ctors[i].getParameterTypes();
            int ctorParamSize = ctorParams.length;
            if (ctorParamSize == paramSize) {
                boolean match = true;
                for (int n = 0; n < ctorParamSize; n++) {
                    if (!MethodUtils
                        .isAssignmentCompatible(
                            ctorParams[n],
                            parameterTypes[n])) {
                        match = false;
                        break;
                    }
                }

                if (match) {
                    // get accessible version of method
                    Constructor ctor = getAccessibleConstructor(ctors[i]);
                    if (ctor != null) {
                        try {
                            ctor.setAccessible(true);
                        } catch (SecurityException se) {}
                        return ctor;
                    }
                }
            }
        }
```
You can see that they do very simplistic matching. Very similar to what we used to do. In fast, I tried this library first when I noticed the bug, but it had the same outcome.

Also, there is nothing in the Java specification about "Choosing the Most Specific Constructor" , so I assume JLS 15.12.2.5 refers constructor matching as well.

Maithem:
> @Maithem if you take a look at the source code:
> 
> ```java
>         // search through all methods 
>         int paramSize = parameterTypes.length;
>         Constructor[] ctors = clazz.getConstructors();
>         for (int i = 0, size = ctors.length; i < size; i++) {
>             // compare parameters
>             Class[] ctorParams = ctors[i].getParameterTypes();
>             int ctorParamSize = ctorParams.length;
>             if (ctorParamSize == paramSize) {
>                 boolean match = true;
>                 for (int n = 0; n < ctorParamSize; n++) {
>                     if (!MethodUtils
>                         .isAssignmentCompatible(
>                             ctorParams[n],
>                             parameterTypes[n])) {
>                         match = false;
>                         break;
>                     }
>                 }
> 
>                 if (match) {
>                     // get accessible version of method
>                     Constructor ctor = getAccessibleConstructor(ctors[i]);
>                     if (ctor != null) {
>                         try {
>                             ctor.setAccessible(true);
>                         } catch (SecurityException se) {}
>                         return ctor;
>                     }
>                 }
>             }
>         }
> ```
> 
> You can see that they do very simplistic matching. Very similar to what we used to do. In fast, I tried this library first when I noticed the bug, but it had the same outcome.
> 
> Also, there is nothing in the Java specification about "Choosing the Most Specific Constructor" , so I assume JLS 15.12.2.5 refers constructor matching as well.

I see.

It seems to be passing the test case though: 
https://gist.github.com/Maithem/cef8d9c63b7504d8dfa8183a7e6f5738

vjeko:
You just got lucky. The ordering of the constructors seems to be undefined. For example, if I reverse constructors in one of the classes, such as:

```
    public static class ExampleInterface {
        public ExampleInterface(BaseInterface base) {
            throw new IllegalStateException("Not suppose to be called");
        }
        
        public ExampleInterface(ChildInterface base) {
            // NOOP.
        }
    }
```

This causes the test to fail:
error

runtime/src/main/java/org/corfudb/util/ReflectionUtils.java
```diff
@@ -92,38 +100,94 @@ public static String getMethodNameOnlyFromString(String s) {
...
+            @NonNull Constructor[] constructors, @NonNull Object[] args)
+            throws IllegalAccessException, InvocationTargetException, InstantiationException {
+        // Figure out the argument types.
+        final List<Class> argTypes = Arrays
+                .stream(args).map(Object::getClass)
+                .collect(Collectors.toList());
+        // Filter out the constructors that do not have the same arity.
+        final List<Constructor> constructorCandidates = Arrays.stream(constructors)
+                .filter(constructor -> constructor.getParameterTypes().length == args.length)
+                .collect(Collectors.toList());
```
Maithem: Btw this breaks support for non-static nested classes, since `getParameterTypes()` will return the enclosing instance as the first parameter, so all the constructors will be filtered out because the sizes won't match. 

From the javadoc:

> The parameterTypes parameter is an array of Class objects that identify the constructor's formal parameter types, in declared order. If this Class object represents an inner class declared in a non-static context, the formal parameter types include the explicit enclosing instance as the first parameter.

Actually, if you remove static modifier from the test types in `ObjectBuilderTest.java` the test will fail.

vjeko: Good point. This is a pretty minor use-case, which we do not have in Corfu right now. Maybe we can address this in a subsequent patch?

The plan was never to implement the specification fully, but rather improve on the previous approach.

PR #2177, opened 2019-10-02
Catch the interrupted exception and retry a few times for cache read op.

xiaoqin2012:
#1885  Overview
Catch the interrupted exception in Corfu Code.

Description:
It has happened a few times in QA's tests for upgrade. The interruption is propagated to CorfuPoller and Dao layer. The stack traces show it all happened for the AddressView.read() operation.

Why should this be merged: 
The corfu layer should retry the operation while the underlying caching layer throw an interrupted exception.

Bug: 2430581

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -29,6 +29,7 @@ import org.corfudb.runtime.exceptions.TrimmedException;
 import org.corfudb.runtime.exceptions.WriteSizeException;
 import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.util.CFUtils;
 import org.corfudb.util.CorfuComponent;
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -275,14 +278,35 @@ public void write(IToken token, Object data) throws OverwriteException {
...
      * @param address An address to read from.
      * @param options Read options for this particular write (i.e. configure caching behavior)
      * @return A result, which be cached.
      */
     public @Nonnull
     ILogData read(long address, @NonNull ReadOptions options) {
+        int count = 0;
+        while (true) {
+            try {
+                return readNative (address, options);
```
lixinchengdu: The name of readNative is misleading in java context. I saw a lot of XXXInner at corfu code.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -275,14 +278,35 @@ public void write(IToken token, Object data) throws OverwriteException {
...
+            } catch (InterruptedException e) {
+                if (++count == MAX_RETRIES) {
+                    log.error ("It has retried {} times, throw an exception {}", MAX_RETRIES, e);
```
lixinchengdu: The last two braces should be deleted otherwise e.toString() would fill the replace holder.

PR #2182, opened 2019-10-08
Schemaless Browser Tool

pankti-m:
Schemaless Browser Tool using Dynamic Messages.

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/corfudb-tools/README
+++ b/corfudb-tools/README
@@ -1,0 +1,37 @@
+CorfuStore Browser Editor Tool
+
+The Browser and Editor tools are used for inspecting and troubleshooting data
+inside Corfu.
+
+The tool is schemaless, i.e., it does not need to be packaged with protobuf
+definitions or proto files to serialize/deserialize the data.
+
+--------
+Browser
+--------
+The browser supports read operation which reads and outputs all entries in a
+given table and namespace.  It can be invoked like the below example-
+
+java -cp "corfudb-tools-0.3.0-SNAPSHOT-shaded.jar" --Dlogback.configurationFile=logback.prod.xml
+org.corfudb.browser.CorfuBrowserMain
+--host=10.160.29.112
+--port=9000
+--namespace=sample_namespace
+--tablename=sample_tablename
+--tlsEnabled=true
+--keystore=/config/cluster-manager/corfu/private/keystore.jks
+--ks_password=/config/cluster-manager/corfu/private/keystore.password
+--truststore=/config/cluster-manager/corfu/public/truststore.jks
+--truststore_password=/config/cluster-manager/corfu/public/truststore.password
+
+Keystore and truststore parameters are not required if tls is disabled.
+
+For usage help,
+java -cp "corfuBrowser/corfudb-tools-0.3.0-SNAPSHOT-shaded.jar" org.corfudb.browser.CorfuTableBrowserMain <--help|-h>
+
+TODO(pmajmudar): Add a user script to invoke the browser.
+
+-------
+Editor
+-------
+Coming Soon
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -1,0 +1,75 @@
+package org.corfudb.browser;
+
+import com.google.common.reflect.TypeToken;
+
+import java.util.Map;
+
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.CorfuTable;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.CorfuDynamicRecord;
+import org.corfudb.runtime.view.TableRegistry;
+import org.corfudb.util.serializer.DynamicProtobufSerializer;
+import org.corfudb.util.serializer.ISerializer;
+import org.corfudb.util.serializer.Serializers;
+
+/**
+ * This is the CorfuStore Browser Tool which prints data in a given namespace and table.
+ *
+ * - Created by pmajmudar on 10/16/2019.
+ */
+@Slf4j
+public class CorfuStoreBrowser {
+    private final CorfuRuntime runtime;
+
+    /**
+     * Creates a CorfuBrowser which connects a runtime to the server.
+     * @param runtime CorfuRuntime which has connected to the server
+     */
+    public CorfuStoreBrowser(CorfuRuntime runtime) {
+        this.runtime = runtime;
+    }
+
+    /**
+     * Fetches the table from the given namespace
+     * @param namespace Namespace of the table
+     * @param tableName Tablename
+     * @return CorfuTable
+     */
+    public CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> getTable(
+        String namespace, String tableName) {
+        log.info("Namespace: {}", namespace);
+        log.info("TableName: {}", tableName);
+        ISerializer dynamicProtobufSerializer =
+            new DynamicProtobufSerializer(runtime);
+        Serializers.registerSerializer(dynamicProtobufSerializer);
+
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> corfuTable =
+            runtime.getObjectsView().build()
+            .setTypeToken(new TypeToken<CorfuTable<CorfuDynamicKey, CorfuDynamicRecord>>() {
+            })
+            .setStreamName(
+                TableRegistry.getFullyQualifiedTableName(namespace, tableName))
+            .setSerializer(dynamicProtobufSerializer)
+            .open();
+        return corfuTable;
+    }
+
+    /**
+     * Prints the payload and metadata in the given table
+     * @param table
+     */
+    public void printTable(
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table) {
+        StringBuilder builder;
+        for (Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry :
+            table.entrySet()) {
+            builder = new StringBuilder("\nKey:\n" + entry.getKey().getKey())
+                .append("\nPayload:\n" + entry.getValue().getPayload())
+                .append("\nMetadata:\n" + entry.getValue().getMetadata())
+                .append("\n====================\n");
+            log.info(builder.toString());
+        }
+    }
+}
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
@@ -1,0 +1,85 @@
+package org.corfudb.browser;
+
+import java.util.Map;
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.CorfuDynamicRecord;
+import org.corfudb.runtime.collections.CorfuTable;
+import org.corfudb.util.GitRepositoryState;
+import org.docopt.Docopt;
+
+/**
+ * Main class for the CorfuStore Browser Tool.
+ * Command line options are documented in the USAGE variable.
+ *
+ * - Created by pmajmudar on 10/16/2019
+ */
+@Slf4j
+public class CorfuStoreBrowserMain {
+    private static final String USAGE = "Usage: corfu-browser --host=<host> " +
+        "--port=<port> --namespace=<namespace> --tablename=<tablename> " +
+        "[--keystore=<keystore_file>] [--ks_password=<keystore_password>] " +
+        "[--truststore=<truststore_file>] [--truststore_password=<truststore_password>] " +
+        "[--tlsEnabled=<tls_enabled>]\n"
+        + "Options:\n"
+        + "--host=<host>   Hostname\n"
+        + "--port=<port>   Port\n"
+        + "--namespace=<namespace>   Namespace\n"
+        + "--tablename=<tablename>   Table Name\n"
+        + "--keystore=<keystore_file> KeyStore File\n"
+        + "--ks_password=<keystore_password> KeyStore Password\n"
+        + "--truststore=<truststore_file> TrustStore File\n"
+        + "--truststore_password=<truststore_password> Truststore Password\n"
+        + "--tlsEnabled=<tls_enabled>";
+
+    public static void main(String[] args) {
+        try {
+            CorfuRuntime runtime;
+            // Parse the options given, using docopt.
+            Map<String, Object> opts =
+                new Docopt(USAGE)
+                    .withVersion(GitRepositoryState.getRepositoryState().describe)
+
+                    .parse(args);
+            String host = opts.get("--host").toString();
+            Integer port = Integer.parseInt(opts.get("--port").toString());
+            boolean tlsEnabled = Boolean.parseBoolean(opts.get("--tlsEnabled")
+                .toString());
+            CorfuRuntime.CorfuRuntimeParameters.CorfuRuntimeParametersBuilder
+                builder = CorfuRuntime.CorfuRuntimeParameters.builder()
+                .useFastLoader(false)
+                .cacheDisabled(true)
+                .tlsEnabled(tlsEnabled);
+            if (tlsEnabled) {
+                String keystore = opts.get("--keystore").toString();
+                String ks_password = opts.get("--ks_password").toString();
+                String truststore = opts.get("--truststore").toString();
+                String truststore_password =
+                    opts.get("--truststore_password").toString();
+                builder.tlsEnabled(tlsEnabled)
+                    .keyStore(keystore)
+                    .ksPasswordFile(ks_password)
+                    .trustStore(truststore)
+                    .tsPasswordFile(truststore_password);
+            }
+            runtime = CorfuRuntime.fromParameters(builder.build());
+            String singleNodeEndpoint = String.format("%s:%d", host, port);
+            runtime.parseConfigurationString(singleNodeEndpoint);
+            runtime.connect();
+            log.info("Successfully connected to {}", singleNodeEndpoint);
+            CorfuStoreBrowser browser = new CorfuStoreBrowser(runtime);
+            CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table =
+                browser.getTable(
+                    opts.get("--namespace").toString(),
+                    opts.get("--tablename").toString()
+                );
+            browser.printTable(table);
+        } catch (Throwable t) {
+            log.error("Error in Browser Execution.", t);
+            throw t;
+        }
+    }
+}
--- a/samples/src/main/java/org/corfudb/samples/CorfuStoreSample.java
+++ b/samples/src/main/java/org/corfudb/samples/CorfuStoreSample.java
@@ -35,9 +35,12 @@ public class CorfuStoreSample {
     private static CorfuRuntime getRuntimeAndConnect(String configurationString) {
 
         return CorfuRuntime.fromParameters(CorfuRuntime.CorfuRuntimeParameters.builder()
-                .build())
-                .parseConfigurationString(configurationString)
-                .connect();
+            .useFastLoader(false)
+            .cacheDisabled(true)
+            .tlsEnabled(false)
+            .build())
+            .parseConfigurationString(configurationString)
+            .connect();
     }
 
     @SuppressWarnings("checkstyle:printLine") // Sample code
@@ -72,6 +75,10 @@ public class CorfuStoreSample {
                         Name.newBuilder().setFirstName("a").setLastName("x").build(),
                         Car.newBuilder().setColor("red").build(),
                         null)
+                .create(tableName,
+                    Name.newBuilder().setFirstName("b").setLastName("y").build(),
+                    Car.newBuilder().setColor("blue").build(),
+                    null)
                 // Transaction is begun and ended here.
                 .commit();
 
--- a/test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
+++ b/test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
@@ -1,0 +1,190 @@
+package org.corfudb.integration;
+
+import com.google.protobuf.UnknownFieldSet;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import org.corfudb.browser.CorfuStoreBrowser;
+import org.corfudb.runtime.collections.CorfuStore;
+import org.corfudb.runtime.collections.CorfuTable;
+import org.corfudb.runtime.collections.CorfuDynamicRecord;
+import org.corfudb.runtime.collections.TableOptions;
+import org.corfudb.runtime.collections.TxBuilder;
+import org.corfudb.test.SampleAppliance;
+import org.corfudb.test.SampleSchema;
+
+
+public class CorfuStoreBrowserIT extends AbstractIT {
+
+    private static String corfuSingleNodeHost;
+
+    private static int corfuStringNodePort;
+
+    private static String singleNodeEndpoint;
+
+    /* A helper method that takes host and port specification, start a single server and
+     *  returns a process. */
+    private Process runSinglePersistentServer(String host, int port) throws
+        IOException {
+        return new AbstractIT.CorfuServerRunner()
+            .setHost(host)
+            .setPort(port)
+            .setLogPath(getCorfuServerLogPath(host, port))
+            .setSingle(true)
+            .runServer();
+    }
+
+    /**
+     * Load properties for a single node corfu server before each test
+     */
+    @Before
+    public void loadProperties() {
+        corfuSingleNodeHost = PROPERTIES.getProperty("corfuSingleNodeHost");
+        corfuStringNodePort = Integer.valueOf(PROPERTIES.getProperty(
+            "corfuSingleNodePort"));
+        singleNodeEndpoint = String.format(
+            "%s:%d",
+            corfuSingleNodeHost,
+            corfuStringNodePort
+        );
+    }
+
+    /**
+     * Create a table and add data to it.  Verify that the browser tool is able
+     * to read its contents accurately.
+     * @throws IOException
+     * @throws NoSuchMethodException
+     * @throws IllegalAccessException
+     * @throws InvocationTargetException
+     */
+    @Test
+    public void browserTest() throws
+        IOException,
+        NoSuchMethodException,
+        IllegalAccessException,
+        InvocationTargetException {
+        final String namespace = "namespace";
+        final String tableName = "table";
+        Process corfuServer = runSinglePersistentServer(corfuSingleNodeHost,
+            corfuStringNodePort);
+
+        // Start a Corfu runtime
+        runtime = createRuntime(singleNodeEndpoint);
+
+        CorfuStore store = new CorfuStore(runtime);
+
+        store.openTable(
+            namespace,
+            tableName,
+            SampleSchema.Uuid.class,
+            SampleSchema.Uuid.class,
+            SampleSchema.Uuid.class,
+            TableOptions.builder().build());
+
+        final long keyUuid = 1L;
+        final long valueUuid = 3L;
+        final long metadataUuid = 5L;
+
+        SampleSchema.Uuid uuidKey = SampleSchema.Uuid.newBuilder()
+            .setMsb(keyUuid)
+            .setLsb(keyUuid)
+            .build();
+        SampleSchema.Uuid uuidVal = SampleSchema.Uuid.newBuilder()
+            .setMsb(valueUuid)
+            .setLsb(valueUuid)
+            .build();
+        SampleSchema.Uuid metadata = SampleSchema.Uuid.newBuilder()
+            .setMsb(metadataUuid)
+            .setLsb(metadataUuid)
+            .build();
+        TxBuilder tx = store.tx(namespace);
+        tx.create(tableName, uuidKey, uuidVal, metadata)
+            .update(tableName, uuidKey, uuidVal, metadata)
+            .commit();
+        runtime.shutdown();
+
+        // Invoke the browser and go through each item
+        runtime = createRuntime(singleNodeEndpoint);
+        CorfuStoreBrowser browser = new CorfuStoreBrowser(runtime);
+        CorfuTable table = browser.getTable(namespace, tableName);
+        browser.printTable(table);
+        Assert.assertEquals(1, table.size());
+        for(Object obj : table.values()) {
+            CorfuDynamicRecord record = (CorfuDynamicRecord)obj;
+            Assert.assertEquals(
+                UnknownFieldSet.newBuilder().build(),
+                record.getPayload().getUnknownFields());
+        }
+    }
+
+    /**
+     * Create a table and add nested protobufs as data to it.  Verify that the
+     * browser tool is able to read the contents accurately.
+     * @throws IOException
+     * @throws NoSuchMethodException
+     * @throws IllegalAccessException
+     * @throws InvocationTargetException
+     */
+    @Test
+    public void nestedProtoTest() throws
+        IOException,
+        NoSuchMethodException,
+        IllegalAccessException,
+        InvocationTargetException {
+        final String namespace = "namespace";
+        final String tableName = "table";
+        runSinglePersistentServer(corfuSingleNodeHost, corfuStringNodePort);
+
+        // Start a Corfu runtime
+        runtime = createRuntime(singleNodeEndpoint);
+
+        CorfuStore store = new CorfuStore(runtime);
+
+        final long keyUuid = 10L;
+        final long ruleIdVal = 50L;
+        final long metaUuid = 100L;
+
+        store.openTable(
+            namespace,
+            tableName,
+            SampleSchema.Uuid.class,
+            SampleSchema.FirewallRule.class,
+            SampleSchema.Uuid.class,
+            TableOptions.builder().build());
+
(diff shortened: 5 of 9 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/DynamicCorfuKey.java
```diff
@@ -0,0 +1,36 @@
...
+        this.payload = new CorfuDynamicMessage(payload);
+    }
+
+    public DynamicMessage getPayload() {
+        return payload.getPayload();
+    }
+
+    @Override
+    public int hashCode() {
+        return payload.hashCode();
```
pankti-m: shouldnt the type url be factored in to calculate the hashcode?

runtime/src/main/java/org/corfudb/runtime/collections/DynamicCorfuKey.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        DynamicCorfuKey dynamicCorfuKey = (DynamicCorfuKey) obj;
```
pankti-m: same here.  Shouldnt the type url be factored in to determine equality?

runtime/src/main/java/org/corfudb/runtime/collections/DynamicCorfuKey.java
```diff
@@ -0,0 +1,36 @@
...
+        return payload.equals(dynamicCorfuKey.payload);
```
pankti-m: same here. Shouldnt the type url be factored in to determine equality?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    @Override
+    public byte getType() {
+        return type;
+    }
+
+    FileDescriptor getDescriptors(String name) throws DescriptorValidationException {
+
```
pankti-m: nit - extra newline

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,59 @@
...
+import org.corfudb.runtime.collections.DynamicCorfuKey;
+import org.corfudb.runtime.collections.DynamicCorfuRecord;
+import org.corfudb.util.serializer.DynamicProtobufSerializer;
+import org.corfudb.util.serializer.ISerializer;
+import org.corfudb.util.serializer.Serializers;
+
+@Slf4j
+public class CorfuBrowser {
+    private CorfuRuntime runtime;
+    String singleNodeEndpoint;
```
zalokhan: private final

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public CorfuBrowser(String host, int port) {
+        singleNodeEndpoint = String.format("%s:%d", host, port);
+        runtime = createRuntime(singleNodeEndpoint);
+    }
+
+    private CorfuRuntime createRuntime(String endpoint) {
+        runtime = new CorfuRuntime(endpoint).setCacheDisabled(true).connect();
```
zalokhan: This is a deprecated constructor. Please use CorfuRuntimeParameters.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,59 @@
...
+        return runtime;
+    }
+
+    public CorfuTable getTable(String namespace, String tableName) {
+        final int code = 25;
```
zalokhan: This number will get difficult to keep track of. We should keep this at some common location.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,59 @@
...
+        ISerializer dynamicProtobufSerializer =
+            new DynamicProtobufSerializer((byte) code, runtime);
+        Serializers.registerSerializer(dynamicProtobufSerializer);
+
+        CorfuTable<DynamicCorfuKey, DynamicCorfuRecord> corfuTable =
+            runtime.getObjectsView().build()
+            .setTypeToken(new TypeToken<CorfuTable<DynamicCorfuKey, DynamicCorfuRecord>>() {
+            })
+            .setStreamName(namespace + "$" + tableName)
```
zalokhan: Use `TableRegistry.getFullyQualifiedTableName`

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public void printTable(CorfuTable table) {
+        for (Object obj : table.values()) {
+            DynamicCorfuRecord record = (DynamicCorfuRecord)obj;
+
+            log.info("Payload: {}", record.getPayload());
+            log.info("Metadata: {}", record.getMetadata());
+            log.info("===================");
+
+            System.out.println("Payload: " + record.getPayload());
```
zalokhan: Remove System outs.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.corfudb.util.GitRepositoryState;
+import org.docopt.Docopt;
+
+public class CorfuBrowserMain {
+    private static final String USAGE = "Usage: corfu-browser --host=<host> " +
+        "--port=<port> --namespace=<namespace> --tablename=<tablename>\n"
+        + "Options:\n"
+        + "--host=<host>   Hostname\n"
+        + "--port=<port>   Port\n";
```
zalokhan: Add options for namespace and table name as well

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: Done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: done.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: Yes.  One way is to keep a configuration file with all constant parameters used by corfu.  Right now I have created a constant in this file itself.  I can work on this cleanup in parallel.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,73 @@
...
+import org.corfudb.runtime.collections.CorfuTable;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.CorfuDynamicRecord;
+import org.corfudb.runtime.view.TableRegistry;
+import org.corfudb.util.serializer.DynamicProtobufSerializer;
+import org.corfudb.util.serializer.ISerializer;
+import org.corfudb.util.serializer.Serializers;
+
+@Slf4j
+public class CorfuBrowser {
```
zalokhan: Add Class level comments.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,73 @@
...
+        connectRuntime(singleNodeEndpoint);
+    }
+
+    private void connectRuntime(String endpoint) {
+        runtime.parseConfigurationString(endpoint);
+        runtime.connect();
+        log.info("Successfully connected to {}", endpoint);
+    }
+
+    public CorfuTable getTable(String namespace, String tableName) {
```
zalokhan: Add comments on all methods.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,73 @@
...
+            .setTypeToken(new TypeToken<CorfuTable<CorfuDynamicKey, CorfuDynamicRecord>>() {
+            })
+            .setStreamName(
+                TableRegistry.getFullyQualifiedTableName(namespace, tableName))
+            .setSerializer(dynamicProtobufSerializer)
+            .open();
+        return corfuTable;
+    }
+
+    @SuppressWarnings("checkstyle:printLine")
```
zalokhan: Since this is only for testing. Remember to remove this before checking in.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Map;
+
+import org.corfudb.util.GitRepositoryState;
+import org.docopt.Docopt;
+
+public class CorfuBrowserMain {
+    private static final String USAGE = "Usage: corfu-browser --host=<host> " +
+        "--port=<port> --namespace=<namespace> --tablename=<tablename> " +
+        "--keystore=<keystore_file> --ks_password=<keystore_password> " +
```
zalokhan: We can make these fields optional.
This will be helpful when we want to browse data on our local setup which might not have the tls requirements.
For optional parameters you can take a look at CorfuServer Docopt. (You need to enclose optional parameters in '[, ]')

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
```diff
@@ -0,0 +1,48 @@
...
+        + "--port=<port>   Port\n"
+        + "--namespace=<namespace>   Namespace\n"
+        + "--tablename=<tablename>   Table Name\n"
+        + "--keystore=<keystore_file> KeyStore File\n"
+        + "--ks_password=<keystore_password> KeyStore Password\n"
+        + "--truststore=<truststore_file> TrustStore File\n"
+        + "--truststore_password=<truststore_password> Truststore Password\n"
+        + "--tlsEnabled=<tls_enabled>";
+
+    @SuppressWarnings("checkstyle:printLine")
```
zalokhan: Remove this after finishing testing.

test/src/test/java/org/corfudb/integration/CorfuBrowserIT.java
```diff
@@ -0,0 +1,174 @@
...
+            "corfuSingleNodePort"));
+        singleNodeEndpoint = String.format(
+            "%s:%d",
+            corfuSingleNodeHost,
+            corfuStringNodePort
+        );
+    }
+
+    @Test
+    public void browserTest() throws
```
zalokhan: Good tests.
Please add high-level comments describing the test.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
pankti-m: done

test/src/test/java/org/corfudb/integration/CorfuBrowserIT.java
pankti-m: done

Maithem:
Can you rebase? there's a bunch of dependencies that were already merged.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+import org.corfudb.runtime.collections.CorfuTable;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.CorfuDynamicRecord;
+import org.corfudb.runtime.view.TableRegistry;
+import org.corfudb.util.serializer.DynamicProtobufSerializer;
+import org.corfudb.util.serializer.ISerializer;
+import org.corfudb.util.serializer.Serializers;
+
+/**
+ * This is the UFO Browser Tool which prints data in a given namespace and table.
```
Maithem: Just to keep the terminology consistent. Maybe use CorfuTable instead ?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * - Created by pmajmudar on 10/16/2019.
+ */
+@Slf4j
+public class CorfuBrowser {
```
Maithem: CorfuTableBrowser

pankti-m:
Rebased.

~ Pankti

corfudb-tools/README
```diff
@@ -0,0 +1,33 @@
...
+
+--------
+Browser
+--------
+The browser supports read operation which reads and outputs all entries in a
+given table and namespace.
+We plan to add a user script to invoke the browser but as of now, it can be used
+with the following command-
+
+java -cp <corfu-tools jar>
```
Maithem: I think the example should be invoked with actual arguments:
`java -cp corfudb-tools-0.3.0-SNAPSHOT-shaded.jar org.corfudb.browser.CorfuBrowserMain`

Since docopt generates the usage with --help, I don't think we also need to document them here

corfudb-tools/src/main/resources/logback.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Maithem: Why do you need both prod and this?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+     * Creates a CorfuBrowser which connects a runtime to the server.
+     * @param host Corfu Server Host
+     * @param port Corfu Server Port
+     * @param keyStore Location of the KeyStore File
+     * @param ksPasswordFile Location of the KeyStore Password File
+     * @param trustStore Location of the Truststore File
+     * @param tsPasswordFile Location of the Truststore Password File
+     * @param tlsEnabled Tls Enabled Status
+     */
+    public CorfuBrowser(String host, int port, String keyStore,
```
Maithem: An endpoint is actually a cluster, not a single address. I suggest you just pass a comma separated connection string. For example, ip1:port1,ip2:port2

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Setup logback logger
+     * - pick the correct logging level before outputting error messages
+     * - add serverEndpoint information
+     *
+     * @throws JoranException logback exception
+     */
+    private static void configureLogger() {
```
Maithem: Why is this needed, aren't you setting this from the *xml files?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
```diff
@@ -0,0 +1,58 @@
...
+import org.docopt.Docopt;
+
+/**
+ * Main class for the UFO Browser Tool.
+ * Command line options are documented in the USAGE variable.
+ *
+ * - Created by pmajmudar on 10/16/2019
+ */
+public class CorfuBrowserMain {
+    private static final String USAGE = "Usage: corfu-browser --host=<host> " +
```
Maithem: Btw, if you wan't an alternative to docopt, you can consider jcommander. I feel like its more cleaner way to specify command args.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
```diff
@@ -0,0 +1,58 @@
...
+        String keystore = opts.get("--keystore") == null ? "" :
+            opts.get("--keystore").toString();
+        String ks_password = opts.get("--ks_password") == null? "" :
+            opts.get("--ks_password").toString();
+        String truststore = opts.get("--truststore") == null ? "" :
+            opts.get("--truststore").toString();
+        String truststore_password =
+            opts.get("--truststore_password") == null? "" :
+                opts.get("--truststore_password").toString();
+        boolean tlsEnabled = opts.get("--tlsEnabled") == null ? false :
```
Maithem: This might be superfluous, since you can actually pass nulls to the constructor.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
```diff
@@ -0,0 +1,58 @@
...
+            Boolean.parseBoolean(opts.get("--tlsEnabled").toString());
+
+        CorfuBrowser browser = new CorfuBrowser(host, port, keystore,
+            ks_password, truststore, truststore_password, tlsEnabled);
+        CorfuTable table = browser.getTable(opts.get("--namespace").toString(),
+            opts.get("--tablename").toString());
+        browser.printTable(table);
```
Maithem: This seems to be crude. Why not have a REPL and support the following options:
1. Get key
2. print all keys
3. print the whole table

Maithem: If the table is too large and we're interested in only a couple of keys, its too much work to first dump the table in a separate file and then grep. It makes sense to support this operation directly.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Prints the payload and metadata in the given table
+     * @param table
+     */
+    public void printTable(CorfuTable table) {
+        for (Object obj : table.values()) {
+            CorfuDynamicRecord record = (CorfuDynamicRecord)obj;
+            log.info("Payload: {}", record.getPayload());
```
Maithem: It will be nice to print this in table format using  width and precision specifiers
==========
Name: table name  Namespace: namespace
==========
Key   Value
k1      v1
k2      v2

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: done.  replaced other references of UFO with CorfuTable.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: done

corfudb-tools/README
```diff
@@ -0,0 +1,33 @@
...
+
+The tool is schemaless, i.e., it does not need to be packaged with protobuf
+definitions or proto files to serialize/deserialize the data.
+
+--------
+Browser
+--------
+The browser supports read operation which reads and outputs all entries in a
+given table and namespace.
+We plan to add a user script to invoke the browser but as of now, it can be used
```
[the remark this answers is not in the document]
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
pankti-m: We decided to use docopt because it is being used elsewhere in the repo.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
pankti-m: I was not sure of the behavior of building CorfuRuntime parameters with null arguments.

If it does not support nulls, will have to add the same checks later.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
pankti-m: for now we decided to start with a simple operation of just reading all the contents.  We plan to add more filtering capabilities later.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
WenbinZhu: Yea passing empty string to runtime parameter looks incorrect. You can pass null to it, and let it fail with NPE in case some parameter is required but not provided if tls is enabled. And if tls is not enabled, these parameters should be ignored.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
```diff
@@ -0,0 +1,58 @@
...
+            opts.get("--truststore").toString();
+        String truststore_password =
+            opts.get("--truststore_password") == null? "" :
+                opts.get("--truststore_password").toString();
+        boolean tlsEnabled = opts.get("--tlsEnabled") == null ? false :
+            Boolean.parseBoolean(opts.get("--tlsEnabled").toString());
+
+        CorfuBrowser browser = new CorfuBrowser(host, port, keystore,
+            ks_password, truststore, truststore_password, tlsEnabled);
+        CorfuTable table = browser.getTable(opts.get("--namespace").toString(),
```
WenbinZhu: Does this mean a single invocation of this tool can only dump one table? If I want to dump multiple tables, I need to re-invoke the tool, create a new runtime and connect?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+        return corfuTable;
+    }
+
+    /**
+     * Prints the payload and metadata in the given table
+     * @param table
+     */
+    public void printTable(CorfuTable table) {
+        for (Object obj : table.values()) {
+            CorfuDynamicRecord record = (CorfuDynamicRecord)obj;
```
WenbinZhu: Why we are only printing the record and metadata, not the keys?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: I based this on the current implementation of the browser tool in the NSX repo.  Will address this in a subsequent patch.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
pankti-m: done

corfudb-tools/README
```diff
@@ -0,0 +1,33 @@
+UFO Browser Editor Tool
```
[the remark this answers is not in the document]
pankti-m: I am changing it to CorfuStoreBrowser

corfudb-tools/README
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowserMain.java
pankti-m: Yes a single invocation only prints contents of a single table.  Every invocation creates a new runtime and connects using it.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: fixed

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+        runtime.connect();
+        log.info("Successfully connected to {}", endpoint);
+    }
+
+    /**
+     * Setup logback logger
+     * - pick the correct logging level before outputting error messages
+     * - add serverEndpoint information
+     *
+     * @throws JoranException logback exception
```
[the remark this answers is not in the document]
pankti-m: done

corfudb-tools/src/main/resources/logback.xml
pankti-m: changed to only use prod

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: I tried this using string.format but to align the keys/values/metadata correctly in their respective columns, there is no certain width which is suitable for all types of data.  Without alignment, it is not readable.

So I am printing it in the form-
Key:
Value:
Meta:
<Row Separator>

Let me know if you want to see a sample output.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
```diff
@@ -0,0 +1,110 @@
...
+            .setSerializer(dynamicProtobufSerializer)
+            .open();
+        return corfuTable;
+    }
+
+    /**
+     * Prints the payload and metadata in the given table
+     * @param table
+     */
+    public void printTable(CorfuTable table) {
```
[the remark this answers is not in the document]
pankti-m: Added.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuBrowser.java
pankti-m: Added keys as well

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -0,0 +1,80 @@
...
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table) {
+        StringBuilder builder;
+        for (Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry :
+            table.entrySet()) {
+            builder = new StringBuilder("\nKey:\n" + entry.getKey().getKey())
+                .append("\nPayload:\n" + entry.getValue().getPayload())
+                .append("\nMetadata:\n" + entry.getValue().getMetadata());
+                //.append("\n====================\n");
+            log.info(builder.toString());
+            /*log.info("\nKey: {}", entry.getKey().getKey());
```
zalokhan: You can remove these comments.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public void printTable(
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table) {
+        StringBuilder builder;
+        for (Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry :
+            table.entrySet()) {
+            builder = new StringBuilder("\nKey:\n" + entry.getKey().getKey())
+                .append("\nPayload:\n" + entry.getValue().getPayload())
+                .append("\nMetadata:\n" + entry.getValue().getMetadata());
+                //.append("\n====================\n");
```
zalokhan: I think you want to uncomment this right?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -0,0 +1,80 @@
...
+import org.corfudb.util.serializer.ISerializer;
+import org.corfudb.util.serializer.Serializers;
+
+/**
+ * This is the CorfuTable Browser Tool which prints data in a given namespace and table.
+ *
+ * - Created by pmajmudar on 10/16/2019.
+ */
+@Slf4j
+public class CorfuStoreBrowser {
```
zalokhan: The comment says CorfuTableBrowser and class name is CorfuStoreBrowser.
Fix either of them.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.CorfuDynamicRecord;
+import org.corfudb.runtime.collections.CorfuTable;
+import org.corfudb.util.GitRepositoryState;
+import org.docopt.Docopt;
+
+/**
+ * Main class for the CorfuTable Browser Tool.
```
zalokhan: Fix this to CorfuStore

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -0,0 +1,81 @@
...
+        + "--namespace=<namespace>   Namespace\n"
+        + "--tablename=<tablename>   Table Name\n"
+        + "--keystore=<keystore_file> KeyStore File\n"
+        + "--ks_password=<keystore_password> KeyStore Password\n"
+        + "--truststore=<truststore_file> TrustStore File\n"
+        + "--truststore_password=<truststore_password> Truststore Password\n"
+        + "--tlsEnabled=<tls_enabled>";
+
+    public static void main(String[] args) {
+        CorfuRuntime runtime;
```
zalokhan: Should we have a wrapping try-catch block?
What happens if there is an exception? Does it log or fail silently?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -0,0 +1,81 @@
...
+        boolean tlsEnabled = Boolean.parseBoolean(opts.get("--tlsEnabled").toString());
+
+        if (tlsEnabled) {
+            String keystore = opts.get("--keystore").toString();
+            String ks_password = opts.get("--ks_password").toString();
+            String truststore = opts.get("--truststore").toString();
+            String truststore_password =
+                opts.get("--truststore_password").toString();
+            runtime = CorfuRuntime.fromParameters(
+                CorfuRuntime.CorfuRuntimeParameters.builder().useFastLoader(false)
```
zalokhan: Start the Parameters builder outside the if loop and build() it after the else to initialize the runtime.
You can then set duplicate options like fast loader and cache disabled only once above.

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
```diff
@@ -0,0 +1,190 @@
...
+            .setLsb(metadataUuid)
+            .build();
+        TxBuilder tx = store.tx(namespace);
+        tx.create(tableName, uuidKey, uuidVal, metadata)
+            .update(tableName, uuidKey, uuidVal, metadata)
+            .commit();
+        runtime.shutdown();
+
+        // Invoke the browser and go through each item
+        runtime.connect();
```
zalokhan: Connecting an already shutdown runtime is not a good idea. Probably we should add a check to avoid that (Not in this PR).
Just create a new runtime.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: yes

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: changed comment to 'CorfuStoreBrowser'

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
pankti-m: runtime.connect() can throw an error and it gets logged.  Otherwise I did not find anything throwing an exception.  Maybe we can wrap this in a generic exception?

samples/src/main/java/org/corfudb/samples/CorfuStoreSample.java
```diff
@@ -35,9 +35,16 @@
...
-                .build())
-                .parseConfigurationString(configurationString)
-                .connect();
+            .useFastLoader(false)
+            .cacheDisabled(true)
+            .tlsEnabled(true)
+            .keyStore("/config/cluster-manager/corfu/private/keystore.jks")
+            .ksPasswordFile("/config/cluster-manager/corfu/private/keystore.password")
+            .trustStore("/config/cluster-manager/corfu/public/truststore.jks")
+            .tsPasswordFile("/config/cluster-manager/corfu/public/truststore.password")
```
WenbinZhu: Why the files are using these paths instead of a local file in the resource folder?

pankti-m: Reverting to not using these parameters.

Doing so will not allow sample app to use tls but that is independent of browser changes.  So can be addressed in a different PR.

PR #2183, opened 2019-10-15
Setup some corfu threads as high priority threads

xiaoqin2012:
## Overview

Description:
Setup node failure detector threads and threads responsible for ping as high priority threads.

Why should this be merged: 
While system is heavy loaded, we see a lot of layout changes due to timeouts. 
Improve threads priority to reduce the ping delay.

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
@@ -166,6 +166,7 @@ public class RemoteMonitoringService implements MonitoringService {
         this.detectionTasksScheduler = Executors.newSingleThreadScheduledExecutor(
                 new ThreadFactoryBuilder()
                         .setDaemon(true)
+                        .setPriority (Thread.MAX_PRIORITY)
                         .setNameFormat(serverContext.getThreadPrefix() + "ManagementService")
                         .build());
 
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -710,6 +710,7 @@ private EventLoopGroup getNewEventLoopGroup() {
                 parameters.nettyEventLoopThreads;
         ThreadFactory factory = new ThreadFactoryBuilder()
                 .setDaemon(true)
+                .setPriority (Thread.MAX_PRIORITY)
```
vjeko: Extra space.

infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -166,6 +166,7 @@ public void increment() {
         this.detectionTasksScheduler = Executors.newSingleThreadScheduledExecutor(
                 new ThreadFactoryBuilder()
                         .setDaemon(true)
+                        .setPriority (Thread.MAX_PRIORITY)
```
vjeko: Extra space "y (".

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
zalokhan: Won't this lead to High priority threads even in the application as well?
Apart from that, state-transfer also uses CorfuRuntime to transfer data. If that uses high priority as well, we might end up in a worse situation where sequencer, log unit and layout threads get starved.

PR #2184, opened 2019-10-15
Lazy Deserialization

zalokhan:
## Overview

Description:

Why should this be merged: 

Referenceing #2079 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
@@ -2,19 +2,18 @@ package org.corfudb.protocols.logprotocol;
 
 import io.netty.buffer.ByteBuf;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
-import lombok.Getter;
+import io.netty.buffer.Unpooled;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
-import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.util.serializer.Serializers;
-
 
 
 /**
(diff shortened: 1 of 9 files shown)


runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -123,8 +133,12 @@ public void serialize(ByteBuf b) {
     @Override
     public void setGlobalAddress(long address) {
         super.setGlobalAddress(address);
-        this.getEntryMap().values().forEach(x -> {
-            x.setGlobalAddress(address);
-        });
+        this.entryMap.values()
```
Maithem: I think this is incorrect. 

This method is called after `getPayload`, but at that point not all SMR entries are deserialized, so by the time `getSMRUpdates` the entries won't have a global address.

runtime/src/main/java/org/corfudb/recovery/FastObjectLoader.java
```diff
@@ -76,6 +76,7 @@
 
 @Slf4j
 @Accessors(chain = true)
+@Deprecated
```
Maithem: Is this required in this patch ?

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -101,20 +92,39 @@ public void serialize(ByteBuf b) {
...
-                    Serializers.CORFU.serialize(x.getValue(), b);
+
+                    int lengthIndex = b.writerIndex();
+                    b.writeInt(0);
+                    b.writeInt(x.getValue().size());
+                    x.getValue().stream().forEach(smrEntry -> Serializers.CORFU.serialize(smrEntry, b));
+                    int length = b.writerIndex() - lengthIndex - 4;
+                    b.writerIndex(lengthIndex);
+                    b.writeInt(length);
+                    b.writerIndex(lengthIndex + length + 4);
```
Maithem: This can be simplified with 
```java

                    int payloadSizeIndex = b.writerIndex();
                    b.writeInt(0);
                    int payloadIndex = b.writerIndex();
                    b.writeInt(x.getValue().size());
                    x.getValue().stream().forEach(smrEntry -> Serializers.CORFU.serialize(smrEntry, b));
                    int length = b.writerIndex() - payloadIndex;
                    b.writerIndex(payloadSizeIndex);
                    b.writeInt(length);
                    b.writerIndex(payloadIndex + length);

```

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -27,51 +26,42 @@
 public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
 
     // map from stream-ID to a list of updates encapsulated as MultiSMREntry
-    @Getter
-    public Map<UUID, MultiSMREntry> entryMap = Collections.synchronizedMap(new HashMap<>());
+    //TODO(Maithem): we should hide entryMap
+    private final Map<UUID, List<SMREntry>> entryMap = Collections.synchronizedMap(new HashMap<>());
```
Maithem: I think we should also remove MultiSMREntry since its redundant with this container, but that can be done later.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -27,51 +26,42 @@
...
      * @param other Object to merge.
      */
     public void mergeInto(MultiObjectSMREntry other) {
         if (other == null) {
             return;
         }
 
-        other.getEntryMap().forEach((streamID, multiSmrEntry) -> {
-            getStreamEntry(streamID).mergeInto(multiSmrEntry);
+        other.entryMap.forEach((streamID, multiSmrEntry) -> {
```
WenbinZhu: rename multiSmrEntry? Now it's a list

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
WenbinZhu: Maybe we can push the `setGlobalAddress` call down to `StreamViewAdapter`, instead of in `getPayload()`, to avoid this issue, since `StreamViewAdapter` will call `getSMRUpdates(id)` and finish deserialization.

WenbinZhu: Also this issue might not be caught in unit tests, because unit test framework does not serialize the payload into data array and deserialize, it's just the payload being copied to server component, so the address i actually set in `useToken()` method before sending to sever, when getting back data from server and calling `getPayload()`, it won't invoke `((LogEntry) actualValue).setGlobalAddress(getGlobalAddress())`. We might need to write a specific unit test or use IT to reveal the problem.

zalokhan:
Addressed by #2192

PR #2189, opened 2019-10-16
Hide Protobuf Serializer TypeCode.

zalokhan:
## Overview

Description:
The type code of the protobuf serializer should not be exposed.
This for the following reasons:
- It should be used via the CorfuStore as it handles the initialization.
- DynamicProtobufSerializer depends on this code to override it to be able the deserialize the data.

Why should this be merged: 
To have the code in a single place and avoid developer errors in future.

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
+++ b/runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
@@ -59,11 +59,6 @@ import static org.corfudb.runtime.view.TableRegistry.getTypeUrl;
 public class DynamicProtobufSerializer implements ISerializer {
 
     /**
-     * This is the serializer code used by the ProtobufSerializer.
-     */
-    private static final byte PROTOBUF_SERIALIZER_CODE = (byte) 25;
-
-    /**
      * Type code of the serializer. In this case the code needs to override the code of {@link ProtobufSerializer}.
      */
     @Getter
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -86,8 +81,8 @@
      */
     private final ConcurrentMap<String, FileDescriptor> fileDescriptorMap = new ConcurrentHashMap<>();
 
-    public DynamicProtobufSerializer(byte type, CorfuRuntime corfuRuntime) {
-        this.type = type;
+    public DynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
+        this.type = ProtobufSerializer.PROTOBUF_SERIALIZER_CODE;
```
pankti-m: is type needed in this class anymore?

test/src/test/java/org/corfudb/integration/CorfuStoreIT.java
```diff
@@ -3,8 +3,15 @@
 import com.google.common.reflect.TypeToken;
 import com.google.protobuf.DynamicMessage;
 
-import org.corfudb.runtime.collections.*;
 import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
```
hisundar: Double import

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
zalokhan: The getType is an interface method and needs to be implemented.
I'll leave this for uniformity.

PR #2193, opened 2019-10-18
Log the address & streams on deserialization exceptions

hisundar:
## Overview

Description:
This change allowed us to identify which streams are responsible for the deserialization exception and is critical to triaging deployment issues.

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -73,18 +73,27 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
                     if (data == null) {
                         this.payload.set(null);
                     } else {
-                        ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
-                        final Object actualValue =
-                                Serializers.CORFU.deserialize(copyBuf, runtime);
-                        if (actualValue instanceof LogEntry) {
-                            ((LogEntry) actualValue).setGlobalAddress(getGlobalAddress());
-                            ((LogEntry) actualValue).setRuntime(runtime);
-                        }
-                        value = actualValue == null ? this.payload : actualValue;
-                        this.payload.set(value);
-                        copyBuf.release();
-                        lastKnownSize = data.length;
-                        data = null;
+                            ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
+                            final Object actualValue;
+                            try {
+                                actualValue =
+                                        Serializers.CORFU.deserialize(copyBuf, runtime);
+                            } catch (Throwable throwable) {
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -307,7 +307,13 @@ private void updatePointer(final ILogData data) {
      */
     protected boolean processEntryForContext(final ILogData data) {
         if (data != null) {
-            final Object payload = data.getPayload(runtime);
+            try {
+                final Object payload = data.getPayload(runtime);
```
Maithem: Why don't you push this down to `LogData::getPayload` ? (since that's when deserialize is actually invoked)

hisundar: done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -73,18 +73,24 @@ public Object getPayload(CorfuRuntime runtime) {
...
-                        ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
-                        final Object actualValue =
-                                Serializers.CORFU.deserialize(copyBuf, runtime);
-                        if (actualValue instanceof LogEntry) {
-                            ((LogEntry) actualValue).setGlobalAddress(getGlobalAddress());
-                            ((LogEntry) actualValue).setRuntime(runtime);
+                        try {
+                            ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
+                            final Object actualValue =
+                                    Serializers.CORFU.deserialize(copyBuf, runtime);
```
Maithem: Limit the scope of the try/catch to the deserialize line ?

hisundar: done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -73,18 +73,25 @@ public Object getPayload(CorfuRuntime runtime) {
...
-                        copyBuf.release();
-                        lastKnownSize = data.length;
-                        data = null;
+                            ByteBuf copyBuf = Unpooled.wrappedBuffer(data);
+                            final Object actualValue;
+                            try {
+                                actualValue =
+                                        Serializers.CORFU.deserialize(copyBuf, runtime);
+                            } catch (Throwable throwable) {
+                                log.error("Exception caught at address {}, {}, {}",
```
Maithem: Need to release the buffer in the catch block. `copyBuf.release();`

hisundar: good catch. thanks!

PR #2194, opened 2019-10-19
CorfuStore: add Streaming API interfaces and stubs

hisundar:
## Overview
Introduces the streaming api for CorfuStore
These are just stubs for clients to code against.
Implementation will follow later.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
@@ -4,6 +4,7 @@ import com.google.protobuf.Message;
 
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
+import java.util.List;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -150,5 +151,28 @@ public class CorfuStore {
                 this.runtime.getTableRegistry(),
                 this.runtime.getObjectsView(),
                 namespace);
+    }
+
+    /**
+     * Subscribe to a specific a table in a namespace or the entire namespace.
+     * Objects returned will honor transactional boundaries
+     * @param streamListener - callback context.
+     * @param namespace - the CorfuStore namespace to subscribe to.
+     * @param tablesOfInterest - If specified, only updates from these tables will be returned.
+     * @param timestamp - If specified, all stream updates from this timestamp will be returned
+     *                  - if null, only future updates will be returned.
+     */
+    public <K extends Message, V extends Message, M extends Message>
+    void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                   @Nonnull List<TableSchema> tablesOfInterest,
+                   @Nullable Timestamp timestamp) {
+    }
+
+    /**
+     * Gracefully shutdown a streamer.
+     * Once this call returns no further stream updates will be returned.
+     * @param streamListener - callback context.
+     */
+    public void unsubscribe(@Nonnull StreamListener streamListener) {
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
@@ -1,0 +1,24 @@
+package org.corfudb.runtime.collections;
+
+import lombok.Getter;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * A single corfu transaction can/may have updates to multiple tables.
+ * Each table in CorfuStore is represented by a stream identified by a UUID.
+ * This class returns those updates to the tables grouped by their stream UUIDs.
+ * This class allows further expansion of the returned values.
+ *
+ * created by hisundar on 2019-10-22
+ */
+public class CorfuStreamEntries {
+    @Getter
+    private final Map<TableSchema, List<CorfuStreamEntry>> entries;
+
+    public CorfuStreamEntries(Map<TableSchema, List<CorfuStreamEntry>> entries) {
+        this.entries = entries;
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
@@ -1,0 +1,59 @@
+package org.corfudb.runtime.collections;
+
+import com.google.protobuf.Message;
+import lombok.Getter;
+
+/**
+ * Entry returned by CorfuStore's StreamListener interface
+ *
+ * @param <K> - type of the protobuf KeySchema defined while the table was created.
+ * @param <V> - type of the protobuf PayloadSchema defined by the table creator.
+ * @param <M> - type of the protobuf metadata schema defined by table creation.
+ *
+ *  NOTE: Ensure that the above protobuf generated classes are accessible to your JVM!
+ *
+ *  Created by hisundar on 2019-10-18
+ */
+public class CorfuStreamEntry<K extends Message, V extends Message, M extends Message> {
+    /**
+     * Key of the UFO stream entry
+     */
+    @Getter
+    private final K key;
+
+    /**
+     * Value of the UFO stream entry
+     */
+    @Getter
+    private final V payload;
+
+    /**
+     * Metadata (ManagedResource) of the UFO stream entry
+     */
+    @Getter
+    private final M metadata;
+
+    /**
+     * Defines the type of the operation in this stream
+     */
+    public enum OperationType {
+        UPDATE,
+        DELETE;
+    };
+
+    @Getter
+    private final OperationType operation;
+    /**
+     * Stream address of this entry
+     */
+    @Getter
+    private final long address;
+
+    public CorfuStreamEntry(K key, V payload, M metadata, long address, OperationType operation) {
+        this.key = key;
+        this.payload = payload;
+        this.metadata = metadata;
+        this.address = address;
+        this.operation = operation;
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
@@ -1,0 +1,42 @@
+package org.corfudb.runtime.collections;
+
+/**
+ * This is the callback interface that any client subscribing to CorfuStore updates must implement.
+ *
+ * Created by hisundar on 2019-10-18
+ */
+public interface StreamListener {
+    /**
+     * A corfu update can/may have multiple updates belonging to different streams.
+     * This callback will return those updates as a list grouped by their Stream UUIDs.
+     *
+     * @param results is a map of stream UUID -> list of entries of this stream.
+     */
+    void onNext(CorfuStreamEntries results);
+
+    /**
+     * Callback to indicate that an error or exception has occurred while streaming or that the stream is
+     * shutting down. Some exceptions can be handled by restarting the stream (TrimmedException) while
+     * some errors (SystemUnavailableError) are unrecoverable.
+     * @param throwable
+     */
+    void onError(Throwable throwable);
+
+    /**
+     * Return if two streamer objects are equal.
+     * This is to allow CorfuStore may have to put these callback objects into a Set or Map.
+     */
+    boolean equals(StreamListener o);
+
+    /**
+     * Return a unique integer representing this StreamListener
+     * This is to allow CorfuStore may have to put these callback objects into a Set or Map.
+     */
+    int hashcode();
+
+    /**
+     * Return a human readable string representing this StreamListener.
+     * Log messages within CorfuStore will use this to identify different StreamListeners.
+     */
+    String toString();
+}
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TableSchema.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TableSchema.java
@@ -1,0 +1,35 @@
+package org.corfudb.runtime.collections;
+
+import com.google.protobuf.Message;
+import lombok.EqualsAndHashCode;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Definition of the fully qualified Corfu Store Table initialized with default protobuf instances.
+ * This class is necessary for StreamListener's deserialization of the stream entries.
+ * It also allows a way to apply a filter and only specify the tables of interest to streaming.
+ *
+ * @param <K> - Default instance type of the protobuf generated key schema
+ * @param <V> - Default instance type of the protobuf generated payload schema
+ * @param <M> - Default instance type of the protobuf generated metadata schema
+ *
+ * created by hisundar on 2019-10-29
+ */
+
+@EqualsAndHashCode
+public class TableSchema<K extends Message, V extends Message, M extends Message> {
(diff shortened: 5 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -0,0 +1,53 @@
...
+        this.payload = payload;
+        this.metadata = metadata;
+        this.operation = operation;
+    }
+
+    /**
+     * Defines the type of the operation in this stream
+     */
+    public enum OperationType {
+        INSERT,
```
zalokhan: Create, update and delete.
We had a requirement to provide an explicit operation for 'create' and 'update'.

runtime/src/main/java/org/corfudb/runtime/collections/Streamer.java
```diff
@@ -0,0 +1,46 @@
...
+ * Created by hisundar on 2019-10-18
+ */
+public interface Streamer {
+    /**
+     * A corfu update may have multiple updates belonging to different streams.
+     * This callback will return those updates as a list grouped by their Stream UUIDs.
+     *
+     * @param results is a map of stream UUID -> list of entries of this stream.
+     */
+    void onNext(Map<UUID, List<CorfuStreamEntry>> results);
```
zalokhan: We should hide the map data structure behind a result object of our own.
Let me know your thoughts.

runtime/src/main/java/org/corfudb/runtime/collections/Streamer.java
```diff
@@ -0,0 +1,46 @@
...
+     * Return a unique integer representing this streamer
+     * This is to allow CorfuStore may have to put these callback objects into a Set or Map.
+     */
+    int hashcode();
+
+    /**
+     * Return a human readable string representing this streamer.
+     * Log messages within CorfuStore will use this to identify different streamers.
+     */
+    String toString();
```
zalokhan: Should we rename this as `getName`?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -151,4 +151,25 @@ public Query query(@Nonnull final String namespace) {
...
+    /**
+     * Subscribe to a specific a table in a namespace or the entire namespace.
+     * Objects returned will honor transactional boundaries
+     * @param streamer - callback context.
+     * @param namespace - the CorfuStore namespace to subscribe to.
+     * @param tableName - If specified, only updates from this table will be returned.
+     * @param timestamp - If specified, all stream updates from this timestamp will be returned
+     *                  - if null, only future updates will be returned.
+     */
+    public void subscribe(@Nonnull Streamer streamer, @Nonnull String namespace,
```
zalokhan: Not sure if it's a good idea to make the caller pass null.
Should we expose multiple signatures?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -0,0 +1,53 @@
...
+ *  NOTE: Ensure that the above protobuf generated classes are accessible to your JVM!
+ *
+ *  Created by hisundar on 2019-10-18
+ */
+public class CorfuStreamEntry<K extends Message, V extends Message, M extends Message> {
+    /**
+     * Key of the UFO stream entry
+     */
+    @Getter
+    K key;
```
zalokhan: Make all of these private and final.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/Streamer.java
hisundar: good idea!

runtime/src/main/java/org/corfudb/runtime/collections/Streamer.java
hisundar: sure, toString() is the java convention right?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: without null, the number of signatures multiples quickly, that's why I just use null

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
zalokhan: Ok.

runtime/src/main/java/org/corfudb/runtime/collections/Streamer.java
```diff
@@ -0,0 +1,42 @@
+package org.corfudb.runtime.collections;
+
+/**
+ * This is the callback interface that any client subscribe()ing to CorfuStore updates must implement.
+ *
+ * Created by hisundar on 2019-10-18
+ */
+public interface Streamer {
```
sneginhal: nit: StreamListener might be a better name than Streamer

hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -151,4 +151,25 @@ public Query query(@Nonnull final String namespace) {
...
                 namespace);
     }
+
+    /**
+     * Subscribe to a specific a table in a namespace or the entire namespace.
+     * Objects returned will honor transactional boundaries
+     * @param streamListener - callback context.
+     * @param namespace - the CorfuStore namespace to subscribe to.
+     * @param tableName - If specified, only updates from this table will be returned.
+     * @param timestamp - If specified, all stream updates from this timestamp will be returned
```
medhavidhawan: Did not understand the timestamp doc.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
```diff
@@ -0,0 +1,22 @@
+package org.corfudb.runtime.collections;
+
+import lombok.Getter;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * A wrapper around a map of stream UUID -> List of CorfuStreamEntry objects
```
medhavidhawan: Comment is unclear. What is the intention of this class and what is it going to be used for in expansion part.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -0,0 +1,60 @@
+package org.corfudb.runtime.collections;
+
+import com.google.protobuf.Message;
+import lombok.Getter;
+
+/**
+ * This will be the type of each entry returned by CorfuStore's Streamer Interface
```
medhavidhawan: I think you can reword the first sentence to:
Entry returned by CorfuStore's Streamer Interface.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Metadata (ManagedResource) of the UFO stream entry
+     */
+    @Getter
+    private final M metadata;
+
+    /**
+     * Defines the type of the operation in this stream
+     */
+    public enum OperationType {
```
medhavidhawan: hopefully there are no migration / upgrade implications if a new operation is added later on.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
```diff
@@ -0,0 +1,42 @@
+package org.corfudb.runtime.collections;
+
+/**
+ * This is the callback interface that any client subscribe()ing to CorfuStore updates must implement.
```
medhavidhawan: remove the paranthesis in subscribing

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Created by hisundar on 2019-10-18
+ */
+public interface StreamListener {
+    /**
+     * A corfu update may have multiple updates belonging to different streams.
```
medhavidhawan: A corfu update can/may have multiple entries belonging to different streams.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
```diff
@@ -0,0 +1,42 @@
...
+    void onError(Throwable throwable);
+
+    /**
+     * Return if two streamer objects are equal.
+     * This is to allow CorfuStore may have to put these callback objects into a Set or Map.
+     */
+    boolean equals(StreamListener o);
+
+    /**
+     * Return a unique integer representing this streamer
```
medhavidhawan: replace: streamer/StreamListener.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
```diff
@@ -0,0 +1,42 @@
...
+     * This is to allow CorfuStore may have to put these callback objects into a Set or Map.
+     */
+    int hashcode();
+
+    /**
+     * Return a human readable string representing this streamer.
```
medhavidhawan: Replace Streamer everywhere. [ Let us keep the terminology as lean and as close to code as possible.

hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
hisundar: We should be able to add/remove from this Type as long as we can fix up the consumers.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
hisundar: It is basically a wrapper over MultiSMRObject where updates from different tables are returned, grouped by their stream UUIDs. Let me try to reword it better.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: The usage is 
1. take a snapshot which returns a timestamp
2. process all table entries at that timestamp
3. start streaming from that timestamp to process all remaining updates that happened after the snapshot was taken
perhaps I can put the above in the comment?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * A single corfu transaction can/may have updates to multiple tables.
+ * Each table in CorfuStore is represented by a stream identified by a UUID.
+ * This class returns those updates to the tables grouped by their stream UUIDs.
+ * This class allows further expansion of the returned values.
+ *
+ * created by hisundar on 2019-10-22
+ */
+public class CorfuStreamEntries {
```
sneginhal: Do we need CorfuStreamEntries if subscription is per table only?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
```diff
@@ -0,0 +1,24 @@
...
+    @Getter
+    private final Map<TableSchema, List<CorfuStreamEntry>> entries;
```
zalokhan: Shouldn't the key be just TableName - (namespace, tableName)?

runtime/src/main/java/org/corfudb/runtime/collections/TableSchema.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+@EqualsAndHashCode
+public class TableSchema<K extends Message, V extends Message, M extends Message> {
+    private final String tableName;
+    private final Class<K> keyClass;
+    private final Class<V> payloadClass;
+    private final Class<M> metadataClass;
+
+    public TableSchema(String tableName, Class<K> keyClass, Class<V> payloadClass, Class<M> metadataClass) {
```
zalokhan: Can you add `@Nonnull` to the parameters except metadataClass which can be `@Nullable`?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -0,0 +1,60 @@
...
+        CREATE,
```
zalokhan: As discussed offline, can we remove the CREATE for the first phase.
Sorry for the confusion.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
hisundar: as discussed, we support both subscription per table and per selected tables in a namespace

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntries.java
hisundar: callers would need the TableSchema anyway for processing the onNext() call, so I thought it just saves an extra parameter by just indexing on the TableSchema itself?

runtime/src/main/java/org/corfudb/runtime/collections/TableSchema.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
hisundar: no problem. done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
medhavidhawan: Yes please add it as comment.

PR #2201, opened 2019-10-24
[DO NOT SQUASH] Disk-backed CorfuTable Implementation + Performance Patches

vjeko:
## Overview

Description:

 Introduce context-aware SMR Objects

SMR Object are now capable of understanding the context under which an
operation is being executed. This allows data-structures to optimize how
they want to internally represent their data depending on the context.

 Introduce version update policy concept

Allow the underlying object to define what type of versioning update
policy they want to adopt. For some data-structures, allowing the
version updates to be unbounded can cause severe performance issues.

Why should this be merged: Fixes performance related issues with disk-backed CorfuTable.

- [ ] Public API has Javadoc


annotations/src/main/java/org/corfudb/runtime/object/ICorfuVersionPolicy.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    class VersionPolicy {
+    }
+
+    /**
+     * Defines how the underlying object versioning should behave.
+     *
+     * @return The version policy
+     */
+    default VersionPolicy getVersionPolicy() {
```
Maithem: This will generate an accessor. Need to annotate with DontInstrument

annotations/src/main/java/org/corfudb/runtime/object/ISMRObject.java
```diff
@@ -0,0 +1,13 @@
...
+
+
+/**
+ * All objects that need to be understood by Corfu have to implement this
+ * interface, whose purpose is to facilitate better integration between
+ * different Corfu components and the underlying object.
+ *
+ * @param <T> The underlying object type
+ */
+public interface ISMRObject<T>
```
Maithem: I think this needs to be reconciled with `ICorfuSMR`:  On master the `ISMRObject` is actually unnecessary and should have been collapesed with `ICorfuSMR` and concrete Classes should override hashcode, toString,etc.. if they need to.

annotations/src/main/java/org/corfudb/runtime/object/IObjectBuilder.java
```diff
@@ -8,7 +8,7 @@
  * @param <T> The type of the object to build.
  * Created by mwei on 11/12/16.
  */
-public interface IObjectBuilder<T> {
+public interface IObjectBuilder<T extends ISMRObject<T>> {
```
Maithem: On master this should have been `T extends ICorfuSMR`, if you look at `getWrapper` the type will actually get recasted as `(Class<ICorfuSMR<T>>)`. Look at the next comment: this change shouldn't be necessary.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
```diff
@@ -67,7 +68,7 @@ public Token appendCheckpoints(CorfuRuntime rt, String author) {
 
                 CheckpointWriter<T> cpw = new CheckpointWriter(rt, streamId, author, (T) map);
                 ISerializer serializer =
-                        ((CorfuCompileProxy<Map>) map.getCorfuSMRProxy())
+                        ((CorfuCompileProxy<SMRMap<String, String>>) map.getCorfuSMRProxy())
```
Maithem: I don't understand this change, why is it needed?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -197,65 +197,85 @@ public int hashCode() {
...
     }
 
+    // The "main" map which contains the primary key-value mappings.
+    private final StreamingMap<K,V> mainMap;
+    private final Set<Index<K, V, ? extends Comparable>> indexSpec;
+    private final Map<String, Map<Comparable, Map<K, V>>> secondaryIndexes;
+    private final CorfuTable<K, V> optimisticTable;
+    private final VersionPolicy versionPolicy;
+
+    public CorfuTable(StreamingMap<K,V> mainMap,
```
Maithem: Why overload the CorfuTable definition and not create a new type?

If this is being used for the large table use cases, allowing secondary indexes (using this implementation) can become an issue. I think not allowing secondary indexes is better than hoping that it won't be used.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -457,30 +477,30 @@ public void insert(@ConflictParameter K key, V value) {
...
-    public Collection<Map.Entry<K, V>> scanAndFilterByEntry(Predicate<? super Map.Entry<K, V>>
-                                                                    entryPredicate) {
-        return mainMap.entrySet().parallelStream()
-                                    .filter(entryPredicate)
-                                    .collect(Collectors.toCollection(ArrayList::new));
+    public Collection<Map.Entry<K, V>> scanAndFilterByEntry(
+            Predicate<? super Map.Entry<K, V>> entryPredicate) {
+        return mainMap.entryStream()
+                .filter(entryPredicate)
+                .collect(Collectors.toCollection(ArrayList::new));
```
Maithem: `entryStream` can return a parallel stream and it will execute on the common pool, which is a forkjoin pool. This can cause two issues: 
1. If the threads block on io (via rocksdb), i think the forkjoin pool will spawn new threads to maintain "parallelism" 
2. If the stream is too large and since the forkjoin starts with `Runtime.getRuntime().availableProcessors() - 1 ` this might cause CPU starvation issues

I think better thread management is required here, or at least have this documented.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -197,65 +197,85 @@ public int hashCode() {
...
     /**
-     * Default constructor. Generates a table without any secondary indexes.
+     * Generates a table with the given {@link Map} implementation and
+     * without any secondary indexes.
      */
-    public CorfuTable() {
-        this(IndexRegistry.empty(), new HashMap<>());
+    public CorfuTable(Map<K, V> mapImpl) {
+        this(IndexRegistry.empty(), mapImpl);
     }
```
Maithem: Hmm, it seems like this is also changing the current behavior for the in memory CorfuTable with respect to `scanAndFilter` . Now the `scan* ` methods will use more resources, which is not expected.

I think this is another reason to split the types and not overload.

annotations/src/main/java/org/corfudb/runtime/object/ICorfuVersionPolicy.java
vjeko: Sounds good. I also need to check if the SRM Proxy Compiler actually works with inheritance.

annotations/src/main/java/org/corfudb/runtime/object/ISMRObject.java
vjeko: Yep, makes sense.

annotations/src/main/java/org/corfudb/runtime/object/IObjectBuilder.java
vjeko: I see.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
vjeko: Does not compile otherwise. Don't quite know why the compiler was not happy. I think `Map<String, String>` should work as well.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: No good reason -- mostly to avoid code duplication.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
Maithem: You can drop the "map" cast and just use it a wrapped object. Why not use:

> `ISerializer serializer = ((CorfuCompileProxy) map.getCorfuSMRProxy())
                                .getSerializer();`

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: The disk-backed CorfuTable always has to be single threaded. The actual behavior is defined in the mapImp. 

For memory (parallel):
```
    public Stream<Entry<K, V>> entryStream() {
        return mapImpl.entrySet().stream().parallel();
    }
```

For disk (single threaded):
```
    public Stream<Entry<K, V>> entryStream() {
        final RocksIterator rocksIterator = rocksDb.newIterator();
        rocksIterator.seekToFirst();
        return Streams.stream(new RocksDbIterator(rocksIterator));
    }
```

RocksDB itself does not allow multi-threaded access either.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Not sure I follow. Can you elaborate?

runtime/pom.xml
```diff
@@ -139,6 +139,11 @@
             <artifactId>sizeof</artifactId>
             <version>0.3.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.rocksdb</groupId>
+            <artifactId>rocksdbjni</artifactId>
```
Maithem: @vjeko @xnull maybe you should also consider lmdb

Its extremely reliable and stable [https://github.com/lmdbjava/benchmarks/blob/master/results/20160710/README.md](url)

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -428,8 +448,8 @@ protected void undoPut(CorfuTable<K, V> table, V undoRecord, K key, V value) {
      * @param undoRecord    The undo record generated by undoRemoveRecord
      */
     @DontInstrument
-     void undoPutAll(CorfuTable<K, V> table, Map<K,V> undoRecord,
-                            Map<? extends K, ? extends V> m) {
+    void undoPutAll(CorfuTable<K, V> table, Map<K,V> undoRecord,
```
Maithem: This can actually be omitted for `ICorfuVersionPolicy.MONOTONIC`

runtime/pom.xml
vjeko: One of the problems that we ran into is the lack of flexibility from these embedded databases (EHCache to be more specific). One of the reasons why we opted for Rocks is because of the amount of configurable parameters.

But yeah, we will look into it.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Ok.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -566,14 +588,22 @@ public void clear() {
...
     public @Nonnull Set<Entry<K, V>> entrySet() {
-        List<Entry<K, V>> copy = new ArrayList<>(mainMap.size());
-        for (Map.Entry<K, V> entry : mainMap.entrySet()) {
-            copy.add(new AbstractMap.SimpleImmutableEntry<>(entry.getKey(),
-                    entry.getValue()));
-        }
-        return new ImmuableListSetWrapper<>(copy);
+        return mainMap.entryStream().map(entry ->
+                new AbstractMap.SimpleImmutableEntry<>(entry.getKey(), entry.getValue()))
+                .collect(ImmutableSet.toImmutableSet());
```
Maithem: This the objects have to be rehashed this can be expensive ~5-10x 

For a map of 1 million entries it:
1 million

iter 0
[guava] elapsed(ms) :                  463
[ImmuableListSetWrapper] elapsed(ms) : 61
iter 1
[guava] elapsed(ms) :                  402
[ImmuableListSetWrapper] elapsed(ms) : 63
iter 2
[guava] elapsed(ms) :                  388
[ImmuableListSetWrapper] elapsed(ms) : 52
iter 3
[guava] elapsed(ms) :                  394
[ImmuableListSetWrapper] elapsed(ms) : 54


For a map of 500 thousand entries:
iter 0
[guava] elapsed(ms) :                  233
[ImmuableListSetWrapper] elapsed(ms) : 33
iter 1
[guava] elapsed(ms) :                  184
[ImmuableListSetWrapper] elapsed(ms) : 38
iter 2
[guava] elapsed(ms) :                  197
[ImmuableListSetWrapper] elapsed(ms) : 31


In this experiment, the map uses a UUID for both key and value, but for complex types, this can be much more expensive.

vjeko: Can you share you benchmarking code? Previously we were doing `copy.add(...)`, so I am not sure I am able to explain the difference.

vjeko: I see. `ImmuableListSetWrapper` is our implementation.

vjeko: Alright, I will collect it to a `List` and convert it to `ImmuableListSetWrapper`, so there should not be any performance difference.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,317 @@
...
+    public void putAll(@NonNull Map<? extends K, ? extends V> map) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void clear() {
+        entryStream().map(Entry::getKey).forEach(this::remove);
```
Maithem: You can use `DestroyDB` here

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,317 @@
...
+        return optimisticMap;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Stream<Entry<K, V>> entryStream() {
+        final RocksIterator rocksIterator = rocksDb.newIterator();
+        rocksIterator.seekToFirst();
```
Maithem: Need to double check this, but there used to be an issue in RocksDB where the seek cost will also included deleted keys, which can be slow over time. Keys are tombstoned, but only removed on compaction. If that's still the case, you can do a direct seek to the smallest key which you can keep track of.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
Maithem: I think we should split it, the way its structured right now is error-prone. Actually, there's a leak. If you look at `TableRegistry::openTable` an instance of `PersistedStreamingMap` is passed, which itself isn't SMR, but passed to CorfuTable to be used during construction. This means that if the outer object (CorfuTable) needs to be reconstructed (i.e. TrimmedException/NoRollBackException) it will use state from the old instance. In other words, the same RocksDB instances can be used multiple times for different CorfuTable instances. Imagine the following scenario:
1. CorfuTable is created
2. CorfuTable is access at irregular intervals which results in intermittent TrimmedExceptions (handled internally)
3. Every time a trim exception is encountered, the VLO will invoke the CorfuTable constructor (to create a new object)
4. The object from step 3 will have state from the previous instance (since the PersistedStreamingMap is not constructed as a part of CorfuTable, but passed as a reference)

In essence, the underlying resources will never be closed/cleaned and therefore can grow overtime. 
Decoupling those two doesn't make a lot of sense (to me at least), I think Its possible to do code sharing in a different way.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -47,8 +48,7 @@
  * <p>Created by mwei on 11/13/16.
  */
 @Slf4j
-public class VersionLockedObject<T> {
-
+public class VersionLockedObject<T extends ISMRObject<T>> {
```
Maithem: Nice. 

You can make `ISMRObject` closeable and then call `close` in `resetUnsafe` before the new object is constructed.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,317 @@
...
+    private final CorfuRuntime corfuRuntime;
+    private final ISerializer serializer;
+    private final RocksDB rocksDb;
+
+    public PersistedStreamingMap(@NonNull Path dataPath,
+                                 @NonNull Options options,
+                                 @NonNull ISerializer serializer,
+                                 @NonNull CorfuRuntime corfuRuntime) {
+        try {
+            FileUtils.deleteDirectory(dataPath.toFile());
```
Maithem: RIP. This will recursively delete the files one-by-one because java "can't" delete a directory that isn't empty. 
Alternatively, you can call DbDestroy if the path exists, the underlying implementation might be more efficient, not sure. Need to investigate.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,317 @@
...
+
+            return next != null;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Entry<K, V> next() {
+            checkInvariants();
```
Maithem: You can remove this since `hasNext` will call `checkInvariants`

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,317 @@
...
+         */
+        @Override
+        public boolean hasNext() {
+            // If we have the next element pipelined, go ahead and return true.
+            if (next != null) {
+                return true;
+            }
+
+            // If the iterator is valid, this means that the next entry exists.
+            checkInvariants();
```
Maithem: Can you omit `iterator.status` from every call? its a native call (i.e an extra sys call)

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,317 @@
...
+public class PersistedStreamingMap<K, V> implements StreamingMap<K, V> {
+
+    private final StreamingMap<K, V> optimisticMap = new StreamingMapDecorator<>();
+    private final AtomicInteger dataSetSize = new AtomicInteger();
+    private final CorfuRuntime corfuRuntime;
+    private final ISerializer serializer;
+    private final RocksDB rocksDb;
+
+    public PersistedStreamingMap(@NonNull Path dataPath,
+                                 @NonNull Options options,
```
Maithem: Probably can enable compression by default and disable checksums.
If data can be lost (i.e. fsync not enabled), then why not also tolerate incorrect checksums ?

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -83,7 +84,7 @@ public Table(@Nonnull final String namespace,
                 .setTypeToken(CorfuTable.<K, CorfuRecord<V, M>>getTableType())
                 .setStreamName(this.fullyQualifiedTableName)
                 .setSerializer(serializer)
-                .setArguments(new ProtobufIndexer(valueSchema))
+                .setArguments(new ProtobufIndexer(valueSchema), streamingMap)
```
Maithem: Creating `StreamingMap` should happen via CorfuTable, look at my comment above about the leak.

runtime/src/main/java/org/corfudb/util/serializer/JsonSerializer.java
```diff
@@ -58,7 +58,7 @@ public Object deserialize(ByteBuf b, CorfuRuntime rt) {
             try {
                 return rt.getObjectsView().build()
                         .setStreamID(new UUID(b.readLong(), b.readLong()))
-                        .setType(Class.forName(smrClassName))
+                        .setType((Class<? extends ISMRObject>) Class.forName(smrClassName))
```
Maithem: Yea, these casts can be cleaned up once we directly use ISMRObject.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -667,7 +718,7 @@ protected void optimisticRollbackUnsafe() {
      * @param entry smr entry
      */
     public void applyUpdateToStreamUnsafe(SMREntry entry, long globalAddress) {
-        applyUpdateUnsafe(entry);
+        applyUpdateUnsafe(entry, globalAddress);
```
Maithem: SMREntry has `getGlobalAddress`

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -209,7 +208,7 @@ public void gc(long trimMark) {
                 if (directAccessCheckFunction.apply(this)) {
                     log.trace("Access [{}] Direct (optimistic-read) access at {}",
                             this, getVersionUnsafe());
-                    R ret = accessFunction.apply(object);
+                    R ret = accessFunction.apply(object.getContext(ICorfuExecutionContext.DEFAULT));
```
Maithem: Might be missing something, but what's the point of staging/upstaging uncommitted changes, if the access methods can never access the upstaged context? Essentially all that work is unnecessary, since it has no consequence on the read path. 

It seems like this PR introduces changes to implement some sort of read-committed semantics, but in many systems read committed actually returns upstaged changes (created by yourself/same transaction), so this seems a bit strange: you can't read you're own writes?  

You can easily support read-your-write semantics by re-routing the access function to either the `ICorfuExecutionContext.DEFAULT`, or `ICorfuExecutionContext.Optimistic`

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -566,14 +588,22 @@ public void clear() {
...
     }
 
+    /**
+     * Present the content of a {@link CorfuTable} via the {@link Stream} interface.
+     *
+     * @return stream of entries
+     */
+    @Accessor
+    public @Nonnull Stream<Entry<K, V>> entryStream() {
+        return mainMap.entryStream();
```
Maithem: I believe you need to change the `CheckpointWriter` to consume this API if the underlying `StreamingMap` is using `PersistedStreamingMap`, otherwise the whole RocksDB instance will be recreated in memory.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,317 @@
...
+                                 @NonNull ISerializer serializer,
+                                 @NonNull CorfuRuntime corfuRuntime) {
+        try {
+            FileUtils.deleteDirectory(dataPath.toFile());
+            this.rocksDb = RocksDB.open(options, dataPath.toFile().getAbsolutePath());
```
Maithem: RocksDB will internally allocate a couple of caches, what are the defaults, are they sane?
Maybe document it here?

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: That is correct  -- tombstoned entries are only removed on a full compaction. Maybe a cleaner approach would be to set `compaction_options_universal.max_size_amplification_percent`. For example, take a look at https://github.com/facebook/rocksdb/wiki/Universal-Compaction (Compaction Triggered by Space Amplification).

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: From their FAQ:

_**Q: What is the difference between DestroyDB() and directly deleting the DB directory manually?**

A: The major difference is that DestroyDB() will take care of the case where the RocksDB database is stored in multiple directories. For instance, a single DB can be configured to store its data in multiple directories by specifying different paths to DBOptions::db_paths, DBOptions::db_log_dir, and DBOptions::wal_dir._

For our use-case, there is not any difference, but it does seem cleaner if we resort to DbDestroy. Also, on any POSIX system, you cannot simply remove a non-empty directory. This is always a two step procedure:

1) Traverse the hierarchy and remove non-directories.
2) Remove the empty directories.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Fixed.

runtime/pom.xml
medhavidhawan: I have only one comment here:
We need to split the effort:
1. Work on POC to see if we can work directly off the logs and not persist the data in map form on disk. At least something to that effect.
2. Look for a technology which is more optimal for the implementation we have and also refine our integration such that it is more performant over time.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: According to their documentation, calling `DestroyDB()` on a live DB is an undefined behavior. The correct approach is to close the DB then destroy the DB. However, opening and closing the DB mid-air seems somewhat dangerous to me, since we now also have to deal with all IO exceptions, from which we might not be able to recover.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: The compression is enabled by default:

_Use options.compression to specify the compression to use. By default it is Snappy. We believe LZ4 is almost always better than Snappy. We leave Snappy as default to avoid unexpected compatibility problems to previous users. LZ4/Snappy is lightweight compression so it usually strikes a good balance between space and CPU usage._

Looks like we can disable checksum verification, but I am unable to pinpoint on how to disable checksum being persisted.

I'll be more explicit and add both of these options.

vjeko: Looks like setting checksum verification on the read path has to be done on every read operation by passing the corresponding `ReadOptions`. Seems pretty invasive to me, and I am pretty sure we won't see any difference (maybe if we were using memory backed file-system the situation would be different), so I am just going to go with the defaults.

Compression is now explicitly set to LZ4.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: There are two concerns here:

1) Resetting the SMR object when we hit some sort of an unrecoverable error. The bug you are referring to can be addressed by passing a Supplier<T> instead of T.

2) Figuring out how to split memory and disk implementations into two different classes.

I agree that we should do (2), but my suggestion is to do it in a subsequent patch, mostly because the blast radius is pretty large, and figuring out how to do it cleanly is going to take some time.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: The documentation says to make sure that the iterator is valid before calling `next()` on it, so I am not sure if I have a choice.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
vjeko: Same thing. Supplier<T> should fix it.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
vjeko: You are not missing anything. The plan is to implement read-your-own-dirty-writes, but since the current clients do not care about this type of behavior, I left it as unimplemented.

This will have to be implemented as a two-phase lookup.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
vjeko: Ok.

annotations/src/main/java/org/corfudb/runtime/object/ICorfuVersionPolicy.java
vjeko: Done.

annotations/src/main/java/org/corfudb/runtime/object/ISMRObject.java
vjeko: Consolidated.

annotations/src/main/java/org/corfudb/runtime/object/IObjectBuilder.java
vjeko: Fixed. `ISMRObject` does not exist anymore.

runtime/src/main/java/org/corfudb/runtime/MultiCheckpointWriter.java
vjeko: Done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Fixed. `CorfuTable` now takes `Supplier<T>`.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: Turns out that we end up recreating the object in case of a reset-type call. So `DestroyDB()` is called implicitly in this case.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: Done -- `universal.max_size_amplification_percent` is set to 50%.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
vjeko: Yep, `ICorfuSMR` is not `AutoClosable`.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
vjeko: Turns out we cannot rely on the `getGlobalAddress()`, since in some cases, special timestamp is being passed:

```
       long syncTo = (timestamp == Address.OPTIMISTIC) ? Address.MAX : timestamp;
        stream.streamUpTo(syncTo)
                .forEachOrdered(entry -> {
                    try {
                        Object res = applyUpdateUnsafe(entry, timestamp);
```

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Good catch. I tried changing the `CheckpointWriter` to consume the `StreamingMap`, but the problem is that there is about 20 different tests that need to be modified as a result of this change.

I suppose I will have to address this separately.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: It's documented now:

```
    /**
     * A set of options defined for disk-backed {@link CorfuTable}.
     *
     * For a set of options that dictate RocksDB memory usage can be found here:
     * https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB
     *
     * Block Cache:  Which can be set via Options::setTableFormatConfig.
     *               Out of box, RocksDB will use LRU-based block cache
     *               implementation with 8MB capacity.
     * Index/Filter: Is a function of the block cache. Generally it infates
     *               the block cache by about 50%. The exact number can be
     *               retrieved via "rocksdb.estimate-table-readers-mem"
     *               property.
     * Write Buffer: Also known as memtable is defined by the ColumnFamilyOptions
     *               option. The default is 64 MB.
     */
```

runtime/src/main/java/org/corfudb/util/serializer/JsonSerializer.java
vjeko: Looks like they are still needed. The current implementation is:

`<R extends ICorfuSMR<R>> IObjectBuilder<R> setType(Class<R> type);`

annotations/src/main/java/org/corfudb/runtime/object/ICorfuSMR.java
```diff
@@ -8,40 +8,60 @@
...
      * @return The undo map. */
-    Map<String, IUndoFunction<T>> getCorfuUndoMap();
+    default Map<String, IUndoFunction<T>> getCorfuUndoMap() {
+        throw new IllegalStateException("");
+    }
 
     /** Get a map from strings (function names) to undoRecord methods.
      * @return The undo record map. */
-    Map<String, IUndoRecordFunction<T>> getCorfuUndoRecordMap();
+    default Map<String, IUndoRecordFunction<T>> getCorfuUndoRecordMap() {
```
xnull: why you need to make all these methods default and throw an exception at the same time. You can just get rid of `default` modifier and then you don't need to throw exceptions

vjeko: The implementation has to be defined this way since the real implementation is generated by Corfu Compile Proxy. If you leave these function unimplemented, you are forcing all SMR objects to define these functions.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,343 @@
...
+                    keyPayload.array(), keyPayload.arrayOffset(), keyPayload.readableBytes(),
+                    valuePayload.array(), valuePayload.arrayOffset(), valuePayload.readableBytes());
+        } catch (RocksDBException ex) {
+            throw new UnrecoverableCorfuError(ex);
+        } finally {
+            keyPayload.release();
+            valuePayload.release();
+        }
+
+        dataSetSize.incrementAndGet();
```
hisundar: LSM cannot compute the size of elements until full merge happens. So do we know if this value is accurate? We might have to do a rocksDb.get() to see if the value exists and then selectively increment right? Please correct me if this is already handled. thanks

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileWrapperBuilder.java
```diff
@@ -44,6 +43,9 @@
         ICorfuSMR<T> wrapperObject = (ICorfuSMR<T>) ReflectionUtils.
                             findMatchingConstructor(wrapperClass.getDeclaredConstructors(), args);
 
+        // Close this object since a new one will be instantiated by the VersionLockedObject.
```
Maithem: I think we should clean this up. Please look at #2225 and #2234 , I pushed some fixes and now your tests passes. 

The problem is to create an SMRObject you need to create two objects, one servers as a facade, which is instrumented and passed to the user and the other is managed by the proxy via VersionLockedObject.

What was happening was that the `close` method was being executed on the facade and not the underlying object. Actually, the underlying object which contains the actual state can't be invoked without invoking a facade method that has been annotated as either @Mutator @Accessor @MutatorAccessor, or @PassThrough. 

`close` which ideally isn't a mutator/accessor needs to execute via PassThrough and NOT "not instrumented", but that functionality seems to be broken on master, so i fixed it here #2234 .

Another issue is, the specification for an SMRObject is not well defined. For example, a consumer can access/mutate instrumented code and non-instrumented code, which execute on two different objects (not expected by the user, hence this bug #2225). Based on Java's typing and how our code generation happens, I think the cleanest way to break this circular dependency is to stipulate that the Object definition reserve the default constructor to generate an identity instance (i.e. an empty object that serves as a facade), so line 43 will use that object only as a facade and then the Version locked object can create the other instance with an other constructor/params it requires. Another requirement that we should enforce is that every public method should be annotated, everything else should be private. Right now we can address that by documentation, but later on it would be nice to support it in the annotation processor directly.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
hisundar: https://github.com/facebook/rocksdb/releases
you can get an estimate but count isn't accurate in RocksDB...
https://stackoverflow.com/questions/25538750/getting-total-number-of-key-value-pairs-in-rocksdb

vjeko: For now I am using `RocksDB::keyMayExist(...)` which is more light weight. Technically, in order to be compliant with the `Map` interface we should be returning the value to which this map previously associated the key,

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileWrapperBuilder.java
vjeko: Adopted. Rather than overloading the semantics of a default constructor, I added a new API call `closeWrapper()` which releases the resources just before the VLO instantiates a new object. But I agree with you -- we need some extensive cleanup in that area.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,323 @@
...
+                    keyPayload.array(), keyPayload.arrayOffset(), keyPayload.readableBytes(),
+                    valuePayload.array(), valuePayload.arrayOffset(), valuePayload.readableBytes());
+        } catch (RocksDBException ex) {
+            throw new UnrecoverableCorfuError(ex);
+        } finally {
+            keyPayload.release();
+            valuePayload.release();
+        }
+
+        return value;
```
lixinchengdu: The return value of put is previous value if it exists.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
lixinchengdu: clear() does not end up recreating new object.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: This is by design. Due to significant performance penalty I decided to deviate from the original `Map` interface and return the current inserted value instead.

annotations/src/main/java/org/corfudb/runtime/object/ICorfuExecutionContext.java
```diff
@@ -0,0 +1,41 @@
...
+     * execute under the specified context.
+     *
+     * For the data-structures that do not care about different contexts,
+     * they should simply return "this".
+     *
+     * @param context context under which the operations will be applied
+     * @return the underlying object for the specified context
+     */
+    @DontInstrument
+    T getContext(Context context);
```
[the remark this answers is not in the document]
vjeko: Except when you use it within VLO and do `object.getObject(...)` which ends up being very confusing. Maybe `withinContext()` would be better?

vjeko:
@lixinchengdu it does not let me reply to you comment, but reset of the object happens in the VLO. Take a look at this code:

```
        } else if (resetSet.contains(entry.getSMRMethod())) {
            // If this is a reset, undo by restoring the
            // previous state.
            object = (T) entry.getUndoRecord();
            // clear the undo record, since it is now
            // consumed (the object may change)
            entry.clearUndoRecord();
            return;
        }
```

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -198,45 +199,60 @@ public int hashCode() {
...
     /**
      * Generate a table with a given implementation for the {@link StreamingMap}.
      */
     public CorfuTable(IndexRegistry<K, V> indices,
                       Supplier<StreamingMap<K, V>> streamingMapSupplier) {
+        this.indexSpec = new HashSet<>();
+        this.secondaryIndexes = new HashMap<>();
+        this.mainMap = streamingMapSupplier.get();
+
+        this.optimisticTable = new CorfuTable<>(this.mainMap.getOptimisticMap(), this.indexSpec,
```
lixinchengdu: optimisitcTable breaks the abstraction of SMR. All optimstic updates assume that the initial table is empty. This is correct for CorfuTable because different keys are orthogonal, but it is not true for general SMR.

lixinchengdu: Moreover, one single optimisticTable prevent us developing nested transactions.

lixinchengdu:
> // If this is a reset, undo by restoring the

Thanks! Got it.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -293,6 +293,10 @@ public void gc(long trimMark) {
      *                             the supplied version.
      */
     public void rollbackObjectUnsafe(long timestamp) {
+        if (object.getVersionPolicy() == ICorfuVersionPolicy.MONOTONIC) {
```
lixinchengdu: If going back in time is not allowed, how could the upper layer client know the restriction. Is there any exception thrown?

vjeko: Well, when the client creates the SMR object, they also defines the versioning policy. In other words, this option is explicitly defined by the user.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Good catch! The implementation is incomplete -- there is a JIRA ticket that references this issue. Long story short, the TransactionalContext and the OptimisticContext need to be reconciled. This is going to be a non-trivial change, so I opted not to tackle this issue in this PR.

PR #2192, opened 2019-10-18
Lazy Deserialization

Maithem:
## Overview
When a LogEntry is deserialized, stream updates are only deserilized
on access. In essence, allowing a stream to only deserilize its updates.
This is useful when a CorfuRuntime doesn't have the serializers for all
other stream updates within the same LogEntry.
Description:

Why should this be merged: Addresses the problem where a client needs to read stream entries that exist as a part of a bigger transaction, but doesn't have a serilizer for all other streams within the same transaction. 

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #2205, opened 2019-10-25
Lazy Deserialization Part 2

Maithem:
## Overview
PR #2192 changed the serialization format of MultiObjectSMREntry.
In order to avoid data migration, this patch reverts those format
changes. Those changes are convienent, but not necessary to
implement lazy deserialization. This patch does that: lazy
deserialization without format changes.

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
@@ -8,7 +8,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.util.serializer.Serializers;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -92,8 +91,14 @@ public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
         for (int i = 0; i < numStreams; i++) {
             UUID streamId = new UUID(b.readLong(), b.readLong());
 
-            int updateLength = b.readInt();
-            byte[] streamUpdates = new byte[updateLength];
+            // The MultiObjectSMREntry payload is structure as follows:
+            // LogEntry Type | number of MultiSMREntry entries | MultiSMREntry id | serialized MultiSMREntry | ...
+            // Therefore we need to unpack the MultiSMREntry entries one-by-one
+            int start = b.readerIndex();
+            MultiSMREntry.seekToEnd(b);
+            int multiSMRLen = b.readerIndex() - start;
+            b.readerIndex(start);
+            byte[] streamUpdates = new byte[multiSMRLen];
             b.readBytes(streamUpdates);
             streamBuffers.put(streamId, streamUpdates);
         }
@@ -105,19 +110,9 @@ public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
         b.writeInt(streamUpdates.size());
         streamUpdates.entrySet().stream()
                 .forEach(x -> {
-                    // Stream payload structure
-                    // | stream id | payload size | SMR Update 1 | ... | SMR Update N|
                     b.writeLong(x.getKey().getMostSignificantBits());
                     b.writeLong(x.getKey().getLeastSignificantBits());
-                    int payloadSizeIndex = b.writerIndex();
-                    b.writeInt(0);
-                    int payloadIndex = b.writerIndex();
-                    b.writeInt(x.getValue().getUpdates().size());
-                    x.getValue().getUpdates().stream().forEach(smrEntry -> Serializers.CORFU.serialize(smrEntry, b));
-                    int length = b.writerIndex() - payloadIndex;
-                    b.writerIndex(payloadSizeIndex);
-                    b.writeInt(length);
-                    b.writerIndex(payloadIndex + length);
+                    Serializers.CORFU.serialize(x.getValue(), b);
                 });
     }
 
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -92,8 +91,11 @@ public void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
         for (int i = 0; i < numStreams; i++) {
             UUID streamId = new UUID(b.readLong(), b.readLong());
 
-            int updateLength = b.readInt();
-            byte[] streamUpdates = new byte[updateLength];
+            int start = b.readerIndex();
+            MultiSMREntry.seekToEnd(b);
```
xnull: why it's a static method? Can we extract that to a utility class?
If not, I'm ok with current impl

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -92,8 +91,11 @@ public void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
...
+            int multiSMRLen = b.readerIndex() - start;
+            b.readerIndex(start);
+            byte[] streamUpdates = new byte[multiSMRLen];
```
xnull: do we really need a copy constructor here?

runtime/src/main/java/org/corfudb/util/serializer/CorfuSerializer.java
```diff
@@ -12,7 +12,7 @@
     private final byte type;
 
     /* The magic that denotes this is a corfu payload */
-    final byte corfuPayloadMagic = 0x42;
+    public static final byte corfuPayloadMagic = 0x42;
```
xnull: why public?

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
```diff
@@ -117,6 +120,30 @@ void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
...
     }
 
+    /**
+     * Given a buffer with the reader index pointing to a serialized SMREntry, this method will
+     * seek the buffer's reader index to the end of the entry.
+     */
+    public static void seekToEnd(ByteBuf b) {
+        // Magic
+        byte magicByte = b.readByte();
+        checkState(magicByte == CorfuSerializer.corfuPayloadMagic, "Not a ICorfuSerializable object");
```
WenbinZhu: This doesn't look like a clean implementation under the current abstraction. I think we should do this inside CorfuSerializer (add a method in CorfuSerializer), and not exposing the handling of the magic.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiSMREntry.java
```diff
@@ -64,6 +65,23 @@ void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
...
     }
 
+    /**
+     * Given a buffer with the reader index pointing to a serialized MultiSMREntry, this method will
+     * seek the buffer's reader index to the end of the entry.
+     */
+    public static void seekToEnd(ByteBuf b) {
+        // Magic
+        byte magicByte = b.readByte();
+        checkState(magicByte == CorfuSerializer.corfuPayloadMagic, "Not a ICorfuSerializable object");
```
WenbinZhu: Same here.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
Maithem: It can be. But since seeking logic is specific to a container, I feel like its appropriate to place here.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
Maithem: I can slice the ByteBuf and avoid this allocation/copy, but that would  hold a reference to the parent buffer. That the parent buffer, which can be much larger, can't be de-allocated and reused.

runtime/src/main/java/org/corfudb/util/serializer/CorfuSerializer.java
Maithem: Good point. I'll just add a seek method to ISerializer and change this to private. We shouldn't expose this.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
Maithem: Good point. I'll just add a seek method to ISerializer and change this to private. We shouldn't expose this.

Maithem: > This seems like a short-term solution right? Because all these seeks might incur performance regression.

Unless we change the format (i.e. prefix the whole container's length with a length field), we can't just skip bytes, and the point of this patch is to not change the format. Let me know if you have a more optimal way.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
annym: Can we add a minor comment on the stream payload structure (and that this is required because the payload size is not explicitly given)

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
Maithem: @WenbinZhu @xnull By hiding the magic byte, I need to make a couple of components "seek aware", given that this is only a short term solution I think this patch is a reasonable one. 

I tried to the other implementation here:
https://github.com/CorfuDB/CorfuDB/compare/multiObjectSMRFormat2

Also, once we change the format, this won;t be needed, so I'm kind of hesitant to make the uber change.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
Maithem: done.

Maithem:
> Having some tests for `seekToEnd(...`) would be great,

Done.

PR #2213, opened 2019-10-29
Removed redundant isMetricsEnabled in CorfuCompileProxy.

zhangn49:
## Overview
Cleaned up redundant part in CorfuCompileProxy.

Description:
1. Removed redundant boolean isMetricsEnabled, since it refers metricsCollectionEnabled from MetricsUtils then applies it to function of MetricsUtils.
2. Removed unnecessary global variables.
3. Replaced anonymous classes with lambda in MetricsUtils.

Why should this be merged: 
Removed redundant and confused part.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -154,16 +151,15 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
                 upcallTargetMap, undoRecordTargetMap,
                 undoTargetMap, resetSet);
 
-        metrics = CorfuRuntime.getDefaultMetrics();
```
lixinchengdu: why MetricRegistry becomes non final?

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
```diff
@@ -288,24 +292,14 @@ public static void incConditionalCounter(boolean enabled, @NonNull Counter count
      * return a gauge on direct memory used by netty's PooledByteBufAllocator
      */
     private static Gauge<Long> getNettyPooledDirectMemGauge() {
-        return new Gauge<Long>() {
-            @Override
-            public Long getValue() {
-                return PooledByteBufAllocator.DEFAULT.metric().usedDirectMemory();
-            }
-        };
+        return () -> PooledByteBufAllocator.DEFAULT.metric().usedDirectMemory();
```
lixinchengdu: Even though the new one is cleaner, the old on is easier to understand.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
zhangn49: I thought local variable does not need to be final. Fixed.

runtime/src/main/java/org/corfudb/util/MetricsUtils.java
zhangn49: I agree. I always use IntelliJ to translate lambda to anonymous class when I don't understand. Modified these to keep a unified code style.

PR #2216, opened 2019-11-04
CorfuQueue: Support for custom serializer & Secondary Indexes

hisundar:
## Overview

Consumers of the CorfuQueue need a way to set a custom serializer based on the integration framework.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
@@ -4,9 +4,12 @@ import com.google.common.reflect.TypeToken;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.collections.CorfuTable.IndexRegistry;
 import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
 import org.corfudb.runtime.view.CorfuGuidGenerator;
+import org.corfudb.util.serializer.ISerializer;
+import org.corfudb.util.serializer.Serializers;
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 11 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -39,16 +41,23 @@
...
                 .setStreamName(streamName)
-                .setArguments(CorfuTable.IndexRegistry.empty(), new LinkedHashMap<Long, E>())
+                .setArguments(indices, new LinkedHashMap<Long, E>())
+                .setSerializer(serializer)
                 .open();
         guidGenerator = CorfuGuidGenerator.getInstance(runtime);
     }
 
+    public CorfuQueue(CorfuRuntime runtime, String streamName) {
+        this(runtime, streamName, Serializers.JSON, CorfuTable.IndexRegistry.empty());
```
xnull: Why it's JSON not CORFU?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -39,16 +41,23 @@
...
+
```
xnull: Please import `CorfuTable.IndexRegistry` you can just use `IndexRegistry` without prefix class

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
```diff
@@ -67,6 +68,20 @@ public void basicQueueOrder() {
         assertThat(records.get(0).getRecordId().compareTo(records2.get(1).getRecordId())).isLessThan(0);
     }
 
+    @Test
+    public void queueWithSecondaryIndexCheck() {
```
xnull: can you write more tests and cover more border cases if they exist?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: I am just retaining the default serializer which is JSON if not specified.

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
hisundar: right now it's only the serializer which is settable in this PR, so I only have unit test for that. Later if we add more features with queries, then I will add tests along with that feature.

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
```diff
@@ -67,6 +68,20 @@ public void basicQueueOrder() {
...
+        CorfuQueue<String>
+                corfuQueue = new CorfuQueue<>(getDefaultRuntime(), "test", Serializers.JAVA,
+                CorfuTable.IndexRegistry.empty());
+
+        CorfuRecordId idC = corfuQueue.enqueue("c");
+        CorfuRecordId idB = corfuQueue.enqueue("b");
+        CorfuRecordId idA = corfuQueue.enqueue("a");
+
+        List<CorfuQueueRecord<String>> records = corfuQueue.entryList();
+        assertThat(records.size()).isEqualTo(3);
```
annym: I think this statement is giving the compilation issues. "3" a magic number

hisundar: thanks! fixed

PR #2217, opened 2019-11-04
CorfuStore: auto retry on SEQUENCER_OVERFLOW transaction aborts

hisundar:
## Overview

For high concurrency use cases where only new keys are getting inserted transactionally the current cache eviction (LFU) in sequencer server can cause transaction aborts which can be consumed and retried internally in CorfuStore.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
@@ -5,11 +5,15 @@ import com.google.protobuf.Message;
 import java.util.ArrayList;
 import java.util.List;
 
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.Token;
 import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+import org.corfudb.runtime.exceptions.AbortCause;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.object.transactions.Transaction;
 import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.runtime.view.ObjectsView;
@@ -23,33 +27,22 @@ import org.corfudb.runtime.view.TableRegistry;
  *
  * Created by zlokhandwala on 2019-08-05.
  */
+@Slf4j
+@Builder
 public class TxBuilder {
-
+    @NonNull
     private final ObjectsView objectsView;
+    @NonNull
     private final TableRegistry tableRegistry;
+    @NonNull
     private final String namespace;
-    private Timestamp timestamp;
-    private final List<Runnable> operations;
-
-    /**
-     * Creates a new TxBuilder.
-     *
-     * @param objectsView   ObjectsView from the Corfu client.
-     * @param tableRegistry Table Registry.
-     * @param namespace     Namespace boundary defined for the transaction.
-     */
-    @Nonnull
-    TxBuilder(@Nonnull final ObjectsView objectsView,
-              @Nonnull final TableRegistry tableRegistry,
-              @Nonnull final String namespace) {
-        this.objectsView = objectsView;
-        this.tableRegistry = tableRegistry;
-        this.namespace = namespace;
-        this.operations = new ArrayList<>();
-    }
+    @Builder.Default
+    private final List<Runnable> operations = new ArrayList<>();
+    @Builder.Default
+    private final int numRetries = 3;
 
     private <K extends Message, V extends Message, M extends Message>
-    Table<K, V, M> getTable(@Nonnull final String tableName) {
+    Table<K, V, M> getTable(@NonNull final String tableName) {
         return this.tableRegistry.getTable(this.namespace, tableName);
     }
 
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -172,15 +175,27 @@ public void commit() {
...
+        if (numRetries <= 0) {
+            throw new IllegalArgumentException("TxBuilder::commit: numRetries <=0"+numRetries);
+        }
+        while (numRetries-- > 0) {
+            this.timestamp = timestamp;
+            try {
+                txBegin();
+                operations.forEach(Runnable::run);
+            } catch(TransactionAbortedException txAbort) {
+                if (txAbort.getAbortCause() == AbortCause.SEQUENCER_OVERFLOW) {
```
WenbinZhu: Looks like your implementation is retrying with the same timestamp, but in case of sequencer overflow, retrying with same timestamp would still fail right?

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -30,6 +32,7 @@
     private final String namespace;
     private Timestamp timestamp;
     private final List<Runnable> operations;
+    private static final int DEFAULT_NUM_TX_RETRIES = 3;
```
xnull: can you please make it not static? When it's not static we can make this parameter configurable, whet it's static it's hard coded

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -172,15 +175,27 @@ public void commit() {
...
      */
-    public void commit(final Timestamp timestamp) {
-        this.timestamp = timestamp;
-        try {
-            txBegin();
-            operations.forEach(Runnable::run);
-        } finally {
-            txEnd();
-            operations.clear();
+    public void commit(final Timestamp timestamp, int numRetries) {
```
xnull: if you make `DEFAULT_NUM_TX_RETRIES` non-static, then you don't need to pass  `numRetries` as a parameter.
And you can check that DEFAULT_NUM_TX_RETRIES is correct in TxBuilder's constructor. 
Which makes your code more resilient and secure

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -172,15 +175,27 @@ public void commit() {
...
+        if (numRetries <= 0) {
+            throw new IllegalArgumentException("TxBuilder::commit: numRetries <=0"+numRetries);
+        }
+        while (numRetries-- > 0) {
```
xnull: Please don't `shadow` numRetries variable. Use another variable, like `remainingRetries`

xnull: And also please don't reduce the variable and check the result at the same time. It's confusing and makes code more error prone

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -172,15 +175,27 @@ public void commit() {
...
+            this.timestamp = timestamp;
+            try {
+                txBegin();
+                operations.forEach(Runnable::run);
+            } catch(TransactionAbortedException txAbort) {
+                if (txAbort.getAbortCause() == AbortCause.SEQUENCER_OVERFLOW) {
+                    continue;
```
xnull: Please add a record in the log that the tx can't be finished

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -172,15 +175,27 @@ public void commit() {
...
+                }
+                throw txAbort;
+            } finally {
+                txEnd();
```
xnull: as I know `txEnd()` never used, we always use `txCommit()` or something like that. Please double-check

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -172,15 +175,27 @@ public void commit() {
...
+                operations.clear();
+                break;
```
xnull: if `finally` block is always get executed after you finish your operation, that means you will always exit after the first iteration and retry will never happen

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
hisundar: done. thanks

hisundar: done. will use separate variable.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
hisundar: the parameter is already configurable at runtime via an explicit argument right?
This is an internal value if the explicit argument is not set.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
hisundar: thanks for the catch!

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
hisundar: done. thanks

hisundar:
> Looks like your implementation is retrying with the same timestamp, but in case of sequencer overflow, retrying with same timestamp would still fail right?

Good point, I have modified it and added a check only for non-snapshot transactions.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
hisundar: done.

annym:


```diff

Issues
======
+ Solved 3
           

Complexity increasing per file
==============================
- runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651452&bid=15577763)

annym:


```diff

Issues
======
+ Solved 3
           

Complexity increasing per file
==============================
- runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/CorfuDB/CorfuDB/pullRequest?prid=4653424&bid=15584313)

PR #2220, opened 2019-11-05
[CorfuStore] API to return CorfuStoreEntries, keySet & clear.

zalokhan:
## Overview

Description:
Added API to return key with the payload on scan and filter and queries.
Added API to return the keyset.
Added API to clear a table.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStoreEntry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStoreEntry.java
@@ -1,0 +1,38 @@
+package org.corfudb.runtime.collections;
+
+import com.google.protobuf.Message;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+/**
+ * Encapsulates the key, value and metadata into one object.
+ * This is returned by the query APIs by the Corfu Store.
+ * Created by zlokhandwala on 11/4/19.
+ */
+@EqualsAndHashCode
+public class CorfuStoreEntry<K extends Message, V extends Message, M extends Message> {
+
+    /**
+     * Key of the entry of type K.
+     */
+    @Getter
+    private final K key;
+
+    /**
+     * Value of the entry of type V.
+     */
+    @Getter
+    private final V payload;
+
+    /**
+     * Metadata of the entry of type M.
+     */
+    @Getter
+    private final M metadata;
+
+    public CorfuStoreEntry(K key, V payload, M metadata) {
+        this.key = key;
+        this.payload = payload;
+        this.metadata = metadata;
+    }
+}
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuStoreEntry.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.protobuf.Message;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+/**
+ * Created by zlokhandwala on 11/4/19.
+ */
+@EqualsAndHashCode
+public class CorfuStoreEntry<K extends Message, V extends Message, M extends Message> {
```
zalokhan: Need to add comments.

runtime/src/main/java/org/corfudb/runtime/collections/Query.java
```diff
@@ -163,6 +163,20 @@ public int count(@Nonnull final String tableName,
         }
     }
 
+    @Nonnull
+    private <K extends Message, V extends Message, M extends Message>
+    List<CorfuStoreEntry<K, V, M>> scanAndFilterByEntry(
```
hisundar: Do we need scanAndFilter() above? Can we just have this method only?

zalokhan: Done.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -219,6 +213,21 @@ private void TxEnd() {
...
     }
 
+    /**
+     * Clears the table.
+     */
+    public void clear() {
+        boolean beganNewTxn = false;
+        try {
+            beganNewTxn = TxBegin();
+            corfuTable.clear();
```
pankti-m: we will clear the table even if beganNewTxn is false..

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -236,7 +254,29 @@ int count() {
...
+     */
+    @Nonnull
+    List<CorfuStoreEntry<K, V, M>> scanAndFilterByEntry(
+            @Nonnull final Predicate<CorfuStoreEntry<K, V, M>> entryPredicate) {
+        return corfuTable.scanAndFilterByEntry(recordEntry ->
+                entryPredicate.test(new CorfuStoreEntry<>(
+                        recordEntry.getKey(),
+                        recordEntry.getValue().getPayload(),
+                        recordEntry.getValue().getMetadata())))
+                .parallelStream()
```
medhavidhawan: Just make sure that we do not get in trouble with parallelStream as it uses common fork join pool.

zalokhan: These are not running long-running tasks but just wrapping the entry. Also, this `parallelStream` does not run the user's predicate and hence is bounded.

PR #2225, opened 2019-11-06
Closable bug (not for review)

vjeko:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -203,59 +203,44 @@ public int hashCode() {
...
+    public CorfuTable(IndexRegistry<K, V> indices,
+                      Supplier<StreamingMap<K, V>> streamingMapSupplier) {
         indices.forEach(index -> {
             secondaryIndexes.put(index.getName().get(), new HashMap<>());
             indexSpec.add(index);
         });
         log.info("CorfuTable: creating CorfuTable with the following indexes: {}",
                 secondaryIndexes.keySet());
-        mainMap = mapImpl;
+        mainMap = streamingMapSupplier.get();
```
xnull: why you need a supplier here?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -457,24 +442,24 @@ public void insert(@ConflictParameter K key, V value) {
...
      * @return a view of the values contained in this map meeting the predicate condition.
      */
     @Accessor
-    public List<V> scanAndFilter(Predicate<? super V> p) {
-        return mainMap.values().parallelStream()
-                                    .filter(p)
-                                    .collect(Collectors.toCollection(ArrayList::new));
+    public List<V> scanAndFilter(Predicate<? super V> valuePredicate) {
+        return mainMap.entryStream()
+                .map(Entry::getValue).filter(valuePredicate)
```
xnull: move .filter() on the next line

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -549,14 +534,16 @@ public void clear() {
     @Override
     @Accessor
     public @Nonnull Set<K> keySet() {
-        return ImmutableSet.copyOf(mainMap.keySet());
+        return mainMap.entryStream().map(Entry::getKey)
```
xnull: enter before .map

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,320 @@
...
+    private final CorfuRuntime corfuRuntime;
+    private final ISerializer serializer;
+    private final RocksDB rocksDb;
+
+    public PersistedStreamingMap(@NonNull Path dataPath,
+                                 @NonNull Options options,
+                                 @NonNull ISerializer serializer,
+                                 @NonNull CorfuRuntime corfuRuntime) {
+        try {
+            RocksDB.destroyDB(dataPath.toFile().getAbsolutePath(), options);
```
xnull: destroyingDb in a constructor - it seems like not the best idea.
I would prefer just to check that if the DB is already exists then throw an exception.

PersistedStreamingMap sitting on a different level, it just can't create/delete/manage database, 
it can't know what's going on. It can only check that the enviroment is in correct state or not

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Entry<K, V> next() {
+            if (hasNext()) {
+                current = next;
+                next = null;
```
xnull: I don't get the motivation for the `next` variable?
I see that we can get rid of `next` variable completely

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,320 @@
...
+    @Override
+    public boolean containsKey(@NonNull Object key) {
+        final ByteBuf keyPayload = Unpooled.buffer();
+        serializer.serialize(key, keyPayload);
+        try {
+            byte[] value = rocksDb.get(
+                    keyPayload.array(), keyPayload.arrayOffset(), keyPayload.readableBytes());
+            return value != null;
+        } catch (RocksDBException ex) {
+            throw new UnrecoverableCorfuError(ex);
```
xnull: again, why we always throw errors? It's a very bad practice

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Please use {@link StreamingMap#entryStream()}.
+     */
+    @Override
+    public boolean containsValue(@NonNull Object value) {
+        throw new UnsupportedOperationException();
```
xnull: if I'm not wrong rocks provides the opportunity to check that the value exists in the db
why not add that here?

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,320 @@
...
+            byte[] value = rocksDb.get(
+                    keyPayload.array(), keyPayload.arrayOffset(), keyPayload.readableBytes());
+            if (value == null) {
+                return null;
+            }
+            return (V) serializer.deserialize(Unpooled.wrappedBuffer(value), corfuRuntime);
+        } catch (RocksDBException ex) {
+            throw new UnrecoverableCorfuError(ex);
+        } finally {
+            keyPayload.release();
```
xnull: why you need to release keyPayload if it will be garbage collected after exiting from the method.
It affects performance

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,320 @@
...
+                    keyPayload.array(), keyPayload.arrayOffset(), keyPayload.readableBytes(),
+                    valuePayload.array(), valuePayload.arrayOffset(), valuePayload.readableBytes());
+        } catch (RocksDBException ex) {
+            throw new UnrecoverableCorfuError(ex);
+        } finally {
+            keyPayload.release();
+            valuePayload.release();
+        }
+
+        dataSetSize.incrementAndGet();
```
xnull: why you need atomic if the map is not threadsafe? Or if it is, then RocksDbIterator also should be thread-safe?

vjeko:
@xnull this patch is not meant for review. If you want to review this, please take a look at this PR: https://github.com/CorfuDB/CorfuDB/pull/2201

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -0,0 +1,320 @@
...
+    public void putAll(@NonNull Map<? extends K, ? extends V> map) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void clear() {
+        entryStream().map(Entry::getKey).forEach(this::remove);
```
xnull: can we just clean the database? In the other case, you need to traverse across entire db and delete record by record, wich is very slow

xnull:
:) cool, why you didn't explain this in the PR?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Why?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Why?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
vjeko: Because SMR objects can be recreated at any point in time -- in that case we want to create a new object, and not use the old reference. Take a look at @Maithem 's comment: https://github.com/CorfuDB/CorfuDB/pull/2201#discussion_r342355330

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
vjeko: The client does not understand the how the persistence is achieved. Asking clients to cleanup the resources after a hard-crash would constitute a leaky design.

PR #2227, opened 2019-11-07
Compression Codecs

Maithem:
## Overview
This patch adds compression capabilities to the StreamLogFiles. 

Why should this be merged: Can improve performance (LogUnit reboot/write/read) 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/LZ4Compression.java
```diff
@@ -0,0 +1,68 @@
...
+        checkArgument(compressed.remaining() > Integer.BYTES);
+
+        int decompressedSize = compressed.getInt();
+
+        ByteBuffer restored = ByteBuffer.allocate(decompressedSize);
+
+        int restoredBytes = decompressor.decompress(compressed, Integer.BYTES, restored, 0, decompressedSize);
+
+        if (restoredBytes != decompressedSize) {
+            // Why is this always off-by one?
```
medhavidhawan: what is this

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -421,9 +428,20 @@ private void trimPrefix() {
     }
 
     private LogData getLogData(LogEntry entry) {
-        ByteBuf data = Unpooled.wrappedBuffer(entry.getData().toByteArray());
+        // unnecessary alloc
+
```
medhavidhawan: comments please

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/LZ4Compression.java
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.ByteBuffer;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ *
+ * Created by Maithem on 11/6/19.
+ */
+public class LZ4Compression implements Codec {
```
medhavidhawan: comments please

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.function.Supplier;
+
+/**
+ *
+ * Created by Maithem on 11/6/19.
+ */
+
+public interface Codec {
```
medhavidhawan: comments

test/src/test/java/org/corfudb/runtime/clients/LogUnitHandlerTest.java
```diff
@@ -195,6 +195,7 @@ public void quotaExceededTest() throws Exception {
                 .setLogPath(dirPath)
                 .setServerRouter(serverRouter)
                 .setLogSizeLimitPercentage(Double.toString(maxLogSizeInPercentage))
```
medhavidhawan: should we add more targetted tests ?

test/src/test/java/org/corfudb/runtime/clients/LogUnitHandlerTest.java
```diff
@@ -195,6 +195,7 @@ public void quotaExceededTest() throws Exception {
...
+                .setCompressionCodec("None")
```
Maithem: Use enum directly.

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
```diff
@@ -46,6 +46,7 @@
     String handshakeTimeout = "10";
     String prefix = "";
     String retention = "1000";
+    String compressionCodec = "LZ4";
```
Maithem: Use enum directly.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/LZ4Compression.java
```diff
@@ -0,0 +1,70 @@
...
+
+        final int decompressedLength = uncompressed.remaining();
+
+        int maxCompressedLength = compressor.maxCompressedLength(decompressedLength);
+
+        ByteBuffer compressed = ByteBuffer.allocate(maxCompressedLength + Integer.BYTES);
+
+        int compressedLen = compressor.compress(uncompressed, 0, decompressedLength, compressed,
+                Integer.BYTES, maxCompressedLength);
+        compressed.putInt(0, decompressedLength);
```
Maithem: decompressedLength doesn't depend on the compression step, its clearer to set it up directly after the buffer allocation

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/LZ4Compression.java
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/LZ4Compression.java
Maithem: Need to investigate why the restored bytes are always off by one byte with respect to the original size.
Although the restored buffer is identical, the returned decompressed size is one byte smaller.

test/src/test/java/org/corfudb/runtime/clients/LogUnitHandlerTest.java
Maithem: Added more tests in StreamLogFilesTest

Maithem:
@medhavidhawan This patch is backward compatible with previous log formats, so there is no need for migration

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Supported codecs
+     */
+    enum Type {
```
xnull: it seems like a weird singleton, what's the profit? I see it as more complex solution than just using plain old simple singleton

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private static ZSTDCompression INSTANCE = new ZSTDCompression();
+
+    public static ZSTDCompression getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public ByteBuffer compress(ByteBuffer uncompressed) {
+        checkNotNull(uncompressed);
```
xnull: why not `Objects.requireNonNull()` ?

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
```diff
@@ -0,0 +1,67 @@
...
+import com.github.luben.zstd.Zstd;
+
+import java.nio.ByteBuffer;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ZSTDCompression implements Codec {
+
+    private static final int DEFAULT_COMPRESSION_LEVEL = 3;
```
xnull: can you make this parameter non-static? When a parameter is static it gets not configurable, it is hardcoded

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public ByteBuffer compress(ByteBuffer uncompressed) {
+        checkNotNull(uncompressed);
+        checkArgument(uncompressed.hasRemaining());
+
+        final int decompressedLength = uncompressed.remaining();
+        final int maxCompressedLength = (int) Zstd.compressBound(decompressedLength);
+
+        //TODO(Maithem): remove unpooled direct buffer allocations
```
xnull: please get rid of TODO.
P.S. never leave todos in code

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
```diff
@@ -0,0 +1,67 @@
...
+        return INSTANCE;
+    }
+
+    @Override
+    public ByteBuffer compress(ByteBuffer uncompressed) {
+        checkNotNull(uncompressed);
+        checkArgument(uncompressed.hasRemaining());
+
+        final int decompressedLength = uncompressed.remaining();
+        final int maxCompressedLength = (int) Zstd.compressBound(decompressedLength);
```
xnull: why not use long instead of int?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -128,6 +130,9 @@ public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
             throw new LogUnitException(msg);
         }
 
+        String codec = (String) serverContext.getServerConfig().get("--compression-codec");
+        compressionCodec = Codec.Type.valueOf(codec);
```
xnull: this way of using Codec.Type makes code less flexible. It comes harder to write a unit/benchmark test, in this case.
Please consider passing Codec.Type in the constructor instead of using singletons

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
```diff
@@ -46,6 +47,7 @@
     String handshakeTimeout = "10";
     String prefix = "";
     String retention = "1000";
+    String compressionCodec = Codec.Type.LZ4.toString();
```
xnull: why we need it here? Why not use an enum instead of a string?

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
Maithem: Its not. This serves as a mapping between int -> codec type which is later used when the LogData is persisted, so from a binary log prospective 1 -> None, 2 -> LZ4, 3 -> ZSTD.
It also serves as a sanity check, where no two codecs are defined with the same id (since enum prevents that).

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: Because its used to allocate an array which requires an int. The cast can be moved to where the allocation happens L:29

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: This is coming from docopt, so a conversion is needed at some point. I don't want to overload the constructor in this patch, but in a subsequent patch I'd like to introduce StreamLogConfig class.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: Bet.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: I can make it configurable, but it can't be used unless its piped all the way through the server configs, which I think is very messy. I will create a different patch to try to clean up the Config part. Based on a cursory survey, level 3 seems like a sane value.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: Bet.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
xnull: you just need an enum + a factory method somewhere, but right now your enum is a factory method for himself which makes code much more complex. Also, the lazy initialization doesn't work here,  which brings more complexity but in reality, there is no lazy initialization in this case.

As I said it can be simplified to:
 - enum `Type` contains only values
 - a factory method somewhere: 
   ```
    public Codec getByType(Type codecType) {...}
  ```

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
Maithem: Bet.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: LogUnitServerConfig already exists, in reality, the config is actually LUConfig + StreamLogConfig, so you can move the code into the config class

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
xnull: using a std library vs 3rd party library for the same behavior. You should always choose the std lib

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
xnull: It's not. It is already proven that using TODO is a bad practice that never works. You can see that check even in IntelliJ IDEA (it's enabled by default), it checks that there are no todos in the change.

What I can find on the internet for a couple of minutes http://wiki.c2.com/?TodoCommentsConsideredHarmful
For the exhaustive explanation/motivation read "Robert Martin clean code"

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
Maithem: > Also, the lazy initialization doesn't work here, which brings more complexity but in reality, there is no lazy initialization in this case.

I understand and I wasn't optimizing for that. I can do double checked locking, but it seems like an overkill. If you think that optimizing two very small objects away is crucial, then I'll do it, otherwise I opt for simplicity. 

> lazy initialization doesn't work here, which brings more complexity 

I'm a java neophyte, can you please explain how the lack of lazy initialization introduces complexity? and what type of complexity; Is it performance, efficiency, code organization, correctness,etc?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: Bet.

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
xnull: NO
It can break your system:
1. ServerContext is a god object
2. lack of incapsulation which can easily break your system or anyone who is not very careful can break your system by dynamically updating some parameter in the ServerContext

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: I believe I was misunderstood. Bet means for sure. Thanks for explanation.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: I believe I was misunderstood. Bet means for sure. Thanks for explanation.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: again? :)

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
xnull: oh man )) ok

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -421,9 +426,16 @@ private void trimPrefix() {
...
+
+        if (entry.hasCodecType() && entryData.hasRemaining()) {
+            Codec.Type codec = Codec.Type.values()[entry.getCodecType()];
+            entryData = codec.getInstance().decompress(entryData);
+        }
+
         LogData logData = new LogData(org.corfudb.protocols.wireprotocol
-                .DataType.typeMap.get((byte) entry.getDataType().getNumber()), data);
+                .DataType.typeMap.get((byte) entry.getDataType().getNumber()),
+                Unpooled.wrappedBuffer(entryData.array()));
```
WenbinZhu: I understand the necessity to convert between ByteBuf  and ByteBuffer here, but we should add a TODO to later to clean it up.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
xnull: have discussed offline

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -108,6 +108,9 @@
                     + " -R <retention>, --metadata-retention=<retention>                         "
                     + "              Maximum number of system reconfigurations (i.e. layouts)    "
                     + "retained for debugging purposes [default: 1000].\n"
+                    + " -C <codec>, --compression-codec=<codec>           "
+                    + "                        The type of compression the LogUnit to use to     "
+                    + "                        compress the binary log. [default: None].\n"
```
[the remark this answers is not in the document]
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public Codec getInstance() {
+            return func.get();
+        }
+    }
+
+    /**
+     * Returns the compressed ByteBuffer as a ByteBuffer with the size of
+     * the uncompressed buffer as the first 4-bytes.
```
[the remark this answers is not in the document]
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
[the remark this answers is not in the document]
Maithem: done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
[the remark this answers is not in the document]
Maithem: Added a comment.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: done

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/Codec.java
```diff
@@ -0,0 +1,46 @@
...
+public interface Codec {
+
+    /**
+     * Supported codecs
+     */
+    enum Type {
+        None(NoCompression::getInstance),
+        LZ4(LZ4Compression::getInstance),
+        ZSTD(ZSTDCompression::getInstance);
+        private Supplier<Codec> func;
```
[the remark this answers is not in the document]
Maithem: done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
Maithem: Ok/

infrastructure/src/main/java/org/corfudb/infrastructure/log/Compression/ZSTDCompression.java
```diff
@@ -0,0 +1,67 @@
...
+
+        //TODO(Maithem): remove unpooled direct buffer allocations
+        byte[] compressed = new byte[maxCompressedLength + Integer.BYTES];
+        ByteBuffer wrappedBuf = ByteBuffer.wrap(compressed);
+        wrappedBuf.putInt(decompressedLength);
+
+
```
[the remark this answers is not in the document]
Maithem: done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: Ostensibly its not good practice to add TODOs, but I have another patch that will eliminate some unnecessary buffer allocations and I'm planning to clean this up.

PR #2233, opened 2019-11-09
CorfuStore: Metadata retains previous values if not set

hisundar:
Some fields in managedResources are set only at the time the object is created and not really available on updates. This means clients would need to do an explicit get() to retrieve the old value.
This change makes the merging of previous value the default behavior for metadata fields.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
@@ -310,6 +310,27 @@ public class Table<K extends Message, V extends Message, M extends Message> {
             Descriptors.FieldDescriptor.Type.SFIXED64
     ));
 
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -274,7 +295,7 @@ int count() {
             }
             if (!versionTypes.contains(fieldDescriptor.getType())) {
                 throw new IllegalArgumentException("Version field needs to be an Integer or Long type."+
-                " Current type="+fieldDescriptor.getType());
+                        " Current type="+fieldDescriptor.getType());
```
xnull: please code reformat

hisundar: done. thanks

Issue #2235, opened 2019-11-14
LogData lastKnownSize is inconsistent in two scenarios.


WenbinZhu:
## Overview

LogData as a `lastKnownSize` which could be when during serialization or deserialization. During deserialization (`getPayload()`), it's updated to `data.length`, which is the length of the serialized payload. However during serialization, it's updated to `serializedCache.array().length` which is 
not the exact size of the buffer, and usually greater than it because array grows dynamically and ususally 2x the previous size.

PR #2236, opened 2019-11-14
GC: fix potential compactor issue that it may not have all serializers.

WenbinZhu:
# Overview

Log unit server might not have all serializers of different streams, but
the current compaction process deserializes all entries which could fail.
However compactor actually does not need to get the exact deserialized
content, so this patch adds a support to only deserialize the entry into
a list of byte arrays and re-assemble after compaction.

Related issue(s) (if applicable): #2235 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -28,6 +28,8 @@
 
     private ByteBuf serializedCache = null;
 
+    // last know serialized payload size.
+    @Getter
     private int lastKnownSize = NOT_KNOWN;
```
lixinchengdu: Add comments about this variable and also states your concern mentioned in the lasted issue.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecord.java
```diff
@@ -193,8 +212,33 @@ void serialize(ByteBuf b) {
...
+    }
+
+    /**
+     * Get the bytes of one {@link SMRRecord} from a larger byte buffer
+     * and transfer it to a new byte array.
+     *
+     * @param b byte buffer to transfer from
+     * @return a new byte array containing the transferred bytes
+     */
+    static byte[] transfer(ByteBuf b) {
```
lixinchengdu: This name is confusing. Probably toByte is a better name.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: There is a comment on this variable.
As for the concern in that issue, I addressed it in this PR. I will reference it in PR description.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMRRecord.java
WenbinZhu: Changed name to slice, as it's not only converting to bytes, it slices from a larger buffer.

PR #2237, opened 2019-11-15
Make (Multi)CheckpointWriter consume StreamingMap

vjeko:
In order for checkpointing to work with large (disk-backed) tables,
(Multi)CheckpointWriter needs to consume the streaming interface.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -15,6 +15,8 @@ import org.corfudb.protocols.wireprotocol.DataType;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.protocols.wireprotocol.Token;
 import org.corfudb.protocols.wireprotocol.TokenResponse;
+import org.corfudb.runtime.collections.ISMRMap;
+import org.corfudb.runtime.collections.StreamingMap;
 import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.runtime.view.CacheOption;
 import org.corfudb.runtime.view.StreamsView;
@@ -27,10 +29,10 @@ import java.time.LocalDateTime;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.UUID;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
+import java.util.stream.Stream;
 
 
 /** Checkpoint writer for SMRMaps: take a snapshot of the
@@ -40,7 +42,7 @@ import java.util.function.Function;
  *  TODO: Generalize to all SMR objects.
  */
 @Slf4j
-public class CheckpointWriter<T extends Map> {
+public class CheckpointWriter<T extends StreamingMap> {
     /** Metadata to be stored in the CP's 'dict' map.
      */
     private final UUID streamId;
@@ -60,9 +62,8 @@ public class CheckpointWriter<T extends Map> {
     private final Timer appendCheckpointTimer = metricRegistry.timer(CHECKPOINT_TIMER_NAME);
 
     @SuppressWarnings("checkstyle:abbreviation")
-    final UUID checkpointStreamID;
-
-    final Map<CheckpointEntry.CheckpointDictKey, String> mdkv = new HashMap<>();
+    private final UUID checkpointStreamID;
+    private final Map<CheckpointEntry.CheckpointDictKey, String> mdkv = new HashMap<>();
 
     /** Mutator lambda to change map key.  Typically used for
      *  testing but could also be used for type conversion, etc.
@@ -163,18 +164,18 @@ public class CheckpointWriter<T extends Map> {
             // A checkpoint writer will do two accesses one to obtain the object
             // vlo version and to get a shallow copy of the entry set
             log.info("appendCheckpoint: Started checkpoint for {} at snapshot {}", streamId, snapshotTimestamp);
-            Set<Map.Entry> entries = this.map.entrySet();
+            Stream<Map.Entry> entries = this.map.entryStream();
             // The vloVersion which will determine the checkpoint START_LOG_ADDRESS (last observed update for this
             // stream by the time of checkpointing) is defined by the stream's tail instead of the stream's version,
             // as the latter discards holes for resolution, hence if last address is a hole it would diverge
             // from the stream address space maintained by the sequencer.
             startCheckpoint(snapshotTimestamp, streamTail);
-            appendObjectState(entries);
+            int entryCount = appendObjectState(entries);
             finishCheckpoint();
             long cpDuration = System.currentTimeMillis() - start;
             log.info("appendCheckpoint: completed checkpoint for {}, entries({}), " +
                             "cpSize({}) bytes at snapshot {} in {} ms",
-                    streamId, entries.size(), numBytes, snapshotTimestamp, cpDuration);
+                    streamId, entryCount, numBytes, snapshotTimestamp, cpDuration);
         } finally {
             rt.getObjectsView().TXEnd();
         }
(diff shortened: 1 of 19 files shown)


test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
```diff
@@ -532,13 +533,13 @@ public void checkpointAndTrimAtDifferentPoint() throws Exception {
             final int sizeMap1 = 4;
             final int sizeMap2 = 6;
 
-            Map<String, String> map1 = defaultRT.getObjectsView().build()
+            StreamingMap<String, String> map1 = defaultRT.getObjectsView().build()
```
hisundar: these will involve about 168 changes in upstream since Map<> is used in those places.

vjeko: There is nothing forbidding you from using it as a Map. The only place where this change will be needed is wherever MultiCheckpointWriter is being consumed.

vjeko: In other words, the change in the contract happens here:

`public class MultiCheckpointWriter<T extends StreamingMap> `

vjeko: I checked the corfu-compactor logic, we consume `MultiCheckpointWriter` using `ICorfuMap`:

`MultiCheckpointWriter<ICorfuMap> mcWriter = new MultiCheckpointWriter<>();`

And `ICorfuMap` is defined as:

```
public interface ICorfuMap<K, V>
    extends StreamingMap<K, V> {
```

Thus, from what I can tell no changes should be necessary on the client side.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -40,7 +42,7 @@
  *  TODO: Generalize to all SMR objects.
  */
 @Slf4j
-public class CheckpointWriter<T extends Map> {
+public class CheckpointWriter<T extends StreamingMap> {
```
Maithem: StreamingMap is too low level to expose on a public API. I think `ICorfuSMR` is more appropriate, `ISMRMap`.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -163,18 +165,18 @@ public Token appendCheckpoint(Token snapshotTimestamp, Long streamTail) {
             // A checkpoint writer will do two accesses one to obtain the object
             // vlo version and to get a shallow copy of the entry set
             log.info("appendCheckpoint: Started checkpoint for {} at snapshot {}", streamId, snapshotTimestamp);
-            Set<Map.Entry> entries = this.map.entrySet();
+            Stream<Map.Entry> entries = this.map.entryStream();
```
Maithem: `ICorfuSMR` can expose a `getStateMethod`, or `ISMRMap` can expose `getEntryStream`

runtime/src/main/java/org/corfudb/runtime/collections/ICorfuMap.java
```diff
@@ -5,7 +5,7 @@
 import java.util.function.Predicate;
 
 public interface ICorfuMap<K, V>
-    extends Map<K, V> {
+    extends StreamingMap<K, V> {
```
Maithem: I think these should be reverse. `ICorfuMap` is the "more general" type.

runtime/src/main/java/org/corfudb/runtime/collections/SMRMap.java
```diff
@@ -31,7 +32,7 @@
 @SuppressWarnings("checkstyle:abbreviation")
 public class SMRMap<K, V>
         extends HashMap<K, V>
-        implements ISMRMap<K,V>, ICorfuSMR<SMRMap<K, V>> {
+        implements StreamingMap<K, V>, ISMRMap<K,V>, ICorfuSMR<SMRMap<K, V>> {
```
Maithem: The type relationships is already complex.

I think it would be cleaner to have ISMRMap extend ICorfuSMR. How can an object be SMR if it doesn't implement ICorfuSMR, all SMRObject types should extend/implement that type.

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -305,8 +306,8 @@ public static CorfuRuntime createRuntimeWithCache() {
         return rt;
     }
 
-    public static Map<String, Integer> createMap(CorfuRuntime rt, String streamName) {
-        Map<String, Integer> map = rt.getObjectsView()
+    public static StreamingMap<String, Integer> createMap(CorfuRuntime rt, String streamName) {
```
Maithem: Why is this needed?

test/src/test/java/org/corfudb/integration/LogSizeQuotaIT.java
```diff
@@ -118,7 +119,7 @@ public void testLogUnitQuota() throws Exception {
         privilegedRt.connect();
 
 
-        Map<String, String> map2 = privilegedRt.getObjectsView()
+        StreamingMap<String, String> map2 = privilegedRt.getObjectsView()
```
Maithem: Why is this needed?

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
Maithem: MultiCheckpointWriter shouldn't even use generics since we never need them.

I think its more appropriate to introduce new tests to test `StreamingMap` and not change existing ones.

Maithem:
I think its better to introduce new tests for the new functionality and not overload what's currently there.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
vjeko: Consumed via `ISMRMap` now.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
vjeko: Consumed via `ISMRMap` now.

runtime/src/main/java/org/corfudb/runtime/collections/SMRMap.java
vjeko: I was trying to avoid the diamond problem that arises due to `AutoClosable` interface. One solution would be to force all concrete implementation of `ICorfuSMR` to implement `close()`, but due those tricky annotation we have (`@PassThrough @DontInstrument`), I am hesitant to go down that path.

test/src/test/java/org/corfudb/integration/AbstractIT.java
vjeko: `Checkpointer` needs to consume an interface that exposes the streaming API, so `Map` will have to be replaced with something else. Most likely `ISMRMap`.

test/src/test/java/org/corfudb/integration/LogSizeQuotaIT.java
vjeko: Take a look at the previous comment: https://github.com/CorfuDB/CorfuDB/pull/2237#discussion_r351501861

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
vjeko: Turns out we cannot use `ISMRMap`, since this interface is only consumed only `SMRMap` and not `CorfuTable`. Making `CorfuTable` inherit from `ISMRMap` makes little sense as it contains bunch of logic only specific to `SMRMap`.

If we were to remove `SMRMap`, this change ought to be possible. For now I ensured that `StreamingMap` does not expose any Corfu specific internal -- it just contains `Stream<Map.Entry<K, V>> entryStream()`.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
vjeko: You look at the previous comment regarding `CorfuSMR`.

As for `ICorfuSMR` exposing `getStateMethod` -- I entertained that idea and I concluded that the complexity is not worth the effort. In order to achieve this, we would have to add and extra generic parameter to `ICorfuSMR` `E`, which in case of `CorfuTable` would be `Stream<Map.Entry<K, V>>`, but in other cases it (`CorfuCompoundObject`, `CorfuSharedCounter`) would not be a sequence. Also, after new GC comes in, all of this code is going to be irrelevant.

runtime/src/main/java/org/corfudb/runtime/collections/ICorfuMap.java
vjeko: `StreamingMap` is not a simplified interface.

test/src/test/java/org/corfudb/integration/StreamAddressDiscoveryIT.java
vjeko: You technically don't need the generic parameter. In this case it serves more as a type alias.

The tests have not been changed, just the upper bound type associated with the constructed objects.

PR #2239, opened 2019-11-19
Fix VM cloning in the Universe Framework

vjeko:
Leveraging a new template and also disable unnecessary
VM customizations.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/it/src/main/java/org/corfudb/universe/universe/vm/ApplianceManager.java
+++ b/it/src/main/java/org/corfudb/universe/universe/vm/ApplianceManager.java
@@ -176,14 +176,19 @@ public class ApplianceManager {
                 folderR.setVal(prop);
 
                 Folder folder = new Folder(vmTemplate.getServerConnection(), folderR);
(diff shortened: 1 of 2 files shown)


it/src/main/java/org/corfudb/universe/universe/vm/ApplianceManager.java
```diff
@@ -240,41 +245,6 @@ private VirtualMachineCloneSpec createLinuxCustomization(
         //Clone is powered on, not a template.
         vmCloneSpec.setPowerOn(true);
         vmCloneSpec.setTemplate(false);
-
```
xnull: why this has removed?

vjeko: Because it is not needed. These options are defined by the template itself. In fact, vSphere refuses to clone the template if customization option are set.

xnull: ok

PR #2243, opened 2019-11-21
DynamicProtobufSerializer: Debug message for serializer error

hisundar:
## Overview

If a table was opened with a different type but carries entries of different types, then we should be able to catch and log this error.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
+++ b/runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
@@ -151,6 +151,13 @@ public class DynamicProtobufSerializer implements ISerializer {
             return fileDescriptorMap.get(name);
         }
 
+        if (!fdProtoMap.containsKey(name)) {
+            log.error("DynamicProtobufSerializer failed to deserialize unknown file {}",
+                    name);
+            log.error(this.toString());
+            throw new SerializerException(
+                    "DynamicProtobufSerializer file " + name + " was never seen in registry");
+        }
         List<FileDescriptor> fileDescriptorList = new ArrayList<>();
         for (String s : fdProtoMap.get(name).getDependencyList()) {
             FileDescriptor descriptor = getDescriptor(s);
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -151,6 +151,9 @@ private FileDescriptor getDescriptor(String name) throws DescriptorValidationExc
             return fileDescriptorMap.get(name);
         }
 
+        if (!fdProtoMap.containsKey(name)) {
```
Maithem: Is it possible to collapse this with fileDescriptorMap?

It's also seems a little bit strange to call getFullyQualifiedTableName(CORFU_SYSTEM_NAMESPACE, REGISTRY_TABLE_NAME) inside the constructor I feel like its cleaner to open that map and extract the serializers from a higher layer.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -151,6 +151,9 @@ private FileDescriptor getDescriptor(String name) throws DescriptorValidationExc
...
+            throw new SerializerException("DynamicProtobuf type "+name+" was never seen in registry");
```
xnull: code reformat

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: I have been exploring this, looks like right there isn't an obvious way to directly convert from a type descriptor to a type. That is why we are having this extra layer of indirection to capture the File itself. It definitely needs more digging that a debugging patch.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -151,6 +151,11 @@ private FileDescriptor getDescriptor(String name) throws DescriptorValidationExc
             return fileDescriptorMap.get(name);
         }
 
+        if (!fdProtoMap.containsKey(name)) {
+            log.error(this.toString());
```
Maithem: Can you add a log error message here, maybe log.error("cannot deserialize object {}", this, exception)

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -284,5 +291,26 @@ public void serialize(Object o, ByteBuf b) {
             throw new SerializerException(ie);
         }
     }
+
+    public String toString() {
+        StringBuilder stringBuilder = new StringBuilder();
```
xnull: it will be very hard to parse these kinds of messages in log aggregation platforms like ELK.
Can you use one string for that?

hisundar: hi @xnull sorry I missed this comment before merge. I would like to address it. Could you please share details on what is the ELK format for strings?

xnull: It's just a best practice, never use more than one line in the log. In the other case, you lose the opportunity to parse your logs as a whole thing

PR #2244, opened 2019-11-21
AddressSpaceView cache fix

vjeko:
When the cache is disabled, entries still end up being cached due to a
bug in AddressSpaceView. This patch fixes that issue and simplifies the
logic.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -7,6 +7,7 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.RemovalNotification;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.netty.handler.timeout.TimeoutException;
 import lombok.NonNull;
@@ -41,16 +42,16 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.NavigableSet;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 
 /**
@@ -61,15 +62,14 @@ import java.util.concurrent.TimeUnit;
 @Slf4j
 public class AddressSpaceView extends AbstractView {
 
+    private final static long CACHE_KEY_SIZE = MetricsUtils.sizeOf.deepSizeOf(0L);
+    private final static long DEFAULT_MAX_CACHE_ENTRIES = 5000;
+    private final static boolean NO_THROW = false;
+
     /**
      * A cache for read results.
      */
-    final Cache<Long, ILogData> readCache;
-
-    final private long cacheKeySize = MetricsUtils.sizeOf.deepSizeOf(new Long(0));
-
-    final private long defaultMaxCacheEntries = 5000;
-
+    private final Cache<Long, ILogData> readCache;
     private final ReadOptions defaultReadOptions = ReadOptions.builder()
             .ignoreTrim(false)
             .waitForHole(true)
@@ -83,28 +83,29 @@ public class AddressSpaceView extends AbstractView {
     public AddressSpaceView(@Nonnull final CorfuRuntime runtime) {
         super(runtime);
 
-        CacheBuilder cacheBuilder = CacheBuilder.newBuilder();
+        CacheBuilder<Object, Object> cacheBuilder = CacheBuilder.newBuilder();
 
-        long maxCacheEntries = runtime.getParameters().getMaxCacheEntries();
-        long maxCacheWeight = runtime.getParameters().getMaxCacheWeight();
-
-        if (maxCacheEntries != 0) {
-            cacheBuilder.maximumSize(runtime.getParameters().getMaxCacheEntries());
-        }
+        final boolean cacheDisabled = runtime.getParameters().isCacheDisabled();
+        final long maxCacheEntries = runtime.getParameters().getMaxCacheEntries();
+        final long maxCacheWeight = runtime.getParameters().getMaxCacheWeight();
+        final int concurrencyLevel = runtime.getParameters().getCacheConcurrencyLevel();
 
         if (maxCacheWeight != 0) {
-            cacheBuilder.maximumWeight(runtime.getParameters().getMaxCacheWeight());
-            cacheBuilder.weigher((k, v) -> (int) (cacheKeySize + MetricsUtils.sizeOf.deepSizeOf(v)));
+            cacheBuilder.maximumWeight(maxCacheWeight);
+            cacheBuilder.weigher((k, v) -> (int) (CACHE_KEY_SIZE + MetricsUtils.sizeOf.deepSizeOf(v)));
         }
 
-        if (maxCacheEntries == 0 && maxCacheWeight == 0) {
-            // If cache weight/size are not set, then we default to using
-            // size based cache
-            cacheBuilder.maximumSize(defaultMaxCacheEntries);
+        if (cacheDisabled) {
+            cacheBuilder.maximumSize(0); // Do not allocate memory when cache is disabled.
+        } else if (maxCacheEntries != 0) {
+            cacheBuilder.maximumSize(maxCacheEntries);
+        } else if (maxCacheWeight == 0) {
+            // If cache weight/size are not set, then we default to using size based cache.
+            cacheBuilder.maximumSize(DEFAULT_MAX_CACHE_ENTRIES);
         }
 
-        if (runtime.getParameters().getCacheConcurrencyLevel() != 0) {
-            cacheBuilder.concurrencyLevel(runtime.getParameters().getCacheConcurrencyLevel());
+        if (concurrencyLevel != 0) {
+            cacheBuilder.concurrencyLevel(concurrencyLevel);
         }
 
         readCache = cacheBuilder.expireAfterAccess(runtime.getParameters().getCacheExpiryTime(), TimeUnit.SECONDS)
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -85,26 +85,28 @@ public AddressSpaceView(@Nonnull final CorfuRuntime runtime) {
 
         CacheBuilder cacheBuilder = CacheBuilder.newBuilder();
 
-        long maxCacheEntries = runtime.getParameters().getMaxCacheEntries();
-        long maxCacheWeight = runtime.getParameters().getMaxCacheWeight();
-
-        if (maxCacheEntries != 0) {
-            cacheBuilder.maximumSize(runtime.getParameters().getMaxCacheEntries());
-        }
+        final boolean cacheDisabled = runtime.getParameters().isCacheDisabled();
```
xnull: private

Maithem:
> Cache that is disabled will still allocate segments based on the
> defaultMaxCacheEntries. This causes unnecessary memory bloat -- about
> 300MB at the time of this patch.

Hmm, this doesn't sound correct: a quick experiment shows otherwise. Here's a heap dump of a cache with the size set to 5,000,000 (i.e. 1000x the default size):

"Name","Retained Size","Shallow Size","Level"
"com.google.common.cache.LocalCache","**2536**","128","1"
"maxWeight = long **5,000,000**  0x00000000004C4B40","","8","2"

The retained size and max cache size are highlighted in bold respectively. 

From Guava's point-of-view it actually doesn't make sense to pre-allocate the segments simply because they can't compute the size of each segment ahead of time. Guava implements both the weighted and size based eviction policies on top of weight based eviction. What this means is the size based cache just uses a weigher function that always returns 1. 

Why can't it pre-allocate for a weight base cache?

Since the weigher function is a lambda, the weigher can partition the max weight in any arbitrary manner which makes "pre-allocating a segment" unfeasible because the number of slots per segment can't be efficiently determined. 

Guava will create an array for segment locks (based on concurrency level) and each segment will have a map collection that can grow dynamically constrained by max segment weight which is max cache weight divided by the concurrency level. 

Its possible that your client code or the read/write methods in the `AddressSpaceView` has a bug that's causing the bloat.

vjeko:
I did not spend much time trying to figure out how the cache ends up getting populated.  Here is what I did notice:

After running the check-pointer over night and despite the cache being disabled, the total memory consumption for the cache ends up being **4GB**:

cache-leak

When the number of max entries is set to 0, the total heap consumption goes down to `50MB`.

There is **only** one place in the AddressSpaceView where read cache gets populated:

```
        // Cache the successful write
        if (!runtime.getParameters().isCacheDisabled() && cacheOption == CacheOption.WRITE_THROUGH) {
            //readCache.put(token.getSequence(), ld);
        }
```
Even after commenting out `readCache.put(...)` the memory consumption does not go down. This is why I concluded this is somehow related to read cache initialization. It could also be the case that read/invalidate operations have side effects.

There is something really fishy happening here.

vjeko:
Alright, there seems to be a bug in `ILogData cacheLoadAndGet(Cache<Long, ILogData> cache, long address, ILogData loadedValue)`.

I am looking into a fix.

vjeko:
From the documentation:

```
  /**
   * Returns the value associated with {@code key} in this cache, obtaining that value from {@code
   * loader} if necessary. The method improves upon the conventional "if cached, return; otherwise
   * create, cache and return" pattern. For further improvements, use {@link LoadingCache} and its
   * {@link LoadingCache#get(Object) get(K)} method instead of this one.
```

So look like `return cache.get(address, () -> loadedValue)` will actually cache the value.

Maithem:
> From the documentation:
> 
> ```
>   /**
>    * Returns the value associated with {@code key} in this cache, obtaining that value from {@code
>    * loader} if necessary. The method improves upon the conventional "if cached, return; otherwise
>    * create, cache and return" pattern. For further improvements, use {@link LoadingCache} and its
>    * {@link LoadingCache#get(Object) get(K)} method instead of this one.
> ```
> 
> So look like `return cache.get(address, () -> loadedValue)` will actually cache the value.

It seems like when some of the patches that went in to fix the checkpoint stream trimmed exception issues this method was introduced `checkForTrimmedAddresses`. It invokes the cache even though it shouldn't, since the verification should be post read/load. Essentially, after the reads bypass the cache, they get re-populated when `checkForTrimmedAddresses` is called and this code path is being invoked from `AbstractQueuedStreamView::resolveCheckpoint`.

If you look here: https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java#L659  there shouldn't be any cache related logic since it should just check for trimmed exceptions.

I suggest you remove `cacheLoadAndGet` from `checkForTrimmedAddresses`. When cacheLoadAndGet was introduced it was only meant to be used in public `read` methods and not helper functions. Also, note that `ReadOptions::isClientCacheable` is not sufficient to make a decision on whether to cache or not, and `AddressSpaceView::cacheReadRequest` should be used instead.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -420,34 +412,33 @@ private ILogData cacheLoadAndGet(Cache<Long, ILogData> cache, long address, ILog
      * @return A map of addresses read, which will be cached if caching is enabled
      */
     public Map<Long, ILogData> read(Iterable<Long> addresses, @NonNull ReadOptions options) {
-        // We need this map to preserve ordering as entries are expected in ascending order.
-        Map<Long, ILogData> result = new HashMap<>();
+        if (!options.isClientCacheable()) {
+            return fetchAll(addresses, options);
```
annym: In the refactoring we lost the check for trimmed addresses, meaning that, in the event of cache disable we won't throw trimmed exceptions if any address was trimmed.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -332,20 +320,20 @@ private boolean cacheReadRequest(ReadOptions options) {
...
+     * @param addresses batch of addresses to read (bring into the cache) in case
+     *                  there is a cache miss (includes nextRead)
      * @param options  options for this read request
      * @return data for current 'address' of interest.
      */
     public @Nonnull
     ILogData read(@NonNull Long nextRead, @NonNull NavigableSet<Long> addresses,
                   @NonNull ReadOptions options) {
-        if (cacheReadRequest(options)) {
+        if (options.isClientCacheable()) {
```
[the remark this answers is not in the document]
vjeko: This option is only used in the constructor. if the cache is disabled, we simple reduce the cache size to zero. No special path is required in that case.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -235,7 +236,7 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
         }, true);
 
         // Cache the successful write
-        if (!runtime.getParameters().isCacheDisabled() && cacheOption == CacheOption.WRITE_THROUGH) {
+        if (cacheOption == CacheOption.WRITE_THROUGH) {
```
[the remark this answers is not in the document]
vjeko: Same logic applies here as in this comment: https://github.com/CorfuDB/CorfuDB/pull/2244/files/1d4b61b1aaff032d990fc6563d9d8fa27f7ab1dd#r353488105

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
vjeko: Good catch! Thanks!

vjeko: Turns out that this if statement is not necessary. We already have a check in `cacheLoadAndGet`:

```
        if (!options.isClientCacheable()) {
            return loadedValue;
        }
```

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -614,64 +601,43 @@ public void invalidateClientCache() {
...
-        if (!trimmedAddresses.isEmpty()) {
-            if (!options.isIgnoreTrim()) {
-                throw new TrimmedException(trimmedAddresses);
-            }
-
-            log.warn("read: ignoring trimmed addresses {}", trimmedAddresses);
-        }
-
-        return result;
+    private List<Long> filterTrimmedAddresses(Map<Long, ILogData> allData, ReadOptions options) {
```
annym: Codacy Issue found: [Avoid unused method parameters such as 'options'.](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651449)

annym:


```diff

Issues
======
+ Solved 1
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651449&bid=15577759)

annym:


```diff

Issues
======
+ Solved 1
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651449&bid=15577759)

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: Codacy Issue found: [Avoid unused method parameters such as 'options'.](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651449)

annym:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651449&bid=15577759)

annym:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/CorfuDB/CorfuDB/pullRequest?prid=4653421&bid=15584309)

Maithem:
This patch didn't seem to work when @hisundar tried it. @hisundar maybe you can comment on what you saw.

vjeko:
I ran a unit test before merging to verify the functionality. The patch is working as expected -- cache size remains at zero.

PR #2246, opened 2019-11-23
Fix garbage informer bugs

lixinchengdu:
## Overview
There are two bugs in garbage informer:

1. writeGarbage does not wait on the future.

2. gcUnsafe is called in multiple threads.


runtime/src/main/java/org/corfudb/runtime/GarbageInformer.java
```diff
@@ -156,10 +160,24 @@ public void submitGCTask() {
         }
     }
 
+    @VisibleForTesting
+    public void waitUntilAllTasksFinish() {
+        // waits until all task in the executor finish
+        drainExecutor.shutdown();
+        try {
+            if (!drainExecutor.awaitTermination(TERMINATION_WAIT_TIME.getSeconds(), TimeUnit.SECONDS)) {
+                drainExecutor.shutdownNow();
```
WenbinZhu: You can put ShutdownNow() to the `finally` block.

test/src/test/java/org/corfudb/runtime/view/AbstractViewTest.java
```diff
@@ -524,8 +524,13 @@ protected void startCompaction(CorfuRuntime rt, LogUnitServer logUnitServer) {
...
         rt.getGarbageInformer().stop();
 
+        // wait until all garbage decisions are sent to logUnit servers.
+        rt.getGarbageInformer().waitUntilAllTasksFinish();
+
         // send garbage decisions to logUnit servers
-        rt.getGarbageInformer().gcUnsafe();
+        while (rt.getGarbageInformer().getGarbageReceivingQueue().size() > 0) {
+            rt.getGarbageInformer().gcUnsafe();
+        }
```
WenbinZhu: Put this logic into GarbageInformer together with waitUntilAllTasksFinish  as we discussed.

WenbinZhu:
@lixinchengdu The unit tests are still failing, did you push Friday's change?

runtime/src/main/java/org/corfudb/runtime/GarbageInformer.java
lixinchengdu: done.

test/src/test/java/org/corfudb/runtime/view/AbstractViewTest.java
lixinchengdu: done.

test/src/test/java/org/corfudb/runtime/object/transactions/SnapshotTransactionContextTest.java
```diff
@@ -136,23 +137,30 @@ public void snapshotReadBeforeCompactionMark() {
...
                 .assertThrows().hasCauseInstanceOf(TrimmedException.class);
+        t(2, () -> assertThat(getRuntime().getAddressSpaceView().getCompactionMark().get())
+                .isEqualTo(RECORDS_PER_SEGMENT));
         t(2, this::TXEnd);
     }
 
     @Test
     public void snapshotReadAfterCompactionMark() {
+        t(2, this::SnapshotTXBegin);
+        t(2, () -> get("k2"));
```
WenbinZhu: You need a put before get right?

runtime/src/main/java/org/corfudb/runtime/GarbageInformer.java
```diff
@@ -156,10 +160,28 @@ public void submitGCTask() {
         }
     }
 
+    @VisibleForTesting
+    public void waitUntilAllTasksFinish() {
+        // send garbage decisions to logUnit servers
+        while (getGarbageReceivingQueue().size() > 0) {
```
WenbinZhu: This should be after awaitTermination.

runtime/src/main/java/org/corfudb/runtime/GarbageInformer.java
```diff
@@ -156,10 +160,28 @@ public void submitGCTask() {
...
+        
+        // waits until all task in the executor finish
+        drainExecutor.shutdown();
+        try {
+            drainExecutor.awaitTermination(TERMINATION_WAIT_TIME.getSeconds(), TimeUnit.SECONDS);
+        } catch (InterruptedException ex) {
+            log.error("Encounter interruption exception when {} is await termination", drainExecutor, ex);
+            throw new UnrecoverableCorfuInterruptedError(ex);
+        } finally {
+            drainExecutor.shutdownNow();
```
WenbinZhu: You need to throw any exception happened in this method, I think add `throw Exception` in the method signature and remove shutdownNow would be better.

runtime/src/main/java/org/corfudb/runtime/GarbageInformer.java
```diff
@@ -36,12 +37,13 @@
     private final static int CORE_POOL_SIZE = 1;
     private final static int MAX_POOL_SIZE = 1;
     private final static long KEEP_ALIVE_TIME = 0L;
+    private final static Duration TERMINATION_WAIT_TIME = Duration.ofHours(1);
```
WenbinZhu: Please also change the receiving and sending queue size as we discussed.

runtime/src/main/java/org/corfudb/runtime/GarbageInformer.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.*;
```
WenbinZhu: Expand

WenbinZhu:
Superseded by #2249

PR #2261, opened 2019-12-06
Streaming API needs to operate on a snapshot

vjeko:
When multiple threads are operating on the same table, and one of them
is utilizing the streaming API, ConcurrentModificationException will
occur since the Stream was operating on a shared reference.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/ImmutableListSetWrapper.java
+++ b/runtime/src/main/java/org/corfudb/util/ImmutableListSetWrapper.java
@@ -1,9 +1,16 @@
 package org.corfudb.util;
 
+import com.google.common.collect.ImmutableList;
+import lombok.NonNull;
+
 import javax.annotation.Nonnull;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.AbstractSet;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/util/ImmutableListSetWrapper.java
```diff
@@ -14,13 +21,17 @@
...
  * too expensive.
- *
  */
+public class ImmutableListSetWrapper<E> extends AbstractSet<E> implements Set<E> {
+    private final List<E> internalList;
 
-public class ImmuableListSetWrapper<E> implements Set<E> {
-    final List<E> internalList;
+    public static <K, V> ImmutableListSetWrapper<Entry<K, V>> fromMap(@NonNull Map<K, V> original) {
+        return new ImmutableListSetWrapper<>(original.entrySet().stream().map(entry ->
```
xnull: Please reformat code to something like
```
ImmutableList<SimpleImmutableEntry<Integer, Integer>> data = original
                .entrySet()
                .stream()
                .map(entry -> new SimpleImmutableEntry<>(entry.getKey(), entry.getValue()))
                .collect(ImmutableList.toImmutableList());

        return new ImmuableListSetWrapper<>(data);
```

current version is not quite well readable

vjeko: Good point, fixed.

PR #2265, opened 2019-12-08
Dead Code

Maithem:
## Overview
Removed dead code.

Why should this be merged: Dead code. 

- [x] Public API has Javadoc


common/src/main/java/org/corfudb/common/result/Result.java
```diff
@@ -126,13 +126,4 @@ public void ifError(Consumer<E> consumer) {
             return new Result<>(null, error);
         }
     }
-
-    public<U> Result<U, E> flatMap(Function<T, Result<U, E>> mapper) {
```
xnull: this code is requred

Maithem: where is it being used?

From `Supplier.java`

```java

@FunctionalInterface
public interface Supplier<T> {

    /**
     * Gets a result.
     *
     * @return a result
     */
    T get();
} ```

xnull: It's not used yet, but it's a part of `Result` API and should not be deleted

Maithem: ok

Maithem: Done.

PR #2269, opened 2019-12-09
Code Coverage

annym:


annym:
> Good change.
> 
> I think the reports are generated per module, is there another plugin to aggregate the reports?

Yes, I'm trying to use jacoco report-aggregate. Let me try pushing some other changes to get it working properly. I'll remove the 'Do not review' once its working.

codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  precision: 2
+  round: nearest
+  range: "40...70"
+  ignore:
+    - "generator/src/main/java/**/*"
+    - "samples/**/*"
```
vjeko: Newline.

PR #2270, opened 2019-12-10
Layout Committed Bug

Maithem:
## Overview
LayoutView::committed can return without collecting all responses.
This misuse of CompletableFuture.anyOf can cause the method to
return without waiting for other ongoing requests even if they
don't fail.

Why should this be merged: fixes a bug

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -295,11 +295,12 @@ public void committed(long epoch, Layout layout, boolean force)
...
         int timeouts = 0;
         int responses = 0;
-        while (responses < commitList.length) {
-            // wait for someone to complete.
+
+        for (CompletableFuture cf : commitList) {
             try {
-                CFUtils.getUninterruptibly(CompletableFuture.anyOf(commitList),
-                        WrongEpochException.class, TimeoutException.class, NetworkException.class);
+                CFUtils.getUninterruptibly(cf, WrongEpochException.class,
```
Maithem: remove CFUtils usage and check other uses of anyOf in this file.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -295,11 +295,12 @@ public void committed(long epoch, Layout layout, boolean force)
 
         int timeouts = 0;
         int responses = 0;
-        while (responses < commitList.length) {
```
vjeko: I am failing to see the bug. You will be iterating while (responses < commitList.length). `CompletableFuture.anyOf()` will be invoked for all CF which do not execute exceptionally.

annym:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651435&bid=15577744)

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
Maithem: On every iteration only one future can bump up the (via CompletableFuture.anyOf) responses counter.
The response counter should be incremented once per future, but one cf can increment it multiple times, which is the bug.

vjeko: I see.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
medhavidhawan: This seems to have the issue that it will not break out to check the next future till the current one in the loop is complete.
This make things timeout if the nodes early in the loop are slow

Maithem: Are you referring to committed or other methods?

The wait time is bounded by the RPC timeout, not RPC timeout * number of requests, since the requests are pipelined

annym:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651435&bid=15577744)

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
medhavidhawan: The reason i bring this up is that the protocol does not need a success from all nodes for committed. This is more stringent than what is needed for correctness.

Maithem: This change doesn't change the semantics, it just fixes a logic bug.

We wait on all futures to complete, not on all servers to respond. The router will timeout requests, so the max time is bounded by the RPC timeout of a single request.

annym:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/CorfuDB/CorfuDB/pullRequest?prid=4653408&bid=15584295)

annym:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/CorfuDB/CorfuDB/pullRequest?prid=4653408&bid=15584295)

annym:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4651435&bid=15577744)

Maithem:
> The `propose()` method in this file is also using `anyOf()`. That logic might be correct (not sure), but I think not efficient: it's wasting a lot of while cycles since `CFUtils.getUninterruptibly(CompletableFuture.anyOf(proposeList)` would always return immediately after the first succeeded future as it's never removed from `proposeList`.
> Can you also change that part?

I think there are more serious problems than performance. I'm extremely surprised all this kinda "works", I will rewrite and introduce more testing for prepare and propose.

PR #2278, opened 2019-12-12
CorfuStore: openTable must let caller know if table exists

hisundar:
Callers wish to know if a table exists in the system and never
create it in the process of checking.
To enable such an usecase, they use the openTable() short form
which returns an error of NoSuchElement if that CorfuRuntime
has never ever seen this table.
However this does not tell the caller if the table exists in the
cluster.
So query the TableRegistry & find out if the table exists first
and return a different error which will let caller know that the
table they wish to open does exist in the system.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -350,8 +351,20 @@ private Options getPersistentMapOptions() {
     Table<K, V, M> getTable(String namespace, String tableName) {
         String fullyQualifiedTableName = getFullyQualifiedTableName(namespace, tableName);
         if (!tableMap.containsKey(fullyQualifiedTableName)) {
-            throw new NoSuchElementException(
-                    String.format("No such table found: namespace: %s, tableName: %s", namespace, tableName));
+            // Table has not been opened, but let's first find out if this table even exists
+            // in the corfu cluster
+            UUID tableStreamId = CorfuRuntime.getStreamID(fullyQualifiedTableName);
+            long tableStreamTail = runtime.getSequencerView().query(tableStreamId);
+            if (tableStreamTail != Address.NON_EXIST) {
```
Maithem: The checkpointer can bump the tail even if the table is empty, so not sure if this check is reliable.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -350,8 +351,20 @@ private Options getPersistentMapOptions() {
...
+                // If table does exist then the caller must use the long form of the openTable()
+                // since there are too few arguments to open a table not seen by this runtime.
+                throw new IllegalArgumentException("Insufficient parameters to re-open existing table "
```
annym: should this message be a bit more informative? as it states "unable to 'reopen' existing table" it does not provide a notion that this table has never been opened, instead it gives the idea that it has been opened but as of now we can't reopen.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
Maithem: I think the consumer might be more interested in the life cycle of the table, so an explicit register operation from the application and not the checkpointer (since that's just a consequence of how checkpointing is done today, but it really shouldnt be that way).

hisundar: The goal is to just distinguish between a table that was never created from the case where a table was created. Empty table is also a valid table. Checkpointer cannot checkpoint a table that has never been created and that's all the caller needs to worry about for this change.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Right now the caller is just polling on the Exception returned by this api - IllegalArgumentException vs NoSuchElementException. But sure should I instead say please provide Key, Value and Metadata schemas to re-open the existing table?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Lifecycle management is outside the scope of this bug fix.

Maithem: Whether you want to implement the lifecycle of the stream or not, this patch doesn't seem to solve the issue that you are trying to fix. The application itself cannot differentiate between whether a table was created by the application, or the checkpointer. Before the application opens the table, the checkpointer can write an empty checkpoint for that stream and `tableStreamTail != Address.NON_EXIST` will always be false.

hisundar: If the checkpointer is able to see a table it means that the table has been created in which case this api would correctly return that the table exists. Note that the caller is only interested in checking if a table exists not if it is empty. So why won't this patch work?

hisundar:
Hi @Maithem , @annym please note this patch is about the existence of a table. Your comments indicate you are referring to a table being empty which is different. The bug fix is to just allow caller to check for the existence of a table and nothing more.

annym:
> Hi @Maithem , @annym please note this patch is about the existence of a table. Your comments indicate you are referring to a table being empty which is different. The bug fix is to just allow caller to check for the existence of a table and nothing more.

@hisundar I see. So regardless if it has been written to or if it's empty, we don't care. This is just telling if it exists or not, with no info of being written or not.
Now, one question. If a table was created by another client and never written to, will the sequencer know about it? cause in that case if we query before the checkpointer runs it will get as tail NON_ADDRESS (meaning it will tell it does not exist, when it really does), but after the checkpointer runs it will get some valid address (telling it does exist). So it's not consistent.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -350,8 +351,20 @@ private Options getPersistentMapOptions() {
...
+            // Table has not been opened, but let's first find out if this table even exists
+            // in the corfu cluster
+            UUID tableStreamId = CorfuRuntime.getStreamID(fullyQualifiedTableName);
+            long tableStreamTail = runtime.getSequencerView().query(tableStreamId);
+            if (tableStreamTail != Address.NON_EXIST) {
+                // If table does exist then the caller must use the long form of the openTable()
+                // since there are too few arguments to open a table not seen by this runtime.
+                throw new IllegalArgumentException("Please provide Key, Value & Metadata schemas to re-open"
+                + " this existing table " + tableName + " in namespace " + namespace);
+            } else { // If the table is completely unheard of return NoSuchElementException
```
pankti-m: nit - comment in new line

hisundar:
> > Hi @Maithem , @annym please note this patch is about the existence of a table. Your comments indicate you are referring to a table being empty which is different. The bug fix is to just allow caller to check for the existence of a table and nothing more.
> 
> @hisundar I see. So regardless if it has been written to or if it's empty, we don't care. This is just telling if it exists or not, with no info of being written or not.
> Now, one question. If a table was created by another client and never written to, will the sequencer know about it? cause in that case if we query before the checkpointer runs it will get as tail NON_ADDRESS (meaning it will tell it does not exist, when it really does), but after the checkpointer runs it will get some valid address (telling it does exist). So it's not consistent.

Yes this is a good point although hypothetical - we do not have a use case that opens a table but not write anything to it. Let me see if there is another way to address this case nevertheless.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: done.

hisundar:
> > Hi @Maithem , @annym please note this patch is about the existence of a table. Your comments indicate you are referring to a table being empty which is different. The bug fix is to just allow caller to check for the existence of a table and nothing more.
> 
> @hisundar I see. So regardless if it has been written to or if it's empty, we don't care. This is just telling if it exists or not, with no info of being written or not.
> Now, one question. If a table was created by another client and never written to, will the sequencer know about it? cause in that case if we query before the checkpointer runs it will get as tail NON_ADDRESS (meaning it will tell it does not exist, when it really does), but after the checkpointer runs it will get some valid address (telling it does exist). So it's not consistent.

hi @annym - I have modified the change to take care of the case, where even an empty table can be detected as long as the openTable call successfully completes. Could you please help take a look again? thanks

PR #2281, opened 2019-12-13
Transaction Poller Bug

Maithem:
## Overview
Every time the transaction poller polls the transaction stream
it will create a new StreamView object, which leaks stream
objects.

Why should this be merged: Fixes a memory leak

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
@@ -35,16 +35,29 @@ public class TransactionPoller implements Runnable {
     private final List<StreamingSubscriptionContext> streamContexts;
 
     /**
+     * A reference to the transaction stream
+     */
+    private final IStreamView txnStream;
+
+    /**
      * Constructor.
      *
      * @param streams The list of StreamingSubscriptionContexts to process.
      */
     public TransactionPoller(@Nonnull CorfuRuntime runtime,
-            @Nonnull List<StreamingSubscriptionContext> streams) {
+                             @Nonnull List<StreamingSubscriptionContext> streams) {
         this.runtime = runtime;
         this.streamContexts = streams.stream()
                 .sorted(Comparator.comparingLong(sc -> sc.getLastReadAddress()))
                 .collect(Collectors.toList());
+
+        StreamOptions options = StreamOptions.builder()
+                .ignoreTrimmed(true)
+                .cacheEntries(false)
+                .build();
+
+        txnStream = runtime.getStreamsView()
+                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
     }
 
     /**
@@ -69,18 +82,12 @@ public class TransactionPoller implements Runnable {
  
         long lastReadAddress = streamContexts.get(0).getLastReadAddress();
 
-        StreamOptions options = StreamOptions.builder()
-                .ignoreTrimmed(true)
-                .cacheEntries(false)
-                .build();
-        IStreamView txStream = runtime.getStreamsView()
-                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
         log.trace("Seeking txStream to {}", lastReadAddress + 1);
-        txStream.seek(lastReadAddress + 1);
+        txnStream.seek(lastReadAddress + 1);
         log.trace("txStream current global position after seeking {}, hasNext {}",
-                txStream.getCurrentGlobalPosition(), txStream.hasNext());
+                txnStream.getCurrentGlobalPosition(), txnStream.hasNext());
 
-        List<ILogData> updates = txStream.remaining();
+        List<ILogData> updates = txnStream.remaining();
 
         log.trace("{} updates remaining in the txStream", updates.size());
 


runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
```diff
@@ -34,17 +34,30 @@
...
     public TransactionPoller(@Nonnull CorfuRuntime runtime,
-            @Nonnull List<StreamingSubscriptionContext> streams) {
+                             @Nonnull List<StreamingSubscriptionContext> streams) {
         this.runtime = runtime;
         this.streamContexts = streams.stream()
                 .sorted(Comparator.comparingLong(sc -> sc.getLastReadAddress()))
                 .collect(Collectors.toList());
+
+        StreamOptions options = StreamOptions.builder()
+                .ignoreTrimmed(true)
```
Maithem: Hmm, @sneginhal why is this true ?

xiaoqin2012: We need to define a generic API to handle TrimmedException for the caller.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
```diff
@@ -69,18 +82,12 @@ private void pollTxStream() {
...
-                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
         log.trace("Seeking txStream to {}", lastReadAddress + 1);
-        txStream.seek(lastReadAddress + 1);
+        txnStream.seek(lastReadAddress + 1);
         log.trace("txStream current global position after seeking {}, hasNext {}",
-                txStream.getCurrentGlobalPosition(), txStream.hasNext());
+                txnStream.getCurrentGlobalPosition(), txnStream.hasNext());
 
-        List<ILogData> updates = txStream.remaining();
+        List<ILogData> updates = txnStream.remaining();
```
xiaoqin2012: This call can throw ExecutionError due to Guava leakage we have seen in some testing.  ExecutionError thrown by google.localcache library. It can throw InterruptedException too if it gets interrupted.

Maithem: `pollTxStream` is called by `run` which has a try/catch block around it.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
Maithem: Trimmed exceptions for objects are automatically handled, but for raw streams usually not much can be done because the user expects a delta, but the delta has been trimmed. I don't see how we can automatically handle these exception because a trimmed exception indicates a loss of deltas, so the consumer has to first observe this exception, then do remedial actions related to their business logic and then re-seek and track deltas from the the current point in time.

hisundar: I can try to answer on behalf of @sneginhal  - this call is made by an internal transaction poller that is common to all the streams. If a stream tries to open something which is trimmed that error will be handled separately. We should not let the transaction poller itself fail with trimmed exception and simply proceed with the latest point.

Maithem:
> The premise mentioned in the PR is not true. The txpoller is not called every time a streamer starts. So technically this is not a leak. It is only once per lifetime of the CorfuServer. I would fix the title of the bug if possible.

Every time pollTxStream is called, it creates a new StreamView object. Polling happens through the `run` method which is called periodically. Because its called periodically, It will keep on creating StreamView objects without closing/removing the ones from older poll runs, so it is a leak.

You are probably referring to the application: Its the consuming client that creates TransactionPoller, not the server.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
Maithem: On a trim exception, the stream pointer has to be re-adjusted, but it doesn't seem like the logic is doing that.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
xiaoqin2012: For ExecutionError thrown by Guava, it is kind of a Guava bug, we should just ignore it. The current implementation propagate any throwable to each subscriber' contexts which will block them polling again in future.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
vjeko: There is no ways for you to detect omissions if youe ignoe `TrimmedException`. How do you detect gaps in the stream?

medhavidhawan: Can we resolve this ? I am not sure if I completely understand this and would want to make sure the correct behaviour is implemented.

hisundar: I have fixed the ignoreTrimmed exception in the latest patch 
https://github.com/CorfuDB/CorfuDB/pull/2287
thanks

Maithem: @hisundar please squash this in your PR. I'm just going to close this.

Maithem:
superseded by  #2287

PR #2254, opened 2019-12-04
Remove ISMRMap and SMRMap

vjeko:
Remove the deprecated (I)SMRMap interfaces and classes. Original tests
have been kept and adopted to use (I)CorfuTable.

## Overview

Description:

Why should this be merged: 

Related issue(s) (if applicable): #<number>

## Checklist (Definition of Done):

- [ ] There are no TODOs left in the code
- [ ] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [ ] Change is covered by automated tests
- [ ] Public API has Javadoc

PR #2284, opened 2019-12-14
Do not resolve execution context on object reset

vjeko:
When calling setUndoRecord(...) during object reset do not try to
resolve the execution context.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -601,7 +601,7 @@ public class VersionLockedObject<T extends ICorfuSMR<T>> {
                 // This entry actually resets the object. So here
                 // we can safely get a new instance, and add the
                 // previous instance to the undo log.
-                entry.setUndoRecord(object.getContext(context));
+                entry.setUndoRecord(object);
                 object.close();
                 object = newObjectFn.get();
                 log.trace("Apply[{}] Undo->RESET", this);


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -601,7 +601,7 @@ private Object applyUpdateUnsafe(SMREntry entry, long timestamp) {
                 // This entry actually resets the object. So here
                 // we can safely get a new instance, and add the
                 // previous instance to the undo log.
-                entry.setUndoRecord(object.getContext(context));
+                entry.setUndoRecord(object);
```
xnull: is it possible to have a test for this logic which demonstrates that this logic is ok and correct?

Maithem: This doesn't look right. The in-memory object itself shouldn't be referenced, instead undo records should be explicitly generated. If the map map's state at v1 is {a,b,c} then a clear happens at v2 and v3 adds {d, e, f}, then undoing the clear would result in an empty map, but it should have restored {a, b, c}. 

Assuming clears are rarely called, you can just set the undo record to false, that would trigger an object rebuild from scratch, which would be correct, but potentially slow.

vjeko: This bug was caught as part of this PR: https://github.com/CorfuDB/CorfuDB/pull/2254

So, there is a test, but it is currently leveraging SMRMap and not CorfuTable.

vjeko: Here are the set of events:

Going Forwards:
```
V1: Add(A, B, C) | Map@{ReferenceX}{A, B, C]
V2: Clear()      | Map@{ReferenceY}{]
                  (Forces a creation of a new object)
                  (Undo record is now pointing at Map@{ReferenceX})
V3: Add(D, E, F) | Map@{ReferenceY}{D, E, F]
```

Going Backwards:
```
From V3 -> V2: Undo Record D, E and F      | Map@{ReferenceY}{]
From V2 -> V1: Restore map at reference X  | Map@{ReferenceX}{A, B, C]
```

What am I missing?

Maithem: The `clear` method doesn't specify an undo function in `CorfuTable`, but the VLO will populate it anyways (special handling), but now I see that the reference is swapped and a new object is created.

I don't think the special handling is needed, since that functionality can be captured with an undofunction, might have been an optimization to eliminate the map copy when generating the undo record.

vjeko: We should definitely look into it since the cost (added complexity) does not give us much benefit. At this point I don't necessarily want to change the behavior, since some of the clients rely on `clear()` to be O(1). If we were to change it, this might case a performance regression on their side.

PR #2282, opened 2019-12-13
Streams view bug

Maithem:
## Overview

Since streams can be opened during a runtime gc cycle, the collection
holding the opened StreamView references should be thread-safe, otherwise
adding to the collection during iteration would cause exceptions.

## Checklist (Definition of Done):

- [ ] There are no TODOs left in the code
- [ ] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [ ] Change is covered by automated tests
- [ ] Public API has Javadoc

PR #2287, opened 2019-12-18
Txn stream leak fix

hisundar:
## Overview

Description:
CorfuStore just has 1 Transaction poller. But to ensure that new
subscribers are taken care of, periodically this transaction poller
is refreshed and a new one is created. But this causes a leak in the
underlying runtime's TRANSACTION StreamsView's openedStreams
resulting in the JVM going out of Memory after a while.
Simple fix is to remove the prior StreamView when recreating.

+Test case to both catch the memory leak bug and validate the fix.

Why should this be merged: 
Results in OOM otherwise 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
@@ -35,16 +35,29 @@ public class TransactionPoller implements Runnable {
     private final List<StreamingSubscriptionContext> streamContexts;
 
     /**
+     * A reference to the transaction stream
+     */
+    private final IStreamView txnStream;
+
+    /**
      * Constructor.
      *
      * @param streams The list of StreamingSubscriptionContexts to process.
      */
     public TransactionPoller(@Nonnull CorfuRuntime runtime,
-            @Nonnull List<StreamingSubscriptionContext> streams) {
+                             @Nonnull List<StreamingSubscriptionContext> streams) {
         this.runtime = runtime;
         this.streamContexts = streams.stream()
                 .sorted(Comparator.comparingLong(sc -> sc.getLastReadAddress()))
                 .collect(Collectors.toList());
+
+        StreamOptions options = StreamOptions.builder()
+                .ignoreTrimmed(true)
+                .cacheEntries(false)
+                .build();
+
+        txnStream = runtime.getStreamsView()
+                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
     }
 
     /**
@@ -69,18 +82,12 @@ public class TransactionPoller implements Runnable {
  
         long lastReadAddress = streamContexts.get(0).getLastReadAddress();
 
-        StreamOptions options = StreamOptions.builder()
-                .ignoreTrimmed(true)
-                .cacheEntries(false)
-                .build();
-        IStreamView txStream = runtime.getStreamsView()
-                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
         log.trace("Seeking txStream to {}", lastReadAddress + 1);
-        txStream.seek(lastReadAddress + 1);
+        txnStream.seek(lastReadAddress + 1);
         log.trace("txStream current global position after seeking {}, hasNext {}",
-                txStream.getCurrentGlobalPosition(), txStream.hasNext());
+                txnStream.getCurrentGlobalPosition(), txnStream.hasNext());
 
-        List<ILogData> updates = txStream.remaining();
+        List<ILogData> updates = txnStream.remaining();
 
         log.trace("{} updates remaining in the txStream", updates.size());
 
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
@@ -1,5 +1,6 @@
 package org.corfudb.runtime.collections;
 
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.List;
@@ -37,6 +38,7 @@ public class TransactionPoller implements Runnable {
     /**
      * A reference to the transaction stream
      */
+    @Getter
     private final IStreamView txnStream;
 
     /**
@@ -52,12 +54,11 @@ public class TransactionPoller implements Runnable {
                 .collect(Collectors.toList());
 
         StreamOptions options = StreamOptions.builder()
-                .ignoreTrimmed(true)
                 .cacheEntries(false)
                 .build();
 
         txnStream = runtime.getStreamsView()
-                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
+                .getUnsafe(ObjectsView.TRANSACTION_STREAM_ID, options);
     }
 
     /**
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TxnStreamingManager.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TxnStreamingManager.java
@@ -148,6 +148,7 @@ public class TxnStreamingManager {
         if (!lockedStreamingSubscriptionContexts.isEmpty()) {
             log.trace("Locked {} StreamingSubscriptionContexts for processing",
                     lockedStreamingSubscriptionContexts.size());
+
             pollerExecutor.submit(new TransactionPoller(runtime, lockedStreamingSubscriptionContexts));
         }
 
--- a/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
@@ -1,7 +1,6 @@
 package org.corfudb.runtime.view;
 
 import com.google.common.annotations.VisibleForTesting;
-import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.DataType;
 import org.corfudb.protocols.wireprotocol.ILogData;
--- a/test/src/test/java/org/corfudb/integration/StreamingIT.java
+++ b/test/src/test/java/org/corfudb/integration/StreamingIT.java
@@ -144,8 +144,9 @@ public class StreamingIT extends AbstractIT {
         store.subscribe(s1n1t1, "n1",
                 Collections.singletonList(new TableSchema("t1", Uuid.class, Uuid.class, Uuid.class)), ts1);
 
-        // After a brief wait verify that the listener go all the updates.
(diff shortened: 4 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/TxnStreamingManager.java
```diff
@@ -148,7 +154,13 @@ private synchronized void scheduleTxnPolling() {
         if (!lockedStreamingSubscriptionContexts.isEmpty()) {
             log.trace("Locked {} StreamingSubscriptionContexts for processing",
                     lockedStreamingSubscriptionContexts.size());
-            pollerExecutor.submit(new TransactionPoller(runtime, lockedStreamingSubscriptionContexts));
+
+            // To prevent leaking transaction stream caches, explicitly remove the prior entry
+            if (transactionPoller != null) {
+                runtime.getStreamsView().getOpenedStreams().remove(transactionPoller.getTxnStream());
```
hisundar: Not a big fan of directly removing things from the underlying StreamsView. Am open to other working alternatives or suggestions.

runtime/src/main/java/org/corfudb/runtime/collections/TxnStreamingManager.java
```diff
@@ -41,6 +41,12 @@
      */
     private final Map<Integer, StreamingSubscriptionContext> subscriptions = new HashMap<>();
 
+    /**
+     * TransactionPoller context to pull updates from the streams.
+     * There should be only 1 for all subscribers.
+     */
+    TransactionPoller transactionPoller;
```
annym: Codacy Issue found: [Use explicit scoping instead of the default package private level](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4695667)

annym:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- runtime/src/main/java/org/corfudb/runtime/collections/TxnStreamingManager.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4695667&bid=15695407)

xiaoqin2012:
I don't see the need to have a new Poller for each scheduled polling. There are other options to support new subscriptions:
 make poll/run function accept the subscription list instead of passing it as arg of constructor.

hisundar:
> I don't see the need to have a new Poller for each scheduled polling. There are other options to support new subscriptions:
> make poll/run function accept the subscription list instead of passing it as arg of constructor.

This was my thought too and I started off making that change only to realize that when the subscription list changes, most of that object isn't valid anyway. So I can see why @sneginhal probably did it this way. So it's sort of like 6 of one and half a dozen of another.

runtime/src/main/java/org/corfudb/runtime/collections/TxnStreamingManager.java
```diff
@@ -148,7 +154,13 @@ private synchronized void scheduleTxnPolling() {
         if (!lockedStreamingSubscriptionContexts.isEmpty()) {
             log.trace("Locked {} StreamingSubscriptionContexts for processing",
                     lockedStreamingSubscriptionContexts.size());
-            pollerExecutor.submit(new TransactionPoller(runtime, lockedStreamingSubscriptionContexts));
+
+            // To prevent leaking transaction stream caches, explicitly remove the prior entry
+            if (transactionPoller != null) {
```
Maithem: I think we need to get rid of this. Why create a new stream on every poll ?

If you really need to create a new stream object on every poll, then get an unsafe stream reference, so that when you create one, the JVM GC will GC it.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -38,15 +37,21 @@
...
 
+    /**
+     * This list should have references to all opened streams. The ViewsGarbageCollector
+     * will use this list to clear trimmed addresses from streams. Since new streams
+     * can be opened during a runtime garbage collection cycle, the collection has
+     * to be thread-safe.
+     */
     @Getter
-    Multimap<UUID, IStreamView> streamCache = Multimaps.synchronizedMultimap(HashMultimap.create());
+    List<IStreamView> openedStreams = new CopyOnWriteArrayList<>();
```
Maithem: Please rebase on #2282

Maithem:
> > I don't see the need to have a new Poller for each scheduled polling. There are other options to support new subscriptions:
> > make poll/run function accept the subscription list instead of passing it as arg of constructor.
> 
> This was my thought too and I started off making that change only to realize that when the subscription list changes, most of that object isn't valid anyway. So I can see why @sneginhal probably did it this way. So it's sort of like 6 of one and half a dozen of another.

What does this mean:

> most of that object isn't valid anyway. 

Are you saying that the stream's prefix at that point isn't needed? If that's the case, then you can just readjust the point via `seek` and keep going. The other layers will handle removing that prefix (caches, runtime gc, etc...)

hisundar:
> > > I don't see the need to have a new Poller for each scheduled polling. There are other options to support new subscriptions:
> > > make poll/run function accept the subscription list instead of passing it as arg of constructor.
> > 
> > 
> > This was my thought too and I started off making that change only to realize that when the subscription list changes, most of that object isn't valid anyway. So I can see why @sneginhal probably did it this way. So it's sort of like 6 of one and half a dozen of another.
> 
> What does this mean:
> 
> > most of that object isn't valid anyway.
> 
> Are you saying that the stream's prefix at that point isn't needed? If that's the case, then you can just readjust the point via `seek` and keep going. The other layers will handle removing that prefix (caches, runtime gc, etc...)

Yes, while that is doable it would mean adding a lot more code for concurrency which isn't needed right now since we create new objects every time. I will look into this though but not sure if its going to be any more less complex than what it is.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/TxnStreamingManager.java
hisundar: fixed, please help review latest patch. thanks!

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
hisundar: done. thanks

annym:


```diff

Issues
======
+ Solved 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4695667&bid=15695407)

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
```diff
@@ -69,18 +83,12 @@ private void pollTxStream() {
...
 
-        StreamOptions options = StreamOptions.builder()
-                .ignoreTrimmed(true)
-                .cacheEntries(false)
-                .build();
-        IStreamView txStream = runtime.getStreamsView()
-                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
         log.trace("Seeking txStream to {}", lastReadAddress + 1);
-        txStream.seek(lastReadAddress + 1);
+        txnStream.seek(lastReadAddress + 1);
```
Maithem: If a trim exception happens, then lastReadAddress won't be updated, no?

At some point you want to seek to the stream tail, so you need to query the sequencer and seek to that point.

Maithem: Just make sure that when you seek, next will return lastReadAddress + 1 and not lastReadAddress + 2, otherwise you might lose an update.

Maithem: Just verify the seek semantics, I can't remember of the top of my head.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
```diff
@@ -69,18 +83,12 @@ private void pollTxStream() {
  
         long lastReadAddress = streamContexts.get(0).getLastReadAddress();
 
-        StreamOptions options = StreamOptions.builder()
```
Maithem: I don't think having the try/catch in run is a good idea. Maybe only catching a trimmed exception is better.

Based on the comment, I think the problem is if you pass a runnable to a scheduler executor and it ends up throwing an exception, then the second run won't schedule that runnable to run. I think its cleaner for the executor owner to actually wrap this runnable with a try/catch and log the exception.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
```diff
@@ -69,18 +83,12 @@ private void pollTxStream() {
...
-                .get(ObjectsView.TRANSACTION_STREAM_ID, options);
         log.trace("Seeking txStream to {}", lastReadAddress + 1);
-        txStream.seek(lastReadAddress + 1);
+        txnStream.seek(lastReadAddress + 1);
         log.trace("txStream current global position after seeking {}, hasNext {}",
-                txStream.getCurrentGlobalPosition(), txStream.hasNext());
+                txnStream.getCurrentGlobalPosition(), txnStream.hasNext());
 
-        List<ILogData> updates = txStream.remaining();
+        List<ILogData> updates = txnStream.remaining();
```
Maithem: If the delta is huge, then this might create a memory issue. Why don't you just continuously call next?

That way you shift the memory issue with a latency issue, later you can introduce a remaining method to the stream that can consume the delta in chunks instead of constructing all the delta in memory first and then consuming it.

annym:


```diff

Issues
======
+ Solved 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/annymartinez.m/CorfuDB/pullRequest?prid=4695667&bid=15695407)

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
hisundar: This is a very good observation - in fact in some consumers the heap memory available is so small that its quite possible with large transactions reconstruction of the deltas cause OOM.
I think this warrants a simpler next() call based streaming. thanks!

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
hisundar: Tracking this comment, for addressing in an upcoming PR. thanks

PR #2300, opened 2019-12-30
Xq/seqcache01

xiaoqin2012:
## Overview
Implement SequencerCache with hashtable and priority queue.

Description:
The legacy used the caffein library, but the replacement policy is not feasible for our use case.
While remove entries from the cache, we need to remove all entries that are smaller than the given address. We need this consistency all the time for the transaction streams. The caffein library doesn't support this consistency. 

Why should this be merged: 
The current cache can be overflowed. It doesn't have feasible eviction policy.

- [X] Public API has Javadoc


annym:


```diff

Issues
======
+ Solved 1
- Added 4
           

Complexity increasing per file
==============================
- infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/annymartinez.m/CorfuDB/pullRequest?prid=4748148&bid=15842070)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,16 +1,14 @@
...
-import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.RemovalCause;
-import com.google.common.annotations.VisibleForTesting;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.view.Address;
 
-import javax.annotation.Nonnull;
+import java.nio.BufferOverflowException;
```
annym: Codacy Issue found: [Unused import - java.nio.BufferOverflowException.](https://app.codacy.com/manual/annymartinez.m/CorfuDB/pullRequest?prid=4748148)

annym: Codacy Issue found: [Avoid unused imports such as 'java.nio.BufferOverflowException'](https://app.codacy.com/manual/annymartinez.m/CorfuDB/pullRequest?prid=4748148)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +74,41 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
      * @return global address
      */
     public Long getIfPresent(ConflictTxStream conflictKey) {
-        return conflictCache.getIfPresent(conflictKey);
+        return cacheHashtable.get(conflictKey);
+    }
+
+    private long firstAddress() {
+        if (cacheEntries.peek() == null)
+            return Address.NOT_FOUND;;
```
annym: Codacy Issue found: [An empty statement (semicolon) not part of a loop](https://app.codacy.com/manual/annymartinez.m/CorfuDB/pullRequest?prid=4748148)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -167,15 +128,24 @@ public long size() {
      * @param newTail        global tail
      */
     public void put(ConflictTxStream conflictStream, long newTail) {
-        conflictCache.put(conflictStream, newTail);
+        if (cacheHashtable.size() == cacheSize) {
+            ConflictTxStreamEntry entry = cacheEntries.peek();
```
annym: Codacy Issue found: [Avoid unused local variables such as 'entry'.](https://app.codacy.com/manual/annymartinez.m/CorfuDB/pullRequest?prid=4748148)

annym:


```diff

Issues
======
+ Solved 1
- Added 1
           

Complexity increasing per file
==============================
- infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/annymartinez.m/CorfuDB/pullRequest?prid=4748148&bid=15842070)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -167,15 +126,24 @@ public long size() {
...
```
annym: Codacy Issue found: [Avoid unused local variables such as 'entry'.](https://app.codacy.com/manual/annymartinez.m/CorfuDB/pullRequest?prid=4748148)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -207,4 +171,26 @@ public String toString() {
             return streamId.toString() + conflictParam;
         }
     }
+
+    public static class ConflictTxStreamEntry {
```
pankti-m: does it need to be public?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -34,8 +30,10 @@
      * <p>
      * a cache of recent conflict keys and their latest global-log position.
      */
-    private final Cache<ConflictTxStream, Long> conflictCache;
-
+    private final Hashtable<ConflictTxStream, Long> cacheHashtable;
```
lixinchengdu: Why Hashtable is used? Hashtable is not as performant as its counterpart. If synchronization is not needed, hashmap is recommended. If the map is used in a multi-threaded context, concurrenthashmap is more performant.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -34,8 +30,10 @@
...
+    private final PriorityQueue<ConflictTxStreamEntry> cacheEntries; //sorted according to address
+    private final int cacheSize;
+    private long maxSequencer = Address.NOT_FOUND; //the max sequence in cacheEntries.
```
lixinchengdu: Probably maxSequence is a better name?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -89,6 +89,12 @@ private void invalidateFirst() {
                 maxConflictWildcard, entry.txVersion, entry.conflictTxStream);
     }
 
+    long firstAddress() {
```
pankti-m: can we add an access specifier here?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -97,8 +103,9 @@ private void invalidateFirst() {
     public void invalidateUpTo(long trimMark) {
         log.debug("Invalidate sequencer cache. Trim mark: {}", trimMark);
         AtomicLong entries = new AtomicLong();
-        ConflictTxStreamEntry entry;
-        while((entry = cacheEntries.peek())!= null && entry.txVersion >= trimMark) {
+        long first;
+
+        while((first = firstAddress()) != Address.NOT_FOUND && first <= trimMark) {
```
pankti-m: I think it is better to initialize first earlier before using it here

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -190,13 +190,12 @@ public ConflictTxStreamEntry(ConflictTxStream stream, long address) {
     }
 
     class ConflictTxStreamEntryComparator implements Comparator<ConflictTxStreamEntry> {
-        // Overriding compare()method of Comparator
-        // for descending order of cgpa
+        // Overriding compare()method of Comparator for ascending order
         public int compare(ConflictTxStreamEntry s1, ConflictTxStreamEntry s2) {
```
pankti-m: can be made concise by returning (s1.txVersion-s2.txVersion)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +72,41 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
      * @return global address
      */
     public Long getIfPresent(ConflictTxStream conflictKey) {
-        return conflictCache.getIfPresent(conflictKey);
+        return cacheHashtable.get(conflictKey);
```
lixinchengdu: HashTable could not contain null keys or values.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -34,8 +30,10 @@
      * <p>
      * a cache of recent conflict keys and their latest global-log position.
      */
-    private final Cache<ConflictTxStream, Long> conflictCache;
-
+    private final Hashtable<ConflictTxStream, Long> cacheHashtable;
+    private final PriorityQueue<ConflictTxStreamEntry> cacheEntries; //sorted according to address
+    private final int cacheSize;
```
lixinchengdu: why not long to make it consistent with the previous version of cache?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +72,41 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
      * @return global address
      */
     public Long getIfPresent(ConflictTxStream conflictKey) {
-        return conflictCache.getIfPresent(conflictKey);
+        return cacheHashtable.get(conflictKey);
+    }
+
+    private long firstAddress() {
+        if (cacheEntries.peek() == null)
```
lixinchengdu: The same as the previous comment.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +72,41 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+            return Address.NOT_FOUND;
+        return cacheEntries.peek().txVersion;
     }
 
     /**
-     * Invalidate all records up to a trim mark.
+     * Invalidate one record.
```
lixinchengdu: It would be better if the JavaDoc is more descriptive. Does this function invalidate the first record?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: what was the motivation for replacing caffeine cache by hasttable?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
PavelZaytsev: Long.compare(s1.txVersion, s2.txVersion)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -167,15 +126,24 @@ public long size() {
...
+        maxSequencer = Math.max(newTail, maxSequencer);
     }
 
     /**
      * Discard all entries in the cache
      */
     public void invalidateAll() {
-        log.info("Invalidate sequencer server cache");
-        conflictCache.invalidateAll();
+        cacheHashtable.clear();
```
PavelZaytsev: Probably better to leave the logging statement

xnull:
>1. The legacy used the caffein library, but the replacement policy is not feasible for our use case.

do you have any evidence that your statement is true? I mean benchmarks or unit tests or anything like that. 
Can you also please explain in more details what exactly caffeine can't do?

> 2. While remove entries from the cache, we need to remove all entries that are smaller than the >given address. We need this consistency all the time for the transaction streams. The caffein >library doesn't support this consistency.

What type of consistency are we talking about? What I know about caffeine is - when I changed it last time it worked well and provided all the needed functionality to solve all the tasks that we had.
Can you please provide more information and details about why we need to replace it with a hashtable?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: The address is a not a member and could not be a member either.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: The priorityQueue constructor only accept int as the argument.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: Why this conversation is resolved?

xiaoqin2012: I have explained before, we have seen overflown issues with caffein cache. The eviction mechanism like LRU etc is not feasible for our use case.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +73,40 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+        return cacheEntries.peek().txVersion;
     }
 
     /**
-     * Invalidate all records up to a trim mark.
+     * Invalidate the first record with the minAddress.
+     */
+    private void invalidateFirst() {
+        ConflictTxStreamEntry entry = cacheEntries.poll();
+        cacheHash.remove(entry.conflictTxStream);
```
xnull: Looks like you can use `TreeMap` instead of `PriorityQueue + HashMap`
Please consider can we replace it and if we can't then why?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +73,40 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
      * @return global address
      */
     public Long getIfPresent(ConflictTxStream conflictKey) {
-        return conflictCache.getIfPresent(conflictKey);
+        return cacheHash.get(conflictKey);
+    }
+
+    private long firstAddress() {
+        if (cacheEntries.peek() == null)
```
xnull: Please replace by `if(!cacheEntries.isEmpty())`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -52,70 +51,18 @@
...
-                if (globalAddress == null) {
-                    globalAddress = Address.NOT_FOUND;
-                }
-                maxConflictWildcard = Math.max(globalAddress, maxConflictWildcard);
-            }
-        };
+    public SequencerServerCache(int cacheSize) {
+        this.cacheSize = cacheSize;
+        cacheHash = new HashMap<>();
+        cacheEntries = new PriorityQueue<> (cacheSize, new ConflictTxStreamEntryComparator());
```
xnull: please reformat code

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +73,40 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
      * @return global address
      */
     public Long getIfPresent(ConflictTxStream conflictKey) {
-        return conflictCache.getIfPresent(conflictKey);
+        return cacheHash.get(conflictKey);
```
xnull: please prevent the method returning `null`, use wether Optional or return a default value

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +73,40 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
         AtomicLong entries = new AtomicLong();
+        long first;
 
-        conflictCache.asMap().forEach((key, txVersion) -> {
-            if (txVersion == null || txVersion >= trimMark) {
-                return;
-            }
-
-            conflictCache.invalidate(key);
+        while((first = firstAddress()) != Address.NOT_FOUND && first < trimMark) {
```
xnull: simplify: `Address.isAddress(firstAddress())`

xnull: Actually, if you migrate on a TreeMap, you don't even need this `while` cycle, you can use methods provided by the treemap, like: `map.headMap(monDate).clear();`

See:
 - https://stackoverflow.com/questions/23936636/remove-treemap-entries-less-than-x
 - https://stackoverflow.com/questions/22976970/treemap-removing-all-keys-greater-than-a-certain-key

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -157,7 +116,7 @@ public void invalidateUpTo(long trimMark) {
      * @return cache size
      */
     public long size() {
-        return conflictCache.estimatedSize();
+        return cacheHash.size();
```
xnull: can we change a return type to `int` ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -167,15 +126,25 @@ public long size() {
      * @param newTail        global tail
      */
     public void put(ConflictTxStream conflictStream, long newTail) {
-        conflictCache.put(conflictStream, newTail);
+        if (cacheHash.size() == cacheSize) {
```
xnull: I don't understand this logic. Can you please explain?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +73,40 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+        log.trace("Updating maxConflictWildcard. Old = '{}', new ='{}' conflictParam = '{}'.",
+                maxConflictWildcard, entry.txVersion, entry.conflictTxStream);
+    }
+
+    /**
+     * Invalidate all records up to a trim mark (not included).
      *
      * @param trimMark trim mark
-     */
+     * */
```
xnull: get rid of extra *

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -167,15 +126,25 @@ public long size() {
...
+        maxAddress = Math.max(newTail, maxAddress);
     }
 
     /**
      * Discard all entries in the cache
      */
     public void invalidateAll() {
-        log.info("Invalidate sequencer server cache");
-        conflictCache.invalidateAll();
+        log.info("Invalidate all entries in sequencer server cache and update maxConflictWildcard to {}", maxAddress);
```
xnull: Since you have replaced caffeine cache by a hashmap, would be great if you also get rid of this method.
Instead of spending linear time O(n) on cleaning maps, you can just create a new instance of a SequencerServerCache and let garbage collector do his work

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -207,4 +177,21 @@ public String toString() {
...
+        private final ConflictTxStream conflictTxStream;
+        private final long txVersion;
+
+        public ConflictTxStreamEntry(ConflictTxStream stream, long address) {
+            conflictTxStream = stream;
+            txVersion = address;
+        }
+    }
+
+    class ConflictTxStreamEntryComparator implements Comparator<ConflictTxStreamEntry> {
```
xnull: The entire class has to be replaced by a variable
```
private static final Comparator<ConflictTxStreamEntry> CONFLICT_TX_COMPARATOR = Comparator.comparing(ConflictTxStreamEntry::txVersion);
```

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: Actually `size` based eviction should work. 
Instead of `size` you can use global addresses.
https://github.com/ben-manes/caffeine/wiki/Eviction

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -155,7 +155,7 @@ public SequencerServer(ServerContext serverContext) {
 
         globalLogTail = config.getInitialToken();
 
-        this.cache = new SequencerServerCache(config.getCacheSize());
+        this.cache = new SequencerServerCache((int)config.getCacheSize());
```
Maithem: You can make the cache size an int instead of down casting.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -19,7 +15,7 @@
  * Contains transaction conflict-resolution data structures.
  * <p>
  * The SequencerServer use its own thread/s. To guarantee correct tx conflict-resolution,
- * the {@link SequencerServerCache#conflictCache} must be updated
+ * the {@link SequencerServerCache#cacheHash} must be updated
```
Maithem: Those are keys not hashes, maybe `conflictKeys` might be a better name.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -34,8 +30,11 @@
      * <p>
      * a cache of recent conflict keys and their latest global-log position.
      */
-    private final Cache<ConflictTxStream, Long> conflictCache;
-
+    //As the sequencer cache is used by a single thread, it is safe to use hashmap.
```
Maithem: Just mark the class as `@NotThreadSafe`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -34,8 +30,11 @@
...
+    private final HashMap<ConflictTxStream, Long> cacheHash;
+    private final PriorityQueue<ConflictTxStreamEntry> cacheEntries; //sorted according to address
```
Maithem: I suggest you remove one type. You can have ConflictKeyEntry that also encapsulates the long and can also have a comparator .

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: I don't think Optional should be used in lieu of null "everywhere", especially in a hot code path. The reason being is it adds to memory pressure.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: This seems to be incorrect. If the priority queue fills up with [1, 2, 3, 4, 5], then `invalidateUpTo` is called to 6, then it fills up with [6, 7, 8, 9], `firstAddress` will return `Address.NOT_FOUND`. Then a put cannot cause an eviction because 6 is greater than `firstAddress()` + 1. Essentially, the window cannot slide.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: @xnull A `TreeMap` cannot have duplicates, where as a `PriorityQueue` can. Although, its possible to use a tree map, but the modelling has to be changed. Yes, you can can a "free" range remove method, but will it be simpled? I don't know, maybe you should propose something.

Maithem: The remove is not necessary and its also inefficient. `poll` will `reheapify`  and then `remove` will do a linear scan. It also doesn't evict multiple keys which belong to the same version, so if the heap looks like [k1:1, k2:1], only k1 will be evicted, but k2 cannot be used, so it also needs to be evicted.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +73,40 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+            return Address.NOT_FOUND;
+        return cacheEntries.peek().txVersion;
     }
 
     /**
-     * Invalidate all records up to a trim mark.
+     * Invalidate the first record with the minAddress.
+     */
+    private void invalidateFirst() {
+        ConflictTxStreamEntry entry = cacheEntries.poll();
```
Maithem: Alternatively, you can first `peek`.

```java
ConflictTxStreamEntry entry = cacheEntries.peek();
if (entry == null) return;
while (cacheEntries.peek().txVersion == entry.txVersion) cacheEntries.poll();
```

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: Please enable auto-formatting with intellij.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: You can do this instead.

https://github.com/CorfuDB/CorfuDB/pull/2300/files#r364366983

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,28 +73,40 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
 
-        conflictCache.asMap().forEach((key, txVersion) -> {
-            if (txVersion == null || txVersion >= trimMark) {
-                return;
-            }
-
-            conflictCache.invalidate(key);
+        while((first = firstAddress()) != Address.NOT_FOUND && first < trimMark) {
+            invalidateFirst();
             entries.incrementAndGet();
```
Maithem: Why is this an AtomicLong ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: @xnull This means when the cache is full, you put must first evict (make space in the cache), before you can add the entry.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -167,15 +126,25 @@ public long size() {
      * @param newTail        global tail
      */
     public void put(ConflictTxStream conflictStream, long newTail) {
-        conflictCache.put(conflictStream, newTail);
+        if (cacheHash.size() == cacheSize) {
+            ConflictTxStreamEntry entry = cacheEntries.peek();
+            invalidateUpTo(firstAddress() + 1);
```
Maithem: At some point the cache will fill up and every put will translate into an eviction first, instead of having this code being invoked on every put, you can just do a  "bulk" eviction which is a % of the cache. For example, evict 25% of the cache whenever evict is invoked.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -167,15 +126,25 @@ public long size() {
...
+        }
+
+        ConflictTxStreamEntry entry = new ConflictTxStreamEntry(conflictStream, newTail);
+        cacheEntries.add(entry);
+        cacheHash.put(conflictStream, entry.txVersion);
+        maxAddress = Math.max(newTail, maxAddress);
```
Maithem: You don't need to track a `maxAddress` because `updateMaxConflictAddress` is exposed and it will be called with `cache.updateMaxConflictAddress(initialToken - 1);`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: +1, I would also suggest to get rid of this method. The caller can recreate these objects.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -187,6 +156,7 @@ public void updateMaxConflictAddress(long newMaxConflictWildcard) {
         log.info("updateMaxConflictAddress, new address: {}", newMaxConflictWildcard);
```
Maithem: You can also remove `updateMaxConflictAddress` once you recreate the cache on reset.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,16 +1,12 @@
...
-import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.RemovalCause;
-import com.google.common.annotations.VisibleForTesting;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.view.Address;
-
-import javax.annotation.Nonnull;
+import java.util.Comparator;
```
Maithem: This is a new data structure, please add specific tests that test eviction logic. So, a test to verify that the window can keep sliding, and eviction of a version will remove multiple keys, etc.

Maithem:
Can you also add cache stats like weight, to get a sense of how much memory its using.
In general, computing the in-mem size of a java object can be expensive, but ConflictTxStream can easily be computed, its just a byte array + uuid.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: But the eviction (even one by one) is very cheap. I wouldn't make the eviction logic more complex. I would prefer the current approach.
But up to you guys

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: Yes, but using `OptionalInt` will be equivalent to using Long. 
If you need better performance then let's return default value and use a primitive long (we need to avoid auto wrapping).
In the other case when we use `Long` is the worst of all possible ways to go.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Good point, we can replace priorityque with treemap that will be much smaller size if the number of entries are big, it can save some place. Will change the logic accordingly.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: @xnull 
The cache map maps conflict keys (stream id + key) to versions (long), illustrated below:

Conflict Key | ck1 | ck2 | ck3 | ck4
Version        | v1   | v1   |  v2  |  v3

Consider the case where we need to insert a new conflict key (ck), but the cache is full, then we need to evict the oldest conflict keys. In the example above, we can't just evict ck1, need to evict all keys that map to v1, so we need to evict ck1 and ck2, this eviction policy is FIFO on the version number. Notice that we also can't evict ck3 before the keys for v1, that's because it will create holes in the resolution window and can lead to incorrect resolutions. 

How can we use the caffeine cache to achieve this behavior ?

Note that this is a sliding window on the versions, where a version can map to multiple keys, so we also need to maintain the begining of the window which is the `maxConflictWildcard` variable.


@xiaoqin2012 Please capture this in the class comments, because its not obvious to everyone why we need to have this new data structure.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: It is easier to remove stuff from treeMap, but we still need to know all the keys to remove stuff from HashSet. So treeMap is not feasible in our case.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,56 +81,76 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
-            }
-
-            conflictCache.invalidate(key);
-            entries.incrementAndGet();
-        });
-
-        log.info("Invalidated entries: {}", entries.get());
+        while(Address.isAddress(firstAddress()) && firstAddress() < trimMark) {
+            invalidateFirst();
+            entries++;
```
zhangn49: Is the counter entries accurate? because invalidateFirst() might remove multiple entries.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -406,7 +401,6 @@ public void resetServer(CorfuPayloadMsg<SequencerRecoveryMsg> msg,
         // It is necessary because we reset the sequencer.
         if (!bootstrapWithoutTailsUpdate) {
             // Evict all entries from the cache. This eviction triggers the callback modifying the maxConflictWildcard.
```
zhangn49: should we remove this comment line?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -152,13 +152,8 @@ public SequencerServer(ServerContext serverContext) {
         this.executor = Executors.newSingleThreadExecutor(
                 new ServerThreadFactory("sequencer-", new ServerThreadFactory.ExceptionHandler()));
 
-
         globalLogTail = Address.getMinAddress();
-
         this.cache = new SequencerServerCache(config.getCacheSize());
-
-
-        setUpTimerNameCache();
```
Maithem: Why remove the metrics timers?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,16 +1,18 @@
...
-import com.github.benmanes.caffeine.cache.CacheWriter;
-import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.RemovalCause;
-import com.google.common.annotations.VisibleForTesting;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
+import net.jcip.annotations.NotThreadSafe;
```
Maithem: Use `import javax.annotation.concurrent.NotThreadSafe;`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,56 +81,76 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
-        return conflictCache.getIfPresent(conflictKey);
+        return cacheConflictKeys.get(conflictKey);
+    }
+
+    /**
+     * The first address in the priority queue.
+     */
+    public long firstAddress() {
+        if (cacheEntries.isEmpty())
+            return Address.NOT_FOUND;
```
Maithem: Please use curly brackets:

https://stackoverflow.com/a/15786982

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: Good point @zhangn49 

invalidateFirst can return number of entries evicted and then that can be used in the counter.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,56 +81,76 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+        }
+
+        int i = 0;
+        while (!cacheEntries.isEmpty() && cacheEntries.peek().txVersion == firstEntry.txVersion) {
+            ConflictTxStream entry = cacheEntries.poll();
+            cacheConflictKeys.remove(entry);
+            i++;
+        }
+
+        log.debug("Evict {} entries", i);
```
Maithem: Please change this to trace level

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: Well if the entries refer to the sequence number, then this counter is correct. If they strictly refer to cache entries, then it isnt correct because it under counts.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,56 +81,76 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
      * Put a value in the cache
      *
      * @param conflictStream conflict stream
-     * @param newTail        global tail
      */
-    public void put(ConflictTxStream conflictStream, long newTail) {
-        conflictCache.put(conflictStream, newTail);
-    }
+    public void put(ConflictTxStream conflictStream) {
+        if (cacheConflictKeys.size() == cacheSize) {
```
Maithem: Can you add a check that the conflict stream's version is greater than the old key's value (if it exists) and the corresponding tests. This cache implements a sliding window, but just based on these API's is possible to regress the key versions. Essentially, the put API always assumes that what ever key you put in, the caller has to guarantee that the corresponding value is greater,  but I think we should verify that regarderless of whether the caller does it or not.

Its a subtle point, but will catch any bugs in the sequencer server.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -185,6 +160,7 @@ public void invalidateAll() {
      */
     public void updateMaxConflictAddress(long newMaxConflictWildcard) {
         log.info("updateMaxConflictAddress, new address: {}", newMaxConflictWildcard);
+        invalidateUpTo(newMaxConflictWildcard);
         maxConflictWildcard = newMaxConflictWildcard;
         maxConflictNewSequencer = newMaxConflictWildcard;
```
Maithem: I think we can remove this variable, since holds redundant information. We only need maxConflictWildCard every time we boot strap the sequencer, we can set maxConflictWildCard = globalAddress(tail from bootstrap msg). 

@annym can further comment.

annym: Mmm perhaps we should keep them. Because imagine we have maxConflictNewSequencer = 60 (when we elected this sequencer as primary, that was the log's tail) and quickly maxConflictWildcard becomes 200 (because there was sequencer overflow) and we have a txSnapshot @ 50. Should the one @ 50 fail due to new sequencer or overflow? Also, we would need to remove the cause NEW_SEQUENCER cause we wouldn't be able to distinguish if the eviction was because cache invalidation or overflow.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Yer. It is two different type of cache miss for different reasons.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Added more tests.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -704,15 +696,15 @@ public void shutdown() {
...
 
         @Default
-        private final long cacheSize = DEFAULT_CACHE_SIZE;
+        private final int cacheSize = DEFAULT_CACHE_SIZE;
 
         public static Config parse(Map<String, Object> opts) {
-            long cacheSize = opts.containsKey("--sequencer-cache-size") ?
-                    Long.parseLong((String) opts.get("--sequencer-cache-size")) : DEFAULT_CACHE_SIZE;
-
+            int cacheSize = (int)(opts.containsKey("--sequencer-cache-size") ?
```
xnull:
```
nt cacheSize = (int) opts.getOrDefault("--sequencer-cache-size", DEFAULT_CACHE_SIZE);
```

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +96,126 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
-            if (txVersion == null || txVersion >= trimMark) {
-                return;
-            }
+    /**
+     * Invalidate the records with the minAddress. It could be one or multiple records
+     */
+    private int invalidateFirst() {
+        ConflictTxStream firstEntry = cacheEntries.peek();
+        if (firstEntry == null) {
+            return 0;
```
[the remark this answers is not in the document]
xiaoqin2012: This return the number of entries, not an address.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +96,126 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+        }
 
-            conflictCache.invalidate(key);
-            entries.incrementAndGet();
-        });
+        int i = 0;
```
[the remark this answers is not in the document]
xiaoqin2012: "i" is the number of entries has been removed.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +96,126 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
-     * @param newMaxConflictWildcard new conflict wildcard
+     * @param conflictStream conflict stream
      */
-    public void updateMaxConflictAddress(long newMaxConflictWildcard) {
-        log.info("updateMaxConflictAddress, new address: {}", newMaxConflictWildcard);
-        maxConflictWildcard = newMaxConflictWildcard;
-        maxConflictNewSequencer = newMaxConflictWildcard;
+    public boolean put(ConflictTxStream conflictStream) {
+
+        Long val = cacheConflictKeys.get(conflictStream);
```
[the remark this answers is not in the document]
xiaoqin2012: get function return an object or null. Could not use long.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xiaoqin2012: getOrDefault will return a string for the key exists, for default it is an Integer. We need to parse differently.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -260,7 +256,7 @@ private TxResolutionResponse txnCanCommit(TxResolutionInfo txInfo) {
             // for each key pair, check for conflict; if not present, check against the wildcard
             for (byte[] conflictParam : conflictParamSet) {
 
-                Long keyAddress = cache.getIfPresent(new ConflictTxStream(conflictStream.getKey(), conflictParam));
+                Long keyAddress = cache.getIfPresent(new ConflictTxStream(conflictStream.getKey(), conflictParam, Address.NON_ADDRESS));
```
xnull: please use `long'

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xnull: my example works correctly, you can use it

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
xnull: so then use `entries` or `numEntries` or `removedEntries`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
xnull: then you can use `cacheConflictKeys.getOrDefault(conflictStream, Address.NON_ADDRESS);`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +101,127 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+            txVersion = address;
         }
 
         @Override
         public String toString() {
             return streamId.toString() + conflictParam;
         }
+
+        public long size() {
+            return MetricsUtils.sizeOf.deepSizeOf(this);
```
xnull: Why we need to measure the size of the object in memory?
It's a very rare and fragile operation in java to measure the object size. Is it possible to avoid that?
What we actually measuring here and how we are going to use it?

test/src/test/java/org/corfudb/runtime/object/transactions/SequencerServerCacheTest.java
```diff
@@ -68,36 +64,149 @@ public void testSequencerCacheTrim() {
...
+        }
+        cache.put(key);
+        assertThat(cache.getIfPresent(key) != null);
+        recordMap.put(key, address);
+        assertThat(cache.size() <= cacheSize);
+    }
+
+    // Verify cache contains all the data in recordMap that address >= firstAddress.
+    void verifyData(HashMap<ConflictTxStream, Long> recordMap, SequencerServerCache cache) {
+        for (ConflictTxStream oldKey : recordMap.keySet()) {
```
xnull: code reformat please

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +101,127 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
      */
-    public void updateMaxConflictAddress(long newMaxConflictWildcard) {
-        log.info("updateMaxConflictAddress, new address: {}", newMaxConflictWildcard);
-        maxConflictWildcard = newMaxConflictWildcard;
-        maxConflictNewSequencer = newMaxConflictWildcard;
+    public boolean put(ConflictTxStream conflictStream) {
+
+        Long val = cacheConflictKeys.get(conflictStream);
+        if (val != null && val >= conflictStream.txVersion) {
+            log.error("For key {}, the value {} is not smaller  than the expected value {} or maxWildCard",
```
WenbinZhu: Please remove extra space after "smaller"
Also the log says "or maxWildCard" but the code does not check maxWildCard.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
pankti-m: nit - line length or 80 chars

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +101,127 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+        ConflictTxStream firstEntry = cacheEntries.peek();
+        if (firstEntry == null) {
+            return 0;
+        }
 
-            conflictCache.invalidate(key);
-            entries.incrementAndGet();
-        });
+        int i = 0;
+        while (!cacheEntries.isEmpty() && cacheEntries.peek().txVersion == firstEntry.txVersion) {
```
WenbinZhu: Looks like you can use the `firstAddress()` method you added to simplify this logic: `while (firstAddress() == firstEntry.txVersion)`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -704,15 +696,14 @@ public void shutdown() {
...
         @Default
-        private final long cacheSize = DEFAULT_CACHE_SIZE;
+        private final int cacheSize = DEFAULT_CACHE_SIZE;
 
         public static Config parse(Map<String, Object> opts) {
-            long cacheSize = opts.containsKey("--sequencer-cache-size") ?
-                    Long.parseLong((String) opts.get("--sequencer-cache-size")) : DEFAULT_CACHE_SIZE;
-
+            int cacheSize = (int)(opts.containsKey("--sequencer-cache-size") ?
+            Long.parseLong((String)opts.get("--sequencer-cache-size")) : DEFAULT_CACHE_SIZE);
```
pankti-m: why parse into a long and then typecast to an integer?  Can we use Integer.parseInt() instead?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
WenbinZhu: I agree to change the variable name instead of `i`  to make it meaningful as this varible will be returned by the method.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,47 +1,66 @@
...
 
-import com.github.benmanes.caffeine.cache.Cache;
-import com.github.benmanes.caffeine.cache.CacheWriter;
-import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.RemovalCause;
-import com.google.common.annotations.VisibleForTesting;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import javax.annotation.concurrent.NotThreadSafe;
```
pankti-m: nit - alphabetic order of imports

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +101,127 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+     * @param trimMark trim mark
      */
-    public long size() {
-        return conflictCache.estimatedSize();
+    public void invalidateUpTo(long trimMark) {
+        log.debug("Invalidate sequencer cache. Trim mark: {}", trimMark);
+        int entries = 0;
+        int pqEntries = 0;
+        while(Address.isAddress(firstAddress()) && firstAddress() < trimMark) {
+            pqEntries +=invalidateFirst();
```
WenbinZhu: NIT: Add space after `while` and after `+=`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -70,52 +83,14 @@ public SequencerServerCache(long cacheSize, CacheWriter<ConflictTxStream, Long>
...
-                if (globalAddress == null) {
-                    globalAddress = Address.NOT_FOUND;
-                }
-                maxConflictWildcard = Math.max(globalAddress, maxConflictWildcard);
-            }
-        };
+    public SequencerServerCache(int cacheSize, long maxConflictNewSequencer) {
+        this.cacheSize = cacheSize;
+        cacheConflictKeys = new HashMap<> ();
+        cacheEntries = new PriorityQueue<> (cacheSize, Comparator.comparingLong((ConflictTxStream a) -> a.txVersion));
```
WenbinZhu: You can remove the casting to ConflictTxStream which is redundant.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,47 +1,66 @@
...
 public class SequencerServerCache {
     /**
      * TX conflict-resolution information:
-     * <p>
      * a cache of recent conflict keys and their latest global-log position.
      */
-    private final Cache<ConflictTxStream, Long> conflictCache;
+
+    // As the sequencer cache is used by a single thread, it is safe to use hashmap.
+    private final HashMap<ConflictTxStream, Long> cacheConflictKeys;
```
WenbinZhu: I think you do not need a hashmap here, using a HashSet should be enough because ConflictTxStream already contains the version.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,85 +101,127 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
      * @return global address
      */
     public Long getIfPresent(ConflictTxStream conflictKey) {
-        return conflictCache.getIfPresent(conflictKey);
+        return cacheConflictKeys.get(conflictKey);
```
WenbinZhu: Per previous comment, a HashSet could fulfill this purpose.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: HashSet doesn't have get function. Given the key is part of the object, we need to retrieve the address.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xnull: So, the method by itself doesn't have any meaning and brings confusion. Please rename it to `get()` because the method is just a wrapper for the get method in a hashmap. 
Always, try to avoid having null as much as possible. In this case, it is pretty trivial to do.

I see several ways to go:
 1. use java.util.OptionalLong - the most convenient way in my opinion. It consumes the same amount of memory as Long but prevents you from null pointer exception. So, your method `getIfPresent` will return OptionalLong.
 2. use defaults, `long keyAddress = cache.getOrDefault(key, -1)` which also prevents you from NPE.
3. use https://bitbucket.org/trove4j/trove/src/master/ to avoid `autoboxing`. Since SequencerServer is a heavy loaded class, avoid using object when you keep only primitives dramatically improves performance and reduces memory consumption.

My opinion is to choose the first approach

 - Why you have to avoid using nulls https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/
 - autoboxing https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -50,18 +66,15 @@
...
-                .executor(Runnable::run)
-                .recordStats()
-                .build();
-    }
+    private long maxConflictNewSequencer;
+
+    /* It is used to calculate the size of ServerCache.
+     * Each entry relates two pointers used by HashMap, one pointer in PriorityQueue.
+     */
+    static final int Entry_Overhead = 24;
```
xnull: please rename according to code convention
also if it possible make it private

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -50,18 +66,15 @@
...
+
+    //As calculating object size is expensive, used the value calculated by deepSize
+    static final int ConflictTxStream_Obj_Size = 80; //by calculated by deepSize
```
xnull: please rename according to code convention
also if it possible make it private

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -50,18 +66,15 @@
...
-                .maximumSize(cacheSize)
-                .writer(writer)
-                //Performing periodic maintenance using current thread (synchronously)
-                .executor(Runnable::run)
-                .recordStats()
-                .build();
-    }
+    private long maxConflictNewSequencer;
+
+    /* It is used to calculate the size of ServerCache.
```
xnull: make it JavaDoc, please, not multiline comment

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -70,52 +83,13 @@ public SequencerServerCache(long cacheSize, CacheWriter<ConflictTxStream, Long>
...
-                        maxConflictWildcard, globalAddress, key, cause
-                );
-
-                if (globalAddress == null) {
-                    globalAddress = Address.NOT_FOUND;
-                }
-                maxConflictWildcard = Math.max(globalAddress, maxConflictWildcard);
-            }
-        };
+    public SequencerServerCache(int cacheSize, long maxConflictNewSequencer) {
```
xnull: please remove part of javadoc (caffeine not used anymore)
image

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -70,52 +83,13 @@ public SequencerServerCache(long cacheSize, CacheWriter<ConflictTxStream, Long>
...
+        this.cacheSize = cacheSize;
+        cacheConflictKeys = new HashMap<> ();
+        cacheEntries = new PriorityQueue<> (cacheSize, Comparator.comparingLong(a -> a.txVersion));
```
xnull: code reformat

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -70,52 +83,13 @@ public SequencerServerCache(long cacheSize, CacheWriter<ConflictTxStream, Long>
...
-
-                if (globalAddress == null) {
-                    globalAddress = Address.NOT_FOUND;
-                }
-                maxConflictWildcard = Math.max(globalAddress, maxConflictWildcard);
-            }
-        };
+    public SequencerServerCache(int cacheSize, long maxConflictNewSequencer) {
+        this.cacheSize = cacheSize;
+        cacheConflictKeys = new HashMap<> ();
```
xnull: code reformat

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,80 +100,120 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
-            if (txVersion == null || txVersion >= trimMark) {
-                return;
-            }
+    /**
+     * Invalidate the records with the minAddress. It could be one or multiple records
+     * @return the number of entries has been invalidated and removed from the cache.
+     */
+    private int invalidateFirst() {
+        ConflictTxStream firstEntry = cacheEntries.peek();
+        if (firstEntry == null) {
```
xnull: just check if `cacheEntries` isEmpty, instead of checking that the entry is null

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,80 +100,120 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+        }
 
-            conflictCache.invalidate(key);
-            entries.incrementAndGet();
-        });
+        int numEntries = 0;
+        while (firstAddress() == firstEntry.txVersion) {
+            ConflictTxStream entry = cacheEntries.poll();
+            cacheConflictKeys.remove(entry);
+            numEntries ++;
```
xnull: code reformat

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,80 +100,120 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+     * priority queue and hashmap to store the pointers
+     * @return the memory space used in bytes:
      */
-    public void invalidateAll() {
-        log.info("Invalidate sequencer server cache");
-        conflictCache.invalidateAll();
+    public long byteSize() {
+        log.debug("the cache has {} entries,  the object size used {}, calculated by beepSize {}",
+                size(), ConflictTxStream_Obj_Size,
+                cacheEntries.isEmpty() ? 0 : MetricsUtils.sizeOf.deepSizeOf(cacheEntries.peek()));
```
xnull: You can't use `deepSizeOf ` so often, it's not desired operation. Avoid it as much as possible, only in very rare and limited situations when there is no choice then you can use it, otherwise not.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,80 +100,120 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+        return size()*(Entry_Overhead + ConflictTxStream_Obj_Size);
```
xnull: code reformat

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,80 +100,120 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
+     * @param conflictStream conflict stream
      */
-    public void updateMaxConflictAddress(long newMaxConflictWildcard) {
-        log.info("updateMaxConflictAddress, new address: {}", newMaxConflictWildcard);
-        maxConflictWildcard = newMaxConflictWildcard;
-        maxConflictNewSequencer = newMaxConflictWildcard;
+    public boolean put(ConflictTxStream conflictStream) {
+
+        Long val = cacheConflictKeys.get(conflictStream);
+        if (val != null && val > conflictStream.txVersion) {
```
xnull: avoid using null (explained above)

test/src/test/java/org/corfudb/runtime/object/transactions/SequencerServerCacheTest.java
```diff
@@ -68,36 +64,149 @@ public void testSequencerCacheTrim() {
...
+            long oldAddress = oldKey.txVersion;
+            if (oldAddress < cache.firstAddress())
```
xnull: please add braces

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,47 +1,63 @@
...
 import java.util.UUID;
-import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * Sequencer server cache.
  * Contains transaction conflict-resolution data structures.
- * <p>
  * The SequencerServer use its own thread/s. To guarantee correct tx conflict-resolution,
- * the {@link SequencerServerCache#conflictCache} must be updated
+ * the {@link SequencerServerCache#cacheConflictKeys} must be updated
```
xnull: Halve of the JavaDoc is not relevant anymore (because you replaced caffeine by a hashmap)

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xnull: please rename `a` -> `conflictStream` or to something similar

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -126,80 +100,120 @@ public void delete(@Nonnull ConflictTxStream key, Long globalAddress, @Nonnull R
...
 
-        conflictCache.asMap().forEach((key, txVersion) -> {
-            if (txVersion == null || txVersion >= trimMark) {
-                return;
-            }
+    /**
+     * Invalidate the records with the minAddress. It could be one or multiple records
+     * @return the number of entries has been invalidated and removed from the cache.
+     */
+    private int invalidateFirst() {
```
xnull: let's rename to `invalidateSmallestTxVersion()`

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: This is the debug mode. Also, deepSize is costly only when you pass in an array like structure and the object has embeded deep pointers.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Clean up the information

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xiaoqin2012: Will not address.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Made the change.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,47 +1,66 @@
...
+
+    // As the sequencer cache is used by a single thread, it is safe to use hashmap.
+    private final HashMap<ConflictTxStream, Long> cacheConflictKeys;
+    private final PriorityQueue<ConflictTxStream> cacheEntries; //sorted according to address
+
+    @Getter
+    private final int cacheSize; // the max number of entries
+
+    @Getter
+    public long cacheEntriesBytes; // the memorySpace used.
```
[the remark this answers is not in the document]
xiaoqin2012: done

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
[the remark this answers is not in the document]
xiaoqin2012: Made the change.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
xiaoqin2012: Will not address.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,47 +1,63 @@
...
 public class SequencerServerCache {
     /**
      * TX conflict-resolution information:
-     * <p>
      * a cache of recent conflict keys and their latest global-log position.
      */
-    private final Cache<ConflictTxStream, Long> conflictCache;
+
+    // As the sequencer cache is used by a single thread, it is safe to use hashmap.
+    private final HashMap<ConflictTxStream, Long> cacheConflictKeys;
```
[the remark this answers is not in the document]
xiaoqin2012: Done.

test/src/test/java/org/corfudb/runtime/object/transactions/SequencerServerCacheTest.java
```diff
@@ -68,36 +64,150 @@ public void testSequencerCacheTrim() {
...
-        final ConflictTxStream secondKey = new ConflictTxStream(UUID.randomUUID(), new byte[]{});
+        SequencerServerCache cache = new SequencerServerCache(1, Address.NOT_FOUND);
         final long firstValue = 1L;
         final long secondValue = 2L;
         final int iterations = 10;
+        final ConflictTxStream firstKey = new ConflictTxStream(UUID.randomUUID(), new byte[]{}, firstValue);
+        final ConflictTxStream secondKey = new ConflictTxStream(UUID.randomUUID(), new byte[]{}, secondValue);
 
         for (int i = 0; i < iterations; i++) {
             criticalVariable.set(false);
```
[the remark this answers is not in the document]
xiaoqin2012: It is not my change. It is the legacy code.

test/src/test/java/org/corfudb/runtime/object/transactions/SequencerServerCacheTest.java
```diff
@@ -68,36 +64,150 @@ public void testSequencerCacheTrim() {
...
+
+        // put entries to the cache, make it full, some entries have the same address
+        while (cache.size() < cacheSize) {
+            for (int j = 0; j < entryPerAddress; j++) {
+                generateData(recordMap, cache, address++, false);
+            }
+        }
+
+        for (ConflictTxStream entry : recordMap.keySet()) {
+            assertThat(cache.put(new ConflictTxStream(entry.getStreamId(), entry.getConflictParam(), entry.txVersion)) == false);
```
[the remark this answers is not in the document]
xiaoqin2012: It is different logic.

Maithem:
@xnull @WenbinZhu @PavelZaytsev please approve if your comments have been addressed.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -50,18 +61,16 @@
...
-                .recordStats()
-                .build();
-    }
+    private long maxConflictNewSequencer;
+
+    /**
+     * It is used to calculate the size of ServerCache. Each entry relates two pointers
+     * used by HashMap, one pointer in PriorityQueue.
+     */
+    static final private int ENTRY_OVERHEAD = 24;
```
xnull: *-> private static final

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -50,18 +61,16 @@
...
+
+    //As calculating object size is expensive, used the value calculated by deepSize
+    static final private int CONFLICTTXSTREAM_OBJ_SIZE = 80; //by calculated by deepSize
```
xnull: *-> private static final

PR #2302, opened 2020-01-07
State Transfer Codec overwrite fix

annym:
## Overview

In the case of data written with a different codec than that used by the State Transfer runtime, we would overwrite the codec of compressed data. Making it imposible for a runtime reader to decompress correctly. 

A test has been added to confirm this.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
@@ -221,7 +221,6 @@ public class StateTransfer {
                 throw new IllegalStateException("Missing address");
             }
             LogData ld = (LogData) dataMap.get(address);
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/StateTransfer.java
```diff
@@ -221,7 +221,6 @@ private static void transferChunk(Layout layout,
                 throw new IllegalStateException("Missing address");
             }
             LogData ld = (LogData) dataMap.get(address);
-            ld.setPayloadCodecType(Codec.Type.valueOf(runtime.getParameters().getCodecType()));
```
Maithem: I actually did this intentionally; assuming that the client and the server bits have the same codecs. The reason is, to load the new server/client version with an old data log (i.e. uncompressed), and the data can be compressed by bouncing the servers one-by-one to trigger state transfer and compress the data without having to rewrite it via checkpoint. Maybe that's not a deal breaker, but its just something for you to consider.

PR #2306, opened 2020-01-11
CorfuStoreBrowser: Support infoTable and dropTable

hisundar:
## Overview

Callers want a way to clear the contents of a CorfuStore table via CLI.
They also want a way to just show the size of a table without actually dumping the whole content.

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -3,6 +3,7 @@ package org.corfudb.browser;
 import com.google.common.reflect.TypeToken;
 
 import java.util.Map;
+import java.util.UUID;
 
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
@@ -81,10 +82,23 @@ public class CorfuStoreBrowser {
     {
         log.info("\n=====Tables=======\n");
         for (CorfuStoreMetadata.TableName tableName : runtime.getTableRegistry()
-            .listTables(namespace)) {
+                .listTables(namespace)) {
             log.info("Table: " + tableName.getTableName());
             log.info("Namespace: " + tableName.getNamespace());
         }
+        log.info("\n======================\n");
+    }
+
+    /**
+     * Print information about a specific table in CorfuStore
+     */
+    public void printTableInfo(String namespace, String tablename) {
+        log.info("\n======================\n");
+        String fullName = TableRegistry.getFullyQualifiedTableName(namespace, tablename);
+        UUID streamUUID = UUID.nameUUIDFromBytes(fullName.getBytes());
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(namespace, tablename);
+        log.info("Table {} in namespace {} with ID {} has {} entries",
+                table, namespace, streamUUID.toString(), table.size());
         log.info("\n======================\n");
     }
 }
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
@@ -2,6 +2,7 @@ package org.corfudb.browser;
 
 import java.util.Map;
 
+import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.runtime.CorfuRuntime;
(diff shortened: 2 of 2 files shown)


corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -81,10 +82,23 @@ public void listTables(String namespace)
...
             log.info("Namespace: " + tableName.getNamespace());
         }
         log.info("\n======================\n");
     }
+
+    /**
+     * Print information about a specific table in CorfuStore
+     */
+    public void printTableInfo(String namespace, String tablename) {
+        log.info("\n======================\n");
```
xnull: what if namespace or/and table doesn't exist?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -75,17 +75,23 @@ public static void main(String[] args) {
             log.info("Successfully connected to {}", singleNodeEndpoint);
 
             CorfuStoreBrowser browser = new CorfuStoreBrowser(runtime);
+            CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table;
+            String namespace = opts.containsKey("--namespace") ?
```
xnull: let's use `Optional` here and avoid using null-s

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -75,17 +75,23 @@ public static void main(String[] args) {
...
+                    opts.get("--namespace").toString() : null;
+            String tableName = opts.containsKey("--tablename") ?
```
xnull: same here

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -75,17 +75,23 @@ public static void main(String[] args) {
...
+                    opts.get("--tablename").toString() : null;
             switch (operation) {
                 case "listTables":
-                    String namespace = opts.containsKey("--namespace") ?
-                        opts.get("--namespace").toString() : null;
                     browser.listTables(namespace);
                     break;
-
+                case "infoTable":
```
xnull: please use an enum instead of strings

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -81,10 +82,23 @@ public void listTables(String namespace)
...
+        String fullName = TableRegistry.getFullyQualifiedTableName(namespace, tablename);
+        UUID streamUUID = UUID.nameUUIDFromBytes(fullName.getBytes());
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(namespace, tablename);
+        log.info("Table " + tablename + " in namespace " + namespace
```
lixinchengdu: Use {} as a placeholder to make this statement clearer.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: Good point, I think its better to detect and fail gracefully. fixed in latest patch. thanks

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
hisundar: actually using null makes it easier to detect and fail gracefully.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
hisundar: done. thanks

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: done. thanks

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -19,6 +20,28 @@
...
+    }
+
+    private static void verifyNamespace(String namespace) {
+        if (namespace == null) {
+            throw new IllegalArgumentException("Please specify --namespace");
+        }
+    }
+
+    private static void verifyNamespaceAndTablename(String namespace, String tableName) {
+        if (namespace == null) {
```
pankti-m: we can just call verifyNamespace here

PR #2307, opened 2020-01-14
Change checkpoint initialization logic

vjeko:
This patch fixes a bug that gets triggers when IStreamView is being
consumed directly. When initializing a stream, the condition which
dictates when a checkpoint should be loaded needs to take into account
that a stream has been previously seek-ed to some valid address.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -3,6 +3,7 @@ package org.corfudb.runtime.view.stream;
 import com.google.common.annotations.VisibleForTesting;
 import lombok.Data;
 import lombok.Getter;
+import lombok.Setter;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -911,6 +913,8 @@ synchronized void seek(long globalAddress) {
      */
     @Data
     static class StreamCheckpoint {
+        public static StreamCheckpoint UNINITIALIZED = new StreamCheckpoint();
+        public static StreamCheckpoint INITIALIZED = new StreamCheckpoint();
```
zhangn49: Could they be final? Looks we don't change their references.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
```diff
@@ -174,6 +182,68 @@ public void testSuccessiveCheckpointTrim() throws Exception {
...
+     *
+     * @param token point at which to trim the address space.
+     */
+    private void trim(Token token) {
+        getRuntime().getAddressSpaceView().prefixTrim(token);
+        getRuntime().getAddressSpaceView().gc();
+        getRuntime().getAddressSpaceView().invalidateServerCaches();
+        getRuntime().getAddressSpaceView().invalidateClientCache();
+    }
+
```
zhangn49: I see `testCheckpointTrim` and `testSuccessiveCheckpointTrim` use these four lines to trim.
Can we replace them with trim()?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
vjeko: Yep.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointTrimTest.java
vjeko: Sounds good.

PR #2309, opened 2020-01-14
CorfuBrowser: Use Nullable, refactor code

hisundar:
Move all functionality out of the Main class.
Address review comments left out in previous review.
Add code coverage tests for all the new functions.

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
@@ -1,6 +1,7 @@
 package org.corfudb.browser;
 
 import java.util.Map;
+import java.util.Optional;
 
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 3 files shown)


hisundar:
> Miss some @return tags in Javadoc comments.

Thanks nan, let me address those.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -98,28 +84,25 @@ public static void main(String[] args) {
             log.info("Successfully connected to {}", singleNodeEndpoint);
 
             CorfuStoreBrowser browser = new CorfuStoreBrowser(runtime);
-            CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table;
-            String namespace = opts.containsKey("--namespace") ?
-                    opts.get("--namespace").toString() : null;
-            String tableName = opts.containsKey("--tablename") ?
-                    opts.get("--tablename").toString() : null;
+            String namespace = Optional.ofNullable(opts.get("--namespace"))
+                    .map(n -> n.toString())
```
pankti-m: will this not throw an NPR if namespace is null?

hisundar: No, that's why we use the ofNullable call right?

PR #2310, opened 2020-01-14
Get rid of a warning during unmap secondary indexes

xnull:
## Overview

Description:
Delete a log message in unmapSecondaryIndexes method to prevent generating Many massages in the log for any new entry in the corfu table.

Why should this be merged: 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -624,7 +624,7 @@ public V merge(K key, V value, BiFunction<? super V, ? super V, ? extends V>
     @SuppressWarnings("unchecked")
     protected void unmapSecondaryIndexes(K key, V value) {
         if (value == null) {
-            log.warn("Attempting to build an index with a null value, skipping.");
+            log.trace("Attempting to build an index with a null value, skipping.");
```
Maithem: This is not cheap and will be called very frequently. I think we should have a condition on the caller and if a secondary index doesn't exist then shouldn;t be called. Its much cleaner than calling it and "implementing a no-op"

xnull: It is already there https://github.com/CorfuDB/CorfuDB/blob/671c55568feba427e327bdb89fa1f4409ca5bb2a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java#L274

The `if` condition checks if an old value is present, if not (which means value == null) then we don't do anything.

muren102: Hi 
If the table has secondary index, the secondaryIndexes is not empty when the table is created. But when secondaryIndexes is not empty, if a new record with key value is created, e.g. in put method, V previous = mainMap.put(key, value); the previous value is null. so the method unmapSecondaryIndexes  is still invoked and the log will be printed. I think it would be better to check value null before invoke unmapSecondaryIndexes.

xnull: `unmapSecondaryIndexes` called by many methods. If I extract this check from the method I need to add it to all caller methods, which is not so good decision. Can you please explain why calling a method is a problem?

xnull:
As discussed offline, we don't need any messages in the log for those events

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
muren102: Hi
Calling the method is not a problem. But it may bring lots of trace logs "Attempting to build an index with a null value, skipping."

xnull: I still don't see how it can be a problem. Anyways we decided to get rid of it

Issue #2312, opened 2020-01-15
LayoutView: Prepare and Propose Issue


zhangn49:
## Overview

[CompletableFuture.anyOf](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#anyOf-java.util.concurrent.CompletableFuture...-) returns a new CompletableFuture, but following logic (line 151 - 154) could remove multiple cfs that completed with exceptions.

## Code Segment
https://github.com/CorfuDB/CorfuDB/blob/671c55568feba427e327bdb89fa1f4409ca5bb2a/runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java#L140-L154

## Explanation 
Let's assume we have prepareList [cf1, cf2, cf3], where cf1 is done normally, cf2 is completed with OutrankedException, cf3 is timeout.

If anyOf returns cf1, the stream operation will remove cf2 and cf3. In this case, it swallows TimeoutException, which makes timeout counter not accurate, and OutrankedException, which makes proposer stuck with a lower rank if OutrankedException is majority.

Besides, if some cfs are not completed yet, the while loop will keep consuming cpu time, because it does not remove completed cf1 from the list, and `CFUtils.getUninterruptibly(CompletableFuture.anyOf(prepareList))` returns immediately.

## More Concerns
For current implementation, we do not catch OutrankedException in prepare and propose. If the anyOf happens to return a cf with OutrankedException, prepare/propose phase will be terminated, which does not comply with Paxos.

medhavidhawan:
I am not sure if I understand this completely. I understand the concerns but removal of anyof makes the implementation susceptible to the slowest responder.

PR #2313, opened 2020-01-15
Fix the issue in LayoutView and some cleanup

zhangn49:
## Overview

Fix the issue in LayoutView and some cleanup

Description:

This misuse of CompletableFuture.anyOf in prepare and propose will result in:
1. Upper level can not observe OutrankedException and make its rank stuck.
2. Counters  {timeouts, wrongEpochRejected} are not accurate.
3. Waste cpu time. 

Detailed explanation is in #2312

Why should this be merged: 

I removed CompletableFuture.anyOf like #2270, and changed some variable names like 'response' -> 'promises/accepts' to make logic clear.

Related issue(s) (if applicable): #2312 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -126,66 +124,64 @@ public Layout prepare(long epoch, long rank)
...
+            if (promises >= getQuorumNumber()) {
+                break;
             }
 
-            // wait for someone to complete.
             try {
-                CFUtils.getUninterruptibly(CompletableFuture.anyOf(prepareList),
-                        OutrankedException.class, TimeoutException.class, NetworkException.class,
-                        WrongEpochException.class);
+                CFUtils.getUninterruptibly(cf, OutrankedException.class, TimeoutException.class,
```
Maithem: Great catch, this is so subtle. Using `CompletableFuture.anyOf` will return a `CompletableFuture` with a `CompletionException` on failure, but `getUninterruptibly` only operates on `Future` and thus only checks `ExecutionException`

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -126,66 +124,64 @@ public Layout prepare(long epoch, long rank)
...
-        while (true) {
-            // do we still have enough for a quorum?
-            if (prepareList.length < getQuorumNumber()) {
-                log.debug("prepare: Quorum unreachable, remaining={}, required={}", prepareList,
-                        getQuorumNumber());
-                throw new QuorumUnreachableException(prepareList.length, getQuorumNumber());
+        long outrankedRecord = Long.MIN_VALUE;
+        for (CompletableFuture cf : prepareList) {
+            if (promises >= getQuorumNumber()) {
+                break;
```
Maithem: Could this ever be true? 
- Promises are generated from a committed layout
- If the layout was committed then there must have been a quorum

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -126,66 +124,64 @@ public Layout prepare(long epoch, long rank)
                     return cf;
                 })
                 .toArray(CompletableFuture[]::new);
-        LayoutPrepareResponse[] acceptList;
+
+        long promises = 0L;
```
Maithem: Since these promises are generated from a java collection has is of size int, why not just use an int instead?

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -126,66 +124,64 @@ public Layout prepare(long epoch, long rank)
...
-                break;
+        // when quorum is unreachable, check if outranks is majority and throw OutrankedException
+        if (promises < getQuorumNumber()) {
+            log.debug("prepare: Quorum unreachable, promises={}, required={}, timeouts={}, " +
+                            "wrongEpochRejected={}, outranks={}",
+                    promises, getQuorumNumber(), timeouts, wrongEpochRejected, outranks);
+            if (outranks >= getQuorumNumber()) {
+                throw new OutrankedException(outrankedRecord);
+            } else {
+                throw new QuorumUnreachableException(Math.toIntExact(promises), getQuorumNumber());
```
Maithem: You can use an int here.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -126,66 +124,64 @@ public Layout prepare(long epoch, long rank)
...
-                    .toArray(LayoutPrepareResponse[]::new);
-
-            log.debug("prepare: Successful responses={}, needed={}, timeouts={}, "
-                            + "wrongEpochRejected={}",
-                    acceptList.length, getQuorumNumber(), timeouts, wrongEpochRejected);
-
-            if (acceptList.length >= getQuorumNumber()) {
-                break;
+        // when quorum is unreachable, check if outranks is majority and throw OutrankedException
+        if (promises < getQuorumNumber()) {
```
Maithem: I know its not your code, but it would be nice if you address this issue. 

Promises can be generated from a layout that is different from the one that is being used by `getQuorumNumber`.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -126,66 +124,64 @@ public Layout prepare(long epoch, long rank)
                     return cf;
                 })
                 .toArray(CompletableFuture[]::new);
-        LayoutPrepareResponse[] acceptList;
+
+        long promises = 0L;
         long timeouts = 0L;
+        long outranks = 0L;
         long wrongEpochRejected = 0L;
-        while (true) {
```
Maithem: This this while `loop` has been removed, are you assuming that upper layers will retry the phases again ?
Could this impact latency, when two fault detectors detect the same error and try to compete on updating the layout ?

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -126,66 +124,64 @@ public Layout prepare(long epoch, long rank)
...
             Layout layoutWithHighestRank = null;
-
-            for (LayoutPrepareResponse layoutPrepareResponse : list) {
-                if (layoutPrepareResponse.getRank() > highestReturnedRank) {
-                    highestReturnedRank = layoutPrepareResponse.getRank();
+            for (LayoutPrepareResponse layoutPrepareResponse : acceptedLayouts) {
+                if (layoutPrepareResponse.getRank() > highestRank) {
+                    highestRank = layoutPrepareResponse.getRank();
                     layoutWithHighestRank = layoutPrepareResponse.getLayout();
                 }
```
Maithem: You can replace this logic with `TreeSet<Layout> layouts = new TreeSet<>(Layout.LAYOUT_COMPARATOR);` and then `layouts.last()` will return the largest LayoutPrepareResponse

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
medhavidhawan: Why are we doing this ?
If the CFUtils is stuck on the wrong response it will wait even when the quorum is complete ?

zhangn49:
> I am not convinced with this cleanup yet. This takes away the concurrency from the protocol. I am not convinced, this can affect protocol convergence time. Maybe I am missing something.

@medhavidhawan Thanks for review.

The worst case does not change, since max time is bounded by RPC request timeout.
The best case is affected by the for loop, as you said it takes away the concurrency.

If we keep the while loop, and keep the anyOf function, we need to remove the returned CF by anyOf. Since we called `CFUtils.getUninterruptibly`, the returned CF must be done or completedExceptionally. In each iteration, we remove all CFs is done or completedExceptionally, and bump corresponding counters.

In this way, we can keep concurrency as well as observe all exceptions before getting quorum, and won't waste cpu cycles.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
zhangn49: > Great catch, this is so subtle. Using `CompletableFuture.anyOf` will return a `CompletableFuture` with a `CompletionException` on failure, but `getUninterruptibly` only operates on `Future` and thus only checks `ExecutionException`

In CompletableFuture.java, get will call this function.
`    /**
     * Reports result using Future.get conventions.
     */
    private static <T> T reportGet(Object r)
        throws InterruptedException, ExecutionException {
        if (r == null) // by convention below, null means interrupted
            throw new InterruptedException();
        if (r instanceof AltResult) {
            Throwable x, cause;
            if ((x = ((AltResult)r).ex) == null)
                return null;
            if (x instanceof CancellationException)
                throw (CancellationException)x;
            if ((x instanceof CompletionException) &&
                (cause = x.getCause()) != null)
                x = cause;
            throw new ExecutionException(x);
        }
        @SuppressWarnings("unchecked") T t = (T) r;
        return t;
    }`

So getUninterruptibly is fine here.
I also did an experiment that blocks all prepare requests and timeout counter was bumped once.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
zhangn49: Yes, I am assuming that, and I think this while loop assumes too.. While loop doesn't retry either.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
zhangn49: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
zhangn49: Should I note down the quorum number when sending requests? Then they are consistent.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
zhangn49: Not sure I understand correctly. 

Do you mean prepare/propose requests are sent to layout servers, which is from a committed layout?

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -112,93 +112,95 @@ public void updateLayout(Layout layout, long rank)
...
-                        OutrankedException.class, TimeoutException.class, NetworkException.class,
-                        WrongEpochException.class);
-            } catch (TimeoutException | NetworkException e) {
-                timeouts++;
-            } catch (WrongEpochException we) {
-                wrongEpochRejected++;
+                CFUtils.getUninterruptibly(CompletableFuture.anyOf(prepareList));
+            } catch (Exception e) {
+                log.trace("prepare: ", e);
             }
```
zhangn49: Is there a more elegant way to get a cf and ignore its exception?

PR #2315, opened 2020-01-17
Server Configuration Revamp

Maithem:
## Overview

This patch introduces a ServerConfiguration type that captures
all server related configuration. This clean up removes the
old style of configuration of passing around Map<String, Object>
and having multiple parsers to create sub-configurations for
different components. This patch also removes the docopt
madness of creating configuration via 10s of lines of string
formating. This change also removes the ServerContext usage
from classes that don't need it.

Why should this be merged:  
This patch is needed to pave the way to allow for server
configuration via config files.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -643,16 +591,11 @@ private EventLoopGroup getNewBossGroup() {
...
-            ? Runtime.getRuntime().availableProcessors() * 2
-            : requestedThreads;
-        EventLoopGroup group = getChannelImplementation().getGenerator()
-            .generate(numThreads, threadFactory);
+        EventLoopGroup group = configuration.getChannelImplementation().getGenerator()
+            .generate(6, threadFactory);
 
         log.info("getClientGroup: Type {} with {} threads",
-            group.getClass().getSimpleName(), numThreads);
+            group.getClass().getSimpleName(), 1);
```
zhangn49: Confused.. Why log with 1 thread while generating with 6 threads?

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
```diff
@@ -13,92 +15,84 @@
...
+                .setChannelImplementation(implementation);
+
 
         // Provide the server with event loop groups
-        if (implementation.equals("local")) {
-            builder.put("client", TestThreadGroups.NETTY_CLIENT_GROUP.get());
-            builder.put("boss", TestThreadGroups.NETTY_BOSS_GROUP.get());
-            builder.put("worker", TestThreadGroups.NETTY_CLIENT_GROUP.get());
+        if (implementation == ChannelImplementation.LOCAL) {
+            conf.setTestClientEventLoop(TestThreadGroups.NETTY_CLIENT_GROUP.get().next())
```
zhangn49: > next() Returns one of the EventExecutors managed by this EventExecutorGroup.

So does `TestThreadGroups.NETTY_CLIENT_GROUP ` manage one EventExecutor?

xnull:
I strongly recommend using YAML format

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -198,38 +57,54 @@ public static void main(String[] args) {
         }
     }
 
-    private static void startServer(String[] args) {
-        // Parse the options given, using docopt.
-        Map<String, Object> opts = new Docopt(USAGE)
-                .withVersion(GitRepositoryState.getRepositoryState().describe)
-                .parse(args);
+    private static void startServer(String[] args) throws ParseException {
```
xnull: can we have a separate PR for cmd line refactoring?

Maithem:
> I strongly recommend using YAML format

Sure, but this PR doesn't make a determination about the config file. If you look at `PropertiesConfiguration` it actually can parse `yaml` in addition to different format.s

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
Maithem: The point of this PR is to load command lines into a Configuration type, what's the point of doing them in two steps ?

If we split this over two PRs, then we need to have a converter from Map<String, Obect> to ServerConfigurations and then another patch to remove that an remplace it with Apache CLI options?

PR #2317, opened 2020-01-21
PersistedStreamingMap SIGSEGV Fix

Maithem:
This patch fixes a bug where the iterator is accessed after it
has been closed, which ends up causing a segfault.

## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
@@ -8,19 +8,17 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
 import org.corfudb.util.serializer.ISerializer;
+import org.rocksdb.CompactionOptionsUniversal;
+import org.rocksdb.CompressionType;
 import org.rocksdb.ReadOptions;
 import org.rocksdb.RocksDB;
 import org.rocksdb.Options;
 import org.rocksdb.RocksDBException;
-import org.rocksdb.RocksIterator;
 
 import java.io.File;
 import java.nio.file.Path;
-import java.util.AbstractMap;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Stream;
@@ -38,6 +36,37 @@ public class PersistedStreamingMap<K, V> implements ContextAwareMap<K, V> {
 
     static {
         RocksDB.loadLibrary();
+    }
+
+    /**
+     * A set of options defined for disk-backed {@link PersistedStreamingMap}.
+     *
+     * For a set of options that dictate RocksDB memory usage can be found here:
+     * https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB
+     *
+     * Block Cache:  Which can be set via Options::setTableFormatConfig.
+     *               Out of box, RocksDB will use LRU-based block cache
+     *               implementation with 8MB capacity.
+     * Index/Filter: Is a function of the block cache. Generally it inflates
+     *               the block cache by about 50%. The exact number can be
+     *               retrieved via "rocksdb.estimate-table-readers-mem"
+     *               property.
+     * Write Buffer: Also known as memtable is defined by the ColumnFamilyOptions
+     *               option. The default is 64 MB.
+     */
+    public static Options getPersistedStreamingMapOptions() {
+        final int maxSizeAmplificationPercent = 50;
+        final Options options = new Options();
+
+        options.setCreateIfMissing(true);
+        options.setCompressionType(CompressionType.LZ4_COMPRESSION);
+
+        // Set a threshold at which full compaction will be triggered.
+        // This is important as it purges tombstoned entries.
+        final CompactionOptionsUniversal compactionOptions = new CompactionOptionsUniversal();
+        compactionOptions.setMaxSizeAmplificationPercent(maxSizeAmplificationPercent);
+        options.setCompactionOptionsUniversal(compactionOptions);
+        return options;
     }
 
     private final ContextAwareMap<K, V> optimisticMap = new StreamingMapDecorator<>();
@@ -58,79 +87,6 @@ public class PersistedStreamingMap<K, V> implements ContextAwareMap<K, V> {
         }
         this.serializer = serializer;
         this.corfuRuntime = corfuRuntime;
-    }
-
-    /**
-     * A Java compatible {@link RocksIterator} implementation
-     */
-    public class RocksDbIterator implements Iterator<Entry<K, V>> {
-        private RocksIterator iterator;
-        private Entry<K, V> current;
-        private Entry<K, V> next;
-
-        RocksDbIterator(RocksIterator iterator) {
-            this.iterator = iterator;
-        }
-
-        /**
-         * Ensure that this iterator is operating under the correct assumptions.
-         */
-        private void checkInvariants() {
-            // RocksDB does not support multi-threaded access.
-            // If the iterator was created by some thread, it also has to be consumed by it.
-            if (!iterator.isOwningHandle()) {
-                throw new IllegalStateException("Detected multi-threaded access to this iterator.");
-            }
-
-            try {
-                iterator.status();
-            } catch (RocksDBException e) {
-                throw new UnrecoverableCorfuError(
-                        "There was an error reading the persisted map.", e);
-            }
-        }
-
-        /**
-         * {@inheritDoc}
-         */
-        @Override
-        public boolean hasNext() {
-            // If we have the next element pipelined, go ahead and return true.
-            if (next != null) {
-                return true;
-            }
-
-            // If the iterator is valid, this means that the next entry exists.
-            checkInvariants();
-            if (iterator.isValid()) {
-                // Go ahead and cache that entry.
-                next = new AbstractMap.SimpleEntry(
-                        serializer.deserialize(Unpooled.wrappedBuffer(iterator.key()), corfuRuntime),
-                        serializer.deserialize(Unpooled.wrappedBuffer(iterator.value()), corfuRuntime));
-                // Advance the underlying iterator.
-                iterator.next();
-            } else {
-                // If there is no more elements to consume, we should release the resources.
-                iterator.close();
-            }
-
-            return next != null;
-        }
-
-        /**
-         * {@inheritDoc}
-         */
-        @Override
-        public Entry<K, V> next() {
-            if (hasNext()) {
-                current = next;
-                next = null;
-                return current;
-            } else {
-                throw new NoSuchElementException();
-            }
-
-        }
     }
 
     /**
@@ -313,14 +269,10 @@ public class PersistedStreamingMap<K, V> implements ContextAwareMap<K, V> {
      */
     @Override
     public Stream<Entry<K, V>> entryStream() {
-        final ReadOptions readOptions = new ReadOptions();
-        // If ReadOptions.snapshot is given, the iterator will return data as of the snapshot.
-        // If it is nullptr, the iterator will read from an implicit snapshot as of the time the
-        // iterator is created. The implicit snapshot is preserved by pinning resource.
-        readOptions.setSnapshot(null);
-        final RocksIterator rocksIterator = rocksDb.newIterator(readOptions);
-        rocksIterator.seekToFirst();
-        return Streams.stream(new RocksDbIterator(rocksIterator));
+        final RocksDbEntryIterator entryIterator = new RocksDbEntryIterator(rocksDb, serializer);
+        Stream<Entry<K, V>> resStream = Streams.stream(entryIterator);
+        resStream.onClose(entryIterator::close);
+        return resStream;
     }
 
     /**
--- a/runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
@@ -1,0 +1,96 @@
+package org.corfudb.runtime.collections;
+
+import io.netty.buffer.Unpooled;
+import org.corfudb.util.serializer.ISerializer;
+import org.rocksdb.ReadOptions;
+import org.rocksdb.RocksDB;
+
+import javax.annotation.concurrent.NotThreadSafe;
+import java.util.AbstractMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+/**
+ *
+ * A wrapper class that provides an iterator over the RocksDb map entries. It is compatible with Java's {@link Iterator}.
+ *
+ * Created by Maithem on 1/21/20.
+ */
+@NotThreadSafe
+public class RocksDbEntryIterator<K, V> implements Iterator<Map.Entry<K, V>>, AutoCloseable {
+
+    /**
+     * A reference to the underlying RocksDb iterator
+     */
+    final private WrappedRocksIterator wrappedRocksIterator;
+
+    /**
+     * Serializer to serialize/deserialize the key/values
+     */
+    final private ISerializer serializer;
+
+    /**
+     * place holder for the current value
+     */
+    private Map.Entry<K, V> next;
+
+    final private ReadOptions readOptions;
+
+    public RocksDbEntryIterator(RocksDB rocksDB, ISerializer serializer) {
+        // Start iterator at the current snapshot
+        readOptions = new ReadOptions();
+        readOptions.setSnapshot(null);
+        this.wrappedRocksIterator = new WrappedRocksIterator(rocksDB.newIterator(readOptions));
+        this.serializer = serializer;
+        wrappedRocksIterator.seekToFirst();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean hasNext() {
+        if (next == null && wrappedRocksIterator.isOpen() && wrappedRocksIterator.isValid()) {
+            // Retrieve entry if it exists and move the iterator
+            next = new AbstractMap.SimpleEntry(
+                    serializer.deserialize(Unpooled.wrappedBuffer(wrappedRocksIterator.key()), null),
+                    serializer.deserialize(Unpooled.wrappedBuffer(wrappedRocksIterator.value()), null));
+            wrappedRocksIterator.next();
+        }
+
+        if (next == null && wrappedRocksIterator.isOpen()) {
+            // close the iterator if it has fully consumed.
+            wrappedRocksIterator.close();
+            readOptions.close();
+        }
+
+        return next != null;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Map.Entry<K, V> next() {
+        if (hasNext()) {
+            Map.Entry<K, V> res = next;
+            next = null;
+            return res;
+        } else {
+            throw new NoSuchElementException();
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void close() {
+        // Release the underlying RocksDB resources
+        if (wrappedRocksIterator.isOpen()) {
+            wrappedRocksIterator.close();
+            readOptions.close();
+        }
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/collections/WrappedRocksIterator.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/WrappedRocksIterator.java
@@ -1,0 +1,140 @@
+package org.corfudb.runtime.collections;
+
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
+import org.rocksdb.RocksDBException;
+import org.rocksdb.RocksIterator;
+import org.rocksdb.RocksIteratorInterface;
+
+/**
+ *
+ * A wrapper class that makes a access to the RocksIterator safe: the rocks library does some checks but they
+ * are enforced via assert, but assert checking is disabled on many jvms by default, hence the explicit checking.
+ *
+ * Created by Maithem on 1/24/20.
+ */
+public class WrappedRocksIterator implements RocksIteratorInterface {
+
+    private final RocksIterator iterator;
+    public WrappedRocksIterator(RocksIterator iterator) {
+        this.iterator = iterator;
+    }
+
+    /**
+     * A helper method that is called before accessing the iterator.
+     * This is necessary to make sure that the iterator hasn't been closed.
+     */
+    private void checkHandle() {
+        if (!iterator.isOwningHandle()) {
+            throw new IllegalStateException("RocksIterator has been closed");
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isValid() {
+        checkHandle();
+        boolean res =  iterator.isValid();
+        status();
+        return res;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void seekToFirst() {
+        checkHandle();
+        iterator.seekToFirst();
+        status();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void seekToLast() {
+        throw new UnsupportedOperationException("seekToLast");
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void seek(byte[] target) {
+        throw new UnsupportedOperationException("seek");
+    }
+
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void seekForPrev(byte[] target){
+        throw new UnsupportedOperationException("seekForPrev");
+    }
+
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void next() {
+       checkHandle();
+       iterator.next();
+       status();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void prev() {
+        throw new UnsupportedOperationException("prev");
+    }
+
+    /**
+     * Returns the key of the current position of the iterator.
+     */
+    public byte[] key() {
+        checkHandle();
+        byte[] res = iterator.key();
+        status();
+        return res;
+    }
+
+    /**
+     * Returns the value of the current position of the iterator.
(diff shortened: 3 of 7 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -40,6 +45,37 @@
...
+     *               Out of box, RocksDB will use LRU-based block cache
+     *               implementation with 8MB capacity.
+     * Index/Filter: Is a function of the block cache. Generally it infates
+     *               the block cache by about 50%. The exact number can be
+     *               retrieved via "rocksdb.estimate-table-readers-mem"
+     *               property.
+     * Write Buffer: Also known as memtable is defined by the ColumnFamilyOptions
+     *               option. The default is 64 MB.
+     */
+    public static Options getPersistedStreamingMapOptions() {
```
xnull: Can you make all these parameters configurable? 
At least pass `Options` as a parameter in the constructor?
Or any other way that helps make all the parameters configurable

vjeko:
Where was this being observed? AFAIK, this feature is not being used ATM.

Maithem:
> Where was this being observed? AFAIK, this feature is not being used ATM.

Which feature are you referring to `PersistedStreamingMap` or `entryStream` ?

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
Maithem: Ok.

vjeko:
I am running your test on the master branch, and getting the following assertion error:

```
java.lang.IllegalStateException: Detected multi-threaded access to this iterator.

	at org.corfudb.runtime.collections.PersistedStreamingMap$RocksDbIterator.checkInvariants(PersistedStreamingMap.java:82)
	at org.corfudb.runtime.collections.PersistedStreamingMap$RocksDbIterator.hasNext(PersistedStreamingMap.java:104)
	at java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1811)
	at java.util.Spliterators$1Adapter.hasNext(Spliterators.java:681)
	at com.google.common.collect.Iterators$4.hasNext(Iterators.java:600)
	at org.corfudb.runtime.collections.CorfuTableTest.PersistedCorfuTableTests(CorfuTableTest.java:277)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.corfudb.AbstractCorfuTest$1$1.lambda$evaluate$0(AbstractCorfuTest.java:106)
	at java.lang.Thread.run(Thread.java:748)
```

Can you explain in more detail what the bug is?

Maithem:
> I am running your test on the master branch, and getting the following assertion error:
> 
> ```
> java.lang.IllegalStateException: Detected multi-threaded access to this iterator.
> 
> 	at org.corfudb.runtime.collections.PersistedStreamingMap$RocksDbIterator.checkInvariants(PersistedStreamingMap.java:82)
> 	at org.corfudb.runtime.collections.PersistedStreamingMap$RocksDbIterator.hasNext(PersistedStreamingMap.java:104)
> 	at java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1811)
> 	at java.util.Spliterators$1Adapter.hasNext(Spliterators.java:681)
> 	at com.google.common.collect.Iterators$4.hasNext(Iterators.java:600)
> 	at org.corfudb.runtime.collections.CorfuTableTest.PersistedCorfuTableTests(CorfuTableTest.java:277)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.lang.reflect.Method.invoke(Method.java:498)
> 	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
> 	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
> 	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
> 	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
> 	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
> 	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
> 	at org.corfudb.AbstractCorfuTest$1$1.lambda$evaluate$0(AbstractCorfuTest.java:106)
> 	at java.lang.Thread.run(Thread.java:748)
> ```
> 
> Can you explain in more detail what the bug is?

There are two bugs, one that where `hasNext` closes the iterator and then the subsequent call fails with `IllegalStateException` (the bug which the Unit Test exercise). The second issue, is there is a race between closing the iterator and calling `status()`, It doesn't always produce an `AssertionError`, it can segfault. I'm guessing its bug in `RocksJava` that's related to `iterator.close()` and the GC

vjeko:
> There are two bugs, one that where `hasNext` closes the iterator and then the subsequent call fails with `IllegalStateException` (the bug which the Unit Test exercise). The second issue, is there is a race between closing the iterator and calling `status()`, It doesn't always produce an `AssertionError`, it can segfault. I'm guessing its bug in `RocksJava` that's related to `iterator.close()` and the GC

How can there be a race? Race implies multi-threaded access. `PersistedStreamingMap` is meant to be consumed in a single threaded environment.

vjeko:
> Which feature are you referring to `PersistedStreamingMap` or `entryStream` ?

I am referring to `PersistedStreamingMap`. The code path that this patch touches is not being used anywhere except for IDPS.

Maithem:
> PersistedStreamingMap



> > There are two bugs, one that where `hasNext` closes the iterator and then the subsequent call fails with `IllegalStateException` (the bug which the Unit Test exercise). The second issue, is there is a race between closing the iterator and calling `status()`, It doesn't always produce an `AssertionError`, it can segfault. I'm guessing its bug in `RocksJava` that's related to `iterator.close()` and the GC
> 
> How can there be a race? Race implies multi-threaded access. `PersistedStreamingMap` is meant to be consumed in a single threaded environment.

There can be a race between a user thread and the GC thread. For example, the `RocksDbIterator` can have unsafe references to objects inside `RocksDB` and when calling methods on that iterator, it will try to deference and object that has been de-allocated (i.e. "finalized" by the GC).

Maithem:
> > There are two bugs, one that where `hasNext` closes the iterator and then the subsequent call fails with `IllegalStateException` (the bug which the Unit Test exercise). The second issue, is there is a race between closing the iterator and calling `status()`, It doesn't always produce an `AssertionError`, it can segfault. I'm guessing its bug in `RocksJava` that's related to `iterator.close()` and the GC
> 
> How can there be a race? Race implies multi-threaded access. `PersistedStreamingMap` is meant to be consumed in a single threaded environment.

Maithem:
> > Which feature are you referring to `PersistedStreamingMap` or `entryStream` ?
> 
> I am referring to `PersistedStreamingMap`. The code path that this patch touches is not being used anywhere except for IDPS.

This path can be invoked via https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java#L161

vjeko:
> There can be a race between a user thread and the GC thread. For example, the `RocksDbIterator` can have unsafe references to objects inside `RocksDB` and when calling methods on that iterator, it will try to deference and object that has been de-allocated (i.e. "finalized" by the GC).

Maybe I am not fully understanding the use case. 

In order for the JVM to garbage collect `RocksDbIterator`, the object reference count has to be zero. However, if reference count is zero, this means that it is impossible for anyone to call `Iterator` methods since no one has a reference to it.

Maithem:
> > There can be a race between a user thread and the GC thread. For example, the `RocksDbIterator` can have unsafe references to objects inside `RocksDB` and when calling methods on that iterator, it will try to deference and object that has been de-allocated (i.e. "finalized" by the GC).
> 
> Maybe I am not fully understanding the use case.
> 
> In order for the JVM to garbage collect `RocksDbIterator`, the object reference count has to be zero. However, if reference count is zero, this means that it is impossible for anyone to call `Iterator` methods since no one has a reference to it.

Not the  `RocksDbIterator` itself, but I was guessing some underlying resources. I haven't looked into the native code, but it seems like its encountering a segfault where it should throw an AssertionError. Possibly a race between underlying `dispose` (caused by close) and the `status`. Still need to investigate, but with this patch the issue doesn't seem to happen. 

Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  0x00007f58a40000f8
J 6323  org.rocksdb.RocksIterator.status0(J)V (0 bytes) @ 0x00007f59295d25e8 [0x00007f59295d2540+0xa8]
J 6322 C1 org.rocksdb.AbstractRocksIterator.status()V (30 bytes) @ 0x00007f592958475c [0x00007f5929584640+0x11c]
v  ~StubRoutines::call_stub
V  [libjvm.so+0x68825b]  JavaCalls::call_helper(JavaValue*, methodHandle*, JavaCallArguments*, Thread*)+0xddb
V  [libjvm.so+0x6d20d8]  jni_invoke_nonstatic(JNIEnv_*, JavaValue*, _jobject*, JNICallType, _jmethodID*, JNI_ArgumentPusher*, Thread*) [clone .constprop.110]+0x208
V  [libjvm.so+0x6d2fe5]  jni_CallVoidMethodA+0xd5
C  [libjdwp.so+0x1c014]  invoker_doInvoke+0xd94
C  [libjdwp.so+0x145ad]  reportEvents.part.2+0xad
C  [libjdwp.so+0x1498a]  event_callback+0x1da
C  [libjdwp.so+0x177f8]  cbBreakpoint+0xc8
V  [libjvm.so+0x772035]  JvmtiExport::post_raw_breakpoint(JavaThread*, Method*, unsigned char*)+0x315
V  [libjvm.so+0x67aedf]  InterpreterRuntime::_breakpoint(JavaThread*, Method*, unsigned char*)+0x1f
j  org.corfudb.runtime.collections.PersistedStreamingMap$RocksDbIterator.checkInvariants()V+36
J 6316 C1 org.corfudb.runtime.collections.PersistedStreamingMap$RocksDbIterator.hasNext()Z (122 bytes) @ 0x00007f5929b7772c [0x00007f5929b77660+0xcc]

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Ensure that this iterator is still valid.
+     */
+    private void checkInvariants() {
+        try {
+            iterator.status();
+        } catch (RocksDBException e) {
+            throw new UnrecoverableCorfuError("There was an error reading the persisted map.", e);
```
hisundar: should the iterator be closed in case of an exception?

Maithem: It doesn't matter because that exception should crash the process. On exit the resources should be cleaned.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -318,9 +281,11 @@ public void clear() {
         // If it is nullptr, the iterator will read from an implicit snapshot as of the time the
         // iterator is created. The implicit snapshot is preserved by pinning resource.
         readOptions.setSnapshot(null);
-        final RocksIterator rocksIterator = rocksDb.newIterator(readOptions);
-        rocksIterator.seekToFirst();
-        return Streams.stream(new RocksDbIterator(rocksIterator));
+        final RocksDbEntryIterator entryIterator = new RocksDbEntryIterator(rocksDb.newIterator(readOptions), serializer);
+        Stream<Entry<K, V>> resStream = StreamSupport.stream(
+                Spliterators.spliteratorUnknownSize(entryIterator, Spliterator.ORDERED), false);
+        resStream.onClose(entryIterator::close);
```
Maithem: Stream consumers should use try-resource to invoke the close method.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.AbstractMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+/**
+ *
+ * Created by Maithem on 1/21/20.
+ */
+
```
vjeko: Please remove this and add a proper JavaDoc comment.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
```diff
@@ -0,0 +1,84 @@
...
+        iterator.seekToFirst();
+        checkInvariants();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean hasNext() {
+        while (next == null && iterator.isValid()) {
```
vjeko: Why are you looping here? Can you give a use-case?

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void close() {
+        // Release the underlying RocksDB resources
+        if (!iterator.isOwningHandle()) {
```
vjeko: `close()` should be idempotent and an extra check can be done here to see if the iterator has been closed before. From the documentation:

```
     * However, implementers of this interface are strongly encouraged
     * to make their {@code close} methods idempotent.
```

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
```diff
@@ -0,0 +1,84 @@
...
+            return res;
+        } else {
+            throw new NoSuchElementException();
+        }
+    }
+
+    /**
+     * Ensure that this iterator is still valid.
+     */
+    private void checkInvariants() {
```
vjeko: All methods should check for multi-threaded access.  Also, this method should be executed on every API call.

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
```diff
@@ -245,4 +255,47 @@ public void snapshotInvariant() {
         final Stream<Map.Entry<Integer, Integer>> result = map.entryStream();
         result.forEach(e -> map.put(new Random().nextInt(), 0));
     }
+
+    @Test
+    public void PersistedCorfuTableTests() {
```
vjeko: This test is using checkpointer specific logic, so it should be moved to `CheckpointTest`. A unit test should be checking for more specific functionality -- for example -- calling `hasNext()` on a closed iterator behaves as expected.

vjeko:
> Not the `RocksDbIterator` itself, but I was guessing some underlying resources. I haven't looked into the native code

Can you share the core dump?

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -40,6 +45,37 @@
         RocksDB.loadLibrary();
     }
 
+    /**
+     * A set of options defined for disk-backed {@link CorfuTable}.
```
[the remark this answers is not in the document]
Maithem: Yes.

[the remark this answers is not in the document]
Maithem: Changed.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
Maithem: Its kind of messy, because its passed to the constructor through a supplier. CorfuTable can be refactored in a different patch.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -318,9 +281,11 @@ public void clear() {
         // If it is nullptr, the iterator will read from an implicit snapshot as of the time the
         // iterator is created. The implicit snapshot is preserved by pinning resource.
         readOptions.setSnapshot(null);
-        final RocksIterator rocksIterator = rocksDb.newIterator(readOptions);
-        rocksIterator.seekToFirst();
-        return Streams.stream(new RocksDbIterator(rocksIterator));
+        final RocksDbEntryIterator entryIterator = new RocksDbEntryIterator(rocksDb.newIterator(readOptions), serializer);
+        Stream<Entry<K, V>> resStream = StreamSupport.stream(
+                Spliterators.spliteratorUnknownSize(entryIterator, Spliterator.ORDERED), false);
```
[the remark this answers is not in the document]
Maithem: Not needed.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
Maithem: It is
```
    public void close() {
        if (this.owningHandle_.compareAndSet(true, false)) {
            this.disposeInternal();
        }

    }
```

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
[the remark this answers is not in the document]
Maithem: Done.

Maithem: Its not checkpoint specific, although the checkpointer uses a CorfuTable that doesn't work.

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
```diff
@@ -245,4 +255,47 @@ public void snapshotInvariant() {
...
+        for (List<Map.Entry<String, String>> partition : partitions) {
+            for (Map.Entry<String, String> entry : partition) {
+                result.put(entry.getKey(), entry.getValue());
+            }
+        }
+
+        assertThat(diskBackedMap.size()).isEqualTo(numKeys);
+        assertThat(result.size()).isEqualTo(numKeys);
+        for (Map.Entry<String, String> entry : result.entrySet()) {
+            assertThat(diskBackedMap.get(entry.getKey())).isEqualTo(diskBackedMap.get(entry.getKey()));
```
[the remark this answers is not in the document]
Maithem: Good catch.

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
vjeko: This code was pretty much copied from `CheckpointWriter::appendObjectState`. Can you please add proper unit tests? Both positive and negative tests. You can add them to `DiskBackedCorfuClientTest`.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -318,9 +274,13 @@ public void clear() {
...
         readOptions.setSnapshot(null);
-        final RocksIterator rocksIterator = rocksDb.newIterator(readOptions);
-        rocksIterator.seekToFirst();
-        return Streams.stream(new RocksDbIterator(rocksIterator));
+        final RocksDbEntryIterator entryIterator =
+                new RocksDbEntryIterator(rocksDb.newIterator(readOptions), serializer);
+        // It's necessary to use the RocksDbEntryIterator as try-with resource to ensure
+        // the iterator is close, otherwise underlying RocksDB resources can leak
+        Stream<Entry<K, V>> resStream = Streams.stream(entryIterator);
+        resStream.onClose(entryIterator::close);
```
WenbinZhu: According to this: https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/ReadOptions.java#L11
I think we need to call `readOptions.close()` before returning to prevent memory leak. (dispose is deprecated, so using close())

runtime/src/main/java/org/corfudb/runtime/collections/CheckedRocksIterator.java
```diff
@@ -0,0 +1,139 @@
...
+public class CheckedRocksIterator implements RocksIteratorInterface {
+
+    private final RocksIterator iterator;
+    public CheckedRocksIterator(RocksIterator iterator) {
+        this.iterator = iterator;
+    }
+
+    private void checkHandle() {
+        if (!iterator.isOwningHandle()) {
+            throw new IllegalStateException("Handle is invalid");
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
Maithem: Didn't know about this. Thanks!

It will get released through finalized, but we shouldn't relay on that.

Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
```diff
@@ -0,0 +1,83 @@
...
+        this.serializer = serializer;
+        iterator.seekToFirst();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean hasNext() {
+        if (next == null && checkedIterator.isValid()) {
```
vjeko: If we detect that the iterator has reached the end, we should go ahead and relinquish the resources immediately, rather than relying on `try-with-resource` or GC. This is a more robust solution since it is not guaranteed that the client will use `try-with-resource`.

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
Maithem: This test belongs to CheckpointTest. Also, added more tests.

runtime/src/main/java/org/corfudb/runtime/collections/WrappedRocksIterator.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * A helper method that is called before accessing the iterator.
+     * This is necessary to make sure that access happens on an owned handle.
+     */
+    private void checkHandle() {
+        if (!iterator.isOwningHandle()) {
+            throw new IllegalStateException("RocksIterator handle is not valid");
+        }
```
vjeko: This message should be more descriptive. The only was for the handle to be invalid is either if:

1) There is some other thread trying to access the iterator.
2) The iterator has been previously closed.

The message should reflect the above conditions.

runtime/src/main/java/org/corfudb/runtime/collections/WrappedRocksIterator.java
```diff
@@ -0,0 +1,143 @@
...
+        return res;
+    }
+
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void status() throws RocksDBException {
+        throw new UnsupportedOperationException("status");
```
vjeko: `checkStatus()` and `status()` should behave the same. The implementation of this interface can be less restrictive, thus you are not required to throw `RocksDBException`, allowing you to merge these two methods.

runtime/src/main/java/org/corfudb/runtime/collections/WrappedRocksIterator.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Verifies that the iterator is in a valid state.
+     */
+    private void checkStatus() {
+        try {
+            iterator.status();
```
vjeko: You should document what constitutes a valid status. You can take a look at:

https://github.com/facebook/rocksdb/wiki/Iterator

And reference some parts of the Error Handling section.

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
```diff
@@ -1,21 +1,31 @@
...
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import org.assertj.core.api.Assertions;
 import org.assertj.core.data.MapEntry;
+import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
+import org.corfudb.runtime.object.ICorfuVersionPolicy;
 import org.corfudb.runtime.view.AbstractViewTest;
+import org.corfudb.util.serializer.Serializers;
```
vjeko: None of these changes are necessary.

runtime/src/main/java/org/corfudb/runtime/collections/RocksDbEntryIterator.java
Maithem: Not really, that assumes all stream consumers have to consume the full stream, which is an incorrect assumption. Consumers have to the stream whether explicitly or with try-with-resources

PR #2321, opened 2020-01-22
Add epoch when building local NodeState

zhangn49:
## Overview

Description:

Add epoch when building local NodeState, and mark connectivity as VisibleForTesting only

Why should this be merged: 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/management/ClusterStateCollector.java
```diff
@@ -122,9 +122,12 @@ private NodeState collectLocalNodeState(long epoch, SequencerMetrics sequencerMe
         });
 
         //Build local NodeState based on pings.
-        NodeConnectivity localConnectivity = NodeConnectivity.connectivity(
-                localEndpoint, ImmutableMap.copyOf(localNodeConnections)
-        );
+        NodeConnectivity localConnectivity = NodeConnectivity.builder()
```
xnull: good catch. 
 - Also please add a test (or an additional check) in `ClusterStateCollectorTest` which checks that the epoch is set.
 - Also would be great to have an additional check inside `NodeConnectivity` that the epoch is set. The easiest way could be to change the primitive value to a class and check that it is not null:
  ```
    @Getter
    @NonNull
    private final Long epoch;
```

File: runtime/src/main/java/org/corfudb/protocols/wireprotocol/failuredetector/NodeConnectivity.java [hunk 2]
```diff
@@ -148,6 +149,7 @@ public int compareTo(NodeConnectivity other) {
      * @param connectivity connectivity matrix
      * @return NodeConnectivity
      */
+    @VisibleForTesting
```
xnull: it is already public, no need for `@VisibleForTesting`

infrastructure/.flattened-pom.xml
```diff
@@ -0,0 +1,138 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
xiaoqin2012: This file is not needed.

PR #2322, opened 2020-01-23
CorfuStore: Record if table is opened as disk based

hisundar:
This is needed for the checkpointer to make sure that
a disk based table is opened as a disk based table and
not an in-memory table.

### DIFF
--- a/runtime/proto/corfu_store_metadata.proto
+++ b/runtime/proto/corfu_store_metadata.proto
@@ -28,8 +28,10 @@ message TableDescriptors {
 // Metadata.
 message TableMetadata {
     TableName table_name = 1;
+
     // True if the CorfuStore is disk based and False if it can be contained in memory.
     bool disk_based = 2;
+
     // True if these streams are to be cached, False otherwise.
     bool cache = 3;
 }
--- a/test/src/test/java/org/corfudb/integration/CorfuStoreIT.java
+++ b/test/src/test/java/org/corfudb/integration/CorfuStoreIT.java
@@ -1,31 +1,44 @@
 package org.corfudb.integration;
 
+import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 import com.google.protobuf.DynamicMessage;
 
 import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata;
 import org.corfudb.runtime.MultiCheckpointWriter;
 import org.corfudb.runtime.collections.CorfuDynamicKey;
 import org.corfudb.runtime.collections.CorfuDynamicRecord;
 import org.corfudb.runtime.collections.CorfuRecord;
 import org.corfudb.runtime.collections.CorfuStore;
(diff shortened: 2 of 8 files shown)


runtime/proto/corfu_store_metadata.proto
```diff
@@ -30,8 +30,12 @@ message TableMetadata {
     TableName table_name = 1;
     // True if the CorfuStore is disk based and False if it can be contained in memory.
     bool disk_based = 2;
+
+    // Full path to disk based location if this table is disk based
+    string disk_based_path = 3;
```
Maithem: The metadata is for a distributed object, but the path of is local information. Does it make sense to store it in this registry ?

hisundar: Technically true, but in practice all distributed clients pass the same path. 
In any way we require a way to communicate the location of the disk path from the client's table to the checkpointer so the latter can use the exact same path as a prefix, otherwise we risk in IO overload if the disk path points to a separate disk.

Maithem: I see.

But what happens if the client and the checkpointer run on the same machine. Wouldn't they both point to the same path? 

Since during open the DB in the destination path is destroyed https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java#L54 , this will crash one of the clients.

hisundar: This path location added in this PR is info-only. It is up to the clients to ensure that they do not pass conflicting paths and open the same table from multiple clients. For example the checkpointer client will suffix the table with a subfolder like _compactor_tablename/ to ensure that it does not conflict. We can at best add a few comments explaining this.

test/src/test/java/org/corfudb/integration/CorfuStoreIT.java
```diff
@@ -232,4 +244,122 @@ public void readDataWithDynamicMessages() throws Exception {
...
+            boolean diskBased = tableRegistryCT.get(tableName).getMetadata().getDiskBased();
+            if (diskBased) {
+                final Path persistedCacheLocation = Paths.get(tempDiskPath + tableName.getTableName());
+                final Options options = new Options().setCreateIfMissing(true);
+                final Supplier<StreamingMap<CorfuDynamicKey, CorfuDynamicRecord>> mapSupplier = () -> new PersistedStreamingMap<>(
+                        persistedCacheLocation, options,
+                        dynamicProtobufSerializer, runtimeC);
+                corfuTableBuilder.setArguments(mapSupplier, ICorfuVersionPolicy.MONOTONIC);
+            }
+
```
Maithem: I know you are just using the provided API, but we really need to clean this interface. Creating a disk back CorfuTable should be as easy as creating an in-memory table.

Maithem:
Is this still needed?

test/src/test/java/org/corfudb/integration/CorfuStoreIT.java
hisundar: I completely agree. The api is super messy right now.

PR #2323, opened 2020-01-23
Mask TrimmedException in CorfuCompileProxy

vjeko:
Encountering a TrimmedException in CorfuCompileProxy is part of a normal
workflow. The current message causes a lot of confusion during
root-cause analysis, since developers incorrectly assume that Corfu
ended up in an exceptional state.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -193,8 +193,8 @@ public class CorfuCompileProxy<T extends ICorfuSMR<T>> implements ICorfuSMRProxy
                         o -> o.syncObjectUnsafe(timestamp),
                         o -> accessMethod.access(o));
             } catch (TrimmedException te) {
-                log.warn("accessInner: Encountered a trim exception while accessing version {} on attempt {}",
-                        timestamp, x);
+                log.info("accessInner: Encountered trimmed address space " +
+                                "while accessing version {} on attempt {}", timestamp, x);
                 // We encountered a TRIM during sync, reset the object
                 underlyingObject.update(o -> {
                     o.resetUnsafe();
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -296,8 +296,8 @@ private long logUpdateInner(String smrUpdateFunction, final boolean keepUpcallRe
                             + " and we don't have a copy");
                 });
             } catch (TrimmedException ex) {
-                log.warn("getUpcallResultInner: Encountered a trim exception while accessing version {} on attempt {}",
-                        timestamp, x);
+                log.info("accessInner: Encountered trimmed address space " +
```
WenbinZhu: accessInner -> getUpcallResultInner, this seems to be incorrectly copy-pasted.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -172,7 +172,7 @@ public void seek(long globalAddress) {
      *                              writing.
      * @return The (global) address the object was written at.
      */
-    public long append(SMREntry entry,
+    public long a Dixon Ly ppend(SMREntry entry,
```
WenbinZhu: Please fix typo.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
vjeko: Good catch. Fixed.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
vjeko: Fixed.

PR #2324, opened 2020-01-23
Change logSizeQuotaPercentage from int to double

xnull:
## Overview

Description:
Currently Universe framework didn't allow us to use double in logSizeQuotaPercentage. 
This change fixes that

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/it/src/main/java/org/corfudb/universe/node/server/CorfuServerParams.java
+++ b/it/src/main/java/org/corfudb/universe/node/server/CorfuServerParams.java
@@ -79,7 +79,7 @@ public class CorfuServerParams implements NodeParams {
     private final String dockerImage = DOCKER_IMAGE_NAME;
 
     @Default
-    private final int logSizeQuotaPercentage = 100;
+    private final double logSizeQuotaPercentage = 100;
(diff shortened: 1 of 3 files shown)


it/src/main/java/org/corfudb/universe/node/server/CorfuServerParams.java
```diff
@@ -79,7 +79,7 @@
     private final String dockerImage = DOCKER_IMAGE_NAME;
 
     @Default
-    private final int logSizeQuotaPercentage = 100;
+    private final double logSizeQuotaPercentage = 100;
```
hisundar: PR comment says universe framework does not support double but here the parameter is being changed from an int to double right?

xnull: yes
I just updated the pr comment.
We need doubles

xnull: I mean logSizeQuotaPercentage is int but we need double

PR #2325, opened 2020-01-25
VersionLockedObject and CorfuTable Deadlock

Maithem:
## Overview
This patch removes a circular dependency between threads acquiring
the VLO lock and threads used for parallel stream operations (i.e
scanAndFilter).

Why should this be merged: Fixes a deadlock bug

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -14,6 +14,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinWorkerThread;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
@@ -60,6 +62,20 @@ import org.corfudb.runtime.object.ICorfuVersionPolicy;
 @CorfuObject
 public class CorfuTable<K ,V> implements
         ICorfuTable<K, V>, ICorfuSMR<CorfuTable<K, V>> {
+
+    // Accessor/Mutator threads can interleave in a way that create a deadlock because they can create a
+    // circular dependency between the VersionLockedObject(VLO) lock and the common forkjoin thread pool. In order
+    // to break the dependency, parallel stream operations have to execute on a separate pool that applications
+    // cant use. For example, if there are 4 accessor threads all using the common forkjoin pool, one of the threads
+    // can acquire the VLO lock and cause the other 3 threads to wait, but after acquiring the VLO lock, the thread
+    // gets block on parallel stream, because the pool is exhausted with threads that are trying to acquire the VLO
+    // look, which creates a circular dependency. In other words, a deadlock.
+    private final static ForkJoinPool pool = new ForkJoinPool(Runtime.getRuntime().availableProcessors() - 1,
+            pool -> {
+                final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
+                worker.setName("CorfuTable-Forkjoin-pool-" + worker.getPoolIndex());
+                return worker;
+            }, null, true);
 
     // The "main" map which contains the primary key-value mappings.
     private final ContextAwareMap<K,V> mainMap;
@@ -358,9 +374,9 @@ public class CorfuTable<K ,V> implements
      */
     @Accessor
     public List<V> scanAndFilter(Predicate<? super V> valuePredicate) {
-        return mainMap.entryStream()
+        return pool.submit(() -> mainMap.entryStream()
                 .map(Entry::getValue).filter(valuePredicate)
-                .collect(Collectors.toCollection(ArrayList::new));
+                .collect(Collectors.toCollection(ArrayList::new))).join();
     }
 
     /** {@inheritDoc} */
@@ -368,9 +384,9 @@ public class CorfuTable<K ,V> implements
     @Accessor
     public Collection<Map.Entry<K, V>> scanAndFilterByEntry(
             Predicate<? super Map.Entry<K, V>> entryPredicate) {
-        return mainMap.entryStream()
+        return pool.submit(() -> mainMap.entryStream()
                 .filter(entryPredicate)
-                .collect(Collectors.toCollection(ArrayList::new));
+                .collect(Collectors.toCollection(ArrayList::new))).join();
     }
 
     /** {@inheritDoc} */
@@ -446,7 +462,8 @@ public class CorfuTable<K ,V> implements
     @Override
     @Accessor
     public @Nonnull Set<K> keySet() {
-        return mainMap.entryStream().map(Entry::getKey)
+        return mainMap.keySet()
+                .stream()
                 .collect(ImmutableSet.toImmutableSet());
     }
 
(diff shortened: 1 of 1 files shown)


medhavidhawan:
Go out and have fun don’t work so hard. Trust me you will still be fine!!

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -61,6 +63,22 @@
 public class CorfuTable<K ,V> implements
         ICorfuTable<K, V>, ICorfuSMR<CorfuTable<K, V>> {
 
+    // Accessor/Mutator threads can interleave in away that creates a deadlock because the can create a
```
medhavidhawan: r/away/a way/

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -61,6 +63,22 @@
...
+    // circular dependency between the VersionLockedObject(VLO) lock and the common forkjoin thread pool. In order
+    // to break the dependency, parallel stream operations have to execute on a separate pool that applications
+    // cant use. For example, if there are 4 accessor threads all using the common forkjoin pool, one of the threads
+    // can acquire the VLO lock and cause the other 3 threads to wait, but after acquiring the VLO lock, the thread
+    // gets block on parallel stream, because the pool is exhauseted with threads that are trying to acquire the VLO
```
medhavidhawan: r/exhauseted/exhausted

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -61,6 +63,22 @@
...
+    // look, which creates a circular decadency. In other words, a deadlock.
```
medhavidhawan: r/decadency/dependency

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -61,6 +63,22 @@
...
+    private static final ForkJoinPool.ForkJoinWorkerThreadFactory factory = pool -> {
+        final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
+        worker.setName("CorfuTable-Forkjoin-pool-" + worker.getPoolIndex());
+        return worker;
+    };
+
+    private final static ForkJoinPool pool = new ForkJoinPool(Runtime.getRuntime().availableProcessors() - 1,
```
medhavidhawan: Should this be a parameter ? or is this formula always good.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -358,19 +376,19 @@ public void insert(@ConflictParameter K key, V value) {
...
+                .collect(Collectors.toCollection(ArrayList::new))).join();
     }
 
     /** {@inheritDoc} */
     @Override
     @Accessor
     public Collection<Map.Entry<K, V>> scanAndFilterByEntry(
             Predicate<? super Map.Entry<K, V>> entryPredicate) {
-        return mainMap.entryStream()
+        return pool.submit(() -> mainMap.entryStream()
```
medhavidhawan: I know i have seen this problem, but is there a way to create a test that will cause this deadlock. I am not holding the review for this but just want to see if something is possible.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
Maithem: This maintains the current behavior, which uses the common thread pool. The common thread pool uses # of cpus - 1. 

 I don't like it, but making this configurable is part of a bigger change to be able to pass a client context to different layers.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
Maithem: Its a timing issue and there isn't a lot of control over how threads access the VLO. 

Let me think about how this can be tested.

xiaoqin2012: Without a test case it is hard to be convinced. 
A small test may help such as with a thread pool of 4 and each start with try to hold a common lock, then the successful one will try to do parallel work, this could lead to a possible deadlock.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -61,6 +63,22 @@
...
+    // gets block on parallel stream, because the pool is exhauseted with threads that are trying to acquire the VLO
+    // look, which creates a circular decadency. In other words, a deadlock.
+    private static final ForkJoinPool.ForkJoinWorkerThreadFactory factory = pool -> {
+        final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
+        worker.setName("CorfuTable-Forkjoin-pool-" + worker.getPoolIndex());
+        return worker;
+    };
+
+    private final static ForkJoinPool pool = new ForkJoinPool(Runtime.getRuntime().availableProcessors() - 1,
+            factory, null, true);
```
WenbinZhu: Is it cleaner to make the factory an anomymous class instead of a private class field since it's only used here?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -358,19 +376,19 @@ public void insert(@ConflictParameter K key, V value) {
      */
     @Accessor
     public List<V> scanAndFilter(Predicate<? super V> valuePredicate) {
-        return mainMap.entryStream()
+        return pool.submit(() -> mainMap.entryStream()
```
WenbinZhu: scanAndFilter/scanAndFilterByEntry are not the only functions using entryStream(), which uses parallel stream. Other functions like keySet() are using it as well, why only scanAndFilter/scanAndFilterByEntry are changed in the patch?

Maithem:
> Can we completely get rid of parallelstream everywhere? Seems like a simpler approach right?

That would cause a performance regression for scanAndFilter operations. On big tables, some applications need parallelism.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
Maithem: I didn't notice.

I'm going to change them to use keySet and values instead, because those two methods shouldn't be exposed by `PersistedStreamingMap`

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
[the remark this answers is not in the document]
Maithem: done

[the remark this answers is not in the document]
Maithem: done

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
Maithem: This race condition is really tricky to reproduce because of timing. I wasn't able to do it with latches. 

Apparent, the usage of parallel stream with the common fork-join pool has unspecified behavior
https://stackoverflow.com/a/29272776

Issue #2328, opened 2020-01-27
Corfu IT tests are broken on Mac OS X 10.15+


hisundar:
Starting with Mac OS X version 10.15 the temporary directory returned by `System.getProperty("java.io.tmpdir")` has read-only permissions on the parent filesystem.
LogUnitServer in corfu however fails to come up if the filesystem is read-only.
This prevents CorfuServer from being tested on Mac OS X 10.15+ and it fails with exception like this
`Cannot start Corfu on a read-only filesystem: /var/folders/qx/045nmsxx55575_gqycvvs1480000gp/T/1580163123321-0`
Note that while the parent filesystem indicates that it is read-only, the actual corfu log directories are still writable, so testing the parent filesystem for writability might be a false panic.

Maithem:
Fix merged.

PR #2329, opened 2020-01-27
Remove Parent Filesystem read-only check for LogUnitServer

hisundar:
## Overview

Description:

Why should this be merged: 

Starting with Mac OS X 10.15 the parent filesystem returned
by the System.getProperty("java.io.tmpdir") is marked as
read-only. However this does not mean that the returned
temporary directory itself is read-only.
So given that this can happen in production as well,
removing the redundant parent filesystem read-only
check would prevent false panics and allow IT tests
to run on Mac as well.

Related issue(s) (if applicable): #2328 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -168,10 +168,6 @@ private long initStreamLogDirectory() {
             String corfuDir = logDir.getParent().toString();
             FileStore corfuDirBackend = Files.getFileStore(Paths.get(corfuDir));
 
-            if (corfuDirBackend.isReadOnly()) {
```
xiaoqin2012: We skip this check if it is MAC os by checking OS property.

hisundar: We could but it would be better to avoid adding platform specific code as far as possible.

hisundar:
> The reason this was added was because there were systems that were mounting in read-only mode, but it was still part of the chain. This blocks the write path and the fault detector wouldn't change it.

It shouldn't because we still test the folders we care about for write permissions `/config/corfu` and `/config/corfu/log` for example. Note that the test case to validate these still passes with this change.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xiaoqin2012: If we don't do this check on Linux, and the file system is really in read-only mode. The following call new File(), it will throw an IO.readonly.exception. it is better to handle it.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -168,10 +168,6 @@ private long initStreamLogDirectory() {
...
-                throw new LogUnitException("Cannot start Corfu on a read-only filesystem:" + corfuDir);
-            }
-
             File corfuDirFile = new File(corfuDir);
             if (!corfuDirFile.canWrite()) {
```
WenbinZhu: Just saw this comment in JDK source code:
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/io/File.java#L780

Looks like the parent filesystem check may still be necessary  in some cases in addition to the directory check here, though I understand it's still not 100% guarantee.
Can we try to increase JVM privilege to overcome this issue? Or use another directory check API with stronger guarantee?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
hisundar: if the new File() throws an IO.readonly.exception that is also acceptable. Either case the objective of preventing a LogUnitServer from coming up would be accomplished, right?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
hisundar: From the Java doc..
On some platforms it may be possible to start the Java virtual machine with special privileges that allow it to modify files that are marked read-only.
We are only interested in capability to write to a file. If the permissions allow modification of the file, that is going to be good enough for LogUnitServer.
Besides this is a small best effort check. It would be quite difficult to cover all the possibilities of file based permissions mismatch that can occur.

PR #2330, opened 2020-01-29
Layout Bootstrap and ClusterId Validation for the Server Message Routing 

PavelZaytsev:
## Overview

Description:
This is a bug fix to prevent writes from happening to the non bootstrapped corfu server. 

Why should this be merged: 
We had this bug happening a couple of times in production for the following scenario:
1) The node belongs to cluster A.
2) The node is getting reset to join to cluster B.
3) Some client, who used a cluster A in the past, writes a record stamped with epoch 0 to the log unit.
4) The node is getting reset to join a cluster B.
5) Although the routers of this node are reset, the packet comes out of limbo (either because it got stuck in the netty caches or just delayed during routing), gets accepted to the node's log unit *before* the node is bootstrapped with a new layout of cluster B.
6) Since the default sealed epoch is still the same (0), the record goes through the batch processor and gets written finally in the stream log.
7) As you may have guessed, this creates data inconsistency that usually manifests itself in the broken state transfer or corrupted data.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
@@ -80,11 +80,17 @@ public class CorfuServerNode implements AutoCloseable {
      * @param serverMap     Server Map with all components.
      */
     public CorfuServerNode(@Nonnull ServerContext serverContext,
-                           @Nonnull Map<Class, AbstractServer> serverMap) {
+                           @Nonnull ImmutableMap<Class, AbstractServer> serverMap) {
         this.serverContext = serverContext;
         this.serverMap = serverMap;
-        router = new NettyServerRouter(new ArrayList<>(serverMap.values()));
+        router = new NettyServerRouter(serverMap.values().asList(), serverContext);
         this.serverContext.setServerRouter(router);
+        // If the node is started in the single node setup and was bootstrapped,
+        // set the server epoch as well.
+        if(serverContext.isSingleNodeSetup() && serverContext.getCurrentLayout() != null){
+            serverContext.setServerEpoch(serverContext.getCurrentLayout().getEpoch(),
+                    router);
+        }
         this.close = new AtomicBoolean(false);
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
@@ -2,13 +2,26 @@ package org.corfudb.infrastructure;
 
 import io.netty.channel.ChannelHandlerContext;
 import org.corfudb.protocols.wireprotocol.CorfuMsg;
+import org.corfudb.protocols.wireprotocol.CorfuMsgType;
+import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
+import org.corfudb.protocols.wireprotocol.WrongClusterMsg;
+import org.corfudb.runtime.view.Layout;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
 
 /**
  * Created by mwei on 12/13/15.
  */
 public interface IServerRouter {
+
+    // Lombok annotations are not allowed for the interfaces.
+    Logger log = LoggerFactory.getLogger(IServerRouter.class);
+
     void sendResponse(ChannelHandlerContext ctx, CorfuMsg inMsg, CorfuMsg outMsg);
 
     /**
@@ -22,8 +35,13 @@ public interface IServerRouter {
     void setServerEpoch(long newEpoch);
 
     /**
+     * Get the currently bootstrapped layout.
+     */
+    Optional<Layout> getCurrentLayout();
+
+    /**
      * Register a server to route messages to.
-     * @param server    The server to route messages to
+     * @param server The server to route messages to
      */
     void addServer(AbstractServer server);
 
@@ -31,4 +49,113 @@ public interface IServerRouter {
      * Get a list of registered servers.
      */
     List<AbstractServer> getServers();
+
+    /**
+     * Set a serverContext for this router.
+     * @param serverContext A current server context.
+     */
+    void setServerContext(ServerContext serverContext);
+
+    /**
+     * Send WRONG_EPOCH message.
+     *
+     * @param msg The incoming message.
+     * @param ctx The context of the channel handler.
+     */
+    default void sendWrongEpochMessage(CorfuMsg msg, ChannelHandlerContext ctx) {
+        sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH,
+                getServerEpoch()));
+        log.trace("Incoming message with wrong epoch, got {}, expected {}, message was: {}",
+                msg.getEpoch(), getServerEpoch(), msg);
+    }
+
+    /**
+     * Send LAYOUT_NOBOOTSTRAP message.
+     *
+     * @param msg The incoming message.
+     * @param ctx The context of the channel handler.
+     */
+    default void sendNoBootstrapMessage(CorfuMsg msg, ChannelHandlerContext ctx) {
+        sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.LAYOUT_NOBOOTSTRAP));
+        log.trace("Received message but not bootstrapped! Message={}", msg);
+    }
+
+    /**
+     * Send WRONG_CLUSTER_ID message.
+     *
+     * @param msg              The incoming message.
+     * @param ctx              The context of the channel handler.
+     * @param currentClusterID The current cluster id.
+     */
+    default void sendWrongClusterIdMessage(CorfuMsg msg, ChannelHandlerContext ctx, UUID currentClusterID) {
+        sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_CLUSTER_ID,
+                new WrongClusterMsg(currentClusterID, msg.getClusterID())));
+        log.trace("Incoming message with a wrong cluster ID, got {}, expected {}, message was {}",
+                msg.getClusterID(), currentClusterID, msg);
+    }
+
+    /**
+     * Validate the epoch of a CorfuMsg, and send a WRONG_EPOCH response if
+     * the server is in the wrong epoch. Ignored if the message type is reset (which
+     * is valid in any epoch).
+     *
+     * @param msg The incoming message to validate.
+     * @param ctx The context of the channel handler.
+     * @return True, if the epoch is correct, but false otherwise.
+     */
+    default boolean epochIsValid(CorfuMsg msg, ChannelHandlerContext ctx) {
+        long serverEpoch = getServerEpoch();
+        if (msg.getEpoch() != serverEpoch) {
+            sendWrongEpochMessage(msg, ctx);
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * Validate that the message's cluster ID is equal to the cluster ID of a bootstrapped layout.
+     *
+     * @param msg           The incoming message.
+     * @param ctx           The context of the channel handler.
+     * @param currentLayout The layout a server was bootstrapped with.
+     * @return True, if the message's cluster ID is equal to this node's cluster ID, otherwise false.
+     */
+    default boolean clusterIdIsValid(CorfuMsg msg, ChannelHandlerContext ctx, Layout currentLayout) {
+        UUID currentClusterID = currentLayout.getClusterId();
+
+        boolean clusterIdsMatch = msg.getClusterID()
+                .equals(currentClusterID);
+
+        if (!clusterIdsMatch) {
+            sendWrongClusterIdMessage(msg, ctx, currentClusterID);
+        }
+        return clusterIdsMatch;
+    }
+
+    /**
+     * Validate the incoming message. The message is valid if:
+     * 1) The flag ignoreEpoch is set to true or it's set to false, and the epoch is valid for all the messages.
+     * 2) Also if the flag ignoreClusterId is set to false,
+     *      a. The current layout server should be bootstrapped and
+     *      b. the message's cluster ID should be equal to the bootstrapped layout's cluster ID.
+     *
+     * @param msg The incoming message.
+     * @param ctx The context of the channel handler.
+     * @return True, if it's a valid message, and false otherwise.
+     */
+    default boolean messageIsValid(CorfuMsg msg, ChannelHandlerContext ctx) {
+        if (!msg.getMsgType().ignoreEpoch && !epochIsValid(msg, ctx)) {
+            return false;
+        }
+
+        if(!msg.getMsgType().ignoreClusterId){
+            return getCurrentLayout()
+                    .map(layout -> clusterIdIsValid(msg, ctx, layout))
+                    .orElseGet(() -> {
+                        sendNoBootstrapMessage(msg, ctx);
+                        return false;
+                    });
+        }
+        return true;
+    }
 }
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
@@ -1,22 +1,21 @@
 package org.corfudb.infrastructure;
 
 import com.google.common.collect.ImmutableList;
-
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.wireprotocol.CorfuMsg;
+import org.corfudb.protocols.wireprotocol.CorfuMsgType;
+import org.corfudb.runtime.view.Layout;
 
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
-
-import lombok.Getter;
-import lombok.Setter;
-import lombok.extern.slf4j.Slf4j;
-
-import org.corfudb.protocols.wireprotocol.CorfuMsg;
-import org.corfudb.protocols.wireprotocol.CorfuMsgType;
-import org.corfudb.protocols.wireprotocol.CorfuPayloadMsg;
+import java.util.Optional;
+import java.util.Set;
 
 
 /**
@@ -34,6 +33,10 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter implements I
     private final Map<CorfuMsgType, AbstractServer> handlerMap;
 
     /**
+     * This node's server context.
+     */
+    private final ServerContext serverContext;
+    /**
      * The epoch of this router. This is managed by the base server implementation.
      */
     @Getter
@@ -41,20 +44,23 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter implements I
     volatile long serverEpoch;
 
     /** The {@link AbstractServer}s this {@link NettyServerRouter} routes messages for. */
-    final List<AbstractServer> servers;
+    private final ImmutableList<AbstractServer> servers;
 
     /** Construct a new {@link NettyServerRouter}.
      *
      * @param servers   A list of {@link AbstractServer}s this router will route
      *                  messages for.
      */
-    public NettyServerRouter(List<AbstractServer> servers) {
-        // Initialize the router epoch from the persisted server epoch
-        this.serverEpoch = ((BaseServer) servers.get(0)).serverContext.getServerEpoch();
-        this.servers = ImmutableList.copyOf(servers);
+    public NettyServerRouter(ImmutableList<AbstractServer> servers, ServerContext serverContext) {
+        this.serverContext = serverContext;
+        this.serverEpoch = serverContext.getServerEpoch();
+        this.servers = servers;
         handlerMap = new EnumMap<>(CorfuMsgType.class);
-        servers.forEach(server -> server.getHandler().getHandledTypes()
-            .forEach(x -> handlerMap.put(x, server)));
+
+        servers.forEach(server -> {
+            Set<CorfuMsgType> handledTypes = server.getHandler().getHandledTypes();
+            handledTypes.forEach(handledType -> handlerMap.put(handledType, server));
+        });
     }
 
     /**
@@ -77,6 +83,11 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter implements I
         return servers;
     }
 
+    @Override
+    public void setServerContext(ServerContext serverContext) {
+        throw new UnsupportedOperationException("The operation is not supported.");
+    }
+
     /**
      * Send a netty message through this router, setting the fields in the outgoing message.
      *
@@ -90,25 +101,9 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter implements I
         log.trace("Sent response: {}", outMsg);
     }
 
-    /**
-     * Validate the epoch of a CorfuMsg, and send a WRONG_EPOCH response if
-     * the server is in the wrong epoch. Ignored if the message type is reset (which
-     * is valid in any epoch).
-     *
-     * @param msg The incoming message to validate.
-     * @param ctx The context of the channel handler.
-     * @return True, if the epoch is correct, but false otherwise.
-     */
-    public boolean validateEpoch(CorfuMsg msg, ChannelHandlerContext ctx) {
-        long serverEpoch = getServerEpoch();
-        if (!msg.getMsgType().ignoreEpoch && msg.getEpoch() != serverEpoch) {
-            sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH,
-                    serverEpoch));
-            log.trace("Incoming message with wrong epoch, got {}, expected {}, message was: {}",
-                    msg.getEpoch(), serverEpoch, msg);
-            return false;
-        }
-        return true;
+    @Override
+    public Optional<Layout> getCurrentLayout() {
+        return Optional.ofNullable(serverContext.getCurrentLayout());
     }
 
     /**
@@ -129,7 +124,7 @@ public class NettyServerRouter extends ChannelInboundHandlerAdapter implements I
                 // The message was unregistered, we are dropping it.
                 log.warn("Received unregistered message {}, dropping", m);
             } else {
-                if (validateEpoch(m, ctx)) {
+                if (messageIsValid(m, ctx)) {
                     // Route the message to the handler.
                     if (log.isTraceEnabled()) {
                         log.trace("Message routed to {}: {}", handler.getClass().getSimpleName(), msg);
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
@@ -28,6 +28,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
@@ -93,7 +94,7 @@ public class FailureDetector implements IDetector {
 
         // Perform polling of all responsive servers.
         return pollRound(
-                layout.getEpoch(), allServers, routerMap, sequencerMetrics,
+                layout.getEpoch(), layout.getClusterId(), allServers, routerMap, sequencerMetrics,
                 ImmutableList.copyOf(layout.getUnresponsiveServers())
         );
     }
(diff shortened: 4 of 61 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
```diff
@@ -90,6 +89,7 @@ public CorfuServerNode(@Nonnull ServerContext serverContext,
         this.serverMap = serverMap;
         router = new NettyServerRouter(new ArrayList<>(serverMap.values()));
         this.serverContext.setServerRouter(router);
+        this.serverContext.setServerEpochForSingleNodeLayout(serverContext.getServerEpoch(), router);
```
Maithem: I dont think this is needed.

The root cause of the problem is: the epoch's uninitialized value is a valid value. Therefore, we need to change `ServerContext:getServerEpoch` to return an invalid value if epoch == null. Instead of returning 0 for null, we should return `Layout.INVALID_EPOCH`

PavelZaytsev: I tried doing just that but a lot of tests failed, since it's necessary for some servers to have the context server epoch to start at 0.

PavelZaytsev: It possible to pass -1 to the BatchProcessor on start up, which will probably work, but then the client would see a WrongEpochException(-1), which is misleading, since they write to the log unit with the epoch 0. So I thought the better approach would be to start with None Epoch and then just seal it after the bootstrap. If the node was never bootstrapped, it was never sealed, so all the writes are rejected.

Maithem: I still think the uninitialized epoch value shouldn't have a valid epoch number.

But even if that is done, I don't think its a generic way to deal with this issue. If you look at `LayoutServer` , you'll see that each handler (except bootstrap) will try to check `isBootstrapped`. Essentially, its the same problem: If a server is not bootstrapped it shouldn't service RPCs. I think the cleanest way to do this is at the message routing level. The `NettyServerRouter` should reject all messages as long as the server hasn't been bootstrapped.

Also, I don;t think its necessary to add a new exception type. `NoBootstrapException` is exactly what you need.

WenbinZhu: We might also need to check clusterId on BaseServer.

PavelZaytsev: @Maithem sounds good. 
@WenbinZhu that sounds good, it will entail adding clusterId as one of the fields of CorfuMsg, some changes with respect to serialization/deserialization and checking of the clusterId from the router. Hopefully, not a lot of work

PavelZaytsev:
@Maithem @WenbinZhu I have addressed the comments, please review when you get the chance

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -135,16 +129,16 @@ private void processor() {
                     currOp.getFutureResult().completeExceptionally(
                             new QuotaExceededException("Quota of "
                                     + streamLog.quotaLimitInBytes() + " bytes"));
-                    log.warn("batchprocessor: quota exceeded, dropping msg {}", currOp.getMsg());
-                } else if (currOp.getType() == Type.SEAL && currOp.getMsg().getEpoch() >= sealEpoch) {
+                    log.warn("batchWriteProcessor: quota exceeded, dropping msg {}", currOp.getMsg());
```
Maithem: It seems like these changes were based on an older master, undo this file's changes?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
```diff
@@ -80,10 +80,10 @@ public CorfuServerNode(@Nonnull ServerContext serverContext) {
      * @param serverMap     Server Map with all components.
      */
     public CorfuServerNode(@Nonnull ServerContext serverContext,
-                           @Nonnull Map<Class, AbstractServer> serverMap) {
+                           @Nonnull ImmutableMap<Class, AbstractServer> serverMap) {
         this.serverContext = serverContext;
         this.serverMap = serverMap;
-        router = new NettyServerRouter(new ArrayList<>(serverMap.values()));
+        router = new NettyServerRouter(serverMap, serverContext.getServerEpoch());
```
Maithem: `serverContext.getServerEpoch()` can still return 0 if it wasn't initialized. Why not just use an invalid value like -1 ?

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
```diff
@@ -90,25 +91,13 @@ public void sendResponse(ChannelHandlerContext ctx, CorfuMsg inMsg, CorfuMsg out
...
-            log.trace("Incoming message with wrong epoch, got {}, expected {}, message was: {}",
-                    msg.getEpoch(), serverEpoch, msg);
-            return false;
+    @Override
+    public Optional<Layout> getCurrentLayout() {
+        if(!serverMap.containsKey(LayoutServer.class)){
+            throw new IllegalStateException("Layout server is not in a server map.");
         }
-        return true;
+        LayoutServer layoutServer = (LayoutServer) serverMap.get(LayoutServer.class);
```
Maithem: The router shouldn't depend on the `LayoutServer` , its cleaner to just use `serverContext`.

runtime/src/main/java/org/corfudb/runtime/view/RuntimeLayout.java
```diff
@@ -109,8 +110,9 @@ private IClient getClient(final Class<? extends IClient> clientClass,
             endpointClientMap.computeIfAbsent(endpoint, s -> {
                 try {
                     Constructor<? extends IClient> ctor =
-                            clientClass.getDeclaredConstructor(IClientRouter.class, long.class);
-                    IClient inst = ctor.newInstance(getRuntime().getRouter(endpoint), layout.getEpoch());
+                            clientClass.getDeclaredConstructor(IClientRouter.class, long.class, UUID.class);
```
Maithem: Just pass the whole layout instead of passing the epoch and cluster id separately ?

test/src/test/java/org/corfudb/infrastructure/AbstractServerTest.java
```diff
@@ -15,6 +15,8 @@
 import org.corfudb.runtime.clients.TestClientRouter;
 import org.junit.Before;
 
+import java.util.AbstractMap;
```
Maithem: Undo ?

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
```diff
@@ -39,6 +39,13 @@ public AbstractServer getDefaultServer() {
                 .build());
     }
 
+    @Override
+    public void setServer(AbstractServer server) {
+        router.reset();
+        router.setIgnoreMessageValidation(true);
```
Maithem: Why ignore the epoch/cluster id check?

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -93,7 +94,7 @@ public PollReport poll(
 
         // Perform polling of all responsive servers.
         return pollRound(
-                layout.getEpoch(), allServers, routerMap, sequencerMetrics,
+                layout.getEpoch(), layout.getClusterId(), allServers, routerMap, sequencerMetrics,
```
Maithem: Just pass the whole Layout?

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
```diff
@@ -21,14 +32,123 @@
...
+        if (!epochIsValid(msg, ctx)) {
+            return false;
+        }
+
+        return getCurrentLayout()
+                .map(layout -> clusterIdIsValid(msg, ctx, layout))
+                .orElseGet(() -> {
+                    sendNoBootstrapMessage(msg, ctx);
+                    return false;
+                });
```
Maithem: This seems incorrect. 

Imagine a fault detector from one cluster trying to seal a node from another cluster. Since a seal message is marked as ignore epoch, the second check won't be executed and thus allowing fault detectors from different cluster to reconfigure each other's nodes.

test/src/test/java/org/corfudb/infrastructure/AbstractServerTest.java
PavelZaytsev: Done

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
PavelZaytsev: That was the first choice that I tried but I have encountered a couple of issues. 
This will create a circular dependency between a context and a router, and force one more component to depend on it, making it even more of a god object. 
This will also require more changes to be done to the TestServerRouter and the TestServer, which are messy to implement. 
Since the LayoutServer is the owner of the layout, and the router has access to the servers, I think that it's viable to retrieve a layout from a server.

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
PavelZaytsev: Because for some tests we just test a specific server, and not the stack. It's easier to ignore the validation in the router then to bootstrap a node with a specific layout, since epoch or a presence of a layout or a cluster id is irrelevant in the test.

PavelZaytsev:
```
What happens in the following scenario:

A client's runtime connects to two clusters at the same time and the sequencer is on the tail node.
Then it would keep acquiring a token from the sequencer but fail to write to it because of WrongClusterException, this will cause so many holes that can impact other clients.

Maybe you can add a check so that the runtime cannot create different clients for different clusters. Right now the validation happens as a post check, but it could even be done earlier so that the client cannot even discover layouts with different cluster ids.
```

@Maithem 
I am a bit confused as to how to recreate this scenario. Right now, there is an assumption that a layout corresponds to a unique clusterId. We retrieve a layout from the layout server of the nodes via fetchLayout() either when the node is bootstrapped for the first time or when the node invalidates a layout and fetches one from the previous layout server list. This would mean that we would have to swap the layout on one of the nodes, and assign it a different clusterId. 
Lets say this adversity happens. So we connect in the loop to the first node with a clusterId1, all the clients (sequencer, logunit, manager, etc) are connected to the node from this cluster too, the clusterId1 in the runtime will be assigned to a runtime's clusterId. In the next iteration of the loop we, will try to connect to a node with a clusterId2, the current check will throw an exception, and the runtime will not be able to connect to all the nodes in the cluster, so the connect will fail. What am I missing to recreate the scenario that you have described?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
PavelZaytsev: Since we don't bootstrap a single node layout from a client, but rather do it on startup from the `installSingleNodeLayoutIfAbsent` method, we set the node's layout, but we never seal it. So we would need to seal both the router and epoch with the singleNodeLayout's epoch before a single node can start.

runtime/src/main/java/org/corfudb/runtime/view/RuntimeLayout.java
PavelZaytsev: I feel like for router it is a sufficient enough information

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
PavelZaytsev: I feel like it is sufficient enough for the poll round to perform iterations. I would extract servers, epoch and clusterId in a separate DTO, for example, PollData but it's not enough the effort, since it's used only in two functions.

Maithem:
> Right now, there is an assumption that a layout corresponds to a unique clusterId. We retrieve a layout



> ```
> What happens in the following scenario:
> 
> A client's runtime connects to two clusters at the same time and the sequencer is on the tail node.
> Then it would keep acquiring a token from the sequencer but fail to write to it because of WrongClusterException, this will cause so many holes that can impact other clients.
> 
> Maybe you can add a check so that the runtime cannot create different clients for different clusters. Right now the validation happens as a post check, but it could even be done earlier so that the client cannot even discover layouts with different cluster ids.
> ```
> 
> @Maithem
> I am a bit confused as to how to recreate this scenario. Right now, there is an assumption that a layout corresponds to a unique clusterId. We retrieve a layout from the layout server of the nodes via fetchLayout() either when the node is bootstrapped for the first time or when the node invalidates a layout and fetches one from the previous layout server list. This would mean that we would have to swap the layout on one of the nodes, and assign it a different clusterId.
> Lets say this adversity happens. So we connect in the loop to the first node with a clusterId1, all the clients (sequencer, logunit, manager, etc) are connected to the node from this cluster too, the clusterId1 in the runtime will be assigned to a runtime's clusterId. In the next iteration of the loop we, will try to connect to a node with a clusterId2, the current check will throw an exception, and the runtime will not be able to connect to all the nodes in the cluster, so the connect will fail. What am I missing to recreate the scenario that you have described?

I was thinking about the possibility of two RuntimeLayouts created with different cluster IDs for two endpoints, but it seems like its not possible because the layout is updated serially and checkClusterId checks if cluster id changed.

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
PavelZaytsev: Resolving for now, as this situation is impossible since a fault detector's management runtime will not start trying to connect to another cluster, failing the clusterId check.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
PavelZaytsev: Fixed

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
PavelZaytsev: Edit: Although there is a circular dependency, this approach seems to be more consistent with the testing we are doing for the single servers and removes the incorrect approach of ignoring the routing validation.

test/src/test/java/org/corfudb/infrastructure/LogUnitCacheTest.java
PavelZaytsev: Edit: Was removed because was inconsistent in a couple of places. Now we explicitly bootstrap in the test cases, set cluster Ids and epochs on client and server routers.

PavelZaytsev:
@Maithem @WenbinZhu I have addressed the latest comments, please review when you get the chance.

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
```diff
@@ -21,14 +33,130 @@
...
+    /**
+     * Validate that the message's cluster ID is equal to the cluster ID of a bootstrapped layout.
+     *
+     * @param msg           The incoming message.
+     * @param ctx           The context of the channel handler.
+     * @param currentLayout The layout a server was bootstrapped with.
+     * @return True, if the message's cluster ID is equal to this node's cluster ID, otherwise false.
+     */
+    default boolean clusterIdIsValid(CorfuMsg msg, ChannelHandlerContext ctx, Layout currentLayout) {
+        UUID currentClusterID = currentLayout.getClusterId();
```
xnull: please rename currentClusterId
note: never use consecutive capital letters

infrastructure/src/main/java/org/corfudb/infrastructure/management/FailureDetector.java
```diff
@@ -126,7 +127,7 @@ PollReport pollRound(
         List<PollReport> reports = new ArrayList<>();
         for (int iteration = 0; iteration < failureThreshold; iteration++) {
             PollReport currReport = pollIteration(
-                    allServers, router, epoch, sequencerMetrics, layoutUnresponsiveNodes
+                    allServers, router, epoch, clusterID, sequencerMetrics, layoutUnresponsiveNodes
```
xnull: *clusterId

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
Maithem: Which clusterId check are referring to ? `messageIsValid` isn't sufficient for RPCs that ignore epoch checks, because ignore epoch is within a context of a cluster, so returning true only based on that flag is incorrect.

I'm thinking of the following scenario: The management agent has a layout and connected to 3 nodes (A, B and C). Since reset is an ad-hoc call it can happen at any time. Now, some one resets node C, the netty channel will auto reconnect by-passing all checks. Now the failure detector sees that C node doesn't have the latest layout and tries to patch it with LAYOUT_COMMITTED (which will pass messageIsValid because it sets ignoreEpoch=true). 

This is incorrect because C got re-added to the cluster after being reset. First of all why cant this scenario happen ? and at what point would message validation fail for the next step (i.e. LAYOUT_COMMITTED). 

Obviously, this isn't an issue for msgs that don't ignore epochs and its not specific to LAYOUT_COMMITTED, but all messages that choose to ignore epoch checks.

PavelZaytsev: I see, I did not know that the netty channel reconnects bypassing the clusterId checks. This way, FD can interfere with the workflows that have resets. In that case, the correct behavior would be to ignore the epochIsValid check for the messages that have the ignoreEpoch=True but continue checking the clusterId nonetheless.

PavelZaytsev: @Maithem 
I fixed the issue related to checking the layout and clusterId for the ignoreEpoch messages, however it introduced another issue that these checks should actually be ignored for some of the message types. This makes sense because these message types have to deal with pings, cluster discovery, cluster bootstrapping or ADD_NODE workflow related (A cluster should not/might not be bootstrapped, when the server router receives these messages). I also had to fix some of the tests accordingly. Let me know if you got any questions, and please review when you get the chance.

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
```diff
@@ -41,20 +45,41 @@
     volatile long serverEpoch;
 
     /** The {@link AbstractServer}s this {@link NettyServerRouter} routes messages for. */
-    final List<AbstractServer> servers;
+    private final ImmutableList<AbstractServer> servers;
+
+    /**
+     * The set of messages for which the layout and cluster id validations should be ignored.
+     */
+    private final Set<CorfuMsgType> ignoreLayoutMsgTypes = ImmutableSet.of(
```
Maithem: Can you put this in CorfuMsgType. IgnoreClusterId should be specified in similar way to ignoreEpoch

annym:
This PR is breaking the IF, I'll filter it out for now, so other PRs run. Let me know when you want to test it or you can test I separately.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -81,6 +81,7 @@
     /**
      * The server context of the node.
      */
+    @Getter
```
WenbinZhu: Is this necessary? Looks like no one use it other than itself.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
```diff
@@ -80,11 +80,17 @@ public CorfuServerNode(@Nonnull ServerContext serverContext) {
...
-                           @Nonnull Map<Class, AbstractServer> serverMap) {
+                           @Nonnull ImmutableMap<Class, AbstractServer> serverMap) {
         this.serverContext = serverContext;
         this.serverMap = serverMap;
-        router = new NettyServerRouter(new ArrayList<>(serverMap.values()));
+        router = new NettyServerRouter(serverMap.values().asList(), serverContext);
         this.serverContext.setServerRouter(router);
+        // If the node is started in the single node setup and was bootstrapped,
+        // set the server epoch as well.
+        if(serverContext.isSingleNodeSetup() && serverContext.getCurrentLayout() != null){
```
WenbinZhu: Can you explain the necessity of this block?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsg.java
```diff
@@ -119,6 +126,15 @@ public void serialize(ByteBuf buffer) {
...
         }
+        if(clusterID == null){
+            buffer.writeLong(0L);
+            buffer.writeLong(0L);
+        }
+        else{
+            buffer.writeLong(clusterID.getMostSignificantBits());
+            buffer.writeLong(clusterID.getLeastSignificantBits());
+        }
+
```
WenbinZhu: NIT: add spaces after if and else

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
PavelZaytsev: Fixed

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
```diff
@@ -82,41 +84,42 @@
...
 
     // EXTRA CODES
-    LAYOUT_ALREADY_BOOTSTRAP(60, TypeToken.of(CorfuMsg.class), true),
-    LAYOUT_PREPARE_ACK(61, new TypeToken<CorfuPayloadMsg<LayoutPrepareResponse>>(){}, true),
-    RESTART(62, TypeToken.of(CorfuMsg.class), true),
-    KEEP_ALIVE(63, TypeToken.of(CorfuMsg.class), true),
+    LAYOUT_ALREADY_BOOTSTRAP(60, TypeToken.of(CorfuMsg.class), true, false),
+    LAYOUT_PREPARE_ACK(61, new TypeToken<CorfuPayloadMsg<LayoutPrepareResponse>>(){}, true, false),
+    RESTART(62, TypeToken.of(CorfuMsg.class), true, false),
+    KEEP_ALIVE(63, TypeToken.of(CorfuMsg.class), true, true),
```
WenbinZhu: I think for KEEP_ALIVE, it's better not to ignore cluster ID, if a client is from a different cluster, it's unnecessary for the servers in this cluster to keep alive the channel to that client.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WrongClusterMsg.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+
+import java.util.UUID;
+
+@Builder
+@AllArgsConstructor
+@Getter
+public class WrongClusterMsg implements ICorfuPayload<WrongClusterMsg> {
```
WenbinZhu: Add javadoc

runtime/src/main/java/org/corfudb/runtime/RebootUtil.java
```diff
@@ -51,13 +61,16 @@ public static void reset(@NonNull String endpoint,
...
      * @param retryDuration          Duration between retries.
+     * @param clusterId              Optional cluster Id. If None is provided, then the default one
+     *                               will be used to create a base client.
      */
     public static void reset(@NonNull String endpoint,
                              @NonNull CorfuRuntimeParameters corfuRuntimeParameters,
                              int retries,
-                             @NonNull Duration retryDuration) {
+                             @NonNull Duration retryDuration,
+                             Optional<UUID> clusterId) {
```
WenbinZhu: The APIs in this file are public APIs consumed by our verticals, you probably need to modify the veritcal codes when you check-in to production code. Is this clusterId field really needed here, can we just use the default?

runtime/src/main/java/org/corfudb/runtime/RebootUtil.java
```diff
@@ -25,6 +27,11 @@ private RebootUtil() {
         // prevent instantiation of this class
     }
 
+    /**
+     * A default cluster Id for resets if none are provided by a user.
+     */
+    private static final UUID DEFAULT_CLUSTER_ID =
```
WenbinZhu: Can we reuse Layout.INVALID_CLUSTER_ID?

test/src/test/java/org/corfudb/infrastructure/TestLayoutBuilder.java
```diff
@@ -34,6 +34,7 @@ public TestLayoutBuilder() {
         layoutServers = new ArrayList<>();
         unresponsiveServers = new ArrayList<>();
         segments = new ArrayList<>();
+        clusterId = UUID.fromString("00000000-0000-0000-0000-000000000000");
```
WenbinZhu: You can use Layout.INVALID_CLUSTER_ID

test/src/test/java/org/corfudb/infrastructure/TestLayoutBuilder.java
```diff
@@ -49,7 +50,7 @@ public static Layout single(int port) {
                 .addLogUnit(port)
                 .addToSegment()
                 .addToLayout()
-                .setClusterId(UUID.randomUUID())
+                .setClusterId(UUID.fromString("00000000-0000-0000-0000-000000000000"))
```
WenbinZhu: You can use Layout.INVALID_CLUSTER_ID, please replace occurences in other test files.

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
```diff
@@ -21,14 +34,128 @@
...
+    /**
+     * Send WRONG_EPOCH message.
+     *
+     * @param msg The incoming message.
+     * @param ctx The context of the channel handler.
+     */
+    default void sendWrongEpochMessage(CorfuMsg msg, ChannelHandlerContext ctx) {
+        sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH,
+                getServerEpoch()));
+        log.warn("Incoming message with wrong epoch, got {}, expected {}, message was: {}",
```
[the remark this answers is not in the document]
PavelZaytsev: I guess I can change it back to trace, since it happens fairly often.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
PavelZaytsev: Removed

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
PavelZaytsev: ServerContext is now initialized with the invalid epoch. In the normal node to set it to the first epoch (0), we bootstrap it with a layout and seal the epoch from the client. After this happens, the client can start routing messages to the node safely. In the single node, we don't bootstrap it since the node self-bootstraps on a startup from the layout server. This block ensures that we self-seal the node in the single node too after everything has been initialized. Otherwise, in the single node setup the log unit seal epoch would be invalid, and all the messages won't be able to go through.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsg.java
PavelZaytsev: Fixed

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
PavelZaytsev: The reason for keep alive ignoring the clusterId checks is because there is a possibility of sending a keep alive messages after the handshake is completed but before the cluster was formed. Basically this scenario might occur:

1. Handshake occurs to establish a trust (making sure a client and a server have correct certificates). We are not concerned with cluster level here yet because there is no cluster per se. 
2. Channel connection is established after a successful handshake. 
3. Channel is idle, and a client has not bootstrapped a cluster yet. A keep alive message is sent to the server router. If we did not ignore, this would have thrown an exception because of the clusterId check but there was never a cluster in the first place, we just established the connection.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WrongClusterMsg.java
PavelZaytsev: Done

runtime/src/main/java/org/corfudb/runtime/RebootUtil.java
PavelZaytsev: It's a fair point. Unfortunately, a reset message is the one we should check the layout and id for the reset messages. I checked whether this API is used in the newest release, and it seems like it's not (they just use our internal client code instead). For the previous releases this API is used in a few places in the cluster management code.

runtime/src/main/java/org/corfudb/runtime/RebootUtil.java
PavelZaytsev: Fixed

test/src/test/java/org/corfudb/infrastructure/TestLayoutBuilder.java
PavelZaytsev: Fixed

test/src/test/java/org/corfudb/infrastructure/TestLayoutBuilder.java
PavelZaytsev: Fixed

PR #2335, opened 2020-01-30
Change Codec String Types to Enums

lixinchengdu:
Small typo fix.


test/src/test/java/org/corfudb/runtime/CompressionTest.java
```diff
@@ -73,7 +73,7 @@ private Object read(CorfuRuntime rt, long address) {
     private List<String> getReadersCodec(String codec) {
         return Stream.of(Codec.Type.values())
                 .map(Codec.Type::name)
-                .filter(type -> type != codec)
+                .filter(type -> !type.equals(codec))
```
xnull: can we pass an enum in the method instead of string?

Maithem: Changed strings to Enums.

PR #2337, opened 2020-01-31
Do not log empty secondary indexes when creating CorfuTable.

pankti-m:
## Overview

Description: When there are no secondary indexes, this log message is unnecessary.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -100,8 +100,12 @@ public class CorfuTable<K ,V> implements
             indexSpec.add(index);
         });
 
-        log.info("CorfuTable: creating CorfuTable with the following indexes: {}",
-                secondaryIndexes.keySet());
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -100,8 +100,12 @@ public CorfuTable(Index.Registry<K, V> indices,
             indexSpec.add(index);
         });
 
-        log.info("CorfuTable: creating CorfuTable with the following indexes: {}",
-                secondaryIndexes.keySet());
+        if (secondaryIndexes.keySet().size() > 0) {
```
Maithem: Just use !secondaryIndexes.isEmpty()?

pankti-m: done

xiaoqin2012:
Not sure if we need this change. With the current code, we know the index is empty. Without logging anything, we are not sure it has secondary indexes or not.

PR #2342, opened 2020-01-31
CorfuStore: Add epoch, CLEAR to streaming API

hisundar:
Consumers of the streaming api need the epoch to
reconstruct the token that represents a Corfu snapshot. 
So modify streaming api to support epoch info too.

Table.clear() is a supported UFO operation.
This must be sent to streaming api too so that it can be handled appropriately.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
@@ -45,21 +45,28 @@ public class CorfuStreamEntry<K extends Message, V extends Message, M extends Me
      */
     public enum OperationType {
         UPDATE,
-        DELETE;
+        DELETE,
+        CLEAR;
     };
 
     @Getter
     private final OperationType operation;
+    /**
+     * Version number of the layout at the time of this entry.
+     */
+    @Getter
+    private final long epoch;
     /**
      * Stream address of this entry
      */
     @Getter
     private final long address;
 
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -68,12 +75,26 @@ public CorfuStreamEntry(K key, V payload, M metadata, long address, OperationTyp
...
+        OperationType operationType;
+        log.trace("fromSRMEntry: Table {} streamer got SMR {}", keyClass.getName(), entry.getSMRMethod());
+        switch(entry.getSMRMethod()) {
+            case "put":
+                operationType = OperationType.UPDATE;
+                break;
+            case "clear":
+                operationType = OperationType.CLEAR;
+                break;
+            default:
```
vjeko: You might want to throw an exception on the default path, and have an explicit delete.

hisundar: good point, done!

hisundar:
> Adding epoch to to the class members of `CorfuStreamEntry` will not automatically fix the collision issue.
> 
> For many java collection, if there exists a hash collision, they will check equality. I think its cleaner to just override equals.

epoch and the collision issue are different. Sorry for the misleading PR comment. I have updated the PR title and comment to reflect the change made. thanks

PR #2369, opened 2020-02-06
Specify log-replication artifact version to revision

dinkar2:
log-replication needs to be at the same ${revision} as all its sibling projects and dependencies.


xnull:
Obsolete?

PR #2372, opened 2020-02-06
Cluster status: add unit tests

xnull:
## Overview

Description:
Cluster status: add unit tests

Why should this be merged: 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -598,4 +505,136 @@ private Layout getHighestEpochLayout(Map<String, Layout> layoutMap) {
                     return true;
                 });
     }
+
+    public static class ClusterHealth {
```
vjeko: Where these functions just moved, or was there any change in logic?

xnull: Just moved from ManagementView to ClusterHealth

runtime/src/test/java/org/corfudb/runtime/view/ClusterHealthTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.corfudb.runtime.view.ClusterStatusReport.ClusterStatus;
+import org.corfudb.runtime.view.ManagementView.ClusterHealth;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ClusterHealthTest {
```
vjeko: Can you please explain what the test does? What is the expected behavior and why.

runtime/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ManagementViewTest {
```
vjeko: Same here.

runtime/src/test/java/org/corfudb/runtime/view/ClusterHealthTest.java
xnull: done

runtime/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
xnull: done

runtime/src/test/java/org/corfudb/runtime/view/ClusterHealthTest.java
```diff
@@ -0,0 +1,36 @@
...
+        final List<String> servers = Arrays.asList(server1, server2, server3);
+        final Layout layout = layoutUtil.getLayout(servers);
+        layout.setUnresponsiveServers(Arrays.asList(server3));
+
+        ClusterHealth clusterHealth = new ClusterHealth();
+        ClusterStatus status = clusterHealth.getClusterHealth(
+                layout, layout.getAllActiveServers()
+        );
+
+        assertThat(status).isEqualTo(ClusterStatus.STABLE);
```
hisundar, line 33: This test is failing in CI. Seems timing related? Can it be made reliable?

xnull: will do it, thank you

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -133,160 +138,41 @@ public void mergeSegments(@Nonnull String endpointToRestoreRedundancy, int retry
...
-     *   reliability of this state.
-     *   For example, in the absence of quorum the system might be in an unstable state which
-     *   cannot converge due to lack of consensus. This is reflected in the
-     *   cluster status report as clusterStatusReliability.
-     *
+     * layout. This information is presented along each node's status (up, down, db_sync).
+     * <p>
+     * It is important to note that as the cluster state is obtained from the layout,
+     * when quorum is not available (majority of nodes) there are lower guarantees on the
+     * reliability of this state.
```
pankti-m, line 150: shouldnt the cluster state always be 'unstable' in the absence of a quorum?

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
```diff
@@ -133,160 +138,41 @@ public void mergeSegments(@Nonnull String endpointToRestoreRedundancy, int retry
...
+     * CLUSTER-SPECIFIC STATUS
+     * - clusterStatus: the cluster status a perceived by the system's layout.
+     * STABLE, DEGRADED, DB_SYNCING or UNAVAILABLE
+     * - nodeStatusMap: each node's status as perceived by the system's layout.
+     * (UP, DOWN or DB_SYNC)
+     * - Cluster Status Reliability: STRONG_QUORUM, WEAK_NO_QUORUM or UNAVAILABLE
+     * <p>
+     * CLIENT-CLUSTER SPECIFIC STATUS:
+     * - clientServerConnectivityStatusMap: the connectivity status of this client to the cluster.
+     * (RESPONSIVE, UNRESPONSIVE).
```
pankti-m, line 175: Can you add an example output in the comment?  It is not clear what all info is displayed/collected..

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
```diff
@@ -375,10 +375,10 @@ public IReplicationProtocol getReplicationProtocol(CorfuRuntime r) {
             }
 
             @Override
-            public ClusterStatus getClusterHealthForSegment(LayoutSegment layoutSegment,
-                                                            Set<String> responsiveNodes) {
-                return !responsiveNodes.containsAll(layoutSegment.getAllLogServers())
-                        ? ClusterStatus.UNAVAILABLE : ClusterStatus.STABLE;
+            public ClusterStatus getClusterHealthForSegment(
+                    LayoutSegment layoutSegment, Set<String> responsiveNodes) {
+                return responsiveNodes.containsAll(layoutSegment.getAllLogServers())
```
pankti-m, line 380: what about the degraded state?

runtime/src/test/java/org/corfudb/runtime/view/ClusterHealthTest.java
zhangn49, line 33: Looks like server3 is still in the layout segment, so log unit servers' health is unavailable. 

The expected overall status should be degraded, right? Since the layout server status is 2/3 active.

runtime/src/main/java/org/corfudb/runtime/view/Layout.java
xnull: it's not possible for a chain replication. If at least one node failed, the status gets UNAVAILABLE

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
xnull: As I know, on the client-side it's not possible to know exactly what the cluster state in case of a failure, but if the client can't connect to a server in the cluster it doesn't mean the cluster in the DEGRADED state.
@annym can explain better than me.

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
xnull: Didn't get your question. There could be only one of those two states RESPONSIVE or UNRESPONSIVE

runtime/src/test/java/org/corfudb/runtime/view/ClusterHealthTest.java
xnull: Sundar, there couldn't be any timing issues (unit test, no external invocations), my test was failing intentionally, I wanted to add more unit tests (that's why I made it failed) when I created this PR and then just forgot to add more tests.
So, I just added more cases for ClusterHealth, now everything is covered

xnull: I just added more tests, your case also covered

>The expected overall status should be degraded, right?

yes

xnull:
@hisundar @pankti-m @zhangn49 I have addressed your comments

runtime/src/main/java/org/corfudb/runtime/view/ManagementView.java
annym: exactly, the fact that a client cannot connect to all 3 servers, does not necessarily imply that the Cluster is UNSTABLE, probably it's all up and running, it's just this client who cannot connect to it. That is why reading the cluster status should consider reliability and connection status, to get an overall look.

runtime/src/test/java/org/corfudb/runtime/view/ClusterHealthTest.java
```diff
@@ -0,0 +1,117 @@
...
+        );
+        assertThat(status).isEqualTo(ClusterStatus.STABLE);
+    }
+
+    /**
+     * Get cluster status from a layout with all responsive nodes.
+     * Expected result is STABLE status
+     */
+    @Test
+    public void testClusterHealth() {
```
annym, line 100: How different is this test from queryClusterStatus in ManagementViewTest ?

This test already exists and it actually is a bit more complete...

runtime/src/test/java/org/corfudb/runtime/view/LayoutUtil.java
```diff
@@ -0,0 +1,30 @@
+package org.corfudb.runtime.view;
+
+import org.corfudb.runtime.view.Layout.LayoutSegment;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+
+public class LayoutUtil {
```
annym, line 10: Please add class description...

runtime/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -0,0 +1,59 @@
...
+        assertThat(quorumLayout).isEqualTo(Optional.of(layout));
+
+        quorumLayout = managementView.getLayoutFromQuorum(layouts, layouts.size());
+        assertThat(quorumLayout).isEqualTo(Optional.of(layout));
+
+        quorumLayout = managementView.getLayoutFromQuorum(layouts, layouts.size() + 1);
+        assertThat(quorumLayout).isEqualTo(Optional.empty());
+    }
+
+    @Test
```
annym, line 46: This test already exists in queryClusterStatus in this same file.

runtime/src/test/java/org/corfudb/runtime/view/ClusterHealthTest.java
xnull: That test is more an Integration Test, this one is a unit test. 
No need to start an entire cluster, no side effects from the environment, the unit test is much simpler. 
I think we need both - integration and unit tests even if some tests are similar.

runtime/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
xnull: The advantages of the unit test:
 - no side-effects (no need to start external services - nutty, no need for network communication, etc)
 - much faster
 - simpler to understand, simpler to find what causes the error.

I could delete those checks in `queryClusterStatus` method if you would like to. 

Also, I think it's ok to have integration and unit tests even if they check similar parameters.

runtime/src/test/java/org/corfudb/runtime/view/LayoutUtil.java
xnull: done

PR #2380, opened 2020-02-07
Redundant stamped lock validation.

WenbinZhu:
## Overview

Remove redundant lock validation as the exclusive write lock is already held.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -244,12 +244,8 @@ public class VersionLockedObject<T extends ICorfuSMR<T>> {
             if (directAccessCheckFunction.apply(this)) {
                 log.trace("Access [{}] Direct (writelock) access at {}", this, getVersionUnsafe());
                 R ret = accessFunction.apply(object.getContext(ICorfuExecutionContext.DEFAULT));
-
-                long versionForCorrectness = getVersionUnsafe();
-                if (lock.validate(ts)) {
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -243,13 +243,7 @@ public void gc(long trimMark) {
...
             if (directAccessCheckFunction.apply(this)) {
                 log.trace("Access [{}] Direct (writelock) access at {}", this, getVersionUnsafe());
-                R ret = accessFunction.apply(object.getContext(ICorfuExecutionContext.DEFAULT));
-
-                long versionForCorrectness = getVersionUnsafe();
-                if (lock.validate(ts)) {
-                    correctnessLogger.trace("Version, {}", versionForCorrectness);
-                    return ret;
-                }
+                return accessFunction.apply(object.getContext(ICorfuExecutionContext.DEFAULT));
```
Maithem: I don't think you need to check if direct access is possible since it is called right after the write lock is acquired. 
Remove it?

Its possible that from the time the optimistic read happened to the time the lock was acquired the version can change back to the original access version, but this if-call is not necessary because if the VLO is at the right version then `updateFunction.accept(this);` will just be a no-op

WenbinZhu: Discussed offline, this can be an optimization in non-transactional read cases.

Maithem:
This seems to be failing the injection framework. Removing my approval till it can be verified.

PR #2385, opened 2020-02-08
move scanAndFilter to StreamingMap

zhangn49:
## Overview

Description:

From a scale test, observed performance regression in scanAndFilter function.
entryStream calls ImmutableListSetWrapper.fromMap, which causes 1-2x more latency than directly using values to filter.

So I add scanAndFilter & scanAndFilterByEntry to StreamingMap Interface to make StreamingMapDecorator and PersistedStreamingMap have different implementation.

Why should this be merged: 

Fix performance regression.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/collections/StreamingMapDecorator.java
```diff
@@ -131,4 +135,25 @@ public void clear() {
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<V> scanAndFilter(Predicate<? super V> valuePredicate) {
+        return mapImpl.values().parallelStream()
+                .filter(valuePredicate)
+                .collect(Collectors.toCollection(ArrayList::new));
+    }
```
vjeko: This is using Java's common pool. I believe this will introduce a regression -- this is why a separate thread pool was introduced.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMapDecorator.java
```diff
@@ -131,4 +135,25 @@ public void clear() {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Collection<Entry<K, V>> scanAndFilterByEntry(
+            Predicate<? super Entry<K, V>> entryPredicate) {
+        return mapImpl.entrySet().parallelStream()
+                .filter(entryPredicate)
+                .collect(Collectors.toCollection(ArrayList::new));
+    }
```
vjeko: Same comment here.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -33,6 +38,19 @@
...
+    // cant use. For example, if there are 4 accessor threads all using the common forkjoin pool, one of the threads
+    // can acquire the VLO lock and cause the other 3 threads to wait, but after acquiring the VLO lock, the thread
+    // gets block on parallel stream, because the pool is exhausted with threads that are trying to acquire the VLO
+    // look, which creates a circular dependency. In other words, a deadlock.
+    private final static ForkJoinPool pool = new ForkJoinPool(Runtime.getRuntime().availableProcessors() - 1,
+            pool -> {
+                final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
+                worker.setName("PersistedStreamingMap-Forkjoin-pool-" + worker.getPoolIndex());
+                return worker;
+            }, null, true);
```
vjeko: `PersistedStreamingMap` cannot take advantage of parallel processing, so there is no point of spawning more than one thread. Is there a reason why `ForkJoinPool` was moved out of `CorfuTable`?

zhangn49: I was not very clear about PersistedStreamingMap.

ForkJoinPool was moved out of CorfuTable because I need `StreamingMapDecorator.scanAndFilter` to get rid of fromMap. If PersistedStreamingMap cannot take advantage of parallel processing, I should put ForkJoinPool to StreamingMapDecorator.

vjeko: Just to give you a bit of context.

`PersistedStreamingMap` uses `RocksDB` underneath, which by design does not allow multi-threaded data iteration. So in this case, you can just scan the data using the same client thread -- no need for a new thread pool executor.

`StreamingMapDecorator` on the other hand has no such restrictions, and this is why `parallelStream()` can be used. However, there was a bug with using a common thread pool that would cause deadlock in some circumstances. This is why `ForkJoinPool` was introduced.

zhangn49: Thanks for the context. I thought RocksDB iterator could somehow split data and operate in parallel so moved it here.

I have moved it to `StreamingMapDecorator`

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMapDecorator.java
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMapDecorator.java
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
```diff
@@ -275,6 +278,27 @@ public void clear() {
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Collection<Entry<K, V>> scanAndFilterByEntry(
+            Predicate<? super Entry<K, V>> entryPredicate) {
+        return entryStream()
+                .filter(entryPredicate)
+                .collect(Collectors.toCollection(ArrayList::new));
+    }
+
```
Maithem: Maybe we can do this in another patch, but we need to get rid of this.

If the table set cannot fit in memory, then its possible that the resulting set won't fit in memory.
Applications can just directly use the Stream exposed by entryStream.

vjeko:
I am hoping that we clean up the interface a bit.

The idea behind the `StreamingMap` was that it provides a streaming interface on top of a map. With this patch, we would be adding `scanAndFilter` functionality which breaks the existing encapsulation (`scanAndFilter` is a higher level functionality that `StreamingMap` should not concern itself with).

I am suggesting you augment the `StreamingMap`:
```
public interface StreamingMap<K, V> extends Map<K, V> {

    // Maybe this can be renamed to immutableEntryStream()
    Stream<Map.Entry<K, V>> entryStream();

    default Stream<Map.Entry<K, V>> mutableEntryStream() {
        return entryStream();
    }
}
```

Where for the StreamingMapDecorator, you would override the default functionality: 
```
    @Override
    public Stream<Entry<K, V>> mutableEntryStream() {
        return mapImpl.entrySet().stream().parallel();
    }
```

From there on, within the `CorfuTable` for `scanAndFilter()`, you can do:

```
        return pool.submit(() -> mainMap.mutableEntryStream()
                .map(Entry::getValue).filter(valuePredicate)
                .collect(Collectors.toCollection(ArrayList::new))).join();
```

This way, you don't even need to touch `ForkJoinPool`.

WenbinZhu:
If the problem is entryStream() being slow, then every caller of it suffers from this performace issue, not only scanAndFilter. For example, I saw checkpointWriter is also using entryStream(). So is it better to fix entryStream() instead of hacking scanAndFilter()?

vjeko:
> If the problem is entryStream() being slow, then every caller of it suffers from this performace issue, not only scanAndFilter. For example, I saw checkpointWriter is also using entryStream(). So is it better to fix entryStream() instead of hacking scanAndFilter()?

I suggest we deal with `checkpointWriter` separately. Most of the time taken by the `checkpointWriter` is taken by the write path, so the performance hit on the read path should be much less visible.

zhangn49:
> I am hoping that we clean up the interface a bit.
> 
> The idea behind the `StreamingMap` was that it provides a streaming interface on top of a map. With this patch, we would be adding `scanAndFilter` functionality which breaks the existing encapsulation (`scanAndFilter` is a higher level functionality that `StreamingMap` should not concern itself with).
> 
> I am suggesting you augment the `StreamingMap`:
> 
> ```
> public interface StreamingMap<K, V> extends Map<K, V> {
> 
>     // Maybe this can be renamed to immutableEntryStream()
>     Stream<Map.Entry<K, V>> entryStream();
> 
>     default Stream<Map.Entry<K, V>> mutableEntryStream() {
>         return entryStream();
>     }
> }
> ```
> 
> Where for the StreamingMapDecorator, you would override the default functionality:
> 
> ```
>     @Override
>     public Stream<Entry<K, V>> mutableEntryStream() {
>         return mapImpl.entrySet().stream().parallel();
>     }
> ```
> 
> From there on, within the `CorfuTable` for `scanAndFilter()`, you can do:
> 
> ```
>         return pool.submit(() -> mainMap.mutableEntryStream()
>                 .map(Entry::getValue).filter(valuePredicate)
>                 .collect(Collectors.toCollection(ArrayList::new))).join();
> ```
> 
> This way, you don't even need to touch `ForkJoinPool`.

Yes, it looks cleaner now.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -368,17 +365,11 @@ public void insert(@ConflictParameter K key, V value) {
...
-     * @param valuePredicate java predicate (function to evaluate)
-     * @return a view of the values contained in this map meeting the predicate condition.
-     */
+    /** {@inheritDoc} */
+    @Override
     @Accessor
     public List<V> scanAndFilter(Predicate<? super V> valuePredicate) {
-        return pool.submit(() -> mainMap.entryStream()
+        return pool.submit(() -> mainMap.mutableEntryStream()
                 .map(Entry::getValue).filter(valuePredicate)
```
Maithem: Can you use try-resources on the stream so that It is explicitly closed after its used.

Like this 

https://github.com/CorfuDB/CorfuDB/commit/e8f154e9330bf164003147b16c50fd77c6df58a5#diff-6f23471c25711fa551fa3b560d27d49dR154

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMap.java
```diff
@@ -21,4 +20,8 @@
      * @return stream of entries
      */
     Stream<Map.Entry<K, V>> entryStream();
+
+    default Stream<Map.Entry<K, V>> mutableEntryStream() {
```
Maithem: I think `unsafeStream` is a more appropriate adjective. It also makes it clear that it can only be used under an `Accessor` lock.

I would also prefer to not have a default method. Just stub it out in the implementation.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
zhangn49: done.

PR #2387, opened 2020-02-10
A workaround to hole in tail issue.

zhangn49:
## Overview

Description:

When a stream's tail is a hole, VLO will try to sync up to tail but never add hole to its resolved queue, which means each sync will query sequencer again.

@lixinchengdu and I bring a workaround that adds hole to resolved queue to avoid re-sync.

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
@@ -96,6 +96,7 @@ public class StreamViewSMRAdapter implements ISMRStream {
      */
     public List<SMREntry> current() {
         ILogData data = streamView.current();
+
         if (data == null
                 || data.getType() != DataType.DATA
                 || !(data.getPayload(runtime) instanceof ISMRConsumable)) {
@@ -113,6 +114,7 @@ public class StreamViewSMRAdapter implements ISMRStream {
      */
     public List<SMREntry> previous() {
         ILogData data = streamView.previous();
+
         while (Address.isAddress(streamView.getCurrentGlobalPosition())
                 && data != null) {
             if (data.getType() == DataType.DATA
--- a/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
@@ -2,6 +2,7 @@ package org.corfudb.runtime.object;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.google.common.annotations.VisibleForTesting;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.logprotocol.SMREntry;
@@ -627,7 +628,10 @@ public class VersionLockedObject<T extends ICorfuSMR<T>> {
 
         List<SMREntry> entries = stream.current();
 
-        while (!entries.isEmpty()) {
+        while (stream.pos() > rollbackVersion) {
+            if (Address.nonAddress(stream.pos())) {
+                throw new NoRollbackException(stream.pos(), rollbackVersion);
+            }
             if (entries.stream().allMatch(SMREntry::isUndoable)) {
                 // start from the end, process one at a time
                 ListIterator<SMREntry> it = entries.listIterator(entries.size());
@@ -646,13 +650,7 @@ public class VersionLockedObject<T extends ICorfuSMR<T>> {
             }
 
             entries = stream.previous();
-
-            if (stream.pos() <= rollbackVersion) {
-                return;
-            }
         }
-
-        throw new NoRollbackException(stream.pos(), rollbackVersion);
     }
 
     /**
@@ -719,6 +717,11 @@ public class VersionLockedObject<T extends ICorfuSMR<T>> {
     public void applyUpdateToStreamUnsafe(SMREntry entry, long globalAddress) {
         applyUpdateUnsafe(entry, globalAddress);
         seek(globalAddress + 1);
+    }
+
+    @VisibleForTesting
+    public ISMRStream getSmrStream() {
+        return smrStream;
     }
 
     /**
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
@@ -290,9 +290,7 @@ public abstract class AbstractContextStreamView<T extends AbstractStreamContext>
      */
     private void updatePointer(final ILogData data) {
         // Update the global pointer, if it is non-checkpoint data.
-        if ((data.isData() && !data.hasCheckpointMetadata()) ||
-                (data.isHole() && (data.getBackpointerMap().isEmpty() || data.containsStream(getCurrentContext().id)))
-        ) {
+        if ((data.isData() || data.isHole()) && !data.hasCheckpointMetadata()) {
             // Note: here we only set the global pointer and do not validate its position with respect to the trim mark,
             // as the pointer is expected to be moving step by step (for instance when syncing a stream up to maxGlobal)
             // The validation is deferred to these methods which call it in advance based on the expected final position
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
@@ -9,7 +9,6 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-import org.corfudb.protocols.wireprotocol.DataType;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.view.Address;
@@ -134,6 +133,13 @@ public abstract class AbstractContextStreamView<T extends AbstractStreamContext>
         if (entry != null) {
             // Update the pointer.
             updatePointer(entry);
+
+            // We added hole to StreamView layer, in order to enable VLO sync to a hole
+            // and in this way, we can avoid unnecessary sync that call sequencer every time.
+            // It can expose a hole to sv consumer, so check if entry is a hole.
+            if (entry.isHole()) {
+                return nextUpTo(maxGlobal);
+            }
 
             // Process the next entry, checking if the context has changed.
             // If the context has changed, we read again, since this entry
@@ -284,7 +290,9 @@ public abstract class AbstractContextStreamView<T extends AbstractStreamContext>
      */
     private void updatePointer(final ILogData data) {
         // Update the global pointer, if it is non-checkpoint data.
-        if (data.getType() == DataType.DATA && !data.hasCheckpointMetadata()) {
+        if ((data.isData() && !data.hasCheckpointMetadata()) ||
+                (data.isHole() && (data.getBackpointerMap().isEmpty() || data.containsStream(getCurrentContext().id)))
+        ) {
             // Note: here we only set the global pointer and do not validate its position with respect to the trim mark,
             // as the pointer is expected to be moving step by step (for instance when syncing a stream up to maxGlobal)
             // The validation is deferred to these methods which call it in advance based on the expected final position
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
@@ -84,11 +84,13 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
                 // to be serviced immediately, rather than reading one entry at a time.
                 ld = read(currentRead, queue);
 
-                if (queue == getCurrentContext().readQueue && ld != null) {
+                if (queue == getCurrentContext().readQueue) {
                     // Validate that the data entry belongs to this stream, otherwise, skip.
                     // This verification protects from sequencer regression (tokens assigned in an older epoch
                     // that were never written to, and reassigned on a newer epoch)
-                    if (ld.containsStream(this.id) && ld.getType() == DataType.DATA) {
+                    if ((ld.containsStream(this.id) && ld.isData()) ||
+                            (ld.isHole() && (ld.getBackpointerMap().isEmpty() || ld.containsStream(this.id)))
+                    ) {
                         addToResolvedQueue(getCurrentContext(), currentRead);
                         readNext = false;
                     } else {
--- a/test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
+++ b/test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
@@ -8,17 +8,26 @@ import java.util.Collection;
 import java.util.ConcurrentModificationException;
 import java.util.Map;
 import java.util.Random;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import org.assertj.core.api.Assertions;
 import org.assertj.core.data.MapEntry;
+import org.corfudb.protocols.wireprotocol.LogData;
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
+import org.corfudb.runtime.object.CorfuCompileProxy;
+import org.corfudb.runtime.object.ICorfuSMR;
+import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.runtime.view.AbstractViewTest;
 import org.junit.Test;
 
 public class CorfuTableTest extends AbstractViewTest {
+
+    private static final int ITERATIONS = 20;
 
     Collection<String> project(Collection<Map.Entry<String, String>> entries) {
         return entries.stream().map(entry -> entry.getValue()).collect(Collectors.toCollection(ArrayList::new));
@@ -244,5 +253,50 @@ public class CorfuTableTest extends AbstractViewTest {
 
         final Stream<Map.Entry<Integer, Integer>> result = map.entryStream();
         result.forEach(e -> map.put(new Random().nextInt(), 0));
+    }
+
+    @Test
+    @SuppressWarnings({"unchecked", "checkstyle:magicnumber"})
+    public void canHandleHoleInTail() {
+        UUID streamID = UUID.randomUUID();
+
+        CorfuTable<String, String> corfuTable = getDefaultRuntime()
+                .getObjectsView()
+                .build()
+                .setTypeToken(new TypeToken<CorfuTable<String, String>>() {})
+                .setStreamID(streamID)
+                .open();
+
+        corfuTable.put("k1", "dog fox cat");
+        corfuTable.put("k2", "dog bat");
+        corfuTable.put("k3", "fox");
+
+        // create a hole
+        TokenResponse tokenResponse =  getDefaultRuntime()
+                .getSequencerView()
+                .next(streamID);
+
+        Token token = tokenResponse.getToken();
+
+        getDefaultRuntime().getAddressSpaceView()
+                .write(tokenResponse, LogData.getHole(token));
+
+        assertThat(getDefaultRuntime().getAddressSpaceView()
+                .read(token.getSequence()).isHole()).isTrue();
+
+        for (int i = 0; i < ITERATIONS; i++) {
+            getDefaultRuntime().getObjectsView().TXBuild()
+                    .type(TransactionType.SNAPSHOT)
+                    .snapshot(token)
+                    .build()
+                    .begin();
+
+            corfuTable.scanAndFilter(item -> true);
+            getDefaultRuntime().getObjectsView().TXEnd();
+        }
+
+        assertThat(((CorfuCompileProxy) ((ICorfuSMR) corfuTable).
+                getCorfuSMRProxy()).getUnderlyingObject().getSmrStream().pos()).isEqualTo(3);
+
     }
 }
--- a/runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
+++ b/runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
@@ -71,4 +71,11 @@ public class Serializers {
             throw new RuntimeException(msg);
         }
     }
+
+    /**
+     * Clear custom serializers.
+     */
+    public static synchronized void clearCustomSerializers() {
+        customSerializers.clear();
+    }
 }
--- a/test/src/test/java/org/corfudb/protocols/logprotocol/LogEntryTest.java
+++ b/test/src/test/java/org/corfudb/protocols/logprotocol/LogEntryTest.java
@@ -6,6 +6,7 @@ import org.corfudb.CustomSerializer;
 import org.corfudb.runtime.exceptions.SerializerException;
 import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
+import org.junit.After;
 import org.junit.Test;
 
 import java.util.Arrays;
@@ -17,6 +18,11 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class LogEntryTest {
+
+    @After
+    public void tearDown() {
+        Serializers.clearCustomSerializers();
+    }
 
     @Test
     public void seekToEndSMREntry () {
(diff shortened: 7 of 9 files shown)


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -629,27 +631,40 @@ protected void rollbackStreamUnsafe(ISMRStream stream, long rollbackVersion) {
...
+                    }
+                } else {
+                    Optional<SMREntry> entry = entries.stream().findFirst();
+                    if (log.isTraceEnabled()) {
+                        log.trace("rollbackStreamUnsafe: one or more stream entries in address @{} are not undoable. " +
+                                        "Undoable entries: {}/{}", stream.pos(),
+                                (int) entries.stream().filter(SMREntry::isUndoable).count(),
+                                entries.size());
+                    }
+                    throw new NoRollbackException(entry, stream.pos(), rollbackVersion);
```
Maithem: I think its cleaner to just have an SMR that is a whole which is always undoable. When applyUndoRecordUnsafe is called, it just results in a no-op.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -96,6 +97,10 @@ public void gc(long trimMark) {
      */
     public List<SMREntry> current() {
         ILogData data = streamView.current();
+        if (data != null && data.isHole()) {
+            throw new HoleInStreamException();
+        }
+
```
vjeko: Rather than signaling this condition via an exception, maybe a better approach would be to have a globally unique instance that signals a hole.

Maithem:
Is this ready to be reviewed ?

zhangn49:
> Is this ready to be reviewed ?

Yes, it is ready.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -571,6 +577,10 @@ private Object applyUpdateUnsafe(SMREntry entry, long timestamp) {
                 Address.isAddress(entry.getGlobalAddress()) ? entry.getGlobalAddress() : "OPT",
                 entry.getSMRArguments());
 
+        if (SMREntry.isNOP(entry)) {
```
Maithem: Can you move this to the caller? if its a no-op, then you skip calling applyUpdateUnsafe

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -84,11 +86,16 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
...
                     // This verification protects from sequencer regression (tokens assigned in an older epoch
                     // that were never written to, and reassigned on a newer epoch)
-                    if (ld.containsStream(this.id) && ld.getType() == DataType.DATA) {
+                    if (ld.getType() == DataType.HOLE || (ld.containsStream(this.id) &&
+                            ld.getType() == DataType.DATA)) {
+                        // Convert hole to NOP
+                        if (ld.isHole()) {
+                            ld = new LogData(DataType.DATA, SMREntry.getNOP(), ld.getMetadataMap());
+                        }
                         addToResolvedQueue(getCurrentContext(), currentRead);
```
Maithem: This layer only understands `ILogData` can you move this conversion to `StreamViewSMRAdapter` which actually maps `ILogData` to `SMREntry`

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
```diff
@@ -245,4 +254,46 @@ public void snapshotInvariant() {
...
+        corfuTable.put("k3", "fox");
+
+        // create a hole
+        TokenResponse tokenResponse =  getDefaultRuntime()
+                .getSequencerView()
+                .next(streamID);
+
+        Token token = tokenResponse.getToken();
+
+        getDefaultRuntime().getAddressSpaceView()
```
Maithem: Can you read back the address and make sure its a hole?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
Maithem: Actually, if `StreamViewSMRAdapter` filter's these holes out then you dont need any changes in the VLO.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
```diff
@@ -68,6 +77,11 @@
     @Getter
     public transient boolean haveUpcallResult = false;
 
+    /** If smrEntry is NOP for hole */
+    public static boolean isNOP(SMREntry smrEntry) {
+        return smrEntry == NOP;
+    }
+
```
Maithem: Yea can get rid of the NOP notion by propagating the holes to the Stream layer but then filtering them out just before consumption in the `StreamViewSMRAdapter`

test/src/test/java/org/corfudb/runtime/collections/CorfuTableTest.java
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
zhangn49: previous and current will return an empty list<SMREntry>, so need to change that while loop a little bit.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
zhangn49: We changed to put hole into ResolvedQueue, and return an empty list<SMREntry> to be ignored by VLO.

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -96,6 +96,7 @@ public void gc(long trimMark) {
      */
     public List<SMREntry> current() {
         ILogData data = streamView.current();
+
```
Maithem: @lixinchengdu I think your initial argument that the object layer should always see `SMREntry` makes more sense. But even on master, it seems like an empty list is returned (look at line 100). 

I think returning an empty list is error-prone can hide issues, maybe we should explicitly fail if the entry is not `ISMRConsumable` or a `DataType.DATA`

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -96,6 +96,7 @@ public void gc(long trimMark) {
...
         if (data == null
                 || data.getType() != DataType.DATA
                 || !(data.getPayload(runtime) instanceof ISMRConsumable)) {
```
Maithem: Sorry for going back and forth on this, but can we reintroduce the NO-OP SMR ? I think its better to be more concise and handle holes explicitly in the object layer instead of just using an empty list as a signal.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -631,7 +632,10 @@ protected void rollbackStreamUnsafe(ISMRStream stream, long rollbackVersion) {
 
         List<SMREntry> entries = stream.current();
 
-        while (!entries.isEmpty()) {
+        while (stream.pos() > rollbackVersion) {
```
Maithem: If the SMREntry is a no-op we can skip calling applyUndo, likewise we can skip the apply on stream forward?

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -88,7 +88,8 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
                     // Validate that the data entry belongs to this stream, otherwise, skip.
                     // This verification protects from sequencer regression (tokens assigned in an older epoch
                     // that were never written to, and reassigned on a newer epoch)
-                    if (ld.containsStream(this.id) && ld.getType() == DataType.DATA) {
+                    if (ld.getType() == DataType.HOLE || (ld.containsStream(this.id) &&
```
Maithem: By the way this will break existing consumers of IStreamView (i.e. transaction pollers), because now see unexpected holes.

vjeko: They should not IMO. What entries are seen by the client is dictated by getNextEntries() and that function filters out the holes.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -631,7 +632,10 @@ protected void rollbackStreamUnsafe(ISMRStream stream, long rollbackVersion) {
...
+            if (Address.nonAddress(stream.pos())) {
+                throw new NoRollbackException(stream.pos(), rollbackVersion);
+            }
             if (entries.stream().allMatch(SMREntry::isUndoable)) {
```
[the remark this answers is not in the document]
zhangn49: Yes, hole entry will skip this if block.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -84,6 +84,7 @@
     /**
      * The stream view this object is backed by.
      */
+    @Getter
```
annym: Can you please add a note that this getter is for testing purposes or perhaps the annotation VisibleForTesting

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -84,11 +84,11 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
                 // to be serviced immediately, rather than reading one entry at a time.
                 ld = read(currentRead, queue);
 
-                if (queue == getCurrentContext().readQueue && ld != null) {
+                if (queue == getCurrentContext().readQueue) {
                     // Validate that the data entry belongs to this stream, otherwise, skip.
                     // This verification protects from sequencer regression (tokens assigned in an older epoch
                     // that were never written to, and reassigned on a newer epoch)
-                    if (ld.containsStream(this.id) && ld.getType() == DataType.DATA) {
+                    if (ld.isHole() || (ld.containsStream(this.id) && ld.isData())) {
```
annym: I am a bit confused, why do we want to surface the hole itself to the object layer? wouldn't that give the wrong behavior? If we are enforcing holes for different reasons (materialize a stream or any other reason), a client should not perceive it...If we are directly consuming the stream it would perceive them right?

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -284,7 +284,8 @@ protected boolean doesEntryUpdateContext(final ILogData data) {
      */
     private void updatePointer(final ILogData data) {
         // Update the global pointer, if it is non-checkpoint data.
-        if (data.getType() == DataType.DATA && !data.hasCheckpointMetadata()) {
+        if ((data.getType() == DataType.DATA || data.getType() == DataType.HOLE)
```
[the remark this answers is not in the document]
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
zhangn49: Sure, I will add a VisibleForTesting function.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
zhangn49: Do you mean we should add hole to ResolvedQueue in AbstractQueuedStreamView?

annym: We are adding it here, right? Which I believe is correct... My question is more like, when it's a hole readNext is being set to false meaning that the upper layer receives this hole. So if you are directly manipulating a stream a nextUpTo would return a hole at some point, or is it further filtered? Maybe I did not see when we filter it in upper layers

annym: @zhangn49 This is what I mean (run this test). To someone consuming the stream directly we are exposing holes (null entries) which should be transparent to a reader (these holes are our way to avoid other issues with checkpoints):


    @Test
    public void testNextUpTo() {
        final int numWrites = 10;

        byte[] testPayload = "hello world".getBytes();

        String stream = "stream1";
        UUID id1 = CorfuRuntime.getStreamID(stream);

        StreamingMap<String, String> map = r.getObjectsView()
                .build()
                .setStreamName(stream)
                .setTypeToken(new TypeToken<CorfuTable<String, String>>() {})
                .open();

        CorfuRuntime producer = getNewRuntime(getDefaultNode())
                .connect();

        CorfuRuntime consumer = getNewRuntime(getDefaultNode())
                .connect();

        IStreamView svProducer = producer.getStreamsView().get(id1);
        IStreamView svConsumer = consumer.getStreamsView().get(id1);


        for (int x = 0; x < numWrites; x++) {
            svProducer.append(testPayload);
        }

        // Emulate the extra Token the Checkpointer requests before appending a checkpoint
        // producer.getSequencerView().next(id1);
        // producer.getAddressSpaceView().write(new Token(0, numWrites), LogData.getHole(numWrites));

        // Get two entries before checkpointing (so we don't load from checkpoint on next access)
        assertThat(svConsumer.nextUpTo(numWrites).getPayload(consumer)).isEqualTo(testPayload);
        assertThat(svConsumer.nextUpTo(numWrites).getPayload(consumer)).isEqualTo(testPayload);

        MultiCheckpointWriter checkpointWriter = new MultiCheckpointWriter();
        checkpointWriter.addMap(map);
        checkpointWriter.appendCheckpoints(r, "Author");

        // Consume Until Tail
        long tail = consumer.getSequencerView().query(id1);
        System.out.println("Tail is: " + tail);

        ILogData data;
        while (svConsumer.hasNext()) {
            data = svConsumer.nextUpTo(tail);
            System.out.println("Next @" + data.getGlobalAddress() + " bp is: " + data.getBackpointer(id1));
            assertThat(data.getPayload(consumer)).isEqualTo(testPayload);

        }

    }

zhangn49: I see. Since we need VLO's version to point to a hole to avoid unnecessary sync, we are adding hole to streamView. 

getNextEntries is safe now, since we filter hole out.
For nextUpTo/next, it will return a hole to direct sv consumers. It seems we don't have direct consumer right now.

I am thinking if I can do this in nextUpTp, to skip hole.
```
        if (entry != null) {
            // Update the pointer.
            updatePointer(entry);

            // Check if entry is a hole
            if (entry.isHole()) {
                return nextUpTo(maxGlobal);
            }

            // Process the next entry, checking if the context has changed.
            // If the context has changed, we read again, since this entry
            // does not contain any data, and we need to follow the new
            // context.
            if (processEntryForContext(entry)) {
                return nextUpTo(maxGlobal);
            }
        }
```

In this case, hasNext() will return true when only hole left in this stream, but looks still consistent with its definition.

```
/** Returns whether or not there are potentially more entries in this
     * stream - this function may return true even if there are no entries
     * remaining, as addresses may have been acquired by other clients
     * but not written yet, or the addresses were hole-filled, or just failed.
     * @return      True, if there are potentially more entries in the stream.
     */
    boolean hasNext();
```

annym: I Think all you need to do, is in the AddressMapStreamView removeFromQueue, if it is a hole, set readNext=True... and this will automatically filter the hole for you to upper layers. So do something like:


if (ld.isHole() || (ld.containsStream(this.id) && ld.isData())) {
                        addToResolvedQueue(getCurrentContext(), currentRead);
                        readNext = ld.isHole()? true: false;		                         
}

zhangn49: Discussed offline, skip hole in nextUpTo to update pointer.
Added comment why we add hole and skip it, also added this unit test to StreamViewTest.

zhangn49:
@hisundar @xnull @WenbinZhu 
Could you please help merge it? Thank you

zhangn49:
> Did Vjko's patch already solved this issue? He updates the global pointer to the hole position, so next time it won't go to sequencer right?

We discussed last time, and found his patch has an issue about rollback. So I think this one is still necessary.

WenbinZhu:
> > Did Vjko's patch already solved this issue? He updates the global pointer to the hole position, so next time it won't go to sequencer right?
> 
> We discussed last time, and found his patch has an issue about rollback. So I think this one is still necessary.

I think we need to do several things before merging this as changes in stream layer are risky:
1) discuss the problem with Vjeko path and difference between this patch
2) revert Vjeko's patch if necessary
3) merge this patch

zhangn49:
> > > Did Vjko's patch already solved this issue? He updates the global pointer to the hole position, so next time it won't go to sequencer right?
> > 
> > 
> > We discussed last time, and found his patch has an issue about rollback. So I think this one is still necessary.
> 
> I think we need to do several things before merging this as changes in stream layer are risky:
> 
> 1. discuss the problem with Vjeko path and difference between this patch
> 2. revert Vjeko's patch if necessary
> 3. merge this patch

As I explained in Slack, this patch is a supplement of 2420. Please help review it when you get a chance, and I will modify it according to your feedback.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -135,6 +134,13 @@ public final synchronized ILogData nextUpTo(final long maxGlobal) {
             // Update the pointer.
             updatePointer(entry);
 
+            // We added hole to StreamView layer, in order to enable VLO sync to a hole
+            // and in this way, we can avoid unnecessary sync that call sequencer every time.
+            // It can expose a hole to sv consumer, so check if entry is a hole.
+            if (entry.isHole()) {
+                return nextUpTo(maxGlobal);
```
WenbinZhu, line 141: Resursive call is not good practice in production, it could blow up stack. I know you are referencing from the recursive code right below, but that is basically dead code.

Also, why don't we do this in `removeFromQueue`?  We can read the next one if current is a hole in `removeFromQueue` right? That seems cleaner.

annym:
> Did Vjko's patch already solved this issue? He updates the global pointer to the hole position, so next time it won't go to sequencer right?

So, Vjeko's patch does this but only for the path of direct stream access. Nan's patch does it for the Object Layer access. And we agreed he would also solve an issue that derives from this pointer being moved to a hole.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
WenbinZhu, line 91: I feel this might not be completely correct in case of sequencer regress.  I think we need to check containsStream() even for a hole, as a hole could be a real hole, which has no stream info, or a no-op which still belongs to some stream. Consider this case:
1) In epoch 0, sequencer issues a token with address 100 to stream A.
2) No one writes to this token, but the client queries sequencer for stream A and get back a bitmap that 100 belongs to stream A.
3) Epoch changes to 1, new sequencer boots up, recovers the stream tail map from log unit, since no one writes to that token with address 100, log unit does not know about that token, neither does new sequencer.
4) New sequencer issues token 100 to stream B to new client 
5) New client uses this token to enforce a hole for stream B (the checkpoint hole).
6) Now the old client which operates on stream A finish the read, but it reads a hole with backpointer of stream B, since it's a hole, you will update the stream A's pointer to point to 100, which belongs to stream B and also put it in to the resolved queue.

So in this case we end up pointing stream A's pointer to a hole that belongs to stream B and put it into A's resolved queue.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -284,7 +290,7 @@ protected boolean doesEntryUpdateContext(final ILogData data) {
      */
     private void updatePointer(final ILogData data) {
         // Update the global pointer, if it is non-checkpoint data.
-        if (data.getType() == DataType.DATA && !data.hasCheckpointMetadata()) {
+        if ((data.isData() || data.isHole()) && !data.hasCheckpointMetadata()) {
```
WenbinZhu, line 293: Same here, we might end up pointing to an address belongs to a different stream.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
zhangn49: Yes, I agree that recursive call is not good.

If we have a hole in streamA's tail, we read this hole can continue to read the next one in removeFromQueue, and return null. Then it won't call updatePointer(entry), so pointer is still behind this hole and we do not improve performance.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
zhangn49: Yes, I added check containsStream() for a no-op hole

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
WenbinZhu: This the problem of existing code organization, we need to clean it up later.
Though I think this does not matter in your case, as long as you put the hole into the resolvedQueue via `addToResolvedQueue`, the maxResolution will be updated: https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java#L87
 then next time maxResolution will prevent extra sequencer call. But using globalPointer instead of maxResolution can prevent unnecessary TrimmedException, so it's ok for now. I think later on we should reorg this part of code.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -284,7 +290,9 @@ protected boolean doesEntryUpdateContext(final ILogData data) {
      */
     private void updatePointer(final ILogData data) {
         // Update the global pointer, if it is non-checkpoint data.
-        if (data.getType() == DataType.DATA && !data.hasCheckpointMetadata()) {
+        if ((data.isData() && !data.hasCheckpointMetadata()) ||
+                (data.isHole() && (data.getBackpointerMap().isEmpty() || data.containsStream(getCurrentContext().id)))
```
WenbinZhu, line 294: Same as last comment, this can be simplified.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -84,11 +84,13 @@ protected ILogData removeFromQueue(NavigableSet<Long> queue) {
...
                 ld = read(currentRead, queue);
 
-                if (queue == getCurrentContext().readQueue && ld != null) {
+                if (queue == getCurrentContext().readQueue) {
                     // Validate that the data entry belongs to this stream, otherwise, skip.
                     // This verification protects from sequencer regression (tokens assigned in an older epoch
                     // that were never written to, and reassigned on a newer epoch)
-                    if (ld.containsStream(this.id) && ld.getType() == DataType.DATA) {
+                    if ((ld.containsStream(this.id) && ld.isData()) ||
+                            (ld.isHole() && (ld.getBackpointerMap().isEmpty() || ld.containsStream(this.id)))
```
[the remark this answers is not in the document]
WenbinZhu: Sorry please ignore the previous commnet, I forgot we also need to include a real hole.

I think we need to add a special NO_OP type, it's much cleaner in logic, and could solve the recursive call issue (nextUpTo does not need to retry for a NO_OP entry, just return to upper layer), if you would like to do...

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
```diff
@@ -284,7 +290,9 @@ protected boolean doesEntryUpdateContext(final ILogData data) {
...
+        ) {
```
WenbinZhu, line 295: I think we probably don't need to check containsStream here because when we call this method, logdata that does not belong to this stream should already be filtered out, could you double check?

zhangn49: yes, removed.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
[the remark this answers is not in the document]
zhangn49: NO_OP type is cleaner in logic, but it involves more code reorg... I think we can combine it with further cleanup.

PR #2388, opened 2020-02-11
Implement keySet

Maithem:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java
@@ -1,6 +1,5 @@
 package org.corfudb.runtime.collections;
 
-import com.google.common.collect.Streams;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import lombok.NonNull;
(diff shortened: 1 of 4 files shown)


hisundar:
> This does not look like the right approach since we are not guaranteed that the key-set will fit in-memory.
If key-set does not fit in memory that would be the consumer's problem and corfu cannot be penalized for providing it, right?
Either way it does not seem right to have a public API that uses this feature and keep it as Unsupported.

vjeko:
> > If key-set does not fit in memory that would be the consumer's problem and corfu cannot be penalized for providing it, right?
> > Either way it does not seem right to have a public API that uses this feature and keep it as Unsupported.

All APIs that are implemented as part of `PersistentStreamingMap` should _always_ work regardless of the data-set size. This patch breaks this assumption.

If you really want to provide this kind of an API, I suggest this change be done in the UFO layer where you can retrieve all keys by doing `entryStream().map(Entry::getKey)collect(...)`.

hisundar:
> > > If key-set does not fit in memory that would be the consumer's problem and corfu cannot be penalized for providing it, right?
> > > Either way it does not seem right to have a public API that uses this feature and keep it as Unsupported.
> 
> All APIs that are implemented as part of `PersistentStreamingMap` should _always_ work regardless of the data-set size. This patch breaks this assumption.
> 

That is subjective. One can just as well say that given that `PersistentStreamingMap` which appears to be a type of `CorfuTable` implementation to all clients should implement all API that CorfuTable does. This fixes that assumption for one of these API.
If that's not a valid assumption then the right approach would be that PersistentStramingMap should be completely disassociated from CorfuTable and be a new type altogether.
 
> If you really want to provide this kind of an API, I suggest this change be done in the UFO layer where you can retrieve all keys by doing `entryStream().map(Entry::getKey)collect(...)`.

This is a good suggestion. I will work on this separately too. They will still ask for `keySet()` because that gives them full control over the batching and pagination.

PR #2396, opened 2020-02-12
Add searchable objects and DCN streams map to CorfuRuntime

pankti-m:
Add searchable objects and DCN streams map to CorfuRuntime and add them to affected streams for a transaction.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -52,9 +52,11 @@ import java.lang.Thread.UncaughtExceptionHandler;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


Maithem:
I think we need to introduce a more general "Transaction Tagging" feature. When an object is opened, a transaction stream, or streams are passed to the builder. For example:

```java
r.getObjectsView()
               .build()
               .setStreamName(name) 
               .setTypeToken(tType)  
               .setTransactionStreams(uuid1, uuid2) 
               .open(); 
```


Then when an object is part of the transaction, the TransactionalContext will keep track of the objects that the transaction has touched. Just before committing you can iterate through the affected objects and collect the transaction streams that were passed to the builder.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -108,6 +110,9 @@
          */
         @Default
         Duration fastLoaderTimeout = Duration.ofMinutes(30);
+
+        @Default
+        Map<UUID, Set<UUID>> streamsMap = new HashMap<>();
```
hisundar: how is this being consumed?

PR #2401, opened 2020-02-13
Add support for Stream Tagging.

pankti-m:
## Overview

Description:  Add support for stream tagging.

Why should this be merged: 
Tags will be user-defined and used to log writes for a table in the appropriate
stream.  This can be leveraged to have more control over what gets added to the
transaction stream and other streams.

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
@@ -20,6 +20,7 @@ import org.corfudb.runtime.object.ICorfuVersionPolicy;
 import org.corfudb.runtime.object.transactions.TransactionType;
 
 import org.corfudb.runtime.object.transactions.TransactionalContext;
+import org.corfudb.runtime.view.SMRObject;
 import org.corfudb.util.serializer.ISerializer;
 import lombok.Getter;
 
@@ -58,6 +59,9 @@ public class Table<K extends Message, V extends Message, M extends Message> {
      * @param valueSchema             Value schema to identify secondary keys.
      * @param corfuRuntime            Connected instance of the Corfu Runtime.
      * @param serializer              Protobuf Serializer.
+     * @param streamingMapSupplier    Streaming Map Supplier
+     * @param versionPolicy           Corfu Version Policy
+     * @param streamTags              Stream Tags
      */
     @Nonnull
     public Table(@Nonnull final String namespace,
@@ -67,7 +71,8 @@ public class Table<K extends Message, V extends Message, M extends Message> {
                  @Nonnull final CorfuRuntime corfuRuntime,
                  @Nonnull final ISerializer serializer,
                  @Nonnull final Supplier<StreamingMap<K, V>> streamingMapSupplier,
-                 @NonNull final ICorfuVersionPolicy.VersionPolicy versionPolicy) {
+                 @NonNull final ICorfuVersionPolicy.VersionPolicy versionPolicy,
+                 @Nullable final UUID[] streamTags) {
 
         this.corfuRuntime = corfuRuntime;
         this.namespace = namespace;
@@ -79,12 +84,15 @@ public class Table<K extends Message, V extends Message, M extends Message> {
                         .build())
                 .orElse(MetadataOptions.builder().build());
 
-        this.corfuTable = corfuRuntime.getObjectsView().build()
+        SMRObject.Builder<CorfuTable<K, CorfuRecord<V, M>>> builder = corfuRuntime.getObjectsView().build()
                 .setTypeToken(CorfuTable.<K, CorfuRecord<V, M>>getTableType())
                 .setStreamName(this.fullyQualifiedTableName)
                 .setSerializer(serializer)
-                .setArguments(new ProtobufIndexer(valueSchema), streamingMapSupplier, versionPolicy)
-                .open();
+                .setArguments(new ProtobufIndexer(valueSchema), streamingMapSupplier, versionPolicy);
+        if (streamTags != null) {
+            builder.setStreamTags(streamTags);
+        }
+        this.corfuTable = builder.open();
     }
 
     /**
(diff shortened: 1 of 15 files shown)


runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -239,9 +244,19 @@ public long getConflictSetAndCommit(ConflictSetInfo conflictSet) {
...
+        Set<UUID> streamTags = new HashSet<>();
+        affectedStreamsIds.forEach( s ->
+        {
+            if (streamToProxyMap.containsKey(s)) {
+                streamTags.addAll(streamToProxyMap.get(s).getStreamTags());
+            }
+        });
+        affectedStreamsIds.addAll(streamTags);
+
+        // Write to the transaction stream if transaction logging is enabled
```
Maithem: You can do something like this

```java
        Set<UUID> affectedStreamsIds = new HashSet<>(getWriteSetInfo().getWriteSet().getEntryMap().keySet());
        modifiedProxies.stream().forEach(proxy -> affectedStreamsIds.add(proxy.getStreamTags()));
```

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -239,9 +244,19 @@ public long getConflictSetAndCommit(ConflictSetInfo conflictSet) {
...
         if (this.transaction.isLoggingEnabled()) {
             affectedStreamsIds.add(TRANSACTION_STREAM_ID);
         }
```
Maithem: You can remove this TRANSACTION_STREAM_ID and use the new method to add a transaction stream

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
pankti-m: done

xiaoqin2012:
Can you write some description about why we need stream tag to give some context.

pankti-m:
> Can you write some description about why we need stream tag to give some context.



> Can you write some description about why we need stream tag to give some context.

Done

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -180,6 +179,7 @@
                 this, proxy, updateEntry.getSMRMethod(),
                 updateEntry.getSMRArguments(), conflictObjects);
 
+        modifiedProxies.add(proxy);
         return addToWriteSet(proxy, updateEntry, conflictObjects);
```
WenbinZhu: If you look at the implementation of `addToWriteSet`, it does a thread synchronization on root context:
```java
synchronized (getRootContext().getTransactionID()) {
        ....
}
```
Which makes me wonder if `modifiedProxies.add(proxy);` is sufficient. Also if you look at how we deal with nested transactions: 
```java
public void addTransaction(AbstractTransactionalContext tc) {
        ......
        mergeWriteSetInto(tc.getWriteSetInfo());
        ......
}
```
It merges writeSet, but the modifiedProxies does not do that. It might lose the tagging information. So maybe we should figure out a way to reuse `writeSet` for tagging or solve the above two issues.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -239,12 +239,13 @@ public long getConflictSetAndCommit(ConflictSetInfo conflictSet) {
...
-        // Write to the transaction stream if transaction logging is enabled
-        Set<UUID> affectedStreamsIds = new HashSet<>(getWriteSetInfo().getWriteSet().getEntryMap().keySet());
+        Set<UUID> affectedStreamsIds = new HashSet<>(getWriteSetInfo()
+            .getWriteSet().getEntryMap().keySet());
 
-        if (this.transaction.isLoggingEnabled()) {
-            affectedStreamsIds.add(TRANSACTION_STREAM_ID);
-        }
+        modifiedProxies.forEach( p ->
+        {
```
WenbinZhu: NIT: move this curly bracket to the above line.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -66,7 +67,8 @@ public Table(@Nonnull final String namespace,
                  @Nonnull final CorfuRuntime corfuRuntime,
                  @Nonnull final ISerializer serializer,
                  @Nonnull final Supplier<StreamingMap<K, V>> streamingMapSupplier,
-                 @NonNull final ICorfuVersionPolicy.VersionPolicy versionPolicy) {
+                 @NonNull final ICorfuVersionPolicy.VersionPolicy versionPolicy,
+                 final UUID[] streamTags) {
```
hisundar, line 71: Why does stream tag have to be 16 full bytes per tag? Seems super expensive given that the update itself might be just 4-8 bytes. Some table updates are pretty small but will need to be tagged in different stream tags. So say a 8 bytes of user data needs to be tagged in 3 stream tags, then user data will be 8 bytes but the overhead of just the tagging alone will be 48 bytes an amplification of 6 X !!

hisundar: sorry I realized that these are stream IDs, the only suggestion would be to use strings instead of UUIDs since verticals who consume deal with strings much more readily. thanks

test/src/test/java/org/corfudb/runtime/object/transactions/OptimisticTransactionContextTest.java
```diff
@@ -895,4 +909,89 @@ public void txnCanCommitIfOverwrittenAndRetry() throws Exception {
...
+            p.getStreamTags().contains(CorfuRuntime.getStreamID("test-1"))));
+        });
+
+        t2(() -> {
+            OptimisticTransactionalContext context =
+                (OptimisticTransactionalContext) TransactionalContext
+                    .getCurrentContext();
+            assertThat(context.getModifiedProxies().isEmpty());
+            assertThat(context.getWriteSetInfo().getAffectedStreams().contains(
+                CorfuRuntime.getStreamID("test-2")));
```
zhangn49, line 953: assertThat(condition) does not guarantee condition is true. It needs a terminal operation like `assertThat(condition).isTrue()` or `assertThat(condition).isZero()`

Issue #2409, opened 2020-02-14
systemDownHandler blocks runtime trying in vain to reconnect to WrongCluster


hisundar:
## Overview

CorfuRuntime will throw a WrongClusterException if it is trying to connect to the cluster with a wrong clusterId. Normally this happens when there is a reconfiguration event and the old cluster is no longer valid. In such circumstances, the systemDownHandler should quickly give up instead of trying to reconnect to the wrong cluster otherwise this will hold up the process and prevent it from restarting.

Maithem:
WrongClusterException is a non-retriable error it should be propagated.

PR #2418, opened 2020-02-19
Runtime: exit systemDownHandler on WrongClusterException

hisundar:
## Overview

Description:
Unblock systemDownHandler if WrongClusterException 

Why should this be merged: 
If a node has been added to a cluster the existing
clients should exit early so they can be restarted.
Currently they just spin in systemDownHandler.

Related issue(s) (if applicable): #2409 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -982,6 +982,10 @@ public class CorfuRuntime {
                     } catch (InterruptedException ie) {
                         throw new UnrecoverableCorfuInterruptedError(
                                 "Interrupted during layout fetch", ie);
+                    } catch (WrongClusterException we) {
+                        // It is futile trying to re-connect to the wrong cluster
+                        log.warn("Giving up since cluster is incorrect or reconfigured!");
+                        throw we;
                     } catch (ExecutionException ee){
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -982,6 +982,9 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
                     } catch (InterruptedException ie) {
                         throw new UnrecoverableCorfuInterruptedError(
                                 "Interrupted during layout fetch", ie);
+                    } catch (WrongClusterException we) {
+                        // It is futile trying to re-connect to the wrong cluster
+                        throw new UnrecoverableCorfuError("WrongClusterException:", we);
```
Maithem: Doesn't `AbstractView` have the same issue ?

hisundar:
> Its cleaner to rethrow the exception instead of converting it o an Error. Clients can handle the exception, so I don't think UnrecoverableCorfuError applies.

Done.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
hisundar: thanks fixed.

PR #2452, opened 2020-02-28
Thread Clean Up

Maithem:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
@@ -1,6 +1,7 @@
 package org.corfudb.runtime.clients;
 
 import com.codahale.metrics.Timer;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -672,11 +680,11 @@ public NettyClientRouter(String host, Integer port, Boolean tls,
...
         @Nonnull CorfuRuntimeParameters parameters) {
         this(node, parameters.getSocketType()
-            .getGenerator().generate(Runtime.getRuntime().availableProcessors() * 2,
+            .getGenerator().generate(parameters.getNettyEventLoopThreads(),
                 new ThreadFactoryBuilder()
                     .setDaemon(true)
                     .setNameFormat(parameters.getNettyEventLoopThreadFormat())
-                    .build()), parameters);
+                    .build()), parameters, true);
     }
```
WenbinZhu: Can we remove the `manageEventLoop` flag, and set the `eventLoopGroup` field in this contructor only? Then in close() we can simply check if `eventLoopGroup` is null and close it it not null.

Maithem: Discussed offline. This change is less invasive.

Issue #2454, opened 2020-03-02
CorfuQueue: ordering is lost if Checkpointer runs


hisundar:
CorfuQueue preserves order of the entries by using a LinkedHashMap.
However when checkpointer runs, it uses a regular HashMap and rewrites entries away from the original order.
Fix is to sort the entries back in order before returning them in `CorfuQueue`'s `entryList()` method.

PR #2455, opened 2020-03-02
CorfuQueue: Sort entries before returning in entryList()

hisundar:
This is to preserve order if checkpointer runs first.
Checkpointer uses a normal HashMap and not a LinkedHashMap
causing a loss of ordering if it runs first.
Fix is to sort the entries first.
This also enables the api to allow entries to be returned
only after a user specified previous entry.

## Overview

Description:

Why should this be merged: 

Related issue(s) (if applicable): #2454 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
@@ -261,10 +261,9 @@ public class CorfuQueue<E> {
         );
 
         int index = 0;
-        for (Long entryId : corfuTable.keySet().stream().sorted().collect(Collectors.toList())) {
-            if (entryId <= entriesAfter) {
-                continue;
-            }
+        for (Long entryId : corfuTable.keySet().stream()
+                .filter(e -> e > entriesAfter)
+                .sorted().collect(Collectors.toList())) {
             if (++index >= maxEntries) {
                 break;
             }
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -257,16 +259,19 @@ public int compareTo(CorfuQueueRecord<? extends E> o) {
         List<CorfuQueueRecord<E>> copy = new ArrayList<>(
                 Math.min(corfuTable.size(), maxEntries)
         );
+
         int index = 0;
-        for (Map.Entry<Long, E> entry : corfuTable.entrySet()) {
+        for (Long entryId : corfuTable.keySet().stream().sorted().collect(Collectors.toList())) {
```
zhangn49: Should we filter keys out before collecting to a list?

zhangn49: For maxEntries, maybe we can use Stream.limit()

hisundar: I thought about this too, but if we specify stream.limit() then it is possible that the stream has only a small number of entries and if the user has specified a start point towards the end, then it will miss out on the entries.
here is the example,
Q: 1, 2, 3, 4, 5
User wants 2 entries AFTER item number 3:
We should return: 4, 5
With Stream.limit() our collect() will have only 2 entries : 1, 2
which means 4, 5 cannot be processed.
Hope this helps?

hisundar: Fixed, I have used the stream.filter as suggested. thanks!

PR #2460, opened 2020-03-04
Use Writer Thread For Eviction

Maithem:
## Overview
Allow the caffeine cache to use the logunit thread pool 

Why should this be merged: Reduces the number of threads

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
@@ -1,5 +1,7 @@
 package org.corfudb.infrastructure;
 
+import static org.corfudb.util.MetricsUtils.sizeOf;
+
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.RemovalCause;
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
```diff
@@ -24,11 +26,12 @@
     private final LoadingCache<Long, ILogData> dataCache;
     private final StreamLog streamLog;
 
-    public LogUnitServerCache(LogUnitServerConfig config, StreamLog streamLog) {
+    public LogUnitServerCache(LogUnitServerConfig config, StreamLog streamLog, ExecutorService executorService) {
```
pankti-m: you can keep this of type Executor(more generic) instead of ExecutorService

Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
```diff
@@ -32,11 +34,12 @@
     //Empirical threshold of number of streams in a logdata beyond which server performance may be slow
     private final int MAX_STREAM_THRESHOLD = 20;
 
-    public LogUnitServerCache(LogUnitServerConfig config, StreamLog streamLog) {
+    public LogUnitServerCache(LogUnitServerConfig config, StreamLog streamLog, Executor executor) {
         this.streamLog = streamLog;
         this.dataCache = Caffeine.newBuilder()
                 .<Long, ILogData>weigher((addr, logData) -> getLogDataTotalSize(logData))
                 .maximumWeight(config.getMaxCacheSize())
+                .executor(executor)
```
Maithem, line 42: @ben-manes can a self-Induced deadlock happen if the writers and eviction threads use the same executor?

ben-manes: That should be fine. Some turn off the executor and use `Runnable::run` so that the work is amortized on the callers. That's fast on our side, but penalizes the caller when invoking your removal listener, etc. We defer to the common threadpool because it is available and lets us avoid penalizing the caller if your callbacks are expensive, but it isn't necessary nor should you observe deadocks based on how it is configured.

Maithem: K, thanks.

achinmishra:
Looks good. The conversation on a possible deadlock is already taken care off. Cant find "upvote" or "approved" button here.

Maithem:
IF seems to be stuck. Cloned it to #2842

PR #2473, opened 2020-03-12
Get rid of ServerHandshakeHandler log messages

xnull:
## Overview

Description:
ServerHandshakeHandler spams the log with current status (it is too verbose).
I left only 'error' and important messages. All the others don't seem useful.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
@@ -78,8 +78,6 @@ public class ServerHandshakeHandler extends ChannelDuplexHandler {
 
         try {
             handshake = (CorfuPayloadMsg<HandshakeMsg>) m;
-            log.debug("channelRead: Handshake Message received. Removing {} from pipeline.",
-                    READ_TIMEOUT_HANDLER);
             // Remove the handler from the pipeline. Also remove the reference of the context from
             // the handler so that it does not disconnect the channel.
             ctx.pipeline().remove(READ_TIMEOUT_HANDLER).handlerRemoved(ctx);
@@ -104,11 +102,7 @@ public class ServerHandshakeHandler extends ChannelDuplexHandler {
         UUID clientId = handshake.getPayload().getClientId();
         UUID serverId = handshake.getPayload().getServerId();
 
-        // Validate handshake, but first verify if node identifier is set to default (all 0's)
-        // which indicates node id matching is not required.
-        if (serverId.equals(UUID.fromString("00000000-0000-0000-0000-000000000000"))) {
-            log.info("channelRead: node id matching is not requested by client.");
-        } else if (!serverId.equals(this.nodeId)) {
+        if (!serverId.equals(this.nodeId)) {
             log.error("channelRead: Invalid handshake: this is {} and client is trying to connect to {}",
                     this.nodeId, serverId);
             this.fireHandshakeFailed(ctx);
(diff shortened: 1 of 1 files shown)


annym:
I have blocked this PR from the Injection Framework as it is failing and got stuck. Please discuss it with @PavelZaytsev he knows why he needs the comparison with a NULL UUID. Let me know when it is fixed I can re-enable IF for this PR.

xnull:
@annym I rolled back clientId changes, can you please active injection-framework again, please?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -117,22 +115,15 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
 
         // Store clientID as a channel attribute.
         ctx.channel().attr(clientIdAttrKey).set(clientId);
-        log.info("channelRead: Handshake validated by Server.");
-        log.debug("channelRead: Sending handshake response: Node Id: {} Corfu Version: {}",
```
xiaoqin2012, line 121: Should we keep the debug log?

annym: +1 this Corfu Version allows to catch when there is a Corfu mismatch

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -145,7 +136,7 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
      */
     @Override
     public void channelActive(final ChannelHandlerContext ctx) throws Exception {
-        log.info("channelActive: Incoming connection established from: {} Start Read Timeout.",
+        log.trace("channelActive: Incoming connection established from: {} Start Read Timeout.",
```
annym, line 139: This log statement I believe is also relevant when we're not sure if the connection from one client to a server is established. It is useful, I think we should leave it as INFO or DEBUG.

WenbinZhu: I think one log line per connection establish is not too many, this tells us when the connection begins, and help to trace the life cycle of a connection when debugging.

hisundar: +1 we have used this message to debug client connection issues.

runtime/src/main/java/org/corfudb/security/tls/SslContextConstructor.java
```diff
@@ -33,11 +33,11 @@ public static SslContext constructSslContext(boolean isServer,
                                                  String ksPasswordFile,
                                                  @NonNull String trustStorePath,
                                                  String tsPasswordFile) throws SSLException {
-        log.info("Construct ssl context based on the following information:");
-        log.info("Key store file path: {}.", keyStorePath);
-        log.info("Key store password file path: {}.", ksPasswordFile);
-        log.info("Trust store file path: {}.", trustStorePath);
-        log.info("Trust store password file path: {}.", tsPasswordFile);
+        log.trace("Construct ssl context based on the following information:");
```
hisundar, line 36: The presence of this message helped distinguish a firewall misconfiguration from a connectivity issue.
What had happened is that the Corfu Server port was blocked but SSL which uses 80/443 went through fine. So wondering if this message can still remain?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
xnull: Ok, if a majority of people believe we need let's change it back

runtime/src/main/java/org/corfudb/security/tls/SslContextConstructor.java
xnull: Sundar, these messages in the log are literally a flood. Even if it helped to find a bug once I don't believe that the price we are paying for that is worth it.
These messages we print so often that every single time when you read logs you read these messages and it gets pretty hard to find all other messages in the log.

xnull:
Fixed

xnull:
Also, for the info, here are the top 20 messages in the logs in Corfu:

image

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
apavan: "If there is a mismatch let's add more check to prevent inconsistency rather than printing an endless number of messages" - i can't agree more with you Slava but at the same time we cannot lose debug ability of the product without a work around.  Please do not merge this change without the work around that you have suggested.

runtime/src/main/java/org/corfudb/security/tls/SslContextConstructor.java
apavan: Slava, your arguments are very valid and I like the thought process. Please have a work around for these issues as we just cannot afford to lose out on debug ability of the product.

PR #2477, opened 2020-03-12
LayoutView don't print status messages

xnull:
## Overview

Description:
LayotView class is the biggest contributor to heavy logging. 
It prints status messages hundreds of thousand times.

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
@@ -245,10 +245,6 @@ public class LayoutView extends AbstractView {
                     .filter(x -> x)
                     .count();
 
-            log.debug("propose: Successful responses={}, needed={}, timeouts={}, "
-                            + "wrongEpochRejected={}",
-                    count, getQuorumNumber(), timeouts, wrongEpochRejected);
-
             if (count >= getQuorumNumber()) {
                 break;
             }


zhangn49:
The reason why LayoutView prints thousands of times is current while loop keeps running.
The issue is here. #2312 
Can we change it to trace for now rather than delete it?

xnull:
trace also ok, but what's the point for that? Those messages don't provide any valuable information

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -164,10 +164,6 @@ public Layout prepare(long epoch, long rank)
                     .filter(x -> x != null)
                     .toArray(LayoutPrepareResponse[]::new);
 
-            log.debug("prepare: Successful responses={}, needed={}, timeouts={}, "
-                            + "wrongEpochRejected={}",
-                    acceptList.length, getQuorumNumber(), timeouts, wrongEpochRejected);
-
```
Maithem: Instead of logging this on each iteration, or removing the logs, you can just move it outside the loop. These logs show how long each phase takes which is useful when debugging.

xnull: 1. I don't know anyone who used it at least ones when debugging
2. We have all necessary logs on the server-side
Mostly those messages are minor. 

Anyway, we can move it. Will do

xnull: Again these logs printing tens of thousands of times to the log ( for the last 3 years) and no one paid attention and no one fixed it, which means the messages were  always ignored when debugging

zhangn49: @xnull Actually I am trying to fix it if you remember... and I did meet a problem that cluster kept bumping epoch from 0 to 300+ in a short time. For that issue, I don't know what happened because those counters are inaccurate.

PR #2478, opened 2020-03-12
Propagate exceptions when bootstrapping a management server during the AddNode workflow.

PavelZaytsev:
## Overview

Description:

Why should this be merged: 

There is a join performed on the bootstrapManagementServer in the AddNode SendRequest method. If it fails, it throws the exception wrapped in the CompletionException, which does not get handled on the upper layers. Even if it is not wrapped in this exception, the root exception is still wrapped in the RuntimeException, so it still wont get caught. The goal is to propagate TimeOut and Network exceptions from this call, so it can get retried. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
+++ b/runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
@@ -19,6 +19,7 @@ import org.corfudb.util.Sleep;
 
 import java.time.Duration;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
 
 /**
  * Utility to bootstrap a cluster.
@@ -54,18 +55,22 @@ public class BootstrapUtil {
      * @param layout Layout to bootstrap with
      */
     private static void bootstrapLayoutServer(IClientRouter router, Layout layout)
-            throws ExecutionException, InterruptedException, AlreadyBootstrappedException {
+            throws ExecutionException, InterruptedException, AlreadyBootstrappedException, TimeoutException {
         LayoutClient layoutClient = new LayoutClient(router, layout.getEpoch());
 
         try {
             CFUtils.getUninterruptibly(layoutClient.bootstrapLayout(layout),
-                    AlreadyBootstrappedException.class);
+                    AlreadyBootstrappedException.class, TimeoutException.class);
         } catch (AlreadyBootstrappedException abe) {
             if (!layoutClient.getLayout().get().equals(layout)) {
                 log.error("BootstrapUtil: Layout Server {}:{} already bootstrapped with different "
                         + "layout.", router.getHost(), router.getPort());
                 throw abe;
             }
+        } catch (TimeoutException toe){
+            log.error("BootstrapUtil: Layout Server {}:{} timeout occurred.",
+                    router.getHost(), router.getPort());
+            throw toe;
         }
     }
 
@@ -76,19 +81,23 @@ public class BootstrapUtil {
      * @param layout Layout to bootstrap with
      */
     private static void bootstrapManagementServer(IClientRouter router, Layout layout)
-            throws ExecutionException, InterruptedException, AlreadyBootstrappedException {
+            throws ExecutionException, InterruptedException, AlreadyBootstrappedException, TimeoutException {
         ManagementClient managementClient
                 = new ManagementClient(router, layout.getEpoch());
 
         try {
             CFUtils.getUninterruptibly(managementClient.bootstrapManagement(layout),
-                    AlreadyBootstrappedException.class);
+                    AlreadyBootstrappedException.class, TimeoutException.class);
         } catch (AlreadyBootstrappedException abe) {
             if (!managementClient.getLayout().get().equals(layout)) {
                 log.error("BootstrapUtil: Management Server {}:{} already bootstrapped with "
                         + "different layout.", router.getHost(), router.getPort());
                 throw abe;
             }
+        } catch (TimeoutException toe) {
+            log.error("BootstrapUtil: Management Server {}:{} timeout occurred.",
+                    router.getHost(), router.getPort());
+            throw toe;
         }
     }
 
@@ -122,9 +131,11 @@ public class BootstrapUtil {
                         bootstrapManagementServer(router, layout);
                         break;
                     } catch (AlreadyBootstrappedException abe) {
-                        log.error("Bootstrapping node: {} failed with exception:", server, abe);
                         log.error("Cannot retry since already bootstrapped.");
                         throw new RuntimeException(abe);
+                    } catch (TimeoutException toe) {
+                        log.error("Timeout exception occurred while bootstrapping. " +
+                                        "Retrying: {} times.", retry);
                     } catch (Exception e) {
                         log.error("Bootstrapping node: {} failed with exception:", server, e);
                         if (retry == 0) {
--- a/runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
+++ b/runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
@@ -19,7 +19,6 @@ import org.corfudb.util.Sleep;
 
 import java.time.Duration;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeoutException;
 
 /**
  * Utility to bootstrap a cluster.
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -99,17 +100,26 @@ public void handleFailure(IReconfigurationHandlerPolicy failureHandlerPolicy,
...
+
+        try {
+            boolean bootstrapped = runtime.getLayoutView().bootstrapLayoutServer(endpoint, layout);
+            future.complete(bootstrapped);
+            return future;
+        }
+        catch (TimeoutException toe){
+            future.completeExceptionally(toe);
+            return future;
+        }
```
WenbinZhu: This seems hacky, and besides why this is needed?  `bootstrapLayoutServer` has an exceptionally block:

```java
.exceptionally(throwable -> {
        try {
                CFUtils.unwrap(throwable, AlreadyBootstrappedException.class);
        } catch (AlreadyBootstrappedException e) {
                log.info("bootstrapLayoutServer: Layout Server {} already bootstrapped.", endpoint);
        }
        return true;
})
```
Notice this `CFUtils.unwrap(throwable, AlreadyBootstrappedException.class);` It will throw a `RutimeException` in case of exceptions other than `AlreadyBootstrappedException`, which makes the returned future exceptionally completed . Thus in this `bootstrapNewNode` method you will get a exceptionally completed future of `RutimeException` with a `TimeoutException` wrapped inside.

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
```diff
@@ -46,7 +46,12 @@ public AddNode(@NonNull String endpointToAdd, @NonNull CorfuRuntime runtime,
     protected UUID sendRequest(@NonNull ManagementClient managementClient) throws TimeoutException {
         // Bootstrap a management server first.
         Layout layout = new Layout(runtime.getLayoutView().getLayout());
-        runtime.getManagementView().bootstrapManagementServer(nodeForWorkflow, layout).join();
+        boolean bootstrapped =
+                runtime.getManagementView().bootstrapManagementServer(nodeForWorkflow, layout);
```
pankti-m: shouldnt timeoutException thrown from bootstrapManagementServer be caught here?

runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java
```diff
@@ -54,18 +55,22 @@ public static void bootstrap(@NonNull Layout layout,
...
         } catch (AlreadyBootstrappedException abe) {
             if (!layoutClient.getLayout().get().equals(layout)) {
                 log.error("BootstrapUtil: Layout Server {}:{} already bootstrapped with different "
                         + "layout.", router.getHost(), router.getPort());
                 throw abe;
             }
+        } catch (TimeoutException toe){
+            log.error("BootstrapUtil: Layout Server {}:{} timeout occurred.",
+                    router.getHost(), router.getPort());
+            throw toe;
```
WenbinZhu: I don't understand why we need to explicitly handle TimeoutException here. What about NetworkExceptions and other exceptions? Besides, the caller of this, namely BootstrapUtil.bootstrap() already logs and handles all exceptions: https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/BootstrapUtil.java#L128-L134

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
```diff
@@ -1640,7 +1640,7 @@ public void testPartialBootstrapNodeSuccess() throws ExecutionException, Interru
         corfuRuntime.getLayoutView().getRuntimeLayout().getLayoutClient(SERVERS.ENDPOINT_1).bootstrapLayout(layout);
         // Attempt bootstrapping the node. The node should attempt bootstrapping both the components Layout Server and
         // Management Server.
-        assertThat(corfuRuntime.getLayoutManagementView().bootstrapNewNode(SERVERS.ENDPOINT_1).get()).isTrue();
+        assertThat(corfuRuntime.getLayoutManagementView().bootstrapNewNode(SERVERS.ENDPOINT_1).get()).isFalse();
```
pankti-m: you can change the test comments to reflect this new behaviour

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
```diff
@@ -319,21 +319,26 @@ public void committed(long epoch, Layout layout, boolean force)
...
+            log.info("bootstrapLayoutServer: Layout Server {} bootstrap successful.",
+                    endpoint);
+            return true;
+        } catch (AlreadyBootstrappedException abe) {
+            log.warn("bootstrapLayoutServer: Layout Server {} already bootstrapped.",
+                    endpoint);
+            return false;
+        } catch (TimeoutException te) {
+            log.error("bootstrapLayoutServer: Timeout occurred.");
+            throw te;
```
WenbinZhu: Why specifically handling TimeoutException, there are other exceptions like NetworkException right? As explained in previous comment, the existing bootstrapLayoutServer implementation does return an exceptionally completed future due to CFUtil.unwrap(), I think the problem might not be here.

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
```diff
@@ -46,7 +46,12 @@ public AddNode(@NonNull String endpointToAdd, @NonNull CorfuRuntime runtime,
...
+
+        if (!bootstrapped) {
```
pankti-m: minor - the variable name bootstrapped is confusing here.  If the server is bootstrapped, we log the warning.  But the check appears to be doing the opposite.  You can change the name to something else.

annym:
@PavelZaytsev Can you please add as description a clear explanation of the bug/issue? perhaps a clear example of what we have seen eases the review and also for reference in the future is a good idea.

runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java
PavelZaytsev: Ok, I see. I won't touch it then. I'll leave management and layout views as is but will unwrap the exception in the AddNode sendRequest I guess.

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
PavelZaytsev: Outdated

test/src/test/java/org/corfudb/runtime/view/ManagementViewTest.java
PavelZaytsev: Outdated

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
PavelZaytsev: Outdated

PavelZaytsev:
@annym I have updated the description.
@WenbinZhu I have addressed the comments and updated the description.

Please review when you get the chance.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -141,7 +141,7 @@ public void invoke() {
                     throw new IllegalStateException("Orchestrator can not be selected");
                 }
 
-            } catch (NetworkException | TimeoutException | IllegalStateException e) {
+            } catch (Exception e) {
```
Maithem: I don't think swallowing all exceptions is a good idea. The retry only needs to happen on timeout and network exceptions.

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
```diff
@@ -46,7 +47,7 @@ public AddNode(@NonNull String endpointToAdd, @NonNull CorfuRuntime runtime,
     protected UUID sendRequest(@NonNull ManagementClient managementClient) throws TimeoutException {
         // Bootstrap a management server first.
         Layout layout = new Layout(runtime.getLayoutView().getLayout());
-        runtime.getManagementView().bootstrapManagementServer(nodeForWorkflow, layout).join();
+        CFUtils.getUninterruptibly(runtime.getManagementView().bootstrapManagementServer(nodeForWorkflow, layout));
```
Maithem: You can pass Network/Timeout exceptions types.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
Maithem: Also, if this thread is interrupted, the interrupt flag isn't restored.

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
PavelZaytsev: Ok, I see. I'll pass the Network and TimeOut types here and return the 3 exceptions.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
PavelZaytsev: Ok, I'll bring it back in.

PavelZaytsev:
@annym 
@Maithem 
@pankti-m 
@WenbinZhu 

I have addressed the comments and updated the description.
Please review when you get the chance.

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
```diff
@@ -44,9 +46,12 @@ public AddNode(@NonNull String endpointToAdd, @NonNull CorfuRuntime runtime,
 
     @Override
     protected UUID sendRequest(@NonNull ManagementClient managementClient) throws TimeoutException {
```
WenbinZhu: I think later on there should a cleanup patch on all the inherited children of sendRequest because others are not taking care of NetworkException, and the signature suggests it only throws TimeoutException.

WenbinZhu:
@PavelZaytsev please rebase the commits.

runtime/src/main/java/org/corfudb/runtime/view/workflows/AddNode.java
PavelZaytsev: Sounds good.

PR #2480, opened 2020-03-14
Fix LogData lastKnownSize which is inaccurate and inconsistent.

WenbinZhu:
## Overview

- When set during serialization and deserialization, the LogData
lastKnownSize is set to different values. It is also inaccurate
as it is set to ByteBuf's internal array length during serialization,
which is usually larger the actually (it is dynamically resized and
usually 2x of previous size). It needs to be set to the payload's
serialized size. This incorrect behavior affects transaction size
limit calculation, which results in unnecessary WriteSizeException.

- Also fixed a test where compression should be ruled out.

Related issue(s) (if applicable): #2235 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -129,7 +129,6 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
         if (serializedCache == null) {
             serializedCache = Unpooled.buffer();
             doSerializeInternal(serializedCache);
-            lastKnownSize = serializedCache.array().length;
         } else {
             serializedCache.retain();
         }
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -129,7 +129,6 @@ public synchronized void acquireBuffer() {
         if (serializedCache == null) {
             serializedCache = Unpooled.buffer();
             doSerializeInternal(serializedCache);
-            lastKnownSize = serializedCache.array().length;
```
Maithem: can't you just set lastKnownSize = buf.readableBytes() on acquireBuffer  instead of setting it twice in doSerializeInternal ?

WenbinZhu: That's what I initially did, but later I realized that that way the lastKnownSize will include the metadata map, but when deserializing, it does include metadata.

xnull:
Can you please write a test (would be great to have at least one unit and one integration test for that)?
The reason why we need it is that the issue was there for the last 3 years and no one paid attention to that because thousands of those log messages printed in the log for years, were hidden by thousands of other messages.

I found that issue just because I was going through all the logs and was filtering a big amount of other logs.

If we wouldn't be very careful here with the fix we will end with the same situation we had for the last 3 years.

Issue #2470, opened 2020-03-11
Inconsistent read due to index update before batch fsync()


WenbinZhu:
## Overview

The current batchProcessor on log unit server performs a batch of write() and then calls fsync(), however because the segment index (knownAddress) is updated immediately after a successful write() (before fsync()), any read() before fsync() can still retrieve the update. This could cause inconsistent read in the following case:

1. write(100L, x)
2. read(100L) -> x because index is updated, so the write becomes visible
3. process restarts without performing fync() 
4. read(100L) -> null because write(x) is not fully persisted before restart

The consequence is two reads on the same address are inconsistent. 
This is not the only case, there are other cases due to this issue.

PR #2482, opened 2020-03-18
Fix potential inconsistent read due to index update before fsync().

WenbinZhu:
## Overview

Description:

There is a potential inconsistent read issue. The current batchProcessor on log unit server performs a batch of write() and then calls fsync(), however because the segment index (knownAddress) is updated immediately after a successful write(), but before fsync(), any read() before fsync() can still retrieve the update. This could cause inconsistent read. The first read could see the data, while the second read, after process restarts, cannot see the data because data is not guaranteed to be persisted in secondary storage without fsync(). 

Related issue(s) (if applicable): #2470 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
@@ -134,8 +134,7 @@ public class InMemoryStreamLog implements StreamLog, StreamLogWithRankedAddressS
      * @return Set of known addresses.
      */
     @Override
-    public Set<Long> getKnownAddressesInRange(long rangeStart, long rangeEnd) {
-
+    public Set<Long> getWrittenAddressesInRange(long rangeStart, long rangeEnd) {
         Set<Long> result = new HashSet<>();
         for (long address = rangeStart; address <= rangeEnd; address++) {
             if (logCache.containsKey(address)) {
@@ -155,8 +154,8 @@ public class InMemoryStreamLog implements StreamLog, StreamLogWithRankedAddressS
     }
 
     @Override
-    public void sync(boolean force){
-        //no-op
+    public void sync(boolean force) {
+        // No-op
     }
 
     @Override
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
@@ -1,18 +1,20 @@
 package org.corfudb.infrastructure.log;
 
-import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.IOUtils;
 
 import java.io.IOException;
 import java.nio.channels.FileChannel;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * The global log is partition into segments, each segment contains a range of consecutive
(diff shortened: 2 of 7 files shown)


WenbinZhu:
A new test will be added.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
```diff
@@ -1,18 +1,21 @@
 package org.corfudb.infrastructure.log;
 
 import lombok.Data;
```
zhangn49, line 3: Unused right now.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
```diff
@@ -21,36 +24,97 @@
...
 
+    // File channel for log writes.
     @NonNull
-    final FileChannel readChannel;
+    final FileChannel writeChannel;
 
+    // File channel for reads.
     @NonNull
-    String fileName;
+    final FileChannel readChannel;
```
zhangn49, line 47: All could be private

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
```diff
@@ -21,36 +24,97 @@
...
     private final Map<Long, AddressMetaData> knownAddresses = new ConcurrentHashMap<>();
-    private final Set<Long> trimmedAddresses = Collections.newSetFromMap(new ConcurrentHashMap<>());
-    private final Set<Long> pendingTrims = Collections.newSetFromMap(new ConcurrentHashMap<>());
-    private volatile int refCount = 0;
 
+    // Index of log entries that are written but not guaranteed to be persisted.
+    private final Map<Long, AddressMetaData> pendingAddresses = new ConcurrentHashMap<>();
+
+    // Reference count of this segment.
+    private AtomicInteger refCount = new AtomicInteger(0);
```
zhangn49, line 56: Could be final

infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
```diff
@@ -155,8 +154,8 @@ public LogData read(long address) {
     }
 
     @Override
-    public void sync(boolean force){
-        //no-op
+    public void sync(boolean metadata){
```
zhangn49, line 157: why call metadata here?

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
WenbinZhu: Carried from an intermidate change, reverted.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
WenbinZhu: Done.

WenbinZhu:
This fix is probably incorrect under current hole fill impl.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -376,13 +375,15 @@ private void verifyLogs() {
 
     @Override
     public void sync(boolean force) throws IOException {
-        if (force) {
-            for (FileChannel ch : channelsToSync) {
-                ch.force(true);
+        for (SegmentHandle sh : segmentsToSync) {
```
xnull, line 378: if `force` is false no need to go through the collection. So prev version is better

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -757,15 +758,15 @@ SegmentHandle getSegmentHandleForAddress(long address) {
         filePath += segment;
         filePath += ".log";
 
-        SegmentHandle handle = writeChannels.computeIfAbsent(filePath, a -> {
+        SegmentHandle handle = openSegments.computeIfAbsent(filePath, a -> {
```
xnull, line 761: please rename single letter variable

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
```diff
@@ -736,4 +744,34 @@ public void estimateSizeTest() throws IOException {
         assertThat(parentSize).isEqualTo(parentDirFilePayloadSize + childDirFilePayloadSize);
         assertThat(childDirSize).isEqualTo(childDirFilePayloadSize);
     }
+
+    @Test
+    public void readConsistencyTest() throws Exception {
```
xnull, line 749: Even if `fsync` is not executed yet it doesn't mean that data is corrupted. 
In very rare cases data can get corrupted on disk and we can't do anything with that, there is no point to bring additional latency and wait until fsync happens - it's a pretty expensive operation.
Also, these changes can affect reads, which can cause unexpected behavior.

WenbinZhu:
> Even if fsync is not executed yet it doesn't mean that data is corrupted.
> In very rare cases data can get corrupted on disk and we can't do anything with that, there is no point to bring additional latency and wait until fsync happens - it's a pretty expensive operation.
> Also, these changes can affect reads, which can cause unexpected behavior.

That is not true, this is an inconsistency issue that write succeeds, but reader reads nothing if the server restarts quickly, but we were not detecting it, while we can. For corruption, most of them are detectable and we can throw error if detected, maybe there are cases that we cannnot detect, but the difference is here we can detect it and ought to do it. Also this patch does not introduce additional latency, it does not wait. and that's actually a problem. Anyway this approach is incorrect under current hole fill mechanism as mention in previous comment.

xnull:
> this is an inconsistency issue that write succeeds, but reader reads nothing if the server restarts quickly

It's not possible. Essentially your argument is: if we write to disk without sync, the data could stay in Page cache and when we read next time we will read from disk but data is not there yet. 
That is a false statement, you will try to read from page cache even if data not synced yet and only then from the disk, and it doesn't matter how fast you restart a server. In the other case, it would be a huge flaw in the Linux kernel.

WenbinZhu:
> > this is an inconsistency issue that write succeeds, but reader reads nothing if the server restarts quickly
> 
> It's not possible. Essentially your argument is: if we write to disk without sync, the data could stay in Page cache and when we read next time we will read from disk but data is not there yet.
> That is a false statement, you will try to read from page cache even if data not synced yet and only then from the disk, and it doesn't matter how fast you restart a server. In the other case, it would be a huge flaw in the Linux kernel.

This is not my argument, I suggest you read the PR description one more time, it's about segment index.

PR #2483, opened 2020-03-18
Fix Trimmed Addresses Message

annym:


runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
```diff
@@ -30,7 +30,7 @@ public TrimmedException() {
     }
 
     public TrimmedException(List<Long> trimmed) {
-        super(String.format("Trimmed addresses {}", trimmed));
+        super(String.format("Trimmed addresses %s", trimmed));
```
Maithem: Is it necessary to call the super constructor ?

Since the list is being consumed directly on exception, is the message really necessary?

annym: We want a message to be printed relevant to the trimmed addresses, as this exception is directly caught by upper layers, otherwise, we're subject to them printing this info.

PR #2484, opened 2020-03-19
Layout and ClusterId Validation for the Server Router

PavelZaytsev:
## Overview

Description:
This PR was previously merged but got reverted to introduce other changes.
Why should this be merged: 
This PR introduces clusterId and layout validation on the server router level.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -951,7 +951,7 @@ public class CorfuRuntime {
                         IClientRouter router = getRouter(s);
                         // Try to get a layout.
                         CompletableFuture<Layout> layoutFuture =
-                                new LayoutClient(router, Layout.INVALID_EPOCH).getLayout();
+                                new LayoutClient(router, Layout.INVALID_EPOCH, Layout.INVALID_CLUSTER_ID).getLayout();
                         // Wait for layout
                         Layout l = layoutFuture.get();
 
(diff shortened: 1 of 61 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -887,11 +887,13 @@ private void checkClusterId(@Nonnull Layout layout) {
...
+                log.info("Connected to a new cluster {}", UuidUtils.asBase64(clusterId));
             }
-        } else if (!clusterId.equals(layout.getClusterId())) {
-            // We connected but got a cluster id we didn't expect.
-            throw new WrongClusterException(clusterId, layout.getClusterId());
+            // ClusterId has changed, connect to a new cluster.
+        } else if (!clusterId.equals(layout.getClusterId()) && layout.getClusterId() != null) {
+            log.warn("Connecting to a new cluster {}. Previous cluster was: {}",
+                    UuidUtils.asBase64(layout.getClusterId()), UuidUtils.asBase64(clusterId));
+            clusterId = layout.getClusterId();
```
WenbinZhu: I feel this could have an issue. Think about this case:

1) the client connects to A, B, C
2) C joined new cluster
3) the client runtime is invalidated, and when it tries to fetch layout, the layout servers it tries to connect is still A, B, C becuase this is based on last layout. 
4) C responded with a new layout and new cluster id. The client will take this layout as in the current CorfuRuntime, it only needs to successfuly fetch from one layout server.
5) the client connects to the new cluster, which is not intended.

Can you think about this case and see if it's true?

PavelZaytsev: I'll remove this, as the prod issues were not attributed to this change.

achinmishra: sounds good!

PR #2485, opened 2020-03-19
CorfuStore: Build-to-Build upgrade support

hisundar:
## Overview

Description:
This is to help QE with their build to build upgrades.
Updating the schema on openTable allows them to migrate to
a new protobuf version without having to redeploy their setups.

Disk based tables are too huge to be iterated by conventional methods like `entrySet()`
This is the only way a large table that won't fit in memory can be reasonably iterated.
Switched browser to use this existing api speeds it up several fold.

- [x] Public API has Javadoc

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -1,18 +1,22 @@
 package org.corfudb.browser;
 
+import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuStoreMetadata;
 import org.corfudb.runtime.CorfuStoreMetadata.TableName;
 import org.corfudb.runtime.collections.CorfuRecord;
+import org.corfudb.runtime.collections.CorfuStoreEntry;
 import org.corfudb.runtime.collections.CorfuTable;
 import org.corfudb.runtime.collections.CorfuDynamicKey;
 import org.corfudb.runtime.collections.CorfuDynamicRecord;
@@ -124,12 +128,18 @@ public class CorfuStoreBrowser {
 
         CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(namespace, tablename);
         int size = table.size();
-        for (CorfuDynamicKey entry : table.keySet()) {
-            builder = new StringBuilder("\nKey:\n" + entry.getKey())
-                .append("\nPayload:\n" + table.get(entry).getPayload())
-                .append("\nMetadata:\n" + table.get(entry).getMetadata())
-                .append("\n====================\n");
-            log.info(builder.toString());
+        final int batchSize = 50;
+        Stream<Map.Entry<CorfuDynamicKey, CorfuDynamicRecord>> entryStream = table.entryStream();
+        final Iterable<List<Map.Entry<CorfuDynamicKey, CorfuDynamicRecord>>> partitions =
+                Iterables.partition(entryStream::iterator, batchSize);
+        for (List<Map.Entry<CorfuDynamicKey, CorfuDynamicRecord>> partition : partitions) {
+            for (Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry : partition) {
+                builder = new StringBuilder("\nKey:\n" + entry.getKey())
+                        .append("\nPayload:\n" + entry.getValue().getPayload())
+                        .append("\nMetadata:\n" + entry.getValue().getMetadata())
+                        .append("\n====================\n");
+                log.info(builder.toString());
+            }
         }
         return size;
     }
--- a/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
@@ -17,6 +17,7 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuStoreMetadata.TableDescriptors;
 import org.corfudb.runtime.CorfuStoreMetadata.TableName;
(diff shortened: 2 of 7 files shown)


PavelZaytsev:
lgtm

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -124,12 +128,18 @@ public int printTable(String namespace, String tablename) {
 
         CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(namespace, tablename);
         int size = table.size();
-        for (CorfuDynamicKey entry : table.keySet()) {
-            builder = new StringBuilder("\nKey:\n" + entry.getKey())
-                .append("\nPayload:\n" + table.get(entry).getPayload())
-                .append("\nMetadata:\n" + table.get(entry).getMetadata())
-                .append("\n====================\n");
-            log.info(builder.toString());
+        final int batchSize = 50;
```
xnull: please make batchSize configurable

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
```diff
@@ -264,6 +267,19 @@ public void checkMetadataTransactions() throws Exception {
         assertThatThrownBy(() ->
                 table.update(key1, EventInfo.getDefaultInstance(), wrongRevisionMetadata))
                 .isExactlyInstanceOf(RuntimeException.class);
+
+        // Verify the table is readable using entryStream()
+        final int batchSize = 50;
```
xnull: please make it configurable

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: this is an internal setting that should not be tunable but I will expose it anyway, thanks

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
hisundar: this is a unit test

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: actually just had a conversation with the folks using this, we feel exposing this to the browser can cause a lot more confusion than help. this is really an internal batching for RocksDB partitions.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
```diff
@@ -264,6 +267,19 @@ public void checkMetadataTransactions() throws Exception {
         assertThatThrownBy(() ->
                 table.update(key1, EventInfo.getDefaultInstance(), wrongRevisionMetadata))
                 .isExactlyInstanceOf(RuntimeException.class);
+
```
pankti-m: Can we have a test where the number of entries > batchsize and then verify that the number of partitions is correct and all entries are included in the entryStream api?

hisundar: this api is using the same parameters as checkpoint writer
I think these good suggestions but are tests we should add to the checkpointer writer probably not here right?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -163,10 +165,35 @@ void registerTable(@Nonnull String namespace,
...
+        // Or no modification to the protobuf files.
+        boolean hasSchemaChanged = false;
+        CorfuRecord<TableDescriptors, TableMetadata> oldRecord = this.registryTable.get(tableNameKey);
+        if (oldRecord != null) {
+            if (!oldRecord.getPayload().getFileDescriptorsMap().equals(tableDescriptors.getFileDescriptorsMap())) {
+                hasSchemaChanged = true;
+                int oldTableEntries = tableMap.get(
+                        TableRegistry.getFullyQualifiedTableName(namespace, tableName)
+                ).count();
+                log.error("registerTable: Schema update detected for table "+namespace+" "+ tableName);
```
WenbinZhu: The this be log.warn?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -238,6 +265,7 @@ public static String getFullyQualifiedTableName(String namespace, String tableNa
      */
     private <T extends Message> void addTypeToClassMap(T msg) {
         String typeUrl = getTypeUrl(msg.getDescriptorForType());
+        log.error("Adding "+ typeUrl + " serializer map..");
```
WenbinZhu: log.error? Is this used when you were debugging and forgot to delete?

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
```diff
@@ -86,6 +86,13 @@ public Object deserialize(ByteBuf b, CorfuRuntime rt) {
             bbis.readFully(data);
             Record record = Record.parseFrom(data);
             Any payload = record.getPayload();
+            if (!classMap.containsKey(payload.getTypeUrl())) {
+                log.error("Deserialization error: Unknown type "+payload.getTypeUrl()+ " dumping map..\n");
+                for (String entry: classMap.keySet()) {
+                    log.error(entry + "=>" + classMap.get(entry));
```
WenbinZhu: Same here?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: I kept it as error because it is unsupported at this time and folks should really never see this.
In next release, we can move this to info once a migration path has been established.
Does that sound ok?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Good catch! thanks - this is from the debugging patch.

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
hisundar: No, we want this line to be printed out even going forward, since it helps diagnose a tricky deserialization issue, so this is intentional!

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
WenbinZhu: Sure.

runtime/src/main/java/org/corfudb/util/serializer/ProtobufSerializer.java
WenbinZhu: Okay, byw can you make it more descriptive?

hisundar: ok I will try

PR #2486, opened 2020-03-20
Bug fix + better errorhandling for writeRecords

PavelZaytsev:
## Overview

Description:

- When we calculate a difference between the known addresses and the remaining addresses, there is a possibility of IndexOutOfBoundsException. This is now fixed by copying all the required range to the immutable list at the beginning and then calculating delta between this list and the known addresses in range.
- For the writeRecords method, during a state transfer we need to retry only for the specific exceptions. If there are other exceptions, for example, WrongEpochException or IllegalArgumentException, we need to fail faster. 

Why should this be merged: 
- Bug fix, error handling improvements.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
@@ -7,6 +7,10 @@ import org.corfudb.infrastructure.log.statetransfer.batch.TransferBatchRequest;
 import org.corfudb.infrastructure.log.statetransfer.batch.TransferBatchResponse;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.runtime.clients.LogUnitClient;
+import org.corfudb.runtime.exceptions.NetworkException;
+import org.corfudb.runtime.exceptions.OverwriteException;
+import org.corfudb.runtime.exceptions.RetryExhaustedException;
+import org.corfudb.util.CFUtils;
 import org.corfudb.util.Sleep;
 
 import java.time.Duration;
@@ -15,6 +19,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
 
 import static org.corfudb.infrastructure.log.statetransfer.batch.TransferBatchResponse.TransferStatus.SUCCEEDED;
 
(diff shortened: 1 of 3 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
```diff
@@ -43,54 +48,68 @@
      */
     default TransferBatchResponse writeRecords(ReadBatch readBatch, LogUnitClient logUnitClient,
                                                int writeRetriesAllowed, Duration writeSleepDuration) {
-        List<Long> totalAddressesInRequest = readBatch.getAddresses();
-        Optional<String> destination = readBatch.getDestination();
+        if (readBatch == null || readBatch.getAddresses().isEmpty()) {
```
xnull: can we use Optional?

infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
```diff
@@ -43,54 +48,68 @@
...
+            throw new IllegalArgumentException("The read batch is empty.");
```
xnull: can we return `Result` rather than throwing an exception?

infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
```diff
@@ -43,54 +48,68 @@
...
+                transferBatchResponse = Optional.of(TransferBatchResponse
+                        .builder()
+                        .transferBatchRequest(new TransferBatchRequest(totalAddressesInRequest,
+                                readBatch.getDestination()))
+                        .status(SUCCEEDED)
+                        .build());
+                break;
+            } catch (OverwriteException | TimeoutException | NetworkException e) {
+                lastWriteException = Optional.of(e);
+                Sleep.sleepUninterruptibly(writeSleepDuration);
```
xnull: It would be great if you bring CFUtils changes from your second StateTransfer PR.  
Then you can get rid of sleep's (at least here)

P.S. also those changes in CFUtils are very useful, I miss it often

infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
```diff
@@ -43,54 +48,68 @@
...
+                // Get all the addresses that were supposed to be written to a stream log.
+                long start = totalAddressesInRequest.get(0);
```
xnull: can you use something like `findFirst`?

infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
```diff
@@ -43,54 +48,68 @@
...
                 // Create a new read batch with the missing data and retry.
                 ImmutableList<LogData> nonWrittenData = readBatch.getData()
                         .stream()
                         .filter(data -> nonWrittenAddresses.contains(data.getGlobalAddress()))
                         .collect(ImmutableList.toImmutableList());
                 readBatch = readBatch.toBuilder().data(nonWrittenData).build();
+            } catch (RuntimeException re) {
+                // If it's some other exception, e.g. IllegalArgumentException,
+                // WrongEpochException -> Propagate it to the caller.
+                throw new IllegalStateException(re);
```
xnull: this seems like a mess, would be great if you can come up with a better solution but I understand if the current approach makes smaller impact on the system

Issue #2489, opened 2020-03-26
CorfuTable.values() returns a Set which does not conform to Java map API


WenbinZhu:
## Overview

CorfuTable.values() used to return a list, but patch #2201 changed it to return a set. This does not conform to Java map API because it is supposed to return a Collection with possible duplicates.

WenbinZhu:
Addressed in #2490

PR #2490, opened 2020-03-26
Fix CorfuTable values() API that incorrectly deduplicates on return.

WenbinZhu:
## Overview

CorfuTable values() API is inherited from Java Map interface, which
is not supposed to perform deduplication on return. If CorfuTable
contains duplicates in values, the current values() API would return
less values than expected. This patch fixes the API to return an
immutable list instead of a set.

Related issue(s) (if applicable): #2489 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -1,5 +1,6 @@
 package org.corfudb.runtime.collections;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -473,7 +474,7 @@ public void clear() {
     public @Nonnull Collection<V> values() {
         return mainMap.values()
                 .stream()
-                .collect(ImmutableSet.toImmutableSet());
+                .collect(ImmutableList.toImmutableList());
```
xnull, line 477: please replace by 
`return ImmutableList.copyOf(mainMap.values());`

WenbinZhu: Done.

PR #2492, opened 2020-03-30
Fix Trim Exception on Raw Stream Access

annym:
## Overview

Description:

When a stream is being consumed directly (remaining/remainingUpTo
in the StreamViews), the readCpQueue is not being cleared.

If on the first access there is a valid CP to load from, the addresses
belonging to it will always remain in memory in the readCPQueue.
If other entries are found, these will be read along the ones in the readCPQueue (always,
as they're not cleared), when they're trimmed we will throw a TrimmedException on an old CP.

Why should this be merged:  TrimmedExceptions being thrown on direct stream access, provokes application restarts in a cycle.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
@@ -352,6 +352,7 @@ public abstract class AbstractQueuedStreamView extends
                 // to be thrown. By updating the global pointer to include the hole,
                 // we prevent that from happening.
                 .filter(x -> x.isData() || x.isHole())
+                // Case of Checkpoint will never load these addresses, cause this is another stream
                 .filter(x -> x.containsStream(context.id))
                 .collect(Collectors.toList());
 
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -370,6 +371,7 @@ private void processTrimmedException(TrimmedException te) {
         } else {
             // Clear the entries which were read
             context.readQueue.headSet(maxGlobal, true).clear();
+            context.readCpQueue.headSet(maxGlobal, true).clear();
```
WenbinZhu: Actually for raw streams, looks like we never load the checkpoints: in line 356 there is a filter `.filter(x -> x.containsStream(context.id))`, which filters out the entries on checkpoint stream, but we still read it because the filter is applied after the read, which is unnecessary. If that is the case, I think we need a clean up, otherwise it's ineffiecient and the logic is fragile, but maybe be can do later as we are short of time now.

annym: so, that's the comment I added as I reminder for us to discuss on what is the right behavior. We do "load" the checkpoint (in the sense that we get the addresses for it and we read them, but we don't return them to the consumer of the remaining API). So we need to define the correct behavior, if we decide not to return them we should do a bit of refactoring, cause that means that trying to reuse the stream layer In the same way for object consumption vs. raw consumption was actually not correct and is inefficient, as they behave differently.

PR #2495, opened 2020-03-31
CorfuQueue: Capture commit token & fix up Queue for ordering

hisundar:
## Overview

The Queue needs to capture the exact address its entries were
written to, so it can then order its entries, of its orderless
hashmap (corfuTable), in the same order that its transactions
materialized.
    
To do this register a preCommitCallback that intercepts the
Sequencer's blessed tokenResponse and fills it up into
the payload of the fields of the Queue.
    
+ Randomize sleep in test case to catch out of order reliably.

Add test case for both inorder and out of order between enqueue op and txEnd.
Co-authored-by: Xiaoqin Ma <maxiaoqin2005@gmail.com>

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
@@ -1,16 +1,21 @@
 package org.corfudb.runtime.collections;
 
 import com.google.common.reflect.TypeToken;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.runtime.CorfuRuntime;
-import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
+import org.corfudb.runtime.object.transactions.TransactionalContext.PreCommitListener;
+import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.CorfuGuidGenerator;
 import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -39,7 +44,7 @@ public class CorfuQueue<E> {
     /**
      * The main CorfuTable which contains the primary key-value mappings.
      */
-    private final CorfuTable<Long, E> corfuTable;
+    private final CorfuTable<Long, CorfuQueueRecord<E>> corfuTable;
     private final CorfuRuntime runtime;
     private final CorfuGuidGenerator guidGenerator;
 
@@ -49,7 +54,7 @@ public class CorfuQueue<E> {
                 () -> new StreamingMapDecorator<>(new LinkedHashMap<Long, E>());
         this.runtime = runtime;
         corfuTable = runtime.getObjectsView().build()
-                .setTypeToken(new TypeToken<CorfuTable<Long, E>>() {})
+                .setTypeToken(new TypeToken<CorfuTable<Long, CorfuQueueRecord<E>>>() {})
                 .setStreamName(streamName)
                 .setArguments(indices, mapSupplier)
                 .setSerializer(serializer)
@@ -75,59 +80,57 @@ public class CorfuQueue<E> {
      * with their snapshot+index id (also a long) which represents a global comparable ordering.
      * This class encapsulates these two longs into one Id and add rules on comparability.
      */
+    @EqualsAndHashCode
     public static class CorfuRecordId implements Comparable<CorfuRecordId> {
-        private final UUID id;
-        public CorfuRecordId(long ordering, long uniqueId) {
-            this.id = new UUID(ordering, uniqueId);
+        @Setter
+        @Getter
+        private long epoch;
+
+        @Setter
+        @Getter
+        private long sequence;
+
+        @Getter
+        private long entryId;
+
+        public CorfuRecordId(long epoch, long sequence, long entryId) {
+            this.epoch = epoch;
+            this.sequence = sequence;
+            this.entryId = entryId;
+        }
+
+        final static int BITS_FOR_SEQUENCE = 40;
+        public CorfuRecordId(UUID from) {
+            this.epoch = from.getMostSignificantBits()>>BITS_FOR_SEQUENCE;
+            this.sequence = from.getMostSignificantBits()&((1L<<BITS_FOR_SEQUENCE) - 1);
+            this.entryId = from.getLeastSignificantBits();
         }
 
         /**
-         * @return Return only the unique part of the id without the ordering
+         * @return Pack CorfuRecordId into a 16 byte UUID
+         * Q: Is it safe?
+         * At the rate of 2ms per transaction, sequence will take  69 years to rollover.
+         * At the rate of 1 cluster reconfig per minute, epoch will take 31 years to rollover.
          */
-        public long getEntryId() {
-            return id.getLeastSignificantBits();
-        }
-
-        /**
-         * @return Return only the ordering part of the entry without the id.
-         */
-        public long getOrdering() {
-            return id.getMostSignificantBits();
+        public UUID asUUID() {
+            return new UUID(epoch<<BITS_FOR_SEQUENCE | sequence, entryId);
         }
 
         /**
          * It's NOT ok to compare two objects if their ordering metadata is dissimilar.
          * @param o object to compare against.
-         * @throws IllegalArgumentException if the two Ids are not comparable.
          * @return results of comparison.
          */
         @Override
         public int compareTo(CorfuRecordId o) {
-            if (this.id.getMostSignificantBits() == 0 && o.id.getMostSignificantBits() != 0) {
-                throw new IllegalArgumentException(
-                        "Incompatible CorfuRecordId comparison: ordering unavailable");
-            }
-            if (this.id.getMostSignificantBits() !=0 && o.id.getMostSignificantBits() == 0) {
-               throw new IllegalArgumentException(
-                       "Incompatible CorfuRecordId comparison: order of compared object unknown");
-            }
-            if (this.id.getLeastSignificantBits() == o.id.getLeastSignificantBits()) {
-                return 0;
-            }
-            return id.compareTo(o.id);
-        }
-
-        /**
-         * It is ok to check equality of a CorfuRecordId with ordering data against one without.
-         * @param o object to compare against.
-         * @return
-         */
-        public boolean equals(CorfuRecordId o) {
-            return id.getLeastSignificantBits() == o.id.getLeastSignificantBits();
+            return Comparator.comparing(CorfuRecordId::getEpoch)
+                    .thenComparing(CorfuRecordId::getSequence)
+                    .thenComparing(CorfuRecordId::getEntryId)
+                    .compare(this, o);
         }
 
         public String toString() {
-            return id.toString();
+            return String.format("%s|%s|%s",epoch, sequence, entryId);
         }
     }
 
@@ -138,21 +141,48 @@ public class CorfuQueue<E> {
      * Capacity restrictions and backoffs must be implemented outside this
      * interface. Consider validating the size of the queue against a high
      * watermark before enqueue.
+     * WARNING: CorfuQueue won't work if transactional enqueues are mixed with
+     *          non-transactional enqueues.
+     *          Either use transactions for all enqueues or none.
      *
      * @param e the element to add
      * @throws IllegalArgumentException if some property of the specified
      *         element prevents it from being added to this queue
-     * @return Unique ID representing this entry in the persistent queue
-     * WARNING: The ID returned by this function will NOT be the comparable with the ID
-     *          returned by the CorfuQueueRecord from entryList() method because if this
-     *          method executes within a transaction, the comparable ordering is not
-     *          known until the transaction commits.
-     *          The ID returned here is only really useful for remove() operations.
      */
-    public CorfuRecordId enqueue(E e) {
+    public void enqueue(E e) {
         final Long id = guidGenerator.nextLong();
-        corfuTable.put(id, e);
-        return new CorfuRecordId(0, id);
+        CorfuQueueRecord<E> queueEntry;
+
+        // If we are in a transaction, then we need the commit address of this transaction
+        // to fix up as the ordering
+        if (TransactionalContext.isInTransaction()) {
+            queueEntry = new CorfuQueueRecord<>(Address.NON_ADDRESS, Address.NON_ADDRESS, id, e);
+            /**
+             * This is a callback that is placed into the root transaction's context on
+             * the thread local stack which will be invoked right after this transaction
+             * is deemed successful and has obtained a final sequence number to write.
+             */
+            class QueueEntryAddressGetter implements PreCommitListener {
+                private CorfuQueueRecord<E> queueRecord;
+                private QueueEntryAddressGetter(CorfuQueueRecord<E> queueRecord) {
+                    this.queueRecord = queueRecord;
+                }
+                @Override
+                public void preCommitCallback(TokenResponse tokenResponse) {
+                    queueRecord.getRecordId().setEpoch(tokenResponse.getEpoch());
+                    queueRecord.getRecordId().setSequence(tokenResponse.getSequence());
+
+                    log.trace("preCommitCallback for Queue: " + queueRecord.getRecordId().toString());
+                }
+            }
+            QueueEntryAddressGetter addressGetter = new QueueEntryAddressGetter(queueEntry);
+            log.trace("enqueue: Adding preCommitListener for Queue: " + queueEntry.getRecordId().toString());
+            TransactionalContext.getRootContext().addPreCommitListener(addressGetter);
+        } else {
+            queueEntry = new CorfuQueueRecord<>(0, 0, id, e);
+        }
+
+        corfuTable.put(queueEntry.getRecordId().getEntryId(), queueEntry);
     }
 
     /**
@@ -172,6 +202,7 @@ public class CorfuQueue<E> {
          * because if this method is wrapped in a transaction, the order is established only later.
          */
         @Getter
+        @Setter
         private final CorfuRecordId recordId;
 
         @Getter
@@ -181,8 +212,8 @@ public class CorfuQueue<E> {
             return String.format("%s=>%s", recordId, entry);
         }
 
-        CorfuQueueRecord(long ordering, long entryId, E entry) {
-            this.recordId = new CorfuRecordId(ordering, entryId);
+        CorfuQueueRecord(long epoch, long sequence, long entryId, E entry) {
+            this.recordId = new CorfuRecordId(epoch, sequence, entryId);
             this.entry = entry;
         }
 
@@ -190,39 +221,24 @@ public class CorfuQueue<E> {
         public int compareTo(CorfuQueueRecord<? extends E> o) {
             return this.recordId.compareTo(o.getRecordId());
         }
-    }
 
-    /**
-     * We need to encode 2 pieces of information into an 8 byte long to represent ordering.
-     * 1. We use 40 bits for CorfuQueue's snapshot version information.
-     * 2. We use 24 bits for the index within that CorfuQueue's snapshot.
-     */
-    final private static int MAX_BITS_FOR_INDEX = 24;
-    final private static int MAX_INDEX_ENTRIES = (1<<MAX_BITS_FOR_INDEX) - 1;
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            CorfuQueueRecord<?> that = (CorfuQueueRecord<?>) o;
+            return getRecordId().equals(that.getRecordId());
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(getRecordId());
+        }
+    }
 
     /**
      * Returns a List of CorfuQueueRecords sorted by the order in which the enqueue materialized.
      * This is the primary method of consumption of entries enqueued into CorfuQueue.
-     *
-     * To re-constitute the commit order in the presence of transactions the CorfuRecords returned
-     * here have their UUID fields composed of two parts:
-     *     +----------------------------------------------------------------------------+
-     *     | Stream snapshot |Index in snapshot|     ID of the entry in the map        |
-     *     +----------------------------------------------------------------------------+
-     *     <----5 bytes------><----3 bytes-----><------- 8 bytes ----------------------->
-     *
-     * <p>Note: The ordering returned can be different based on the call. For example:
-     * Queue contents at start:
-     *         --- id1->R1 ----
-     * thread1: snapshot at 99
-     * id2 = enqueue(R2);
-     * list1 = entryList(); // contents returned are [99|0+id1 => R1, 99|1+id2 => R2]
-     *
-     * thread2: snapshot at 100
-     * list2 = entryList(); // contents returned are [100|0+id1 => R1]
-     *
-     * As seen above R1 returned to thread1 and thread2 have same id but different order.
-     * </p>
      *
      * <p>This function currently does not return a view like the java.util implementation,
      * and changes to the entryList will *not* be reflected in the map. </p>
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -76,23 +77,21 @@ public int size() {
      * This class encapsulates these two longs into one Id and add rules on comparability.
      */
     public static class CorfuRecordId implements Comparable<CorfuRecordId> {
-        private final UUID id;
-        public CorfuRecordId(long ordering, long uniqueId) {
-            this.id = new UUID(ordering, uniqueId);
-        }
+        @Setter
```
xnull: is it possible to get rid of setters?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -123,11 +114,11 @@ public int compareTo(CorfuRecordId o) {
          * @return
          */
         public boolean equals(CorfuRecordId o) {
-            return id.getLeastSignificantBits() == o.id.getLeastSignificantBits();
+            return (epoch == o.getEpoch()) && (sequence == o.getSequence()) && entryId == o.getEntryId();
```
xnull: When you implement `equals` method you have to implement a hashCode method, if not then functionality will be broken.

xnull: Also, please add `@Owerride`
image

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -172,6 +180,7 @@ public CorfuRecordId enqueue(E e) {
          * because if this method is wrapped in a transaction, the order is established only later.
          */
         @Getter
+        @Setter
```
xnull: You must be very careful with setters. They always bring much more complexity to the code, make it fragile, more complicated to analyze and there are always problems with concurrency.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -181,8 +190,8 @@ public String toString() {
             return String.format("%s=>%s", recordId, entry);
         }
 
-        CorfuQueueRecord(long ordering, long entryId, E entry) {
-            this.recordId = new CorfuRecordId(ordering, entryId);
+        CorfuQueueRecord(long epoch, long ordering, long entryId, E entry) {
```
xnull: why not to pass `CorfuRecordId` as a parameter?

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -248,6 +250,17 @@ public void syncWithRetryUnsafe(VersionLockedObject vlo,
...
+    /**
+     * Add an object that needs extra processing right before commit happens
+     *
+     * @param preCommitListener The context of the object that needs extra processing
+     *                         along with its lambda.
+     */
+    public abstract void addPreCommitListener(TransactionalContext.PreCommitListener preCommitListener);
+
+    @Getter
+    private List<TransactionalContext.PreCommitListener> preCommitListeners = new ArrayList<>();
```
xnull: Please don't use TransactionalContext as a prefix. Please add TransactionalContext to imports

runtime/src/main/java/org/corfudb/runtime/object/transactions/SnapshotTransactionalContext.java
```diff
@@ -85,4 +85,9 @@ public SnapshotTransactionalContext(Transaction transaction) {
     public void addTransaction(AbstractTransactionalContext tc) {
         throw new UnsupportedOperationException("Can't merge into a readonly txn (yet)");
     }
+
+    @Override
+    public void addPreCommitListener(TransactionalContext.PreCommitListener preCommitListener) {
```
xnull: please add TransactionalContext to imports

hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: because these fields need to be accessed and set individually at different times I thought it would be cleaner to separate those two.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: yes unfortunately we do not have access to the commit sequence and epoch until the transaction actually commits, so we cannot go with immutable construct. It has to be set later, creating a new copy would break it away from the SMR version.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: no, it breaks SMR if we create a new object - we *must* modify the exact value that was added to the table's hash map.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks!

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
xnull: ok

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
xnull: sad to hear that we can't afford immutability, so, ok

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -76,58 +78,56 @@ public int size() {
...
-         * @param o object to compare against.
+         * It is NOT ok to check equality of a CorfuRecordId with ordering data against one without.
+         * @param otherObj object to compare against.
          * @return
          */
-        public boolean equals(CorfuRecordId o) {
-            return id.getLeastSignificantBits() == o.id.getLeastSignificantBits();
+        @Override
+        public boolean equals(Object otherObj) {
+            if (this == otherObj) return true;
```
xnull: Actually instead of bolierplate code you should use @EqualsAndHashCode annotation from lombok. Can you make that change?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -149,10 +149,27 @@ public String toString() {
      *          known until the transaction commits.
      *          The ID returned here is only really useful for remove() operations.
      */
-    public CorfuRecordId enqueue(E e) {
+    public void enqueue(E e) {
         final Long id = guidGenerator.nextLong();
-        corfuTable.put(id, e);
-        return new CorfuRecordId(0, id);
+        CorfuQueueRecord<E> queueEntry = new CorfuQueueRecord<>(0, 0, id, e);
```
WenbinZhu: Is it better to use invalid epoch and sequence instead of 0 ?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -149,10 +149,27 @@ public String toString() {
...
+        class QueueEntryAddressGetter implements PreCommitListener {
+            CorfuQueueRecord<E> queueRecord;
```
WenbinZhu: Private

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -149,10 +149,27 @@ public String toString() {
...
+            QueueEntryAddressGetter(CorfuQueueRecord<E> queueRecord) {
```
WenbinZhu: Private

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -181,15 +199,28 @@ public String toString() {
             return String.format("%s=>%s", recordId, entry);
         }
 
-        CorfuQueueRecord(long ordering, long entryId, E entry) {
-            this.recordId = new CorfuRecordId(ordering, entryId);
+        CorfuQueueRecord(long epoch, long ordering, long entryId, E entry) {
```
WenbinZhu: ordering -> sequence ?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -242,41 +253,20 @@ public int compareTo(CorfuQueueRecord<? extends E> o) {
...
-                    .begin();
-            snapshotVersion = TransactionalContext.getCurrentContext()
-                    .getSnapshotTimestamp().getSequence();
-            startedNewTransaction = true;
-        }
         List<CorfuQueueRecord<E>> copy = new ArrayList<>(
                 Math.min(corfuTable.size(), maxEntries)
         );
 
+    Comparator<Map.Entry<Long, CorfuQueueRecord<E>>> recordIdComparator = (r1, r2) ->
```
WenbinZhu: Fix Indent

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -242,41 +253,20 @@ public int compareTo(CorfuQueueRecord<? extends E> o) {
...
+                r1.getValue().getRecordId().compareTo(r2.getValue().recordId);
         int index = 0;
-        for (Long entryId : corfuTable.keySet().stream()
-                .filter(e -> e > entriesAfter)
-                .sorted().collect(Collectors.toList())) {
+        for (Map.Entry<Long, CorfuQueueRecord<E>> entry : corfuTable.entryStream()
+                .filter(e -> e.getValue().getRecordId().compareTo(entriesAfter) > 0)
+                .sorted(recordIdComparator).collect(Collectors.toList())) {
             if (++index >= maxEntries) {
```
WenbinZhu: You can put a `.limit(maxEntries)` after `.filter()` to avoid collecting and sorting all entries and ++index. This should be more efficient.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -285,23 +275,23 @@ public int compareTo(CorfuQueueRecord<? extends E> o) {
      * @return all the entries in the Queue
      */
     public List<CorfuQueueRecord<E>> entryList() {
-        return this.entryList(0L, MAX_INDEX_ENTRIES);
+        return this.entryList(new CorfuRecordId(0,0,0), MAX_INDEX_ENTRIES);
```
WenbinZhu: Is it better to use invalid epoch and sequence instead of 0 ? Same for the rests

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -248,6 +251,17 @@ public void syncWithRetryUnsafe(VersionLockedObject vlo,
      */
     public abstract void addTransaction(AbstractTransactionalContext tc);
 
+    /**
+     * Add an object that needs extra processing right before commit happens
+     *
+     * @param preCommitListener The context of the object that needs extra processing
+     *                         along with its lambda.
+     */
+    public abstract void addPreCommitListener(PreCommitListener preCommitListener);
```
WenbinZhu: I'm wondering if this is correct in case of nested transactions. In that case we would merge the read/write set in the child transaction's context into parent's context, however the preCommitListeners seems not doing this.

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -286,4 +286,9 @@ public long getConflictSetAndCommit(ConflictSetInfo conflictSet) {
         log.trace("Commit[{}] Written to {}", this, address);
         return address;
     }
+
+    @Override
+    public void addPreCommitListener(TransactionalContext.PreCommitListener preCommitListener) {
+        this.getPreCommitListeners().add(preCommitListener);
```
annym: add trace or debug logging?

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -168,6 +168,17 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
...
             try {
+                if (TransactionalContext.isInTransaction()) {
+                    // If this transaction has entries that wish to capture the committed address
+                    // invoke its preCommitCallbacks with the tokenResponse from the sequencer.
+                    // Note that we might invoke the same method multiple times on retries,
+                    // which means the preCommitCallback must be idempotent.
+                    TokenResponse finalTokenResponse = tokenResponse;
+                    TransactionalContext.getRootContext()
+                            .getPreCommitListeners()
+                            .forEach(e -> e.preCommitCallback(finalTokenResponse));
```
annym: add some logging?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: fixed. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: would it matter? keeping it as zero allows these non-transactional entries to go to the first of the queue - might make it easier for testing

WenbinZhu: But then if you compare the a non-tranactional entry with a tranactional entry who was committed at epoch 0, sequence 0,  it would be confusing right?

hisundar: yes, that isn't supported. ok let me catch that case and assert.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
hisundar: I thought about this - that is the reason I always pick the RootContext to store these preCommitListeners. As you said nesting transactions will keep merging their writeSets() anyway which is is unnecessary because only the address of the parent transaction matters anyway.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -149,10 +140,27 @@ public String toString() {
      *          known until the transaction commits.
      *          The ID returned here is only really useful for remove() operations.
      */
-    public CorfuRecordId enqueue(E e) {
+    public void enqueue(E e) {
         final Long id = guidGenerator.nextLong();
-        corfuTable.put(id, e);
-        return new CorfuRecordId(0, id);
+        CorfuQueueRecord<E> queueEntry = new CorfuQueueRecord<E>(0, 0, id, e);
+        class QueueEntryAddressGetter implements TransactionalContext.PreCommitListener {
```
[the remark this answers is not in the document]
hisundar: fixed. thanks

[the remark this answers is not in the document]
hisundar: fixed. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -149,10 +140,27 @@ public String toString() {
...
+            CorfuQueueRecord<E> queueRecord;
+            QueueEntryAddressGetter(CorfuQueueRecord<E> queueRecord) {
+                this.queueRecord = queueRecord;
+            }
+            @Override
+            public void preCommitCallback(TokenResponse tokenResponse) {
+                queueRecord.getRecordId().setEpoch(tokenResponse.getEpoch());
```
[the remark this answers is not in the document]
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -149,10 +140,27 @@ public String toString() {
...
+                queueRecord.getRecordId().setSequence(tokenResponse.getSequence());
+            }
+        }
+        // If we are in a transaction, then we need the commit address of this transaction
```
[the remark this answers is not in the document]
hisundar: not just that, I am enforcing it via an explicit check.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -242,41 +253,20 @@ public int compareTo(CorfuQueueRecord<? extends E> o) {
...
-            long ordering = (snapshotVersion << MAX_BITS_FOR_INDEX) | index;
-            CorfuQueueRecord<E> record = new CorfuQueueRecord<>(
-                    ordering, entryId, corfuTable.get(entryId)
-            );
-            copy.add(record);
-        }
-        // Given that we are using a WRITE_AFTER_WRITE on a read-only txn, we expect no aborts.
-        if (startedNewTransaction) {
-            runtime.getObjectsView().TXEnd();
+            copy.add(entry.getValue());
```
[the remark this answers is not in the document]
hisundar: I think your comment is on old version of the code - the whole loop is fixed up based on a previous comment, so I guess it can't happen anymore?

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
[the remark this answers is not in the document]
hisundar: added trace, thanks

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -75,59 +80,55 @@ public int size() {
...
-            }
-            if (this.id.getMostSignificantBits() !=0 && o.id.getMostSignificantBits() == 0) {
-               throw new IllegalArgumentException(
-                       "Incompatible CorfuRecordId comparison: order of compared object unknown");
-            }
-            if (this.id.getLeastSignificantBits() == o.id.getLeastSignificantBits()) {
-                return 0;
+            if ((this.sequence != Address.NON_ADDRESS && o.sequence == Address.NON_ADDRESS) ||
+                    (this.sequence == Address.NON_ADDRESS && o.sequence != Address.NON_ADDRESS)) {
+                throw new IllegalStateException("CorfuQueue cannot mix transactional and non-transactional types");
```
WenbinZhu, line 122: Looks like you are still using 0, 0 for non-transactional entries (the else branch in enqueue) ?

hisundar: exactly, Queue is still usable if all enqueue() are non-transactional (unit tests only at this point)

hisundar: Sorry good catch

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
```diff
@@ -88,11 +93,13 @@ public Record(CorfuRecordId id, String data) {
...
+                    corfuQueue.enqueue(queueData);
+                    // Each transaction may or may not sleep to simulate out of order between enQ & commit
+                    TimeUnit.MILLISECONDS.sleep(coinToss);
                     lock.lock();
-                    CorfuRecordId id = corfuQueue.enqueue(queueData);
                     final long streamOffset = TXEnd();
-                    validator.add(new Record(id, queueData));
-                    log.debug("ENQ:" + id + "=>" + queueData + " at " + streamOffset);
+                    validator.add(new Record(new CorfuRecordId(0,0,i), queueData));
+                    log.debug("ENQ:" + i + "=>" + queueData + " at " + streamOffset);
```
xnull, line 102: Please fix
```
log.debug("ENQ: {} => {} at {}", i, queueData, streamOffset);
```

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
xnull: not sure about SMR, but if you are going to change these fields and use them to calculate hashCode then it is very dangerous operation you can break HashMap by that.
I mean if you change those fields after you add it to any java data structure like HashMap, Queue and so on, you will not be able to find that entry in the data structure

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
hisundar: done. thanks

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
```diff
@@ -88,11 +93,13 @@ public Record(CorfuRecordId id, String data) {
                     TXBegin(txnType);
                     Long coinToss = new Random().nextLong() % numConflictKeys;
                     conflictMap.put(coinToss, coinToss);
+                    corfuQueue.enqueue(queueData);
+                    // Each transaction may or may not sleep to simulate out of order between enQ & commit
+                    TimeUnit.MILLISECONDS.sleep(coinToss);
                     lock.lock();
```
xiaoqin2012, line 99: We should remove the lock and generated some tx abort scenarios. You can define validator as a hashtable instead of an array. I am not clear if tx aborted, if the record will be removed from the queue or not?

runtime/src/main/java/org/corfudb/runtime/object/transactions/SnapshotTransactionalContext.java
```diff
@@ -85,4 +85,9 @@ public SnapshotTransactionalContext(Transaction transaction) {
...
+        throw new UnsupportedOperationException("Can't register precommit hooks in readonly txn");
```
pankti-m, line 91: are snapshot transactions considered read-only?

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
pankti-m, line 260: Shouldnt it be overridden in WriteAfterWriteTransactionalContext as well?

hisundar: WriteAfterWrite inherits from OptimisticTransaction and will end up inheriting this method too.

runtime/src/main/java/org/corfudb/runtime/object/transactions/SnapshotTransactionalContext.java
hisundar: yes, there are similar exceptions for attempting to write in a snapshot transaction

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
hisundar: Aborts are definitely happening because we conflict on one of two values, for example in a test with 2000 transactions only about 1500 succeed. But I will add an assert to verify this too.

hisundar: Actually it turns out travis sometimes serializes the runs which means there might not be any aborts due to concurrency. So I can't add an assert or test will become very non-deterministic :(

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
WenbinZhu: @hisundar Is this fixed?

hisundar: yes I removed the check as you suggested.

PR #2497, opened 2020-03-31
Take metadata size into account when calculating the logData size.

pankti-m:
The current method only considers the data size, which can give
innacurate result if the metadata size is significant.

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -6,11 +6,13 @@ import io.netty.buffer.Unpooled;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.util.EnumMap;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+import org.apache.commons.collections.iterators.EntrySetMapIterator;
 import org.corfudb.common.compression.Codec;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.LogEntry;
@@ -35,6 +37,8 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     private ByteBuf serializedCache = null;
 
     private int lastKnownSize = NOT_KNOWN;
+
+    final EnumMap<LogUnitMetadataType, Object> metadataMap;
 
     private final transient AtomicReference<Object> payload = new AtomicReference<>();
 
@@ -64,6 +68,7 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
 
     /**
      * Return the payload.
+     * This method is invoked by the clients to fetch the payload from the LogData
      */
     public Object getPayload(CorfuRuntime runtime) {
         Object value = payload.get();
@@ -95,7 +100,6 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
                             }
                             value = actualValue == null ? this.payload : actualValue;
                             this.payload.set(value);
-                            lastKnownSize = data.length;
                         } catch (Throwable throwable) {
                             log.error("Exception caught at address {}, {}, {}",
                                     getGlobalAddress(), getStreams(), getType());
@@ -129,6 +133,7 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
         if (serializedCache == null) {
             serializedCache = Unpooled.buffer();
             doSerializeInternal(serializedCache);
+            lastKnownSize = serializedCache.readableBytes();
         } else {
             serializedCache.retain();
         }
@@ -136,24 +141,22 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
 
     @Override
     public int getSizeEstimate() {
-        byte[] tempData = data;
-        if (tempData != null) {
-            return tempData.length;
-        } else if (lastKnownSize != NOT_KNOWN) {
+        if (lastKnownSize != NOT_KNOWN) {
             return lastKnownSize;
         }
-        log.warn("getSizeEstimate: LogData size estimate is defaulting to 1,"
-                + " this might cause leaks in the cache!");
         return 1;
     }
 
-    @Getter
-    final EnumMap<LogUnitMetadataType, Object> metadataMap;
+    @Override
+    public EnumMap<IMetadata.LogUnitMetadataType, Object> getMetadataMap() {
+        return metadataMap;
+    }
 
     /**
      * Return the payload.
      */
     public LogData(ByteBuf buf) {
+        lastKnownSize = buf.readableBytes();
         type = ICorfuPayload.fromBuffer(buf, DataType.class);
         if (type == DataType.DATA) {
             data = ICorfuPayload.fromBuffer(buf, byte[].class);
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -138,7 +138,7 @@ public synchronized void acquireBuffer() {
     public int getSizeEstimate() {
         byte[] tempData = data;
         if (tempData != null) {
-            return tempData.length;
+            return tempData.length + metadataMap.size() + DataType.values().length;
```
WenbinZhu: metadataMap.size() is the number of entries, not the size.

pankti-m: i could not find a straightforward way to calculate the number of bytes used by metadataMap.  size() is not supported on the key and value types.  Similary for DataType - had to use the number of data types in the enum.  

I can get the size from the byteBuffer in the constructor.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -147,13 +150,21 @@ public int getSizeEstimate() {
         return 1;
     }
 
+    private int calculateMetaDataMapSize() {
+        int size = 0;
+        for (Map.Entry<LogUnitMetadataType, Object> entry : metadataMap.entrySet()) {
+            entry.
+        }
```
xiaoqin2012, line 157: Is something missing here?

pankti-m: yes, it was to be deleted.  Removed now.  Thanks.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -154,6 +149,7 @@ public int getSizeEstimate() {
      * Return the payload.
      */
     public LogData(ByteBuf buf) {
+        lastKnownSize = buf.readableBytes();
```
Maithem, line 152: Serialized is usually more compact than in-memory representation, not will be sufficient.

The size is used by the server cache only. Since the logunit doesn't deserialize the payload, I think a better estimation function would be to use the byte array size + sizeOf(metadata size).

sizeOf can be slow and consume a lot of memory, but since the metadata structure is very flat, I expect this function to have very low overhead and it will produce an accurate estimation of in memory size.

xnull: Can we write a benchmark for `sizeOf`? Because you can assume that the sizeOf operation is fast enough but in reality, it would be deadly slow. 
We have benchmarks module, specially designed for that.
This can be used as an example https://github.com/CorfuDB/corfudb-tests/blob/master/benchmarks/src/jmh/java/org/corfudb/benchmarks/util/DataGeneratorBenchmark.java

WenbinZhu: @Maithem Is it worth to use sizeOf? Because metadata is not the only one that has difference between in-memory representation and serialized size. LogUnit caches LogData, even though we calculate the exact in-memory size of metadata plus the payload size, LogData object itself has overheads, which means we can never have a precise size, unless use sizeOf(LogData), which would be more expensive. So does it worth to calculate the accurate in-memory size of metadata?

xnull: So, I have some numbers.
I have created a LogData object:
```
ByteBuf b = Unpooled.buffer();
            byte[] streamEntry = DataGenerator.generateDataString(1024).getBytes();
            Serializers.CORFU.serialize(streamEntry, b);
            LogData ld = new LogData(DataType.DATA, b);
            ld.setCheckpointedStreamId(UUID.randomUUID());
            ld.setClientId(UUID.randomUUID());
            ld.setCheckpointedStreamStartLogAddress(321L);
            ld.setGlobalAddress(45678L);
            Map<UUID, Long> bm = new HashMap<>();
            for (int i = 0; i < 100; i++) {
                bm.put(UUID.randomUUID(), rnd.nextLong());
            }

            ld.setBackpointerMap(bm);
```
I have compared serialized size vs in-memory:
`Serialized: 4096, sizeof: 11272`
So, 3x difference.

Regarding SizeOf:
The more you have records in the `metadataMap` the slower sizeOf operation.
Without any records in `metadataMap` the performance is about 300k operations on my laptop.
Having just 100 records in the `BackpointerMap<UUID, Long>` makes it much slower ~ 20k op/s

So, both ways (measuring serialized size vs sizeOf) are not good enough.
But measuring the serialized size gives us at least something.

xnull: And as you can see using: `sizeOf(metadataMap)` is the worst idea

WenbinZhu:
If we truly worry about the overheads of the in-memory size compared to serialized size, one thing we can do is use softReference on the cache's value:
```java
                .softValues()
                .build(...);
```
softValues() will wrap the value of cache in SoftReference, which will be automatically garbage collected by JVM on memory pressure, though there might be some minor impact on performance, however this guarantees no OOM happens.

pankti-m:
One option is to use the Instrumentation package in java - https://www.baeldung.com/java-size-of-object

But this also does not calculate the sizes of references of an object.  To do that, we have to traverse through them.

"Returns an implementation-specific approximation of the amount of storage consumed by the specified object. The result may include some or all of the object's overhead, and thus is useful for comparison within an implementation but not between implementations. The estimate may change during a single invocation of the JVM."

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
xiaoqin2012: Instead of use useOf(a Map), should use sizeOf(logdata). Not sure it will help. Some benchmark may help.

xiaoqin2012: Also, it is possible that the overhead is static : (11272 - 4096)? Will the overhead change for different payload objects? If the overhead is static, we can calculate the serialized size + the static overhead.

xiaoqin2012: I would suggest to do some benchmark for calculate metadata and logdata:
comparing the total serialized data size and the size we get with sizeOf api.
Is it 40% or 80%.

pankti-m: yes, it should be possible to test that using the heapdump we get with the new changes.  We will have lastKnownSize(serialized bytebuf size) in the dump and also the actual logData object with the in-mem size.  

Let us know if you have any other/better ideas.

ben-manes:
> softValues() will wrap the value of cache in java's SoftReference, which will be automatically garbage collected by JVM on memory pressure, though there might be some impact on performance, however this guarantees no OOM happens.

The is a myth and its painful when it is debunked. These objects get promoted to old generation, pollute the heap, and uses valuable memory. When a GC is required it causes an expensive full collection. This can cause death spirals of many GCs as the heap is filled to capacity, until eventually an out-of-memory error. Even though this could be collected, it exhausts the quota and the JVM crashes. You get much better, predictable performance by avoiding soft references if possible. After fighting with Apache PdfBox's usage due to crashing under load, I moved application logic to Lambdas just to deal with the memory problems it caused. That was on JDK11 with G1.

> One option is to use the Instrumentation package in java

[MemoryMeter](https://github.com/jbellis/jamm) was a popular package that does the traversal with a choice of estimation strategies. These are always best efforts and not exact, but could be good enough.

pankti-m:
Obsolete.

PR #2499, opened 2020-04-02
Clean definition of Direct Stream Access APIs

annym:
## Overview

Description:

The remaining and remainingUpTo APIs were thought to provide
    a direct access to the stream for the case of pollers. We have not
    taken care of the specifics of these APIs and this change aims to revisit
    or redefine the behavior.

- remaining should be able to load from a checkpoint and return the entries represented by it.
- remaining should be able to load from a checkpoint and the regular stream in a single call.
- remaining should not load from a checkpoint if the stream is seeked (pointer is moved) as
    we might return duplicates.
- If loading from a checkpoint the stream's global pointer should point to the address it represents
    in the regular stream and not the actual address of the checkpoint entry itself.

Pending: because remaining/remainingUpTo can return entries that come from the checkpoint stream or the regular stream (transactional or not), these APIs should return a List(SMREntry) instead of List(ILogData) as it is not expected that a consumer deals with the complexities of a checkpoint/regular stream.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractContextStreamView.java
@@ -9,6 +9,9 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
+import org.corfudb.protocols.logprotocol.LogEntry;
+import org.corfudb.protocols.logprotocol.SMREntry;
 import org.corfudb.protocols.wireprotocol.DataType;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
@@ -197,7 +200,16 @@ public abstract class AbstractContextStreamView<T extends AbstractStreamContext>
             getCurrentContext().setGlobalPointerCheckGCTrimMark(maxGlobal);
         } else {
             // Update pointer from log data and then validate final position of the pointer against GC trim mark.
-            updatePointer(entries.get(entries.size() - 1));
+
+            // Entries can come from a checkpoint or a regular stream, in the case it comes from a checkpoint only
+            // (last entry is checkpoint entry type) we should accordingly set the pointer to point to the
+            // address this checkpoint represents for the regular stream and not for the checkpoint.
+            Object lastEntry = entries.get(entries.size() - 1).getPayload(runtime);
+            if (lastEntry instanceof CheckpointEntry) {
+                getCurrentContext().setGlobalPointer(Long.parseLong(((CheckpointEntry)lastEntry).getDict().get(CheckpointEntry.CheckpointDictKey.START_LOG_ADDRESS)));
+            } else {
+                updatePointer(entries.get(entries.size() - 1));
+            }
             getCurrentContext().validateGlobalPointerPosition(getCurrentGlobalPosition());
         }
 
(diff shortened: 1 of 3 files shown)


annym:
Now, that we’re cleaning and redefining the behavior for direct stream access, I think we should change the signature of these APIs and instead of returning a List<ILogData> it should return a List<SMREntry>. The reason is that, we can load from a checkpoint or from a regular stream (transactional data or non-transactional data). Meaning that the payload of this ILogData could point to any of:

// ILogData.getPayload(rt) can be of three types:
    // CheckpointEntry: if it is a checkpoint
    // SMREntry: if it is a non-transactional write
    // MultiObjectSMREntry: if it is a transactional write

Each case requires a different processing to extract the actual updates, and handling the complexity of Checkpoint and Regular streams to the consumer I don’t think is correct. 

List<SMREntry> smrEntries = new ArrayList<>();
// Process Case of Checkpoint
    MultiSMREntry multiSMREntry = ((CheckpointEntry)data.getPayload(rt)).getSmrEntries();
    if (multiSMREntry != null) {
        smrEntries.addAll(multiSMREntry.getSMRUpdates(CorfuRuntime.getStreamID(streamA)));
    }
}

// Process Case MultiObjectSMREntry (transactional writes)
         smrEntries.addAll(((MultiObjectSMREntry)data.getPayload(rt)).getSMRUpdates(CorfuRuntime.getStreamID(streamA)));

//  Process Case SMREntry
/        smrEntries.add((SMREntry)data.getPayload(rt));


Pending: re-visit nextUpTo API which is also provided from direct stream access. This PR is taking into account only remaining/remainingUpTo

vjeko:
@annym I've been thinking a bit about the streaming layer. I believe we should take a second look at how this layer is actually composed and see if we can reduce the complexity.  For example, mixing coalesced (compacted) and linearizable (non-compacted) state might make sense from the VLO's perspective, but not if you are consuming the stream directly. There are also some open questions as to how to implement/expose virtual streams (entire-namespace stream, stream groups) that the current architecture does not facilitate.

Anyways, just something to consider.

annym:
> @annym I've been thinking a bit about the streaming layer. I believe we should take a second look at how this layer is actually composed and see if we can reduce the complexity. For example, mixing coalesced (compacted) and linearizable (non-compacted) state might make sense from the VLO's perspective, but not if you are consuming the stream directly. There are also some open questions as to how to implement/expose virtual streams (entire-namespace stream, stream groups) that the current architecture does not facilitate.
> 
> Anyways, just something to consider.

@vjeko That's precisely the objective to re define the streaming layer for the direct access case as we've never really thought of it thoroughly. I'll arrange a meeting for next week so we can discuss this!

vjeko:
FYI, I'm on PTO till the 10th.

achinmishra:
@vjeko Come on, you got bored and you wanted to have a slice of this :-D

WenbinZhu:
@annym Looks like the patch is assuming remaining/remainingUpTo are the only data consuming API for raw stream, but next/nextUpTo are also exposed, since we can not make it private to Corfu, I think one more thing we need to do is either make sure next/nextUpTo is compatible with raw stream as well or clearly docuement in the Javadoc not to use it with raw stream.

annym:
> @annym Looks like the patch is assuming remaining/remainingUpTo are the only data consuming API for raw stream, but next/nextUpTo are also exposed, since we can not make it private to Corfu, I think one more thing we need to do is either make sure next/nextUpTo is compatible with raw stream as well or clearly docuement in the Javadoc not to use it with raw stream.

@WenbinZhu Exactly! In tomorrow's meeting let's decide what we want to expose, and the expected behavior. This was just to kickoff the re-design and most likely we'll need to change the APIs signature and behavior.

PR #2503, opened 2020-04-05
While remove a router, the client should be removed too.

xiaoqin2012:
Overview
While adding a node to the system, a rounter is created and added to the router pool and a mapping information from endpoint to the router is adding to the layout view. 
While a router is stopped, it is removed from the router pool, but the mapping int he layout hasn't been removed. 

While a router is stopped and removed, the mapping information should be removed too to avoid to reuse the stale router.  

Description:
 * If the client is not removed, the stopped router will be reused.
 * Accessing the stopped router will trigger an exception.

Why should this be merged: 
While removing a router, the client on the clientMap should be removed too.
Otherwise, the client bounding the router will be reused and the stopped router will be reused again. 

Add a test case to simulate the race condition:
1. The cluster is setup with one server localhost:9000.
2. l = get the layout with one server and one router.
3. setup the baseclient and the router for server localhost:9001.
4. call invalidate layout to trigger the prune of router for localhost:9001.
5. The following call with router for localhost:9001 will caught an exception all the time. Even after retry with getBaseclient as the client is already in the map. 
With the fix, the retry will succeed to add a new router.

### DIFF
--- a/it/src/test/java/org/corfudb/universe/scenario/AddNodeDuringInvalidateLayoutIT.java
+++ b/it/src/test/java/org/corfudb/universe/scenario/AddNodeDuringInvalidateLayoutIT.java
@@ -1,0 +1,157 @@
+package org.corfudb.universe.scenario;
+
+import org.corfudb.runtime.clients.BaseClient;
+import org.corfudb.runtime.clients.BaseHandler;
+import org.corfudb.runtime.clients.LayoutHandler;
+import org.corfudb.runtime.clients.ManagementHandler;
+import org.corfudb.runtime.clients.NettyClientRouter;
+import org.corfudb.runtime.exceptions.NetworkException;
+import org.corfudb.runtime.view.Layout;
+import org.corfudb.universe.GenericIntegrationTest;
+import org.corfudb.universe.UniverseManager.UniverseWorkflow;
+import org.corfudb.universe.group.cluster.docker.DockerCorfuCluster;
+import org.corfudb.universe.node.client.LocalCorfuClient;
+import org.corfudb.universe.node.server.CorfuServer;
+import org.corfudb.universe.scenario.fixture.Fixture;
+import org.corfudb.universe.universe.UniverseParams;
+import org.corfudb.util.CFUtils;
+import org.corfudb.util.NodeLocator;
+import org.corfudb.util.Sleep;
+import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+import org.mockito.Mockito;
+
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -912,6 +912,7 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
                         .contains(endpoint.toEndpointUrl()))
                 .forEach(endpoint -> {
                     try {
+                        getLayoutView().getRuntimeLayout().removeClient(BaseClient.class, endpoint.toEndpointUrl());
```
PavelZaytsev, line 915: Why don't we remove other clients too?

test/src/test/java/org/corfudb/integration/ClusterReconfigIT.java
```diff
@@ -1206,4 +1214,67 @@ private void testFailoverWithStreamSubsumedByCheckpoint(boolean trim) throws Exc
...
+            try {
+                log.debug("\n*****retry : " + retry + " clientMap " + runtime.getLayoutView().getRuntimeLayout().getSenderClientMap());
+                BaseClient baseClient = runtime.getLayoutView().getRuntimeLayout()
+                        .getBaseClient("localhost:9001");
+                IClientRouter router = baseClient.getRouter();
+                log.debug("\nrouter " + router + " map " + runtime.getLayoutView().getRuntimeLayout().getSenderClientMap());
+                if (retry == 0) {
+                    runtime.invalidateLayout();
+                }
+                router.sendMessageAndGetCompletable(new CorfuMsg(CorfuMsgType.RESET)
```
PavelZaytsev, line 1260: I think it would be nice to simulate the layout invalidation concurrently somehow while resetting a node. I know that this is very hard to make it in a such a way that it produces the same test output all the time especially with our code structure but I think it will move us closer to the reproduction.

it/src/test/java/org/corfudb/universe/scenario/AddNodeDuringInvalidateLayoutIT.java
```diff
@@ -0,0 +1,157 @@
...
+                        .addClient(new BaseHandler());
+
+                Mockito.doAnswer(invoke -> {
+                    client.getRuntime().invalidateLayout();
+                    return invoke.callRealMethod();
+                }).when(spyRouter).sendMessageAndGetCompletable(Mockito.any(), Mockito.any());
+
+                // Run add node.
+                client.getManagementView()
+                        .addNode(endpoint, numRetries, timeOutDuration, pollDuration);
```
PavelZaytsev, line 147: I fixed this line here in my branch, right now here it does don't really tests the invalidation during any part of the addNode.

PR #2504, opened 2020-04-06
Store metada map in serialized form in the LogUnit Cache.

pankti-m:
Store metadata map in serialized form in the LogUnit Cache.

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
@@ -6,9 +6,12 @@ import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.google.common.annotations.VisibleForTesting;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.infrastructure.LogUnitServer.LogUnitServerConfig;
+import org.corfudb.infrastructure.log.InMemoryStreamLog;
 import org.corfudb.infrastructure.log.StreamLog;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.LogData;
+
+import static org.corfudb.util.MetricsUtils.sizeOf;
 
 /**
  * LogUnit server cache.
@@ -45,7 +48,11 @@ public class LogUnitServerCache {
      */
     private ILogData handleRetrieval(long address) {
         LogData entry = streamLog.read(address);
-        log.trace("handleRetrieval: Retrieved[{} : {}]", address, entry);
+        log.debug("handleRetrieval: Retrieved[{} : {}]", address, entry);
+        if (entry != null && !(streamLog instanceof InMemoryStreamLog)) {
+            //System.out.println("Here2");
+            entry.serializeMetadata();
+        }
         return entry;
     }
 
@@ -68,17 +75,17 @@ public class LogUnitServerCache {
             ILogData ld = dataCache.getIfPresent(address);
             return ld != null ? ld : handleRetrieval(address);
         }
-
         return dataCache.get(address);
     }
 
     /**
-     * Returns the log entry form the cache or retrieves it from the underlying storage.
+     * Returns the log entry from the cache or retrieves it from the underlying storage.
      *
      * @param address the address of the log entry to retrieve
      * @return the log entry read from cache or retrieved the underlying storage
      */
     public ILogData get(long address) {
+        //System.out.println("Here1");
         return get(address, true);
     }
 
@@ -90,7 +97,13 @@ public class LogUnitServerCache {
      * @param entry   the log entry to write
      */
     public void put(long address, ILogData entry) {
-        log.trace("LogUnitServerCache.put: Cache write[{} : {}]", address, entry);
+        log.info("LogUnitServerCache.put: Cache write[{} : {}]", address, entry);
+        if (entry instanceof LogData && !(streamLog instanceof InMemoryStreamLog)) {
+            //System.out.println("In mem metadata deepSizeOf {}" + ((LogData) entry).getDeepSizeOfMetadata());
+            ((LogData)entry).serializeMetadata();
+            //System.out.println("Serialized Metadata numbytes {}" + ((LogData) entry).getSerializedMetadataBytes());
+            //System.out.println("Total Size Estimate - " + entry.getSizeEstimate() + "Total Deep Size - " + sizeOf.deepSizeOf(entry));
+        }
         dataCache.put(address, entry);
     }
 
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -17,6 +17,9 @@ import org.corfudb.protocols.logprotocol.LogEntry;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.WriteSizeException;
 import org.corfudb.util.serializer.Serializers;
+import org.ehcache.sizeof.SizeOf;
+
+import static org.corfudb.util.MetricsUtils.sizeOf;
 
 /**
  * Created by mwei on 8/15/16.
@@ -37,6 +40,10 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     private int lastKnownSize = NOT_KNOWN;
 
     private final transient AtomicReference<Object> payload = new AtomicReference<>();
+
+    private EnumMap<LogUnitMetadataType, Object> metadataMap;
+
+    private byte[] serializedMetadata;
 
     public static LogData getTrimmed(long address) {
         LogData logData = new LogData(DataType.TRIMMED);
@@ -95,7 +102,6 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
                             }
                             value = actualValue == null ? this.payload : actualValue;
                             this.payload.set(value);
-                            lastKnownSize = data.length;
                         } catch (Throwable throwable) {
                             log.error("Exception caught at address {}, {}, {}",
                                     getGlobalAddress(), getStreams(), getType());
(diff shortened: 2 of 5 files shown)


runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -62,6 +62,11 @@ public static LogData getEmpty(long address) {
         return logData;
     }
 
+    @Override
+    public EnumMap<IMetadata.LogUnitMetadataType, Object> getMetadataMap() {
+        return ICorfuPayload.enumMapFromBuffer(serializedMetadataMap, LogUnitMetadataType.class);
```
xnull, line 67: Are we going to deserialize a metadataMap every single time when we need it?
It's gonna be a huge performance impact

WenbinZhu: @pankti-m To solve the oom problem, we only need to serialize metadata before it was put into cache, so I think we need both a serialized and unserialized field for metamap, just like `payload` and `data`.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -129,44 +133,35 @@ public synchronized void acquireBuffer() {
...
         if (lastKnownSize != NOT_KNOWN) {
             return lastKnownSize;
         }
-
         return 1;
     }
 
     @Getter
-    final EnumMap<LogUnitMetadataType, Object> metadataMap;
+    private final ByteBuf serializedMetadataMap = Unpooled.buffer();
```
xiaoqin2012, line 152: I suppose we still need metadataMap that is null at the server side and only get populated at the client/runtime side.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -221,6 +221,7 @@ public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx,
 
         batchWriter.addTask(WRITE, msg)
                 .thenRunAsync(() -> {
+                    logData.serializeMetadata();
```
xiaoqin2012, line 224: This should be in the cache put function. Otherwise it is a bug.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -136,32 +135,33 @@ public synchronized void acquireBuffer() {
...
+        serializedMetadata = Unpooled.buffer();
+        ICorfuPayload.serialize(serializedMetadata, metadataMap);
+        metadataMap = null;
+    }
 
     /**
      * Return the payload.
      */
     public LogData(ByteBuf buf) {
+        lastKnownSize = buf.readableBytes();
```
xiaoqin2012, line 158: should use buf.readerIndex()

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -265,6 +265,7 @@ public void doSerialize(ByteBuf buf) {
     }
 
     void doSerializeInternal(ByteBuf buf) {
+        int startAddress = buf.readerIndex();
```
xiaoqin2012, line 268: should use buf.writerIndex()

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
xiaoqin2012: I did some benchmark with the serialized version. It did save some space. With a 8 streams transaction: 
LogData in the cache (metaDataMap is null):getSizeEstimate 543 deepSize 1568
Same LogData at Runtime (metaDataMap is not null: getSizeEstimate 543 deepSize 4584
As now the LogData at the Cache is a flat data structure, we can still sizeOf.deepSize to get its accurate size. 

Please add a test case and also do a benchmark to see if there are performance issues. You may refer my branch for the testcase : xq/logUnitCache01

pankti-m: ok, will do.  Right now I am hitting errors in existing unit tests

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
pankti-m: This is a fresh buffer, so readableBytes() should be correct?  Let me run a couple of tests...

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
pankti-m: if we see the implementation of buf.readableBytes, it uses writerIndex-readerIndex, which is what I have in the function.  Let me know if I have missed something.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
pankti-m: Done, thanks!

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
xiaoqin2012: It is not guarantee a fresh buffer. As runtime ask for log data to the server, the LogData will be sent in a batch in one message. The message payload buffer contains a batch of LogData.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
xiaoqin2012: I fixed some issues in my branch xq/logUnitCache01. Replace serializedMetadata with byte[]. The problem with ByteBuf is that after eaching reading, the readerIndex will be changed, the further read operations will hit the out of boundary exception. Another issue is that if there are two threads are reading the bytebuf at the same time, it messes the readerIndex.

pankti-m: Thanks!  Yes I replaced it similarly in an earlier patch.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
```diff
@@ -45,7 +46,10 @@ public LogUnitServerCache(LogUnitServerConfig config, StreamLog streamLog) {
      */
     private ILogData handleRetrieval(long address) {
         LogData entry = streamLog.read(address);
-        log.trace("handleRetrieval: Retrieved[{} : {}]", address, entry);
+        log.debug("handleRetrieval: Retrieved[{} : {}]", address, entry);
+        if (entry != null && !(streamLog instanceof InMemoryStreamLog)) {
+            entry.serializeMetadata();
```
xiaoqin2012, line 51: the checking of log type is memory should be removed. As many of our test cases are tested against memory log.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
```diff
@@ -90,7 +93,10 @@ public ILogData get(long address) {
      * @param entry   the log entry to write
      */
     public void put(long address, ILogData entry) {
-        log.trace("LogUnitServerCache.put: Cache write[{} : {}]", address, entry);
+        log.info("LogUnitServerCache.put: Cache write[{} : {}]", address, entry);
+        if (entry instanceof LogData  && !(streamLog instanceof InMemoryStreamLog)) {
+            ((LogData)entry).serializeMetadata();
+        }
```
xiaoqin2012, line 99: Should remove the checking of memory log.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -136,36 +135,60 @@ public synchronized void acquireBuffer() {
...
+
+    public void serializeMetadata() {
+        if (serializedMetadata != null) {
+            return;
+        }
+        ByteBuf serializedBuf = Unpooled.buffer();
+        //System.out.println("Metadata map: " + metadataMap);
+        ICorfuPayload.serialize(serializedBuf, metadataMap);
+        serializedMetadata = byteArrayFromBuf(serializedBuf);
+        metadataMap = null;
```
xiaoqin2012, line 156: Look into byteArrayFromBuf, it is creating another buf. As the serializedBuf is fresh, you can just do : byte[] serializedMetadata = new byte[.readableBytes()];

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -281,17 +305,37 @@ void doSerializeInternal(ByteBuf buf) {
...
                 ICorfuPayload.serialize(buf, data);
-                lastKnownSize = data.length;
             }
         }
-
         if (type.isMetadataAware()) {
-            ICorfuPayload.serialize(buf, metadataMap);
+            if (serializedMetadata == null) {
+                int lengthIndex = buf.writerIndex();
+                buf.writeInt(0);
```
xiaoqin2012, line 318: Why does it need write a dummy Int?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -329,7 +373,7 @@ public boolean equals(Object o) {
 
     @Override
     public String toString() {
-        return "LogData[" + getGlobalAddress() + "]";
+        return "LogData["; //+ getGlobalAddress() + "]";
```
xiaoqin2012, line 376: Should change back?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
pankti-m: By doing that, I will have to repeat the same thing done in byteArrayFromBuf - 
serializedMetadata = new byte[serializedBuf.readableBytes()];
serializedBuf.readBytes(serializedMetadata);

If creating  the bytebuf is not an expensive operation, I would like to keep this unchanged.  Let me know your thoughts.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
pankti-m: If we look at how the payload is serialized in this function, it writes the size of payload followed by bytes.  This is to keep parity with it.

test/src/test/java/org/corfudb/integration/CacheTest.java
```diff
@@ -0,0 +1,89 @@
...
+                    .setTypeToken(new TypeToken<CorfuTable<Integer, byte[]>>() {})
+                    .setStreamName("table" + i)
+                    .serializer(Serializers.JAVA)
+                    .open();
+        }
+
+        start = System.currentTimeMillis();
+        for(int i=0; i<iterations; i++) {
+            assertThat(maps_new[i].get(i).equals(smallPayload));
+        }
```
xiaoqin2012, line 81: I would suggest to make the iterations bigger for one map, like 1 million. Also do the same for the put operation.

pankti-m: Changed the test to iterate 10K times, each time writing to 100 tables in 1 transaction.  So the total writes will be    1 million

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
pankti-m: Removed

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
```diff
@@ -90,8 +94,13 @@ public ILogData get(long address) {
...
      */
     public void put(long address, ILogData entry) {
-        log.trace("LogUnitServerCache.put: Cache write[{} : {}]", address, entry);
+        log.info("LogUnitServerCache.put: Cache write[{} : {}]", address, entry);
+        if (entry instanceof LogData) {
+            ((LogData)entry).serializeMetadata();
+            log.info("Size Estimate - " + entry.getSizeEstimate() + "Deep Size - " + sizeOf.deepSizeOf(entry));
+        }
         dataCache.put(address, entry);
+        log.info("Size Estimate After - " + entry.getSizeEstimate() + " Deep Size After - " + sizeOf.deepSizeOf(entry));
```
xiaoqin2012, line 103: This should be trace level?

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -429,7 +429,9 @@ public void checkUnCachedWrites() throws Exception {
         setServer(s1);
 
         ILogData entry = s1.getDataCache().get(globalAddress);
-
+        /*if (entry instanceof LogData) {
+            ((LogData) entry).deserializeMetadata();
+        }*/
         // Verify that the meta data can be read correctly
```
xiaoqin2012, line 435: remove the not used code.

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -429,7 +429,9 @@ public void checkUnCachedWrites() throws Exception {
         setServer(s1);
 
         ILogData entry = s1.getDataCache().get(globalAddress);
-
+        /*if (entry instanceof LogData) {
```
xiaoqin2012, line 432: remove these not used code.

PR #2509, opened 2020-04-17
CorfuStore: retry internal txns on openTable

hisundar:
## Overview

Description:
Consumers of CorfuStore are ok expecting their transactions aborting due to conflict but prefer that internal transactions like the one in openTable() be retried automatically on failure

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
@@ -29,6 +29,7 @@ import org.corfudb.runtime.collections.StreamingMap;
 import org.corfudb.runtime.collections.StreamingMapDecorator;
 import org.corfudb.runtime.collections.Table;
 import org.corfudb.runtime.collections.TableOptions;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.object.ICorfuVersionPolicy;
 import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.util.serializer.ISerializer;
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -178,17 +179,21 @@ void registerTable(@Nonnull String namespace,
...
+            try {
+                this.runtime.getObjectsView().TXBuild().type(TransactionType.OPTIMISTIC).build().begin();
+                if (hasSchemaChanged) {
+                    this.registryTable.put(tableNameKey,
+                            new CorfuRecord<>(tableDescriptors, metadataBuilder.build()));
+                } else {
+                    this.registryTable.putIfAbsent(tableNameKey,
+                            new CorfuRecord<>(tableDescriptors, metadataBuilder.build()));
+                }
+                this.runtime.getObjectsView().TXEnd();
```
xiaoqin2012, line 193: Is it possible that before calling TXEnd(), while doing put, it will throw a transaction abort exception and the tx.end() is never called?

hisundar: No, we can't throw transaction abort exceptions until a transaction ends because only when a transaction ends will we have all the keys whose conflicts need resolution.

xiaoqin2012: Ok. How about other exceptions? If before call TXEnd(), the above code throws other exceptions but not transaction abort exception, the the TXEnd() will not be called, is it safe to start a transaction without a TXEnd() call? Also after 3 retries, if it still catch a transaction abort exception, it should throw it, right?

hisundar: Only TransactionAborted exception is caught and retried which is not thrown before TXEnd() is called as of this codebase.
If there are other exceptions then we will not catch or retry them, it will go up to caller as usual.

xiaoqin2012: Two issues:
1. The original code, always execute TXEnd() as it is in the finally. So with any exception, the TXEnd() will be executed. 
2. In the change, if transactionAborted happens three times after retry, the exception should be thrown again to the caller.

hisundar: 1. Calling TXEnd() on an other exception that occurred before it would be as undefined/wrong behavior as calling TXEnd() in the finally block - (one might end up committing a transaction that was supposed to fail). I suppose it's better to abort the transaction in that case to prevent corruption. Let me make that change instead.
2. Sure, while this will almost never occur in real life because the transaction abort itself is rare, I suppose it should be cleaner to re-throw the exception.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreTest.java
```diff
@@ -11,6 +11,7 @@
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuStoreMetadata;
 import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
```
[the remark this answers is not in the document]
pankti-m: nit - please remove this import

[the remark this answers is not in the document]
hisundar: will be removing this on the next patch.. thanks

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Addressed 1. and 2. thanks

PR #2514, opened 2020-04-18
Include deep size of metadataMap in LogData size calculation for Caffeine

pankti-m:
Include deep size of metadataMap in LogData size calculation for Caffeine.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
@@ -10,6 +10,8 @@ import org.corfudb.infrastructure.log.StreamLog;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.protocols.wireprotocol.LogData;
 
+import static org.corfudb.util.MetricsUtils.sizeOf;
+
 /**
  * LogUnit server cache.
  * <p>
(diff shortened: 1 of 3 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
```diff
@@ -27,7 +29,8 @@
     public LogUnitServerCache(LogUnitServerConfig config, StreamLog streamLog) {
         this.streamLog = streamLog;
         this.dataCache = Caffeine.newBuilder()
-                .<Long, ILogData>weigher((addr, logData) -> logData.getSizeEstimate())
+                .<Long, ILogData>weigher((addr, logData) -> logData.getSizeEstimate() +
```
xiaoqin2012, line 32: Can you replace this with a function call? For now, we suppose there is no performance impact if number of streams in a transaction is less than 10 tables? But if there are more tables, it may decrease the performance. Please put comments to the function and check how many streams in the map, if it is bigger than a threshold like 20, print a warn log.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
```diff
@@ -91,6 +94,7 @@ public ILogData get(long address) {
      */
     public void put(long address, ILogData entry) {
         log.trace("LogUnitServerCache.put: Cache write[{} : {}]", address, entry);
+        //System.out.println("Metadata map deep size - " + sizeOf.deepSizeOf(entry.getMetadataMap()));
```
WenbinZhu, line 97: Remove this.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
WenbinZhu, line 32: Add 8 bytes for the key

pankti-m: done

pankti-m: done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServerCache.java
pankti-m: done

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -19,10 +19,7 @@
 import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
WenbinZhu, line 22: Expand

PR #2518, opened 2020-04-20
Enhance log unit server logging.

WenbinZhu:
## Overview

Enhance log unit server logging.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -142,9 +142,13 @@ public class LogUnitServer extends AbstractServer {
      */
     @ServerHandler(type = CorfuMsgType.TAIL_REQUEST)
     public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.debug("handleTailRequest: received a tail request {}", msg);
+        log.trace("handleTailRequest: received a tail request {}, type: {}", msg.getPayload().getReqType());
         batchWriter.<TailsResponse>addTask(TAILS_QUERY, msg)
-                .thenAccept(tailsResp -> r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp)))
+                .thenAccept(tailsResp -> {
+                    log.debug("handleTailRequest: global tail: {}, number of streams: {}",
+                            tailsResp.getLogTail(), tailsResp.getStreamTails().size());
+                    r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp));
+                })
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -142,9 +142,13 @@ protected void processRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRo
      */
     @ServerHandler(type = CorfuMsgType.TAIL_REQUEST)
     public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.debug("handleTailRequest: received a tail request {}", msg);
+        log.trace("handleTailRequest: received a tail request {}, type: {}", msg.getPayload().getReqType());
         batchWriter.<TailsResponse>addTask(TAILS_QUERY, msg)
-                .thenAccept(tailsResp -> r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp)))
+                .thenAccept(tailsResp -> {
+                    log.debug("handleTailRequest: global tail: {}, number of streams: {}",
```
xnull, line 148: So, essentially we have the same number of log messages flooding the log. I don't see any improvements here.

WenbinZhu: The perpose of this patch is not to reduce logs, which you already did in a earlier patch. This is what we discussed earlier to replace initial logs with more meaning messages which include the query results.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,7 +183,9 @@ public void handleLogAddressSpaceRequest(CorfuMsg msg, ChannelHandlerContext ctx
     @ServerHandler(type = CorfuMsgType.TRIM_MARK_REQUEST)
     public void handleTrimMarkRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
         log.trace("handleTrimMarkRequest: received a trim mark request {}", msg);
-        r.sendResponse(ctx, msg, CorfuMsgType.TRIM_MARK_RESPONSE.payloadMsg(streamLog.getTrimMark()));
+        long trimMark = streamLog.getTrimMark();
+        log.debug("handleTrimMarkRequest: trim mark response: {}", trimMark);
```
annym, line 187: @WenbinZhu won't this message in debug add too much logging? which does not add much, perhaps trace would be more suitable for this one?

WenbinZhu: Initially the first log line was in debug mode until Salva's patch changed it to trace, this patch keeps the same result as before: one debug log line for this method.  Also trim mark request is not quite offen right?

PR #2524, opened 2020-04-24
State Transfer Part 1 Fixes

### LINKED ISSUE #2521: RestoreRedundancyMergeSegments action is unable to make progress after OutrankedException
## Overview

If the seal of the servers occurred but the new layout is not yet committed during the RestoreRedundancyMergeSegments action, there is a chance of task not making any progress during the retries. This showcased in the following scenario: 

- There are nodes A, B and C. Nodes B, C belong to the cluster and node A does not.
- We are adding node A. Node A, B and C are at epoch 3. 
- Node A performs a state transfer and it finishes ok.
- Node A now wants to seal every server in the system (move every server in the system to epoch 4). 
- It needs only a quorum of seal responses to continue. Let's say it asynchronously seals servers A, B and C. B and C seal immediately and send the responses to A. A receives 2 out of 3 responses and continues. 
- Node A is now at epoch 3, Node B is at epoch 4 and Node C is at epoch 4.
- Node A forms a new layout (at epoch 4) where it is a part of a cluster, and now tries to attempt consensus on this layout. - Node A sends prepare request with a rank 1 to servers A, B and C. It needs only a quorum of prepare responses to continue the Paxos. 
- Node B at epoch 4 (recall that it was previously sealed) receives a prepare request at epoch 4. Sends a response -> ok.
- Node C at epoch 4 (recall that it was previously sealed) receives a prepare request at epoch 4. Sends a response -> ok.
- Node A at epoch 3 (it is not sealed yet, that's where the race is happening) is at epoch 3. Rejects a prepare request with a wrong epoch exception.  
- Node A continues with the second phase of Paxos - propose requests to all the nodes A, B and C since now it got the majority of prepare responses.
- Node B at epoch 4 (recall that it was previously sealed) receives a propose request at epoch 4. Sends a response -> ok.
- Node C at epoch 4 (recall that it was previously sealed) receives a propose request at epoch 4. Sends a response -> ok.
- Node A receives a propose request at epoch 4 (now it's sealed). It does not have a phase 1 rank but it got a propose request, so it throws an OuntrankedException. 
- Paxos fails.
- The cluster is in this state: 
  - Node A, B and C all have the layouts at epoch 3.
  - Node A, B and C are all sealed with epoch 4 and their routers are at epoch 4 too.
  - Primary sequencer is not bootstrapped. 
- Node A retries the entire workflow. One of the first steps is to retrieve a trim mark. Node A keeps on failing with a wrong epoch exception because it's runtime and the layout of every node is at epoch 3, but every server router is at epoch 4. 

TLDR: The issue arises because
1) The node that drives the RestoreRedundancyMergeSegments action is not guaranteed to be sealed before attemptConsensus.
2) A datapath logic and the layout reconfiguration logic are retried together.

### LINKED ISSUE #2523: Retries for writeRecords during the state transfer might result in IOOBE
## Overview

- When we calculate a difference between the known addresses and the remaining addresses, there is a possibility of IndexOutOfBoundsException.

- For writeRecords method we need to retry only for the specific exceptions. If there are other exceptions, for example, WrongEpochException or IllegalArgumentException, we need to fail faster.

PavelZaytsev:
## Overview

Description:
Bug fixes for the latest state transfer:
1. Unsealed local server during Paxos results in liveness Issue.

The node that drives the RestoreRedundancyMergeSegments action is not guaranteed to be sealed before attemptConsensus, resulting in a partial seal and RestoreRedundancyAndMergeSegment action getting stuck due to the WrongEpochExceptions.
(Please refer to #2523).

This is now fixed by decoupling the retries for the state transfer and the merge segments logic as well as invoking the seal on the local node before attempting the merge.  

2. IndexOutOfBounds during state transfer writes.

When we calculate a difference between the known addresses and the remaining addresses, there is a possibility of IndexOutOfBoundsException. 
(Please refer to #2521).

This is now fixed by copying all the required range to the immutable list at the beginning and then calculating the delta between this list and the known addresses in the range.

Why should this be merged: 
Bug fixes, improvements.

Related issue(s) (if applicable): 
Fixes #2521
Fixes #2523

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
@@ -7,6 +7,10 @@ import org.corfudb.infrastructure.log.statetransfer.batch.TransferBatchRequest;
 import org.corfudb.infrastructure.log.statetransfer.batch.TransferBatchResponse;
 import org.corfudb.protocols.wireprotocol.LogData;
 import org.corfudb.runtime.clients.LogUnitClient;
+import org.corfudb.runtime.exceptions.NetworkException;
+import org.corfudb.runtime.exceptions.OverwriteException;
+import org.corfudb.runtime.exceptions.RetryExhaustedException;
+import org.corfudb.util.CFUtils;
 import org.corfudb.util.Sleep;
 
 import java.time.Duration;
@@ -15,6 +19,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
 
 import static org.corfudb.infrastructure.log.statetransfer.batch.TransferBatchResponse.TransferStatus.SUCCEEDED;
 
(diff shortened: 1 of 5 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/statetransfer/batchprocessor/StateTransferBatchProcessor.java
```diff
@@ -43,54 +48,69 @@
      */
     default TransferBatchResponse writeRecords(ReadBatch readBatch, LogUnitClient logUnitClient,
                                                int writeRetriesAllowed, Duration writeSleepDuration) {
-        List<Long> totalAddressesInRequest = readBatch.getAddresses();
-        Optional<String> destination = readBatch.getDestination();
+        if (readBatch == null || readBatch.getAddresses().isEmpty()) {
+            throw new IllegalArgumentException("The read batch is empty.");
```
xnull, line 52: why we throw an exception if there is an empty list of addresses? Essentially, having an empty request is not an error, it should be ok just do nothing?

PR #2530, opened 2020-05-04
CorfuStore: Lightweight Streaming Layer

hisundar:
Streaming needs to work in JVMs that have very little heap space.
In such processes use of a separate txPoller with buffering of
updates creates an overhead that causes the JVM to go OOM.
This change simplifies streaming layer avoids & avoids buffering.
Each subscriber gets a separate tx stream and thread whose fate is just bound to its calls.
A sanity limit of 8 subscribers means for advanced use cases of several table subscriptions
it would be up to the callers to work together to prevent system hog.
The stream layer will not assume upper layer needs.

## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc


hisundar:
> I think this is design change and we should discuss this as a team. We might have future use cases where this model will open way too many threads or end up being restrictive. Also an easy quick fix to try would be to limit the number of entries in the processing queue.

Sure, because as of now, use of a single transaction poller and separate thread for subscription is broken as demonstrated by some verticals who have low heap space. This will be true even after we make the TRANSACTION_STREAM more selective, because the core issue is holding the backlog itself can up being too costly for some JVMs. Also once we break up TRANSACTION_STREAM, use of a single transaction poller won't work anyway. So the goal of this PR is to start with the minimal simplest way to get this working without breaking anyone and add complexity if and only if required.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Subscribe to updates.
+     *
+     * @param streamListener   Client listener.
+     * @param namespace        Namespace of interest.
+     * @param tablesOfInterest Only updates from these tables will be returned.
+     * @param startAddress     Address to start the notifications from.
+     */
+    public synchronized <K extends Message, V extends Message, M extends Message>
+    void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
```
xnull, line 64: Since you implement pub-sub, it would be great if you could use https://www.reactive-streams.org/ instead of self-written protocols.
reactive-streams is a part of Java 9, so when we migrate to newer versions of java it can be used natively

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+            throw new SubscriberAlreadyExistsException(
+                    "subscribe: Max "+MAX_SUBSCRIBERS+" hit! Consider consolidating "+streamListener);
+        }
+        if (subscriptions.containsKey(streamListener)) {
+            throw new SubscriberAlreadyExistsException(
+                    "StreamManager:subscrbie already registered "+streamListener+". Maybe call unregister");
+        }
+        log.info("StreamManager::subscribe {}, startAddress {}, namespace {}, tables {}", streamListener,
+                startAddress, namespace, tablesOfInterest.toString());
+        SubscriptionThread st = new SubscriptionThread(this, streamListener,
```
xnull, line 76: Creating a new thread could be an expensive operation, probably reactive-streams already solved this problem internally in an efficient way

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+import org.corfudb.runtime.view.stream.IStreamView;
+
+/**
+ *
+ * Created by hisundar on 04/28/2020.
+ */
+@Slf4j
+public class StreamManager {
+
+    private final int MAX_SUBSCRIBERS = 12; // Arbitrary sanity check value.
```
xnull, line 35: it should be static in the other case make it non-caps

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+        if (subscriptions.containsKey(streamListener)) {
+            throw new SubscriberAlreadyExistsException(
+                    "StreamManager:subscrbie already registered "+streamListener+". Maybe call unregister");
+        }
+        log.info("StreamManager::subscribe {}, startAddress {}, namespace {}, tables {}", streamListener,
+                startAddress, namespace, tablesOfInterest.toString());
+        SubscriptionThread st = new SubscriptionThread(this, streamListener,
+                namespace, tablesOfInterest, startAddress);
+        subscriptions.put(streamListener, st);
```
xnull, line 78: could be race condition here, please use ConcurrentHashMap

xnull: Oh I see, you use synchronized. But what's the point for that if we can use ConcurrentHashMap? Also I suspect all these things already implemented and tested in reactive-streams

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+
+    /**
+     * Unsubscribe a prior subscription.
+     *
+     * @param streamListener Client listener.
+     */
+    public synchronized void unsubscribe(@Nonnull StreamListener streamListener) {
+        SubscriptionThread st = subscriptions.remove(streamListener);
+        if (st == null) {
+            log.warn("StreamManager::unsubscribe has no context for "+streamListener.toString());
```
xnull, line 90: log.warn("StreamManager::unsubscribe has no context for {}", streamListener);

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+            return;
+        }
+        st.setShutdown(true);
+        st.interrupt();
+        log.info("Unsubscribed StreamListener {}", streamListener.toString());
+    }
+
+    class SubscriptionThread extends Thread {
```
xnull, line 98: I'm concerned, probably we gonna create too many threads

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+                TableSchema<? extends Message, ? extends Message, ? extends Message>> tablesOfInterest;
+
+        /**
+         * Starting address.
+         */
+        private final long startAddress;
+
+        @Getter
+        @Setter
+        private volatile boolean shutdown = false;
```
xnull, line 132: Should be AtomicBoolean

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
PavelZaytsev, line 90: NIT: When concating the strings I think its better to have spaces around `+`

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+    public synchronized <K extends Message, V extends Message, M extends Message>
+    void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                   @Nonnull List<TableSchema> tablesOfInterest, long startAddress) {
+        if (subscriptions.size() >= MAX_SUBSCRIBERS) {
+            throw new SubscriberAlreadyExistsException(
+                    "subscribe: Max "+MAX_SUBSCRIBERS+" hit! Consider consolidating "+streamListener);
+        }
+        if (subscriptions.containsKey(streamListener)) {
+            throw new SubscriberAlreadyExistsException(
+                    "StreamManager:subscrbie already registered "+streamListener+". Maybe call unregister");
```
WenbinZhu, line 72: The exception name makes sense for the second case, but seems confusing for the first case.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+                txnStream.seek(startAddress + 1);
+            } catch (Throwable throwable) {
+                log.warn("{}:onError at seek to {}", listener.toString(), startAddress + 1);
+                listener.onError(throwable);
+                streamManager.unsubscribe(listener);
+                return;
+            }
+
+            while (!isShutdown()) {
+                try {
```
WenbinZhu, line 152: I think it's cleaner to put the try catch block outside the while loop, because once it hits exception, it is stopped. And preferrably also combine the the first try catch block for txnStream.seek(), they seem to be the same.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+                    while (!txnStream.hasNext()) {
+                        log.debug("{}::run() all caught up. Taking a 50ms nap", listener.toString());
```
WenbinZhu, line 154: This should be trace right?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: We have a cap on it at MAX_SUBSCRIBERS currently set to an arbitrary value of 12.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: Threads are not going to be created and destroyed often, it should be just once per lifecycle of the process.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: sure - we would be asked to move to Java 11 shortly anyway, but as of Java 8 this functionality is not available. Also this gives us fine grained control over how the streams should behave.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+     * @param namespace        Namespace of interest.
+     * @param tablesOfInterest Only updates from these tables will be returned.
+     * @param startAddress     Address to start the notifications from.
+     */
+    public synchronized <K extends Message, V extends Message, M extends Message>
+    void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                   @Nonnull List<TableSchema> tablesOfInterest, long startAddress) {
+        if (subscriptions.size() >= MAX_SUBSCRIBERS) {
+            throw new SubscriberAlreadyExistsException(
+                    "subscribe: Max "+MAX_SUBSCRIBERS+" hit! Consider consolidating "+streamListener);
```
pankti-m, line 68: nit - space after + signs

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
pankti-m: also, typo in 'subscriber'

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,219 @@
...
+                }
+            }
+        }
+
+        public SubscriptionThread(@Nonnull StreamManager streamManager,
+                                  @Nonnull StreamListener listener,
+                                  @Nonnull String namespace,
+                                  @Nonnull List<TableSchema> tableSchemas,
+                                  long startAddress) {
+            super("CorfuSubscriber: "+namespace+":"+listener.toString());
```
pankti-m, line 196: nit: spaces around '+'

pankti-m: also isnt it more clear this way? "corfusubscriber: listener.toString() + namesapace"?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: subscribe() unsubscribe() where the synchronized is used are lifecycle methods, they won't be called often. It's more like once per lifecycle of the process so concurrency is not required, the goal of these concurrency primitives is just correctness.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: Also we are talking about 1 thread per business vertical so number of threads is currently just one. However one can argue that even if there are multiple threads since they correspond to actual business verticals they should be fully justified.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -173,7 +170,7 @@ public Query query(@Nonnull final String namespace) {
     void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
                    @Nonnull List<TableSchema> tablesOfInterest,
                    @Nullable Timestamp timestamp) {
-        txnStreamingManager.subscribe(streamListener, namespace, tablesOfInterest,
+        streamManager.subscribe(streamListener, namespace, tablesOfInterest,
                 (timestamp == null) ? getTimestamp().getSequence() : timestamp.getSequence());
```
pankti-m, line 174: is the client expected to know the start address of the stream of interest?  Is it possible for the client to have this info in a clean way?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: yes, just wanted to prevent adding another exception since each exception raises more questions that need explanation and answering. Thought the message in the exception makes the difference right?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. I put it in debug because it was helpful for debugging this on a live setup.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: I will put the try catch outside the while, but can't combine it with the first.
The first exception is distinctly different. It happens when a subscriber is trying to start from a trimmed address which is different from a subscriber hitting a trimmed exception during normal operation.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: > Looks like this is related to the stream tagging feature? I see that the StreamManager uses the txn stream but filters out based on the tables of interest. Do you plan to change it once stream tagging is available?

Yes as soon as your change to have stream tagging is merged we just have to modify the TRANSACTION_STREAM to that specific stream tag. Which is exactly why this change is needed in the first place. In the previous (current) implementation there is just one publisher  which means we can't have selective polling.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: fixed, thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: yes it is, fixed, thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
WenbinZhu: What I meant was changing the name of exception to a more genral one and the messages in it can be different.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
WenbinZhu: Actually seek() should not raise any exception as it just moves the pointer, but it's ok to be defensive.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: CorfuStore exposes an api called getTimestamp() which abstracts the internal start address of the stream. Please check out Timestamp in CorfuStoreMetadata.

hisundar:
> Looks like this is related to the stream tagging feature? I see that the StreamManager uses the txn stream but filters out based on the tables of interest. Do you plan to change it once stream tagging is available?

Yes as soon as your change to have stream tagging is merged we just have to modify the TRANSACTION_STREAM to that specific stream tag. Which is exactly why this change is needed in the first place. In the previous (current) implementation there is just one publisher which means we can't have selective polling.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,227 @@
+package org.corfudb.runtime.collections;
+
+import com.google.protobuf.Message;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+import java.sql.Time;
```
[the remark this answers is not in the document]
xnull: btw never use java.sql.Time

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
xnull: you can add a dependency when we migrate to the new version of java we just get rid of that  dependency.
Also you write a lot of code which makes migration to reactive streams more and more complicated over time, we would end up with our custom version, now is the best time to migrate, it's cheap and save and takes couple of hours of work

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,227 @@
...
+                listener.onError(throwable);
+                streamManager.unsubscribe(listener);
+                return;
+            }
+
+            try {
+                while (!shutdown.get()) {
+                    while (!txnStream.hasNext()) {
+                        log.trace("{}::run() all caught up. Taking a 50ms nap", listener.toString());
+                        int DEFAULT_TIMEOUT_IN_MILLI = 50;
```
xnull, line 159: please rename the variable, it is not a constant

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: understood, sure will do that. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,227 @@
...
+                        long onNextStart = System.nanoTime();
+                        listener.onNext(callbackResult);
+                        long onNextEnd = System.nanoTime();
+                        if (TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart) > 0) {
+                            log.debug("{}::onNext took {}s", listener.toString(),
+                                    TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart));
+                        }
+                    }
+                }
+            } catch (Throwable throwable) {
```
xnull, line 190: looks like there is a problem here, you do not handle InterruptedException it means there could be problems when other threads will try to stop this thread. Intellij IDEA should makes this line yellow with the warning/explanation.

P.S. I know you execute interrupt() in `unsubscribe` method. I can't require you or claim but the way how this while cycle is written seems fragile to me

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,227 @@
...
+                log.warn("{}::onError: {}", listener.toString(), throwable.toString());
+                listener.onError(throwable);
+                streamManager.unsubscribe(listener);
+                return;
```
xnull, line 194: no need for return

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,227 @@
...
+            }
+        }
+
+        public SubscriptionThread(@Nonnull StreamManager streamManager,
+                                  @Nonnull StreamListener listener,
+                                  @Nonnull String namespace,
+                                  @Nonnull List<TableSchema> tableSchemas,
+                                  long startAddress) {
+            super("CorfuSubscriber: " + listener.toString() + ":" + namespace);
```
xnull, line 203: please use String.format

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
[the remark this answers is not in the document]
hisundar: removed, thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: removed, thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: Well this implementation is 600 lines of lesser code compared to previous (existing). It's just one file, so can't think of an easier way to migrate to RxJava if we should decide to use it in the future. 

Also its not just dependency, RxJava internally uses a thread pool based on a publisher subscriber model. The use of separate publisher + subscriber is the root cause of the very bug which prompted this change, so we have to be careful about adding back a publisher-subscriber based model. For now, it makes sense to have something that works and more importantly does not break anything.

xnull: I'm not talking about RxJva, only reactive-streams, it's much smaller

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: StreamingIT.java has a test case that covers unsubscribe() and validate that it does not receive any future updates. That test seems to be working as expected.
But if you have an alternate better way please suggest. For now I am just relying on the fact that InterruptedException is also a type of Throwable
https://docs.oracle.com/javase/7/docs/api/java/lang/InterruptedException.html

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
xnull: I would vote for `log.warn("StreamManager::unsubscribe has no context for {}", streamListener);`

At least it won't create an additional instance of a String, and appears prettier

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,225 @@
...
+
+    /**
+     * Subscribe to updates.
+     *
+     * @param streamListener   Client listener.
+     * @param namespace        Namespace of interest.
+     * @param tablesOfInterest Only updates from these tables will be returned.
+     * @param startAddress     Address to start the notifications from.
+     */
+    public synchronized <K extends Message, V extends Message, M extends Message>
```
xnull, line 66: It would be great if you fix all yellows in the code, some of them could indicate about fragility in the code some bout bugs, thanks

image

xnull: looks like you forgot to add K,V,T into TableSchema signature

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,225 @@
...
+                streamManager.unsubscribe(listener);
+            }
+        }
+
+        public SubscriptionThread(@Nonnull StreamManager streamManager,
+                                  @Nonnull StreamListener listener,
+                                  @Nonnull String namespace,
+                                  @Nonnull List<TableSchema> tableSchemas,
+                                  long startAddress) {
+            super(String.format("CorfuSubscriber: {}:{}", listener.toString(), namespace));
```
xnull, line 201: Sorry my bad, the format has to be
`super(String.format("CorfuSubscriber: %s:%s", listener, namespace));`

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,225 @@
...
+                txnStream.seek(startAddress + 1);
+            } catch (Throwable throwable) {
+                log.warn("{}:onError at seek to {}", listener.toString(), startAddress + 1);
+                listener.onError(throwable);
+                streamManager.unsubscribe(listener);
+                return;
+            }
+
+            try {
+                while (!shutdown.get()) {
```
xnull, line 155: If I understand it properly what you are trying to do is to have scheduler which will run your task periodically.
You can use java scheduler and it simplifies code. You will not need to check `shutdown` variable, no need for thread.sleep, catching exceptions to stop the thread, and so on.

Please take a look https://github.com/CorfuDB/CorfuDB/compare/simplifyStreaming...simplifyStreaming-scheduler

xnull: Also you can have as many threads as you need (scheduler threadfctory)

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: Your implementation seems to be creating one thread factory in every thread. Why?

hisundar: Please see https://stackoverflow.com/a/50077162/4513389 for why this is an overkill for this use case.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
xnull: It's just a dirty example, you can put scheduler and factory into places where the should be

xnull: > Please see https://stackoverflow.com/a/50077162/4513389

A very weird explanation, not sure what that guy tried to say :)

hisundar: scheduler type model with a thread pool is exactly what we have currently (please check out the code that was removed by this PR) - the problem is when you have varying rates of consumption we have multiple tasks executing in the same thread and the heap (from thread local data) grows too quickly for JVMs with low heap space to handle. This is the very bug which prompted this simplification in the first place.

Adding back a scheduler and factory model where multiple tasks are going to execute on the same thread will simply bring back the very problem we are trying to fix.

xnull: I don't believe that scheduler could be a problem in this case. I don't see how it could be possible in java when "the heap (from thread local data) grows too quickly" it sounds a bit odd, also if you just replace scheduler by threads I'm pretty sure it doesn't improve anything just by definition it can't improve performance/resource utilization/consumption.
So, it's up to you, but replacing scheduler by threads not a good idea imo (usually threads gett replaced by scheduler)

hisundar: It's not a problem with scheduler, but a problem with having multiple tasks executing on the same thread pulling different set of log data objects from stream layer.
So to avoid that we really need only have one task running on one thread. Given this requirement a scheduler is really pointless which is what the stackoverflow link above is trying to explain. Once we move to stream tagged selective streaming we will now be able to selectively pull items, at which point it might make sense to consider multiple tasks on a thread or thread factory or pools or even the reactive-streams framework if it makes sense. Technologies have to be applied to serve use cases not the other way around right?

xiaoqin2012:
The current transaction poller polls all the outstanding log in one operation that could use a lot of memory if the poller is slow. The solution you propose still has the similar problem if there are a lot of transactions for a particular stream. 

How about letting the poller polls the transaction log in a batch motion, like polling next 200 for each operation. It can be easily implemented by passing a parameter while calling streamUpto(current offset + batchsize).

hisundar:
> The current transaction poller polls all the outstanding log in one operation that could use a lot of memory if the poller is slow. The solution you propose still has the similar problem if there are a lot of transactions for a particular stream.
> 
Are you referring to existing code or this PR? In this PR there is no transaction poller that pulls all outstanding operations. We have the subscriber thread directly pulling from the transaction stream, one item at a time not in batches.

> How about letting the poller polls the transaction log in a batch motion, like polling next 200 for each operation. It can be easily implemented by passing a parameter while calling streamUpto(current offset + batchsize).

This is exactly what the existing code did before this PR. Please see the files that are deleted by this PR. The problem with it is when the batch has too many outstanding items the JVM goes OOM. That said once we have selective stream tagging we will be able to revisit a similar concept.

runtime/src/main/java/org/corfudb/runtime/collections/TransactionPoller.java
```diff
@@ -1,111 +0,0 @@
...
-    private void pollTxStream() {
- 
-        long lastReadAddress = streamContexts.get(0).getLastReadAddress();
-
-        log.trace("Seeking txStream to {}", lastReadAddress + 1);
-        txnStream.seek(lastReadAddress + 1);
-        log.trace("txStream current global position after seeking {}, hasNext {}",
-                txnStream.getCurrentGlobalPosition(), txnStream.hasNext());
-
-        List<ILogData> updates = txnStream.remaining();
```
xiaoqin2012, line 91: Here we call remaining(). that could cause the high memory usage. Instead of call remaining(), we can call remainUpto(offset + 40).

We can do something like this:
pollTxStream() {
   n = 0;
  while {
    n = polltxstream(upto);
} do (n > 0)

hisundar: remainingUpto() was the first thing I tried when trying to fix the OOM issue before trying this rewrite.
The problem is when you have a publisher subscriber model even if you poll 40 items ahead of current cursor and if only one of those 40 is needed while 39 are not, that whole batch cannot be garbage collected. Since we have one big TRANSACTION_STREAM to which the whole world tags transactions any attempt at batching hits this problem. Which is why to avoid the issue we just go one at a time so we are sure that nothing is held back.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
xnull: Let's clarify.
Invariants:
 - a thread can execute one task at a time, it's not possible to execute many tasks by one thread at a time
 - your task must not keep any state in the thread, you shouldn't use ThreadLocal variables if you want to use a common pool of threads.

If I understand it properly, we have some ThreadLocal variables which bring side effects?
If that true, then it's poor design/implementation in Corfu and we are trying to hack it here and make it work.

If I'm wrong then ok, but I don't see how using a scheduler can be in any way worse than using threads. I would prefer to fix a problem in the design than replace a better solution by a more complex and fragile solution

hisundar: Here is a simpler question - why do you need a thread scheduler/factory if a thread only needs to ever execute one task?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: reactive-streams is also based on a publisher-subscriber model with functionality for backpressure both of which are anti-patterns given the current use case where we have JVMs with low heap space needing to selectively poll. In the current form where we still have a big global TRANSACTION_STREAM anything other than simple threads processing one mutation at a time will hit issues. 

Now that is not to say reactive-streams should not be used. We need the stream tagging feature first. That will allow us to be selective in streaming at which point we would be able to have better control over the subscriptions and we can avoid OOMs. Until then we do need fine grained control over how stream listeners work. Using direct threads achieves just that in probably the simplest way possible.

hisundar:
> Even I still disagree with the approach, the code seems good,

Your comments are important and will be addressed once stream tagging feature is merged.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,224 @@
...
+ * A simple thread based subscription engine where each subscriber or listener gets
+ * a thread which will listen on the tables of interest to it and moves at a rate
+ * which is the same as the consumption.
+ *
+ * Created by hisundar on 04/28/2020.
+ */
+@Slf4j
+public class StreamManager {
+
+    private static final int MAX_SUBSCRIBERS = 12; // Arbitrary sanity check value.
```
Maithem, line 38: I think its cleaner to pass a config type that has this as a default.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,224 @@
...
+            throw new StreamSubscriptionException(
+                    "StreamManager:subscriber already registered "
+                            + streamListener + ". Maybe call unregister first?");
+        }
+        log.info("StreamManager::subscribe {}, startAddress {}, namespace {}, tables {}",
+                streamListener, startAddress, namespace, tablesOfInterest.toString());
+        SubscriptionThread st = new SubscriptionThread(this, streamListener,
+                namespace, tablesOfInterest, startAddress);
+        subscriptions.put(streamListener, st);
+        st.start();
```
Maithem, line 84: These are not deamon threads, they can hang the JVM on exit.

Maithem, line 84: Please name these threads accordingly. Otherwise, its can be hard to profile/debug.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,224 @@
...
+            ));
+
+            this.startAddress = startAddress;
+            StreamOptions options = StreamOptions.builder()
+                    .cacheEntries(false)
+                    .build();
+
+            this.txnStream = runtime.getStreamsView()
+                    .get(ObjectsView.TRANSACTION_STREAM_ID, options);
+        }
```
Maithem, line 222: On unsubscribe the stream is not removed from the StreamsView. A series of subscribe/unsubscribe can create a stream leak.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,224 @@
...
+            } catch (Throwable throwable) {
+                log.warn("{}:onError at seek to {}", listener.toString(), startAddress + 1);
+                listener.onError(throwable);
+                streamManager.unsubscribe(listener);
+                return;
+            }
+
+            try {
+                while (!shutdown.get()) {
+                    while (!txnStream.hasNext()) {
```
Maithem, line 156: I wouldn't use this API. It's not used and might be buggy. I recommend that you just try to sync anyways. If there are no deltas, the callbacks won't trigger.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,224 @@
...
+                        if (TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart) > 0) {
+                            log.debug("{}::onNext took {}s", listener.toString(),
+                                    TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart));
+                        }
+                    }
+                }
+            } catch (Throwable throwable) {
+                log.warn("{}::onError: {}", listener.toString(), throwable.toString());
+                listener.onError(throwable);
+                streamManager.unsubscribe(listener);
```
Maithem, line 192: This will eventually call interrupt on its own self. After the interrupt there is a log line. Once a thread is interrupted it's possible that the subsequent io call fails with ClosedByInterruptException or similar interrupted exceptions. 

Anyways, it seems like the unsubscribe was designed to be called from a different thread, not the thread itself.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,224 @@
...
+        if (st == null) {
+            log.warn("StreamManager::unsubscribe has no context for {}", streamListener.toString());
+            return;
+        }
+        st.shutdown.set(true);
+        st.interrupt();
+        log.info("Unsubscribed StreamListener {}", streamListener.toString());
+    }
+
+    class SubscriptionThread extends Thread {
```
Maithem, line 103: Spawning a thread per subscriber seems to be too expensive for what's being achieved. 
I think its better to multiplex these over an executor. The polling logic and life-cycle management is also cleaner with a scheduler executor. You can submit a periodic job that syncs the subscription streams, that will give you a task future that can be canceled. I think that's cleaner than managing the life cycle of the thread yourself. 

I think the code already has some bugs related to the thread life cycle management, so maybe consider just using a scheduled executor.

runtime/src/main/java/org/corfudb/runtime/exceptions/StreamSubscriptionException.java
```diff
@@ -0,0 +1,12 @@
+package org.corfudb.runtime.exceptions;
+
+/**
+ * CorfuStore only allows one subscriber for a given namespace.
+ * If an attempt is made to register another subscriber without terminating or unsubscribing
+ * the existing subscriber on that namespace this exception will be thrown.
+ */
+public class StreamSubscriptionException extends RuntimeException {
```
Maithem, line 8: Instead of extending from RuntimeException I think we should introduce CorfuClientExceptions and extend from that. Right now our API with the application is not well-defined and that would be the right step forward.

Maithem:
The IF seems to be bricked. Its failing on all builds.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,224 @@
+package org.corfudb.runtime.collections;
```
Maithem, line 1: Can you also implement a close that runtime calls on shutdown ?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: a RuntimeParameter? 
ok will do that.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: making them daemon threads create other problems, it will ends up masking issues where the vertical subscription thread which should be shutdown before JVM shutdown aren't told to exit. we saw this problem with a few consumers.

hisundar: will do this. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: thanks will do this.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: ok - will choose to sleep if there are no deltas. btw we should consider either fixing this buggy api or removing it 
I see it used in other places as well.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: this was the same comment by @xnull and while it is valid - the plan was to specifically limit this to one thread per subscriber for 
1. keeps the memory taken by a subscriber completely local to a thread
2. limits the number of threads/subscribers very explicitly

but the pool idea is definitely valid, I had planned to use it after we have stream tagging because that will change the threading model again.

runtime/src/main/java/org/corfudb/runtime/exceptions/StreamSubscriptionException.java
hisundar: ok - will explore this. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: currently runtime has no visibility into CorfuStore, since CorfuStore runs above.
need to figure out a way to tie those two together. if you have any ideas on this, please share.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
Maithem: Maybe just pass it through the constructor. I think we need to reduce the CorfuRuntimeParamters configs significantly.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
Maithem: The streamManager should have some shutdown/close logic where the threads are stopped properly and an exit callback is fired for the subscribers. Not exposing a close method and not managing the life-cycle inside the runtime is kind of problematic. Its not clear whose responsible for releasing those resources.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
Maithem: `AbstractContextStreamView::hasNext` is actually not being used by the VLO, or internal streaming that's why I'm recommending not to use it, we just dont know it it works. Maybe you can just remove it from the stream API itself.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
Maithem: I see, but we still need a close method, so that the consumer can explicitly call close if not the our runtime.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
Maithem: A scheduled executor will achieve point 1 and 2. I just don't see a strong reason for not using a scheduled executor. 

Might be missing something, but I don't see this as an intermediate step and I don't see it's relationship with stream tagging.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: just realized what you are describing as a problem is exactly the solution we want - we DO want to hang the JVM on exit otherwise corfu becomes responsible for an application level bug where the caller did not cleanly unsubscribe() and shutdown the subscription threads.
Callers are supposed to explicitly detect JVM shutdown and invoke unsubscribe() 
unsubscribe() is the shutdown/close logic.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: fixed.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: tied up the runtime.close() to shutdown the executor pool.

hisundar:
> Even I still disagree with the approach, the code seems good,

Just an fyi - All your comments are addressed in the same way you suggested in latest change. thanks

hisundar:
> There are a couple of bugs in this patch, please address them. Otherwise, it looks good.

All your comments are addressed in latest patch. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+            tableSchemas.forEach(tableSchema -> tablesOfInterest.put(
+                    CorfuRuntime.getStreamID(
+                            TableRegistry.getFullyQualifiedTableName(namespace,
+                                    tableSchema.getTableName())
+                    ), tableSchema
+            ));
+
+            this.startAddress = startAddress;
+            StreamOptions options = StreamOptions.builder()
+                    .cacheEntries(!runtime.getParameters().isCacheDisabled())
```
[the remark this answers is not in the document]
hisundar: sounds good. will do.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+                    if (!entries.isEmpty()) {
+                        CorfuStreamEntries callbackResult = new CorfuStreamEntries(entries);
+                        log.debug("{}::onNext with {} updates", listener.toString(), entries.size());
+                        long onNextStart = System.nanoTime();
+                        listener.onNext(callbackResult);
+                        long onNextEnd = System.nanoTime();
+                        if (TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart) > 0) {
+                            log.debug("{}::onNext took {}s", listener.toString(),
+                                    TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart));
+                        }
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+                    }
+                } catch (Throwable throwable) {
+                    log.warn("{}::onError: {}", listener.toString(), throwable.toString());
+                    listener.onError(throwable);
+                    streamManager.unsubscribe(listener);
+                    break;
+                }
+            } while (logData != null);
```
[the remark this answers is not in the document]
hisundar: true, just wanted a conditional loop instead of an infinite loop.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+                        long onNextEnd = System.nanoTime();
+                        if (TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart) > 0) {
+                            log.debug("{}::onNext took {}s", listener.toString(),
+                                    TimeUnit.NANOSECONDS.toSeconds(onNextEnd - onNextStart));
+                        }
+                    }
+                } catch (Throwable throwable) {
+                    log.warn("{}::onError: {}", listener.toString(), throwable.toString());
+                    listener.onError(throwable);
+                    streamManager.unsubscribe(listener);
```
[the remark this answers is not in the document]
hisundar: hmm unsubscribe() calls the executor pool's thread cancel will might interrupt the thread. I don't know the implications of executing code after interrupt is invoked on the thread.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+         *
+         * Map of (Stream Id - TableSchema)
+         */
+        @Getter
+        private final Map<UUID,
+                TableSchema<? extends Message, ? extends Message, ? extends Message>> tablesOfInterest;
+        /**
+         * Starting address.
+         */
+        private final long startAddress;
```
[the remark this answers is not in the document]
hisundar: removed. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+ * which is the same as the consumption.
+ *
+ * Created by hisundar on 04/28/2020.
+ */
+@Slf4j
+public class StreamManager {
+
+    private static final int DEFAULT_NUM_SUBSCRIBERS = 6; // Arbitrary sanity check value.
+    private static final int DEFAULT_SLEEP_MILLIS = 50; // Default time to sleep
+    private final int maxSubscribers;
```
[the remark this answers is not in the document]
hisundar: done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+     * @param startAddress     Address to start the notifications from.
+     */
+    public synchronized <K extends Message, V extends Message, M extends Message>
+    void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                   @Nonnull List<TableSchema<K, V, M>> tablesOfInterest, long startAddress) {
+        if (subscriptions.containsKey(streamListener)) {
+            throw new StreamSubscriptionException(
+                    "StreamManager:subscriber already registered "
+                            + streamListener + ". Maybe call unregister first?");
+        }
```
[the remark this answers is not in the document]
hisundar: Subscribing to same namespace and table is allowed. Which is why we caller's hashcode() and equals() are used to validate re-subscription.
If the caller does not want to start another subscriber on the same namespace and table, then the hashcode should be constructed from listener name. If caller wants to allow 2 subscriptions on the same namespace+table, then hashcode and equals can just reflect the subscriber's name or other distinguishing fields.
Mentioning this in the comments too.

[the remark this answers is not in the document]
hisundar: Basically we are saying, "hey you decide if this is a re-subscription or a new subscription"

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,238 @@
...
+     * This is the map of all streaming subscriptions.
+     */
+    @Getter
+    private final Map<StreamListener, StreamSubscriber> subscriptions;
+
+    public StreamManager(@Nonnull CorfuRuntime runtime, int maxSubscribers) {
+        this.runtime = runtime;
+        this.maxSubscribers = maxSubscribers;
+        this.subscriptions = new HashMap<>();
+        this.scheduledExecutorService = Executors.newScheduledThreadPool(maxSubscribers);
```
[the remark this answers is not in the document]
hisundar: That is what I added before, please see comments from @xnull  and @Maithem . I am open to either suggestion. 
Not having a limit means that it will be up to the caller to ensure that they don't create too many subscribers but if they do know what they are doing, then we don't limit that behavior.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -104,6 +104,13 @@ public void clear() {
         openedStreams.clear();
     }
 
+    /**
+     * Remove a specific stream from openedStreams
+     */
+    public void removeStream(IStreamView stream) {
+        openedStreams.remove(stream);
```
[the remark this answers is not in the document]
hisundar: I thought about this, but going by what @apavan said, I wanted to limit the changes made in this PR to just the bug fix.
unsubscribe() resubscribe() should be very infrequent operations anyway.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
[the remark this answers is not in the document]
hisundar: well not the second one, the second one is needed to debug log running tasks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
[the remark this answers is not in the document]
WenbinZhu: Yea my point is the comment in the StreamSubscriptionException should be consistent with the implementation here.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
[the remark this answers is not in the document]
WenbinZhu: This is another comment I had, but that was folded by github, so you probably didn't see it, so pasting here:

"I suggest not pass true to cancel(). This will raise UnrecoverableCorfuInterruptedError. We saw a lot of such bugs, and nowadays we are confident to punt these bugs to verticals because corfu never do thread.interupt() or future.cancel(true)."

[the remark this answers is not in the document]
hisundar: now fixed based on offline discussion about not interrupting thread at all. Brought back the shutdown variable.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -837,6 +837,7 @@ private void handleUncaughtThread(@Nonnull Thread thread, @Nonnull Throwable thr
     public void shutdown() {
         // Stopping async task from fetching layout.
         isShutdown = true;
+        getTableRegistry().shutdown();
```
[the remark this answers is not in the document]
hisundar: fixed to remove lombok and explicitly implement the getTableRegistry() construct

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,261 @@
...
+    @Getter
+    private class StreamSubscriber {
+        private final SubscriberTask subscriberTask;
+        private final ScheduledFuture<?> scheduledFuture;
+    }
+
+    public StreamManager(@Nonnull CorfuRuntime runtime, int maxSubscribers) {
+        this.runtime = runtime;
+        this.subscriptions = new HashMap<>();
+        this.scheduledExecutorService = Executors.newScheduledThreadPool(maxSubscribers);
```
WenbinZhu, line 75: @hisundar I think at least we need to log a warning message if the number of subscribers exceeds limit, otherwise it will be hard to debug when such issues happen. Also is 6 a bit large for default maxSubscribers considering each subscriber will now have a thread?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -0,0 +1,261 @@
...
+        /**
+         * Tables of interest.
+         *
+         * Map of (Stream Id - TableSchema)
+         */
+        @Getter
+        private final Map<UUID,
+                TableSchema<? extends Message, ? extends Message, ? extends Message>> tablesOfInterest;
+
+        public AtomicBoolean shutdown;
```
WenbinZhu, line 179: There is already a `isShutdown` variable on line 168 right?

WenbinZhu (approved):
LGTM, some minor comments.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: ok, so how about a large limit like 32 at which we will throw an exception. It will only be hit in case there is some bug that is calling subscribe() and leaking?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
hisundar: fixed. thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
WenbinZhu: Yea we can do that, but one thing I think that is also important is logging the number of subscribers when there are new subscribes, since the current design is supposed to be used by a small number of subscriptions, so if we hit a performance issue, we can easily tell there are too many subscribers.

annym:
@hisundar This PR is getting stuck on Injection Framework, I'll disable it for now, if you get back to it and need to run IF please let me know and we can test it.

hisundar:
> @hisundar This PR is getting stuck on Injection Framework, I'll disable it for now, if you get back to it and need to run IF please let me know and we can test it.

This change is already merged right? Why does it impact IF ?

annym:
> > @hisundar This PR is getting stuck on Injection Framework, I'll disable it for now, if you get back to it and need to run IF please let me know and we can test it.
> 
> This change is already merged right? Why does it impact IF ?

I was wondering the same as soon as I published the comment and realized it was merged. But now I know why. When you needed to merge this PR the IF was not picking old PRs and I added this one manually to the list. But I just removed it from the list so we should be fine.

PR #2534, opened 2020-05-09
Lower logging level for log writes

xnull:
## Overview
We can not afford so massive amount of logs. 
"Log write" messages often consume up to 80% of all logs in the system overall.
Here are some numbers from a system (3 node cluster), Corfu by itself consumes 80-90% of all logs:

image

image

Description:

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -209,7 +209,7 @@ public class LogUnitServer extends AbstractServer {
     @ServerHandler(type = CorfuMsgType.WRITE)
     public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
         LogData logData = (LogData) msg.getPayload().getData();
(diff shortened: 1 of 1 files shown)


xnull:
> This log is quite useful, it helped triage a lot of bugs in clustering, steam layer and data loss, I don't think it should be lowered.

Hey, Wenbin, could you please provide in a private email that contains the list of those bugs (the aim is to have a discussion in detail how useful were those messages a how they were used to find bugs). 

Also would be great to understand your vision -  how much we can pay for having this information in logs? Is it ok to generate up to 80% of the overall logs in the system to be able to find those bugs? Are those bugs equal to 80% of all problems in all projects? Things like that.

Also would be great to understand what is the motivation for having those log messages in the `production` systems (why we can't enable it only for testing)? How it can be used to find bugs in prod? Do we have any cases for the last year when bugs were found in production using "log write" messages? (please also specify particular bug numbers, it's very important)

Maithem:
I think anyone who fixed/debugged non-trivial bugs will attest that this log has helped them enormously. Although these logs are set to DEBUG, they serve more like a "tracer". I think we can easily reduce the output by more than half by shortening the message. For now, I don't think its a good idea to change it to TRACE. 

TRACE is not enabled in production and too noisy.

Issue #2533, opened 2020-05-07
getTrimMark() is ineffieicent due to sequential and duplicate requests.


WenbinZhu:
## Overview

getTrimMark() sends requests to log unit servers to get the trim marks and aggregrate to return the maximum. The current implementation uses java stream and intend to send out requests in parallel, however steam in java is lazily evaluated, and the current usage  actually makes it send sequentially which is not efficient. 

Another problem is in case of multiple semgents and multiple stripes the current code will send a lot of duplicate requests.

Current code:
<pre>
public Token getTrimMark() {
        return layoutHelper(
                e -> {
                    long trimMark = e.getLayout().segments.stream()
                            .flatMap(seg -> seg.getStripes().stream())
                            .flatMap(stripe -> stripe.getLogServers().stream())
                            .map(e::getLogUnitClient)
                            .map(LogUnitClient::getTrimMark)
                            .map(future -> {
                                // This doesn't look nice, but its required to trigger
                                // the retry mechanism in AbstractView. Also, getUninterruptibly
                                // can't be used here because it throws a UnrecoverableCorfuInterruptedError
                                try {
                                    // <b><i>lazily evaluted, the previous map() call is combined with this map() call, making it sequential</b></i>
                                    return future.join();   
                                } catch (CompletionException ex) {
                                    Throwable cause = ex.getCause();
                                    if (cause instanceof RuntimeException) {
                                        throw (RuntimeException) cause;
                                    } else {
                                        throw new RuntimeException(cause);
                                    }
                                }
                            })
                            .max(Comparator.naturalOrder()).get();
                    return new Token(e.getLayout().getEpoch(), trimMark);
                });
    }

<\pre>

Issue #2117, opened 2019-08-15
Potential race in StreamLogDataStore


WenbinZhu:
## Overview

StreamLogDataStore is using `AtomicLong`, but not exploiting it atomic operation, which could lead to potential race conditions.

For example: 

A reader is executing getStartingAddress()
```java
public long getStartingAddress() {
        if (startingAddress.get() == Address.NON_ADDRESS) {
            startingAddress.set(dataStore.get(STARTING_ADDRESS_RECORD, ZERO_ADDRESS));
        }

        return startingAddress.get();
    }
```

while A writer executing updateStartingAddress()
```java
public void updateStartingAddress(long newStartingAddress) {
        log.info("Update starting address to: {}", newStartingAddress);

        dataStore.put(STARTING_ADDRESS_RECORD, newStartingAddress);
        startingAddress.set(newStartingAddress);
    }
```

After reader executes ` dataStore.get(STARTING_ADDRESS_RECORD, ZERO_ADDRESS)`, thread switches to writer which finishes executing `updateStartingAddress`. Now `startingAddress` is set to the new value. Then thread switches back to reader, and executes `startingAddress.set()`, which sets an stale value, and will not be changed until next update operation. 

This might be fine for now, seems like it's not a case in the current consumer code (StreamLogFiles), but could be a problem later. We should exploit atomic operations given by AtomicLong, e.g. `getAndUpdate()`.

Besides, the last two lines of `updateStartingAddress` are not atomic either, which should be fine for now as we use single threaded writer, but still should be aware of.

PR #2552, opened 2020-05-28
Introducing auto commit mechanism for faster read and state transfer.

WenbinZhu:
## Overview

Description:

Introducing auto commit mechanism for faster read and state transfer.

Related issue(s) (if applicable): #2533, #2117, #2482 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,188 @@
...
+                // commit up to the global tail at the time of last auto commit cycle.
+                if (!Address.isAddress(lastLogTail)) {
+                    initializeTails();
+                    return;
+                }
+
+                // Fetch maximum trim mark from log units and compare with committed tail.
+                // In order not to fetch trim mark for every retry, we only do this when
+                // the epoch changes after last attempt.
+                long currEpoch = getCorfuRuntime().getLayoutView().getLayout().getEpoch();
```
PavelZaytsev, line 99: I understand that this is an optimization, however, I am pondering if it can result in any issues. For example, what if we have a one node setup, where the epochs change infrequently, but log grows fast, so gets trimmed a lot. Will we by accident committing addresses that have already been trimmed?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -471,61 +472,84 @@ private ILogData cacheLoadAndGet(@NonNull Cache<Long, ILogData> cache, long addr
...
+     *
+     * @param start start of address range, inclusive
+     * @param end   end of address range, inclusive
+     */
+    public void commit(long start, long end) {
+        if (start >= end) {
+            return;
+        }
+
+        ContiguousSet<Long> range = ContiguousSet.create(
```
PavelZaytsev, line 539: Are there any advantages of using this rather than a simple ArrayList/ImmutableList, for example?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1186,6 +1197,22 @@ public LogData read(long address) {
         }
     }
 
+    @Override
```
PavelZaytsev, line 1200: I think we need JavaDoc here.

PavelZaytsev: Also maybe we can give a more clear name to the method.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,188 @@
...
+
+            } catch (Throwable t) {
+                log.error("runAutoCommit: encountered unexpected exception", t);
+                lastLogTail = getCorfuRuntime().getSequencerView().query().getSequence();
+                throw t;
+            }
+        }
+
+        // Update lastLogTail no matter commit cycle succeeds or not.
+        lastLogTail = getCorfuRuntime().getSequencerView().query().getSequence();
```
PavelZaytsev, line 154: Right now when we get a token from a sequencer we do it via layoutHelper, which retries up to systemDownHandlerTriggerLimit if something goes wrong. In some cases, for example, when the sequencer requires a full bootstrap from a failure detector, this can get stuck for quite a long period of time. I am wondering if there is a possibility of this interfering with a periodically scheduled auto-commit task. For example, if this update to the lastLogTail can somehow interfere with a next scheduled runAutoCommit task because it takes longer than expected to finish it.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,188 @@
...
+        autoCommitScheduler.scheduleAtFixedRate(
+                () -> LambdaUtils.runSansThrow(this::runAutoCommit),
+                interval.toMillis() / 2,
+                interval.toMillis(),
+                TimeUnit.MILLISECONDS
+        );
+    }
+
+    @VisibleForTesting
+    void runAutoCommit() {
```
PavelZaytsev, line 76: I think it might be beneficial to either make it public or define some public wrapper,so that other code can invoke it. Right now it can only be called from inside a package

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
```diff
@@ -195,6 +206,28 @@ Layout restoreWithBackOff(CorfuRuntime runtime, StateTransferManager transferMan
...
+    private void transferCommittedTail(CorfuRuntime runtime, Layout layout, long committedTail) {
+        final int maxRetry = 3;
+        LogUnitClient logUnitClient = runtime.getLayoutView()
+                .getRuntimeLayout(layout)
+                .getLogUnitClient(currentNode);
+
+        for (int i = 0; i < maxRetry; i++) {
+            try {
+                CFUtils.getUninterruptibly(logUnitClient.updateCommittedTail(committedTail),
+                        TimeoutException.class, NetworkException.class);
```
zhangn49, line 223: Does it only update current node's committedTail? Should it also update source node's ct?

zhangn49: Should we add a log info/debug here to indicate transfer CT succeed? 
updateCommittedTail in LogUnitServer's log is in trace level, and it would be better for us to know retry exhausts or not

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -361,6 +361,7 @@ public void run() {
 
         private boolean single = true;
         private boolean tlsEnabled = false;
+        private boolean autoCommit = false;
```
zhangn49, line 364: Can we use noAutoCommit to be consistent with ServerContextBuilder?

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
```diff
@@ -127,6 +125,49 @@ public ILogData peek(RuntimeLayout runtimeLayout, long globalAddress) {
...
+        List<CompletableFuture<InspectAddressesResponse>> futures = serverAddressMap
+                .entrySet()
+                .stream()
+                .map(entry -> entry.getKey().inspectAddresses(entry.getValue()))
+                .collect(Collectors.toList());
+
+        // Merge the inspect responses from different log unit servers.
+        List<Long> holes = futures.stream()
+                .flatMap(future -> CFUtils.getUninterruptibly(future).getEmptyAddresses().stream())
+                .collect(Collectors.toList());
```
zhangn49, line 150: How it handles TrimException in inspectAddresses?

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -123,6 +133,8 @@ public ManagementAgent(@NonNull SingletonResource<CorfuRuntime> runtimeSingleton
                 localMonitoringService
         );
 
+        this.autoCommitService = new AutoCommitService(serverContext, runtimeSingletonResource);
```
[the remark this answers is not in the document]
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        // Update lastLogTail no matter commit cycle succeeds or not.
+        lastLogTail = getCorfuRuntime().getSequencerView().query().getSequence();
+    }
+
+    private Layout updateLayoutAndGet() {
+        return getCorfuRuntime()
+                .invalidateLayout()
+                .thenApply(serverContext::saveManagementLayout)
```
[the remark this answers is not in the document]
WenbinZhu: I kept it same as what RemoteMonitoringService does. Saving management layout helps server to recover with lastest layout.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: The if we try to inspect the address brfore the trim mark, log unit server will throw a TrimmedException. Also committing addresses that have trimmed should be safe.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
WenbinZhu: Here we need a fast way to create a contiguous list/set given by two endpoints.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
WenbinZhu: The JavaDoc is provided in the public interface.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: I changed to use seqeuncer client instead of sequencer view

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
```diff
@@ -154,6 +174,19 @@ public LogData read(long address) {
         return logCache.get(address);
     }
 
+    @Override
+    public boolean contains(long address) throws TrimmedException {
+        if (isTrimmed((address))) {
+            throw new TrimmedException();
+        }
+
```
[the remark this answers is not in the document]
WenbinZhu: In memory stream log still needs a notion of trimmed.

infrastructure/src/main/java/org/corfudb/infrastructure/log/InMemoryStreamLog.java
```diff
@@ -109,6 +114,21 @@ public synchronized StreamsAddressResponse getStreamsAddressSpace() {
         return new StreamsAddressResponse(logMetadata.getGlobalTail(), logMetadata.getStreamsAddressSpaceMap());
     }
 
+    @Override
+    public long getCommittedTail() {
+        return committedTail.get();
+    }
+
+    @Override
+    public synchronized void updateCommittedTail(long newCommittedTail) {
```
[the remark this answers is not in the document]
WenbinZhu: Removed.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
WenbinZhu: We only need to update current node's committed tail to prevent an edge case that losing all committed tails. We don't update the source because we don't want the state transfer to have too much dependency on the committed tail, if the source is not quite responsive, the state transfer on current node can not be finished.

test/src/test/java/org/corfudb/integration/AbstractIT.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/ChainReplicationProtocol.java
WenbinZhu: TrimmedException is poped up tp AddressSpaceView and then AutoCommitService. Then current commit cycle fails and retries. There's actually a bug in handling retry, I fixed will add a test case.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * An auto-commit service that periodically commits the unwritten addresses
+ * in the global log, continuously consolidating the log prefix.
+ * <p>
+ * Created by WenbinZhu on 5/5/20.
+ */
+@Slf4j
+public class AutoCommitService implements ManagementService {
+
+    private static final int COMMIT_BATCH_SIZE = 500;
```
xnull, line 35: please make these parameters not static - it allows to make it configurable in the future. Right now this is hard coded but must be configureble

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+                interval.toMillis() / 2,
+                interval.toMillis(),
+                TimeUnit.MILLISECONDS
+        );
+    }
+
+    @VisibleForTesting
+    public synchronized void runAutoCommit() {
+        boolean trimmed = false;
+        long lastEpoch = -1;
```
xnull, line 82: why not 0?

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+                    if (committedTail < trimMark.getSequence() - 1) {
+                        Token trimToken = new Token(trimMark.getEpoch(), trimMark.getSequence() - 1);
+                        getCorfuRuntime().getAddressSpaceView().prefixTrim(trimToken, false);
+                        committedTail = trimMark.getSequence() - 1;
+                    }
+                    lastEpoch = currEpoch;
+                    trimmed = false;
+                }
+
+                log.debug("runAutoCommit: trying to commit [{}, {}].", committedTail + 1, lastLogTail);
```
xnull, line 120: please write a message to the log only ones - whether it commited or failed

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+                        i, MAX_COMMIT_RETRY, re);
+
+                if (i >= MAX_COMMIT_RETRY) {
+                    log.warn("runAutoCommit: retry exhausted, abort and wait for next cycle");
+                    break;
+                }
+
+                // When inspecting address, log unit server could throw a TrimmedException if
+                // the address to inspect is trimmed, then we need to adjust committed tail.
+                if (re instanceof TrimmedException) {
```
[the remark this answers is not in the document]
xnull: why this warning is not fixed?

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+                    trimmed = true;
+                }
+
+                if (re instanceof NetworkException || re.getCause() instanceof TimeoutException) {
```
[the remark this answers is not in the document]
xnull: why the warning is not fixed?

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+                updateLastLogTail(currentLayout);
+                throw t;
+            }
+        }
+
+        // Update lastLogTail no matter commit cycle succeeds or not.
+        updateLastLogTail(currentLayout);
+    }
+
+    private Layout updateLayoutAndGet() {
```
xnull, line 171: what's the motivation, why you need to update the layout?
FD is in charge of doing it, not sure if it safe to update the layout here

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+    private final SingletonResource<CorfuRuntime> runtimeSingletonResource;
+    private final ScheduledExecutorService autoCommitScheduler;
+
+    // Global log tail fetched at last auto commit cycle, which
+    // would be the commit upper bound in the current cycle.
+    private long lastLogTail = Address.NON_ADDRESS;
+    // Cached committed tail so that we do not need to fetch every cycle.
+    private long committedTail = Address.NON_ADDRESS;
+
+    AutoCommitService(@NonNull ServerContext serverContext,
```
xnull, line 49: Why do we need to run it as a separate service, why we can't execute it in RemoteMonitoringService after the iteration of failure detection?

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+                    log.trace("runAutoCommit: successfully committed batch [{}, {}]", committedTail, commitEnd);
+                    committedTail = commitEnd;
+                }
+
+                log.debug("runAutoCommit: successfully finished auto commit cycle. " +
+                        "New committed tail: {}.", committedTail);
+                break;
+
+            } catch (RuntimeException re) {
+                log.warn("runAutoCommit: encountered an exception on attempt {}/{}.",
```
xnull, line 142: don't need to print every single time this.
Just print an error in case of you can't finish the operation

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+                        i, MAX_COMMIT_RETRY, re);
+
+                if (i >= MAX_COMMIT_RETRY) {
+                    log.warn("runAutoCommit: retry exhausted, abort and wait for next cycle");
```
xnull, line 146: no need for this

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
```diff
@@ -0,0 +1,217 @@
...
+        autoCommitScheduler.scheduleAtFixedRate(
+                () -> LambdaUtils.runSansThrow(this::runAutoCommit),
+                interval.toMillis() / 2,
+                interval.toMillis(),
+                TimeUnit.MILLISECONDS
+        );
+    }
+
+    @VisibleForTesting
+    public synchronized void runAutoCommit() {
```
xnull, line 80: The method seems too complicated. Since it is complex, it's hard to write tests for this method. Can you make it smaller and add more unit tests?

image

infrastructure/src/main/java/org/corfudb/infrastructure/LocalMonitoringService.java
```diff
@@ -23,7 +23,7 @@
  * Created by zlokhandwala on 11/2/18.
  */
 @Slf4j
-class LocalMonitoringService implements MonitoringService {
+class LocalMonitoringService implements ManagementService {
```
xnull, line 26: what was motivation for renaming it?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -308,6 +330,25 @@ public void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerCo
         }
     }
 
+    @ServerHandler(type = CorfuMsgType.INSPECT_ADDRESSES_REQUEST)
+    public void inspectAddresses(CorfuPayloadMsg<InspectAddressesRequest> msg,
+                                 ChannelHandlerContext ctx, IServerRouter r) {
+        log.trace("inspectAddresses: {}", msg.getPayload().getAddresses());
+        InspectAddressesResponse inspectResponse = new InspectAddressesResponse();
+        try {
+            for (long address : msg.getPayload().getAddresses()) {
```
xnull, line 339: code reformat please

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -44,6 +44,21 @@
      */
     private static final Duration RECOVERY_RETRY_INTERVAL = Duration.ofSeconds(1);
 
+    /**
+     * Locally collected server metrics polling interval.
+     */
+    private static final Duration METRICS_POLL_INTERVAL = Duration.ofSeconds(3);
```
xnull, line 50: please make it not constant

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -187,10 +200,15 @@ private void initializationTask() {
             }
         }
 
-        // Start monitoring services that deals with failure and healing detection.
+        // Start management services that deals with failure & healing detection and auto commit.
         if (!shutdown) {
             localMonitoringService.start(METRICS_POLL_INTERVAL);
             remoteMonitoringService.start(POLICY_EXECUTE_INTERVAL);
+            if (!serverContext.getServerConfig(Boolean.class, "--no-auto-commit")) {
```
xnull, line 207: please don't use the double negative, it confuses a lot

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
```diff
@@ -41,6 +42,15 @@
...
     LogData read(long address);
 
+    /**
+     * Inspect if the stream log contains the entry at given address.
+     *
+     * @param address the address to inspect
+     * @return true if stream log contains the entry at the address, false otherwise
+     * @throws TrimmedException if address less than starting address
+     */
+    boolean contains(long address) throws TrimmedException;
```
xnull, line 52: please don't throw an exception, just return an enum if you really need to know about trimm

xnull: `contains` must not throw any exceptions, else it's is fragile and weird logic

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -118,7 +119,7 @@ public StreamLogFiles(ServerContext serverContext, boolean noVerify) {
         writeChannels = new ConcurrentHashMap<>();
         channelsToSync = new HashSet<>();
         this.verify = !noVerify;
-        this.dataStore = StreamLogDataStore.builder().dataStore(serverContext.getDataStore()).build();
+        this.dataStore = new StreamLogDataStore(serverContext.getDataStore());
```
xnull, line 122: please bring builder back

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
```diff
@@ -195,6 +206,30 @@ Layout restoreWithBackOff(CorfuRuntime runtime, StateTransferManager transferMan
                 .collect(Collectors.toList());
     }
 
+    /**
+     * Send the last transferred address as the committed tail to the target log unit.
+     * This is required to prevent loss of committed tail after state transfer finishes
+     * and then all other log units failed and auto commit service is paused.
+     */
+    private void transferCommittedTail(CorfuRuntime runtime, Layout layout, long committedTail) {
+        final int maxRetry = 3;
```
xnull, line 215: please transform this variable into a filed (in the future probably we would like to make this parameter configurable via the app config)

runtime/src/main/java/org/corfudb/protocols/wireprotocol/InspectAddressesResponse.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * A response message containing a list of uncommitted addresses.
+ *
+ * Created by WenbinZhu on 5/4/20.
+ */
+@Data
+@AllArgsConstructor
+public class InspectAddressesResponse implements ICorfuPayload<InspectAddressesResponse> {
+
+    @Getter
```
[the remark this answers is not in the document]
xnull: please fix the warning

runtime/src/main/java/org/corfudb/protocols/wireprotocol/InspectAddressesResponse.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A response message containing a list of uncommitted addresses.
+ *
+ * Created by WenbinZhu on 5/4/20.
+ */
+@Data
```
xnull, line 16: don't use @data, use @getter, @setter instead

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java
```diff
@@ -186,6 +187,19 @@ private static Object handleReadDataCorruption(CorfuPayloadMsg<Long> msg,
...
 
+    /**
+     * Handle a INSPECT_ADDRESSES_RESPONSE message.
+     *
+     * @param msg Incoming Message
+     * @param ctx Context
+     * @param r   Router
+     */
+    @ClientHandler(type = CorfuMsgType.INSPECT_ADDRESSES_RESPONSE)
+    private static Object handleInspectAddressResponse(CorfuPayloadMsg<InspectAddressesResponse> msg,
```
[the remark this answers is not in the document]
xnull: please fix the warning

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java
```diff
@@ -199,12 +213,33 @@ private static Object handleTailResponse(CorfuPayloadMsg<TailsResponse> msg,
...
 
+    /**
+     * Handle a COMMITTED_TAIL_RESPONSE message.
+     *
+     * @param msg Incoming Message
+     * @param ctx Context
+     * @param r   Router
+     */
+    @ClientHandler(type = CorfuMsgType.COMMITTED_TAIL_RESPONSE)
+    private static Object handleCommittedTailResponse(CorfuPayloadMsg<Long> msg,
```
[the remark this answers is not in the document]
xnull: please fix all the warnings

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
```diff
@@ -117,6 +117,15 @@ public ILogData peek(RuntimeLayout runtimeLayout, long address) {
                 .collect(Collectors.toMap(SimpleImmutableEntry::getKey, SimpleImmutableEntry::getValue));
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void commitAll(RuntimeLayout runtimeLayout, Collection<Long> addresses) {
+        // TODO: parallelize this operation as in chain replication.
```
xnull, line 125: get rid of todo

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: They are fine tuned for now, so they are static final, later if we need to expose them, I will change it accordingly.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: It needs to be initialized to an invalid epoch.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: Layout need to be updated to handle wong epoch, and disable auto commit if not primary sequencer. Safety is guarantee by using layoutHelper. I removed saveManagementLayout as it's not important.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: 1) RemoteMonitoringService executes too often, auto commit needs its own interval
2) Auto commit can block RemoteMonitoringService
3) It's cleaner to have a separate service

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: Done, lowered to trace.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: Since now we don't log on every exception, we need this.

infrastructure/src/main/java/org/corfudb/infrastructure/AutoCommitService.java
WenbinZhu: It's not complited, the core logic of this method is just 30 lines, others are just exception handling and comments.

infrastructure/src/main/java/org/corfudb/infrastructure/LocalMonitoringService.java
WenbinZhu: Now with the addition of AutoCommitService, it's better to call ManagementService instead of MonitoringService.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
WenbinZhu: There is nothing to format here, the entire method is already well formatted. Even IDE tells it's already formatted.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
WenbinZhu: This is your code, I didn't change, just moved up several lines.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
WenbinZhu: Because by default auto commit is on, then if user specifies this flag, it's disabled. So has to use double negative here.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLog.java
WenbinZhu: If you look at methods, like read/write, they throw exceptions like DataCorruptionException, OverwriteException, and catch them. They just don't mention it in the interface. It's actully the same. Here TrimmedException is important for upper layer to catch, so adding it here as a heads up.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
WenbinZhu: The constructor has only one parameter, having a builder is an overkill.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/actions/RestoreRedundancyMergeSegments.java
WenbinZhu: This is quite internal and not worth exposing to user.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/InspectAddressesResponse.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/replication/QuorumReplicationProtocol.java
WenbinZhu: This class is not being used, so not implementing for now, just made it compile.

PR #2560, opened 2020-06-02
CorfuStoreBrowser: add option to load data into a table

hisundar:
This is useful for testing checkpoint issues in live setups.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -3,7 +3,6 @@ package org.corfudb.browser;
 import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
@@ -13,15 +12,15 @@ import java.util.stream.Stream;
 
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
-import org.corfudb.runtime.CorfuStoreMetadata;
 import org.corfudb.runtime.CorfuStoreMetadata.TableName;
-import org.corfudb.runtime.collections.CorfuRecord;
-import org.corfudb.runtime.collections.CorfuStoreEntry;
+import org.corfudb.runtime.collections.CorfuStore;
 import org.corfudb.runtime.collections.CorfuTable;
 import org.corfudb.runtime.collections.CorfuDynamicKey;
 import org.corfudb.runtime.collections.CorfuDynamicRecord;
 import org.corfudb.runtime.collections.PersistedStreamingMap;
 import org.corfudb.runtime.collections.StreamingMap;
+import org.corfudb.runtime.collections.TableOptions;
+import org.corfudb.runtime.collections.TxBuilder;
 import org.corfudb.runtime.object.ICorfuVersionPolicy;
 import org.corfudb.runtime.view.SMRObject;
 import org.corfudb.runtime.view.TableRegistry;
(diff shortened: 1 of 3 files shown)


corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -38,14 +39,16 @@
...
-        + "--operation=<listTables|infoTable|showTable|dropTable> Operation\n"
+        + "--operation=<listTables|infoTable|showTable|dropTable|loadTable> Operation\n"
         + "--namespace=<namespace>   Namespace\n"
         + "--tablename=<tablename>   Table Name\n"
         + "--keystore=<keystore_file> KeyStore File\n"
         + "--ks_password=<keystore_password> KeyStore Password\n"
         + "--truststore=<truststore_file> TrustStore File\n"
         + "--truststore_password=<truststore_password> Truststore Password\n"
         + "--diskPath=<path to temp folder for large tables> Path to Temp Dir\n"
+        + "--numItems=<number of items to load> Number of items to load for loadTable\n"
```
pankti-m, line 50: looks like we dont use these parameters.  Can they be removed?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -203,4 +199,28 @@ public int dropTable(String namespace, String tablename) {
...
+                .setTypeToken(new TypeToken<CorfuTable<byte[], byte[]>>() {})
+                .setStreamName("test")
+                .open();
+        final int batchSize = 1000;
+        final int numItems = 1000;
+        byte[] bigBadBuf = "JUST_A_RANDOMLY_LOOOOOOOOOOOOOOOOOOOOONG_STRING".getBytes();
+        for (int i = numItems; i> 0; i--) {
+            log.info("loadTable: Batches left {}", i);
+            for (int j = 0; j < batchSize; j++) {
+                corfuTable.put(bigBadBuf, bigBadBuf);
```
pankti-m, line 221: this will write the same entry over and over again.  Is it the intention?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -203,4 +199,28 @@ public int dropTable(String namespace, String tablename) {
...
         return tableSize;
     }
+
+    /**
+     * Loads the table with random data
+     * @param namespace - the namespace where the table belongs
+     * @param tablename - table name without the namespace
+     * @return - number of entries in the table
+     */
+    public int loadTable(String namespace, String tablename) {
```
pankti-m, line 209: can we add a comment explaining what the purpose of this operation is and when it should be used?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: yes - it was meant to test checkpointing where we want to see if the checkpointer reduces the space. space will only be reduced if there is only one real key-value pair

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
hisundar: it is used but not as what it looks like on first glance - will also add an option to insert multiple items instead of same key multiple times

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: sure

hisundar:
merged into log-replication-master

PR #2576, opened 2020-06-16
Added Delta Transfer IMPL

zhangn49:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
```diff
@@ -72,6 +72,7 @@
     RESET_LOGUNIT(47, new TypeToken<CorfuPayloadMsg<Long>>(){}, true, false),
     LOG_ADDRESS_SPACE_REQUEST(48, TypeToken.of(CorfuMsg.class)),
     LOG_ADDRESS_SPACE_RESPONSE(49, new TypeToken<CorfuPayloadMsg<StreamsAddressResponse>>(){}),
+    SUFFIX_TRIM(69, new TypeToken<CorfuPayloadMsg<Long>>() {}),
```
PavelZaytsev, line 75: I don't think this needs to be a payloaded message with an epoch. CorfuMsg already has epoch as a part of the base fields.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -343,6 +376,65 @@ public void prefixTrim(long address) {
...
+
+            // Check if committed tail is not less than starting address
+            if (isTrimmed(committedTail)) {
+                throw new SuffixTrimException("local committed tail is trimmed",
+                        committedTail, globalTailCopy);
+            }
+
+            // Check if committed tail is greater than global tail
+            if (committedTail > globalTailCopy) {
+                throw new SuffixTrimException("local committed tail is greater than global tail",
```
PavelZaytsev, line 396: I think the message should start with a capital letter.

PavelZaytsev: Also, how is this exception being handled? Looks like the user will get CorfuMsgType.ERROR_SERVER_EXCEPTION and the log unit server will throw a LogUnitException. Is this what you want? Maybe we could add a handler to the LogUnitHandler and give a client a more specific exception?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -343,6 +376,65 @@ public void prefixTrim(long address) {
...
 
+    @Override
+    public void suffixTrim() {
+
+        // Have a global tail copy is necessary, since degrade to reset is dependent on it
+        long globalTailCopy = logMetadata.getGlobalTail();
+
+        try {
+            long committedTail = dataStore.getCommittedTail();
+
```
PavelZaytsev, line 387: I think its also a good idea to check that the committed tail is initialized. Because otherwise it will throw an isTrimmed exception, since the committed tail is initialized to -1 and the starting address is initialized to 0.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -343,6 +376,65 @@ public void prefixTrim(long address) {
...
+            }
+
+            // Check if committed tail is greater than global tail
+            if (committedTail > globalTailCopy) {
+                throw new SuffixTrimException("local committed tail is greater than global tail",
+                        committedTail, globalTailCopy);
+            }
+
+            // Close all segment handlers and delete suffix log files
+            long committedSegment = committedTail / RECORDS_PER_LOG_FILE;
```
PavelZaytsev, line 401: I know you did not start it but I think this segment calculation should be put into a separate function.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -343,6 +376,65 @@ public void prefixTrim(long address) {
...
+                    long segment = Long.parseLong(segmentStr);
+                    return segment >= committedSegment && segment <= tailSegment;
+                } catch (Exception e) {
+                    log.warn("suffixTrim: ignoring file {}", file.getName());
+                    return false;
+                }
+            });
+
+            // Update tail segment
+            long newTailSegment = Math.max(committedSegment - 1, 0);
```
PavelZaytsev, line 416: I think it's a good idea to explain why the committed segment is also wiped out.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
zhangn49: This Exception only carries error message and data, and all runtime exceptions will be caught by below line 427. Once it gets an RuntimeException, it will degrade to reset. I am not sure if this is too conservative.

For some cases like committedTail > globalTail, maybe we can assume that all data is committed and keep suffixTrim. What do you think?

PR #2578, opened 2020-06-18
Log Replication Design Redefinition

annym:
## Overview

Description:

- Single CorfuLogReplicationRuntime per Cluster
- Leadership Verification as integrated part of Corfu
- Cluster/Site Discovery/Changes managed through Corfu
- Rename site -> cluster (Corfu terminology)
- Cleanup


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+     * Stop Log Replication, regardless of current state.
+     */
+    public void stop() {
+        log.info("Local leadership lost. Log Replication will immediately stop.");
+        input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.LOCAL_LEADER_LOSS));
+    }
+}
```
zhangn49, line 296: It seems like LOCAL_LEADER_LOSS event can not stop log replication, so discovery service is not able to be shut down.

annym: I am reviewing the state machine, let me take special look at this case! I'll let you know!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -0,0 +1,279 @@
...
+         * "snapshotStart": "-1"
+         * "snapshotSeqNum": " -1"
+         * "snapshotTransferred": "-1"
+         * "snapshotApplied": "-1"
+         * "lastLogEntryProcessed": "-1"
+         */
+        if (negotiationResponse.getSnapshotStart() == -1) {
+            negotiationResponse.getLastLogProcessed();
+            fsm.input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.NEGOTIATION_COMPLETE,
+                    new LogReplicationEvent(LogReplicationEvent.LogReplicationEventType.SNAPSHOT_SYNC_REQUEST)));
```
zhangn49, line 198: return here. Otherwise, it will trigger more than one event.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -0,0 +1,279 @@
...
+         * "snapshotTransferred": "-1"
+         * "snapshotApplied": "-1"
+         * "lastLogEntryProcessed": "-1"
+         */
+        if (negotiationResponse.getSnapshotStart() > negotiationResponse.getSnapshotTransferred()) {
+            log.info("Get the negotiation response {} and will start replication.",
+                    negotiationResponse);
+            fsm.input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.NEGOTIATION_COMPLETE,
+                    new LogReplicationEvent(LogReplicationEvent.LogReplicationEventType.SNAPSHOT_SYNC_REQUEST)));
+        }
```
zhangn49, line 217: Same here

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -0,0 +1,279 @@
...
+         * "snapshotTransferred": "100"
+         * "snapshotApplied": "-1"
+         * "lastLogEntryProcessed": "-1"
+         */
+        if (negotiationResponse.getSnapshotStart() == negotiationResponse.getSnapshotTransferred() &&
+                negotiationResponse.getSnapshotTransferred() > negotiationResponse.getSnapshotApplied()) {
+            fsm.input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.NEGOTIATION_COMPLETE,
+                    new LogReplicationEvent(LogReplicationEvent.LogReplicationEventType.SNAPSHOT_WAIT_COMPLETE,
+                            new LogReplicationEventMetadata(LogReplicationEventMetadata.getNIL_UUID(), negotiationResponse.getSnapshotStart()))));
+        }
```
zhangn49, line 238: Same here

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -0,0 +1,279 @@
...
+             */
+            if (logHead <= negotiationResponse.getLastLogProcessed() + 1) {
+                fsm.input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.NEGOTIATION_COMPLETE,
+                        new LogReplicationEvent(LogReplicationEvent.LogReplicationEventType.REPLICATION_START,
+                                new LogReplicationEventMetadata(LogReplicationEventMetadata.getNIL_UUID(), negotiationResponse.getLastLogProcessed())
+                        )));
+            } else {
+                fsm.input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.NEGOTIATION_COMPLETE,
+                        new LogReplicationEvent(LogReplicationEvent.LogReplicationEventType.SNAPSHOT_SYNC_REQUEST)));
+            }
```
zhangn49, line 268: here

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -0,0 +1,279 @@
...
+            log.error("The active site configID {} is bigger than the standby configID {} ",
+                    metadataManager.getSiteConfigID(), negotiationResponse.getSiteConfigID());
+            throw new LogReplicationNegotiationException("Mismatch of configID");
+        }
+
+        /*
+         * The standby site has larger config ID, redo the whole discovery for the active site
+         * it will be triggered by a notification of the site config change.
+         */
+        if (negotiationResponse.getSiteConfigID() > negotiationResponse.getSiteConfigID()) {
```
zhangn49, line 169: negotiationResponse.getSiteConfigID() > metadataManager.getSiteConfigID()

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -0,0 +1,279 @@
...
+            log.error("The active site version {} is different from standby site version {}",
+                    metadataManager.getVersion(), negotiationResponse.getVersion());
+            throw new LogReplicationNegotiationException(" Mismatch of version number");
+        }
+
+        /*
+         * The standby site has a smaller config ID, redo the discovery for this standby site when
+         * getting a new notification of the site config change if this standby is in the new config.
+         */
+        if (negotiationResponse.getSiteConfigID() < negotiationResponse.getSiteConfigID()) {
```
zhangn49, line 159: negotiationResponse.getSiteConfigID() < metadataManager.getSiteConfigID()

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -236,75 +238,75 @@ public boolean setSrcBaseSnapshotStart(long siteConfigID, long ts) {
...
-        s.concat(LogReplicationMetadataType.LastSnapshotTransferred.getVal() + " " + getLastSnapTransferDoneTimestamp() + " ");
-        s.concat(LogReplicationMetadataType.LastSnapshotApplied.getVal() + " " + getLastSrcBaseSnapshotTimestamp() + " ");
-        s.concat(LogReplicationMetadataType.LastSnapshotSeqNum.getVal() + " " + getLastSnapSeqNum() + " ");
-        s.concat(LogReplicationMetadataType.LastLogProcessed.getVal() + " " + getLastProcessedLogTimestamp() + " ");
+        s.concat(LogReplicationMetadataType.TOPOLOGY_CONFIG_ID.getVal() + " " + getSiteConfigID() +" ");
+        s.concat(LogReplicationMetadataType.LAST_SNAPSHOT_STARTED.getVal() + " " + getLastSnapStartTimestamp() +" ");
+        s.concat(LogReplicationMetadataType.LAST_SNAPSHOT_TRANSFERRED.getVal() + " " + getLastSnapTransferDoneTimestamp() + " ");
+        s.concat(LogReplicationMetadataType.LAST_SNAPSHOT_APPLIED.getVal() + " " + getLastSrcBaseSnapshotTimestamp() + " ");
+        s.concat(LogReplicationMetadataType.LAST_SNAPSHOT_SEQ_NUM.getVal() + " " + getLastSnapSeqNum() + " ");
+        s.concat(LogReplicationMetadataType.LAST_LOG_PROCESSED.getVal() + " " + getLastProcessedLogTimestamp() + " ");
```
zhangn49, line 309: concat does not assign new string to s. Should be s = s.concat or use a string builder here.

annym: I see you addressed this in your PR, let's leave it there!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuInterClusterReplicationServer.java
```diff
@@ -299,6 +308,31 @@ private void startServer() {
...
+        this.clusterManagerAdapter = buildClusterManagerAdapter(serverContext.getPluginConfigFilePath());
+
+        // Start LogReplicationDiscovery Service, responsible for
+        // acquiring lock, retrieving Site Manager Info and processing this info
+        // so this node is initialized as Source (sender) or Sink (receiver)
+        replicationDiscoveryService = new CorfuReplicationDiscoveryService(serverContext,
+                clusterManagerAdapter, discoveryServiceCallback);
+
+        Thread replicationDiscoveryThread = new Thread(replicationDiscoveryService);
+        replicationDiscoveryThread.start();
```
zhangn49, line 331: Can we have a thread name here?

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -79,34 +85,80 @@ public void shutdown() {
     private void handleLogReplicationEntry(CorfuPayloadMsg<LogReplicationEntry> msg, ChannelHandlerContext ctx, IServerRouter r) {
         log.info("Log Replication Entry received by Server.");
```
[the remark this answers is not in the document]
annym: Yes, I'll change this one as it is likely to receive a lot of these type of messages. For the negotiation and leadership I'll leave them in this early stage as they should not be common and can help us debug.

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -79,34 +85,80 @@ public void shutdown() {
...
+                    metadata.getTopologyConfigId(),
+                    metadata.getVersion(),
+                    metadata.getLastSnapStartTimestamp(),
+                    metadata.getLastSnapTransferDoneTimestamp(),
+                    metadata.getLastSrcBaseSnapshotTimestamp(),
+                    metadata.getLastProcessedLogTimestamp());
+            log.info("Send Negotiation response");
+            r.sendResponse(msg, CorfuMsgType.LOG_REPLICATION_NEGOTIATION_RESPONSE.payloadMsg(response));
+        } else {
+            log.warn("Dropping negotiation request as this node is not the leader.");
```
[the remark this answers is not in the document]
annym: But, strictly speaking we drop the negotiation request message, and send instead a leadership loss message. We don't process the negotiation request or do anything with it.

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -26,33 +29,36 @@
...
 
-    public LogReplicationServer(@Nonnull ServerContext context, LogReplicationConfig logReplicationConfig) {
+    public LogReplicationServer(@Nonnull ServerContext context, @Nonnull  LogReplicationConfig logReplicationConfig,
+                                @Nonnull  LogReplicationMetadataManager metadataManager, String corfuEndpoint) {
         this.serverContext = context;
+        this.metadataManager = metadataManager;
+        this.sinkManager = new LogReplicationSinkManager(corfuEndpoint, logReplicationConfig, metadataManager, serverContext);
+
         this.executor = Executors.newFixedThreadPool(1,
                 new ServerThreadFactory("LogReplicationServer-", new ServerThreadFactory.ExceptionHandler()));
```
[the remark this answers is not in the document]
annym: Yeah, I think we should use multi-threading here. Let me think it a bit more and get back to you.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,514 @@
...
+
+    /***
+     * The standby cluster's leader change can lead to connection loss.
+     * If the current node is not the active cluster's leader, discard the notification.
+     * If the current node is the the active cluster's leader that is is responsible for the current
+     * replication job, will restart the replication with the remote cluster.
+     *
+     * @param event
+     */
+    private void processConnectionLoss(DiscoveryServiceEvent event) {
```
PavelZaytsev, line 431: Not used

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,514 @@
...
+    }
+
+    public void processSiteChangeNotification(DiscoveryServiceEvent event) {
+        // Stale notification, skip
+        if (event.getTopologyConfig().getTopologyConfigID() < getReplicationManager().getTopologyDescriptor().getTopologyConfigId()) {
+            log.debug("Stale Topology Change Notification, current={}, received={}", topologyDescriptor.getTopologyConfigId(), event.getTopologyConfig());
+            return;
+        }
+
+        TopologyDescriptor newConfig = new TopologyDescriptor(clusterManagerAdapter.fetchTopology());
```
PavelZaytsev, line 410: Do we assume here that this call will always succeed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -0,0 +1,301 @@
...
+        /*
+         * Remove standbys that are not in the new config
+         */
+        for (String siteID : standbysToRemove) {
+            stopLogReplicationRuntime(siteID);
+            topologyDescriptor.removeStandbySite(siteID);
+        }
+
+        //Start the standbys that are in the new config but not in the current config
+        for (String clusterId : newConfig.getStandbyClusters().keySet()) {
```
PavelZaytsev, line 215: I am wondering if it works as expected.
Let's say currentStandbys.keySet() = [2, 3, 4] and newStandbys.keySet() = [1, 2, 3]. So 1 got added and 4 got removed. We do retainAll on newStandbys on line 207, this will make it newStandbys.keySet() = [2, 3] and 1 (a new standby that is not in the current config and is only in the new config) will be left out?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,514 @@
...
+        return runtime;
+    }
+
+    /**
+     * Verify current node belongs to a cluster in the topology.
+     */
+    private boolean clusterPresentInTopology(TopologyDescriptor topology) {
+        localClusterDescriptor = topology.getClusterDescriptor(localEndpoint);
+        if (localClusterDescriptor != null) {
+            localNodeDescriptor = localClusterDescriptor.getNode(localEndpoint);
```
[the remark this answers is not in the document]
annym: fixed... if null we give the same error, not part of topology,,,

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -0,0 +1,295 @@
...
+     */
+    private void consume() {
+        try {
+            if (state.getType() == LogReplicationRuntimeStateType.STOPPED) {
+                log.info("Log Replication Communication State Machine has been stopped. No more events will be processed.");
+                return;
+            }
+
+            //  Block until an event shows up in the queue.
+            LogReplicationRuntimeEvent event = eventQueue.take();
```
xiaoqin2012, line 224: There is a possible starving case that there are two standbys, standby1 call consume, but standby1's eventQueue is  empty, so the thread is blocking here even though standby2's eventQue is not empty.

annym: How can that happen if they don't share the same eventQueue nor run on the same thread? A CorfuLogReplicationRuntime is dedicated to a standby site.

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -79,34 +85,80 @@ public void shutdown() {
     private void handleLogReplicationEntry(CorfuPayloadMsg<LogReplicationEntry> msg, ChannelHandlerContext ctx, IServerRouter r) {
         log.info("Log Replication Entry received by Server.");
 
-        LogReplicationEntry ack = sinkManager.receive(msg.getPayload());
+        if (isLeader(msg, r)) {
```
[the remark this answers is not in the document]
annym: We don't want to process any incoming message, if we are not the leader. If we received a message it's because we are believed to be the leader by the source cluster, however, we aren't.  For this reason we reply that we've lost leadership as a way to trigger the source to re-verify leader on the remote.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuInterClusterReplicationServer.java
```diff
@@ -242,49 +272,28 @@ private void startServer() {
...
                 activeServer.startAndListen();
-
-                if (upgraded) {
-                    replicationDiscoveryService.putEvent(
-                        new DiscoveryServiceEvent(DiscoveryServiceEvent.DiscoveryServiceEventType.Upgrade));
-                }
             } catch (Throwable th) {
-                log.error("CorfuServer: Server exiting due to unrecoverable error: ", th);
-                System.exit(EXIT_ERROR_CODE);
+                if (!test) {
```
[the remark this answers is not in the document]
annym: Here is the thing, in CorfuDB, all our AbstractITs run Corfu as a process.. Here we admitted to run the Log Replication Server in a dedicated thread, so for this to happen we can't force the system to exit with an error code, cause then the test even though successful is terminated with no result. But I honestly believe that the utility of running as a thread is for testing when things are going run (see logs) so it should not interfere our regular / production path. I am going to remove this logic, and anyone wanting to run the server as a thread shall do it for debugging and place it properly in an AbstractIT.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -0,0 +1,300 @@
...
+        for (ClusterDescriptor remoteCluster : topologyDescriptor.getStandbyClusters().values()) {
+            try {
+                startLogReplicationRuntime(remoteCluster);
+            } catch (Exception e) {
+                log.error("Failed to start log replication runtime for remote cluster {}", remoteCluster.getClusterId());
+
+                // Remove cluster from the list of standby's, as the cluster discovery process will receive
+                // change notification when this site becomes stable/available again.
+                topologyDescriptor.getStandbyClusters().remove(remoteCluster.getClusterId());
+            }
```
zhangn49, line 89: It looks like that this catch block is in a dead code path. Because startLogReplicationRuntime will catch all exceptions and input an event to FSM.

Remove elements in an iteration over values also dangerous.

> If the map is modified while an iteration over the collection is in progress (except through the iterator's own remove operation), the results of the iteration are undefined.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -0,0 +1,300 @@
...
+     * Query max stream tail for all streams to be replicated.
+     *
+     * @return max tail of all relevant streams.
+     */
+    private long queryStreamTail() {
+        Set<String> streamsToReplicate = logReplicationConfig.getStreamsToReplicate();
+        long maxTail = Address.NON_ADDRESS;
+        for (String s : streamsToReplicate) {
+            UUID currentUUID = CorfuRuntime.getStreamID(s);
+            Map<UUID, Long> tailMap = corfuRuntime.getAddressSpaceView().getAllTails().getStreamTails();
```
zhangn49, line 233: Query tailMap multiple times is not efficient.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,513 @@
...
+     * msg needs to send out.
+     */
+    @Override
+    public void prepareSiteRoleChange() {
+        replicationManager.prepareSiteRoleChange();
+    }
+
+    /**
+     * Query the current all replication stream log tail and calculate the number of messages to be sent.
+     * If the max tail has changed, give 0%. Otherwise,
```
PavelZaytsev, line 497: The description should be finished here

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        TopologyConfigurationMsg topologyConfig = TopologyConfigurationMsg.newBuilder()
+                .setTopologyConfigID(topologyConfigId)
+                .addAllClusters(clustersConfigs).build();
+
+        return topologyConfig;
+    }
+
+    public NodeDescriptor getNodeInfo(String endpoint) {
```
PavelZaytsev, line 75: not used

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,513 @@
...
+
+        return new LogReplicationConfig(streamsToReplicate);
+    }
+
+    /**
+     * Register interest on Log Replication Lock.
+     *
+     * The node that acquires the lock will drive/lead log replication.
+     */
+    private void registerToLogReplicationLock() {
```
PavelZaytsev, line 294: Right  now this will continue indefinitely in case of constant failures, why do we never want  to timeout?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/VerifyingRemoteLeaderState.java
```diff
@@ -0,0 +1,142 @@
...
+        this.worker.submit(this::verifyLeadership);
+    }
+
+
+    /**
+     * Verify who is the leader node on the remote cluster by sending leadership request to all nodes.
+     *
+     * If no leader is found, the verification will be attempted for LEADERSHIP_RETRIES times.
+     */
+    public synchronized void verifyLeadership() {
```
xiaoqin2012, line 86: Is it doing query the leadership?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Executor service for FSM state tasks
+     */
+    private ExecutorService communicationFSMWorkers;
+
+    /**
+     * Executor service for FSM event queue consume
+     */
+    private ExecutorService communicationFSMConsumer;
```
xiaoqin2012, line 133: Why do we need two thread pools, communicationFSMWorks and communicationFSMConsumer? Those tasks are really small, the overhead of context switch is bigger than running the task. Also, it  adds complexity to the system to make sure these two threads are sync.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -0,0 +1,295 @@
...
+ *                                                                 |             |
+ *                                                           ON_CONNECTION       |   ON_CONNECTION_UP
+ *                                                               DOWN            |     (NON-LEADER)
+ *                                                             (R-LEADER)        |      +-----+
+ *                                                                 |             |      |     |
+ *                                                                 |     +-------v------+-+   |
+ *            +---------------+      ALL STATES                    +-----+                <---+
+ *            |               |                                          |                |
+ *            |   STOPPED     <---- L-LEADER_LOSS                        |  REPLICATING   |
+ *            |               |                     SITE FLIP <-----     |                |
```
xiaoqin2012, line 66: It is possible that while doing the negotiation, there is a leadership change at the standby site.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
```diff
@@ -0,0 +1,432 @@
...
+     * @param message The message to send.
+     */
+    @Override
+    public void sendMessage(CorfuMsg message) {
+        // Get the next request ID.
+        message.setRequestID(requestID.getAndIncrement());
+        // Get Remote Leader
+        if(runtimeFSM.getRemoteLeader().isPresent()) {
+            String remoteLeader = runtimeFSM.getRemoteLeader().get();
+            channelAdapter.send(remoteLeader, CorfuMessageConverter.toProtoBuf(message));
```
xiaoqin2012, line 250: If the connection is lost during channelAdapter.send(), should it detect it and generate an REMOTE_LEADER_LOSS message?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryServiceAdapter.java
```diff
@@ -0,0 +1,23 @@
+package org.corfudb.infrastructure.logreplication.infrastructure;
+
+import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo;
+
```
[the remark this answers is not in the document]
annym: I'm not sure if this is really needed, because the logic to prepare for role change or query replication is always the same as it internally faces Corfu... @xiaoqin2012 since you introduced this, can you confirm?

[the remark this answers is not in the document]
xiaoqin2012: In the new PR https://github.com/CorfuDB/CorfuDB/pull/2586, I added the some description for this interface.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
annym: @xiaoqin2012 please address this...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,513 @@
...
+        this.localHost =  NodeLocator.parseString(serverContext.getLocalEndpoint()).getHost();
+        this.pluginFilePath = serverContext.getPluginConfigFilePath();
+        this.discoveryCallback = discoveryCallback;
+    }
+
+    public void run() {
+        try {
+            startDiscovery();
+
+            while (shouldRun) {
```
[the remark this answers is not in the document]
xiaoqin2012: Will refactor this code for another PR.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
xiaoqin2012: We can move the tailMap query above the for loop. Good catch!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/NodeDescriptor.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Getter
+    private String ipAddress;
+
+    @Getter
+    private String port;
+
+    @Getter
+    @Setter
+    private boolean leader;
```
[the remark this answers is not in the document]
annym: Yes, I also don't like the leader being part of this. I did not place it but let me see cause I saw it consumed from some place and that's why I didn't get rid of it right away.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    /**
+     * Query the current all replication stream log tail and calculate the number of messages to be sent.
+     * If the max tail has changed, give 0%. Otherwise,
+     */
+    @Override
+    public int queryReplicationStatus() {
+        return replicationManager.queryReplicationStatus();
+    }
```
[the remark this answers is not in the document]
zhangn49: @xiaoqin2012 Please have a look at this comment. I know a standby cluster should not call queryReplicationStatus, but what is the expected result if it calls?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -0,0 +1,300 @@
...
+     */
+    private long queryEntriesToSend(long timestamp) {
+        int totalNumEntries = 0;
+
+        for (CorfuLogReplicationRuntime runtime: runtimeToRemoteCluster.values()) {
+            totalNumEntries += runtime.getNumEntriesToSend(timestamp);
+        }
+
+        return totalNumEntries;
+    }
```
zhangn49, line 255: I think this numEntries is not exactly how many entries we need to send. Because we only transfer some streams, and `getNumEntriesToSend` returns (maxStreamTail - ackedTimeStamp), which will count other log entries. Besides, for loop will amplify totalNumEntries multiple times.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -0,0 +1,300 @@
...
+
+        /*
+         * If the tail has been moved, reset the base calculation
+         */
+        if (maxTail > prepareSiteRoleChangeStreamTail) {
+            prepareSiteRoleChange();
+        }
+
+        long currentNumEntriesToSend = queryEntriesToSend(prepareSiteRoleChangeStreamTail);
+        log.debug("maxTail {} totalNumEntriesToSend  {}  currentNumEntriesToSend {}", maxTail, totalNumEntriesToSend, currentNumEntriesToSend);
```
zhangn49, line 284: if the max stream tail moves, it calls `prepareSiteRoleChange()`, which call `queryStreamTail()` one more time, and will update `totalNumEntriesToSend`. Then it call `queryEntriesToSend()` again, will get a pretty close result as `currentNumEntriesToSend`. So percent calculation will always return a 0.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
zhangn49: For example, the timestamp is 20, and we are interested in stream A[1, 3, 4], B[2, 5, 6], C[7, 8, 20]
The result will be (20-4) + (20-6) + (20-20) = 30. The expected result will be 0.

annym: + @xiaoqin2012 please address this

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
annym: @xiaoqin2012 please address this

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
annym: No, the remote leader loss is a message to be sent from the standby not locally. If the connection is lost, we will get a connection_down from the leader and re-verify.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
annym: of course, and that will cause the event R_LEADER_LOSS

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
annym: These two threads do not need to be in sync... I see it this way, imagine this scenario, you are doing negotiation and blocked waiting for the response to come back, but right when you start it, the connection comes down. If you don't have a consumer thread which is dedicated you are not able to process these async events.. so you just block until it timeout, while you could've abort and released resources and go ahead and re-connect and look for a new leader.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/VerifyingRemoteLeaderState.java
annym: yes

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: We can discuss this, currently, we just need a node to be leader... otherwise no workflow is triggered. So somehow it's a requirement. But yes, we can also have a number of retries or a timeout and completely abort if we fail to register to the lock...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: @xiaoqin2012 please address this...

xiaoqin2012: Just remove the "Otherwise".

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -26,33 +28,36 @@
  * The Log Replication Server, handles log replication entries--which
  * represent parts of a Snapshot (full) sync or a Log Entry (delta) sync
  * and also handles negotiation messages, which allows the Source Replicator
- * to get a view of the last synchronized point at the remote site.
+ * to get a view of the last synchronized point at the remote cluster.
  */
 @Slf4j
 public class LogReplicationServer extends AbstractServer {
```
xiaoqin2012, line 34: Is LogReplicationServer per site or per standby?

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -77,36 +82,80 @@ public void shutdown() {
...
-                metadata.getLastSnapStartTimestamp(),
-                metadata.getLastSnapTransferDoneTimestamp(),
-                metadata.getLastSrcBaseSnapshotTimestamp(),
-                metadata.getLastProcessedLogTimestamp());
-        r.sendResponse(ctx, msg, CorfuMsgType.LOG_REPLICATION_NEGOTIATION_RESPONSE.payloadMsg(response));
+
+        if (isLeader(msg, r)) {
+            LogReplicationMetadataManager metadata = sinkManager.getLogReplicationMetadataManager();
+            LogReplicationNegotiationResponse response = new LogReplicationNegotiationResponse(
+                    metadata.getTopologyConfigId(),
```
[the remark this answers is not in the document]
xiaoqin2012: We can wrap up in one call in metadataManager and do error handling.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
xiaoqin2012: It is cleaner that the first event is timeout and retires, and then process the second event. Processing two events at the same time will lead to an un-deterministic result due to timing issues.

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
annym: a standby is a site... and yes a LRS is per standby site.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,514 @@
...
+                // Unblock server initialization retrieving context: topology + configuration
+                discoveryCallback.complete(context);
+
+                registerToLogReplicationLock();
+            } else {
+                // If a cluster descriptor is not found, this node does not belong to any topology... raise an exception
+                String message = String.format("Node[%s] does not belong to any Cluster provided by the discovery service, topology=%s",
+                        localEndpoint, topologyDescriptor);
+                log.warn(message);
+                throw new LogReplicationDiscoveryServiceException(message);
```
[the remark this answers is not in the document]
annym: This is a valid point, similar to how we retry registration to lock until successful. We can think of permitting subscription to Cluster Manager until we appear as part of the cluster and it might make sense in the event of a topology being formed. I've added this and your other comment of the lock for discussion and we can push them in new PRs.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: No, I have changed this. It can certainly fail.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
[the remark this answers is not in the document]
annym: I've added a TODO in the code for this to be addressed.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -0,0 +1,513 @@
...
+     * On discovery, process the topology information and fetch log replication configuration
+     * (streams to replicate) required by an active and standby site before starting
+     * log replication.
+     */
+    private void startDiscovery() throws LogReplicationDiscoveryServiceException {
+
+        try {
+            log.info("Connecting to Cluster Manager adapter...");
+
+            this.clusterManagerAdapter.connect(this);
```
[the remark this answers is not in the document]
annym: Added. I've enclosed the connection and query topology in a loop. If after a given number of retries it fails we throw an exception.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -174,40 +190,27 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
         rxMessageCounter++;
```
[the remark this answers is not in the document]
annym: I believe this is currently single threaded. 
@xiaoqin2012 can you confirm this?

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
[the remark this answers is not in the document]
annym: +1 I have added a TODO @xiaoqin2012 to be addressed in another PR.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -0,0 +1,295 @@
...
+                log.info("Log Replication Communication State Machine has been stopped. No more events will be processed.");
+                return;
+            }
+
+            //  Block until an event shows up in the queue.
+            LogReplicationRuntimeEvent event = eventQueue.take();
+
+            try {
+                LogReplicationRuntimeState newState = state.processEvent(event);
+                transition(state, newState);
```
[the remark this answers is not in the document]
annym: Let's discuss this offline!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
annym: Adding a comment in the code so @zhangn49 and @xiaoqin2012 can work on this!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
annym: Adding as a TODO in code for you guys to work on this on a separate PR.

PR #2579, opened 2020-06-19
Give maximum priority to log unit batch writer thread.

WenbinZhu:
## Overview

Description:

Give maximum priority to log unit batch writer thread, which is mission critical.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
@@ -1,6 +1,5 @@
 package org.corfudb.infrastructure;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import java.util.LinkedList;
@@ -48,6 +47,7 @@ public class BatchProcessor implements AutoCloseable {
     private ExecutorService processorService = Executors
             .newSingleThreadExecutor(new ThreadFactoryBuilder()
                     .setDaemon(false)
+                    .setPriority(Thread.MAX_PRIORITY)
                     .setNameFormat("LogUnit-BatchProcessor-%d")
                     .build());
 


infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -48,6 +47,7 @@
     private ExecutorService processorService = Executors
             .newSingleThreadExecutor(new ThreadFactoryBuilder()
                     .setDaemon(false)
+                    .setPriority(Thread.MAX_PRIORITY)
```
xnull, line 50: Unfortunately, in reality, it doesn't give us any advantages, setting this parameter doesn't affect anything https://www.javamex.com/tutorials/threads/priority_what.shtml

Anyway, good to have it

WenbinZhu: According the article, it does work in linux, just need to run as root and enable the jvm parameter.

xnull: >Under Linux, you have to go through more hoops to get thread priorities to function at all

>for them to work, you must be running as root (or with root privileges via setuid);
which is not our case

>the JVM parameter -XX:UseThreadPriorities must be included.

>And does anyone really run, say, their web server as root?

>The graph shows that there is a correlation between Java priority (nice value) and CPU allocation, although it is far from linear.

So, in our case, no effect at all

WenbinZhu: > the JVM parameter -XX:UseThreadPriorities must be included.
> And does anyone really run, say, their web server as root?
It depends on how you deploy, for example corfu is deployed with nice value -10 in nsx. Also giving root privilege to a service is not uncommon. Anyway it doesn't contradict with this patch, it depends on how you deploy.

> The graph shows that there is a correlation between Java priority (nice value) and CPU allocation, although it is far from linear.
Of course it cannot be linear, also nice value is just a hint, but it does work to some extent.

Maithem:
If the overall goal is to improve latency then thread affinity might be a better approach. Since its not clear how these hints are honored, thread affinity is easy to understand. Also, when  you set thread affinity on a particular thread, the linux scheduler will distribute the load of other threads. So you don't need to have "anti-affinity" logic, or worry about other processes threads.

PR #2583, opened 2020-06-24
Add Unit Test For NegotiationResponse

zhangn49:
## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/receive/LogReplicationMetadataManager.java
```diff
@@ -298,13 +298,13 @@ public void setSnapshotApplied(LogReplicationEntry entry) {
     }
 
     public String getMetadata() {
```
annym, line 300: This is intended to be a friendly print of the metadata, right? can we move it to the override of the toString?

log-replication/src/main/java/org/corfudb/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -77,6 +78,23 @@ public CorfuLogReplicationRuntime(@NonNull LogReplicationRuntimeParameters param
         corfuRuntime.connect();
     }
 
+    /**
+     * Constructor
+     *
+     * @param parameters runtime parameters
+     */
+    @VisibleForTesting
```
annym, line 86: Nan can you please rebase again on top of log-replication-feature-annym, this code changed a bit.

test/src/test/java/org/corfudb/logreplication/runtime/NegotiationResponseTest.java
```diff
@@ -0,0 +1,198 @@
...
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.BDDMockito.given;
+
+@SuppressWarnings("checkstyle:magicnumber")
```
annym, line 19: Can you please add a general description to this suite. What we are "mocking", what is tested, unit? IT?

xiaoqin2012:
I just realized that we need to check two cases:
1. while the receiver shows it has never done a transfer with all values -1 including the lastLogEntryTimestamp -1, and the sender has log tail -1, in this case, we can just start log entry sync and there is no snapshot sync required. We can do a snapshot sync, but the receiver metadata will not change as there is no data sent over , only snapshot-start and snapshot-end will be sent over.
2. From scenario 1, we can have case that all snapshot related metadata values are -1, but lastLogEntryTimestamp is 100, in this case, if logEntry 101 is not trimmed at the sender, we should start logEntrySync, there is no full snapshot sync is required.

PR #2588, opened 2020-06-25
Log replication leadership flipping testing init

PavelZaytsev:
## Overview

Description:

Leadership change testing init. 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
@@ -161,7 +161,15 @@ public class ServerContext implements AutoCloseable {
     public ServerContext(Map<String, Object> serverConfig) {
         this.serverConfig = serverConfig;
         this.dataStore = new DataStore(serverConfig, this::dataStoreFileCleanup);
-        generateNodeId();
+
+        if (serverConfig.get("--node-id") != null) {
+            UUID serverId = UuidUtils.fromBase64(getServerConfig(String.class, "--node-id"));
+            generateNodeId(serverId);
+        }
+        else {
+            generateNodeId();
+        }
+
         this.failureHandlerPolicy = new ConservativeFailureHandlerPolicy();
 
         // Setup the netty event loops. In tests, these loops may be provided by
@@ -294,9 +302,16 @@ public class ServerContext implements AutoCloseable {
      * Generate a Node Id if not present.
      */
     private void generateNodeId() {
+        generateNodeId(UUID.randomUUID());
+    }
+
+    /**
+     * Generate a Node Id if not present.
+     */
+    private void generateNodeId(UUID nodeId) {
         String currentId = getDataStore().get(NODE_ID_RECORD);
         if (currentId == null) {
-            String idString = UuidUtils.asBase64(UUID.randomUUID());
+            String idString = UuidUtils.asBase64(nodeId);
             log.info("No Node Id, setting to new Id={}", idString);
             getDataStore().put(NODE_ID_RECORD, idString);
         } else {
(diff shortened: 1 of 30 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -160,7 +160,15 @@
     public ServerContext(Map<String, Object> serverConfig) {
         this.serverConfig = serverConfig;
         this.dataStore = new DataStore(serverConfig, this::dataStoreFileCleanup);
-        generateNodeId();
+
+        if (serverConfig.get("--node-id") != null) {
```
annym, line 164: Curious, why is this needed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/DefaultSiteConfig.java
```diff
@@ -31,7 +31,7 @@ private DefaultSiteConfig() {
     private static String primaryCorfuPort = "9000";
 
     @Getter
-    private static String primaryLogReplicationPort = "9010";
+    private static List<String> primaryLogReplicationPorts = Arrays.asList("9010", "9011", "9012");
```
annym, line 34: This file is no longer referenced in the code, can you rebase and please modify in DefaultClusterConfig.java instead.

log-replication/src/main/java/org/corfudb/logreplication/infrastructure/CorfuInterClusterReplicationServer.java
```diff
@@ -46,8 +46,8 @@
                     + "\n"
                     + "Usage:\n"
                     + "\tlog_replication_server [-a <address>|-q <interface-name>] "
-                    + "[-c <ratio>] [-d <level>] [-p <seconds>] "
-                    + "[--base-server-threads=<base_server_threads>] "
+                    + "[-c <ratio>] [-d <level>] [-p <seconds>] [--site-manager-config <config>]"
```
annym, line 49: My understanding is that you are adding the option of passing a file that has a fixed topology configuration, right? ... To avoid adding more flags that are merely for testing, probably what is cleaner is the following:

- We have the flag --plugin, which intention is to specify the location of a file that contains  the class path to the external "adapters", which includes site manager adapter and other plugins (transport and configuration)

So the idea is that there you specify your test adapter. Let's say you create a class that you call UnitTestClusterManagerAdapter. This class will always represent your adapter for testing, and now that class reads from a fixed location the configuration (instead of really querying some external entity), And that such file with the config is this file you are trying to pass. 

If you want to change config in between tests, your adapter is the same, you only need to modify the content that the config is retrieving. That way we avoid bringing in running parameters that are test specific.

log-replication/src/main/java/org/corfudb/logreplication/infrastructure/CorfuInterClusterReplicationServer.java
```diff
@@ -215,12 +219,18 @@ private void startServer() {
...
-
         printStartupMsg(opts);
         configureLogger(opts);
 
         log.info("Started with arguments: {}", opts);
-        this.siteManagerAdapter = constructSiteManagerAdapter();
+
+        if (opts.get("--site-manager-config") != null) {
+            String siteConfigPath = (String) opts.get("--site-manager-config");
+            this.siteManagerAdapter = new DefaultSiteManager(siteConfigPath);
```
annym, line 229: there is a path down the pipeline where the DefaultSiteManager would be created, from the specified --plugin (if that is the class you want to represent your adapter)

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
PavelZaytsev: @annym The current lock holder replication server is identified by its clientId/nodeId. Because I need a way to kill the replication server with a specific clientId -- the one that is holding the lock at the moment, I need to know this information beforehand. Otherwise, the clientId is assigned randomly and I am not sure how to extract this information once the server is running.

annym: Ok, sounds reasonable. Would you mind adding something in this line as a comment in the code for future reference? thanks!

PR #2591, opened 2020-06-30
Fix topology config change logic

zhangn49:
## Overview

Description:

Fix topology config change logic and make CorfuReplicationSiteConfigIT work

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -376,11 +376,11 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private void stopLogReplication() {
         switch(localClusterDescriptor.getRole()) {
             case ACTIVE:
-                log.info("This cluster has lost leadership. Stopping lof replication, according to role {}", localClusterDescriptor.getRole());
+                log.info("This cluster has lost leadership. Stopping log replication, according to role {}", localClusterDescriptor.getRole());
                 replicationManager.stop();
                 break;
             case STANDBY:
-                log.info("This cluster has lost leadership. Stopping lof replication, according to role {}", localClusterDescriptor.getRole());
+                log.info("This cluster has lost leadership. Stopping log replication, according to role {}", localClusterDescriptor.getRole());
                 // Signal Log Replication Server/Sink to stop receiving messages, leadership loss
                 interClusterReplicationService.getLogReplicationServer().setLeadership(false);
                 break;
@@ -448,6 +448,9 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     /**
      * Process a topology change as provided by the Cluster Manager
      *
+     * Note: We are assuming that topology configId change implies a role change.
+     *       The number of standby clusters change would not bump config id.
+     *
      * @param event discovery event
      */
     public void processTopologyChangeNotification(DiscoveryServiceEvent event) {
@@ -478,9 +481,6 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
             updateLocalTopology(newTopology);
             return;
         }
-
-        // TODO: Are we sure that when there is a topologyConfigId change it implies a role change
-        //  and not a new standby added??
 
         // New topology config with higher epoch
         processTopologyConfigChange(newTopology);
@@ -544,12 +544,14 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     }
 
     /**
-     * Query the current all replication stream log tail and remeber the max
+     * Query all replicated stream log tails and remember the max
      * and query each standbySite information according to the ackInformation decide all manay total
      * msg needs to send out.
      */
     @Override
     public void prepareClusterRoleChange() {
+        //TODO  It does not restrict ClusterRole change from standby -> active or active->standby however,
+        // our underlying only process one type. Maybe it's the naming? or revising the actual functionality?
         if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && replicationManager != null) {
             replicationManager.prepareClusterRoleChange();
         } else {
@@ -559,11 +561,12 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     }
 
     /**
-     * Query the current all replication stream log tail and calculate the number of messages to be sent.
-     * If the max tail has changed, give 0%.
+     * Query all replicated stream log tails and calculate the number of messages to be sent.
+     * If the max tail has changed, return 0%.
      */
     @Override
     public int queryReplicationStatus() {
+        //TODO make sure caller should query all nodes in the cluster and pick the max of these 3 values
         if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
             if (!isLeader) {
                 log.warn("Illegal queryReplicationStatus when node is not a leader " +
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -411,30 +411,35 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     }
 
     /**
-     * Process Cluster Role Change:
-     *   - Standby to Active, or
-     *   - Active to Standby
+     * Process Topology Config Change:
+     *   - Higher config id
+     *   - Potential cluster role change
      *
      * @param newTopology new discovered topology
      */
-    public void processClusterRoleChange(TopologyDescriptor newTopology) {
-        // stop ongoing replication, stopLogReplication() checks leadership and active
+    public void processTopologyConfigChange(TopologyDescriptor newTopology) {
+        // Stop ongoing replication, stopLogReplication() checks leadership and active
+        // We do not update topology until we successfully stop log replication
         stopLogReplication();
 
         //TODO pankti: read the configuration again and refresh the LogReplicationConfig object
 
-        // update topology, cluster, and node configs
+        // Update topology, cluster, and node configs
         updateLocalTopology(newTopology);
 
-        // update config id in metadata manager
+        // Update topology config id in metadata manager
+        logReplicationMetadataManager.setupTopologyConfigId(topologyDescriptor.getTopologyConfigId());
+        log.debug("Persist new topologyConfigId {}, cluster id={}, status={}", topologyDescriptor.getTopologyConfigId(),
+                localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
+
+        // Update sink manager
         interClusterReplicationService.getLogReplicationServer().getSinkManager()
                 .updateTopologyConfigId(topologyDescriptor.getTopologyConfigId());
-        log.debug("Persist new topologyConfigId {}, status={}", topologyDescriptor.getTopologyConfigId(),
-                localClusterDescriptor.getRole());
 
+        // Update replication server, in case there is a role change
         logReplicationServerHandler.setActive(localClusterDescriptor.getRole().equals(ClusterRole.ACTIVE));
 
-        // On Cluster Role Change, only if this node is the leader take action
+        // On Topology Config Change, only if this node is the leader take action
         if (isLeader) {
             onLeadershipAcquire();
         }
@@ -446,32 +451,39 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      * @param event discovery event
      */
     public void processTopologyChangeNotification(DiscoveryServiceEvent event) {
-        // Stale notification, skip
+        // Skip stale notification
         if (event.getTopologyConfig().getTopologyConfigID() < topologyDescriptor.getTopologyConfigId()) {
             log.debug("Stale Topology Change Notification, current={}, received={}",
                     topologyDescriptor.getTopologyConfigId(), event.getTopologyConfig());
             return;
         }
 
-        TopologyDescriptor newConfig = new TopologyDescriptor(event.getTopologyConfig());
-
-        // Note: should not update topology here. Otherwise, below check is always true.
-        if (newConfig.getTopologyConfigId() == topologyDescriptor.getTopologyConfigId()) {
+        TopologyDescriptor newTopology = new TopologyDescriptor(event.getTopologyConfig());
+        // Process standby add/remove, which will not increment config id
+        // We won't stop ongoing replications in this case
+        if (newTopology.getTopologyConfigId() == topologyDescriptor.getTopologyConfigId()) {
+            log.debug("Processing a new topology with the same config id, previous topology" +
+                    " is {}, and new topology is {}", topologyDescriptor, newTopology);
             if (localClusterDescriptor.getRole() == ClusterRole.STANDBY) {
                 return;
             }
 
-            // If the current node is active, compare with the current topologyConfig, see if there are additional or
-            // removed standbys
+            // If the current node is active, compare with the current topologyConfig, see if there
+            // are additional or removed standbys
             if (replicationManager != null && isLeader) {
-                replicationManager.processStandbyChange(newConfig);
+                replicationManager.processStandbyChange(newTopology);
             }
-        } else {
-            // TODO: Are we sure that when there is a topologyConfigId change it implies a role change
-            //  and not a new standby added??
-            processClusterRoleChange(newConfig);
+
+            // After processing standby change, update local topology
+            updateLocalTopology(newTopology);
+            return;
         }
 
+        // TODO: Are we sure that when there is a topologyConfigId change it implies a role change
+        //  and not a new standby added??
+
+        // New topology config with higher epoch
+        processTopologyConfigChange(newTopology);
     }
 
     private void updateLocalTopology(TopologyDescriptor newConfig) {
@@ -541,7 +553,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
         if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && replicationManager != null) {
             replicationManager.prepareClusterRoleChange();
         } else {
-            log.warn("Illegal prepareSiteRoleChange when cluster{} with role {}",
+            log.warn("Illegal prepareClusterRoleChange when cluster{} with role {}",
                     localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
         }
     }
@@ -552,15 +564,20 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      */
     @Override
     public int queryReplicationStatus() {
-        // TODO (maxi): address Nan's comments
         if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
-            // If not leader, replication manager might be null
-            if (replicationManager != null) {
-                return replicationManager.queryReplicationStatus();
+            if (!isLeader) {
+                log.warn("Illegal queryReplicationStatus when node is not a leader " +
+                        "in an ACTIVE Cluster{} ", localClusterDescriptor.getClusterId());
+                return 0;
             }
-            log.warn("Illegal queryReplicationStatus when replication manager is null " +
-                    "in an ACTIVE Cluster{} ", localClusterDescriptor.getClusterId());
-            return 0;
+
+            if (replicationManager == null) {
+                log.warn("Illegal queryReplicationStatus when replication manager is null " +
+                        "in an ACTIVE Cluster{} ", localClusterDescriptor.getClusterId());
+                return 0;
+            }
+
+            return replicationManager.queryReplicationStatus();
         } else {
             log.warn("Illegal queryReplicationStatus when cluster{} with role {}",
                     localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -59,6 +59,11 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private static final String LOCK_NAME = "Log_Replication_Lock";
 
     /**
+     * Invalid replication status when query a standby cluster
+     */
+    private static final int INVALID_REPLICATION_STATUS = -1;
+
+    /**
      * Used by the active cluster to initiate Log Replication
      */
     @Getter
@@ -121,6 +126,8 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private boolean shouldRun = true;
 
     private boolean isLeader;
+
+    private LogReplicationServer logReplicationServerHandler;
 
     /**
      * Constructor Discovery Service
@@ -229,7 +236,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
         this.logReplicationMetadataManager = new LogReplicationMetadataManager(getCorfuRuntime(),
                 topologyDescriptor.getTopologyConfigId(), localClusterDescriptor.getClusterId());
 
-        LogReplicationServer logReplicationServerHandler = new LogReplicationServer(serverContext, logReplicationConfig,
+        logReplicationServerHandler = new LogReplicationServer(serverContext, logReplicationConfig,
                 logReplicationMetadataManager, localCorfuEndpoint);
         logReplicationServerHandler.setActive(localClusterDescriptor.getRole().equals(ClusterRole.ACTIVE));
 
@@ -340,7 +347,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      * as source (sender/producer) or sink (receiver).
      */
     private void onLeadershipAcquire() {
-        switch (localNodeDescriptor.getRoleType()) {
+        switch (localClusterDescriptor.getRole()) {
             case ACTIVE:
                 log.info("Start as Source (sender/replicator)");
                 if (replicationManager == null) {
@@ -358,35 +365,27 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
                 break;
             default:
                 log.error("Log Replication not started on this cluster. Leader node {} belongs to cluster with {} role.",
-                            localEndpoint, localNodeDescriptor.getRoleType());
+                            localEndpoint, localClusterDescriptor.getRole());
                 break;
         }
-    }
-
-    private void updateTopologyConfigId(boolean active) {
-        // Required only on topology changes
-        interClusterReplicationService.getLogReplicationServer().getSinkManager().updateTopologyConfigId(active, topologyDescriptor.getTopologyConfigId());
-
-        log.debug("Persist new topologyConfigId {}, status={}", topologyDescriptor.getTopologyConfigId(),
-                localNodeDescriptor.getRoleType());
     }
 
     /**
      * Stop Log Replication
      */
     private void stopLogReplication() {
-        switch(localNodeDescriptor.getRoleType()) {
+        switch(localClusterDescriptor.getRole()) {
             case ACTIVE:
-                log.info("This cluster has lost leadership. Stopping lof replication, according to role {}", localNodeDescriptor.getRoleType());
+                log.info("This cluster has lost leadership. Stopping lof replication, according to role {}", localClusterDescriptor.getRole());
                 replicationManager.stop();
                 break;
             case STANDBY:
-                log.info("This cluster has lost leadership. Stopping lof replication, according to role {}", localNodeDescriptor.getRoleType());
+                log.info("This cluster has lost leadership. Stopping lof replication, according to role {}", localClusterDescriptor.getRole());
                 // Signal Log Replication Server/Sink to stop receiving messages, leadership loss
                 interClusterReplicationService.getLogReplicationServer().setLeadership(false);
                 break;
             default:
-                log.warn("Invalid role type {}. Failed to stop replication if any running.", localNodeDescriptor.getRoleType());
+                log.warn("Invalid role type {}. Failed to stop replication if any running.", localClusterDescriptor.getRole());
                 break;
         }
     }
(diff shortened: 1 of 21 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -497,7 +514,9 @@ public void updateSiteConfig(LogReplicationClusterInfo.TopologyConfigurationMsg
      */
     @Override
     public void prepareSiteRoleChange() {
-        replicationManager.prepareSiteRoleChange();
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && replicationManager != null) {
```
xiaoqin2012, line 517: May add log.warn that this call should not be called on non active site. Also why replicationManager can be null?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -506,7 +525,15 @@ public void prepareSiteRoleChange() {
      */
     @Override
     public int queryReplicationStatus() {
-        return replicationManager.queryReplicationStatus();
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
```
xiaoqin2012, line 528: Add log.warn if the roletype is not active.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: Added. If the replication manager is not initialized yet, it will be null. It happens when the cluster role is standby.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: Yeah, I think I need some clarifications for this function.

Case 1: if the role is standby, what is the expected return value?
Case 2: if the role is active, but somehow replication is not initialized, like have not got the lock. What is the expected value?

Will add log.warn

xiaoqin2012: Case 1, this should not be called on the standby side. As standby side doesn't know the active log tail information. 
But for snapshot sync scenario during applying phase, the active can query standby querySnapshotProcessStatus(), the standby side reply how much data has been appied. This can be done at phase II implementation.
Case 2, do you mean the active is not the leader node? For now, we can consider it should be called on  the leader node only. We need to talk to the other group and get their finalized API and implement the complete solution.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -391,32 +390,50 @@ public void processLockRelease() {
...
 
-        if (localNodeDescriptor.isLeader()) {
+        if (isLeader) {
             stopLogReplication();
-            localNodeDescriptor.setLeader(false);
+            isLeader = false;
         }
     }
 
     public void processSiteFlip(TopologyDescriptor newConfig) {
```
annym, line 399: Let's rename processSiteFlip -> processClusterRoleChange

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -391,32 +390,50 @@ public void processLockRelease() {
...
+
+        logReplicationServerHandler.setActive(localClusterDescriptor.getRole().equals(ClusterRole.ACTIVE));
+
+        // we do not need to update replication manager's config, since it will be initialized again
+
         startLogReplication();
     }
 
     public void processSiteChangeNotification(DiscoveryServiceEvent event) {
         // Stale notification, skip
```
annym, line 428: Same here -> processClusterRoleChangeNotification

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -173,6 +173,7 @@ public LogReplicationSourceManager(CorfuRuntime runtime,
 
         this.logReplicationFSM = new LogReplicationFSM(this.runtime, config, params.getRemoteClusterDescriptor(),
                 dataSender, readProcessor, logReplicationFSMWorkers);
+        this.logReplicationFSM.setSiteConfigID(params.getTopologyConfigId());
```
annym, line 176: let's rename -> setTopologyConfigId

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -196,6 +196,8 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
          // It could be caused by an out-of-date sender or the local node hasn't done the site discovery yet.
          // If there is a siteConfig change, the discovery service will detect it and reset the state.
         if (message.getMetadata().getTopologyConfigId() != siteConfigID) {
+            log.trace("Sink manage with config id {} ignored msg id {}", siteConfigID,
```
annym, line 199: typo -> manage -> manager

test/src/test/java/org/corfudb/integration/CorfuReplicationSiteConfigIT.java
```diff
@@ -342,11 +343,11 @@ public void runSiteSwitch() throws Exception {
             sleep(sleepInterval);
 
             siteManager.prepareSiteRoleChange();
-            while (replicationStatus != CorfuReplicationManager.PERCENTAGE_BASE) {
-                replicationStatus = siteManager.queryReplicationStatus();
-                System.out.print("\nreplication percentage done " + replicationStatus);
-                sleep(sleepInterval);
-            }
+//            while (replicationStatus != CorfuReplicationManager.PERCENTAGE_BASE) {
+//                replicationStatus = siteManager.queryReplicationStatus();
```
annym, line 347: Is this required but still not working? or old code we can remove?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -391,32 +390,50 @@ public void processLockRelease() {
...
+        // stop ongoing replication, stopLogReplication() checks leadership and active
         stopLogReplication();
+
         //TODO pankti: read the configuration again and refresh the LogReplicationConfig object
-        replicationManager.setTopologyDescriptor(newConfig);
-        boolean activeCluster = localNodeDescriptor.getRoleType() == ClusterRole.ACTIVE;
-        updateTopologyConfigId(activeCluster);
+
+        // update local topology descriptor
+        topologyDescriptor = newConfig;
```
annym, line 406: Let's group these three into, updateTopology()

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -497,7 +514,12 @@ public void updateSiteConfig(LogReplicationClusterInfo.TopologyConfigurationMsg
      */
     @Override
     public void prepareSiteRoleChange() {
```
annym, line 516: Let's rename -> prepareClusterRoleChange

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -497,7 +514,12 @@ public void updateSiteConfig(LogReplicationClusterInfo.TopologyConfigurationMsg
...
-        replicationManager.prepareSiteRoleChange();
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && replicationManager != null) {
+            replicationManager.prepareSiteRoleChange();
```
annym, line 518: same here... in ReplicationManager let's rename for consistency to prepareClusterRoleChange

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -506,7 +528,17 @@ public void prepareSiteRoleChange() {
      */
     @Override
     public int queryReplicationStatus() {
-        return replicationManager.queryReplicationStatus();
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
```
annym, line 531: Does this calculate how much is pending or how much is done? Can we fix the method description and rename accordingly.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -506,7 +528,17 @@ public void prepareSiteRoleChange() {
...
+            // If not leader, replication manager might be null
+            if (replicationManager != null) {
+                return replicationManager.queryReplicationStatus();
+            }
+            return 0;
```
annym, line 536: can we add some logging warning, before this return 0. Cause then we don't know if it means that 0 is pending as it all was transferred or the replication never started so 0 is pending.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -506,7 +528,17 @@ public void prepareSiteRoleChange() {
...
+        } else {
+            log.warn("Illegal queryReplicationStatus when cluster{} with role {}",
+                    localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
+            return -1;
```
annym, line 540: Can we make -1 a final static variable INVALID_REPLICATION_STATUS.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -134,7 +134,7 @@ public static TopologyDescriptor readConfig() throws IOException {
         for (int i = 0; i < primaryNodeNames.size(); i++) {
             log.info("Primary Site Name {}, IpAddress {}", primaryNodeNames.get(i), primaryIpAddresses.get(i));
             NodeDescriptor nodeInfo = new NodeDescriptor(primaryIpAddresses.get(i),
-                    activeLogReplicationPort, ClusterRole.ACTIVE, PRIMARY_SITE_NAME, UUID.fromString(primaryNodeIds.get(i)));
+                    activeLogReplicationPort, PRIMARY_SITE_NAME, UUID.fromString(primaryNodeIds.get(i)));
```
annym, line 137: Can we rename PRIMARY_SITE_NAME -> ACTIVE_CLUSTER_NAME

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -145,7 +145,7 @@ public static TopologyDescriptor readConfig() throws IOException {
         for (int i = 0; i < standbyNodeNames.size(); i++) {
             log.info("Standby Site Name {}, IpAddress {}", standbyNodeNames.get(i), standbyIpAddresses.get(i));
             NodeDescriptor nodeInfo = new NodeDescriptor(standbyIpAddresses.get(i),
-                    standbyLogReplicationPort, ClusterRole.STANDBY, STANDBY_SITE_NAME, UUID.fromString(standbyNodeIds.get(i)));
+                    standbyLogReplicationPort, STANDBY_SITE_NAME, UUID.fromString(standbyNodeIds.get(i)));
```
annym, line 148: site -> cluster

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -363,10 +365,9 @@ private boolean receivedValidMessage(LogReplicationEntry message) {
      * 1. update the metadata store with the most recent topologyConfigId
      * 2. reset snapshotWriter and logEntryWriter state
      * 3. reset buffer logEntryBuffer state.
-     * @param active
      * @param siteConfigID
      */
-    public void updateTopologyConfigId(boolean active, long siteConfigID) {
+    public void updateTopologyConfigId(long siteConfigID) {
```
annym, line 370: siteConfigID -> topologyConfigId (in the argument)

test/src/test/java/org/corfudb/integration/CorfuReplicationSiteConfigIT.java
zhangn49: It's working. Brought back.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: I think it calculates how much is pending and have put some comments... I think it would be better to get addressed by another PR.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: Yeah, I understand both cases should not happen right now, and just to avoid NPE. I have added warning logs for both cases.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -436,40 +445,51 @@ public void processClusterRoleChange(TopologyDescriptor newTopology) {
...
-                // removed standbys
-                getReplicationManager().processStandbyChange(newConfig);
-            } else {
-                // TODO: Are we sure that when there is a topologyConfigId change it implies a role change
-                //  and not a new standby added??
-                processClusterRoleChange(newConfig);
+            // If the current node is active, compare with the current topologyConfig, see if there are additional or
+            // removed standbys
+            if (replicationManager != null && isLeader) {
+                replicationManager.processStandbyChange(newConfig);
```
zhangn49, line 467: I see the point that only the leader should do replication related jobs.
But discovery service does not only have topologyConfig, it also has cluster, node... etc.
So we need to make sure all fields get updated, and once any node gets a lock, it has updated configs...

xiaoqin2012: Good catch! This bug is introduced by Anny's new change. The new topology information should be updated on all the nodes, not just the leader.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -419,11 +418,21 @@ public void processLockRelease() {
...
-        replicationManager.setTopology(newTopology);
-        boolean activeCluster = localNodeDescriptor.getRoleType() == ClusterRole.ACTIVE;
-        updateTopologyConfigId(activeCluster);
+
+        // update topology, cluster, and node configs
+        updateLocalTopology(newTopology);
+
+        // update config id in metadata manager
+        interClusterReplicationService.getLogReplicationServer().getSinkManager()
+                .updateTopologyConfigId(topologyDescriptor.getTopologyConfigId());
```
annym, line 431: I think updateTopologyConfigId is not the correct naming in the SinkManager, right? Cause that method has added functionality and is really doing much more than that... maybe processClusterRoleChange or resetOnClusterRoleChange?

annym: Also, similar to stop log replication isn't this dependent of the fact that you are currently a standby?... In such case, it might be a good idea to put this statement in the stopLogReplication method in the case STANDBY, right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: Correct, so one question. When a DISCOVERED_TOPOLOGY comes in (change in the topology) we come in this method processTopologyChangeNotification, only if it is stale we can ignore that incoming topology, otherwise, shouldn't the first thing we do is call your updateLocalTopology() to update all local objects? and maybe you can store the old topology in a temp for the further comparisons.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -517,18 +537,35 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
      * msg needs to send out.
      */
     @Override
-    public void prepareSiteRoleChange() {
-        replicationManager.prepareSiteRoleChange();
+    public void prepareClusterRoleChange() {
```
annym, line 540: I see we have a processClusterRoleChange which comes as part of a DISCOVERED_TOPOLOGY event, but who can directly act upon the discovery service to forcefully prepare for Cluster Role Change? Is this some testing API? we should mark it as such and document it if it's the case.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -517,18 +537,35 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && replicationManager != null) {
+            replicationManager.prepareClusterRoleChange();
+        } else {
+            log.warn("Illegal prepareSiteRoleChange when cluster{} with role {}",
```
annym, line 544: prepareSiteRoleChange -> prepareClusterRoleChange in the log comment.

annym: Can't there be a role change from standby to active?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -517,18 +537,35 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
-     * Query all replicated stream log tails and calculate the number of messages to be sent.
-     * If the max tail has changed, return 0%.
+     * Query the current all replication stream log tail and calculate the number of messages to be sent.
+     * If the max tail has changed, give 0%.
      */
     @Override
     public int queryReplicationStatus() {
         // TODO (maxi): address Nan's comments
-        return replicationManager.queryReplicationStatus();
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
```
annym, line 556: You can remove the first TODO in this method as that you addressed.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -517,18 +537,35 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
+            // If not leader, replication manager might be null
+            if (replicationManager != null) {
```
annym, line 558: I think we should only reply if we are the leader. Cause from their point of view most likely we'll tell them to query all nodes in the cluster and they'll pick the max of these 3 values, otherwise, they won't know who is the leader right now as the lock is something internal.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/CorfuReplicationClusterManagerAdapter.java
```diff
@@ -35,7 +35,7 @@ public synchronized void updateTopologyConfig(TopologyConfigurationMsg newTopolo
     }
 
     public void prepareSiteRoleChange() {
-        corfuReplicationDiscoveryService.prepareSiteRoleChange();
+        corfuReplicationDiscoveryService.prepareClusterRoleChange();
```
annym, line 38: I think we have something loose in this API. It does not restrict ClusterRole change from standby -> active or active->standby however, our underlying only process one type. Maybe it's the naming? or revising the actual functionality?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/LogReplicationFSM.java
```diff
@@ -349,7 +349,7 @@ void transition(LogReplicationState from, LogReplicationState to) {
         to.onEntry(from);
     }
 
-    public void setSiteConfigID(long siteConfigID) {
+    public void setTopologyConfigId(long siteConfigID) {
         this.siteConfigID = siteConfigID;
         snapshotReader.setSiteEpoch(siteConfigID);
```
annym, line 354: now that you are unifying the topologyConfigId, please change it also on. the snapshot and logEntry reader to setTopologyConfigId.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -363,14 +365,13 @@ private boolean receivedValidMessage(LogReplicationEntry message) {
      * 1. update the metadata store with the most recent topologyConfigId
      * 2. reset snapshotWriter and logEntryWriter state
      * 3. reset buffer logEntryBuffer state.
-     * @param active
-     * @param siteConfigID
+     * @param topologyConfigId
      */
-    public void updateTopologyConfigId(boolean active, long siteConfigID) {
-        this.siteConfigID = siteConfigID;
+    public void updateTopologyConfigId(long topologyConfigId) {
```
annym, line 370: this is the method I mentioned up in another comment that we might change the name as it is overloaded with other functionalities.

test/src/test/java/org/corfudb/integration/CorfuReplicationSiteConfigIT.java
```diff
@@ -18,6 +18,7 @@
 public class CorfuReplicationSiteConfigIT extends AbstractIT {
     final static int MAX_RETRY = 10;
     final static long sleepInterval = 10000;
+    private final String pluginConfigFilePath = "src/test/resources/transport/nettyConfig.properties";
```
annym, line 21: Nan would you mind adding a brief description on what this test does. So when someone wants to look at it they know in plain English what it is doing. Maybe something in the line: Bring up an active and standby Log Replication Server, start replication. Initiate a cluster role switch in the middle of (snapshot/log entry) sync, while data is still being persisted into the active datastore (or do you stop the writing before the switch).. is the replication still going on when asked to switch or did it finish already? something in that line so we can easily detect which tests we need to add.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
xiaoqin2012: The function of updateTopologyCongId will update the metadata and sinkManager related information. It should be called on all the type of nodes.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
xiaoqin2012: This is an API we talked about that the upper layer like site manager can call us to prepare a site role change. Now seems that it may not be required. We can get rid of it or just keep it for our testing or warning purpose. The goal of this API is that after this call, the tables that should be replicated should not be changed any more. This hook can tell if the data has been changed or not.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: Yeah, it is cleaner. Besides call updateLocalTopology(), I also need to update other components' config. Like logReplicationServerHandler.setActive

test/src/test/java/org/corfudb/integration/CorfuReplicationSiteConfigIT.java
zhangn49: Yeah, we definitely need a description. I will add descriptions for new tests.

Now any IT run replication as a thread will fail in Travis CI, and the current default cluster manager is not able to change the topology config unless running as a thread. 

So I am introducing a new test cluster manager that subscribes a stream and updates config onNext(). In this way, we can easily control config change in test, and new cluster config IT can run as a process and pass Travis build.

I keep this one to verify config change logic, and will remove it after my new cluster config complete.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: Typo fixed.

Yeah, it can be a role change from standby to active. I think it depends on how the cluster manager uses it...

I don't want to touch prepareClusterRoleChange() and queryReplicationStatus()'s functionality in this PR. I just added some logs and make it a little robust. It is better to have a new PR about those two functions.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/LogReplicationFSM.java
zhangn49: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
zhangn49: Moved metadata manager update to discovery service.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: Added check and logs

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: @annym  @xiaoqin2012  I realized that updateLocalTopology() is dependent on config id change.

1. same config id -> add/remove standby, no role change. Process standby first then update configs

2. higher config id -> might have a role change. Need to stop all replications, which will use current status(role, config)
So we can not pdateLocalTopology() until we stop all replications... Besides, notify the metadata manager that we have a higher config id, update sink manager, and update server's role.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: But only the metadata, right? however, not the snapshotWriter/logEntryWriter. 
Those actually have logic that include calls to the database which are not required if you're not the lead standby, as a non-lead is not really doing anything. There is access to the logReplicationMedataManager at the DiscoveryService level, we can remove the updateTopologyConfigId on the logReplicationMetadata from the sink so it applies to all nodes, and the reset of the writers only if you're the leader, or is there something else that is generic to leaders and non-leaders?

xiaoqin2012: 1. For all the nodes, it needs to update the metadata if config changes. As the metadata is protected by the transaction, at each site, only one node will succeed.
2. The related writer state should be updated too. As the configuration changes, a new full snapshot full sync will be needed. This is to reset the writer state. It is safer to reset writer state at all nodes as the leadership can change at anytime. In my local PR, I am trying to make the writer stateless and only using the persistent metadata.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -517,18 +549,40 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
+        } else {
+            log.warn("Illegal prepareClusterRoleChange when cluster{} with role {}",
+                    localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
+        }
     }
 
     /**
-     * Query all replicated stream log tails and calculate the number of messages to be sent.
-     * If the max tail has changed, return 0%.
+     * Query the current all replication stream log tail and calculate the number of messages to be sent.
```
annym, line 562: typo -> Query all current replication streams log tails ...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: Can you add a TODO in the code please so we don't miss this whenever we decide to merge into master.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: ok, let's leave it for another PR, but we should revisit this.

test/src/test/java/org/corfudb/integration/CorfuReplicationSiteConfigIT.java
annym: ok, yes I was thinking on this test and I have some thoughts on how to run it in processes. I'll share it with you offline I think it will let us model several scenarios very quickly.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
xiaoqin2012: Assume the case that, there is a leadership change from node1 to node2. The leadership acquire event is on the queue for node2's discovery service to process, so node2 think it is not leader. Node1 has got the notification of leadership loss and has processed it and node1 think it is not leader either. There are some intermittent time, there is no node that consider it is the  leader and we don't want to lose the update of the metadata with the new config.

annym: Ok ok, yes in that case even though it is redundant it will protect us in these cases.

PR #2594, opened 2020-07-02
New RPC Layer

Maithem:
## Overview
New RPC Layer

Why should this be merged: Major clean ups, fixes some bugs and adds more tests.

- [x] Public API has Javadoc


common/src/main/java/org/corfudb/common/protocol/client/ChannelHandler.java
```diff
@@ -0,0 +1,315 @@
...
+            }
+            request = requestTimeoutQueue.poll();
+            CompletableFuture requestFuture = pendingRequests.remove(request.requestId);
+            if (requestFuture != null && !requestFuture.isDone()) {
+                requestFuture.completeExceptionally(new PeerUnavailable(remoteAddress));
+            } else {
+                // request is already completed successfully.
+            }
+        }
+    }
```
WenbinZhu, line 220: I haven't started reviewing, just took a glance at the class, so I might be wrong.
It seems that using a Java DelayQueue for checking timeout request is simplier and more effiecnt here. The efficiency comes from the fact that current scheduling impl might have unnecessary checks when a request is just issued, and the checked is immediately scheduled, while the delayqueue blocks until the first request expires. 

Also I'm not understanding why we use eventLoopGroup to drive the scheduled checking here? Is it better to use a separate thread?

WenbinZhu: Looked at it again and I think the problem with current scheduling approach is not extra check, it's not fully repecting the request timeout, the real timeout could be 2x the configured. Since the check interval is same as reqeust timeout, we could face the problem that after the request is put into the queue, right before it times out, we scheduled a check, but think that it's not timed out yet, and wait to to be scheduled next time. However by then the time elapsed is almost 2x the configured request timeout.

Maithem: The reason I didn't use a new thread is simply because the task to "too cheap".

Good catch. I think one way to fix this is to lower the timer resolution for the check. Like check every 100ms/200ms for example. I wanted an easy way to cancel the pending requests (before the timeout expires) that's why I didn't use the scheduled ScheduledExecutorService. Its possible to use the ScheduledExecutorService in conjunction with tracking the returned future task and cancel them from the scheduler, but that uses more than double the memory?

WenbinZhu: @Maithem The way I was thinking is to use a delayQueue, and a dedicated thread to block waiting on the head of the queue which is the first one to be timed out. This involves a spearate thread, which may have some overhead, as this is per channel. I think the approach that the current NettyClientRouter uses might be better, it uses `CFUtils.with()`, which internally uses a global (static) single thread ScheduledExecutorService which is invoked only when the request is timed out (similar to DelayQueue, and actually ScheduledExecutorService internally uses DelayQueue). This approach only requires one extra thread that can monitor all the channels, which is much cheaper.

Maithem: But how can you cancel a task without the timer expiring ?

You need to capture the task future returned from ScheduledExecutorService and store it somewhere, which doubles the memory. Since the tasks are being stored inside the executor and also inside your other data structure.

common/pom.xml
```diff
@@ -57,6 +57,38 @@
...
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <protocVersion>3.11.4</protocVersion>
+                            <includeStdTypes>true</includeStdTypes>
+                            <addProtoSources>inputs</addProtoSources>
+                            <outputDirectory>src/main/java</outputDirectory>
+                            <inputDirectories>
+                                <include>src/main/proto</include>
```
Maithem, line 85: Move this to generated-sources

common/src/main/java/org/corfudb/common/protocol/client/ChannelHandler.java
WenbinZhu: You mean cancel a task before it times out like when channel inactive/disconnected? If this is the case, then  `CFUtils.within()` can still do it, when `sendRequest()` is called, you first generate a task future as usual, put this task future into pendingRequest map, and return to client a composed future, i.e. `return CFUtils.within(task, timeout)`. When you need to cancel a task, just cancel the orignal task in the pendingRequest map, and this composed future will be canceled as well.

Maithem: Yep. There are situations, where the future can be canceled before the time elapses. This is required to let upper layers know quickly, so that they can retry, or make other decisions (i.e. Fault detector). 

That's what I was saying, you need another queue/collection to keep the pending a task future in addition to the structures that are maintained inside the ScheduledExecutorService. My point is why have two mechanisms to expire a request, when you can only have one ?

Also, this reduces object allocation and takes less resources. The current code only needs one append to the queue + poller thread. The ScheduledExecutorService maintains an internal min heap that it needs to maintain. Furthermore, I think the ScheduledExecutorService doesn't release the task if you cancel it until the timer kicks in, so It wont be GC'd as fast. I think that's the default behavior.

WenbinZhu: @Maithem Sorry I didn't get the exrtra collection part, what I was referring to actually reduces explicit collecitons. The CFUtil.within() approach only requires a explicit pendingRequest map, which is already there (plus an implicit queue in the scheduler), while in your current code, you have a pendingRequest map as well as requestTimeoutQueue. So in terms of collections, it's actually the same, while the problem in current implementation is it wastes cpu cycles to check timout every 100~200 ms in order to repect the configured timeout, is that right?

common/src/main/java/org/corfudb/common/protocol/client/ChannelHandler.java
```diff
@@ -0,0 +1,316 @@
...
+                /**
+                 ch.pipeline().addLast(new NettyCorfuMessageDecoder());
+                 ch.pipeline().addLast(new NettyCorfuMessageEncoder());
+                 **/
+
+                /**
+                 * //TODO(Maithem): need to implement new handshake logic without netty pipelines
+                 ch.pipeline().addLast(new ClientHandshakeHandler(parameters.getClientId(),
+                 node.getNodeId(), parameters.getHandshakeTimeout()));
+                 **/
```
zfrenette, line 194: @Maithem Why is it better to implement the new handshake logic outside of the Netty pipeline?

Maithem: I think its cleaner and easier to understand. The current code adds a stage in the pipeline that does the handshake and then removes that stage once the handshake complete. I don't see that as necessary.

Anyways, this can be subjective. But I'm fine with either way.

Maithem:
@zfrenette take this code and run with it, or throw it away.

hisundar:
Hi NSBU-Tools,
Corfu team updates these pom.xml files and the maven-deps scrapes them to derive the bazel CORFU_VERSION. If these pom.xml files are deprecated, can you please share the new method to update the corfu dependency?

Thanks and with warm regards,
Sundar

From: Nate West <westn@vmware.com>
Date: Thursday, August 6, 2020 at 8:50 AM
To: NSBU-tools <NSBU-tools@vmware.com>
Subject: REMINDER: MP pom.xml files will be removed by EOW

REMINDER: MP pom.xml files will be removed from the nsx-highline branch by EOW

As of mid-July 2020, NSX Transformers builds on the nsx-highline branch no longer consume Maven builds.  The pom.xml files will be removed from the nsx/mp directory hierarchy this week.  If you have any additional questions, please reply to the nsbu-tools DL.

hisundar:
Hi Sundar,
Thanks for the question.  I have this change staged that will be merged at the same time.

https://gitreview.eng.vmware.com/c/nsx/+/301523

In the future, you’ll need to update the constants in the mave-deps.bzl file instead of the pom files.
-Nate


From: Sundar Sridharan <sundararamas@vmware.com>
Date: Thursday, August 6, 2020 at 10:46 AM
To: Nate West <westn@vmware.com>, NSBU-tools <NSBU-tools@vmware.com>
Cc: CorfuDB/CorfuDB <reply@reply.github.com>
Subject: Re: REMINDER: MP pom.xml files will be removed by EOW

Hi NSBU-Tools,
Corfu team updates these pom.xml files and the maven-deps scrapes them to derive the bazel CORFU_VERSION. If these pom.xml files are deprecated, can you please share the new method to update the corfu dependency?

Thanks and with warm regards,
Sundar

From: Nate West <westn@vmware.com>
Date: Thursday, August 6, 2020 at 8:50 AM
To: NSBU-tools <NSBU-tools@vmware.com>
Subject: REMINDER: MP pom.xml files will be removed by EOW

REMINDER: MP pom.xml files will be removed from the nsx-highline branch by EOW

As of mid-July 2020, NSX Transformers builds on the nsx-highline branch no longer consume Maven builds.  The pom.xml files will be removed from the nsx/mp directory hierarchy this week.  If you have any additional questions, please reply to the nsbu-tools DL.

PR #2597, opened 2020-07-06
Support any combination of standby/active clusters from Manager

annym:
## Overview

Description:

- Because topology is provided by an external adapter we should support multiple
active/standby clusters or non-active/standby as being provided externally.

Why should this be merged: this is an issue for current Site Manager as no active cluster and multiple standby's are part of valid configurations.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
@@ -1,20 +1,28 @@
 package org.corfudb.infrastructure.logreplication.infrastructure;
 
 import lombok.Getter;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo.TopologyConfigurationMsg;
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo.ClusterRole;
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo.ClusterConfigurationMsg;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
- * This class represents a view of a Multi-Cluster/Site Topology,
- * where one cluster is the active/primary and n cluster's are standby's (backups).
+ * This class represents a view of a Multi-Cluster Topology,
+ *
+ * Ideally, in a given topology, one cluster represents the active cluster (source of data)
+ * while n others are standby clusters (backup's). However, because the topology info is provided by an
+ * external adapter which can be specific to the use cases of the user, a topology might be initialized
+ * with multiple active clusters and multiple standby clusters.
+ *
  */
 @Slf4j
 public class TopologyDescriptor {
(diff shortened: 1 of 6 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
```diff
@@ -1,91 +1,148 @@
...
-        this.standbyClusters = standbyClusters;
+        this.activeClusters = new HashMap<>();
+        this.standbyClusters = new HashMap<>();
+
+        if(activeClusters != null) {
+            activeClusters.forEach(activeCluster -> this.activeClusters.put(activeCluster.getClusterId(), activeCluster));
+        }
+
+        if(standbyClusters != null) {
+            standbyClusters.forEach(activeCluster -> this.standbyClusters.put(activeCluster.getClusterId(), activeCluster));
```
zhangn49, line 91: I think it is better to use 'standbyCluster' as the lambda parameter

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
```diff
@@ -95,11 +152,12 @@ public ClusterDescriptor getClusterDescriptor(String endpoint) {
             }
         }
 
+        log.warn("Endpoint {} does not belong to any cluster defined in {}", clusters);
```
zhangn49, line 155: endpoint missing here.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
```diff
@@ -1,91 +1,148 @@
...
+ * external adapter which can be specific to the use cases of the user, a topology might be initialized
+ * with multiple active clusters and multiple standby clusters.
+ *
  */
 @Slf4j
 public class TopologyDescriptor {
 
     // Represents a state of the topology configuration (a topology epoch)
     @Getter
+    @Setter
```
zhangn49, line 33: Is this setter necessary? I think we should use a new object when we need to update config id...

annym: +1, changed.

zhangn49:
@annym When we convert topologyMsg to topologyDescriptor, those INVALID/NONE/UNRECOGNIZED clusters will be filtered out. So when a node tries to get its local cluster info, it gets nothing. 

Do we need to add a map of invalid clusters in TopologyDescriptor?

annym:
> @annym When we convert topologyMsg to topologyDescriptor, those INVALID/NONE/UNRECOGNIZED clusters will be filtered out. So when a node tries to get its local cluster info, it gets nothing.
> 
> Do we need to add a map of invalid clusters in TopologyDescriptor?

@zhangn49 Yes, Sundar and I were discussing this yesterday. Let me add it as part of this PR as well!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
```diff
@@ -95,11 +160,12 @@ public ClusterDescriptor getClusterDescriptor(String endpoint) {
...
         }
 
+        log.warn("Endpoint {} does not belong to any cluster defined in {}", endpoint, clusters);
         return null;
     }
 
     @Override
     public String toString() {
-        return String.format("Topology[%s] :: Active Cluster=%s :: Standby Clusters=%s", topologyConfigId, activeCluster, standbyClusters);
+        return String.format("Topology[%s] :: Active Cluster=%s :: Standby Clusters=%s", topologyConfigId, activeClusters, standbyClusters);
```
pankti-m, line 169: should we add invalid as well?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -182,55 +181,54 @@ public TopologyConfigurationMsg queryTopologyConfig(boolean useCached) {
...
      */
     public static class SiteManagerCallback implements Runnable {
-        public boolean siteFlip = false;
-        DefaultClusterManager siteManager;
+        public boolean clusterRoleChange = false;
+        DefaultClusterManager clusterManager;
 
         SiteManagerCallback(DefaultClusterManager siteManagerAdapter) {
-            this.siteManager = siteManagerAdapter;
+            this.clusterManager = siteManagerAdapter;
```
pankti-m, line 216: this is confusing.. The way I understand is that clusterManager=siteManager and clusterManagerAdapter=siteManagerAdapter(the plugin).  So how can we assign the adapter to the the clusterManager itself?  I think the variable names need to be changed..

Also, should the adapter be renamed to clusterManagerAdapter?

annym: its just the naming, I'll rename to make it clear!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/TopologyDescriptor.java
annym: Yes we should! adding it!

PR #2600, opened 2020-07-07
Startup TopologyConfigId Fix Receiver

annym:
## Overview

Description:

On receiver startup the topologyConfigId needs to be initialized
or received messages will be filtered out due to different topology
config id when compared to the one in the message.

Why should this be merged: Bug Fix, when starting on a topologyConfigId != 0 all messages are dropped by the receiver.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -14,9 +14,11 @@ import org.corfudb.infrastructure.logreplication.utils.LogReplicationStreamNameT
 import org.corfudb.infrastructure.logreplication.infrastructure.DiscoveryServiceEvent.DiscoveryServiceEventType;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo.ClusterRole;
+import org.corfudb.runtime.exceptions.RetryExhaustedException;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.util.NodeLocator;
 import org.corfudb.util.Sleep;
+import org.corfudb.util.retry.ExponentialBackoffRetry;
 import org.corfudb.util.retry.IRetry;
 import org.corfudb.util.retry.IntervalRetry;
 import org.corfudb.util.retry.RetryNeededException;
@@ -30,6 +32,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * This class represents the Log Replication Discovery Service.
@@ -43,8 +46,10 @@ import java.util.concurrent.LinkedBlockingQueue;
 @Slf4j
 public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicationDiscoveryServiceAdapter {
 
-    private static final int CLUSTER_MANAGER_CONNECT_RETRIES = 10;
-    private static final int CONNECT_SLEEP_DURATION = 5000;
+    /**
+     * Wait interval (in seconds) between consecutive fetch topology attempts to cap exponential back-off.
+     */
+    private static final int FETCH_THRESHOLD = 300;
 
     /**
      * Bookkeeping the topologyConfigId, version number and other log replication state information.
@@ -60,7 +65,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private static final String LOCK_NAME = "Log_Replication_Lock";
 
     /**
-     * Invalid replication status when query a standby cluster
+     * Invalid replication status, when querying a standby cluster
      */
     private static final int INVALID_REPLICATION_STATUS = -1;
 
@@ -126,9 +131,15 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
 
     private boolean shouldRun = true;
 
-    private boolean isLeader;
+    private volatile AtomicBoolean isLeader;
 
     private LogReplicationServer logReplicationServerHandler;
+
+    /**
+     * Indicates the server has been started. A server is started once it is determined
+     * that this node belongs to a cluster in the topology provided by ClusterManager.
+     */
+    private boolean serverStarted = false;
 
     /**
      * Constructor Discovery Service
@@ -146,7 +157,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
         this.serverContext = serverContext;
         this.localEndpoint = serverContext.getLocalEndpoint();
         this.serverCallback = serverCallback;
-        this.isLeader = false;
+        this.isLeader = new AtomicBoolean();
     }
 
     public void run() {
@@ -166,9 +177,6 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
                     }
                 }
             }
-        } catch (LogReplicationDiscoveryServiceException e) {
-            log.error("Exceptionally terminate Log Replication Discovery Service", e);
-            serverCallback.completeExceptionally(e);
         } catch (Exception e) {
             log.error("Unhandled exception caught during log replication service discovery. Retry,", e);
         } finally {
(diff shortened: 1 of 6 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -113,6 +108,7 @@ public LogReplicationSinkManager(String localCorfuEndpoint, LogReplicationConfig
          */
         this.rxState = RxState.LOG_ENTRY_SYNC;
         this.config = config;
+        this.topologyConfigId = topologyConfigId;
```
annym, line 111: I have a test to verify this fix. It will come as part of another PR as it has refactored the CorfuLogReplicationE2ETest to make the methods as utils for other tests.

xiaoqin2012: The clusterManager should query the config and update the config ID properly. Does the clusterManager not start to cause the issue?

annym: The clusterManager does update the persisted topologyConfigId. But within the LogReplicationSinkManager we use the cached value (see for instance the receive method). This cached value is updated on processTopologyConfigChange but not in the initial path (when there is no change but we're starting)... So topologyConfigId on the first run is always 0 instead of the one received by the ClusterManager... We're ok on the update path but fail to set it on the start path...

xiaoqin2012: Then it is better that sinkManager doesn't cache the configureID. If the sinkManager needs the configID, it just queries the metadata table to get the most recent one.

annym: For every received message we verify it is intended for us on that 'epoch'. That would mean accessing the database for every received message.

xiaoqin2012: The data is at runtime, it is a in-memory database, the accessing should be fast. Just one scenario: a node missed the notification of the clusterConfig change due to network partition, this node join the cluster again with the old config and becomes the leader later, will its cached configID be out of date?

annym: One question, why is it an in-memory database? Isn't the metadata table manager doing an access to Corfu itself? which is not running in-memory and which could be doing an RPC call by reading from the tail.

Maybe this scenario that you're highlighting is an indicative that on leadership (lock acquisition) we should re-fetch the topology, cause what if all 3 nodes missed the notification? even if we read from the datastore it is going to rely on an old config. Maybe that's the second change that we should do.

xiaoqin2012: If performance is the concern, we can remove the epoch and epoch check at SinkManager, the snapshotWriter/logEntryWriter will do the check anyway. The checking at sinkManager just brings more complexity. It is nice to pass the epoch down at the startup to the metadata manager.

annym: But that is just moving the problem one layer up, right? cause still we process every message, so for every message we would be accessing the database. 

I would say that maybe we should keep the topologyConfigId cached and it is refreshed on two events:
- On lock acquisition (we need to add this, to avoid the scenario of all nodes missing notification)
- On topology change notification

xiaoqin2012: The writers always do the check no matter what.

annym: But it is a possible case. The writers can continue to do the check always, that is not in question. I'm just concerned of having to access the database on every apply (every message) to check the topologyConfigId. TopologyConfigId changes anyways always come from the discovery service, either because:
(1) the lock was acquired and we check what is the recent topology  (to be added), or
(2) because a changeOfTopologyNotification came in

In this sense, we can always push this value down. Anyways reading from the DB is because we persisted it from the discovery service on (1) or (2) and instead of pushing the new topologyConfigId down, down there we read it from the database. There is where I don't think we need to be doing RPC calls for every received message if we can have it locally by being pushed by the Discovery Service.

xiaoqin2012: I think adding a checking at SinkManager is ok. Just when the message has a higher topology epoch, the current node should send an alarm to DiscoveryService to get the most recent topology information etc. We can address that in another PR.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -396,9 +419,28 @@ private void stopLogReplication() {
...
         log.debug("Process lock acquire event");
         isLeader = true;
+
+        // Regardless of the role of this node, on leadership acquisition
+        // we shall query the topology provider to get the most up to date
+        // topologyConfigId, which might have changed during our time as non-leaders
+        // and for which we might have missed the change notification and update it on
+        // the sink manager so the value is cached to filter messages upon receive.
+        fetchTopologyFromClusterManager();
+        updateTopologyConfigId(topologyDescriptor.getTopologyConfigId());
```
xiaoqin2012, line 429: Just doing updateTopologyConfigID is not enough. While process a topology change event, the discovery service does more work refer to processTopologyChangeNotification. 

While combining two events topology change and leadership change together, it is a bit complex.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -396,9 +419,28 @@ private void stopLogReplication() {
     public void processLockAcquire() {
         log.debug("Process lock acquire event");
         isLeader = true;
+
+        // Regardless of the role of this node, on leadership acquisition
+        // we shall query the topology provider to get the most up to date
+        // topologyConfigId, which might have changed during our time as non-leaders
+        // and for which we might have missed the change notification and update it on
+        // the sink manager so the value is cached to filter messages upon receive.
+        fetchTopologyFromClusterManager();
```
zhangn49, line 428: I think we missed 'logReplicationServerHandler.setActive' after updating localTopology.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -370,6 +370,29 @@ private void onLeadershipAcquire() {
         }
     }
 
+    /**
```
zhangn49, line 373: One more thought is when we process config change event notification, we are using the config from that notification message, which might be delayed/outdated. I was wondering if we can fetch latest one to skip those outdated ones?

xiaoqin2012: If there is another change, it should send another notification. We should stick to one model, either push(get notified) or poll(query). If we sometime use "notification", sometime use "query", it complicate the system.

annym: mmm but that would be really complex right, the same argument could be thought for the one being fetched (that it could be delayed/outdated). Anyways, if it is delayed or outdated we'll soon receive another notification.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: you're right! So, to avoid this being lose and in different places, I refactored a bit as for updates to always happen when verifying if the cluster belongs to the retrieved topology.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
xiaoqin2012: While processing a topology event, it is an optimization we can do to query the most recent topology. But while there is a leadership change, there is no need to query the topology to complicate the scenario.

PR #2603, opened 2020-07-07
Handle the case where Version Table is empty.

pankti-m:
If the version table is dropped using the CorfuStoreBrowser(UFO) for testing, it will be empty and should be re-created.


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/utils/LogReplicationStreamNameTableManager.java
```diff
@@ -124,8 +124,14 @@ private boolean tableVersionMatchesPlugin() {
         corfuStore.openTable(CORFU_SYSTEM_NAMESPACE, LOG_REPLICATION_PLUGIN_VERSION_TABLE);
         LogReplicationStreams.VersionString versionString = LogReplicationStreams.VersionString.newBuilder().setName("VERSION").build();
         Query q = corfuStore.query(CORFU_SYSTEM_NAMESPACE);
+
+        // If the version table is dropped using the CorfuStoreBrowser(UFO) for testing, it will be empty.
+        // In this case, it should be re-created
+        if (q.getRecord(LOG_REPLICATION_PLUGIN_VERSION_TABLE, versionString) == null) {
+            return false;
```
xiaoqin2012, line 131: Add a log message for this case.

PR #2605, opened 2020-07-08
Add suspend and resume API for lock client.

xiaoqin2012:
* Suspend an interest and enforce the lock client entering NoLeaseState.
  * Resume an interest.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/utils/src/main/java/org/corfudb/utils/lock/LockClient.java
+++ b/utils/src/main/java/org/corfudb/utils/lock/LockClient.java
@@ -18,6 +18,7 @@ import org.corfudb.utils.lock.LockDataTypes.LockId;
 import org.corfudb.utils.lock.persistence.LockStore;
 import org.corfudb.utils.lock.states.LockEvent;
 
+import javax.print.attribute.standard.SheetCollate;
 import java.lang.reflect.InvocationTargetException;
 import java.util.*;
 import java.util.concurrent.*;
@@ -48,9 +49,16 @@ public class LockClient {
     // single threaded scheduler to monitor locks
     private final ScheduledExecutorService lockMonitorScheduler;
 
+    // single threaded scheduler for lock listener.
+    private final ExecutorService lockListenerExecutor;
+
+    // single threaded scheduler to process tasks;
+    private final ScheduledExecutorService taskScheduler;
+
+
     // duration between monitoring runs
     @Setter
-    private static int DurationBetweenLockMonitorRuns = 60;
+    private static int DurationBetweenLockMonitorRuns = 10;
 
     // The context contains objects that are shared across the locks in this client.
     private final ClientContext clientContext;
@@ -60,6 +68,9 @@ public class LockClient {
 
     @Getter
     private UUID clientId;
+
+    private Lock lock;
+
 
     /**
      * Constructor
@@ -73,7 +84,7 @@ public class LockClient {
     //TODO need to determine if the application should provide a clientId or should it be internally generated.
     public LockClient(UUID clientId, CorfuRuntime corfuRuntime) throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
 
-        ScheduledExecutorService taskScheduler = Executors.newScheduledThreadPool(1, (r) ->
+        this.taskScheduler = Executors.newScheduledThreadPool(1, (r) ->
         {
             Thread t = Executors.defaultThreadFactory().newThread(r);
             t.setName("LockTaskThread");
@@ -81,7 +92,7 @@ public class LockClient {
             return t;
         });
 
-        ExecutorService lockListenerExecutor = Executors.newFixedThreadPool(1, (r) ->
+        this.lockListenerExecutor = Executors.newFixedThreadPool(1, (r) ->
         {
             Thread t = Executors.defaultThreadFactory().newThread(r);
             t.setName("LockListenerThread");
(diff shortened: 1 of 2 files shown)


utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -154,6 +183,30 @@ private void monitorLocks() {
...
+                    log.error("Lock: Could not cancel the future for lease renewal!!!", lock.getLockId());
+                }
+            }
+            log.warn("Lock: canceled the future for monitor locks!!!", lock.getLockId());
+        }
+    }
+
+
+    public void stop() {
+        lock.input(LockEvent.UNRECOVERABLE_ERROR);
```
pankti-m, line 204: The stop() could have been called in a non-error scenario also.  Can we add another event instead of unrecoverable error?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -154,6 +183,30 @@ private void monitorLocks() {
...
+        taskScheduler.shutdownNow();
```
pankti-m, line 205: why do we call shutdown()?  a clean way to shutdown the executorservice is to call shutdown() and then shutdownnow() - https://www.baeldung.com/java-executor-service-tutorial

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -154,6 +183,30 @@ private void monitorLocks() {
         ));
     }
 
+
+    /**
+     * Stop monitoring the locks.
+     */
+    private void stopMonitorLocks() {
+        synchronized (lockMonitorFuture) {
```
pankti-m, line 191: what if lockMonitorFuture has not yet been initialized?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -154,6 +183,30 @@ private void monitorLocks() {
...
+            if (lockMonitorFuture.isPresent() && !lockMonitorFuture.get().isDone()) {
+                boolean cancelled = lockMonitorFuture.get().cancel(false);
+                if (!cancelled) {
+                    log.error("Lock: Could not cancel the future for lease renewal!!!", lock.getLockId());
```
pankti-m, line 195: cancelled can be false even if the future has already completed.  In that case, it is not an error.  Can this be changed to debug or trace?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -61,6 +69,9 @@
     @Getter
     private UUID clientId;
 
+    private Lock lock;
```
pankti-m, line 72: why are we making this a separate instance variable?  All locks are stored in the hashmap "locks"

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -129,6 +140,24 @@ public void registerInterest(@NonNull String lockGroup, @NonNull String lockName
         monitorLocks();
     }
 
+    /**
+     * Enforce a transition to NoLeaseState and stop monitoring.
+     **/
+    public void suspendInterest() {
+        log.info("LockClient {} suspendInterest.", clientId);
+        stopMonitorLocks();
```
pankti-m, line 148: stopMonitorLocks() only works on locks with expired lease.  Here we want to revoke lease on active locks.  So we need to generate an event "lease_revoked".. right?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -129,6 +140,24 @@ public void registerInterest(@NonNull String lockGroup, @NonNull String lockName
         monitorLocks();
     }
 
+    /**
+     * Enforce a transition to NoLeaseState and stop monitoring.
+     **/
+    public void suspendInterest() {
```
pankti-m, line 146: do we need this functionality or is it only for testing?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -129,6 +140,24 @@ public void registerInterest(@NonNull String lockGroup, @NonNull String lockName
...
+        log.info("LockClient {} suspendInterest.", clientId);
+        stopMonitorLocks();
+        lock.input(LockEvent.LEASE_REVOKED);
+    }
+
+    /**
+     * Resume the interest inserting an LEASE_REVOIKED event and start monitoring.
+     **/
+    public void resumeInterest() {
```
pankti-m, line 155: do we need this functionality or is it only for testing?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
xiaoqin2012: While lockclient suspend the lock, it will need to submit an event through the lock.

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
xiaoqin2012: Now for testing purpose. It is possible in future if we detect a slow node that holding the lock, we can enforce it to release the lock.

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
xiaoqin2012: monitor thread monitors active locks too.

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
xiaoqin2012: Same as above.

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
xiaoqin2012: I did shutdown() first, it generated a lot of exceptions. For this case, shutdownnow() seems work better.

PR #2607, opened 2020-07-08
Add Logging around Negotiation Response Processing

annym:
## Overview

Add logging around negotiation response processing for debugging.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
@@ -72,6 +72,7 @@ public class NegotiatingState implements LogReplicationRuntimeState {
                 fsm.updateConnectedEndpoints(event.getEndpoint());
                 return null;
             case NEGOTIATION_COMPLETE:
+                log.info("Negotiation complete, result={}", event.getNegotiationResult());
                 ((ReplicatingState)fsm.getStates().get(LogReplicationRuntimeStateType.REPLICATING)).setReplicationEvent(event.getNegotiationResult());
                 return fsm.getStates().get(LogReplicationRuntimeStateType.REPLICATING);
             case NEGOTIATION_FAILED:
@@ -142,6 +143,9 @@ public class NegotiatingState implements LogReplicationRuntimeState {
      */
     private void processNegotiationResponse(LogReplicationNegotiationResponse negotiationResponse)
             throws LogReplicationNegotiationException {
+
+        log.debug("Process negotiation response {} from {}", negotiationResponse, fsm.getRemoteClusterId());
+
         /*
          * If the version are different, report an error.
          */
@@ -182,12 +186,13 @@ public class NegotiatingState implements LogReplicationRuntimeState {
          * "topologyConfigId": "10"
          * "version": "release-1.0"
          * "snapshotStart": "-1"
-         * "snapshotSeqNum": " -1"
+         * "snapshotSeqNum": "-1"
          * "snapshotTransferred": "-1"
          * "snapshotApplied": "-1"
          * "lastLogEntryProcessed": "-1"
          */
         if (negotiationResponse.getSnapshotStart() == -1) {
+            log.info("No snapshot available in remote. Initiate SNAPSHOT sync to {}.", fsm.getRemoteClusterId());
             negotiationResponse.getLastLogProcessed();
             fsm.input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.NEGOTIATION_COMPLETE,
                     new LogReplicationEvent(LogReplicationEvent.LogReplicationEventType.SNAPSHOT_SYNC_REQUEST)));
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -256,11 +264,20 @@ private void processNegotiationResponse(LogReplicationNegotiationResponse negoti
...
              */
             if (logHead <= negotiationResponse.getLastLogProcessed() + 1) {
+                log.info("Resume LOG ENTRY sync. Address space has not been trimmed, deltas are guaranteed to be available. " +
+                        "logHead={}, lastLogProcessed={}", logHead, negotiationResponse.getLastLogProcessed());
                 fsm.input(new LogReplicationRuntimeEvent(LogReplicationRuntimeEvent.LogReplicationRuntimeEventType.NEGOTIATION_COMPLETE,
                         new LogReplicationEvent(LogReplicationEvent.LogReplicationEventType.REPLICATION_START,
                                 new LogReplicationEventMetadata(LogReplicationEventMetadata.getNIL_UUID(), negotiationResponse.getLastLogProcessed())
                         )));
             } else {
+                // TODO: it is OK for a first phase, but this might not be efficient/accurate, as the next (+1)
```
annym, line 274: @xiaoqin2012 can you please confirm this TODO note? To be sure keeping it around for future work.

xiaoqin2012: This is not written by me. Also the statement doesn't sound correct. If there are logs trimmed, we don't know if there are entries belong to the replicated tables or not. To be correct and safe, a full snapshot sync should be requested.

annym: It was written by me, as I was adding the logging and had never reviewed this code. I just wanted to have your input on this specific point. 
My point here is that getLastLogProcessed + 1 is not necessarily the next entry, right? so though getLastLogProcessed+1 falls in the trimmed space, maybe the next entry is getLastLogProcessed+2000 ( as this is a global address, which does not fall in the trimmed space). So, for better accuracy in the future (not now) we can leave this note around so we check the real next entry and avoid doing a full snapshot sync when not really required.

xiaoqin2012: Last processed is 10, now the log has been trimmed and the log head is 20. How do you know the entries between 11 and 19 contains the data to be replicated or not?

annym: We inspect txStream.next()and see the global address of the one coming next... and if it is 16, OK let's do snapshot sync. But if the next is 56, no need to.

xiaoqin2012: After trim, the log head is 20, the txStream's head at least will be 20 or bigger, right?
assume txStream has entries at 16, 18, 22. After trim, txStream's head will be 22, the next() call will give 22, right?

annym: mm, I see your point. But now I have a follow up question, the bitmaps keep track of the last trimmed address, so despite this scenario we could know that head is 22 and last trimmed address is 18. Again we don't need this now but I'm seeing if its an option.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -142,6 +143,9 @@ public void setLeaderEndpoint(String endpoint) {
      */
     private void processNegotiationResponse(LogReplicationNegotiationResponse negotiationResponse)
             throws LogReplicationNegotiationException {
+
+        log.debug("Process negotiation response {} from {}", negotiationResponse, fsm.getRemoteClusterId());
```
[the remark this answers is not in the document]
annym: I also thought of this, but negotiation doesn't happen that frequent so I thought it would be useful to actually have it.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/NegotiatingState.java
```diff
@@ -200,14 +205,14 @@ private void processNegotiationResponse(LogReplicationNegotiationResponse negoti
...
-         * "snapshotSeqNum": " 88"
+         * "snapshotSeqNum": "88"
          * "snapshotTransferred": "-1"
          * "snapshotApplied": "-1"
          * "lastLogEntryProcessed": "-1"
          */
         if (negotiationResponse.getSnapshotStart() > negotiationResponse.getSnapshotTransferred()) {
-            log.info("Get the negotiation response {} and will start replication.",
-                    negotiationResponse);
+            log.info("Previous Snapshot Sync transfer did not complete. Restart SNAPSHOT sync, snapshotStart={}, snapshotTransferred={}",
```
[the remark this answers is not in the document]
annym: It can be greater because the transfer did not complete. It started (was set at that moment) but did not complete (so the transferred remains -1).

PR #2610, opened 2020-07-10
Parameterize Replication Configurations

annym:
## Overview

Description:

-Snapshot Sync batch size (number of entries sent on a batch for snapshot sync)
-Lock Lease Duration


annym:
> need to ensure that these parameters are reflected in conf files that start up log-replication-server

Yes. @pankti-m can you help with this on the deploy part? Thanks!

Maithem:
Docopt + Lombok's @Data is really messy and hard to maintain. Please consider using https://commons.apache.org/proper/commons-configuration/userguide/howto_filebased.html

We are already moving to file-based configuration #2315 and it would be nice for new code to adopt that.

annym:
> Docopt + Lombok's @DaTa is really messy and hard to maintain. Please consider using https://commons.apache.org/proper/commons-configuration/userguide/howto_filebased.html
> 
> We are already moving to file-based configuration #2315 and it would be nice for new code to adopt that.

@Maithem We are using a couple of file based configurations for other LR functions. As for these server things, we need a quick fix for a first testing release, but we can definitely adopt it for the second phase and actually decouple from corfu server configurations.

PR #2612, opened 2020-07-10
Unify Read APIs

Maithem:
## Overview
Merged the multi-address read with the single read API.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -326,7 +308,8 @@ public void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerCo
             }
             r.sendResponse(ctx, msg, CorfuMsgType.READ_RESPONSE.payloadMsg(rr));
         } catch (DataCorruptionException e) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_CORRUPTION.msg());
```
Maithem, line 329: This is actually a bug on master, if a multi read encounters a data corruption exception, it will keep timing out. This patches fixes this bug.

zhangn49: Why it will keep timing out? Because it does not pass the current read address?

Maithem: Yes, the message is invalid without passing a long, so the server will fail every time i tries to respond with that message type. The the client's request will start timing out and retrying.

Maithem: Another issue that makes it hard to debug is exceptions throwing inside the netty pipeline stages are not propagated to the producer of the channel write (i.e. client sending message/server sending a response).

WenbinZhu: @Maithem I was helping Chetan debugging a similar issue yesterday where pipeline exceptions were not propagated to producer, failing silently. Eventually I found the problem when waiting on the future returned by channel.writeAndFlush(). So I think in the routers, we need to attach a listener to the future returned by channel.writeAndFlush(). For server's router, the listener can log and error if it fails, on client side, the returned future can be chained with this writeAndFlush future. We can do this in the new RPC project.

Maithem: Yep, its annoying.

The void promise was apparently done for "performance" reasons, but I remember removing it last year and I couldn't see any performance difference.

We need to do something like this:

```java

channel.writeAndFlush(sendBuf).addListener(res -> {
            if (!res.isSuccess()) {
                // fail with unavailable?
                CompletableFuture<TurboAPI.BaseCommand> resp = pendingRequests.remove(requestId);
                if (resp != null && !resp.isDone()) {
                    // res.cause()
                    resp.completeExceptionally(res.getThrowable());
                }
            }
        });
```

Maithem:
> inspectAddresses also sends an invalid `CorfuMsgType.ERROR_DATA_CORRUPTION.msg()`

Good point, thought I already changed it.

Maithem:
> inspectAddresses also sends an invalid `CorfuMsgType.ERROR_DATA_CORRUPTION.msg()`

Fixed and added a new test.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -288,55 +288,46 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
...
-        log.trace("inspectAddresses: {}", msg.getPayload().getAddresses());
+        List<Long> addresses = msg.getPayload().getAddresses();
+        log.trace("inspectAddresses: {}", addresses);
         InspectAddressesResponse inspectResponse = new InspectAddressesResponse();
+
+    long currentAddress = Address.NON_ADDRESS;
         try {
-            for (long address : msg.getPayload().getAddresses()) {
+            for (long address : addresses) {
+                currentAddress = address;
```
WenbinZhu, line 330: `address` is primitive long, do we really need `currentAddress`?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -288,55 +288,46 @@ private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext
...
-                }
-            }
-            r.sendResponse(ctx, msg, CorfuMsgType.READ_RESPONSE.payloadMsg(rr));
-        } catch (DataCorruptionException e) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_CORRUPTION.msg());
+    ReadResponse rr = new ReadResponse();
+    long currentRead = Address.NON_ADDRESS;
+    try {
+      for (Long address : msg.getPayload().getAddresses()) {
+        currentRead = address;
```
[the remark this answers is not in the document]
Maithem: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java

Issue #2611, opened 2020-07-10
Flush Logger On Exit


Maithem:
**Bug report.**
The server's buffered logger needs to flush its logs before exit (via shutdown hook), otherwise critical logs can be lost.

PR #2613, opened 2020-07-10
Flush Async Logger on Exit

Maithem:
## Overview
If not flushed, some crucial logs can be lost.

Why should this be merged: Improves debuggability 

Related issue(s) (if applicable): #2611

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
@@ -4,6 +4,7 @@ import static org.corfudb.util.NetworkUtils.getAddressFromInterfaceName;
 
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.core.joran.spi.JoranException;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
@@ -346,7 +347,18 @@ public class CorfuServer {
     private static void cleanShutdown() {
         log.info("CleanShutdown: Starting Cleanup.");
         shutdownServer = true;
-        activeServer.close();
+        try {
+            CorfuServerNode current = activeServer;
+            if (current != null) {
+                current.close();
+            }
+        } catch (Throwable th) {
+            log.error("cleanShutdown: failed during shutdown", th);
+        }
+
+        // Flush the async appender before exiting to prevent the loss of logs
+        LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
+        loggerContext.stop();
     }
 
     /**
--- a/infrastructure/src/main/resources/logback.prod.xml
+++ b/infrastructure/src/main/resources/logback.prod.xml
@@ -1,9 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration scan="true" scanPeriod="30 seconds">
 
-    <!-- https://logback.qos.ch/manual/configuration.html#stopContext -->
-    <shutdownHook/>
-
     <property name="LOG_DIRECTORY" value="/var/log/corfu" />
 
     <appender name="file" class="ch.qos.logback.core.rolling.RollingFileAppender">
@@ -27,6 +24,7 @@
     <appender name="async_file" class="ch.qos.logback.classic.AsyncAppender">
         <appender-ref ref="file" />
         <queueSize>1024</queueSize>
+        <maxFlushTime>5000</maxFlushTime>
     </appender>
 
     <root level="info">


infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -346,7 +347,18 @@ static void restartServer(boolean resetData) {
     private static void cleanShutdown() {
         log.info("CleanShutdown: Starting Cleanup.");
         shutdownServer = true;
-        activeServer.close();
+        try {
+            CorfuServerNode current = activeServer;
```
xiaoqin2012, line 351: Why do you need a local variable for this?

Maithem: activeServer is a volatile and can be re-assigned by multiple threads.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -346,7 +347,18 @@ static void restartServer(boolean resetData) {
...
+            if (current != null) {
+                activeServer.close();
+            }
+        } catch (Throwable th) {
+            log.error("cleanShutdown: failed during shutdown", th);
+        }
+
+        // Flush the async appender before exiting to prevent the loss of logs
+        LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
```
xnull, line 360: This doesn't help at all, I suspect it can even cause loss of the logs.
Logback already have solved the problem, and we have it enabled https://github.com/CorfuDB/CorfuDB/blob/30c8d154f59dfe22d4cc16b94ab9e2c51851c2e5/infrastructure/src/main/resources/logback.prod.xml#L5

Maithem: This is not the Matrix, but things are not what they seem to be.

The shutdown hook configuration added in #2049 is malformed and actually doesn't register a hook that stops the logger's context on shutdown. This is true despite what logback's documentation says, it might be outdated, there might be a bug, or it might just be inaccurate. At least for version 1.2.3 it doesn't work.

For a shutdown hook to be called, a shutdown thread must be registered with `Runtime.getRuntime().addShutdownHook` for it to be executed. If you run the server and collect a heap dump, you can inspect all the shutdown hooks registered and you won't see any hooks registered by logback. This means that the logger context is not being stopped properly on shutdown, which explains some of the log loss I've seen debugging a couple of bugs. 

Now even if that configuration works, its not the best way of implementing "shutdown" logic for a logger because there is no order on shutdown hooks. This means that logging statements produced during the execution of the shutdown hook can be lost this is due to an undefined execution order of the hooks. If the logback shutdown hook is executed first  then subsequent logs will be lost from other hooks.

If you are not convinced, then you can inspect logback's source code here: https://github.com/qos-ch/logback/blob/a154cd1b564d436c90a26b8cb1a2e8ffff0a4a47/logback-core/src/main/java/ch/qos/logback/core/joran/action/ShutdownHookAction.java#L46

If you inspect the code snippet I linked you'll see that it tries to find a `class` attribute, but fails to do so (because it doesn't exist). This means that even the "default" hook isn't being registered. 

Furthermore, the documentation you link about the default hook is incorrect, if you checkout logback's code base version 1.2.3 you'll see that a `DefaultShutdownHook` type doesn't even exist.

Finally, to fix the configuration bug introduced in #2049  the class and full type path must be supplied. For example, the correct configuration for installing a default hook would be

 `<shutdownHook class="ch.qos.logback.core.hook.DelayingShutdownHook"/>`


Anyways, I think this PR fixes an existing problem on master and improves logging in general (via explicit shutdown ordering) and I think we should go forward with it.

xnull: So,
- let's add DelayingShutdownHook or any other class that is required to the shutdownHook?
- what I see from source code is all the hooks are run in parallel, which means no hooks wait for each other. If you do not shutdown JVM then no reason for shutting down the logger. If the JVM gonna fail you will lose your logs, either way, it doesn't matter if you shutdown it by a hook or explicitly in your code, the effect will be the same.

I still don't see the motivation for adding the code that shouldn't be added. because we have a solution provided by the library of the box.
That's great if you found a misconfiguration bug in the config, let's fix the config?

image

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -346,7 +347,18 @@ static void restartServer(boolean resetData) {
     private static void cleanShutdown() {
         log.info("CleanShutdown: Starting Cleanup.");
         shutdownServer = true;
-        activeServer.close();
+        try {
+            CorfuServerNode current = activeServer;
+            if (current != null) {
+                activeServer.close();
```
xnull, line 353: it has to be `current ` instead of `activeServer `

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
xnull, line 351: what about - replace `volatile` by AtomicReference? Then your code will be cleaner, no need for checking for null and bringing additional mental complexity to the code?

xnull: Ok, looks like AtomicReference doesn't have good API for that. Nevermind

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
Maithem: > I still don't see the motivation for adding the code that shouldn't be added


Please read my response again. 

Even if you ignore the logback bug, the PR doesn't re-implement the "out-of-box" behavior. This PR adds execution order which I already describe in detail.


>  hooks are run in parallel, which means no hooks wait for each other.


I never claimed that. I explicitly mention **undefined execution order** , which is the whole point of having one hook and having serial order, as opposed to having two hooks and no order.

xnull: Ok, I see your point

infrastructure/src/main/resources/logback.prod.xml
```diff
@@ -27,6 +27,7 @@
     <appender name="async_file" class="ch.qos.logback.classic.AsyncAppender">
         <appender-ref ref="file" />
         <queueSize>1024</queueSize>
+        <maxFlushTime>5000</maxFlushTime>
```
WenbinZhu, line 30: Should we remove this line: https://github.com/CorfuDB/CorfuDB/blob/master/infrastructure/src/main/resources/logback.prod.xml#L5 , since we are doing it elsewhere and it's missing class name as well?

Maithem: Yep, I think we should keep maxFlushTime

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
Maithem: Yep.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
Maithem: Ok.

Maithem:
@xnull Please complete your review

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -346,7 +347,18 @@ static void restartServer(boolean resetData) {
     private static void cleanShutdown() {
         log.info("CleanShutdown: Starting Cleanup.");
         shutdownServer = true;
-        activeServer.close();
+        try {
+            CorfuServerNode current = activeServer;
+            if (current != null) {
+                current.close();
```
xnull, line 353: In the best case would be great to get rid of a volatile variable completely. Probably next time

Maithem: I'm planning to get rid of CorfuServerNode its unnecessarily complex, with that the volatile will disappear.

PR #2616, opened 2020-07-11
Prevent file descriptor leaks

xnull:
## Overview

Description:
Prevent file descriptor leaks when restart/reset Corfu server

Why should this be merged: 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1219,7 +1219,18 @@ public void close() {
...
+        for (FileChannel channel : channelsToSync) {
+            try {
+                channel.force(true);
+            } catch (IOException e) {
+                log.debug("Can't force updates in the channel {}", e.getMessage());
+            } finally {
+                IOUtils.closeQuietly(channel);
+            }
+        }
+
```
Maithem, line 1231: `channelsToSync` is a subset of of `writeChannels` what's the point of closing them again ?

it/src/test/java/org/corfudb/universe/scenario/FileDescriptorLeaksIT.java
```diff
@@ -0,0 +1,68 @@
...
+
+import static org.corfudb.universe.scenario.fixture.Fixtures.TestFixtureConst.DEFAULT_STREAM_NAME;
+import static org.junit.jupiter.api.Assertions.fail;
+
+@Slf4j
+public class FileDescriptorLeaksIT extends GenericIntegrationTest {
+
+    /**
+     * Test cluster behavior after one node down.
+     * <p>
```
Maithem, line 23: I think this test is too verbose to what it's trying to check. You can write a unit test and use `OperatingSystemMXBean` to check the number of open file descriptors. Its faster and more succinct.

xnull: `lsof` is a perfectly fine command, I don't see any needs to use anything else.
 - The test is checking that there are no file descriptor leaks, I don't see why we need a unit test here. 
 - it uses docker containers which guaranties same result on different operation systems, which is not possible with a unit test (in case of lsof).
 - I believe lsof is more solid approach than anything else including JMX.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: Because it's a pretty fragile structure - having two collections that could or couldn't contain same elements(instances of filechannels). 

Since it's not guaranteed at compile time that `writeChannels` is always a superset of `channelsToSync` it's safe to close those descriptors (just in case) even if we know that writeChannels should be a superset of channelsToSync.

However, I'm ok not to add this change.

it/src/test/java/org/corfudb/universe/scenario/FileDescriptorLeaksIT.java
xnull: Also if you are talking about [UnixOperatingSystemMXBean](https://docs.oracle.com/javase/8/docs/jre/api/management/extension/index.html?com/sun/management/UnixOperatingSystemMXBean.html )
unfortunately, it doesn't provide enough information regarding what files are open and their statuses.

Maithem:
@xnull  Is this ready for review?

xnull:
> @xnull Is this ready for review?

yes, most of the code has finished, I'm fixing rest of failing tests

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -182,8 +182,6 @@
 
     // Flag if set to true - causes the Corfu Server to shutdown.
     private static volatile boolean shutdownServer = false;
-    // If set to true - triggers a reset of the server by wiping off all the data.
-    private static volatile boolean cleanupServer = false;
```
Maithem, line 186: Nice.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -251,11 +249,6 @@ private static void startServer(String[] args) {
                 System.exit(EXIT_ERROR_CODE);
             }
 
-            if (cleanupServer) {
-                clearDataFiles(serverContext);
-                cleanupServer = false;
-            }
-
```
Maithem, line 258: Move this logic to be executed by a handler thread and not the main thread can introduce a race. Although not clean, this race doesn't seem to be present on master.

The handler thread signaled to the main thread to shutdown and do the clean up.

But what can happen here is, as you are deleting the state, the main thread will break out of the `activeServer.startAndListen()`  and try to create a new server context while the data is being deleted by the handler thread in the `restartServer` method.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -332,12 +325,11 @@ private static void clearDataFiles(ServerContext serverContext) {
...
     static void restartServer(boolean resetData) {
 
-        if (resetData) {
-            cleanupServer = true;
-        }
-
         log.info("RestartServer: Shutting down corfu server");
         activeServer.close();
+        if (resetData) {
+            clearDataFiles(activeServer.getServerContext());
```
Maithem, line 331: There's a race between this and the main thread. While data is being deleted a new context can be created at the same time.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -332,12 +325,11 @@ private static void clearDataFiles(ServerContext serverContext) {
      */
     static void restartServer(boolean resetData) {
 
-        if (resetData) {
-            cleanupServer = true;
-        }
-
         log.info("RestartServer: Shutting down corfu server");
         activeServer.close();
+        if (resetData) {
```
Maithem, line 330: Can you change this variable name. Reset and restart have been confusing to many.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
```diff
@@ -136,30 +139,36 @@ public void close() {
 
         // A executor service to create the shutdown threads
         // plus name the threads correctly.
-        final ExecutorService shutdownService = Executors.newFixedThreadPool(serverMap.size(),
-                new ServerThreadFactory("CorfuServer-shutdown-",
-                        new ServerThreadFactory.ExceptionHandler()));
+        final ExecutorService shutdownService = Executors.newFixedThreadPool(
+                serverMap.size(),
+                new ServerThreadFactory("CorfuServer-shutdown-", new ExceptionHandler())
+        );
```
Maithem, line 145: Honestly having this seems like an overkill. Whoever wrote this probably was trying to use the same design as the system's shutdown hook, but that is a little different that what we are doing during shutdown. Shutdown hooks are system wide and have multiple entry points. But since these are completely controlled by us, don't see a strong reason for parallelism here. I would just execute it in a single thread.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
```diff
@@ -136,30 +139,36 @@ public void close() {
...
-                    }
-                }, shutdownService))
-                .toArray(CompletableFuture[]::new);
+        List<CompletableFuture<Void>> shutdownFutures = serverMap.values()
+                .stream()
+                .map(server -> shutdownServer(shutdownService, server))
+                .collect(Collectors.toList());
 
-        CompletableFuture.allOf(shutdownFutures).join();
+        CFUtils.allOf(shutdownFutures).join();
```
Maithem, line 154: I would suggest we time out and log a warning if its not done within a some period of time. Indefinitely waiting during shutdown seems suspicious.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
```diff
@@ -23,23 +23,22 @@
...
+    private final FileChannel readChannel;
 
     @NonNull
-    String fileName;
+    private final Path fileName;
 
     private final Map<Long, AddressMetaData> knownAddresses = new ConcurrentHashMap<>();
     private final Set<Long> trimmedAddresses = Collections.newSetFromMap(new ConcurrentHashMap<>());
     private final Set<Long> pendingTrims = Collections.newSetFromMap(new ConcurrentHashMap<>());
     private volatile int refCount = 0;
```
Maithem, line 40: You can remove trimmedAddresses and pendingTrims they were used for sparse trimming. But that feature has been removed.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
```diff
@@ -52,18 +51,14 @@ public synchronized void release() {
...
-                Arrays.asList(writeChannel, readChannel)
-        );
-
-        for (FileChannel channel : channels) {
+        Arrays.asList(writeChannel, readChannel).forEach(channel -> {
             try {
                 channel.force(true);
             } catch (IOException e) {
-                log.debug("Can't force updates in the channel", e.getMessage());
+                log.debug("Can't force updates in the channel: {}", e.getMessage());
```
Maithem, line 58: Probably should be a warn and not a debug message. Can you also log the segment number.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * The class manages corfu's file segments
+ */
+@Builder
+@Slf4j
+public class SegmentSupervisor {
+    @Getter
+    private final ConcurrentMap<Path, SegmentHandle> channels = new ConcurrentHashMap<>();
+    private final Set<FileChannel> channelsToSync = ConcurrentHashMap.newKeySet();
+
```
Maithem, line 36: If you can keep a dirty flag inside the segment and on write flag it as true, then you can remove channelsToSync. Scanning the channels is supper cheap considering the open segments number is relatively small. Essentially the scan is cheap.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -723,79 +713,6 @@ private LogData readRecord(SegmentHandle segment, long address) throws IOExcepti
         }
     }
 
-    @Nullable
```
Maithem, line 726: In `private void readAddressSpace(SegmentHandle segment) throws` can you load the address space from the read channel and not the write channel ?

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
```diff
@@ -0,0 +1,181 @@
...
+            } catch (IOException e) {
+                log.error("Error opening file {}", missingFilePath, e);
+                IOUtils.closeQuietly(writeCh);
+                IOUtils.closeQuietly(readCh);
+                throw new IllegalStateException(e);
+            } catch (RuntimeException ex) {
+                //Prevents file resources leaks in case of any RuntimeException.
+                IOUtils.closeQuietly(writeCh);
+                IOUtils.closeQuietly(readCh);
+                throw ex;
```
Maithem, line 76: If this is not expected, then its a good idea to log an error.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
```diff
@@ -0,0 +1,181 @@
...
+                    segment.getSegment(), segment.getRefCount()
+            );
+        }
+
+        channelsToSync.remove(segment.getWriteChannel());
+        channels.remove(segment.getFileName(), segment);
+        segment.close();
+    }
+
+    void sync(boolean force) throws IOException {
```
Maithem, line 124: Can you try to force all the segments instead of short-circuiting and exiting on the first failure?

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
```diff
@@ -0,0 +1,181 @@
...
+        if (force) {
+            for (FileChannel ch : channelsToSync) {
+                ch.force(true);
+                channelsToSync.remove(ch);
+            }
+        }
+        log.trace("Sync'd {} channels", channelsToSync.size());
```
Maithem, line 131: This is called often, maybe wrap it with `log.isTrace()`

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -218,12 +211,11 @@ private void initializeLogMetadata() {
                     logMetadata.update(logEntry, true);
                 }
             } finally {
-                segment.close();
+                segmentSupervisor.close(segment);
```
Maithem, line 214: You can make this a `try -with-resources`

Maithem, line 214: You can make this a `try-with-resources`

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -980,11 +896,10 @@ private AddressMetaData writeRecord(SegmentHandle segment, long address,
         ByteBuffer record = getByteBuffer(metadata, logEntry);
         long channelOffset;
 
-        try (MultiReadWriteLock.AutoCloseableLock ignored =
-                     segmentLocks.acquireWriteLock(segment.getSegment())) {
+        try (AutoCloseableLock ignored = segmentLocks.acquireWriteLock(segment.getSegment())) {
```
Maithem, line 899: You can remove `segmentLocks` it's not used, it will break one test in StreamLogFiles, but that use case isn't supported. We never have multi-writers

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1285,39 +1174,24 @@ private void deleteFilesMatchingFilter(FileFilter fileFilter) {
     @Override
     public void reset() {
         // Trim all segments
-        long endSegment = Math.max(logMetadata.getGlobalTail(), 0L) / RECORDS_PER_LOG_FILE;
+        long endSegment = getEndSegment();
```
Maithem, line 1177: This is not robust. I know you didn't write it, but there is a problem.
If for some reason a node experiences data corruption, it is taken out of the cluster and then re-introduced. On the reset, we should do a "blind" delete and delete everything in /log , but this code depends on retrieving the end segment. what if that metadata is corrupted, then you won't be able to delete the state. 

A cleaner approach would be to just nuke everything in that directory.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1285,39 +1174,24 @@ private void deleteFilesMatchingFilter(FileFilter fileFilter) {
...
-        deleteFilesMatchingFilter(file -> {
-            try {
-                String segmentStr = file.getName().split("\\.")[0];
-                return Long.parseLong(segmentStr) <= endSegment;
-            } catch (Exception e) {
-                log.warn("reset: ignoring file {}", file.getName());
-                return false;
-            }
-        });
+        deleteFiles(endSegment, "reset: ignoring file {}");
```
Maithem, line 1183: Can't you just push this to `SegmentSupervisor` and have a `deleteAllSegments?`

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -419,19 +405,23 @@ private void trimPrefix() {
...
         // Close segments before deleting their corresponding log files
-        closeSegmentHandlers(endSegment);
+        segmentSupervisor.closeSegmentHandlers(endSegment);
 
+        deleteFiles(endSegment, "trimPrefix: ignoring file {}");
+
+        log.info("trimPrefix: completed, end segment {}", endSegment);
+    }
+
+    private void deleteFiles(long endSegment, String errorMsg) {
```
Maithem, line 415: Can make this `final long endSegment`

xnull: there is no reason for that, `final` word is a boilerplate here

test/src/main/java/org/corfudb/test/LsofSpec.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.fail;
+
+@Slf4j
+public class LsofSpec {
```
Maithem, line 15: I think a more general approach for this spec should filter on `'(deleted)'` and not the segment path. That way it will find all deleted files with an open file descriptor.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
```diff
@@ -1,121 +1,160 @@
...
+        segmentSupervisor.close(segment);
+        log.contains(0);
+        // a resource leak could happen in case of clearing channels
+        // after the fact that `contains` method just added a segment into the Map.
+        // To prevent resources leak don't clean the map (SegmentSupervisor.channels).
+        //segmentSupervisor.getChannels().clear();
+        log.close();
+
+        segmentSupervisor.closeSegmentHandlers(0);
+        lsofSpec.check();
```
Maithem, line 105: Can't you execute this after every test with a junit annotation ?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
xnull: Agree, but I wouldn't like to touch it now

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
xnull: Not in this bug fix

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
xnull: let's create another PR for that

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentHandle.java
xnull: It was `debug` all the time if you'd like to change it, we would need another PR

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
xnull: I would like to bring any changes in logic in this PR

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -456,6 +456,7 @@ public void shutdown() {
         executor.shutdown();
```
Maithem, line 456: This doesn't wait for tasks to finish executing which can introduce a race that can cause a file descriptor leak.

When `clearDataFiles`  interleaves with any call to `getSegmentHandleForAddress` (i.e. through read/inspectAddresses).

Consider the following execution. A segment is closed, a read comes in and re-opens the segment, then the segment files are deleted. 

Can this not happen?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: the channel is used for reads and writes, I wouldn't like to touch it in this pr

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
xnull: If we throw an exception then no need for logs. In the other case, we would have a lot of logs on different levels but the logs will be useless. Since the exception will be handled on upper levels, we expect that any necessary logging will happen on those levels

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
Maithem: That's the point, the read channel opens the file in read only mode, so writes will fail. During the load, only reads should be permitted. Its more restrictive.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
xnull: I'm not sure what could be the strategy here. If `sync` throws an exception we stop BatchProcessor.
This is how it works right now.
I'm not sure what we expect if sync fails

Maithem: If segment 1, 2 and 3 need to be synced and syncing 3 fails, that should prevent us from syncing 1 and 2. The idea is to sync as much data regardless if we will fail later on, or not. We should definitely fail, but I don't think we should fail right away.

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
xnull: I think no need for that, `channelsToSync` contains a few channels, `size` always fast

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: I can't because segmentSupervisor contains many segments, I have close a particular one here. 
Also, I can't just execute segment.close() I have to remove the segment from the list

xnull: I can't because segmentSupervisor contains many segments, I have close a particular one here.
Also, I can't just execute segment.close() I have to remove the segment from the list

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: wouldn't like to touch it in this PR, since it changes the logic a bit

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: Yes, it's much simpler to delete everything, actually, I'm pretty sure in 100% cases we delete all the data anyway because `andSegment` should be always equal to the latest segment. 
Let me change this piece of code when we fix all the tests

Maithem: Everything in `*/corfu/log/*` and not everything in `/corfu/*` , still need to keep some of the local datastores data (i.e. managment/layoutserver layouts)

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: Unfortunately, I can't because It uses logSizeQuota. I wouldn't like to add logsizeQuota as a dependency into SegmentSupervisor
At least not right now

test/src/main/java/org/corfudb/test/LsofSpec.java
xnull: The problem that `lsof` output is different on Mac and Linux. Mac version doesn't have '(deleted)' in its output.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
xnull: I do https://github.com/CorfuDB/CorfuDB/blob/e3d65380c753a1104f064f2e353b03cb55144882/test/src/test/java/org/corfudb/AbstractCorfuTest.java#L365

That's why it took so much time to bring this change. We have a big mess in our unit/integration tests and almost never release resources

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: But it works now somehow, let's create another PR for that

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
xnull: I totally agree but didn't want to change it because of uncertainty.
Let me fix that if we ok with the change

xnull:
> Although this patch a great clean up, it introduces a new concurrency issue and doesn't address a major cause of file descriptor leaks. In particular, the interleaving of segment delete, reads and inspect address calls.
> 
> Might be missing something, but I don't see how that case is fixed. My reasons are articulated in my previous review.

I have pretty solid arguments not to agree with you, but before let me finish my fix

Maithem:
> you

Ok. What are they?

Maithem:
> > Although this patch a great clean up, it introduces a new concurrency issue and doesn't address a major cause of file descriptor leaks. In particular, the interleaving of segment delete, reads and inspect address calls.
> > Might be missing something, but I don't see how that case is fixed. My reasons are articulated in my previous review.
> 
> I have pretty solid arguments not to agree with you, but before let me finish my fix

Ok, what are your arguments?

infrastructure/src/main/java/org/corfudb/infrastructure/log/SegmentSupervisor.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: Sure, this is a problem.
Actually we shut down all our netty servers asynchronously, probably I should make all `shutdown` methods synchronous

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
xnull: The previous version also had a similar problem - we delete data and reset streamlog in parallel at the same time.

Fixed, by introducing and using `CorfuServerStateMachine` which brings total ordering between start, stop, reset_log_unit, shutdown operations

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
xnull: The previous version also had a similar problem - we delete data and close streamlog in parallel at the same time.

Fixed, by introducing and using `CorfuServerStateMachine` which brings total ordering between start, stop, reset_log_unit, shutdown operations

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: removed

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
xnull: We are still waiting on this infinitely, but all the servers now fail by default timeout. Please take a look

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: Fixed

xnull:
This PR will be split on smaller pr-s

PR #2621, opened 2020-07-13
Fix Shadow Stream Trimmed Exception (master)

annym:
## Overview

Description:

- Shadow Streams are specific to the snapshot sync cycle, therefore,
  log trims do not affect it in between different cycles.
- Bug Fix: on negotiation properly reset the baseSnapshot so log entry sync
  can resume in between restarts of LR.


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -37,14 +37,14 @@ public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, Str
...
-            metadataTable = this.corfuStore.openTable(namespace,
+            metadataTable = this.corfuStore.openTable(NAMESPACE,
                             metadataTableName,
                             LogReplicationMetadataKey.class,
                             LogReplicationMetadataVal.class,
                             null,
                             TableOptions.builder().build());
         } catch (Exception e) {
-            log.error("Caught an exception while opening the table namespace={}, name={}", namespace, metadataTableName);
+            log.error("Caught an exception while opening the table NAMESPACE={}, name={}", NAMESPACE, metadataTableName, e);
```
pankti-m, line 47: looks like the exception wont get printed?

test/src/test/java/org/corfudb/integration/LogReplicationAbstractIT.java
```diff
@@ -0,0 +1,362 @@
...
+
+        // Verify data is present in Standby Site
+        assertThat(mapAStandby.size()).isEqualTo(expectedConsecutiveWrites);
+
+        for (int i = 0; i < (expectedConsecutiveWrites); i++) {
+            assertThat(mapAStandby.containsKey(String.valueOf(i)));
+        }
+    }
+
+    public void checkpoint(CorfuRuntime rt, List<CorfuTable> toCheckpoint) {
```
pankti-m, line 282: can we rename this also to checkPointAndTrim?

test/src/test/java/org/corfudb/integration/LogReplicationAbstractIT.java
```diff
@@ -0,0 +1,362 @@
...
+        // Lastly restore the regular protobuf serializer and undo the dynamic protobuf serializer
+        // otherwise the test cannot continue beyond this point.
+        Serializers.registerSerializer(protobufSerializer);
+        return trimToken;
+    }
+
+    public void checkpointAndTrimStandby() {
+        // Standby MetadataTable to be checkpointed
+        String metadataTableName = LogReplicationMetadataManager.TABLE_PREFIX_NAME + DefaultClusterManager.DEFAULT_STANDBY_CLUSTER_NAME;
+        CorfuTable<LogReplicationMetadata.LogReplicationMetadataKey, LogReplicationMetadata.LogReplicationMetadataVal> metadataTableStandby =
```
pankti-m, line 347: shouldnt we do the same thing for streamNames table?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
annym: Yes it will, for exceptions we don't need to specify a place holder in log.error

test/src/test/java/org/corfudb/integration/LogReplicationAbstractIT.java
annym: Sure. Done.

PR #2630, opened 2020-07-14
Auto Format Code on Pre-Commit

Maithem:
## Overview
Use google style checks + pre-commit hook to auto-format and fail style check

Why should this be merged: Makes code base more compliant with google's formatting style. 

- [x] Public API has Javadoc


Maithem:
Upload formatter + pre-commit hook and integrate with travis

xnull:
Using `pre-commit hook` is not the best idea. 
Much better IDEA is to add the format checks in maven, the developer will not be able to compile his code until he has issues with the format.

The best plugin for that, that I know is spotless https://github.com/diffplug/spotless

Maithem:
> Using `pre-commit hook` is not the best idea.
> Much better IDEA is to add the format checks in maven, the developer will not be able to compile his code until he has issues with the format.
> 
> The best plugin for that, that I know is spotless https://github.com/diffplug/spotless

The idea behind a pre-commit hook is to **fail** on the git diff (i.e. new changes submitted). This patch reduces ~40k violations to 2k violations based on google style, so until we fix all of those violations we can't enable it from maven. Im assuming you are referring to : https://github.com/CorfuDB/CorfuDB/pull/2630/files#diff-600376dffeb79835ede4a0b285078036R325

Pre-commit hook will try to auto-format the patch and then run the style check (note that the auto formatting cannot address all violations), if the check fails, it will actually fail the build. 

Not sure if there are maven plug-ins that already do that (i.e. run the checks on the git diff), I just couldn't find any.

Maithem:
> > Using `pre-commit hook` is not the best idea.
> > Much better IDEA is to add the format checks in maven, the developer will not be able to compile his code until he has issues with the format.
> > The best plugin for that, that I know is spotless https://github.com/diffplug/spotless
> 
> The idea behind a pre-commit hook is to **fail** on the git diff (i.e. new changes submitted). This patch reduces ~40k violations to 2k violations based on google style, so until we fix all of those violations we can't enable it from maven. Im assuming you are referring to : https://github.com/CorfuDB/CorfuDB/pull/2630/files#diff-600376dffeb79835ede4a0b285078036R325
> 
> Pre-commit hook will try to auto-format the patch and then run the style check (note that the auto formatting cannot address all violations), if the check fails, it will actually fail the build.
> 
> Not sure if there are maven plug-ins that already do that (i.e. run the checks on the git diff), I just couldn't find any.

This can be enabled after we resolve all violations. For now, I think we should enable the checks only on files changed by the diff.

```
<configuration>
                    <configLocation>google_checks.xml</configLocation>
                    <encoding>UTF-8</encoding>
                    <consoleOutput>true</consoleOutput>
                    <failsOnError>false</failsOnError>
                    <failOnViolation>false</failOnViolation>
                    <violationSeverity>warning</violationSeverity>
                    <includeTestSourceDirectory>true</includeTestSourceDirectory>
                </configuration>

```

xnull:
> This patch reduces ~40k violations to 2k violations based on google style, so until we fix all of those violations we can't enable it from maven.

The spotless plugin should have a feature - "check only changed code", so, if it has that feature we don't need to fix all the issues in the code, we can enable it immediately

>Pre-commit hook will try to auto-format the patch and then run the style check (note that the auto formatting cannot address all violations), if the check fails, it will actually fail the build.

this is always a painful process - on Travis, you would have a lot of fun to be able to push changes to GitHub:
 - the first problem will be - to figure what branch is used right now. You know only a PR number when you run you code under Travis (and that's true for any build server - circle, GitHub actions, GitLab and so on)
 - you will need to provide github token and provide permissions to travis to push to the branch, which makes a lot of mess

xnull:
> Pre-commit hook will try to auto-format the patch and then run the style check (note that the auto formatting cannot address all violations), if the check fails, it will actually fail the build.

Also, if you check will apply all those rules without pushing to the branch, and fails, the developer will have hard time to run all the checks again, locally and then fix all the violations, then push changes and then hope that everything is ok. If not ok, then start over again. 

The process seems pretty painful
Also according to the branch name we age going to use python scripts, which makes the process much more painful

Issue #2627, opened 2020-07-13
Add Log Replication Cluster Role ITs


annym:
## Overview

- We need to remove current CorfuReplicationSiteConfigIT as it is thread based and this will not work on travis. 
- We need to support these same tests but based on processes.

PR #2634, opened 2020-07-14
Add Log Replication Cluster Role ITs

zhangn49:
## Overview

Description:

Add Log Replication Cluster Role ITs

* Modified Default Cluster Manager to support process-based ITs
* Removed current CorfuReplicationSiteConfigIT, since it won't work on Travis.

Why should this be merged: 

Related issue(s) (if applicable): #2627 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -489,11 +489,12 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      *   - Higher config id
      *   - Potential cluster role change
      *
+     * Cluster change from active to standby is a two step process, we first confirm that
+     * we are ready to do the cluster role change, so by the time we receive cluster change
+     * notification, nothing needs to be done, other than stop.
      * @param newTopology new discovered topology
      */
     public void onClusterRoleChange(TopologyDescriptor newTopology) {
-        // TODO: confirm prepare to become standby is a two-step process, otherwise,
-        //  we can't just stop on an intention to switch
         // Stop ongoing replication, stopLogReplication() checks leadership and active
         // We do not update topology until we successfully stop log replication
         if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
@@ -507,7 +508,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
 
         // Update topology config id in metadata manager
         logReplicationMetadataManager.setupTopologyConfigId(topologyDescriptor.getTopologyConfigId());
-        log.debug("Persist new topologyConfigId {}, cluster id={}, status={}", topologyDescriptor.getTopologyConfigId(),
+        log.debug("Persist new topologyConfigId {}, cluster id={}, role={}", topologyDescriptor.getTopologyConfigId(),
                 localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
 
         // Update replication manager
@@ -591,7 +592,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
         updateReplicationManagerTopology(discoveredTopology);
         // Update topology config id in metadata manager
         logReplicationMetadataManager.setupTopologyConfigId(topologyDescriptor.getTopologyConfigId());
-        log.debug("Persist new topologyConfigId {}, cluster id={}, status={}", topologyDescriptor.getTopologyConfigId(),
+        log.debug("Persist new topologyConfigId {}, cluster id={}, role={}", topologyDescriptor.getTopologyConfigId(),
                 localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -510,6 +510,9 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
         log.debug("Persist new topologyConfigId {}, cluster id={}, status={}", topologyDescriptor.getTopologyConfigId(),
                 localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
 
+        // Update replication manager
+        updateReplicationManagerTopology(newTopology);
+
         // Update sink manager
         interClusterReplicationService.getLogReplicationServer().getSinkManager()
                 .updateTopologyConfigId(topologyDescriptor.getTopologyConfigId());
@@ -578,15 +581,18 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
         log.debug("Standby Cluster has been added or removed from topology={}", discoveredTopology);
 
         // We only need to process new standby's if your role is of an ACTIVE cluster
-        if (localClusterDescriptor.getRole() == ClusterRole.STANDBY) {
-            return;
-        }
-
-        if (replicationManager != null && isLeader.get()) {
-            replicationManager.processStandbyChange(discoveredTopology);
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
+            if (replicationManager != null && isLeader.get()) {
+                replicationManager.processStandbyChange(discoveredTopology);
+            }
         }
 
         updateLocalTopology(discoveredTopology);
+        updateReplicationManagerTopology(discoveredTopology);
+        // Update topology config id in metadata manager
+        logReplicationMetadataManager.setupTopologyConfigId(topologyDescriptor.getTopologyConfigId());
+        log.debug("Persist new topologyConfigId {}, cluster id={}, status={}", topologyDescriptor.getTopologyConfigId(),
+                localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
     }
 
     /**
@@ -625,6 +631,12 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
 
         // Update local node descriptor
         localNodeDescriptor = localClusterDescriptor.getNode(localEndpoint);
+    }
+
+    private void updateReplicationManagerTopology(TopologyDescriptor newConfig) {
+        if (replicationManager != null) {
+            replicationManager.updateRuntimeConfigId(newConfig);
+        }
     }
 
     /***
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
@@ -1,5 +1,6 @@
 package org.corfudb.infrastructure.logreplication.infrastructure;
 
+import com.google.common.collect.Sets;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
@@ -14,6 +15,7 @@ import org.corfudb.util.retry.IntervalRetry;
 import org.corfudb.util.retry.RetryNeededException;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
@@ -178,34 +180,40 @@ public class CorfuReplicationManager {
     }
 
     /**
+     * Update Log Replication Runtime config id.
+     */
+    public void updateRuntimeConfigId(TopologyDescriptor newConfig) {
+        runtimeToRemoteCluster.values().forEach(runtime -> runtime.updateFSMConfigId(newConfig));
+    }
+
+    /**
      * The notification of change of adding/removing standby's without epoch change.
      *
      * @param newConfig has the same topologyConfigId as the current config
      */
     public void processStandbyChange(TopologyDescriptor newConfig) {
         if (newConfig.getTopologyConfigId() != topology.getTopologyConfigId()) {
-            log.error("Detected changes in the topology. The new topology descriptor {} doesn't have the same " +
+            log.warn("Detected changes in the topology. The new topology descriptor {} doesn't have the same " +
                     "topologyConfigId as the current one {}", newConfig, topology);
-            return;
         }
 
-        Map<String, ClusterDescriptor> newStandbys = newConfig.getStandbyClusters();
-        Map<String, ClusterDescriptor> currentStandbys = topology.getStandbyClusters();
-        newStandbys.keySet().retainAll(currentStandbys.keySet());
-        Set<String> standbysToRemove = currentStandbys.keySet();
-        standbysToRemove.removeAll(newStandbys.keySet());
+        Set<String> currentStandbys = new HashSet<>(topology.getStandbyClusters().keySet());
+        Set<String> newStandbys = new HashSet<>(newConfig.getStandbyClusters().keySet());
+        Set<String> intersection = Sets.intersection(currentStandbys, newStandbys);
 
-        /*
-         * Remove standbys that are not in the new config
-         */
+        //standbysToRemove = currentStandbys - intersection
+        Set<String> standbysToRemove = new HashSet<>(currentStandbys);
+        standbysToRemove.removeAll(intersection);
+
+        //Remove standbys that are not in the new config
         for (String clusterId : standbysToRemove) {
             stopLogReplicationRuntime(clusterId);
             topology.removeStandbyCluster(clusterId);
         }
 
         //Start the standbys that are in the new config but not in the current config
-        for (String clusterId : newConfig.getStandbyClusters().keySet()) {
-            if (runtimeToRemoteCluster.get(clusterId) == null) {
+        for (String clusterId : newStandbys) {
+            if (!runtimeToRemoteCluster.containsKey(clusterId)) {
                 ClusterDescriptor clusterInfo = newConfig.getStandbyClusters().get(clusterId);
                 topology.addStandbyCluster(clusterInfo);
                 startLogReplicationRuntime(clusterInfo);
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
@@ -189,9 +189,10 @@ public class CorfuReplicationManager {
     /**
      * The notification of change of adding/removing standby's without epoch change.
      *
-     * @param newConfig has the same topologyConfigId as the current config
+     * @param newConfig should have the same topologyConfigId as the current config
      */
     public void processStandbyChange(TopologyDescriptor newConfig) {
+        // ConfigId mismatch could happen if customized cluster manager does not follow protocol
         if (newConfig.getTopologyConfigId() != topology.getTopologyConfigId()) {
             log.warn("Detected changes in the topology. The new topology descriptor {} doesn't have the same " +
                     "topologyConfigId as the current one {}", newConfig, topology);
@@ -201,7 +202,6 @@ public class CorfuReplicationManager {
         Set<String> newStandbys = new HashSet<>(newConfig.getStandbyClusters().keySet());
         Set<String> intersection = Sets.intersection(currentStandbys, newStandbys);
 
-        //standbysToRemove = currentStandbys - intersection
         Set<String> standbysToRemove = new HashSet<>(currentStandbys);
         standbysToRemove.removeAll(intersection);
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
@@ -41,9 +41,8 @@ import java.util.concurrent.LinkedBlockingQueue;
  * one active cluster, and one or more standby clusters.
  */
 @Slf4j
-@SuppressWarnings("checkstyle:magicnumber")
 public class DefaultClusterManager extends CorfuReplicationClusterManagerBaseAdapter {
-    public static final String CONFIG_FILE_PATH = "/config/corfu/corfu_replication_config.properties";
+    public static final String CONFIG_FILE_PATH = "src/test/resources/corfu_replication_config.properties";
     private static final String DEFAULT_ACTIVE_CLUSTER_NAME = "primary_site";
     private static final String DEFAULT_STANDBY_CLUSTER_NAME = "standby_site";
     private static final int NUM_NODES_PER_CLUSTER = 3;
@@ -68,7 +67,7 @@ public class DefaultClusterManager extends CorfuReplicationClusterManagerBaseAda
     public static final CommonTypes.Uuid OP_INVALID = CommonTypes.Uuid.newBuilder().setLsb(4L).setMsb(4L).build();
 
     @Getter
-    private long epoch;
+    private long configId;
 
     @Getter
     private boolean shutdown;
@@ -83,7 +82,7 @@ public class DefaultClusterManager extends CorfuReplicationClusterManagerBaseAda
     private CorfuStore corfuStore;
 
     public void start() {
-        epoch = 0L;
+        configId = 0L;
         shutdown = false;
         topologyConfig = constructTopologyConfigMsg();
         clusterManagerCallback = new ClusterManagerCallback(this);
@@ -101,7 +100,7 @@ public class DefaultClusterManager extends CorfuReplicationClusterManagerBaseAda
             );
             table.clear();
         } catch (Exception e) {
-            // Ignore
+            throw new RuntimeException(e);
         }
         corfuStore.subscribe(new ConfigStreamListener(this), CONFIG_NAMESPACE,
                 Collections.singletonList(new TableSchema(CONFIG_TABLE_NAME, CommonTypes.Uuid.class, CommonTypes.Uuid.class, CommonTypes.Uuid.class)), ts);
@@ -239,10 +238,15 @@ public class DefaultClusterManager extends CorfuReplicationClusterManagerBaseAda
         List<ClusterDescriptor> newStandbyClusters = new ArrayList<>();
         currentConfig.getActiveClusters().values().forEach(activeCluster ->
                 newStandbyClusters.add(new ClusterDescriptor(activeCluster, ClusterRole.STANDBY)));
-        currentConfig.getStandbyClusters().values().forEach(standbyCluster ->
-                newActiveClusters.add(new ClusterDescriptor(standbyCluster, ClusterRole.ACTIVE)));
+        for (ClusterDescriptor standbyCluster : currentConfig.getStandbyClusters().values()) {
+            if (newActiveClusters.isEmpty()) {
+                newActiveClusters.add(new ClusterDescriptor(standbyCluster, ClusterRole.ACTIVE));
+            } else {
+                newStandbyClusters.add(new ClusterDescriptor(standbyCluster, ClusterRole.STANDBY));
+            }
+        }
 
-        return new TopologyDescriptor(++epoch, newActiveClusters, newStandbyClusters);
+        return new TopologyDescriptor(++configId, newActiveClusters, newStandbyClusters);
     }
 
     /**
@@ -258,7 +262,7 @@ public class DefaultClusterManager extends CorfuReplicationClusterManagerBaseAda
                 newActiveClusters.add(new ClusterDescriptor(standbyCluster, ClusterRole.ACTIVE)));
         newActiveClusters.add(currentActive);
 
-        return new TopologyDescriptor(++epoch, newActiveClusters, new ArrayList<>());
+        return new TopologyDescriptor(++configId, newActiveClusters, new ArrayList<>());
     }
 
     /**
@@ -273,7 +277,7 @@ public class DefaultClusterManager extends CorfuReplicationClusterManagerBaseAda
         ClusterDescriptor newStandby = new ClusterDescriptor(currentActive, ClusterRole.STANDBY);
         newStandbyClusters.add(newStandby);
 
-        return new TopologyDescriptor(++epoch, new ArrayList<>(), newStandbyClusters);
+        return new TopologyDescriptor(++configId, new ArrayList<>(), newStandbyClusters);
     }
 
     /**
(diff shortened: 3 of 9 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -578,15 +581,18 @@ private void onStandbyClusterAddRemove(TopologyDescriptor discoveredTopology) {
...
+                replicationManager.processStandbyChange(discoveredTopology);
+            }
         }
 
         updateLocalTopology(discoveredTopology);
+        updateReplicationManagerTopology(discoveredTopology);
+        // Update topology config id in metadata manager
+        logReplicationMetadataManager.setupTopologyConfigId(topologyDescriptor.getTopologyConfigId());
+        log.debug("Persist new topologyConfigId {}, cluster id={}, status={}", topologyDescriptor.getTopologyConfigId(),
+                localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
```
zhangn49, line 595: Even if current node's role did not change, it does not guarantee that others' role or config id is not changed...  
So updating sender/receiver's config id is necessary here.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -177,35 +179,41 @@ private void stopLogReplicationRuntime(String remoteClusterId) {
...
+
     /**
      * The notification of change of adding/removing standby's without epoch change.
      *
      * @param newConfig has the same topologyConfigId as the current config
      */
     public void processStandbyChange(TopologyDescriptor newConfig) {
         if (newConfig.getTopologyConfigId() != topology.getTopologyConfigId()) {
-            log.error("Detected changes in the topology. The new topology descriptor {} doesn't have the same " +
+            log.warn("Detected changes in the topology. The new topology descriptor {} doesn't have the same " +
```
zhangn49, line 196: We do not assume that config id change implies a role change, so it is possible that we have a higher config id.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -93,104 +91,520 @@ public void testLogReplicationRoleFlip() throws Exception {
...
+     * This test verify config change with a role switch during a snapshot sync transfer phase.
+     * <p>
+     * 1. Init with corfu 9000 active and 9001 standby
+     * 2. Write 50 entries to active map
+     * 3. Start log replication: Node 9010 - active, Node 9020 - standby
+     * 4. Perform a role switch with corfu store
+     * 5. Standby will drop messages and keep size 0
+     * 6. Verify active map becomes size 0, since source size is 0
+     */
+    //@Test
```
zhangn49, line 218: Since there is no way to make sure apply completes before the writer reset, we do not support DuringTransferPhase & DuringApplyPhase now.

xiaoqin2012:
Just curious, all tests are passing or there are still issues?

zhangn49:
> Just curious, all tests are passing or there are still issues?

All 4 tests passed. 
1.testNewConfigWithSwitchRole
2.testNewConfigWithTwoActive
3.testNewConfigWithAllStandby
4.testNewConfigWithInvalidClusters

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -578,15 +581,18 @@ private void onStandbyClusterAddRemove(TopologyDescriptor discoveredTopology) {
...
-        if (localClusterDescriptor.getRole() == ClusterRole.STANDBY) {
-            return;
-        }
-
-        if (replicationManager != null && isLeader.get()) {
-            replicationManager.processStandbyChange(discoveredTopology);
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
+            if (replicationManager != null && isLeader.get()) {
+                replicationManager.processStandbyChange(discoveredTopology);
+            }
```
xiaoqin2012, line 587: If the topologyConfigId has been change, it will overrule the processStandbyChange. processStandbyChange only is called when the topologyConfigId doesn't change and the list of standbys have been changed. 
Is the logic correct that processStandbyChange called before updating the localTopology?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
xiaoqin2012: We may add more comments to the function to explain that this function only is executed if the configID doesn't change.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -93,104 +91,520 @@ public void testLogReplicationRoleFlip() throws Exception {
...
+                activeRuntime.getAddressSpaceView().getLogTail(), standbyClusterCorfuPort,
+                standbyRuntime.getAddressSpaceView().getLogTail());
+
+        activeReplicationServer = runReplicationServer(activeReplicationServerPort, nettyPluginPath);
+        standbyReplicationServer = runReplicationServer(standbyReplicationServerPort, nettyPluginPath);
+        log.info("Replication servers started, and replication is in progress...");
+
+        // Wait until data is fully replicated
+        waitForReplication(size -> size == firstBatch, mapStandby, firstBatch);
+        log.info("After full sync, both maps have size {}. Current active corfu[{}] log tail " +
```
xiaoqin2012, line 153: Can we also add a check of the log replication metadata at the standby side. It should indicate a status of snapshot full sync has finished and the timestamp for the snapshot full sync should agree with the current active log tail.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
zhangn49: Since we don't assume that config id change implies a role change anymore, processStandbyChange won't be overruled. Now we only detect if local node role changes or not, then perform `onClusterRoleChange`  or `onStandbyClusterAddRemove`



> Is the logic correct that processStandbyChange called before updating the localTopology?

I think the logic is correct, since stopping old replication relies on old config.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: Yeah, it makes sense to check metadata. I need to add helper functions to do some Corfu table queries...
I will make sure this PR's logic is correct, then I will have a look at metadata tables and add extra checks.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -41,28 +57,72 @@
 
     private static final String ACTIVE_CLUSTER_NODE = "primary_site_node";
     private static final String STANDBY_CLUSTER_NODE = "standby_site_node";
-    private boolean ifShutdown = false;
+
+
+    public static final String CONFIG_NAMESPACE = "ns_lr_config_it";
+    public static final String CONFIG_TABLE_NAME = "lr_config_it";
+    public static final CommonTypes.Uuid OP_RESUME = CommonTypes.Uuid.newBuilder().setLsb(0L).setMsb(0L).build();
+    public static final CommonTypes.Uuid OP_SWITCH = CommonTypes.Uuid.newBuilder().setLsb(1L).setMsb(1L).build();
```
annym, line 65: What are these OP? can we have a general description?

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -0,0 +1,619 @@
...
+     * 3. Start log replication: Node 9010 - active, Node 9020 - standby
+     * 4. Wait for Snapshot Sync, both maps have size 10
+     * 5. Write 5 more entries to active map, to verify Log Entry Sync
+     * 6. Perform a active-invalid config update with corfu store
+     * 7. Write 5 more entries to active map
+     * 8. Verify data will not be replicated, since standby is invalid
+     * 9. Resume to standby and verify data is fully replicated again.
+     */
+    @Test
+    public void testNewConfigWithInvalidClusters() throws Exception {
```
annym, line 510: If you take a look at PR #2640 I broke all these functions of writing to active/standby maps/ stopping into generic utilities available in LogReplicationAbstractIT, which would reduce a lot of the code here (just a suggestion)

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -0,0 +1,619 @@
...
+
+        shutdownCorfuServer(activeCorfuServer);
+        shutdownCorfuServer(standbyCorfuServer);
+        shutdownCorfuServer(activeReplicationServer);
+        shutdownCorfuServer(standbyReplicationServer);
+    }
+
+    /**
+     * This test verify config change with a role switch.
+     * <p>
```
annym, line 121: These tests are good! I think one that we might be missing and could be interesting is:
- Snapshot and Log Entry (check it completes)
- Stop both LRs
- Enforce Role Switch
- Bring them back

This way we can verify it performs correctly the site flip even when coming back up after with a new config.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -510,6 +510,9 @@ public void onClusterRoleChange(TopologyDescriptor newTopology) {
         log.debug("Persist new topologyConfigId {}, cluster id={}, status={}", topologyDescriptor.getTopologyConfigId(),
                 localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
 
+        // Update replication manager
+        updateReplicationManagerTopology(newTopology);
```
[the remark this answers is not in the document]
zhangn49: Let's assume that we start with topology[0]: {active: A, standby:B}, then update with topology[1]:{standby: A, B}. If we do not update replication manager topology, cluster B will have outdated epoch 0.

Next time a new active C comes in, topology is [2]{active C, standby:A, B}. For cluster B, its role never changed, and it won't update replication manager topology.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -578,15 +581,18 @@ private void onStandbyClusterAddRemove(TopologyDescriptor discoveredTopology) {
...
-        if (replicationManager != null && isLeader.get()) {
-            replicationManager.processStandbyChange(discoveredTopology);
+        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
+            if (replicationManager != null && isLeader.get()) {
+                replicationManager.processStandbyChange(discoveredTopology);
+            }
         }
 
         updateLocalTopology(discoveredTopology);
+        updateReplicationManagerTopology(discoveredTopology);
```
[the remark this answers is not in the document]
zhangn49: Just like my example above, we need to keep replication manager's topology epoch updated.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -41,28 +57,72 @@
...
+    public static final CommonTypes.Uuid OP_TWO_ACTIVE = CommonTypes.Uuid.newBuilder().setLsb(2L).setMsb(2L).build();
+    public static final CommonTypes.Uuid OP_ALL_STANDBY = CommonTypes.Uuid.newBuilder().setLsb(3L).setMsb(3L).build();
+    public static final CommonTypes.Uuid OP_INVALID = CommonTypes.Uuid.newBuilder().setLsb(4L).setMsb(4L).build();
+
+    @Getter
+    private long epoch;
```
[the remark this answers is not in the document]
zhangn49: It is configId. Renamed.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -41,28 +57,72 @@
...
+        CorfuStoreMetadata.Timestamp ts = corfuStore.getTimestamp();
+        try {
+            Table<Messages.Uuid, Messages.Uuid, Messages.Uuid> table = corfuStore.openTable(
+                    CONFIG_NAMESPACE, CONFIG_TABLE_NAME,
+                    Messages.Uuid.class, Messages.Uuid.class, Messages.Uuid.class,
+                    TableOptions.builder().build()
+            );
+            table.clear();
+        } catch (Exception e) {
+            // Ignore
```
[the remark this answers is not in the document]
zhangn49: Yes, it should be thrown and discovery service will timeout and fail the test.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -0,0 +1,619 @@
...
+        // Wait until data is fully replicated again
+        waitForReplication(size -> size == thirdBatch, mapActive, thirdBatch);
+        log.info("Data is fully replicated again after role switch, both maps have size {}. " +
+                        "Current active corfu[{}] log tail is {}, standby corfu[{}] log tail is {}",
+                thirdBatch, activeClusterCorfuPort, activeRuntime.getAddressSpaceView().getLogTail(),
+                standbyClusterCorfuPort, standbyRuntime.getAddressSpaceView().getLogTail());
+
+        // Double check after 10 seconds
+        TimeUnit.SECONDS.sleep(mediumInterval);
+        //assertThat(mapActive.size()).isEqualTo(thirdBatch);
```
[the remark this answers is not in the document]
zhangn49: Yes, it works now. Forgot to bring it back after debugging.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -0,0 +1,619 @@
...
+
+        // Wait until data is fully replicated again
+        waitForReplication(size -> size == secondBatch, mapStandby, secondBatch);
+        log.info("After delta sync, both maps have size {}. Current active corfu[{}] log tail " +
+                        "is {}, standby corfu[{}] log tail is {}", secondBatch, activeClusterCorfuPort,
+                activeRuntime.getAddressSpaceView().getLogTail(), standbyClusterCorfuPort,
+                standbyRuntime.getAddressSpaceView().getLogTail());
+
+        // Verify data
+        for (int i = 0; i < secondBatch; i++) {
```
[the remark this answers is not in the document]
zhangn49: Yeah, it is cleaner. I think for loop can make sure all entries exist and help understand how many entries should be there.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -181,57 +229,151 @@ public TopologyConfigurationMsg queryTopologyConfig(boolean useCached) {
...
 
-        // Convert the current active to standby
-        ClusterDescriptor oldActive = topologyDescriptor.getActiveClusters().values().iterator().next();
-        ClusterDescriptor newStandby = new ClusterDescriptor(oldActive, ClusterRole.STANDBY);
+        List<ClusterDescriptor> newActiveClusters = new ArrayList<>();
+        List<ClusterDescriptor> newStandbyClusters = new ArrayList<>();
+        currentConfig.getActiveClusters().values().forEach(activeCluster ->
+                newStandbyClusters.add(new ClusterDescriptor(activeCluster, ClusterRole.STANDBY)));
+        currentConfig.getStandbyClusters().values().forEach(standbyCluster ->
+                newActiveClusters.add(new ClusterDescriptor(standbyCluster, ClusterRole.ACTIVE)));
```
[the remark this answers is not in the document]
zhangn49: Yes, done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -181,57 +229,151 @@ public TopologyConfigurationMsg queryTopologyConfig(boolean useCached) {
...
+                    clusterManager.updateTopologyConfig(newConfig.convertToMessage());
+                    log.warn("change the cluster config");
+                } catch (InterruptedException ie) {
+                    throw new UnrecoverableCorfuInterruptedError(ie);
                 }
             }
         }
     }
 
+    public static class ConfigStreamListener implements StreamListener {
```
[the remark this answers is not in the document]
zhangn49: Done.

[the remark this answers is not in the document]
zhangn49: Yes, it is not required.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
zhangn49: It is operations. I think it is kind of ugly to hardcode it and I will try to directly write new config to tables.

zhangn49:
@annym @xiaoqin2012 
Hi, I have addressed most of your comments. And I think it is good to have this PR merged before the cut, so I can test role change cases in the production environment.

I will have a separate PR to address 3 things later:
1. Code cleanup: Remove ClusterManagerCallback, change config operations, and reuse functions in #2640 
2. More test cases: 'Stop LR, Role switch, and Bring back', as well as two cases I commented out.
3. Add helper functions to query and check standby metadata.

zhangn49:
Pending items will be tracked by #2656

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -112,9 +171,8 @@ public static TopologyDescriptor readConfig() throws IOException {
             }
             // TODO: add reading of node id (which is the APH node uuid)
 
-            reader.close();
-        } catch (FileNotFoundException e) {
-            log.warn("Plugin Config File {} does not exist. Using default configs", config_file);
+        } catch (IOException e) {
+            log.warn("Plugin Config File {} does not exist. Using default configs", CONFIG_FILE_PATH);
```
pankti-m, line 175: this can be an inaccurate error message if the exception was not FileNotFound.  Why was the exception type changed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -181,57 +229,161 @@ public TopologyConfigurationMsg queryTopologyConfig(boolean useCached) {
     }
 
     /**
-     * Enforce one of the standby Cluster's to become the new active cluster and current active to become standby
+     * Create a new topology config, which changes one of the standby as the active,
+     * and active as standby. Data should flow in the reverse direction.
```
pankti-m, line 233: can you annotate it with @TestOnly or add it in the comment that this is only used by tests?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -181,57 +229,161 @@ public TopologyConfigurationMsg queryTopologyConfig(boolean useCached) {
...
-        TopologyDescriptor newSiteConf = new TopologyDescriptor(1L, Arrays.asList(newPrimary), standbyClusters);
-        return newSiteConf;
+        return new TopologyDescriptor(++configId, newActiveClusters, newStandbyClusters);
+    }
+
+    /**
+     * Create a new topology config, which marks all standby cluster as active on purpose.
+     * System should drop messages between any two active clusters.
+     **/
+    public TopologyDescriptor generateConfigWithAllActive() {
```
pankti-m, line 257: can you add an annotation or comment that this in only for testing?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -181,57 +229,161 @@ public TopologyConfigurationMsg queryTopologyConfig(boolean useCached) {
...
+        newStandbyClusters.add(newStandby);
+
+        return new TopologyDescriptor(++configId, new ArrayList<>(), newStandbyClusters);
+    }
+
+    /**
+     * Create a new topology config, which marks all standby cluster as invalid on purpose.
+     * System should not send messages in this case.
+     **/
+    public TopologyDescriptor generateConfigWithInvalid() {
```
pankti-m, line 288: maybe we should also add a case with active also as invalid

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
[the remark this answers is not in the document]
pankti-m: why do we need to use the table and a listener on it?  The test can just iterate thru all cases and inject different topologies right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -4,6 +4,7 @@
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.infrastructure.logreplication.LogReplicationConfig;
+import org.corfudb.infrastructure.logreplication.proto.LogReplicationMetadata.LogReplicationMetadataVal;
```
pankti-m, line 7: is this import needed?

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -0,0 +1,618 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+/**
+ * This test suite exercises some topology config change scenarios.
+ * Each test will start with two single node corfu servers, and two single node log replicators.
+ */
+@Slf4j
+@SuppressWarnings("checkstyle:magicnumber")
```
pankti-m, line 30: do you plan to remove this?

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    /**
+     * This test verifies config change with a role switch during a snapshot sync transfer phase.
+     * <p>
+     * 1. Init with corfu 9000 active and 9001 standby
+     * 2. Write 50 entries to active map
+     * 3. Start log replication: Node 9010 - active, Node 9020 - standby
+     * 4. Perform a role switch with corfu store
+     * 5. Standby will drop messages and keep size 0
```
pankti-m, line 214: Actually, role is not expected to change when in transfer phase.  Only when we report transfer as 100%, the flip happens.  Will have to decide how we want to handle this case.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -0,0 +1,618 @@
...
+    /**
+     * This test verifies config change with two standby clusters
+     * <p>
+     * 1. Init with corfu 9000 active and 9001 standby
+     * 2. Write 10 entries to active map
+     * 3. Start log replication: Node 9010 - active, Node 9020 - standby
+     * 4. Wait for Snapshot Sync, both maps have size 10
+     * 5. Write 5 more entries to active map, to verify Log Entry Sync
+     * 6. Perform a two-standby config update with corfu store
+     * 7. Write 5 more entries to active map
```
pankti-m, line 417: actually, writes do not happen on the standby.  Technically they can, but we should decide if from Corfu we want to allow them

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
pankti-m: A similar suggestion is to move the common code from all tests to a common place in this file so the tests themselves are smaller

test/src/test/resources/logback-test.xml
```diff
@@ -44,8 +44,8 @@
     </logger>
 
     <!--<root level="DEBUG">-->
-        <!--&lt;!&ndash;<appender-ref ref="FILE" />&ndash;&gt;-->
```
pankti-m, line 47: do you plan to revert this before merging?

zhangn49: I think the old one is double '<!--'. The new one is cleaner.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
zhangn49: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
zhangn49: Actually the entire DefaultClusterManager is test-only.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
zhangn49: The same. I think I will refactor this part in the next PR.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: I will keep this one, since it is an IT class.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: Yeah, these two cases are not very clear right now. We did know much about workflow when designed these two cases.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
zhangn49: Sounds good, will add it in the next PR.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
zhangn49: Auto close will throw IOException. I think it is fine, since the entire default cluster manager is test only.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -489,11 +489,12 @@ public void processLockRelease() {
      *   - Higher config id
      *   - Potential cluster role change
      *
+     * Cluster change from active to standby is a two step process, we first confirm that
+     * we are ready to do the cluster role change, so by the time we receive cluster change
```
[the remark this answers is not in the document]
zhangn49: I guess we confirm it by query replication status.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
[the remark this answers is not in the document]
zhangn49: @pankti-m  Yes, please see my example at line 515

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
zhangn49: Ideally, there is no config id change when we only processStandbyChange. I will add a short comment.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
```diff
@@ -1,33 +1,49 @@
...
 
-import static java.lang.Thread.sleep;
-
+/**
+ * This class extends CorfuReplicationClusterManagerAdapter, provides topology config API
+ * for integration tests. The initial topology config should be valid, which means it has only
+ * one active cluster, and one or more standby clusters.
+ */
 @Slf4j
+@SuppressWarnings("checkstyle:magicnumber")
```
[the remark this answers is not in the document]
zhangn49: Removed.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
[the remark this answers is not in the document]
zhangn49: @pankti-m Listener enables us not to busy-wait.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: Good to know it

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterManager.java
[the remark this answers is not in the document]
pankti-m: ... not sure if I understand, but we can discuss this offline.  It is not a functional issue.

PR #2637, opened 2020-07-16
Make snapshot message size configurable.

xiaoqin2012:
* Make the message size configurable.
   * While reading a SMR entry, setup its size.
   * While processing an opaque entry, get its size.
   * Send snapshot message with the maxMessageSize.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
```diff
@@ -135,6 +145,23 @@ void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
             b.skipBytes(len);
         }
         SMRArguments = arguments;
+        serializedSize = b.readerIndex() - readIndex;
+    }
+
+
+    public void calculateSerializedSize() {
```
Maithem, line 152: Since this is an opaque entry, then you know for fact that the arguments array is an array of byte[].

So you don't need to allocate a buffer and do the copy just to calculate the size.
```java
int size = 0;
for (byte[] buf : arguments) size +=  buf.length;
size += (SMRMethod.length() * Character.BYTES);
size += Integer.BYTES; 
return size; 
```

test/src/test/java/org/corfudb/infrastructure/logreplication/LogReplicationFSMTest.java
```diff
@@ -349,7 +350,7 @@ public void testSnapshotSyncStreamImplementation() throws Exception {
 
         Queue<LogReplicationEntry> listenerQueue = ((TestDataSender) dataSender).getEntryQueue();
 
-        assertThat(LARGE_NUM_ENTRIES/ StreamsSnapshotReader.MAX_NUM_SMR_ENTRY).isLessThanOrEqualTo(listenerQueue.size());
+        //assertThat(LARGE_NUM_ENTRIES/ StreamsSnapshotReader.MAX_NUM_SMR_ENTRY).isLessThanOrEqualTo(listenerQueue.size());
```
annym, line 353: If this is not needed lets remove.

annym, line 353: Can we have add a test where we lower the size to a reasonable amount, add 5 times that size of data and check that the listenerQueue is always lower than that and close to the limit? so we have a test that verifies the snapshotReader.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +72,115 @@
...
+
+    /**
+     * Read log data from the current stream until the sum of all SMR entries's size reaches the maxDataSize.
      * @param stream
-     * @param numEntries
      * @return
      */
-    private List<SMREntry> next(OpaqueStreamIterator stream, int numEntries) {
-        //if it is the end of the stream, set an end of stream mark, the current
+    private SMREntryList next(OpaqueStreamIterator stream) {
```
annym, line 147: Can we add a unit test on this? as it is core to snapshot sending protocol.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -216,16 +219,30 @@ public String getPluginConfigFilePath() {
...
     }
 
-    public int getSnapshotSyncBatchSize() {
-        Integer snapshotSyncBatchSize = getServerConfig(Integer.class, "--snapshot-batch");
-        return snapshotSyncBatchSize == null ? SnapshotSender.DEFAULT_SNAPSHOT_BATCH_SIZE : snapshotSyncBatchSize;
+    /**
+     * Get the max number of snapshot message can be sent over per run cycle during full snapshot sync.
+     * @return
+     */
+    public int getMaxNumSnapshotDataMsgPerCycle() {
```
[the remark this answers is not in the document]
xiaoqin2012: 64MB is the max data msg size. Also, how many messages we can send for each run cycle in the state machine.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +72,115 @@
...
+                // If the current element could not be included in the message
+                // remember it has the iterator has moved it.
+                if (currentSize + currentEntrySize > maxDataSize) {
+                    lastEntry = currentSMREntries;
+                    break;
+                }
+
+                // Add all entries and update the size
                 list.addAll(entry.getEntries().get(stream.uuid));
+                currentSize += currentEntrySize;
```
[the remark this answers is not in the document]
xiaoqin2012: It preserves the ordering.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
xiaoqin2012: You can create a PR track that.

Issue #2623, opened 2020-07-13
Shadow Streams Legit Trim Exceptions


annym:
## Overview

At the standby cluster (receiver) a replicated stream is temporarily placed in a shadow stream. This shadow stream aims to accumulate data until a full snapshot is available on the standby, to further apply on the real stream. Shadow streams are not checkpointed, therefore, we can hit TrimmedExceptions on this stream. Upon the reasons for not checkpointing this stream, we have (i) avoid multiple copies of the same data in a given period of time and (ii) actually making these stream names available to the Checkpointer is a challenge as they are not opened by the application (which is the way we know which streams to checkpoint).

We can break down the TrimmedException problem in very specific cases:
1. **Invalid Trimmed Exception**: this happens when the snapshot sync succeeds, i.e., we write to the shadow, we apply on real stream and we clear the shadow table. But later in time (after several cycles of checkpoint) we try a new snapshot sync, which attempt to re-open the previous shadow table and hits a trim as no CP is found and the address map for this stream reflect that there was data at some point.
2. **Legit Trimmed Exception**: when the shadow stream is trimmed before applying, either because:
    a. The snapshot duration is larger than the checkpoint cycle. 
    b. The snapshot duration though not larger than the checkpoint cycle, overlapped with the checkpoint/trim point. 

We currently have a temporal solution for (1). In which we use a different shadow stream per snapshot cycle, this avoiding TE completely. The downside of this approach is that we accumulate metadata (sequencer address maps over time) and it does not address (2). 

**SOLUTION**

To solve (2) we propose to have Snapshot Sync Hooks available through the discovery service, so checkpoint/trim is disabled during this time (this requires alarms to avoid the case of long/never ending snapshot syncs) 

This solution can be complemented with the following, to address (1):

Solution 1: **ignore trims**. With checkpointing/trim being blocked during snapshot, trims cannot occur during a snapshot cycle but only in between and since this is the case denoted in (1), i.e., invalid trimmed exceptions. We can just ignore and seek from the latest point of this stream, as trimmed data was part of previous snapshot cycle.

Solution 2: **Drop API**. Having an API to actually drop the address maps for the shadow stream right 
before the apply is finished. 

Solution 3: **Table Aliasing**. Being able to do an atomic switch from the shadow table to the actual stream. This would significantly improve performance and optimize resources. It implies core changes.

[...]

PR #2640, opened 2020-07-17
Handle Trim Exceptions on Shadow Streams

annym:
Overview

Description:

1. Implement a callback mechanism, where snapshot sync start/end is notified
to a plugin, so checkpoint/trim is disabled for the duration of snapshot sync.
2. Bug Fix: properly set base snapshot on renegotiation for two different log entry (delta) sync cycles
3. Tests to verify 1 and 2

Related issue(s) (if applicable): #2623

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultSnapshotSyncPlugin.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultSnapshotSyncPlugin.java
@@ -1,0 +1,23 @@
+package org.corfudb.infrastructure.logreplication.infrastructure.plugins;
+
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.CorfuRuntime;
+
+/**
+ * Default Snapshot Sync Plugin
+ *
+ * This implementation returns immediately for start and end of a snapshot sync.
+ */
+@Slf4j
+public class DefaultSnapshotSyncPlugin implements ISnapshotSyncPlugin {
+
+    @Override
+    public void onSnapshotSyncStart(CorfuRuntime runtime) {
+        log.debug("onSnapshotSyncStart :: OK");
+    }
+
+    @Override
+    public void onSnapshotSyncEnd(CorfuRuntime runtime) {
+        log.debug("onSnapshotSyncEnd :: OK");
+    }
+}
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/LogReplicationFSM.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/LogReplicationFSM.java
@@ -289,16 +289,16 @@ public class LogReplicationFSM {
             LogReplicationEvent event = eventQueue.take();
 
             if (event.getType() == LogReplicationEventType.REPLICATION_START) {
-                baseSnapshot = event.getMetadata().getSyncTimestamp();
-                ackedTimestamp = baseSnapshot;
+                baseSnapshot = event.getMetadata().getLastBaseSnapshot();
+                ackedTimestamp = event.getMetadata().getLastLogEntrySyncedTimestamp();
                 log.info("Log Replication FSM consume event {}", event);
             }
 
             if (event.getType() == LogReplicationEventType.LOG_ENTRY_SYNC_REPLICATED) {
                 if (state.getType() == LogReplicationStateType.IN_LOG_ENTRY_SYNC &&
                         state.getTransitionEventId().equals(event.getMetadata().getRequestId())) {
-                    log.debug("Log Entry Sync ACK, update last ack timestamp to {}", event.getMetadata().getSyncTimestamp());
-                    ackedTimestamp = event.getMetadata().getSyncTimestamp();
+                    log.debug("Log Entry Sync ACK, update last ack timestamp to {}", event.getMetadata().getLastLogEntrySyncedTimestamp());
+                    ackedTimestamp = event.getMetadata().getLastLogEntrySyncedTimestamp();
                     log.debug("Replicated Event ackTs {} ", ackedTimestamp);
                 }
             } else {
@@ -306,9 +306,9 @@ public class LogReplicationFSM {
                     // Verify it's for the same request, as that request could've been canceled and was received later
                     if (state.getType() == LogReplicationStateType.IN_SNAPSHOT_SYNC &&
                             state.getTransitionEventId().equals(event.getMetadata().getRequestId())) {
-                        log.info("Snapshot Sync ACK, update last ack timestamp to {}", event.getMetadata().getSyncTimestamp());
-                        baseSnapshot = event.getMetadata().getSyncTimestamp();
-                        ackedTimestamp = baseSnapshot;
+                        log.info("Snapshot Sync ACK, update last ack timestamp to {}", event.getMetadata().getLastLogEntrySyncedTimestamp());
+                        baseSnapshot = event.getMetadata().getLastBaseSnapshot();
+                        ackedTimestamp = event.getMetadata().getLastLogEntrySyncedTimestamp();
                     }
                 }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
@@ -14,6 +14,8 @@ import org.corfudb.runtime.collections.TableOptions;
 import org.corfudb.runtime.collections.TxBuilder;
 import org.corfudb.runtime.view.Address;
 
+import java.util.UUID;
+
 import static org.corfudb.runtime.view.TableRegistry.CORFU_SYSTEM_NAMESPACE;
 
 /**
@@ -23,7 +25,7 @@ import static org.corfudb.runtime.view.TableRegistry.CORFU_SYSTEM_NAMESPACE;
 @Slf4j
 public class LogReplicationMetadataManager {
 
-    private static final String namespace = CORFU_SYSTEM_NAMESPACE;
+    private static final String NAMESPACE = CORFU_SYSTEM_NAMESPACE;
     private static final String TABLE_PREFIX_NAME = "CORFU-REPLICATION-WRITER-";
 
     private CorfuStore corfuStore;
@@ -39,14 +41,14 @@ public class LogReplicationMetadataManager {
         this.corfuStore = new CorfuStore(runtime);
         metadataTableName = getPersistedWriterMetadataTableName(localClusterId);
         try {
-            metadataTable = this.corfuStore.openTable(namespace,
+            metadataTable = this.corfuStore.openTable(NAMESPACE,
                             metadataTableName,
                             LogReplicationMetadataKey.class,
                             LogReplicationMetadataVal.class,
                             null,
                             TableOptions.builder().build());
         } catch (Exception e) {
-            log.error("Caught an exception while opening the table namespace={}, name={}", namespace, metadataTableName);
+            log.error("Caught an exception while opening the table NAMESPACE={}, name={}", NAMESPACE, metadataTableName, e);
             throw new ReplicationWriterException(e);
         }
         setupTopologyConfigId(topologyConfigId);
@@ -57,16 +59,16 @@ public class LogReplicationMetadataManager {
     }
 
     public TxBuilder getTxBuilder() {
-        return corfuStore.tx(namespace);
+        return corfuStore.tx(NAMESPACE);
     }
 
     private String queryString(CorfuStoreMetadata.Timestamp timestamp, LogReplicationMetadataType key) {
         LogReplicationMetadataKey txKey = LogReplicationMetadataKey.newBuilder().setKey(key.getVal()).build();
         CorfuRecord record;
         if (timestamp == null) {
-            record = corfuStore.query(namespace).getRecord(metadataTableName, txKey);
+            record = corfuStore.query(NAMESPACE).getRecord(metadataTableName, txKey);
         } else {
-            record = corfuStore.query(namespace).getRecord(metadataTableName, timestamp, txKey);
+            record = corfuStore.query(NAMESPACE).getRecord(metadataTableName, timestamp, txKey);
         }
 
         LogReplicationMetadataVal metadataVal = null;
@@ -102,12 +104,11 @@ public class LogReplicationMetadataManager {
         return query(null, LogReplicationMetadataType.LAST_SNAPSHOT_STARTED);
     }
 
-
     public long getLastSnapTransferDoneTimestamp() {
         return query(null, LogReplicationMetadataType.LAST_SNAPSHOT_TRANSFERRED);
     }
 
-    public long getLastSrcBaseSnapshotTimestamp() {
+    public long getLastAppliedBaseSnapshotTimestamp() {
         return query(null, LogReplicationMetadataType.LAST_SNAPSHOT_APPLIED);
     }
 
@@ -140,7 +141,7 @@ public class LogReplicationMetadataManager {
             return;
         }
 
-        TxBuilder txBuilder = corfuStore.tx(namespace);
+        TxBuilder txBuilder = corfuStore.tx(NAMESPACE);
 
         for (LogReplicationMetadataType key : LogReplicationMetadataType.values()) {
             long val = Address.NON_ADDRESS;
@@ -163,7 +164,7 @@ public class LogReplicationMetadataManager {
             return;
         }
 
-        TxBuilder txBuilder = corfuStore.tx(namespace);
+        TxBuilder txBuilder = corfuStore.tx(NAMESPACE);
 
         for (LogReplicationMetadataType key : LogReplicationMetadataType.values()) {
             long val = Address.NON_ADDRESS;
@@ -211,7 +212,7 @@ public class LogReplicationMetadataManager {
             return false;
         }
 
-        TxBuilder txBuilder = corfuStore.tx(namespace);
+        TxBuilder txBuilder = corfuStore.tx(NAMESPACE);
 
         // Update the topologyConfigId to fence all other transactions that update the metadata at the same time
         appendUpdate(txBuilder, LogReplicationMetadataType.TOPOLOGY_CONFIG_ID, topologyConfigId);
@@ -255,7 +256,7 @@ public class LogReplicationMetadataManager {
             return;
         }
 
-        TxBuilder txBuilder = corfuStore.tx(namespace);
+        TxBuilder txBuilder = corfuStore.tx(NAMESPACE);
 
         //Update the topologyConfigId to fence all other transactions that update the metadata at the same time
         appendUpdate(txBuilder, LogReplicationMetadataType.TOPOLOGY_CONFIG_ID, topologyConfigId);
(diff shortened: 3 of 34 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/LogReplicationEventMetadata.java
```diff
@@ -20,12 +20,15 @@
...
-    /* Represents the last synced timestamp.
-     *
-     * For snapshot sync, it represents the base snapshot.
-     * For log entry sync, it represents the last log entry synced.
-    */
-    private long syncTimestamp;
+    /*
+     * Represents the last log entry synced timestamp.
+     */
+    private long lastLogEntrySyncedTimestamp;
```
pankti-m, line 26: does this only apply to logEntrySync?  because for snapshot sync we only ack after all entries are received...

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -556,7 +561,8 @@ public String getOptionsString() {
          * @throws IOException
          */
         public Process runServer() throws IOException {
-            final String serverConsoleLogPath = CORFU_LOG_PATH + File.separator + host + "_" + port + "_consolelog";
+            // final String serverConsoleLogPath = CORFU_LOG_PATH + File.separator + host + "_" + port + "_consolelog";
+            final String serverConsoleLogPath = "/Users/amartinezman/Desktop/gm" + File.separator + host + "_" + port + "_consolelog";
```
pankti-m, line 565: remove?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/SnapshotSinkBufferManager.java
```diff
@@ -84,14 +84,8 @@ public LogReplicationEntryMetadata makeAckMessage(LogReplicationEntry entry) {
      */
     @Override
     public boolean verifyMessageType(LogReplicationEntry entry) {
-        switch (entry.getMetadata().getMessageMetadataType()) {
-            case SNAPSHOT_MESSAGE:
-            case SNAPSHOT_END:
-                return true;
-            default:
-                log.error("wrong message type ", entry.getMetadata());
```
pankti-m, line 92: is this log message required?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -23,7 +23,7 @@
 @Slf4j
 public class LogReplicationMetadataManager {
 
-    private static final String namespace = CORFU_SYSTEM_NAMESPACE;
+    private static final String NAMESPACE = CORFU_SYSTEM_NAMESPACE;
```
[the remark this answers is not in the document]
annym: mmm yeah I thought I had done it previously too. But I checked and it is correctly rebased, I think the only change is that since it's a static it's all caps.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/LogReplicationEventMetadata.java
annym: correct, it only applies for LogEntrySync.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -229,7 +258,13 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
         if (rxState.equals(RxState.LOG_ENTRY_SYNC)) {
             return logEntrySinkBufferManager.processMsgAndBuffer(message);
         } else {
-            return snapshotSinkBufferManager.processMsgAndBuffer(message);
+            LogReplicationEntry ack = snapshotSinkBufferManager.processMsgAndBuffer(message);
+            if (message.getMetadata().getMessageMetadataType() == SNAPSHOT_END) {
+                // Notify end of snapshot sync. This is a blocking call.
+                snapshotSyncPlugin.onSnapshotSyncEnd();
```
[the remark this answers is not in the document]
annym: mmm but the snapshotPlugin is available in this class and not in the snapshotSinkBufferManager.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/SnapshotSinkBufferManager.java
annym: I added it on the caller!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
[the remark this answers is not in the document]
pankti-m: ok, yes since we call it after processMsgAndBuffer(), the apply will have completed.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -59,15 +67,12 @@
     private LogReplicationMetadataManager logReplicationMetadataManager;
     private RxState rxState;
 
-    private LogReplicationConfig config;
+    private final LogReplicationConfig config;
```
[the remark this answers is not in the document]
pankti-m: so we should not make this final?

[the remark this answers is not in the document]
pankti-m: sorry, I dint see you had already made that change

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/LogReplicationFSM.java
```diff
@@ -290,26 +289,26 @@ private void consume() {
             LogReplicationEvent event = eventQueue.take();
 
             if (event.getType() == LogReplicationEventType.REPLICATION_START) {
-                baseSnapshot = event.getMetadata().getSyncTimestamp();
-                ackedTimestamp = baseSnapshot;
+                baseSnapshot = event.getMetadata().getLastBaseSnapshot();
```
PavelZaytsev, line 292: I had questions about it in PR, looks like now its fixed, cool

annym: yes, most likely you were hitting this bug as well!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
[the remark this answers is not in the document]
xiaoqin2012: Snapshot_end : means the data has been transferred to the standby site and the standby will start to apply the entries from shadow stream to the real streams.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -19,84 +19,106 @@
...
+        // For every stream create a shadow stream which name is unique based
+        // on the original stream and a suffix.
+        for (String streamName : config.getStreamsToReplicate()) {
+            String shadowStreamName = streamName + SHADOW_STREAM_SUFFIX;
+            UUID streamId = CorfuRuntime.getStreamID(streamName);
+            UUID shadowStreamId = CorfuRuntime.getStreamID(shadowStreamName);
+            uuidMap.put(streamId, shadowStreamId);
+            uuidMap.put(shadowStreamId, streamId);
+            streamViewMap.put(streamId, streamName);
+
```
xiaoqin2012, line 91: streamViewMap should not change after initialized?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -205,37 +227,43 @@ public void apply(List<LogReplicationEntry> messages) throws Exception {
...
+    private long applyShadowStream(UUID streamId, long seqNum, long snapshot) {
+        UUID shadowStreamId = uuidMap.get(streamId);
+
+        // We can safely ignore trims on the shadow stream, as our protocol guarantees
+        // that the checkpointer/trimmer will not run during a snapshot sync. Therefore,
+        // any observed TrimmedException belongs to different snapshot cycles, for which,
+        // old data is guaranteed to be applied.
         StreamOptions options = StreamOptions.builder()
-                .ignoreTrimmed(false)
+                .ignoreTrimmed(true)
```
xiaoqin2012, line 244: Is it necessary to set ignore the trim? How about there is a bug in the plugin?
If the plugin works correct, we should not get trimmed exception. If the plugin doesn't work properly, the working shadow stream can still be trimmed and we get a trim exception, right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/LogReplicationEventMetadata.java
```diff
@@ -53,16 +56,31 @@ public LogReplicationEventMetadata(UUID requestId) {
      */
     public LogReplicationEventMetadata(UUID requestId, long syncTimestamp) {
         this.requestId = requestId;
```
xiaoqin2012, line 58: Is this used by testing? We better remove this one.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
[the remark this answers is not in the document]
annym: Is this a comment or a response to Pankti's comment? 
Because snapshotSinkBufferManager.processMsgAndBuffer applies the data. So when it comes back we're sure it has been applied.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/LogReplicationEventMetadata.java
```diff
@@ -53,16 +56,31 @@ public LogReplicationEventMetadata(UUID requestId) {
...
-        this.syncTimestamp = syncTimestamp;
+        this.lastLogEntrySyncedTimestamp = syncTimestamp;
```
xiaoqin2012, line 59: We better remove this constructor, or add setting the baseSnapshot.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
annym: If we don't ignore the trim, we'll get TrimmedExceptions on a new snapshot cycle. At the end of the first snapshot cycle we signal the plugin to resume checkpoint/trim, therefore, by the second cycle, the shadow stream has been trimmed one or several times.

xiaoqin2012: Can we still open a stream that has been trimmed but no checkpoint? If there is bug in the plug in, we may lose data.
Can we bookkeeping the log timestamp where the current snapshot start to write to the shadow stream, while start to apply, we open the the stream and see the start timestamp is not trimmed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultSnapshotSyncPlugin.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Default Snapshot Sync Plugin
+ *
+ * This implementation returns immediately for start and end of a snapshot sync.
+ */
+@Slf4j
+public class DefaultSnapshotSyncPlugin implements ISnapshotSyncPlugin {
+
+    @Override
+    public void onSnapshotSyncStart() {
```
hisundar, line 14: Can this method take a CorfuRuntime as a parameter? Without that it's tricky to re-initialize a new runtime just for freezing checkpointing.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultSnapshotSyncPlugin.java
```diff
@@ -0,0 +1,22 @@
...
+        log.debug("onSnapshotSyncStart :: OK");
+    }
+
+    @Override
+    public void onSnapshotSyncEnd() {
```
hisundar, line 19: Can this method take a CorfuRuntime as a parameter? Without that it's tricky to re-initialize a new runtime just for unfreezing checkpointing.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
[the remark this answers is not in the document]
xiaoqin2012: Should check the ack message type.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
annym: mmm, yeah I think this is something we can do. Let me try it out.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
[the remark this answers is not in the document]
annym: @xiaoqin2012 Would it be enough to change that check for:
``if (ack.getMetadata().getMessageMetadataType() == MessageType.SNAPSHOT_REPLICATED)``
as this is the only ACK sent at the end of all received snapshot sync.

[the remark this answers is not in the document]
annym: mm I'm actually thinking that yeah we probably would need to complement this with keeping the latests started snapshot sync TS in the metadata to cover the case of an ongoing snapshot which is canceled (but still completed out of order)

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
hisundar: fyi if you plan to do bookkeeping note that the snashotSync plugin already captures this info. please feel free to add a getter method to get the wall clock timestamp when the last snapshot sync requested checkpoint freezing.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
annym: done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
xiaoqin2012: I mean the log sequence number not the wall time. I am still curious if the stream can be opened if it has been trimmed without checkpointing.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/LogReplicationEventMetadata.java
annym: For several cases this event only has meaning for log entry and doesn't really handle the baseSnapshot, actually from where we call it, it has no notion of it.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -226,49 +256,89 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
             return logEntrySinkBufferManager.processMsgAndBuffer(message);
         } else {
-            return snapshotSinkBufferManager.processMsgAndBuffer(message);
+            LogReplicationEntry ack = snapshotSinkBufferManager.processMsgAndBuffer(message);
+            if (ack.getMetadata().getMessageMetadataType() == MessageType.SNAPSHOT_REPLICATED) {
+                long lastStartedSnapshotSync = logReplicationMetadataManager.getLastSnapStartTimestamp();
+                // Verify this snapshot ACK corresponds to the last initialized/valid snapshot sync
+                // as a previous one could have been canceled but still processed due to messages being out of order
+                if (ack.getMetadata().getSnapshotTimestamp() == lastStartedSnapshotSync) {
+                    // Notify end of snapshot sync. This is a blocking call.
```
xiaoqin2012, line 278: While doing snapshot transfer, all message has the same snapshottimestamp. May check the metadata's seqNum + 1 = ack's seqNum.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -226,49 +256,89 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+                    snapshotSyncPlugin.onSnapshotSyncEnd(runtime);
+                } else {
+                    log.warn("SNAPSHOT_SYNC has completed for {}, but new ongoing SNAPSHOT_SYNC is {}",
+                            ack.getMetadata().getSnapshotTimestamp(), lastStartedSnapshotSync);
```
pankti-m, line 282: ongoing snapshot sync is ack.getMetadata().getSnapshotTimestamp(), right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
xiaoqin2012: Make it more clear, we start a snapshot transfer for snapshot timestamp 10, 
the first message is snapshot_start with seqNum= -1 and snapshot timestamp 10,
the second message is snapshot_msg with seqNum=0, and snapshot timestamp 10.
the last one is snapshot_end with seqNum=1, and snapshot timestamp 10.
So all message has the same snapshot timestamp, but different SeqNum. 
May check the metadata:
if snapshot appliedDone == snapshottimestamp then it is done. make it more robust the ack's seqNum is seqNum + 1 (seqNum from metadata)

xiaoqin2012: If there are leadership change during full snapshot transfer, we need to consider the following steps:
1. older leader make a call to set the trim holder.
2. leadership. change at the standby cluster, the new leader will re-run the full cycle of snapshot sync by reset the persisted metadata. // this will block the old leader to clear the trim holder.
3. the new leader set the trim holder again. 

If we guarantee that reset the persisted metadata first, then put trim holder, may it be correct?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
annym: No, the ongoing one is the one persisted in the datastore. The ack might come from processing some old stuff in the buffer.

Issue #2624, opened 2020-07-13
Support Switchover on Non-leader LR


annym:
## Overview

We need a means to reply from a non-leader LR when the switchover workflow is triggered from this. We can approach this by periodically persisting the replication status in the metadata table and accessing this table in the queryReplicationStatus API.

xiaoqin2012:
There are two clusters: A (active) and B (standby). While B is planning to transition from standby to active role, cluster A's site manager will query LR for log replication status. 
The metadata at the cluster B records the replication status. 
1. In snapshot sync.
2. In delta sync. (only this implemented today) 

Option for implementation:
1. Any LR node at cluster A can send a query message to a node at cluster B to get the log replication status
2. Comparing the query result with its own log tail and do the proper calculation.

pankti-m:
> There are two clusters: A (active) and B (standby). While B is planning to transition from standby to active role, cluster A's site manager will query LR for log replication status.
> The metadata at the cluster B records the replication status.
> 
>     1. In snapshot sync.
> 
>     2. In delta sync. (only this implemented today)
> 
> 
> Option for implementation:
> 
>     1. Any LR node at cluster A can send a query message to a node at cluster B to get the log replication status
> 
>     2. Comparing the query result with its own log tail and do the proper calculation.

Does this mean we do not have metadata on the active to tell the snapshot sync progress?

xiaoqin2012:
The current log replication metadata is per cluster and record the replication status if it is a standby cluster.

PR #2641, opened 2020-07-18
Support queryReplicationStatus() on non-leader Active Node.

pankti-m:
## Overview
Add support on non-leader active node to respond to queries for Replication Status.
 
Description:
The leader on the active site keeps an up-to-date in-memory copy of the last ACK received.  A thread periodically reads this ACK value and updates the replication completion status in the metadata table, which is shared among cluster members.  On receiving a query for the replication status, any node(leader or not) can read the metadata table and get the completion  status.
#2624 

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/proto/log_replication_metadata.proto
+++ b/infrastructure/proto/log_replication_metadata.proto
@@ -25,3 +25,24 @@ message LogReplicationMetadataKey {
 message LogReplicationMetadataVal {
   string val = 1;
 }
+
+/*
+ * Replication Status Key
+ */
+ message ReplicationStatusKey {
+   string clusterId = 1;
+ }
+
+/*
+ * Replication Status Value
+ * Active Site sets the completionPercent, Standby sets the dataConsistent boolean
+ */
+message ReplicationStatusVal {
+  uint64 replicationCompletion = 1;
+  bool dataConsistent = 2;
+  enum SyncType {
+    SNAPSHOT = 0;
+    LOG_ENTRY = 1;
+  }
+  SyncType type = 3;
+}
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -9,6 +9,7 @@ import org.corfudb.infrastructure.logreplication.LogReplicationConfig;
 import org.corfudb.infrastructure.logreplication.infrastructure.plugins.CorfuReplicationClusterManagerAdapter;
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo;
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo.TopologyConfigurationMsg;
+import org.corfudb.infrastructure.logreplication.proto.LogReplicationMetadata;
 import org.corfudb.infrastructure.logreplication.replication.receive.LogReplicationMetadataManager;
 import org.corfudb.infrastructure.logreplication.utils.LogReplicationStreamNameTableManager;
 import org.corfudb.infrastructure.logreplication.infrastructure.DiscoveryServiceEvent.DiscoveryServiceEventType;
@@ -27,6 +28,7 @@ import org.corfudb.utils.lock.LockListener;
 
 import javax.annotation.Nonnull;
 import java.time.Duration;
+import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
@@ -62,11 +64,6 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      */
     private static final String LOCK_GROUP = "Log_Replication_Group";
     private static final String LOCK_NAME = "Log_Replication_Lock";
-
-    /**
-     * Invalid replication status, when querying a standby cluster
-     */
-    private static final int INVALID_REPLICATION_STATUS = -1;
 
     /**
      * Used by the active cluster to initiate Log Replication
@@ -423,7 +420,6 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      * Fetch current topology from Cluster Manager
      */
     private void fetchTopologyFromClusterManager() {
-
         try {
             IRetry.build(ExponentialBackoffRetry.class, () -> {
                 try {
@@ -511,6 +507,10 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
 
         // Update topology config id in metadata manager
         logReplicationMetadataManager.setupTopologyConfigId(topologyDescriptor.getTopologyConfigId());
+
+        // Reset the Replication Status on Active and Standby
+        logReplicationMetadataManager.resetReplicationStatus();
+
         log.debug("Persist new topologyConfigId {}, cluster id={}, role={}", topologyDescriptor.getTopologyConfigId(),
                 localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
 
@@ -664,14 +664,14 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     }
 
     /**
-     * Query all replicated stream log tails and remember the max
-     * and query each standbySite information according to the ackInformation decide all manay total
-     * msg needs to send out.
+     * No work needs to be done here.  If in the Active state, writes to all replicated streams have stopped at this time.
+     * Following this, the ClusterManagerAdapter can query the status of ongoing snapshot sync on the
+     * local(active) cluster.
      */
     @Override
     public void prepareToBecomeStandby() {
-        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && replicationManager != null) {
-            replicationManager.prepareClusterRoleChange();
+        if (ClusterRole.ACTIVE == localClusterDescriptor.getRole()) {
+            log.info("Received a Request to Become Standby");
         } else {
             log.warn("Illegal prepareToBecomeStandby when cluster {} with role {}",
                     localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
@@ -679,31 +679,21 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     }
 
     /**
-     * Query all replicated stream log tails and calculate the number of messages to be sent.
-     * If the max tail has changed, return 0%.
+     * Active Cluster - Read the shared metadata table to find the status of any ongoing snapshot or log entry sync
+     * and return a completion percentage.
+     * Standby Cluster - Read the shared metadata table and find if data is consistent(returns false if
+     * snapshot sync is in the apply phase)
      */
     @Override
-    public int queryReplicationStatus() {
-        //TODO make sure caller should query all nodes in the cluster and pick the max of these 3 values
-        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE) {
-            if (!isLeader.get()) {
-                log.warn("Illegal queryReplicationStatus when node is not a leader " +
-                        "in an ACTIVE cluster={} ", localClusterDescriptor.getClusterId());
-                return 0;
-            }
-
-            if (replicationManager == null) {
-                log.warn("Illegal queryReplicationStatus when replication manager is null " +
-                        "in an ACTIVE cluster={} ", localClusterDescriptor.getClusterId());
-                return 0;
-            }
-
-            return replicationManager.queryReplicationStatus();
-        } else {
-            log.warn("Illegal queryReplicationStatus when cluster={} with role {}",
-                    localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
-            return INVALID_REPLICATION_STATUS;
+    public Map<String, LogReplicationMetadata.ReplicationStatusVal> queryReplicationStatus() {
+        LogReplicationMetadata.ReplicationStatusVal status;
+        if (ClusterRole.ACTIVE == localClusterDescriptor.getRole()) {
+            return logReplicationMetadataManager.getReplicationRemainingPercent();
+        } else if (ClusterRole.STANDBY == localClusterDescriptor.getRole()) {
+            return logReplicationMetadataManager.getDataConsistentOnStandby();
         }
+        log.error("Received Replication Status Query in Incorrect Role {}.", localClusterDescriptor.getRole());
+        return null;
     }
 
     public void shutdown() {
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
@@ -21,12 +21,10 @@ import java.util.Set;
 import java.util.UUID;
 
 /**
- * This class manages Log Replication for multiple remote (standby) cluster's.
+ * This class manages Log Replication for multiple remote (standby) clusters.
  */
 @Slf4j
 public class CorfuReplicationManager {
-
-    public final static int PERCENTAGE_BASE = 100;
 
     // Keep map of remote cluster ID and the associated log replication runtime (an abstract
     // client to that cluster)
@@ -42,11 +40,6 @@ public class CorfuReplicationManager {
 
     private final CorfuRuntime corfuRuntime;
 
-    // TODO (Xiaoqin Ma): can you please add a description on this variable's meaning
-    private long prepareClusterRoleChangeLogTail;
-
-    private long totalNumEntriesToSend;
-
     private final LogReplicationMetadataManager metadataManager;
 
     private final String pluginFilePath;
@@ -61,10 +54,7 @@ public class CorfuReplicationManager {
         this.metadataManager = metadataManager;
         this.pluginFilePath = pluginFilePath;
         this.corfuRuntime = corfuRuntime;
-
         this.localNodeDescriptor = localNodeDescriptor;
-        this.prepareClusterRoleChangeLogTail = Address.NON_ADDRESS;
-        this.totalNumEntriesToSend = 0;
     }
 
     /**
@@ -85,7 +75,6 @@ public class CorfuReplicationManager {
      * Stop log replication for all the standby sites
      */
     public void stop() {
-
         runtimeToRemoteCluster.values().forEach(runtime -> {
             try {
                 log.info("Stop log replication runtime to remote cluster id={}", runtime.getRemoteClusterId());
@@ -94,7 +83,6 @@ public class CorfuReplicationManager {
                 log.warn("Failed to stop log replication runtime to remote cluster id={}", runtime.getRemoteClusterId());
             }
         });
-
         runtimeToRemoteCluster.clear();
     }
 
@@ -110,9 +98,7 @@ public class CorfuReplicationManager {
      * Start Log Replication Runtime to a specific standby Cluster
      */
     private void startLogReplicationRuntime(ClusterDescriptor remoteClusterDescriptor) {
-
         String remoteClusterId = remoteClusterDescriptor.getClusterId();
-
         try {
             if (!runtimeToRemoteCluster.containsKey(remoteClusterId)) {
                 log.info("Starting Log Replication Runtime to Standby Cluster id={}", remoteClusterId);
@@ -219,91 +205,5 @@ public class CorfuReplicationManager {
                 startLogReplicationRuntime(clusterInfo);
             }
         }
-    }
-
-    /**
-     * Query max stream tail for all streams to be replicated.
-     *
-     * @return max tail of all relevant streams.
-     */
-    private long queryStreamTail() {
-        Set<String> streamsToReplicate = context.getConfig().getStreamsToReplicate();
-        long maxTail = Address.NON_ADDRESS;
-        Map<UUID, Long> tailMap = corfuRuntime.getAddressSpaceView().getAllTails().getStreamTails();
-        for (String s : streamsToReplicate) {
-            UUID currentUUID = CorfuRuntime.getStreamID(s);
-            Long currentTail = tailMap.get(currentUUID);
-            if (currentTail != null) {
-                maxTail = Math.max(maxTail, currentTail);
-            }
-        }
-        return maxTail;
-    }
-
-    /**
-     * Given a timestamp, calculate how many entries to be sent for all replicated streams.
-     *
-     * @param timestamp
-     */
-    private long queryEntriesToSend(long timestamp) {
-        // TODO(Xiaoqin Ma / Nan) : is not exactly how many entries we need to send?
-        //  Because we only transfer some streams, and getNumEntriesToSend returns (maxStreamTail - ackedTimeStamp),
-        //  which will count other log entries. Besides, for loop will amplify totalNumEntries multiple times.
-
-        int totalNumEntries = 0;
-
-        for (CorfuLogReplicationRuntime runtime: runtimeToRemoteCluster.values()) {
-            totalNumEntries += runtime.getNumEntriesToSend(timestamp);
-        }
-
-        return totalNumEntries;
-    }
-
-    /**
-     * Query the current all replication stream log tail and remember the max stream tail.
-     * Query each standby site information according to the ack information to calculate the number of
-     * msgs to be sent out.
-     */
-    public void prepareClusterRoleChange() {
-        prepareClusterRoleChangeLogTail = queryStreamTail();
-        totalNumEntriesToSend = queryEntriesToSend(prepareClusterRoleChangeLogTail);
-    }
-
-    /**
-     * Query the all replication stream log tail and calculate the number of messages to be sent.
-     * If the max tail has changed, give 0 percent has done.
-     *
-     * @return Percentage of work has been done, when it return 100, the replication is done.
-     */
-    public int queryReplicationStatus() {
-        long maxTail = queryStreamTail();
-
-        /*
-         * If the tail has been moved, reset the base calculation
-         */
-        if (maxTail > prepareClusterRoleChangeLogTail) {
-            prepareClusterRoleChange();
-        }
-
-        // TODO(Xiaoqin Ma/Nan): if the max stream tail moves, it calls prepareSiteRoleChange(), which
-        //  call queryStreamTail() one more time, and will update totalNumEntriesToSend.
-        //  Then it call queryEntriesToSend() again, will get a pretty close result as currentNumEntriesToSend.
-        //  So percent calculation will always return a 0.
-        long currentNumEntriesToSend = queryEntriesToSend(prepareClusterRoleChangeLogTail);
-        log.debug("maxTail {} totalNumEntriesToSend  {}  currentNumEntriesToSend {}", maxTail, totalNumEntriesToSend, currentNumEntriesToSend);
-
-        if (totalNumEntriesToSend == 0 || currentNumEntriesToSend == 0)
-            return PERCENTAGE_BASE;
-
-        /*
-         * percentage of has been sent
-         * as the currentNumEntriesToSend is not zero, the percent should not be 100%
-         */
-        int percent = (int)((totalNumEntriesToSend - currentNumEntriesToSend)*PERCENTAGE_BASE/totalNumEntriesToSend);
-        if (percent == PERCENTAGE_BASE) {
-            percent = PERCENTAGE_BASE - 1;
-        }
-
-        return percent;
     }
 }
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/CorfuReplicationClusterManagerBaseAdapter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/CorfuReplicationClusterManagerBaseAdapter.java
@@ -5,10 +5,13 @@ import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.infrastructure.logreplication.infrastructure.CorfuReplicationDiscoveryServiceAdapter;
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo.TopologyConfigurationMsg;
+import org.corfudb.infrastructure.logreplication.proto.LogReplicationMetadata;
+
+import java.util.Map;
 
 /***
- * This is the base class for CorfuReplicationSiteManager and implements the basic functionality.
- * Any SiteMangerImplementation should extend this class or implements the interface.
+ * This is the base class for CorfuReplicationClusterManagerAdapter and implements the basic functionality.
+ * Any ClusterManger Adapter implementation should extend this class or implement the interface.
  *
  */
 @Slf4j
@@ -45,7 +48,7 @@ public abstract class CorfuReplicationClusterManagerBaseAdapter implements Corfu
         corfuReplicationDiscoveryService.prepareToBecomeStandby();
     }
 
-    public int queryReplicationStatus() {
+    public Map<String, LogReplicationMetadata.ReplicationStatusVal> queryReplicationStatus() {
         return corfuReplicationDiscoveryService.queryReplicationStatus();
     }
 }
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
@@ -7,24 +7,35 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.infrastructure.LogReplicationRuntimeParameters;
 import org.corfudb.infrastructure.logreplication.DataSender;
 import org.corfudb.infrastructure.logreplication.LogReplicationConfig;
+import org.corfudb.infrastructure.logreplication.infrastructure.ClusterDescriptor;
+import org.corfudb.infrastructure.logreplication.proto.LogReplicationMetadata;
 import org.corfudb.infrastructure.logreplication.replication.fsm.ObservableAckMsg;
-import org.corfudb.infrastructure.logreplication.replication.receive.DataReceiver;
 import org.corfudb.infrastructure.logreplication.replication.fsm.LogReplicationEvent;
 import org.corfudb.infrastructure.logreplication.replication.fsm.LogReplicationFSM;
+import org.corfudb.infrastructure.logreplication.replication.receive.LogReplicationMetadataManager;
 import org.corfudb.infrastructure.logreplication.runtime.LogReplicationClient;
 import org.corfudb.infrastructure.logreplication.replication.send.CorfuDataSender;
 import org.corfudb.infrastructure.logreplication.replication.send.logreader.DefaultReadProcessor;
 import org.corfudb.infrastructure.logreplication.replication.send.LogReplicationEventMetadata;
 import org.corfudb.infrastructure.logreplication.replication.send.logreader.ReadProcessor;
+import org.corfudb.protocols.wireprotocol.StreamAddressRange;
 import org.corfudb.protocols.wireprotocol.logreplication.LogReplicationEntry;
 import org.corfudb.protocols.wireprotocol.logreplication.MessageType;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuRuntime.CorfuRuntimeParameters;
 import org.corfudb.infrastructure.logreplication.replication.fsm.LogReplicationEvent.LogReplicationEventType;
+import org.corfudb.runtime.view.Address;
+import org.corfudb.runtime.view.stream.StreamAddressSpace;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
 
+import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * This class represents the Log Replication Manager at the source cluster.
@@ -37,7 +48,7 @@ import java.util.concurrent.Executors;
  **/
 @Data
 @Slf4j
-public class LogReplicationSourceManager implements DataReceiver {
+public class LogReplicationSourceManager {
 
     private CorfuRuntime runtime;
     /*
@@ -61,6 +72,13 @@ public class LogReplicationSourceManager implements DataReceiver {
      */
     private final LogReplicationConfig config;
 
+    /*
+     * Log Replication MetadataManager.
+     */
+    private final LogReplicationMetadataManager metadataManager;
+
+    private final LogReplicationAckReader ackReader;
+
     @VisibleForTesting
     private int countACKs = 0;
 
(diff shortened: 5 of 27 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -91,61 +91,64 @@ public long query(CorfuStoreMetadata.Timestamp timestamp, LogReplicationMetadata
...
+        // TODO pankti: Change to set the topology config id and not other fields.  This cannot be changed now as it
+        // fails later if anything is found not set.
+        for (LogReplicationMetadataKey.KeyType type : LogReplicationMetadataKey.KeyType.values()) {
             long val = Address.NON_ADDRESS;
-            if (key == LogReplicationMetadataType.TOPOLOGY_CONFIG_ID) {
+            if (type == LogReplicationMetadataKey.KeyType.TOPOLOGY_CONFIG_ID) {
                 val = topologyConfigId;
             }
-            appendUpdate(txBuilder, key, val);
+            if (type != LogReplicationMetadataKey.KeyType.UNRECOGNIZED) {
```
annym, line 149: I'm curious, if we update the topologyConfigId, do we always need to reset all other values? Let's say a new standby is added.. is this still needed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/SinkBufferManager.java
```diff
@@ -161,6 +161,10 @@ public LogReplicationEntry processMsgAndBuffer(LogReplicationEntry dataMessage)
         return null;
     }
 
+    /*private int calculateRemainingEntriesToReplicate() {
```
annym, line 164: We can remove this commented code.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -79,6 +84,13 @@
     @Getter
     private ObservableValue rxMessageCount = new ObservableValue(rxMessageCounter);
 
+    /*
+     * Periodic Thread which reads the sequence number of entries processed
+     */
+    private ScheduledExecutorService lastReplicatedEntryPoller = Executors.newSingleThreadScheduledExecutor();
```
annym, line 90: Do we need an executor service for the standby? Cause it would only be updated to true (or 100) once the snapshot is applied, right? so we can just do it directly, instead of having a thread for this every 15 seconds... anyways its not changing.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -386,4 +402,24 @@ public void shutdown() {
...
     }
+
+    /**
+     * In Snapshot Sync, if the StreamsSnapshotWriter is in the apply phase, the data is not yet
+     * consistent and cannot be read by applications.  Data is always consistent during Log Entry Sync,
+     * so always return true in that case.
+     * @return boolean
+     */
+    private boolean isDataConsistent() {
+        if (RxState.LOG_ENTRY_SYNC == rxState) {
```
annym, line 413: rxState needs to be thread safe right?
Also, do we really need this distinction? anyways if it is in LOG_ENTRY_SYNC , data_consistent was set to true (maybe we need to confirm that it is set to true even if on re-instantiation it enters directly in log entry sync)

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -648,46 +647,26 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
-                return 0;
-            }
-
-            return replicationManager.queryReplicationStatus();
-        } else {
-            log.warn("Illegal queryReplicationStatus when cluster{} with role {}",
-                    localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
-            return INVALID_REPLICATION_STATUS;
+        if (ClusterRole.ACTIVE == localClusterDescriptor.getRole()) {
+            return Integer.parseInt(logReplicationMetadataManager.getReplicationStatus());
```
annym, line 666: Pankti, do we have any test?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -648,46 +647,26 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
         }
+        // TODO pankti - boolean cannot be parsed to int
```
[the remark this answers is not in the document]
pankti-m: I changed it a bit but I think it will be best to change the metadata schema a bit and not have string for all types...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -648,46 +648,29 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
     @Override
-    public void prepareToBecomeStandby() {
-        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && replicationManager != null) {
-            replicationManager.prepareClusterRoleChange();
-        } else {
-            log.warn("Illegal prepareToBecomeStandby when cluster{} with role {}",
-                    localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
-        }
-    }
+    public void prepareToBecomeStandby() { }
```
zhangn49, line 654: Do we still need this function?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,16 +221,61 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+     *
+     * @param ackedTimestamp
+     * Note: This method of calculating the remaining entries to send is inaccurate because we consider the tail
+     * of all streams(replicated and otherwise) in Corfu and compare it with the current entry sent.  It is possible
+     * that the tail is far ahead but replicated streams have few entries to send.
+     */
+    private int calculateRemainingEntriesToSend(long ackedTimestamp) {
+        long timestamp = getMaxReplicatedStreamsTail();
+        long remainingEntriesToSend = timestamp - ackedTimestamp;
+        int percentDone = (int) (ackedTimestamp/timestamp * 100);
```
zhangn49, line 267: Looks like remainingEntriesToSend is not used?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -79,6 +81,8 @@
     @Getter
     private ObservableValue rxMessageCount = new ObservableValue(rxMessageCounter);
 
+    private AtomicInteger dataConsistent = new AtomicInteger(1);
```
zhangn49, line 84: Can we have a short comment on this dataConsistent?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
pankti-m: We do not do anything here, but the SiteManager has a workflow for prepareToBecomeStandby().  So the SM adapter calls this function.  We can change the adapter to not invoke and remove this function, but I thought it is better to keep it in case something needs to be added here in future.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
pankti-m: right.  I intended to use it somehow to calculate the remaining entries in a more accurate way, but it is not needed now.  Will remove.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
pankti-m: I have it in initCommonParams() where it first gets initialized but can move it here.

pankti-m:
Follow-up PR #2657 adds the following improvements - 
1. Calculate percentage of remaining entries to replicate in a more accurate way
 2. Add New Tests

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -311,14 +313,17 @@ private void applySnapshotSync(LogReplicationEntry message) {
         switch (message.getMetadata().getMessageMetadataType()) {
             case SNAPSHOT_MESSAGE:
                 snapshotWriter.apply(message);
-                return;
+                break;
             case SNAPSHOT_END:
+                dataConsistent.set(0);
+                logReplicationMetadataManager.setDataConsistentOnStandby(String.valueOf(dataConsistent));
```
zhangn49, line 319: I am a little confused about why the dataConsistent is initialized as 1.
If a SNAPSHOT_MESSAGE comes, dataConsistent is still 1. Is it correct?

pankti-m: Yes, dataConsistent is 1 in any phase other than apply phase of the snapshot sync.  This is because only during apply phase we have an intermediate state where replicated data is being applied and applications can read inconsistent data.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -61,119 +67,108 @@
...
+        //this.runtime = CorfuRuntime.fromParameters(runtime.getParameters());
+        //this.runtime.parseConfigurationString(runtime.getLayoutServers().get(0)).connect();
+
+        this.logReplicationFSM = new LogReplicationFSM(this.runtime, config, params.getRemoteClusterDescriptor(),
+                dataSender, readProcessor, logReplicationFSMWorkers);
+
+        this.logReplicationFSM.setTopologyConfigId(params.getTopologyConfigId());
+
+        this.metadataManager = metadataManager;
+        lastAckedTsPoller.scheduleWithFixedDelay(new TsPollingTask(), 0, 15, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
annym: Yes, if its too late to address in this PR you can add it to our issues list.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,16 +242,58 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+    }
+
+    /**
+     * Given a timestamp, calculate how many entries to be sent for all replicated streams.
+     *
+     * @param
+     */
+    private int calculateRemainingEntriesToSend(long ackedTimestamp) {
+        long timestamp = queryStreamTail();
+        long remainingEntriesToSend = timestamp - ackedTimestamp;
```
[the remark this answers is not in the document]
annym: Instead of the tails we need to query the address maps for each streams to replicate.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: Pankti I was thinking given this it might be safe to have a logging? .. not that we would do anything, but at least if they claim to have started the workflow we can confirm.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -95,60 +95,63 @@ public long query(CorfuStoreMetadata.Timestamp timestamp, LogReplicationMetadata
...
     }
 
     public void setupTopologyConfigId(long topologyConfigId) {
         CorfuStoreMetadata.Timestamp timestamp = corfuStore.getTimestamp();
-        long persistedTopologyConfigId = query(timestamp, LogReplicationMetadataType.TOPOLOGY_CONFIG_ID);
+        long persistedTopologyConfigId = query(timestamp, LogReplicationMetadataKey.KeyType.TOPOLOGY_CONFIG_ID);
 
         if (topologyConfigId <= persistedTopologyConfigId) {
             log.warn("Skip setupTopologyConfigId. the current topologyConfigId " + topologyConfigId + " is not larger than the persistedTopologyConfigID " + persistedTopologyConfigId);
-            return;
```
zhangn49, line 141: Why don't we return here?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
zhangn49: I think we should not update other values in this function, because it causes more TX conflicts, and discovery service will shutdown. If the version filed is empty, we should initialize version somewhere...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
zhangn49: I think it will bring some concurrent updates to some extent.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
xiaoqin2012: If the active cluster doesn't change, the source of data doesn't change, we don't need to reset the log replication metadata. So, we should have two API s:
1. just  updateTopologyConfigId, if the active cluster doesn't change. 
2. update Topology Id and reset logReplicationMetadata if the active cluster change, as the source of data change.

xiaoqin2012:
I am still not convinced the pros that we keep a table for this. For active cluster, can we just poll the information from standby on demand? Can you give me a list of pros/cons for keeping a table and polling on demand?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -61,119 +70,89 @@
...
+        //this.runtime.parseConfigurationString(runtime.getLayoutServers().get(0)).connect();
 
         this.logReplicationFSM = new LogReplicationFSM(this.runtime, config, params.getRemoteClusterDescriptor(),
                 dataSender, readProcessor, logReplicationFSMWorkers);
+
         this.logReplicationFSM.setTopologyConfigId(params.getTopologyConfigId());
+
+        this.metadataManager = metadataManager;
+        lastAckedTsPoller = Executors.newSingleThreadScheduledExecutor(
+            new ThreadFactoryBuilder().setNameFormat("Last Ackd Ts Reader").build());
```
zhangn49, line 153: Can we use thread name : last-ackd-ts-reader?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -61,119 +70,89 @@
...
+                ThreadFactoryBuilder().setNameFormat("state-machine-worker").build());
+        ReadProcessor readProcessor = new DefaultReadProcessor(runtime);
+
         // If this runtime has opened other streams, it appends non opaque entries and because
         // the cache is shared we end up doing deserialization. We need guarantees that this runtime is dedicated
         // for log replication exclusively.
-        this.runtime = CorfuRuntime.fromParameters(runtime.getParameters());
-        this.runtime.parseConfigurationString(runtime.getLayoutServers().get(0)).connect();
+        //this.runtime = CorfuRuntime.fromParameters(runtime.getParameters());
+        //this.runtime.parseConfigurationString(runtime.getLayoutServers().get(0)).connect();
```
zhangn49, line 144: Should we remove these lines? as well as the above comments?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -61,119 +70,89 @@
...
```
xiaoqin2012, line 145: Remove the not used code.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
xiaoqin2012, line 144: Remove not used code.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -137,21 +145,36 @@ public void setupTopologyConfigId(long topologyConfigId) {
...
 
-        for (LogReplicationMetadataType key : LogReplicationMetadataType.values()) {
-            long val = Address.NON_ADDRESS;
-            if (key == LogReplicationMetadataType.TOPOLOGY_CONFIG_ID) {
-                val = topologyConfigId;
+        for (LogReplicationMetadataType type : LogReplicationMetadataType.values()) {
+            if (type == LogReplicationMetadataType.TOPOLOGY_CONFIG_ID) {
+                appendUpdate(txBuilder, type, topologyConfigId);
+                try {
+                    IRetry.build(IntervalRetry.class, () -> {
```
zhangn49, line 159: Will we add retry to each tx commit in metadata manager? I think we should have a max retry limit here.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
         // Process ACKs from Application, for both, log entry and snapshot sync.
         if(message.getMetadata().getMessageMetadataType() == MessageType.LOG_ENTRY_REPLICATED) {
             log.debug("Log entry sync ACK received on timestamp {}", message.getMetadata().getTimestamp());
+            lastAckedTimestamp.set(message.getMetadata().getTimestamp());
             logReplicationFSM.input(new LogReplicationEvent(LogReplicationEventType.LOG_ENTRY_SYNC_REPLICATED,
                 new LogReplicationEventMetadata(message.getMetadata().getSyncRequestId(), message.getMetadata().getTimestamp())));
         } else if (message.getMetadata().getMessageMetadataType() == MessageType.SNAPSHOT_REPLICATED) {
             log.debug("Snapshot sync ACK received on base timestamp {}", message.getMetadata().getSnapshotTimestamp());
+            lastAckedTimestamp.set(message.getMetadata().getTimestamp());
             logReplicationFSM.input(new LogReplicationEvent(LogReplicationEventType.SNAPSHOT_SYNC_COMPLETE,
```
xiaoqin2012, line 235: Should it be msgSeqNum? All snapshot messages have same timestamp.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -281,7 +304,6 @@ public void setSnapshotApplied(LogReplicationEntry entry) {
         if (siteConfigID != persistSiteConfigID || ts != persistSnapStart || ts != persistSnapTranferDone) {
             log.warn("topologyConfigId " + siteConfigID + " != " + " persist " + persistSiteConfigID +  " ts " + ts +
                     " != " + "persistSnapTransferDone " + persistSnapTranferDone);
-            return;
```
zhangn49, line 284: Is this correct to remove?

pankti-m:
> I am still not convinced the pros that we keep a table for this. For active cluster, can we just poll the information from standby on demand? Can you give me a list of pros/cons for keeping a table and polling on demand?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
         // Process ACKs from Application, for both, log entry and snapshot sync.
         if(message.getMetadata().getMessageMetadataType() == MessageType.LOG_ENTRY_REPLICATED) {
             log.debug("Log entry sync ACK received on timestamp {}", message.getMetadata().getTimestamp());
+            lastAckedTimestamp.set(message.getMetadata().getTimestamp());
             logReplicationFSM.input(new LogReplicationEvent(LogReplicationEventType.LOG_ENTRY_SYNC_REPLICATED,
                 new LogReplicationEventMetadata(message.getMetadata().getSyncRequestId(), message.getMetadata().getTimestamp())));
         } else if (message.getMetadata().getMessageMetadataType() == MessageType.SNAPSHOT_REPLICATED) {
             log.debug("Snapshot sync ACK received on base timestamp {}", message.getMetadata().getSnapshotTimestamp());
+            lastAckedTimestamp.set(message.getMetadata().getTimestamp());
```
xiaoqin2012, line 234: For snapshot messages, should it be msgSeqNum? All snapshot messages have the same timestamp.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -253,15 +277,14 @@ public void setLastSnapTransferDoneTimestamp(long topologyConfigId, long ts) {
         if (topologyConfigId != persistedTopologyConfigId || ts <= persistedTopologyConfigId) {
             log.warn("The metadata is older than the persisted one. Set snapshotStart topologyConfigId " + topologyConfigId + " ts " + ts +
                     " persisteSiteConfigID " + persistedTopologyConfigId + " persistSnapStart " + persistSnapStart);
-            return;
```
zhangn49, line 256: Is this correct to remove?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -130,31 +128,37 @@ public LogReplicationSinkManager(String localCorfuEndpoint, LogReplicationConfig
...
+        initCommonParams(metadataManager, config);
+    }
 
-        /*
-         * When the server is up, it will be at LOG_ENTRY_SYNC state by default.
-         * The sender will query receiver's status and decide what type of replication to start with.
-         * It will transit to SNAPSHOT_SYNC state if it received a SNAPSHOT_START message from the sender.
-         */
-        this.rxState = RxState.LOG_ENTRY_SYNC;
+    private void initCommonParams(LogReplicationMetadataManager metadataManager, LogReplicationConfig config) {
```
zhangn49, line 137: Should we init topologyConfigId in this function?

infrastructure/proto/log_replication_metadata.proto
```diff
@@ -25,3 +25,23 @@ message LogReplicationMetadataKey {
 message LogReplicationMetadataVal {
   string val = 1;
 }
+
+/*
+ * Replication Status Key
+ */
+ message ReplicationStatusKey {
+   enum Status {
+     REPLICATION_COMPLETION = 0;
```
annym, line 34: Not sure I understand why this is an enum. Probably having as a key the Cluster UUID would make sense? Because at the active, one active can have multiple standby's, so how do we differentiate? And for the standby its ok to have its own cluster id as key. It would look something like this:
 
**Key** <uuid>                              |                      **Value**
Cluster_Standby_Paris_0001        |            ReplicationStatusVal
Cluster_London_002                     |            ReplicationStatusVal

infrastructure/proto/log_replication_metadata.proto
```diff
@@ -25,3 +25,23 @@ message LogReplicationMetadataKey {
...
+   Status key = 1;
+ }
+
+/*
+ * Replication Status Value
+ * Active Site sets the completionPercent, Standby sets the dataConsistent boolean
+ */
+message ReplicationStatusVal {
+  uint64 replicationCompletion = 1;
+  bool dataConsistent = 2;
```
annym, line 46: should we also add an enum which indicates if we are in LOG_ENTRY (delta) or SNAPSHOT (full) sync? for completeness.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -660,46 +659,38 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
-                log.warn("Illegal queryReplicationStatus when node is not a leader " +
-                        "in an ACTIVE Cluster{} ", localClusterDescriptor.getClusterId());
-                return 0;
-            }
-
-            if (replicationManager == null) {
-                log.warn("Illegal queryReplicationStatus when replication manager is null " +
-                        "in an ACTIVE Cluster{} ", localClusterDescriptor.getClusterId());
-                return 0;
+        if (ClusterRole.ACTIVE == localClusterDescriptor.getRole()) {
```
annym, line 683: We can leave this for later, but we might need to change this API a bit, because in line with my previous comment, there can be multiple standby's, so we might want to return a map if no cluster Id is specified (so we return the status for all standby's), and accept the cluster Id as argument for another API.

annym, line 683: Should this be returning our proto ReplicationStatusVal?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -160,7 +173,8 @@ public void updateVersion(String version) {
         String  persistedVersion = queryString(timestamp, LogReplicationMetadataType.VERSION);
 
         if (persistedVersion.equals(version)) {
-            log.warn("Skip update the current version {} with new version {} as they are the same", persistedVersion, version);
+            log.warn("Skip update the current version {} with new version {} as they are the same",
```
annym, line 176: nit -> Skip update of current version {} to {}, as they are the same.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+        return remainingEntriesToSend;
+    }
+
+    /**
+     * Task which periodically updates the metadata table with replication completion percentage
+     */
+    private class TsPollingTask implements Runnable {
+        @Override
+        public void run() {
+            long remainingReplicationStatus = calculateRemainingEntriesToSend(lastAckedTimestamp.get());
```
PavelZaytsev, line 296: Perhaps we can add some validation that this indeed a correct percent from 0 to 100. We just store it as a string later and then parse back as an integer. If it's incorrect it will throw a NumberFormatException

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -217,14 +231,14 @@ public boolean setSrcBaseSnapshotStart(long topologyConfigId, long ts) {
         // Update the topologyConfigId to fence all other transactions that update the metadata at the same time
         appendUpdate(txBuilder, LogReplicationMetadataType.TOPOLOGY_CONFIG_ID, topologyConfigId);
 
-        // Setup the LAST_SNAPSHOT_STARTED
+        // Setup the LAST_LAST_SNAPSHOT_STARTEDED
```
annym, line 234: typo -> STARTED

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -253,15 +267,14 @@ public void setLastSnapTransferDoneTimestamp(long topologyConfigId, long ts) {
...
-            return;
         }
 
         TxBuilder txBuilder = corfuStore.tx(NAMESPACE);
 
         //Update the topologyConfigId to fence all other transactions that update the metadata at the same time
         appendUpdate(txBuilder, LogReplicationMetadataType.TOPOLOGY_CONFIG_ID, topologyConfigId);
 
-        //Setup the LAST_SNAPSHOT_STARTED
+        //Setup the LAST_LAST_SNAPSHOT_STARTEDED
```
annym, line 277: same typo -> STARTED

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -301,21 +313,85 @@ public void setSnapshotApplied(LogReplicationEntry entry) {
...
+        txBuilder.update(REPLICATION_STATUS_TABLE, key, val, null);
+        txBuilder.commit();
+    }
+
+    public long getReplicationRemainingPercent() {
+        ReplicationStatusKey key = ReplicationStatusKey.newBuilder().setKey(
+                ReplicationStatusKey.Status.REPLICATION_COMPLETION).build();
+        CorfuRecord record = corfuStore.query(NAMESPACE).getRecord(REPLICATION_STATUS_TABLE, key);
+        if (record == null) {
+            return 0;
```
annym, line 330: maybe we should add a log here so we know that nothing has been updated to the corfuStore and not that it is complete necessarily.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -301,21 +313,85 @@ public void setSnapshotApplied(LogReplicationEntry entry) {
...
+
+    public boolean getDataConsistentOnStandby() {
+        ReplicationStatusKey key = ReplicationStatusKey.newBuilder().setKey(
+                ReplicationStatusKey.Status.IS_DATA_CONSISTENT).build();
+        CorfuRecord record = corfuStore.query(NAMESPACE).getRecord(REPLICATION_STATUS_TABLE, key);
+
+        // TODO: Initially, snapshot sync is pending so the data is not consistent.  But what if everything was applied and
+        // the standby rebooted?  In that case it is consistent
+        if (record == null) {
+            return false;
```
annym, line 353: Also here we can print some warning message or something for us to know its not yet persisted.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -21,12 +21,12 @@
...
 
 /**
- * This class manages Log Replication for multiple remote (standby) cluster's.
+ * This class manages Log Replication for multiple remote (standby) clusters.
  */
 @Slf4j
 public class CorfuReplicationManager {
 
-    public final static int PERCENTAGE_BASE = 100;
+    public final static int PERCENTAGE_HUNDRED = 100;
```
PavelZaytsev, line 29: Does not seem to be used

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -301,21 +313,85 @@ public void setSnapshotApplied(LogReplicationEntry entry) {
...
+        txBuilder.update(REPLICATION_STATUS_TABLE, key, val, null);
+        txBuilder.commit();
+    }
+
+    public boolean getDataConsistentOnStandby() {
+        ReplicationStatusKey key = ReplicationStatusKey.newBuilder().setKey(
+                ReplicationStatusKey.Status.IS_DATA_CONSISTENT).build();
+        CorfuRecord record = corfuStore.query(NAMESPACE).getRecord(REPLICATION_STATUS_TABLE, key);
+
+        // TODO: Initially, snapshot sync is pending so the data is not consistent.  But what if everything was applied and
```
annym, line 350: Even If the standby rebooted it should be persisted, right? unless it rebooted right before updating this value. (Maybe we should ensure it is atomically updated when the last snapshot entry is written)

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -301,21 +313,85 @@ public void setSnapshotApplied(LogReplicationEntry entry) {
...
+                   builder.append(getLastAppliedBaseSnapshotTimestamp());
+                   break;
+                case LAST_SNAPSHOT_SEQ_NUM:
+                   builder.append(getLastSnapSeqNum());
+                   break;
+                case LAST_LOG_ENTRY_PROCESSED:
+                   builder.append(getLastProcessedLogTimestamp());
+                   break;
+                case REMAINING_REPLICATION_PERCENT:
+                    builder.append(getReplicationRemainingPercent());
```
annym, line 384: break;

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -301,21 +313,85 @@ public void setSnapshotApplied(LogReplicationEntry entry) {
...
+                default:
```
annym, line 385: missing DATA_CONSISTENT_ON_STANDBY?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -398,14 +401,17 @@ private void processSnapshotMessage(LogReplicationEntry message) {
         switch (message.getMetadata().getMessageMetadataType()) {
             case SNAPSHOT_MESSAGE:
                 snapshotWriter.apply(message);
-                return;
+                break;
             case SNAPSHOT_END:
+                dataConsistent.set(false);
```
annym, line 406: This class has a receive method where we first receive messages. And there we have some logic to process a SNAPSHOT_START marker, so I think we should set dataConsistent to false in processSnapshotStart. As we now we're starting the transfer of a snapshot sync, so from that point until the snapshot_end is received and applied is when we'll set to true.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -398,14 +401,17 @@ private void processSnapshotMessage(LogReplicationEntry message) {
...
+                logReplicationMetadataManager.setDataConsistentOnStandby(dataConsistent.get());
                 snapshotWriter.snapshotTransferDone(message);
                 completeSnapshotApply(message);
-                return;
+                dataConsistent.set(true);
```
annym, line 410: To keep a sense of atomicity maybe we can have a private method setDataConsistency(boolean consistent) something like that where we update the local variable and the persisted...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
annym: So we can remove it from here and set it on the reception of the start marker.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
annym: +1

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+    private long calculateRemainingEntriesToSend(long ackedTimestamp) {
+        long maxReplicatedStreamTail = getMaxReplicatedStreamsTail();
+
+        if (ackedTimestamp == Address.NON_ADDRESS) {
+            return FULL_REPLICATION_REMAINING_PERCENT;
+        }
+        long remainingEntriesToSend = 0;
+        for (String stream : config.getStreamsToReplicate()) {
+            UUID streamId = CorfuRuntime.getStreamID(stream);
+            StreamAddressRange range = new StreamAddressRange(streamId, maxReplicatedStreamTail, ackedTimestamp);
```
PavelZaytsev, line 282: I just want to clarify. ackedTimestamp is the last timestamp acked by the receiver that we receive when the entry is replicated, so it should be smaller than the maxReplicatedStreamTail, for the range to be valid right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
annym: +1 maybe ack-timestamp-poller or ack-timestamp-reader

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -61,119 +70,89 @@
...
+
+    /*
+     * Periodic Thread which reads the last Acked Timestamp and writes it to the metadata table
+     */
+    ScheduledExecutorService lastAckedTsPoller;
+
+    /*
+     * Interval at which the thread reads the last Acked Timestamp
+     */
+    private static int ACKED_TS_READ_INTERVAL_SECONDS = 15;
```
annym, line 91: no need to do this for this review, but we might add an issue to track the fact that we should move this configuration out to the config file.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
         // Process ACKs from Application, for both, log entry and snapshot sync.
         if(message.getMetadata().getMessageMetadataType() == MessageType.LOG_ENTRY_REPLICATED) {
             log.debug("Log entry sync ACK received on timestamp {}", message.getMetadata().getTimestamp());
+            lastAckedTimestamp.set(message.getMetadata().getTimestamp());
```
annym, line 229: Actually, this is a bit weird. But I realized it the other day while looking at other things. This method for the Source is not really ever called. All the logic for processing ACKs is taken care in the SnapshotSender and LogEntrySender respectively, so it is there where we should be setting this lastAckedTimestamp.

annym: Actually, I think we have a major concern here and its that SnapshotSender does not really accumulate intermediate SNAPSHOT_REPLICATED, so it only updates at the end.. But you can bypass that (for now) by updating when you send something...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+
+    /**
+     * For the given replication runtime, query max stream tail for all streams to be replicated.
+     *
+     * @return max tail of all streams to be replicated for the given runtime
+     */
+    private long getMaxReplicatedStreamsTail() {
+        Map<UUID, Long> tailMap = runtime.getAddressSpaceView().getAllTails().getStreamTails();
+        long maxTail = Address.NON_ADDRESS;
+        for (String s : config.getStreamsToReplicate()) {
```
annym, line 252: s -> streamName

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+            UUID streamUuid = CorfuRuntime.getStreamID(s);
+            if (tailMap.get(streamUuid) != null) {
```
annym, line 254: if tailMap.contains(streamUuid)?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+     * streams.
+     * For Snapshot Sync, each entry sent is a snapshot.  So this function returns the total number of snapshots
+     * remaining to be sent.
+     * If the ack'd timestamp is uninitialized, it returns 100%, which means no replication has been done.
+     */
+    private long calculateRemainingEntriesToSend(long ackedTimestamp) {
+        long maxReplicatedStreamTail = getMaxReplicatedStreamsTail();
+
+        if (ackedTimestamp == Address.NON_ADDRESS) {
+            return FULL_REPLICATION_REMAINING_PERCENT;
```
annym, line 277: Is this true? what if the maxReplicatedStreamTail is -1, that means there is no data so remaining is 0%, right?

annym: We should add tests to cover all these cases!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+        }
+        long remainingEntriesToSend = 0;
+        for (String stream : config.getStreamsToReplicate()) {
+            UUID streamId = CorfuRuntime.getStreamID(stream);
+            StreamAddressRange range = new StreamAddressRange(streamId, maxReplicatedStreamTail, ackedTimestamp);
+            StreamAddressSpace addressSpace = runtime.getSequencerView().getStreamAddressSpace(range);
+            Roaring64NavigableMap map = addressSpace.getAddressesInRange(range);
```
annym, line 284: No need, the address map you requested is already confined to that range, you can directly: addressSpace.getAddressMap().getLongCardinality()

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationSourceManager.java
```diff
@@ -247,17 +226,75 @@ public LogReplicationEntry receive(LogReplicationEntry message) {
...
+            remainingEntriesToSend += map.getLongCardinality();
+        }
+        return remainingEntriesToSend;
```
annym, line 287: should we add a logging before returning?

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1380,6 +1377,11 @@ private void verifyPersistedLogEntryMetadata() {
         assertThat(expectedAckTimestamp == lastLogProcessed).isTrue();
     }
 
+    private void verifyRemainingReplicationPercent(long remainingPercent) {
```
annym, line 1380: Where is this being used?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
pankti-m: we will have to do it now if we change the key in the replication status metadata protobuf

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
pankti-m: sure

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
pankti-m: I did not remove so that it can be used later when we add tests

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/CorfuReplicationClusterManagerBaseAdapter.java
```diff
@@ -45,7 +46,8 @@ public void prepareToBecomeStandby() {
         corfuReplicationDiscoveryService.prepareToBecomeStandby();
     }
 
-    public int queryReplicationStatus() {
-        return corfuReplicationDiscoveryService.queryReplicationStatus();
+    public LogReplicationMetadata.ReplicationStatusVal queryReplicationStatus() {
+        return corfuReplicationDiscoveryService.queryReplicationStatus(
+            topologyConfig.getClusters(0).getId());
```
hisundar, line 51: It's not guaranteed to always place the local cluster in the 0th index position right?
Am I mistaken?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -660,46 +664,36 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
-                        "in an ACTIVE Cluster{} ", localClusterDescriptor.getClusterId());
-                return 0;
-            }
-
-            return replicationManager.queryReplicationStatus();
-        } else {
-            log.warn("Illegal queryReplicationStatus when cluster{} with role {}",
-                    localClusterDescriptor.getClusterId(), localClusterDescriptor.getRole());
-            return INVALID_REPLICATION_STATUS;
+    public LogReplicationMetadata.ReplicationStatusVal queryReplicationStatus(String clusterId) {
```
hisundar, line 687: I only see one type - is it possible to overload the queryReplicationStatus to return the full map?

PR #2643, opened 2020-07-20
Config the message size for both snapshot full sync and delta sync.

xiaoqin2012:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -311,6 +324,16 @@ void verifyTables(HashMap<String, CorfuTable<Long, Long>> tables0, HashMap<Strin
             }
     }
 
+    void waitData(HashMap<String, CorfuTable<Long, Long>> tables, HashMap<String, HashMap<Long, Long>> hashMap) {
```
pankti-m, line 327: what is the purpose of this method?

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1042,9 +1094,12 @@ public void testLogEntrySyncLargeTables() throws Exception {
 
     /* ********************** AUXILIARY METHODS ********************** */
 
+    private void generateTxCrossTables(Set<String> crossTableTransactions, boolean startCrossTx, int numKeys) throws Exception {
+        generateTxCrossTables(crossTableTransactions, startCrossTx, numKeys, 0);
```
pankti-m, line 1098: numKeys -> NUM_KEYS

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1042,9 +1094,12 @@ public void testLogEntrySyncLargeTables() throws Exception {
...
+    }
 
-    // startCrossTx indicates if we start with a transaction across Tables
-    private void testSnapshotSyncCrossTables(Set<String> crossTableTransactions, boolean startCrossTx) throws Exception {
+        // startCrossTx indicates if we start with a transaction across Tables
+    private void generateTxCrossTables(Set<String> crossTableTransactions, boolean startCrossTx, int numKeys, int startValue) throws Exception {
```
pankti-m, line 1102: numKeys -> NUM_KEYS

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1054,20 +1109,20 @@ private void testSnapshotSyncCrossTables(Set<String> crossTableTransactions, boo
...
-        generateTransactionsCrossTables(srcCorfuTables, Collections.singleton(t1), srcDataForVerification, NUM_KEYS, srcDataRuntime, NUM_KEYS);
+        generateTransactionsCrossTables(srcCorfuTables, Collections.singleton(t1), srcDataForVerification, numKeys, srcDataRuntime, numKeys);
 
         // Write data to t2
-        generateTransactionsCrossTables(srcCorfuTables, Collections.singleton(t2), srcDataForVerification, NUM_KEYS, srcDataRuntime, 0);
+        generateTransactionsCrossTables(srcCorfuTables, Collections.singleton(t2), srcDataForVerification, numKeys, srcDataRuntime, 0);
 
         // Write data across to tables specified in crossTableTransactions in transaction
-        generateTransactionsCrossTables(srcCorfuTables, crossTableTransactions, srcDataForVerification, NUM_KEYS, srcDataRuntime, NUM_KEYS*2);
+        generateTransactionsCrossTables(srcCorfuTables, crossTableTransactions, srcDataForVerification, numKeys, srcDataRuntime, numKeys*2);
```
pankti-m, line 1125: numKeys -> NUM_KEYS

test/src/test/resources/logback-test.xml
```diff
@@ -43,9 +43,9 @@
         <!--<appender-ref ref="MetricsRollingFile" />-->
     </logger>
 
-    <!--<root level="DEBUG">-->
-        <!--&lt;!&ndash;<appender-ref ref="FILE" />&ndash;&gt;-->
-        <!--<appender-ref ref="STDOUT" />-->
-        <!--&lt;!&ndash;<appender-ref ref="MetricsRollingFile" />&ndash;&gt;-->
-    <!--</root>-->
+    <root level="INFO">
```
pankti-m, line 46: was this intentional?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterConfig.java
```diff
@@ -45,7 +45,7 @@ private DefaultClusterConfig() {
     private static String standbyLogReplicationPort = "9020";
 
     @Getter
-    private static int logSenderBufferSize = 20;
+    private static int logSenderBufferSize = 1;
```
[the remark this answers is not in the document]
xiaoqin2012: As we increase the size of the message to 64MB, there is no point to have a big buffer.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
```diff
@@ -98,21 +93,37 @@ void processMsg(LogReplicationEntry txMessage) {
...
                     " persistSnapDone " + persistSnapDone + " persistLogTs " + persistLogTS);
             return;
         }
 
+        // Skip Opaque entries with timestamp that are not larger than persistedTs
+        OpaqueEntry[] newOpaqueEntryList = opaqueEntryList.stream().filter(x->x.getVersion() > persistLogTS).toArray(OpaqueEntry[]::new);
+
+        // Check that all opaque entries contain the correct streams
+        for (OpaqueEntry opaqueEntry : newOpaqueEntryList) {
+            if (!streamMap.keySet().containsAll(opaqueEntry.getEntries().keySet())) {
```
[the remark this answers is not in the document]
xiaoqin2012: For tx opqaque entry is a SMRentry that contains all data changed in one transaction.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -36,10 +36,12 @@
      */
     private static final String config_file = "/config/corfu/corfu_replication_config.properties";
 
+    private final int DEFAULT_ACK_CNT = 1;
+
```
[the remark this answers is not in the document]
xiaoqin2012: This is the count of messages. As the message becomes bigger now, so we send an ACK for each message.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +62,116 @@
...
-                OpaqueEntry entry = (OpaqueEntry) stream.iterator.next();
-                stream.maxVersion = Math.max(stream.maxVersion, entry.getVersion());
-                list.addAll(entry.getEntries().get(stream.uuid));
+            while (currentMsgSize < maxDataSizePerMsg) {
+                if (lastEntry != null) {
+                    List<SMREntry> smrEntries = lastEntry.getEntries().get(stream.uuid);
+                    if (smrEntries != null) {
+                        int currentEntrySize = calculateSize(smrEntries);
+
+                        if (currentEntrySize > DEFAULT_LOG_REPLICATION_DATA_MSG_SIZE) {
```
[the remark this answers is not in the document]
xiaoqin2012: If an entry's size is bigger than the user's setup but is smaller than the default that we can handle, we want that the log replication still make progress, but log a warning. But if the entry's size is bigger than we can handle, we will fail.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +62,116 @@
...
+                            log.error("The current entry size {} is bigger than the maxDataSizePerMsg {} supported", currentEntrySize, DEFAULT_LOG_REPLICATION_DATA_MSG_SIZE);
```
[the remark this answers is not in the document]
xiaoqin2012: This is an error and will cause the log replication transport layer failure.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +62,116 @@
...
+                        } else if (currentEntrySize > maxDataSizePerMsg) {
+                            log.warn("The current entry size {} is bigger than the configured maxDataSizePerMsg {}",
```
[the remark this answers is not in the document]
xiaoqin2012: This is a warning.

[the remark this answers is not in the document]
xiaoqin2012: As I rebased, this is what Anny and I discussed from the previous PR.

test/src/test/java/org/corfudb/infrastructure/logreplication/ReplicationReaderWriterTest.java
```diff
@@ -247,4 +257,39 @@ public void testUFOWithLogUpdate() throws NoSuchMethodException, IllegalAccessEx
         assertThat(bSet.containsAll(aSet)).isTrue();
         assertThat(aSet.containsAll(bSet)).isTrue();
     }
+
+    public static void readLogEntryMsgs(List<LogReplicationEntry> msgQ, Set<String> streams, CorfuRuntime rt) throws
```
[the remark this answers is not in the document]
xiaoqin2012: This is a function used by this test. This test the reader/writer directly not through the transport layer.

test/src/test/java/org/corfudb/infrastructure/logreplication/ReplicationReaderWriterTest.java
```diff
@@ -247,4 +257,39 @@ public void testUFOWithLogUpdate() throws NoSuchMethodException, IllegalAccessEx
...
+                msgQ.add(entry);
+            }
+
+            System.out.println(" msgQ size " + msgQ.size());
+
+        } while (entry != null);
+    }
+
+    public static void writeLogEntryMsgs(List<LogReplicationEntry> msgQ, Set<String> streams, CorfuRuntime rt) {
+        org.corfudb.infrastructure.logreplication.LogReplicationConfig config = new LogReplicationConfig(streams);
```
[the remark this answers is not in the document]
xiaoqin2012: Same as the above.

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -60,6 +60,8 @@
     private static final int SHUTDOWN_RETRIES = 10;
     private static final long SHUTDOWN_RETRY_WAIT = 500;
 
+    private static final int MSG_SIZE = 131072;
```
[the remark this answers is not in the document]
xiaoqin2012: This is set for the test case. Not too big as then all entries will be sent over in one message.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
```diff
@@ -98,21 +93,37 @@ void processMsg(LogReplicationEntry txMessage) {
...
-                txMessage.getMetadata().getPreviousTimestamp() != persistLogTS) {
+                entryTS <= persistLogTS) {
             log.warn("Skip write this msg {} as its timestamp is later than the persisted one " +
                     txMessage.getMetadata() +  " persisteSiteConfig " + persistSiteConfigID + " persistSnapStart " + persistSnapStart +
                     " persistSnapDone " + persistSnapDone + " persistLogTs " + persistLogTS);
             return;
         }
 
+        // Skip Opaque entries with timestamp that are not larger than persistedTs
+        OpaqueEntry[] newOpaqueEntryList = opaqueEntryList.stream().filter(x->x.getVersion() > persistLogTS).toArray(OpaqueEntry[]::new);
```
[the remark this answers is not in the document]
xiaoqin2012: For opaque entry, the version is the timestamp.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -84,39 +99,100 @@ boolean shouldProcess(OpaqueEntry entry) throws ReplicationReaderException {
...
+                    break;
+                }
+
+                lastOpaqueEntry = txOpaqueStream.next();
             }
+
+            log.trace("Generate LogEntryDataMessage size {} with {} entries for maxDataSizePerMsg {}. lastEnry size {}",
+                    currentMsgSize, opaqueEntryList.size(), maxDataSizePerMsg, lastOpaqueEntry == null? 0 : currentEntrySize);
+
+            if (opaqueEntryList.size() == 0 && hasNoiseData) {
```
annym, line 181: what happens if there is noisy data but the list is not empty? we should still abort, right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -84,39 +99,100 @@ boolean shouldProcess(OpaqueEntry entry) throws ReplicationReaderException {
...
+                throw new IllegalTransactionStreamsException("There are noisy streams in the transaction log entry");
+            }
+
+            if (opaqueEntryList.size() == 0) {
```
annym, line 185: It's cleaner if generateMessageWithOpaqueEntryList accepts the empty list and returns a null. Less if conditions.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +67,103 @@
     public StreamsSnapshotReader(CorfuRuntime runtime, LogReplicationConfig config) {
         this.rt = runtime;
         this.rt.parseConfigurationString(runtime.getLayoutServers().get(0)).connect();
+        this.maxDataSizePerMsg = config.getMaxDataSizePerMsg();
         streams = config.getStreamsToReplicate();
+        log.debug("The maxDataSizePerMsg {} ", maxDataSizePerMsg);
```
annym, line 72: can this logger be removed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +67,103 @@
...
+                        }
+
+                        smrList.addAll(smrEntries);
+                        currentMsgSize += currentEntrySize;
+                        stream.maxVersion = Math.max(stream.maxVersion, lastEntry.getVersion());
+                    }
+                    lastEntry = null;
+                }
+
+                if (stream.iterator.hasNext()) {
```
annym, line 151: L 151 -157 can be simplified:

if (!stream.iterator.hasNext()) {
    break;
}

lastEntry = (OpaqueEntry) stream.iterator.next();

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +67,103 @@
...
+            while (currentMsgSize < maxDataSizePerMsg) {
+                if (lastEntry != null) {
+                    List<SMREntry> smrEntries = lastEntry.getEntries().get(stream.uuid);
+                    if (smrEntries != null) {
+                        int currentEntrySize = ReaderUtility.calculateSize(smrEntries);
+
+                        if (currentEntrySize > MAX_LOG_REPLICATION_DATA_MSG_SIZE_SUPPORTED) {
+                            log.error("The current entry size {} is bigger than the maxDataSizePerMsg {} supported", currentEntrySize, MAX_LOG_REPLICATION_DATA_MSG_SIZE_SUPPORTED);
+                            throw new IllegalSnapshotEntrySizeException(" The snapshot entry is bigger than the system supported");
+                        } else if (currentEntrySize > maxDataSizePerMsg) {
```
annym, line 133: I'm confused, why do we have this same comparison of a single currentEntrySize twice? shouldn't it only compare with maxDataSizePerMsg, as this is the user configured (which could default to MAX_LOG_REPLICATION_DATA_MSG_SIZE_SUPPORTED)

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +67,103 @@
...
+                            observeBiggerMsg.setValue(observeBiggerMsg.getValue()+1);
+                            log.warn("The current entry size {} is bigger than the configured maxDataSizePerMsg {}",
+                                    currentEntrySize, maxDataSizePerMsg);
+                        }
+
+                        // Skip append this entry in this message. Will process it first at the next round.
+                        if (currentEntrySize + currentMsgSize > maxDataSizePerMsg && currentMsgSize != 0) {
```
annym, line 140: why do we need currentMsgSize != 0? If we already checked that currentEntrySize is bigger previously and we throw an exception.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/logreplication/LogReplicationEntry.java
```diff
@@ -13,41 +21,102 @@
...
 
-    private byte[] payload;
+    private List<OpaqueEntry> opaqueEntryList = new ArrayList<>();
 
-    public LogReplicationEntry(LogReplicationEntryMetadata metadata, byte[] payload) {
-        this.payload = payload;
+
+    // Only used by test cases
+    @VisibleForTesting
+    byte[] payload;
```
annym, line 35: private

test/src/test/java/org/corfudb/integration/AbstractIT.java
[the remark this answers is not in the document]
annym: Can we add this comment in the code so it is easy to follow?

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -313,6 +315,7 @@ public static Process runReplicationServer(int port, String pluginConfigFilePath
                 .setHost(DEFAULT_HOST)
                 .setPort(port)
                 .setPluginConfigFilePath(pluginConfigFilePath)
+                .setMsg_size(MSG_SIZE)
```
annym, line 318: Camel case please.

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -507,6 +510,7 @@ public Process runServer() throws IOException {
         private String compressionCodec = null;
         private String pluginConfigFilePath = null;
         private String logPath = null;
+        private int msg_size = 0;
```
annym, line 513: Java uses CamelCase.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
annym, line 326: private

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -695,7 +742,7 @@ public void testLogEntrySyncValidCrossTablesWithTriggerTimeout() throws Exceptio
         expectedAckMessages =  NUM_KEYS*WRITE_CYCLES;
 
         testConfig.clear().setDropMessageLevel(2);
-        startLogEntrySync(crossTables, WAIT.ON_ERROR);
+        startLogEntrySync(crossTables, WAIT.ON_TIMEOUT_ERROR);
```
annym, line 719: Why did this change?

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -811,16 +832,22 @@ public void testLogEntrySyncValidCrossTablesWithWritingAtSrc() throws Exception
         testConfig.clear();
         testConfig.setWritingSrc(true);
         testConfig.setDeleteOP(true);
-
         testConfig.setWaitOn(WAIT.ON_ACK);
-        startLogEntrySync(crossTables, WAIT.ON_ACK, false);
+
+        HashSet<WAIT> waitHashSet = new HashSet<>();
+        waitHashSet.add(WAIT.ON_ACK);
+        startLogEntrySync(crossTables, waitHashSet, true);
```
annym, line 839: why did this change? there is a method that takes only one condition.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -811,16 +832,22 @@ public void testLogEntrySyncValidCrossTablesWithWritingAtSrc() throws Exception
...
         expectedAckTimestamp = Long.MAX_VALUE;
-        // Verify Data on Destination site
-        System.out.println("****** Verify Data on Destination");
+
         // Because t2 is not specified as a replicated table, we should not see it on the destination
         srcDataForVerification.get(t2).clear();
 
+        // Verify Data on Destination site
+        System.out.println("****** Wait Data on Destination");
+        waitData(dstCorfuTables, srcDataForVerification);
```
annym, line 848: The intention of the observables and the WAIT.ON_ACK is to avoid while loops that are just stuck there until a condition is met. Can't we set the expected number of ACKS by knowing the size of what we wrote and how many messages will be sent, hence how many calks we will receive back?

annym: All you need is to set expectedAckMessages to the correct value.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -971,6 +998,8 @@ public void testLogEntrySyncWithTrim() throws Exception {
         // Setup Environment: two corfu servers (source & destination)
         setupEnv();
 
+        log.info("Have setutEnv Done");
```
annym, line 1001: required?

test/src/test/java/org/corfudb/integration/SourceForwardingDataSender.java
```diff
@@ -50,7 +50,9 @@
 
     final static int DROP_INCREMENT = 4;
 
-    private int firstDrop = DROP_INCREMENT;
+    private int DroppingNum = 2;
```
annym, line 53: droppingNum (not starting with caps)

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
xiaoqin2012: Wait replication data reach at the standby cluster.

test/src/test/resources/logback-test.xml
xiaoqin2012: Change back.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -216,16 +219,30 @@ public String getPluginConfigFilePath() {
         return pluginConfigFilePath == null ? PLUGIN_CONFIG_FILE_PATH : pluginConfigFilePath;
     }
 
-    public int getSnapshotSyncBatchSize() {
-        Integer snapshotSyncBatchSize = getServerConfig(Integer.class, "--snapshot-batch");
-        return snapshotSyncBatchSize == null ? SnapshotSender.DEFAULT_SNAPSHOT_BATCH_SIZE : snapshotSyncBatchSize;
+    /**
```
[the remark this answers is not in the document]
xiaoqin2012: You gave the word earlier. If you want to change, you can change it in your PR.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -56,68 +62,116 @@
...
         return txMsg;
     }
 
     /**
-     * Read numEntries from the current stream.
+     * Given a list of SMREntries, calculate the total sizeInBytes.
+     * @param smrEntries
+     * @return
+     */
+    public static int calculateSize(List<SMREntry> smrEntries) {
```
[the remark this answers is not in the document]
xiaoqin2012: Move to an utility class.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -223,4 +281,21 @@ public void setTopologyConfigId(long topologyConfigId) {
         this.topologyConfigId = topologyConfigId;
     }
 
+    /**
+     * Record a list of SMR entries
+     */
+    static class SMREntryList {
```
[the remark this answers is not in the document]
xiaoqin2012: made the change.

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -556,7 +564,7 @@ public String getOptionsString() {
          * @throws IOException
          */
         public Process runServer() throws IOException {
-            final String serverConsoleLogPath = CORFU_LOG_PATH + File.separator + host + "_" + port + "_consolelog";
+            final String serverConsoleLogPath = "/Users/maxi/Projects/tmp/test.result"; //CORFU_LOG_PATH + File.separator + host + "_" + port + "_consolelog";
```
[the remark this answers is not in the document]
xiaoqin2012: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/LogReplicationConfig.java
```diff
@@ -30,16 +55,21 @@
      */
     public LogReplicationConfig(Set<String> streamsToReplicate) {
         this.streamsToReplicate = streamsToReplicate;
+        this.maxNumMsgPerBatch = DEFAULT_MAX_NUM_MSG_PER_BATCH;
```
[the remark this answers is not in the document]
xiaoqin2012: It is harder for later code review. It is easier logic to init all things together in one place instead of jumping code around for simple logic.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/DefaultClusterConfig.java
```diff
@@ -45,7 +45,7 @@ private DefaultClusterConfig() {
     private static String standbyLogReplicationPort = "9020";
 
     @Getter
-    private static int logSenderBufferSize = 20;
+    private static int logSenderBufferSize = 2;
```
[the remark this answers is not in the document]
xiaoqin2012: At both the sender and receiver we have buffers for messages. As the message is big, the standard double buffering may be enough.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/LogReplicationConfig.java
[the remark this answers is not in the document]
annym: it is actually a good practice, to avoid code duplication, and having all inits in one single place or when a change comes, you'll have to change in several places (prone to errors):

https://stackoverflow.com/questions/7577627/purpose-of-constructor-chaining

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
```diff
@@ -74,4 +75,24 @@ public boolean verifyMessageType(LogReplicationEntry entry) {
...
+         */
+        for (LogReplicationEntry entry : buffer.values()) {
+            LogReplicationEntryMetadata metadata = entry.getMetadata();
+            if (metadata.getTimestamp() < lastProcessedSeq) {
+                //remove it
+                buffer.remove(metadata.getPreviousTimestamp());
+            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq && metadata.getTimestamp() > lastProcessedSeq) {
+                sinkManager.processMessage(entry);
+                ackCnt++;
+                buffer.remove(lastProcessedSeq);
```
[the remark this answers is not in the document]
xiaoqin2012: For each message, it has the previous time stamp and the currentTimestamp( the max ts of all transactions in this message). So if all transactions in one message have been processed, we will remove it. If it has transactions that haven't been processed, we will still buffer it and process it later.

This happens because the possible leadership change and the two leaders at the active cluster may packaging the messages differently as the polling transaction log speed  are different at the different nodes.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
```diff
@@ -74,4 +75,24 @@ public boolean verifyMessageType(LogReplicationEntry entry) {
 
         return true;
     }
+
+    public void processBuffer() {
```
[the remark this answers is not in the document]
xiaoqin2012: In your case, when it receives 5, it will process 5 and right after process it:
1. it will update the lastSeqNum
2. it will look at the buffer to see if there are any messages are in order to be processed.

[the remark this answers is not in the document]
xiaoqin2012: the map is indexed according to the preTs/preSeq number. 
For your case the map: 
after process message 2, the lastTs = 2
for message 7, the preTs is 5
for message 8, the preTs is 7
for message 5, the preTs is 2: we will process it and update the lastTs as 5, use 5 as the index to lookup and find message 7.

test/src/test/java/org/corfudb/infrastructure/logreplication/LogReplicationFSMTest.java
```diff
@@ -349,8 +349,6 @@ public void testSnapshotSyncStreamImplementation() throws Exception {
 
         Queue<LogReplicationEntry> listenerQueue = ((TestDataSender) dataSender).getEntryQueue();
 
-        assertThat(LARGE_NUM_ENTRIES/ StreamsSnapshotReader.MAX_NUM_SMR_ENTRY).isLessThanOrEqualTo(listenerQueue.size());
```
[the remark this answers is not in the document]
xiaoqin2012: Now we are using the size not the number of entries.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/LogEntrySender.java
```diff
@@ -32,7 +33,7 @@
     /*
      * Implementation of Log Entry Reader. Default implementation reads at the stream layer.
      */
-    private LogEntryReader logEntryReader;
+    private StreamsLogEntryReader logEntryReader;
```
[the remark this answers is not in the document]
xiaoqin2012: We should do some clean up to remove the interface and testone.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
xiaoqin2012: I want to put all things clear for each test case instead of using the wrapper, otherwise, while debugging, I need to jump around to see what are the real arguments passed.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
xiaoqin2012: It is not accurate right. Even we know the value size, but not the MSR entry size in advance.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
xiaoqin2012: When the test goes wrong, we know the process and where it stuck.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -84,39 +99,100 @@ boolean shouldProcess(OpaqueEntry entry) throws ReplicationReaderException {
...
+        // For interested entry, if its size is too big we should skip and report error
+        if (currentEntrySize > maxDataSizePerMsg) {
+            log.error("The current entry size {} is bigger than the maxDataSizePerMsg {} supported",
+                    currentEntrySize, MAX_LOG_REPLICATION_DATA_MSG_SIZE_SUPPORTED);
+            hasNoiseData = true;
+            return false;
+        }
+
+        if (currentEntrySize > maxDataSizePerMsg) {
+            log.warn("The current entry size {} is bigger than the configured maxDataSizePerMsg {}",
```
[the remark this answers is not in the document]
annym: I think this was accidentally copied. Same as in L.113.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
annym: ??

annym:
@xiaoqin2012 Also one last thing, the fix to the IT that was failing was reverted by one of the changes here, please change back cause it is failing on your PR:

in testLogEntrySyncValidCrossTablesWithWritingAtSrc you must change:

startLogEntrySync(crossTables, waitHashSet, true); for --->
startLogEntrySync(crossTables, waitHashSet, false);

Otherwise data continues to be written while LogEntry is evaluated and so the active can have more entries than the receiver at the moment of eval.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
xiaoqin2012: For example, we have log0, log1, log2, log3. All logs from 0 to 2 are good, but log3 is bad. Should we transfer log0, log1, log2, first, then abort when process log3?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -84,39 +99,100 @@ boolean shouldProcess(OpaqueEntry entry) throws ReplicationReaderException {
...
+                        break;
+                    }
+
+                    // Add the lastOpaqueEntry to the current message.
+                    opaqueEntryList.add(lastOpaqueEntry);
+                    currentMsgSize += currentEntrySize;
+                    lastOpaqueEntry = null;
+                }
+
+                if (hasNoiseData) {
```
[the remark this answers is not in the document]
xiaoqin2012: Consider that the max_msg_size is 100 and there are only two entrys. entry0 with size  90, entry1 with size 20
after processing the first entry, the while is correct, then we set  lastEntry = entry2.
but entry2 could not go with the first msg.
while constructing the second message, if there are no new entrys, the hasNext gives false, but we need to add entry2 first. So the logic is check the lastEntry first. then calling the next.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -84,39 +99,100 @@ boolean shouldProcess(OpaqueEntry entry) throws ReplicationReaderException {
...
+        hasNoiseData = true;
+        return false;
+    }
+
+    boolean checkSizeOK(OpaqueEntry entry, int currentMsgSize, int currentEntrySize) {
+        // For interested entry, if its size is too big we should skip and report error
+        if (currentEntrySize > maxDataSizePerMsg) {
+            log.error("The current entry size {} is bigger than the maxDataSizePerMsg {} supported",
+                    currentEntrySize, MAX_LOG_REPLICATION_DATA_MSG_SIZE_SUPPORTED);
+            hasNoiseData = true;
```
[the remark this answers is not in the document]
xiaoqin2012: Changed it.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
xiaoqin2012: User can setup a different one. Supported is the one we can support. The user setup one is the one the user thinks it gives the best performance. Like 64MB is the max we can support. But in reality maybe 2M gives the best performance. But it depends the real hardware config that the users are aware of or have done experiments with.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
xiaoqin2012: If it is the only entry and its size is bigger than the configured but smaller than the max_supported, we still continue the replication , right? I thought we have discussed this before.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
xiaoqin2012: It is much clear about what each test is wait on.

xiaoqin2012:
I fixed that test testLogEntrySyncValidCrossTablesWithWritingAtSrc with waitData. The goal of the test is running tx while doing log entry transfer.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
annym: Well from early discussions with Medhavi my understanding is that it's just an invalid state, so we can directly abort without processing anything.

PR #2645, opened 2020-07-21
Fix Log Replication Failing IT

annym:
## Overview

Description:

Fix testLogEntrySyncValidCrossTablesWithWritingAtSrc IT

### DIFF
--- a/test/src/test/java/org/corfudb/integration/LogReplicationIT.java
+++ b/test/src/test/java/org/corfudb/integration/LogReplicationIT.java
@@ -420,7 +420,7 @@ public class LogReplicationIT extends AbstractIT implements Observer {
         crossTables.add(t0);
         crossTables.add(t1);
 
-        testSnapshotSyncCrossTables(crossTables, true);
+        writeCrossTableTransactions(crossTables, true);
 
         // Start Snapshot Sync
         startSnapshotSync(crossTables);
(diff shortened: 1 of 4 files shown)


pankti-m:
Is see a failure in testLogEntrySyncValidCrossTablesWithWritingAtSrc().  Is it a known issue which this is fixing?

annym:
> Is see a failure in testLogEntrySyncValidCrossTablesWithWritingAtSrc(). Is it a known issue which this is fixing?

Yes this is the one I'm trying to fix. Let's wait for results to come back and I'll re-run a couple of times to confirm. Cause locally it does not fail any more.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1190,7 +1192,7 @@ private LogReplicationFSM startLogEntrySync(Set<String> tablesToReplicate, Set<W
         blockUntilExpectedAckTs.acquire();
         expectedAckTimestamp = srcDataRuntime.getAddressSpaceView().getLogTail();
 
-        // Block until the snapshot sync completes == one ACK is received by the source manager, or an error occurs
+        // Block until the expected ACK Timestamp is reached
         System.out.println("\n****** Wait until the wait condition is met");
         if (waitConditions.contains(WAIT.ON_ERROR)) {
```
xiaoqin2012, line 1197: Change the system.out to log.in.

annym: System outs for testing are all over the place so even changing one does not do much. If this is something that we want we can make this in a separate PR where we change all of them.

PR #2648, opened 2020-07-22
Introducing leadership change tests

PavelZaytsev:
## Overview

Description:

* Leadership change during the replication tests.


test/src/test/java/org/corfudb/integration/CorfuReplicationLeaderIT.java
```diff
@@ -0,0 +1,871 @@
...
+            final int third = 3;
+            while (replicationTableStandby.size() < numWriteIterations / third) {
+                log.info("Standby table size: {}", replicationTableStandby.size());
+                Sleep.sleepUninterruptibly(Duration.ofSeconds(1));
+            }
+
+            LockClient lockClient = new LockClient(UUID.randomUUID(), LockConfig.builder().build(),
+                    activeRuntime);
+
+            Uuid firstLeastOwnerId;
```
pankti-m, line 196: typo - Least -> Lease

test/src/test/java/org/corfudb/integration/CorfuReplicationLeaderIT.java
```diff
@@ -0,0 +1,871 @@
...
+    }
+
+    private Tuple<UUID, CorfuInterClusterReplicationServer> runReplicationFuture(int port,
+                                                                                 String pluginConfigPath) {
+        UUID uuid = UUID.randomUUID();
+        return runReplicationFuture(port, uuid, pluginConfigPath);
+    }
+
+    @Test
+    public void testLockHolderChangesSourceDuringLogEntrySync() throws Exception {
```
pankti-m, line 141: can we separate the common code in all tests into a separate method so that the tests are smaller and more readable?

test/src/test/java/org/corfudb/integration/LockIT.java
```diff
@@ -60,55 +61,55 @@
...
-               assertThat(client.getLocks().get(lockId).getState().getType()).isEqualTo(LockStateType.HAS_LEASE);
-               assertThat(lockRevokedObservables.get(clientId).getValue()).isEqualTo(0);
-           }
-       } catch (Exception e) {
-           System.out.println("Unexpected exception: " + e);
-           throw e;
-       } finally {
-           shutdown();
-       }
+            // Since this is the only client, lock should've been acquired, verify, block until condition is met
```
pankti-m, line 82: is this indentation right?

test/src/test/resources/logback-test.xml
```diff
@@ -43,9 +43,9 @@
         <!--<appender-ref ref="MetricsRollingFile" />-->
     </logger>
 
-    <!--<root level="DEBUG">-->
-        <!--&lt;!&ndash;<appender-ref ref="FILE" />&ndash;&gt;-->
-        <!--<appender-ref ref="STDOUT" />-->
-        <!--&lt;!&ndash;<appender-ref ref="MetricsRollingFile" />&ndash;&gt;-->
-    <!--</root>-->
+<!--    <root level="DEBUG">-->
```
pankti-m, line 46: revert to original

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -96,12 +98,75 @@ public LockClient(UUID clientId, CorfuRuntime corfuRuntime) throws NoSuchMethodE
...
+                .setLockGroup(lockConfig.getLockGroup())
+                .setLockName(lockConfig.getLockName())
+                .build();
+        if (!locks.containsKey(lockId)){
+            throw new IllegalStateException("Can only deregister if the known " +
+                    "lock is present in the map.");
+        }
+        Lock lock = locks.get(lockId);
+
+        lock.input(LockEvent.LEASE_EXPIRED);
```
pankti-m, line 133: The LockClient can be in any state when calling deregisterInterest() and LEASE_EXPIRED event can only be received in HAS_LEASE state.  I see we set the state correctly in discoveryService but I think there should be validation here as well

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -96,12 +98,75 @@ public LockClient(UUID clientId, CorfuRuntime corfuRuntime) throws NoSuchMethodE
...
+        if (!locks.containsKey(lockId)){
+            throw new IllegalStateException("Can only resume interest if the known " +
+                    "lock is present in the map.");
+        }
+
+        Lock lock = locks.get(lockId);
+
+        monitorLocks();
+
+        lock.input(LockEvent.LEASE_REVOKED);
```
pankti-m, line 151: why do we revoke the lease on resumeInterest?  Also we should add the same validation here.

utils/src/main/java/org/corfudb/utils/lock/states/LockEvent.java
```diff
@@ -13,5 +13,6 @@
     LEASE_RENEWED,              // lease for the lock has been renewed
     LEASE_REVOKED,              // lease for the lock has been revoked, some other process got the lock
     LEASE_EXPIRED,              // lease for the lock has expired as it could not be renewed.
-    UNRECOVERABLE_ERROR         // unrecoverable error occured
+    UNRECOVERABLE_ERROR,        // unrecoverable error occured
+    FORCE_LEASE_ACQUIRED,       // lease for the lock has been forcefully acquired
```
pankti-m, line 17: can you update the diagram in Lock.java to add this transition?

utils/src/main/java/org/corfudb/utils/lock/LockConfig.java
```diff
@@ -0,0 +1,29 @@
+package org.corfudb.utils.lock;
```
pankti-m, line 1: in discoveryService the LockConfig is generated using the properties file.  Why do we need both?

utils/src/main/java/org/corfudb/utils/lock/persistence/LockStore.java
```diff
@@ -118,6 +130,33 @@ public boolean acquire(LockId lockId) throws LockStoreException {
...
+        else {
+            LockData newLockData = LockData.newBuilder()
+                    .setLockId(lockId)
+                    .setLeaseOwnerId(clientId)
+                    .setLeaseRenewalNumber(0)
+                    .setLeaseAcquisitionNumber(lockInDatastore.get().getLeaseAcquisitionNumber() + 1)
+                    .build();
+            update(lockId, newLockData, timestamp);
+            log.info("Lock: {} Client:{} force-acquired lock.", lockId, clientId);
+        }
```
pankti-m, line 157: what if the lock is acquired by another server and must be revoked?

utils/src/main/java/org/corfudb/utils/lock/states/HasLeaseState.java
```diff
@@ -54,7 +52,8 @@ public HasLeaseState(Lock lock) {
      */
     @Override
     public Optional<LockState> processEvent(LockEvent event) throws IllegalTransitionException {
-        log.debug("Lock: {} lock event:{} in state:{}", lock.getLockId(), event, getType());
+        log.debug("Lock: {} lock event:{} in state:{} for id: {}", lock.getLockId(), event,
+                getType(), lockStore.getClientId());
```
pankti-m, line 56: dont we need a case for forceAcquired?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -708,4 +728,47 @@ public void shutdown() {
...
+                key -> new Lock(lockId, lockListener, lockClient.getClientContext(), lockConfig,
+                        initState));
+
+        return lockClient;
+    }
+
+
+    @VisibleForTesting
+    public void deregisterToLogReplicationLock() throws Exception {
+        LockClient lockClient = createLockClient(LockStateType.HAS_LEASE);
```
pankti-m, line 758: this will create a new lock every time, right?  can we use the same lock keep releasing/acquiring it?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -163,7 +163,14 @@
...
         this.serverConfig = serverConfig;
         this.dataStore = new DataStore(serverConfig, this::dataStoreFileCleanup);
-        generateNodeId();
+
+        if (serverConfig.get("--node-id") != null) {
+            UUID serverId = UuidUtils.fromBase64(getServerConfig(String.class, "--node-id"));
+            generateNodeId(serverId);
+        }
+        else {
+            generateNodeId();
```
xiaoqin2012, line 172: Can we put the if condition to generateNodeId() function?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -180,9 +187,36 @@ public void run() {
         } catch (Exception e) {
             log.error("Unhandled exception caught during log replication service discovery. Retry,", e);
         } finally {
-            if (runtime != null) {
-                runtime.shutdown();
-            }
+            runtime.cleanup(CorfuRuntime::shutdown);
```
xiaoqin2012, line 190: Is it the only place that clean up the runtime?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -180,9 +187,36 @@ public void run() {
...
+        }
+    }
+
+    private LockConfig createLockConfig(String pluginFilePath) {
```
xiaoqin2012, line 194: It is cleaner to have a lock config class.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -447,6 +477,7 @@ private void fetchTopologyFromClusterManager() {
      */
     private void stopLogReplication() {
         if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && isLeader.get()) {
+            isLeader.set(false);
```
xiaoqin2012, line 480: stopLogReplication doesn't mean leader loss, right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -708,4 +728,47 @@ public void shutdown() {
     private String getLocalHost() {
         return NodeLocator.parseString(serverContext.getLocalEndpoint()).getHost();
     }
+
+    private LockClient createLockClient(LockStateType initState) throws Exception {
```
xiaoqin2012, line 732: May define a LockConfig or LockUtility class and move this function there.

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
xiaoqin2012: It has the same logic as registerInterest.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuInterClusterReplicationServer.java
```diff
@@ -389,7 +391,19 @@ private CorfuReplicationClusterManagerAdapter buildClusterManagerAdapter(String
 
         try (URLClassLoader child = new URLClassLoader(new URL[]{jar.toURI().toURL()}, this.getClass().getClassLoader())) {
             Class adapter = Class.forName(config.getTopologyManagerAdapterName(), true, child);
-            return (CorfuReplicationClusterManagerAdapter) adapter.getDeclaredConstructor().newInstance();
+            if (config.getTopologyConfigPath().isPresent()) {
```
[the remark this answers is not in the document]
PavelZaytsev: Right now topology config path is only used for testing in order to create a parameterized constructor for the UnitTestClusterManager. It basically just accepts a static topology configuration.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -180,9 +187,36 @@ public void run() {
...
         } finally {
-            if (runtime != null) {
-                runtime.shutdown();
-            }
+            runtime.cleanup(CorfuRuntime::shutdown);
+        }
+    }
+
+    private LockConfig createLockConfig(String pluginFilePath) {
+        try (InputStream input = new FileInputStream(pluginFilePath)) {
```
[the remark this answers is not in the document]
PavelZaytsev: I'll move the lock configs to the topology

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/UnitTestClusterManager.java
```diff
@@ -0,0 +1,129 @@
...
+    private class SiteConfigKeys {
+        @NonNull
+        private final String rolePrefix;
+
+        @NonNull
+        private final ClusterRole role;
+
+        private String clusterRoleToRolePrefix(ClusterRole role) {
+            if (role == ACTIVE) {
+                return "primary";
```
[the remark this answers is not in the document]
PavelZaytsev: This retrieves a prefix based on the role to access the keys in the config files. For example, if the role is ACTIVE the prefix is "primary", so if we go into corfu_replication_config_two_primaries.properties we will be getting the following keys:

primary_site
primary_site_corfu_portnumber
primary_site_portnumber0
primary_site_portnumber1
primary_site_node0
primary_site_node1
primary_nodes

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/plugins/UnitTestClusterManager.java
```diff
@@ -0,0 +1,129 @@
...
+        private String getIpAddressKey() {
+            return rolePrefix + "_site_node";
+        }
+
+        private String getPortKey() {
+            return rolePrefix + "_site_portnumber";
+        }
+    }
+
+    private static final String STANDBY_SITE_KEY = "standby_site";
```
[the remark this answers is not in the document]
PavelZaytsev: Was not used, so I removed it

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
```diff
@@ -217,7 +217,7 @@ public LogReplicationEntry resend() {
             log.warn("Caught a timeout exception ", e);
             force = true;
         } catch (Exception e) {
-            log.warn("Caught an Exception and will notify discovery service ", e);
+            log.trace("Caught an Exception and will notify discovery service ", e);
```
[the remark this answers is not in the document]
PavelZaytsev: This method is not related to my changes, I just changed the warning to trace because the logging was way too noisy.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/ReplicatingState.java
```diff
@@ -89,9 +95,7 @@ public void onEntry(LogReplicationRuntimeState from) {
 
     @Override
     public void onExit(LogReplicationRuntimeState to) {
-        if (to.getType().equals(LogReplicationRuntimeStateType.STOPPED)) {
```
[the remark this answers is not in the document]
PavelZaytsev: onEntry StoppedState is already doing the same thing, so we were doing the same thing twice.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/VerifyingRemoteLeaderState.java
```diff
@@ -55,6 +55,9 @@ public LogReplicationRuntimeState processEvent(LogReplicationRuntimeEvent event)
                     return fsm.getStates().get(LogReplicationRuntimeStateType.WAITING_FOR_CONNECTIVITY);
                 }
                 return this;
+            case REMOTE_LEADER_LOSS:
```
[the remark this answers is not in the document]
PavelZaytsev: Yes, we are receiving these events from the LogReplicationClientRouter every time a Remote responds with LOG_REPLICATION_LEADERSHIP_LOSS on our every message.

test/src/test/java/org/corfudb/integration/CorfuReplicationLeaderIT.java
```diff
@@ -0,0 +1,871 @@
...
+    private static final String PRIMARY_CORFU_ENDPOINT = DEFAULT_HOST + ":" + PRIMARY_CORFU_PORT;
+    private static final String STANDBY_CORFU_ENDPOINT = DEFAULT_HOST + ":" + STANDBY_CORFU_PORT;
+
+    private static final ImmutableList<Integer> TWO_PRIMARY_REPLICATION_SERVER_PORTS =
+            ImmutableList.of(9010, 9011);
+
+    private static final ImmutableList<Integer> ONE_STANDBY_REPLICATION_SERVER_PORT =
+            ImmutableList.of(9020);
+
+    private static final ImmutableList<Integer> ONE_PRIMARY_REPLICATION_SERVER_PORT =
```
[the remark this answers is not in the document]
PavelZaytsev: It is used for testing. In one scenario we lose the leadership on the primary, so we need two primaries and 1 standby. In another case we lose the leadership on the remote, so we need two standbys and one primary. Just makes it working with them easier.

test/src/test/java/org/corfudb/integration/CorfuReplicationLeaderIT.java
```diff
@@ -0,0 +1,871 @@
...
+                }
+                Sleep.sleepUninterruptibly(mainThreadSleepDuration);
+            }
+
+            while (replicationTableStandby.size() < numWriteIterations) {
+                log.info("Standby table size: {}", replicationTableStandby.size());
+                Sleep.sleepUninterruptibly(mainThreadSleepDuration);
+            }
+
+            backGroundWriter.join();
```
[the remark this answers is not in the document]
PavelZaytsev: The test will timeout if all the data is not written to the standby. The condition verifies it:
while (replicationTableStandby.size() < numWriteIterations) {
...

test/src/test/java/org/corfudb/integration/CorfuReplicationLeaderIT.java
```diff
@@ -0,0 +1,871 @@
...
+                                .stream()
+                                .filter(tuple -> !convertUUID(tuple.first).equals(convertUUID(server.first)))
+                                .findFirst();
+                        if (second.isPresent()) {
+                            Tuple<UUID, CorfuInterClusterReplicationServer> server2 = second.get();
+                            final int waitTime = 3;
+                            log.info("Waiting a few seconds before force acquire.");
+                            Sleep.sleepUninterruptibly(Duration.ofSeconds(waitTime));
+                            server2.second.getReplicationDiscoveryService()
+                                    .forceAcquireLogReplicationLock();
```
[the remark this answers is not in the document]
PavelZaytsev: If we don't forceAcquire, there is no determinism, so a lot of things can happen. All depends on how and when the threads in all the executors will be scheduled. And we have a lot of them in the distributed lock implementation.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: With how the locking is designed, it gets way too complex if we reuse the instance. It is much easier to disable the locking all together, put it in some initial state and then invoke the event that forces the transition. It has a @VisibleForTesting annotation, so this method will not be used in prod anyways.

utils/src/main/java/org/corfudb/utils/lock/persistence/LockStore.java
PavelZaytsev: That's is why the locking timeouts are configured via LockConfig. And we delay all the timeouts so this does not happen.

utils/src/main/java/org/corfudb/utils/lock/LockConfig.java
PavelZaytsev: Property file can be used to supply the needed configuration - for example for testing, or if we want to tune the timeouts. This class has a @Builder, as well as @Default annotation for the fields, in case the properties file is not provided and we don't want to configure anything. Basically we used to have these fields sprinkled around the code as static and now they are available as defaults.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: I thought about it too but it made moving it difficult because some of the things to create a lock are located in this class - for example, Discovery service is being passed here, this will create circular dependencies. I think since this is for testing, its fine to have it there for now. Otherwise, it will entail longer refactoring.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: We should ideally call it in processLockRelease but there is an if condition in this method: 
`localClusterDescriptor.getRole() == ClusterRole.ACTIVE && isLeader.get()`
that will not get executed if we do it there.

PavelZaytsev: It used to be just: `if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE`. Not sure why it was added later.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: I had to modify this method, so it depends on LogReplicationPluginConfig, which is in infrastructure. This will create a circular dependency between the two. I think it's ok to have it here for now because the LockConfig from a file is actually being created only by the CorfuReplicationDiscvoeryService.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: Yes we cleanup runtime only once. The reason why it is wrapped in the SingletonResource is because we create the new lock clients for testing that also use this runtime.

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
PavelZaytsev: I added validations

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
PavelZaytsev: @pankti-m that's how the register/resume interest in lock works. Basically we call the revoke lease in that case to resume the interest. I added validation.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
xiaoqin2012: There are two options to do:
1. Define a new class extends from DiscoveryServiceClass , the new class is used for testing mostly.
2. Pass the instance of discovery service to the LockConfig class. 
We can do it later. Just this class file becomes so big now. It is hard for people to read in future.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -447,6 +491,7 @@ private void fetchTopologyFromClusterManager() {
      */
     private void stopLogReplication() {
         if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && isLeader.get()) {
+            isLeader.set(false);
```
annym, line 494: I think it is there in case it is stopped from processEvent on an exception (doesn't really imply we are leaders) or if called from processTopologyChangeNotification which also does not assume we're leaders.
You could probably set it right when it comes back from stopLogReplication in the processLockRelease, right? ... Cause imagine we get a processTopologyChangeNotification, in this case maybe we stop just because we are no longer active, but that does not mean we lost leadership.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -708,4 +742,46 @@ public void shutdown() {
...
+
+
+    @VisibleForTesting
+    public void deregisterToLogReplicationLock() throws Exception {
+        LockClient lockClient = createLockClient(LockStateType.HAS_LEASE);
+        lockClient.deregisterInterest();
+    }
+
+    @VisibleForTesting
+    public void forceAcquireLogReplicationLock() throws Exception {
```
annym, line 776: Pavel I'm not sure I understand correctly. How does this force to acquire the lock? let's say this method is called on node A. But node B has the lock, in Corfu Store node B is still marked as the lock owner. How does this enforce? or are you saying that you are faking the state machine to believe it has it. But in practice in the corfu store it does not?
If that is the case how do we control in our tests that two nodes don't have the lock at the same time?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -708,4 +742,46 @@ public void shutdown() {
...
+                        initState));
+
+        return lockClient;
+    }
+
+
+    @VisibleForTesting
+    public void deregisterToLogReplicationLock() throws Exception {
+        LockClient lockClient = createLockClient(LockStateType.HAS_LEASE);
+        lockClient.deregisterInterest();
```
annym, line 772: What happens if you deregisterInterest which enforces the LEASE_EXPIRED event which somehow will cause a callback on our listener, but the real lock (the previous to this) acquires the lock at that moment (which will also cause a callback on our DiscoveryService).

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
annym, line 220: I'm curious why is this noisy? I mean why do you see exceptions so often? is it just because you're exercising lock changes and it should fail or some underlying bug we should look into?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/VerifyingRemoteLeaderState.java
annym, line 58: but since in this state you don't have a remote leader, you can just ignore it, right?... It would throw an IllegalTransitionException which just ends up ignoring the event here. BTW we might want to change that logging level of the IllegalTransitionException to something lower as it is not really critical like we discussed the other day.

test/src/test/resources/logback-test.xml
```diff
@@ -43,9 +43,9 @@
         <!--<appender-ref ref="MetricsRollingFile" />-->
     </logger>
 
-    <!--<root level="DEBUG">-->
-        <!--&lt;!&ndash;<appender-ref ref="FILE" />&ndash;&gt;-->
-        <!--<appender-ref ref="STDOUT" />-->
-        <!--&lt;!&ndash;<appender-ref ref="MetricsRollingFile" />&ndash;&gt;-->
-    <!--</root>-->
+    <root level="DEBUG">
```
annym, line 46: Please comment this back

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: I should have documented this somewhere but basically it's just one part for testing. There are multiple issues with testing a leadership change. There are multiple different thread pools in the distributed lock implementation that schedule the periodic tasks all the time, no APIs to control the deregistration/registration to the lock service when needed, fixed timeouts on all the tasks that produce the events for the FSM, and the strong coupling on the discovery service. All of this makes it testing really difficult. I could not find a better way to test the leadership change without rewriting the distributed lock significantly, so I came up with the following, in order to force the leadership change in tests and make them deterministic:

1) Increased these thread pools' timeouts via providing a lock configuration. These will essentially allow deferring the renewals for the lock holder indefinitely as well as postponing the revoking calls from the competitors indefinitely.  
2) Created an API called deregisterToLogReplicationLock - this should be called on the current lock holder. It will force it to release the lock, without renewals (since we have increased these timeouts), and not allow everyone else to get the lock immediately (since we also have increased these timeouts).
3) Created an API called forceAcquireLogReplicationLock - this should be called on some other LR once the previous one has deregistered the interest. 

Doing these three things will guarantee that tests execute exactly the same way all the time without requiring to do a lot of modifications in the actual distributed lock code and breaking something.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: This is why we increase the timeouts via the configuration in tests to prevent some other tasks from acquiring the lock.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
PavelZaytsev: It's noisy on the lock changes when the timeouts start happening. Did not look like a bug to me.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -180,9 +188,49 @@ public void run() {
...
-                runtime.shutdown();
-            }
+            runtime.cleanup(CorfuRuntime::shutdown);
+        }
+    }
+
+    private LockConfig createLockConfig(String pluginFilePath) {
+        LogReplicationPluginConfig logReplicationPluginConfig =
+                new LogReplicationPluginConfig(pluginFilePath);
+        if (!logReplicationPluginConfig.getTopologyConfigPath().isPresent()) {
```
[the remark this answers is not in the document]
PavelZaytsev: Fixed as discussed

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -180,9 +188,49 @@ public void run() {
...
         } finally {
-            if (runtime != null) {
-                runtime.shutdown();
-            }
+            runtime.cleanup(CorfuRuntime::shutdown);
+        }
+    }
+
+    private LockConfig createLockConfig(String pluginFilePath) {
+        LogReplicationPluginConfig logReplicationPluginConfig =
```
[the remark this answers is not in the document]
PavelZaytsev: Now its done from a separate config file

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
PavelZaytsev: Ok makes sense

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/fsm/VerifyingRemoteLeaderState.java
PavelZaytsev: The issue was during the Replicating State. We assume that we can only move to the VerifyingRemoteLeader only if the connection is down. But this is not the only scenario when it can get lost. It can also get lost if it loses leadership on the remote. In that case, it actually explicitly notifies the active about it every time actively tries to send a message.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuInterClusterReplicationServer.java
```diff
@@ -47,6 +48,7 @@
                     + "[--snapshot-batch=<batch-size>]"
                     + "[--lock-lease=<lease-duration>]"
                     + "[-c <ratio>] [-d <level>] [-p <seconds>] "
+                    + "[--node-id=<nodeId>]"
```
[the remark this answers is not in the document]
PavelZaytsev: Removed and replaced a config with a file for other ITs

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuInterClusterReplicationServer.java
```diff
@@ -46,8 +47,9 @@
                     + "\tlog_replication_server (-l <path>|-m) [-nsN] [-a <address>|-q <interface-name>] "
                     + "[--max-num-snapshot-msg-per-batch=<batch-size>] "
                     + "[--max-data-message-size=<msg-size>] "
-                    + "[--lock-lease=<lease-duration>]"
                     + "[-c <ratio>] [-d <level>] [-p <seconds>] "
+                    + "[--node-id=<nodeId>]"
+                    + "[--log-replication-config=<log-replication-config-file-path>]"
```
annym, line 52: Whatever is added on the CorfuInterClusterReplicationServer we need to add in CorfuServer (as optional) because we start CorfuReplicationServer through the CorfuServer script (it's a license limitation that we have)

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -57,7 +57,7 @@
                     + "[-k <seqcache>] [-T <threads>] [-B <size>] [-i <channel-implementation>] "
                     + "[-H <seconds>] [-I <cluster-id>] [-x <ciphers>] [-z <tls-protocols>]] "
                     + "[--metrics] [--metrics-port <metrics_port>]"
-                    + "[--snapshot-batch=<batch-size>] [--lock-lease=<lease-duration>]"
+                    + "[--snapshot-batch=<batch-size>] [--log-replication-config=<log-replication-config-file-path>]"
```
annym, line 60: This is intended to replace the --plugin as well, right? so we can remove the plugin...

PavelZaytsev: I think it's better to have them separately right? Plugin is designed to pass in the jar plugin info. Config is designed for configuring different parameters for log replication.

annym: Ok, well I had understood in the meeting that Sundar suggested to keep them in one file so we have less files. I think it's ok, we only need to keep this in mind then so @hisundar makes the proper changes to include this in the init script when this gets merged and we include this new file in the appliance as well (either way we need to include this new file)

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -329,9 +327,15 @@ public CorfuRuntimeParameters getManagementRuntimeParameters() {
      * Generate a Node Id if not present.
      */
     private void generateNodeId() {
+        UUID serverId;
+        if (serverConfig.get("--node-id") != null) {
```
xnull, line 331: `if(serverConfig.containsKey("--node-id"))...`

PR #2649, opened 2020-07-22
Server-Side Infrastructure for New BaseServer RPCs

zfrenette:
## Overview

Description: Protobuf RPCs for BaseServer.

Why should this be merged: Meant to replace RequestHandler, and allow for the development of other RPC types.

- [X] Public API has Javadoc


common/src/main/java/org/corfudb/common/protocol/client/ResponseHandler.java
```diff
@@ -20,6 +20,7 @@
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
         ByteBuf msgBuf = (ByteBuf) msg;
+        msgBuf.readByte(); // Temporary -- Consume 0x2 marker.
```
Maithem, line 23: Read and check the value instead of assuming its 0x2

common/src/main/proto/CorfuProtocol.proto
```diff
@@ -87,20 +84,23 @@ message PingRequest {
...
-
+    UUID clientId = 1;
+    UUID serverId = 2;
 }
 
+// TODO(Zach): Update handshake logic
 message AuthenticateResponse {
-
+    UUID serverId = 1;
+    string corfuVersion = 2;
```
Maithem, line 96: You can remove this and do the version check on every message instead. The message version can be kept inside the header.

infrastructure/src/main/java/org/corfudb/infrastructure/protocol/NettyServerRouter.java
```diff
@@ -0,0 +1,144 @@
...
+        log.info("channelActive: Incoming connection established from: {}.",
+                ctx.channel().remoteAddress());
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        ByteBuf msgBuf = (ByteBuf) msg;
+
+        // Temporary -- Check Corfu msg marker: 0x1 indicates legacy while 0x2 indicates new
+        if (msgBuf.getByte(msgBuf.readerIndex()) == 0x1) {
```
Maithem, line 100: Can you make these static constants (i.e. 0x1 and 0x2)

Maithem:
looks good, we need to think of a clean way to get rid of the annotations, or at least simplify the code. Code flow is a little hard to follow.

chetangudisagar:
I was thinking about a slightly different approach of having a similar implementation of handlers like the `PeerClient `setup i.e. 

1. Merge the handlers into one class that are currently distributed across different classes (`BaseServer`, `SequencerServer`, `LogUnitServer`, etc.). The actual implementations can be either written there in the handlers or routed to different classes as the implementation can contain lengthy code. This gets rid of the annotation-based handler mechanism.
2. Have a channel handler similar to client one that handles functionalities of the client connections (~`NettyServerRouter`). 

@Maithem What are your views on this?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageDecoder.java
```diff
@@ -19,6 +20,13 @@
     @Override
     protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteBuf,
                           List<Object> list) throws Exception {
+        // Leave message unchanged if is non-legacy
+        if(byteBuf.getByte(byteBuf.readerIndex()) != API.LEGACY_CORFU_MSG_MARK) {
+            list.add(byteBuf);
+            return;
+        }
+
+        byteBuf.readByte();
```
Maithem, line 29: This is a bug. The marker is already consumed in line 24, if that if-statement isnt true, then another byte will be consumed, which is not the marker byte, so this will fail the subsequent call to serialization since it will be missing a byte.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/NettyCorfuMessageDecoder.java
```diff
@@ -19,6 +20,13 @@
     @Override
     protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteBuf,
                           List<Object> list) throws Exception {
+        // Leave message unchanged if is non-legacy
+        if(byteBuf.getByte(byteBuf.readerIndex()) != API.LEGACY_CORFU_MSG_MARK) {
```
Maithem, line 24: Never mind, ignore my comment below. It's wrong.

Unline the readByte, getByte API doesn't move the reader index.

infrastructure/src/main/java/org/corfudb/infrastructure/protocol/NettyServerRouter.java
```diff
@@ -96,14 +97,13 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
-            ctx.fireChannelRead(msgBuf); // Forward legacy corfu msg to next handler
+        // Temporary -- If message is not a new Protobuf message, forward the message.
+        if (msgBuf.getByte(msgBuf.readerIndex()) != API.PROTO_CORFU_MSG_MARK) {
+            ctx.fireChannelRead(msgBuf);
             return;
         }
 
-        msgBuf.readByte(); // Temporary -- Consume 0x2 marker
-
+        msgBuf.readByte();
```
Maithem, line 106: What if this byte is not LEGACY_CORFU_MSG_MARK ?

I think you should explicitly check that it is LEGACY_CORFU_MSG_MARK and throw an IllegalStateException otherwise.

zfrenette: Do you mean throw an exception (instead of passing it along the pipeline) if the message isn't marked with PROTO_CORFU_MSG_MARK?

infrastructure/src/main/java/org/corfudb/infrastructure/protocol/NettyServerRouter.java
Maithem: @zfrenette  Did you delete your comment?

I got an email notification but don't see it on the PR.

> Do you mean throw an exception (instead of passing it along the pipeline) if the message isn't marked with PROTO_CORFU_MSG_MARK?


Yes, I think its better to throw an exception instead of the propagate the error further.

infrastructure/src/main/java/org/corfudb/infrastructure/protocol/NettyServerRouter.java
```diff
@@ -60,6 +62,17 @@ public NettyServerRouter(ImmutableList<AbstractServer> servers, ServerContext se
...
+
+
+    @Override
+    public void sendResponse(Response response, ChannelHandlerContext ctx) {
+        ByteBuf outBuf = PooledByteBufAllocator.DEFAULT.buffer();
+        outBuf.writeByte(API.PROTO_CORFU_MSG_MARK);
+        outBuf.writeBytes(response.toByteArray());
+        //TODO(Zach): Where to remove allocation?
+        ctx.writeAndFlush(outBuf);
+    }
```
Maithem, line 74: Pass the allocator via the constructor. Accessing a resources from multiple entry points makes it hard to change. You can wrap the default allocator and pass that instance to the constructor.

infrastructure/src/main/java/org/corfudb/infrastructure/protocol/NettyServerRouter.java
```diff
@@ -0,0 +1,163 @@
...
+
+
+
+    @Override
+    public void sendResponse(Response response, ChannelHandlerContext ctx) {
+        ByteBuf outBuf = PooledByteBufAllocator.DEFAULT.buffer();
+        outBuf.writeByte(API.PROTO_CORFU_MSG_MARK);
+        outBuf.writeBytes(response.toByteArray());
+        //TODO(Zach): Where to remove allocation?
```
Maithem, line 72: You can use an input stream here and release it on IOException

infrastructure/src/main/java/org/corfudb/infrastructure/protocol/BaseServer.java
```diff
@@ -0,0 +1,84 @@
+package org.corfudb.infrastructure.protocol;
```
Maithem, line 1: Added the new annotated methods in the existing class. I don't see a strong reason to fork the class.

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -180,6 +213,24 @@ private void doReset(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
...
+     * which monitors the exit code of Corfu. If the exit code is 100, then it resets
+     * the server and DELETES ALL EXISTING DATA.
+     * @param req The incoming request message.
+     * @param ctx The channel context.
+     * @param r The server router.
+     */
+    @AnnotatedServerHandler(type = CorfuProtocol.MessageType.RESET)
+    private void handleReset(Request req, ChannelHandlerContext ctx, org.corfudb.infrastructure.protocol.IServerRouter r) {
+        log.warn("Remote reset requested from client {}", req.getHeader().getClientId());
+        //TODO(Zach): checkArgument(req.hasResetRequest());
```
Maithem, line 227: How can you even get routed to this point without having the ResetRequest? I think that check is done before a call is made to this method.

zfrenette: I believe only the MessageType is checked. There hasn't been a check that a proper payload is provided.

zfrenette: I think these checks can all be done in the handle method in RequestHandlerMethods.

common/src/main/java/org/corfudb/common/protocol/API.java
```diff
@@ -131,6 +137,71 @@ public static Response newPingResponse(Header header, ServerError error) {
...
+        ResetRequest resetRequest = ResetRequest.getDefaultInstance();
+        return Request.newBuilder()
+                .setHeader(header)
+                .setResetRequest(resetRequest)
+                .build();
+    }
+
+    public static Response newResetResponse(Header header) {
+        return newRestartResponse(header, newNoServerError());
+    }
```
chetangudisagar, line 171: I think you forgot to change this to -
`return newResetResponse(header, newNoServerError());`

zfrenette: Thanks for pointing this out

infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
zfrenette: I decided to perform these checks in NettyServerRouter since it was originally performed in RequestHandler.

common/src/main/java/org/corfudb/common/protocol/API.java
```diff
@@ -47,4 +124,163 @@ public static Request newPingRequest(Header header) {
...
+    public static Response newErrorResponseNoPayload(Header header, ServerError error) {
+        return Response.newBuilder()
+                .setHeader(header)
+                .setError(error)
+                .build();
+    }
+
+    public static boolean validateRequest(Request request) {
+        switch(request.getHeader().getType()) {
+            case PING:
```
Maithem, line 214: This seems superfluous, just let the message go through and if not recognized the server should reply with an error.

Maithem:
We need to chunk work into smaller, more focused PR.

Maithem:
Ops, clicked closed instead of comment.

common/src/main/java/org/corfudb/common/protocol/API.java
zfrenette: Can you comment a bit more on why this is superfluous? Currently, ResponseHandler/PeerClient performs this check as well using a similar switch statement. Also, I did have this check (as a checkArgument call) in the handle method for each message type (say handlePing when receiving a ping message), but you had mentioned that this was checked prior.
Although no error is currently sent, I do intend to send an error (see NettyServerRouter) if this checks fails.

common/src/main/proto/CorfuProtocol.proto
zfrenette: I'll incorporate this in a separate PR later on.

PR #2650, opened 2020-07-22
Fix NullPointerException in StreamView.

WenbinZhu:
## Overview

When read option is set to ignoreTrim and non cacheable, removeFromQueue
will throw NPE when encountering TrimmedException and trying to get the
trimmed addresses list from TrimmedException.

Why should this be merged:  Fix NPE

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
+++ b/runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
@@ -3,6 +3,7 @@ package org.corfudb.runtime.exceptions;
 import lombok.Getter;
 import lombok.Setter;
 
+import java.util.Collections;
 import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
```diff
@@ -24,7 +25,7 @@
      */
     @Getter
     @Setter
-    private List<Long> trimmedAddresses;
+    private List<Long> trimmedAddresses = Collections.emptyList();
```
Maithem, line 28: I think its cleaner to make this field final and assign the empty list in the constructors that don't need it.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -727,7 +727,7 @@ private boolean isLogDataValid(long address, ILogData logData, boolean throwExce
 
         if (logData.isTrimmed()) {
             if (throwException) {
-                throw new TrimmedException(String.format("Trimmed address %s", address));
+                throw new TrimmedException(Collections.singletonList(address));
```
Maithem, line 730: Cleaner to push down to the constructor.

xnull:
Would be great to fix also other NPE-s
image

runtime/src/main/java/org/corfudb/runtime/exceptions/TrimmedException.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java

WenbinZhu:
> Would be great to fix also other NPE-s
> image

This is supposed to be a quick patch to quickly unblock QE testing.
Besides many of these warnings do not mean there is an actual NPE, a lot of times we have checks to make sure the objects are not null, but IDE static analysis does not know about it and still give a warning.

PR #2661, opened 2020-07-27
File descriptor leaks corfu 0.3.0.2

xnull:
## Overview

Description:
File descriptor leaks fixes

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/checks.xml
+++ b/checks.xml
@@ -235,11 +235,6 @@
         </module>
         <module name="OverloadMethodsDeclarationOrder"/>
         <module name="VariableDeclarationUsageDistance"/>
-        <module name="CustomImportOrder">
-            <property name="sortImportsInGroupAlphabetically" value="true"/>
-            <property name="separateLineBetweenGroups" value="true"/>
-            <property name="customImportOrderRules" value="STATIC###THIRD_PARTY_PACKAGE"/>
-        </module>
         <module name="MethodParamPad"/>
         <module name="NoWhitespaceBefore">
             <property name="tokens"
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -16,13 +16,19 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
 import lombok.Builder;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.infrastructure.ServerThreadFactory.ExceptionHandler;
 import org.corfudb.infrastructure.log.InMemoryStreamLog;
 import org.corfudb.infrastructure.log.StreamLog;
 import org.corfudb.infrastructure.log.StreamLogCompaction;
@@ -102,6 +108,8 @@ public class LogUnitServer extends AbstractServer {
 
     private ExecutorService executor;
 
+    private final ReadWriteLock resetLock = new ReentrantReadWriteLock();
+
     /**
      * Returns a new LogUnitServer.
      *
@@ -110,8 +118,10 @@ public class LogUnitServer extends AbstractServer {
     public LogUnitServer(ServerContext serverContext) {
         this.serverContext = serverContext;
         this.config = LogUnitServerConfig.parse(serverContext.getServerConfig());
-        executor = Executors.newFixedThreadPool(serverContext.getLogunitThreadCount(),
-                new ServerThreadFactory("LogUnit-", new ServerThreadFactory.ExceptionHandler()));
+        executor = Executors.newFixedThreadPool(
+                serverContext.getLogunitThreadCount(),
+                new ServerThreadFactory("LogUnit-", new ExceptionHandler())
+        );
 
         if (config.isMemoryMode()) {
             log.warn("Log unit opened in-memory mode (Maximum size={}). "
@@ -128,13 +138,40 @@ public class LogUnitServer extends AbstractServer {
         dataCache = new LogUnitServerCache(config, streamLog);
         batchWriter = new BatchProcessor(streamLog, serverContext.getServerEpoch(), !config.isNoSync());
 
-        logCleaner = new StreamLogCompaction(streamLog, 10, 45, TimeUnit.MINUTES, ServerContext.SHUTDOWN_TIMER);
+        logCleaner = new StreamLogCompaction(
+                streamLog, resetLock, 10, 45, TimeUnit.MINUTES, ServerContext.SHUTDOWN_TIMER
+        );
     }
 
 
     @Override
     protected void processRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
-        executor.submit(() -> getHandler().handle(msg, ctx, r));
+        executor.submit(
+                () -> {
+                    switch (msg.getMsgType()) {
+                        case READ_REQUEST:
+                        case KNOWN_ADDRESS_REQUEST:
+                        case COMPACT_REQUEST:
+                            Lock lock = resetLock.readLock();
+                            lock.lock();
+
+                            if (msg.getEpoch() != serverContext.getLogUnitEpochWaterMark()) {
+                                r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH,
+                                        serverContext.getServerEpoch()));
+                                return;
+                            }
+
+                            try {
+                                getHandler().handle(msg, ctx, r);
+                            } finally {
+                                lock.unlock();
+                            }
+
+                            break;
+                        default:
+                            getHandler().handle(msg, ctx, r);
+                    }
+                });
     }
 
     /**
@@ -144,7 +181,9 @@ public class LogUnitServer extends AbstractServer {
     public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
         log.debug("handleTailRequest: received a tail request {}", msg);
         batchWriter.<TailsResponse>addTask(TAILS_QUERY, msg)
-                .thenAccept(tailsResp -> r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp)))
+                .thenAccept(tailsResp -> {
+                    r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp));
+                })
                 .exceptionally(ex -> {
                     handleException(ex, ctx, msg, r);
                     return null;
@@ -162,7 +201,8 @@ public class LogUnitServer extends AbstractServer {
         log.trace("handleLogAddressSpaceRequest: received a log address space request {}", msg);
         batchWriter.<StreamsAddressResponse>addTask(LOG_ADDRESS_SPACE_QUERY, payloadMsg)
                 .thenAccept(tailsResp -> r.sendResponse(ctx, msg,
-                        CorfuMsgType.LOG_ADDRESS_SPACE_RESPONSE.payloadMsg(tailsResp)))
+                        CorfuMsgType.LOG_ADDRESS_SPACE_RESPONSE.payloadMsg(tailsResp))
+                )
                 .exceptionally(ex -> {
                     handleException(ex, ctx, payloadMsg, r);
                     return null;
@@ -219,14 +259,16 @@ public class LogUnitServer extends AbstractServer {
             msg.setPriorityLevel(PriorityLevel.HIGH);
         }
 
-        batchWriter.addTask(WRITE, msg)
+        batchWriter
+                .addTask(WRITE, msg)
                 .thenRunAsync(() -> {
                     dataCache.put(msg.getPayload().getGlobalAddress(), logData);
                     r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg());
-                }, executor).exceptionally(ex -> {
-            handleException(ex, ctx, msg, r);
-            return null;
-        });
+                }, executor)
+                .exceptionally(ex -> {
+                    handleException(ex, ctx, msg, r);
+                    return null;
+                });
     }
 
     /**
@@ -239,7 +281,8 @@ public class LogUnitServer extends AbstractServer {
         log.debug("rangeWrite: Writing {} entries [{}-{}]", range.size(),
                 range.get(0).getGlobalAddress(), range.get(range.size() - 1).getGlobalAddress());
 
-        batchWriter.addTask(RANGE_WRITE, msg)
+        batchWriter
+                .addTask(RANGE_WRITE, msg)
                 .thenRun(() -> r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg()))
                 .exceptionally(ex -> {
                     handleException(ex, ctx, msg, r);
@@ -258,7 +301,8 @@ public class LogUnitServer extends AbstractServer {
     private void prefixTrim(CorfuPayloadMsg<TrimRequest> msg, ChannelHandlerContext ctx,
                             IServerRouter r) {
         log.debug("prefixTrim: trimming prefix to {}", msg.getPayload().getAddress());
-        batchWriter.addTask(PREFIX_TRIM, msg)
+        batchWriter
+                .addTask(PREFIX_TRIM, msg)
                 .thenRun(() -> r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg()))
                 .exceptionally(ex -> {
                     handleException(ex, ctx, msg, r);
@@ -321,8 +365,16 @@ public class LogUnitServer extends AbstractServer {
     @ServerHandler(type = CorfuMsgType.COMPACT_REQUEST)
     private void handleCompactRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
         log.debug("handleCompactRequest: received a compact request {}", msg);
-        streamLog.compact();
-        r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+
+        Lock lock = resetLock.writeLock();
+        lock.lock();
+
+        try {
+            streamLog.compact();
+            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+        } finally {
+            lock.unlock();
+        }
     }
 
     @ServerHandler(type = CorfuMsgType.FLUSH_CACHE)
@@ -374,27 +426,39 @@ public class LogUnitServer extends AbstractServer {
     private synchronized void resetLogUnit(CorfuPayloadMsg<Long> msg,
                                            ChannelHandlerContext ctx, IServerRouter r) {
 
+        if (msg.getPayload() <= serverContext.getLogUnitEpochWaterMark()
+                || msg.getPayload() != serverContext.getServerEpoch()) {
+            log.info("LogUnit Server Reset request received but reset already done.");
+            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+            return;
+        }
+
         // Check if the reset request is with an epoch greater than the last reset epoch seen to
-        // prevent multiple reset in the same epoch. and should be equal to the current router
+        // prevent multiple reset in the same epoch and should be equal to the current router
         // epoch to prevent stale reset requests from wiping out the data.
-        if (msg.getPayload() > serverContext.getLogUnitEpochWaterMark()
-                && msg.getPayload() == serverContext.getServerEpoch()) {
-            serverContext.setLogUnitEpochWaterMark(msg.getPayload());
-            batchWriter.addTask(RESET, msg)
+        serverContext.setLogUnitEpochWaterMark(msg.getPayload());
+
+        Lock lock = resetLock.writeLock();
+        lock.lock();
+
+        try {
+            CompletableFuture<Void> cf = batchWriter
+                    .addTask(RESET, msg)
                     .thenRun(() -> {
                         dataCache.invalidateAll();
                         log.info("LogUnit Server Reset.");
                         r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
-                    }).exceptionally(ex -> {
-                handleException(ex, ctx, msg, r);
-                return null;
-            });
-        } else {
-            log.info("LogUnit Server Reset request received but reset already done.");
-            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+                    })
+                    .exceptionally(ex -> {
+                        handleException(ex, ctx, msg, r);
+                        return null;
+                    });
+            //this has to be a synchronous call for the read write lock to take effect
+            cf.join();
+        } finally {
+            lock.unlock();
         }
     }
-
 
     /**
      * Shutdown the server.
@@ -406,6 +470,7 @@ public class LogUnitServer extends AbstractServer {
         executor.shutdown();
         logCleaner.shutdown();
         batchWriter.close();
+        streamLog.close();
     }
 
     @VisibleForTesting
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -1,30 +1,7 @@
 package org.corfudb.infrastructure;
 
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.LOG_ADDRESS_SPACE_QUERY;
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.PREFIX_TRIM;
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.RANGE_WRITE;
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.RESET;
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.SEAL;
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.TAILS_QUERY;
-import static org.corfudb.infrastructure.BatchWriterOperation.Type.WRITE;
-
-
 import com.google.common.annotations.VisibleForTesting;
 import io.netty.channel.ChannelHandlerContext;
-import java.lang.invoke.MethodHandles;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
-
 import lombok.Builder;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
@@ -59,6 +36,25 @@ import org.corfudb.runtime.exceptions.WrongEpochException;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.stream.StreamAddressSpace;
 import org.corfudb.util.Utils;
+
+import java.lang.invoke.MethodHandles;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.LOG_ADDRESS_SPACE_QUERY;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.PREFIX_TRIM;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.RANGE_WRITE;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.RESET;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.SEAL;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.TAILS_QUERY;
+import static org.corfudb.infrastructure.BatchWriterOperation.Type.WRITE;
 
 
 /**
@@ -108,8 +104,6 @@ public class LogUnitServer extends AbstractServer {
 
     private ExecutorService executor;
 
-    private final ReadWriteLock resetLock = new ReentrantReadWriteLock();
-
     /**
      * Returns a new LogUnitServer.
      *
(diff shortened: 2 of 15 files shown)


xnull:
> getSegmentHandleForAddress

Consider the following scenarios:
 - StreamLogCompaction is triggered by the scheduler and executing a cleanup, another thread reads the data at the same time.
 - what if one thread just gets a segment and is using a filechannel, and the second thread is executing the reset method?

xnull:
> I believe there's another problem that we need to fix.
> 
> When `RESET_LOGUNIT` is called readers will wait on the reset to happen and complete after the delete. This means that its possible that a read request from a previous epoch will be serviced after the delete and it would return an empty slot. In turn, this would trigger a whole filling process in a range that the fault detector expects to be empty (since it resets then starts state transfer).
> 
> I think it would be cleaner to move the locking + epoch check to `processRequest` since that is the entry point to all requests.
> 
> In essence, what we need is that all requests from previous epochs should fail with WrongEpochException after the reset happens.

Lock moved to LogUnitServer.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -272,40 +293,44 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
...
 
-        ReadResponse rr = new ReadResponse();
-        try {
-            for (Long address : msg.getPayload().getAddresses()) {
-                ILogData logData = dataCache.get(address, cacheable);
-                if (logData == null) {
-                    rr.put(address, LogData.getEmpty(address));
-                } else {
-                    rr.put(address, (LogData) logData);
+        logUnitLock.acquireReadLockAndExecuteSync(() -> {
```
Maithem, line 318: Where is the epoch checks for reads?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -472,4 +507,57 @@ public static LogUnitServerConfig parse(Map<String, Object> opts) {
...
+        private final ReadWriteLock resetLock = new ReentrantReadWriteLock();
+
+        private void acquireReadLockAndExecuteSync(Runnable action) {
+
+            acquireReadLock()
+                    .thenApply(lock -> {
+                        action.run();
+                        return lock;
+                    })
+                    .whenComplete((lock, err) -> lock.unlock());
```
Maithem, line 521: I think this style is an overkill. It's kind of convoluted. If you still wan't to use this, then I suggest moving the code to the entry point, which is `processRequest`, it's more centralized that way.

Also, wouldn't this swallow the exception ? On master, at least it's logged with `uncaughtException`.

Another cleaner solution  to consider is to shutdown the LogUnit executor and execute the reset logic on the IO thread. After the reset logic completes you can install a new executor. That way in-flight operations will time out on transition. As opposed to be being evaluated with an older epoch.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: >Also, wouldn't this swallow the exception ?
No, because `exceptionally` already catches. any exceptions in CF

anyways, I deleted LogUnitLock

changed

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -272,18 +290,29 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
...
+        lock.lock();
+
         try {
-            ILogData logData = dataCache.get(address, cacheable);
-            if (logData == null) {
-                rr.put(address, LogData.getEmpty(address));
-            } else {
-                rr.put(address, (LogData) logData);
+            if (msg.getEpoch() != serverContext.getServerEpoch()) {
+                throw new WrongEpochException(serverContext.getServerEpoch());
```
Maithem, line 298: Shouldn't this be based on serverContext.getLogUnitEpochWaterMark()?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -383,28 +438,39 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
-            });
-        } else {
-            log.info("LogUnit Server Reset request received but reset already done.");
-            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+                    })
+                    .exceptionally(ex -> {
+                        handleException(ex, ctx, msg, r);
+                        return null;
+                    });
+            cf.join();
```
Maithem, line 468: Can you put a comment to indicate that this has to be a synchronous call for the lock to take effect.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -292,19 +321,30 @@ public void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerCo
...
+                    ILogData logData = dataCache.get(address, cacheable);
+                    if (logData == null) {
+                        rr.put(address, LogData.getEmpty(address));
+                    } else {
+                        rr.put(address, (LogData) logData);
+                    }
+                }
+                r.sendResponse(ctx, msg, CorfuMsgType.READ_RESPONSE.payloadMsg(rr));
+            } catch (DataCorruptionException e) {
+                r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_CORRUPTION.msg());
```
Maithem, line 344: `ERROR_DATA_CORRUPTION.msg()` need to pass a long here. Please pull in #2612

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -292,19 +321,30 @@ public void multiRead(CorfuPayloadMsg<MultipleReadRequest> msg, ChannelHandlerCo
...
-            for (Long address : msg.getPayload().getAddresses()) {
-                ILogData logData = dataCache.get(address, cacheable);
-                if (logData == null) {
-                    rr.put(address, LogData.getEmpty(address));
-                } else {
-                    rr.put(address, (LogData) logData);
+            ReadResponse rr = new ReadResponse();
+            try {
+                if (msg.getEpoch() != serverContext.getServerEpoch()) {
+                    throw new WrongEpochException(serverContext.getServerEpoch());
```
Maithem, line 331: You have to send back a `WRONG_EPOCH` and not throw an exception. Otherwise, the exception wont be handled.

xnull:
> Looks good, just minor bugs.

done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: MULTIPLE_READ_REQUEST and KNOWN_ADDRESS_REQUEST has a bug, please rebase on #2612

xnull: #2612 should be merged into corfu-0.3.0.2 otherwise I can't merge into my PR directly

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: What's the motivation for that?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: So you didn't want to rebase because of the conflicts ? lel

Here's the rebased version #2674

xnull: Did I say that?
Why you don't want more #2612 into 0.3.0.2 branch?

Maithem: This is part of your patch, and you said you were done. It's reasonable to assume that you would do it, so I wasn't sure why you are waiting for me to cherry-pick it. Anyways, in the interest of time I did it on your behalf.

xnull: #2612 is another PR, I don't quite see why you want to mix it with this pr, why not just bring it to 0.3.0.2 separately, then I can get it from 0.3.0.2.
Anyways, let's merge #2674 and finish this conversation, don't you agree?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
Maithem: The motivation is "consistency". The batch processor will fail requests based on `LogUnitEpochWaterMark` and not `ServerEpoch`

xnull: fixed

xnull:
I have addressed all comments, please take a look

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -128,13 +138,40 @@ public LogUnitServer(ServerContext serverContext) {
...
+                () -> {
+                    switch (msg.getMsgType()) {
+                        case READ_REQUEST:
+                        case KNOWN_ADDRESS_REQUEST:
+                        case COMPACT_REQUEST:
+                            Lock lock = resetLock.readLock();
+                            lock.lock();
+
+                            if (msg.getEpoch() != serverContext.getLogUnitEpochWaterMark()) {
+                                r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH,
```
Maithem, line 159: Might be a perf issue, not sure thought.

xnull:
> ```java
>    protected void processRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
>     executor.submit(
>         () -> {
>           switch (msg.getMsgType()) {
>             case READ_REQUEST:
> ```

Unfortunately, the change above didn't  work https://travis-ci.org/github/CorfuDB/CorfuDB/builds/713821757

I rolled back to prev version https://travis-ci.org/github/CorfuDB/CorfuDB/builds/713834236

xnull:
> Slava, can you please provide an explanation on how the FD
@vjeko sure,
 - in the previous version, we watched only IO exceptions, but if a runtime exception is thrown we have a FileChannel open but don't have access to it, because we didn't add the file channel into a hashmap, because of exception. I have added additional `catch` clause https://github.com/CorfuDB/CorfuDB/blob/d360d82af3b0eee553e19a94d26c34edacc00311/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java#L780

 - We have to close streamLog during shutdown https://github.com/CorfuDB/CorfuDB/blob/d360d82af3b0eee553e19a94d26c34edacc00311/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java#L461

 - resetLock prevents us from reading data during logUnitServer.reset or StreamLogCompaction.compact operations. 
    The log files getting deleted (by streamLog.reset operation) and being read at the same time, which causes that deleted files are staying open https://github.com/CorfuDB/CorfuDB/blob/d360d82af3b0eee553e19a94d26c34edacc00311/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java#L110

[streamLog.reset](https://github.com/CorfuDB/CorfuDB/blob/d360d82af3b0eee553e19a94d26c34edacc00311/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java#L1271)

[streamLog.read](https://github.com/CorfuDB/CorfuDB/blob/d360d82af3b0eee553e19a94d26c34edacc00311/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java#L693)

We decided to use ReadWriteLock, when logUnit.reset is happening no readers can read data

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -102,6 +107,8 @@
 
     private ExecutorService executor;
 
+    private final ReadWriteLock resetLock = new ReentrantReadWriteLock();
```
zhangn49, line 110: Could we have some comments on this read-write lock? Like how it should be used in the future?
And looks like this PR is pretty different from #2616 , is this PR a subset?

xnull: comments - done.

This is a `small` fix for 0.3.0.2, yes it's different.

vjeko:
As far as I can tell, the only problem that can occur due to concurrent `read/write` while doing a `close` on the same `FD`, is that you might end up with unintended consequences (corrupt data) due to file descriptor re-use (which infinitesimal probability since the file descriptors are allocated sequentially). I do not understand how we can end up with dangling file descriptors.

Is there any sort of a test that demonstrates this issue?

xnull:
> Is there any sort of a test that demonstrates this issue?
Regarding baseServer.reset there is a universe test https://github.com/CorfuDB/CorfuDB/blob/550fc74a7b6e8494394601f75ce10ba946d0d9b1/it/src/test/java/org/corfudb/universe/scenario/FileDescriptorLeaksIT.java#L50
You can take a look at it.

Regarding exceptions in "getSegmentHandler" some unit tests fail on that issue and in the #2616 it was fixed along with most of the unit tests failed with FD leaks.

Regarding rwLock, I'm working on a test to reproduce that

checks.xml
```diff
@@ -235,11 +235,6 @@
         </module>
         <module name="OverloadMethodsDeclarationOrder"/>
         <module name="VariableDeclarationUsageDistance"/>
-        <module name="CustomImportOrder">
-            <property name="sortImportsInGroupAlphabetically" value="true"/>
-            <property name="separateLineBetweenGroups" value="true"/>
-            <property name="customImportOrderRules" value="STATIC###THIRD_PARTY_PACKAGE"/>
-        </module>
```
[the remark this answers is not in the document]
xnull: It generated a lot of messages that travis-ci couldn't finish the operation.

P.S. I also believe this check is completely useless and it floods all logs with this message, which makes tests analysis quite problematic.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -273,26 +291,34 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
             log.trace("read: {}, cacheable: {}", msg.getPayload().getAddresses(), cacheable);
         }
 
-        ReadResponse rr = new ReadResponse();
+        Lock lock = resetLock.readLock();
+        lock.lock();
```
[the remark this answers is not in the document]
xnull: We can't move it to StreamLog because we need the lock, in StreamLogCompaction class. 
Also, we need to get a lock before we start reading data and we need to check epoch, so we need to prevent a message getting into BatchProcessor. 
We have to use lock in LogUnitServer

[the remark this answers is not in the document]
WenbinZhu: You don't need a lock for StreamLogCompaction, the compactor only deletes files that are less than trim mark, that no readers can read, even you need the lock, you can create lock here, but call lock() inside it.
Reads don't go to BatchProcessor at all, so I don't know where is your epoch check...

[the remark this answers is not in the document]
xnull: >having a lock will be a waste in terms of performance.

can you please explain how it can be possible?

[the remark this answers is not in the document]
xnull: Also, I was thinking about your argument and come up with the statement, which sounds like this - the most important thing is: you probably don't want to read data from the cache if you actually don't have data in your database. 
It is a data inconsistency problem. That's why we need to have rwLock in LogUnitServer, when we reset LoUnit we have to clean cache, and only then we can read from the cache.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -273,26 +295,34 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
...
-            try {
-                ILogData logData = dataCache.get(address, cacheable);
-                if (logData == null) {
-                    rr.put(address, LogData.getEmpty(address));
-                } else {
-                    rr.put(address, (LogData) logData);
+        try {
+
+            ReadResponse rr = new ReadResponse();
+
```
Maithem, line 304: The epoch check for reads was dropped from the last patch. Please re-add it.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -303,26 +333,40 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
...
-            r.sendResponse(ctx, msg,
-                    CorfuMsgType.KNOWN_ADDRESS_RESPONSE.payloadMsg(knownAddresses));
-        } catch (DataCorruptionException dc) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_CORRUPTION
-                    .payloadMsg(Address.NON_ADDRESS));
-            return;
-        } catch (Exception e) {
-            handleException(e, ctx, msg, r);
+            KnownAddressRequest request = msg.getPayload();
+            try {
```
Maithem, line 341: Epoch check after lock.lock().

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -303,26 +333,40 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
...
 
     @ServerHandler(type = CorfuMsgType.COMPACT_REQUEST)
     private void handleCompactRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
         log.debug("handleCompactRequest: received a compact request {}", msg);
-        streamLog.compact();
-        r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+
+        Lock lock = resetLock.writeLock();
+        lock.lock();
+
```
Maithem, line 363: epoch check after lock.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -1275,7 +1281,7 @@ public void reset() {
         dataStore.resetStartingAddress();
         dataStore.resetTailSegment();
         logMetadata = new LogMetadata();
-        writeChannels.clear();
```
Maithem, line 1278: Maybe your test can't trigger the issue because of this. Why was this dropped?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: Forgot to move it from `processRequest`, my bad.
Added that check into processRequest for
```
 READ_REQUEST:
 KNOWN_ADDRESS_REQUEST:
 COMPACT_REQUEST:
```

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: I have changed `processRequest` could you please take a look

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: Could you please explain your question?
It was dropped because if you clean this you will have FD leaks

WenbinZhu: But if you have the locks protecting it, why adding this line could have the leaks? Does that mean the lock doesn't work at all?

Maithem: The leak was happening due to a synchronization problem between when `writeChannels.clear` and `writeChannels.computeIfAbsent` , but the RW lock is suppose to solve this synchronization issue. So, I don't understand why `writeChannels.clear();` has been removed. 

IF there are 4 segment handles [s1, s2, s3, s4] and the compaction mark is at 0, then a reset happens, then the same segment handles will be used to populate the state transfer data, which is incorrect. Essentially, your keeping the file channel handles in the map, but deleting the files underneath. What am I missing ?

WenbinZhu: btw closeSegmentHandlers() will also remove the handle from the wirteChannels

Maithem: Ah, I see.

WenbinZhu: But it still doesn't explain why removing this line could cause test to fail when lock is being added...

Maithem: I was asking @xnull , I didn't see the test failing without the `writeschannel.clear` call. I didn't try it myself.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -128,13 +141,34 @@ public LogUnitServer(ServerContext serverContext) {
...
+                        case KNOWN_ADDRESS_REQUEST:
+                        case COMPACT_REQUEST:
+                            if (msg.getEpoch() != serverContext.getServerEpoch()) {
+                                CorfuPayloadMsg<Long> payload = new CorfuPayloadMsg<>(
+                                        CorfuMsgType.WRONG_EPOCH,
+                                        serverContext.getServerEpoch()
+                                );
+                                r.sendResponse(ctx, msg, payload);
+                                return;
+                            }
```
zfrenette, line 165: Could you elaborate why this is checked here? From my understanding, a message type like READ_REQUEST or KNOWN_ADDRESS_REQUEST already have their epochs checked in the method epochIsValid from IServerRouter, since their ignoreEpoch fields are set as false. What's the difference between the getServerEpoch provided by the router versus the one provided by the context?

zfrenette: Oh I see, the epoch can become outdated as the request waits in the queue.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
xnull: @Maithem I can't say for sure, my guess is somehow new segments get added after we execute `closeSegmentHandlers` and those segments get cleaned afterward by `writeSegments.clean()`.
I reproduced FD leaks one more time right now when I added epoch checks, and unfortunately, it has leaks.
Cleaning writeChannels prevents us from leaks.

P.S. one small question - https://github.com/CorfuDB/CorfuDB/blob/00f40e2e700313ca422a7f03f8fc92da58e0b67b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java#L158

should it be check for `serverContext.getLogUnitEpochWaterMark()` ?

Maithem: Just to be consistent `getLogUnitEpochWaterMark` should be used, because that's what the BatchProcessor uses.

xnull: ok, thanks

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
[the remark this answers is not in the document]
Maithem: I think the concern here is that in the absence of reset calls the cost of acquiring a read lock will still be incurred despite the high probability of a cache hit.

@xnull to address this issue, the lock doesn't have to be moved to StreamLogFiles, it can be moved to `LogUnitServerCache::handleRetrieval` that way the cost of the lock will only be incurred on a cache miss, which should have a low probability (since we usually see cache hit rates of ~90%).

@WenbinZhu I think that there are other bottlenecks in the stack that we would hit before this being a problem. You can get some sense of read lock latency without contention here https://mechanical-sympathy.blogspot.com/2013/08/lock-based-vs-lock-free-concurrent.html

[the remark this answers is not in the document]
xnull: LogUnitServer is an entry point for locks, I don't like the idea of moving it to underlying classes, it will make code more complex.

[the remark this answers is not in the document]
xnull: Locks are super fast, no need to optimize it, according to the article you have provided it's ~5million op/sec, which is ~20-30 times faster than reading from caffeine. The overall performance of reading from Corfu node is 5k per sec, do we really want to optimize lock with millions of ops per sec?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zhangn49: @xnull Is epoch check missing here?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: As discussed twice the check will crash most of the unit tests, here is the build with the change https://travis-ci.org/github/CorfuDB/CorfuDB/builds/716015026

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: As discussed twice the check will crash most of the unit tests, here is the build with the change https://travis-ci.org/github/CorfuDB/CorfuDB/builds/716015026

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: also see my comment https://github.com/CorfuDB/CorfuDB/pull/2661#issuecomment-667404526

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: Unfortunately, the change break tests https://github.com/CorfuDB/CorfuDB/pull/2661#issuecomment-667404526

xnull:
>Hard to review when addressed comments get reverted after an approval. I still think the epoch checks are needed for the read path. If the UT are failing then we need to fix them.

@Maithem I totally agree, but how I can bring that change if it breaks everything else? I tried twice, you can see it, what we can do to make it work, what is your opinion?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -374,28 +424,40 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
+        // prevent multiple reset in the same epoch and should be equal to the current router
         // epoch to prevent stale reset requests from wiping out the data.
-        if (msg.getPayload() > serverContext.getLogUnitEpochWaterMark()
-                && msg.getPayload() == serverContext.getServerEpoch()) {
-            serverContext.setLogUnitEpochWaterMark(msg.getPayload());
-            batchWriter.addTask(RESET, msg)
+        serverContext.setLogUnitEpochWaterMark(msg.getPayload());
+
+        Lock lock = resetLock.writeLock();
+        lock.lock();
```
WenbinZhu, line 440: Locking here is bad, the reset op is added to BatchProcessor, but will be run after all previous write operations in the queue are finished, during that all reads will be blocked. This is a performance issue.

Also if you put lock into streamLogFiles then you don't need to do cf.join() below.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -273,26 +295,40 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
...
 
-        ReadResponse rr = new ReadResponse();
+        if (msg.getEpoch() != serverContext.getServerEpoch()) {
+            r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH,
+                    serverContext.getServerEpoch()));
+            return;
+        }
+
+        Lock lock = resetLock.readLock();
+        lock.lock();
```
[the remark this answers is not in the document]
xnull: - Please see my comment here https://github.com/CorfuDB/CorfuDB/pull/2661#discussion_r467181541
If we allow reading data from the cache when we don't have data in the database, I see it as a big issue.

- 'reset' happens very rarely in comparison to reads,
- reset happens when a node is in the unresponsive list, so, reads not going to be affected.

Could you please explain your motivation for inconsistent reads?
Let's chat privately

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -273,26 +295,40 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
             log.trace("read: {}, cacheable: {}", msg.getPayload().getAddresses(), cacheable);
         }
 
-        ReadResponse rr = new ReadResponse();
+        if (msg.getEpoch() != serverContext.getServerEpoch()) {
+            r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH,
+                    serverContext.getServerEpoch()));
+            return;
+        }
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -273,26 +286,26 @@ public void read(CorfuPayloadMsg<ReadRequest> msg, ChannelHandlerContext ctx, IS
...
+                        rr.put(address, LogData.getEmpty(address));
+                    } else {
+                        rr.put(address, (LogData) logData);
+                    }
+                } catch (DataCorruptionException e) {
+                    log.error(
+                            "Data corruption exception while reading addresses {}",
+                            msg.getPayload().getAddresses(),
+                            e);
+                    r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_CORRUPTION.payloadMsg(address));
```
WenbinZhu, line 304: Please fix the indent.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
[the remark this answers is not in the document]
xnull: resetLock moved into StreamLogFiles

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -374,28 +424,40 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
+                || msg.getPayload() != serverContext.getServerEpoch()) {
+            log.info("LogUnit Server Reset request received but reset already done.");
+            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+            return;
+        }
+
         // Check if the reset request is with an epoch greater than the last reset epoch seen to
-        // prevent multiple reset in the same epoch. and should be equal to the current router
+        // prevent multiple reset in the same epoch and should be equal to the current router
         // epoch to prevent stale reset requests from wiping out the data.
```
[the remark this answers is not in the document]
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: done

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xnull: fixed

Maithem:
Universe tests are failing

PR #2663, opened 2020-07-27
Fix FSM Unit Test and Snapshot Plugin Condition

annym:
## Overview

Description:

- Fix Intermittent Failure on FSM Unit Test
- Fix onSnapshotEnd condition to check for an ack SNAPSHOT_END instead of SNAPSHOT_REPLICATED (intermediate acks)

Why should this be merged: stabilize log replication master branch

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/TestDataSender.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/TestDataSender.java
@@ -4,6 +4,7 @@ import lombok.Getter;
 import org.corfudb.infrastructure.logreplication.DataSender;
 import org.corfudb.infrastructure.logreplication.replication.send.LogReplicationError;
 import org.corfudb.protocols.wireprotocol.logreplication.LogReplicationEntry;
+import org.corfudb.protocols.wireprotocol.logreplication.LogReplicationEntryMetadata;
 import org.corfudb.protocols.wireprotocol.logreplication.MessageType;
 
 import java.util.LinkedList;
@@ -13,7 +14,8 @@ import java.util.concurrent.CompletableFuture;
 
 
 /**
- * Test Implementation of Snapshot Data Sender
+ * Test Implementation of Snapshot Data Sender which emulates sending messages by placing directly
+ * in an entry queue, and sends ACKs right away.
  */
 public class TestDataSender implements DataSender {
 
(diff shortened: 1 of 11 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/TestDataSender.java
```diff
@@ -34,7 +36,18 @@ public TestDataSender() {
         }
 
         CompletableFuture<LogReplicationEntry> cf = new CompletableFuture<>();
-        LogReplicationEntry ack = LogReplicationEntry.generateAck(message.getMetadata());
```
pankti-m, line 37: nit - probably we should rename this to setter of something.. because it is only setting the metadata.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/TestSnapshotReader.java
```diff
@@ -40,23 +40,24 @@ public SnapshotReadMessage read(UUID snapshotRequestId) {
         // Connect to endpoint
         List<LogReplicationEntry> messages = new ArrayList<>();
 
+        int index = globalIndex;
+
+        // Limit to read as max as BatchSize and until the maximum baseSnapshot
+        for (int i=index; (i<(index+config.getBatchSize()) && index<baseSnapshot) ; i++) {
```
pankti-m, line 46: index is never getting updated.  will it ever reach the condition index<baseSnapshot?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
```diff
@@ -272,26 +272,33 @@ private LogReplicationEntry processReceivedMessage(LogReplicationEntry message)
...
+            // If the snapshot sync apply has completed (determined by the end marker) notify
+            // plugin the completion, so checkpoint/trim process is resumed.
+            if (ack.getMetadata().getMessageMetadataType() == SNAPSHOT_END) {
+                processSnapshotSyncApplied(ack);
             }
             return ack;
         }
     }
 
+    private void processSnapshotSyncApplied(LogReplicationEntry ack) {
```
pankti-m, line 284: apply has not been done yet, right?  Only transfer has completed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
```diff
@@ -53,8 +52,7 @@
     /*
      * If there is a timeout for a message, should generate an error or not
      */
-    private boolean errorOnMsgTimeout = true;
-
+    private boolean errorOnMsgTimeout;
```
pankti-m, line 55: default value will be false which can change the behavior.. is it intentional?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/TestDataSender.java
annym: Yes, I had this comment on another PR and that PR is addressing it, so I'm not going to touch this here to not conflict.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/TestSnapshotReader.java
annym: it is in line 43

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
annym: the static code analysis asks to remove this redundant. It is initialized in the constructor.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationSinkManager.java
annym: It has been applied already.

xiaoqin2012:
Can you add a change to this PR that make sure that the snapshot_end has the correct sequence number?

PR #2666, opened 2020-07-28
Log Entry Reader Bug Fix

annym:
## Overview

Description:

When finding streams that we are not interested in replicating
do not modify original Entries from OpaqueEntry.

Why should this be merged: 

- [ ] Public API has Javadoc


xiaoqin2012:
Can we add a test case that the last log entry doesn't contain interested streams and it doesn't loop on it.

annym:
> Can we add a test case that the last log entry doesn't contain interested streams and it doesn't loop on it.

Added.

PR #2673, opened 2020-07-30
Add option to load data into corfu table

hisundar:
Using corfu browser we can now load data into a dummy table.
This is useful for testing checkpoint issues as well as live testing log-replication feature

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -3,7 +3,6 @@ package org.corfudb.browser;
 import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -203,4 +206,45 @@ public int dropTable(String namespace, String tablename) {
...
+                    TableName.getDefaultInstance().getClass(),
+                    TableName.getDefaultInstance().getClass(),
+                    optionsBuilder.build());
+
+            TableName dummyVal = TableName.newBuilder().setNamespace(namespace).setTableName(tablename).build();
+            final int numTxn = (numItems > batchSize) ? numItems / batchSize : 1;
+            for (int i = numTxn; i > 0; i--) {
+                log.info("loadTable: Batches left {}", i);
+                TxBuilder tx = store.tx(namespace);
+                for (int j = batchSize; j > 0; j--) {
```
xiaoqin2012, line 238: i-- should move to here along with j--.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -203,4 +206,45 @@ public int dropTable(String namespace, String tablename) {
...
+                    TableName dummyKey = TableName.newBuilder().setNamespace(Integer.toString(numItems))
+                            .setTableName(Integer.toString(j)).build();
+                    tx.update(tablename, dummyKey, dummyVal, dummyVal);
+                }
+                tx.commit();
+            }
+        } catch (Exception e) {
+            log.error("loadTable: {} {} {} {} failed.", namespace, tablename, numItems, batchSize, e);
+        }
+        return numItems/batchSize;
```
annym, line 248: Probably it should be (int)Math.ceil((double)numItems/batchSize);
Imagine 10 entries, in batches of 8, it would return 1, instead of 2.

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
```diff
@@ -132,6 +132,26 @@ public void browserTest() throws
...
+        runSinglePersistentServer(corfuSingleNodeHost,
+                corfuStringNodePort);
+
+        // Start a Corfu runtime
+        runtime = createRuntime(singleNodeEndpoint);
+        final int numItems = 100;
+        final int batchSize = 10;
+
+        CorfuStoreBrowser browser = new CorfuStoreBrowser(runtime);
+        Assert.assertEquals(browser.loadTable(namespace, tableName, numItems, batchSize), batchSize);
```
annym, line 152: In line with the above comment probably we should explicitly expect numItems/batchSize (in case someone changes numItems)... or use non-divisible numbers for test completeness.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: done. thanks

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
hisundar: the return type is purely for testing - no one using the browser will see it.
thanks

PR #2678, opened 2020-07-31
Sequencer Reconfiguration Corruption Fixes

Maithem:
## Overview
Prior to this patch, the global tail, stream tails and stream bit
sets were retrieved and computed from the first node in the last
segment. This assumption is incorrect and leads to the possibility
of reconfiguring the sequencer with state that doesn't reflect the
log. In consequence, leading to inconsistent reads, data loss and
cluster instability.

The patch introduces multiple fixes to compute the sequencer state
from the right head node (i.e. the head node of all segments) and
forces the computation part and reconfiguration to happen on the
same epoch.

Why should this be merged: Fixes data loss/inconsistency/stability issues

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,7 +117,83 @@ public static String toReadableId(UUID id) {
...
+     * segments.
+     * @param layout layout to use to find the head node
+     * @return head node
+     * @throws IllegalStateException if no such node exists
+     */
+    private static String selectHeadNode(Layout layout) {
+        checkNotNull(layout);
+        validateSegments(layout.getSegments());
+        List<Layout.LayoutSegment> segments = layout.getSegments();
+        // validate segments (start end) contiguous
```
Maithem, line 154: Already done in `validateSegments`

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+    /**
+     * This method selects a head node: the first node (in the open segment) that also exists in all previous
+     * segments.
+     * @param layout layout to use to find the head node
+     * @return head node
+     * @throws IllegalStateException if no such node exists
+     */
+    private static String selectHeadNode(Layout layout) {
+        checkNotNull(layout);
+        validateSegments(layout.getSegments());
```
xnull, line 152: Segments and stripes can't be null https://github.com/CorfuDB/CorfuDB/blob/4ade684cd5f833bfffb26f5a84077b20d7020411/runtime/src/main/java/org/corfudb/runtime/view/Layout.java#L128

No need to check that they are not null one more time.

It's supposed by default that the layout is in a correct state. 
If you see any parameters in the layout that are not validates, it's better to add them into the Layout class

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+        List<Layout.LayoutSegment> segments = layout.getSegments();
+
+        // select open segment (i.e. last segment)
+        Layout.LayoutSegment openSegment = segments.get(segments.size() - 1);
```
xnull, line 156: `layout.getLatestSegment()`

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+        // only support chain replication
+        Layout.LayoutStripe openSegmentStripe = openSegment.getStripes().get(0);
```
xnull, line 158: openSegment.getFirstStripe()

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+
+        if (segments.size() == 1) {
+            // Since there is only one segment we can return the first node in the stripe (i.e. head node)
+            return segments.get(0).getStripes().get(0).getLogServers().get(0);
```
xnull, line 162: `return layout.getFirstSegment().getFirstStripe().getLogServers().get(0);`

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+        }
+
+        // walk the chain of the last segment and find the first node that exists in all previous segments
+        List<Layout.LayoutSegment> reversedSegmentPrefix = segments.subList(0, segments.size() - 1);
+        checkState(!reversedSegmentPrefix.isEmpty());
```
xnull, line 167: It's hard to understand what this check does.
What's the motivation for:
 - checking that `segments` list contains more than one element if we already checked that condition in the previous statement (a couple of lines above)?
 - reverse `reversedSegmentPrefix` collection two times, which gives us the same collection in the end

Cognitive complexity of this method seems unnecessary complicated

xnull: Sorry confused with the name of `reversedSegmentPrefix `

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+        // Reverse the prefix to simplify the indexing when walking the segments backwards (i.e. high to low)
+        Collections.reverse(reversedSegmentPrefix);
+
+        for (String node : openSegmentStripe.getLogServers()) {
+            boolean isNodeMissing = false;
+            for (Layout.LayoutSegment currentSegment : reversedSegmentPrefix) {
```
xnull, line 173: If you are looking for a node that is present in all segments then no need to reverse the collection.
And a `double negative` conditions make the logic more error-prone.

For instance: !isNodeMissing - if the node is not missing, actually means - if the node present in all segments.
Also in the `for` cycle has to be simplified

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+    /**
+     * Verify that the stripes are not empty and that each segment has one stripe.
+     * @param segments segments to validate
+     */
+    private static void validateSegments(List<Layout.LayoutSegment> segments) {
+        checkNotNull(segments);
+        checkArgument(!segments.isEmpty());
+        long previousSegmentEndAddress = 0;
+        for (Layout.LayoutSegment segment : segments) {
+            checkState(segment.getStart() == previousSegmentEndAddress);
```
pankti-m, line 130: shouldnt this be segment.getStart() == previousSegmentAddress + 1 ?

WenbinZhu:
@Maithem Should we also fix the issue of the chain order, because that is the root cause of the this sequncer reconfiguration issue because it's based on the order assumption. And I think we have other places that assumes the order of the chain.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Those are checks that happen in the constructor, but the Layout.java is not immutable. Might move these checks to the builder and make the Layout.java immutable in a subsequent patch.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: I'm trying to find the first node from the last segment that also exists in all segments. Which is different from finding a node that exists in all segments. just consider the case where you have all nodes in all segments. Then obviously not all nodes can be the head segment. 

Actually, although this algorithm is better than what we have on master, it is still in **incorrect.** It's not possible to compute the sequencer state from only **one** node. I think.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: > If you are looking for a node that is present in all segments then no need to reverse the collection.

The algorithm isn't doing that. Although a head node will exist in all segments, merely existing in all segments doesn't make it a "head node". A head node is a node that's suppose to have the largest log prefix.

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -415,8 +416,13 @@ public void reconfigureSequencerServers(Layout originalLayout, Layout newLayout,
                         || !originalLayout.getPrimarySequencer()
                         .equals(newLayout.getPrimarySequencer())) {
 
-                    StreamsAddressResponse streamsAddressesResponse = runtime.getAddressSpaceView()
-                            .getLogAddressSpace();
+                    // The sequencer state needs to be retrieved from the head logunit from the
+                    // new layout. Note that AddressSpaceView::getLogAddressSpace shouldn't be
+                    // used because there is no guarantee that the same head node from newLayout
```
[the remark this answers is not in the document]
Maithem: Done. Hope its clearer.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: I don't think so.
Consider the case of one segment [0, -1) , then previousSegmentEndAddress=0 is equal to segment.getStart() which is 0.
Now consider two segments [0,100) [100, -1) , the check is also true here. When we split a segment, the end address of the segment is **non-inclusive** . If they were inclusive, then we need to check with +1 , but they are not.

runtime/src/main/java/org/corfudb/util/Utils.java
xnull: Regarding current code, I played a bit with the code, it seems to me it can be simplified to
```
LayoutStripe openSegmentStripe = layout.getLatestSegment().getFirstStripe();

        for (String node : openSegmentStripe.getLogServers()) {
            boolean nodeIsPresent = segments.stream()
                    .allMatch(currentSegment -> currentSegment.getAllLogServers().contains(node));

            if (nodeIsPresent) {
                return node;
            }
        }
```
If it would work for you, then great, if it's not correct then ok, no worries

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: > @Maithem Should we also fix the issue of the chain order, because that is the root cause of the this sequncer reconfiguration issue because it's based on the order assumption. And I think we have other places that assumes the order of the chain.

Given how state transfer works, I don't think its possible to maintain the same order of all nodes on all segments.

Consider the following scenario.

S1:{n2} S2:{n2, n3}  epoch = 5

then at epoch 6 we add n1

S1:{n2} S2:{n2, n3} S3:{n2, n3, n1}

At epoch 7 n1's state transfer completes
S1:{n2, n1} S2:{n2, n3, n1} S3:{n2, n3, n1}

At epoch 8, node two dies
S1:{n1} S2:{n3, n1} S3:{n3, n1}

As you can see here, node 1 is the tail for S3, but the head for S1. So its not possible to enforce order on the segments unless we schedule the state transfers properly, but I dont think that's a good idea. 

I'm changing my algorithm to a simpler approach where I select the head node of each segment and then aggregate the results.

Maithem: Actually, in the example above its clear that the sequencer state cannot be computed by selected only **one** node, so the current algorithm is incorrect. That's why we need to aggregate results from multiple nodes.

Maithem: Not sure why this replied to the wrong thread. @WenbinZhu see response above.

WenbinZhu: @Maithem I see, this is an edge case in layout reconfiguration, n1 finishes state transfer earlier than n3.
I'll continue the review.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -163,6 +235,7 @@ public static void prefixTrim(RuntimeLayout runtimeLayout, Token address) {
 
     /**
      * Get the maximum committed log tail from all log units.
+     * Get global log tail. TODO(Maithem): to be executed on same layout as seal
```
WenbinZhu, line 238: Looks like this todo is addressed?

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -415,8 +416,13 @@ public void reconfigureSequencerServers(Layout originalLayout, Layout newLayout,
                         || !originalLayout.getPrimarySequencer()
                         .equals(newLayout.getPrimarySequencer())) {
 
-                    StreamsAddressResponse streamsAddressesResponse = runtime.getAddressSpaceView()
-                            .getLogAddressSpace();
+                    // The sequencer state needs to be retrieved from the head logunit from the
+                    // new layout. Note that AddressSpaceView::getLogAddressSpace shouldn't be
```
WenbinZhu, line 420: Can we remove AddressSpaceView::getLogAddressSpace? That method was only used here and in a couple of unit tests, since we are now using Utils instead, we can remove that method and also change the unit tests to use Utils as well.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+     * @throws IllegalStateException if no such node exists
+     */
+    private static String selectHeadNode(Layout layout) {
+        checkNotNull(layout);
+        validateSegments(layout.getSegments());
+        List<Layout.LayoutSegment> segments = layout.getSegments();
+
+        // select open segment (i.e. last segment)
+        Layout.LayoutSegment openSegment = segments.get(segments.size() - 1);
+        // only support chain replication
```
WenbinZhu, line 157: Is this comment misplaced? Looks like it should be only support single strip.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+            }
+
+            if (!isNodeMissing) {
+                // this node is the first node in the last chain that exists in all previous segments (i.e. head node
+                // of the whole chain across all segments)
+                return node;
+            }
+        }
+
+        // need to log layout
```
WenbinZhu, line 190: Can we reword this comment?

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+        if (segments.size() == 1) {
+            // Since there is only one segment we can return the first node in the stripe (i.e. head node)
+            return segments.get(0).getStripes().get(0).getLogServers().get(0);
+        }
+
+        // walk the chain of the last segment and find the first node that exists in all previous segments
+        List<Layout.LayoutSegment> reversedSegmentPrefix = segments.subList(0, segments.size() - 1);
+        checkState(!reversedSegmentPrefix.isEmpty());
+        // Reverse the prefix to simplify the indexing when walking the segments backwards (i.e. high to low)
+        Collections.reverse(reversedSegmentPrefix);
```
WenbinZhu, line 169: I not understanding why we need to walk the segments backwards? If we need to check all the segments, does it matter if we do forward or backward? If it's for performance consideration, it looks like the first segment has a higher probablility to not having the candidate head node? S1:{n1} S2:{n3, n1} S3:{n3, n1}

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -119,6 +117,80 @@ public static String toReadableId(UUID id) {
...
+    }
+
+    /**
+     * This method selects a head node: the first node (in the open segment) that also exists in all previous
+     * segments.
+     * @param layout layout to use to find the head node
+     * @return head node
+     * @throws IllegalStateException if no such node exists
+     */
+    private static String selectHeadNode(Layout layout) {
```
WenbinZhu, line 150: I think we need to add a small unit test for this method.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Correct. Like I mentioned earlier its always possible to compute the sequencer state from only one node, so we need to aggregate. I will simplify this logic by ORing all heads from all segments.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Yes. Writing new unit tests for the new algorithm

Maithem:
subsumed by #2715

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Changed algorithm. Please see, the logic is cleaner #2715

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -415,8 +416,13 @@ public void reconfigureSequencerServers(Layout originalLayout, Layout newLayout,
...
+                    // used because there is no guarantee that the same head node from newLayout
+                    // will be the same head not selected by the AbstractView
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Yes.

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
Maithem: Done.

PR #2496, opened 2020-03-31
For AddNode create a new client router

PavelZaytsev:
## Overview

Description:
If a client invalidates a layout in another thread while performing an AddNode in the current thread, there is a possibility of accidentally pruning the router to the node being added from the NodeRouterPool, preventing a client from initiating an AddNodeWorkflow on the server's orchestrator. It is better to create the client router on demand for pings and bootstraps before issuing a workflow request since the node being added is not a part of the layout yet.

Why should this be merged: 
Fixes the defect and potential bug. 

Related issue(s) (if applicable): #<number>

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #2680, opened 2020-08-03
Allowing retries when RuntimeException gets raised during AddNode workflow

achinmishra:
## Overview
Description:
Hitting the bug in the AddNode() where we need to retry when RuntimeException is raised from bootstrapManagementServer(). bootstrapManagementServer() wraps TimeoutException as RuntimeException. 

Currently, RuntimeException gets trickled down to the clients. There should be a retry logic when there is a TimeoutException in bootstrapManagementServer().

Why should this be merged: 
Enabling the retry logic during RuntimeException

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -141,6 +141,9 @@ public void invoke() {
                     throw new IllegalStateException("Orchestrator can not be selected");
                 }
 
+            } catch(RuntimeException e) {
```
WenbinZhu, line 144: This catch will mask the next catch block, because NetworkException and IllegalStateException are also RuntimeException, you need to put it after the next catch block. Please also check with @PavelZaytsev if initially he intentionally wanted to not retry on other runtime exceptions, as there might be some edge cases.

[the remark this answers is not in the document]
vjeko: You need to check for the root cause. If the root cause is not an instance of `TImeoutException`, you will want to re-throw.

PavelZaytsev:
I think it would be better to rewrite a bootstrapManagementServer to propagate needed exceptions and ignore AlreadyBootstrappedException, or even better - refactor BootstrapUtil to use its bootstrapManagementServer that is being created from the brand new router instead of a router pool. I had a PR somewhere that you can use for reference.
https://github.com/CorfuDB/CorfuDB/pull/2496

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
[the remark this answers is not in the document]
vjeko: Also, since Corfu's exception handling is pretty screwed up, it might make sense to make exception checking more generic:

```
            } catch (TimeoutException e) {
                log.warn("WorkflowRequest: Error while running {} on attempt {}, cause:", this, x, e);
            } catch (RuntimeException runtimeException) {
                handleException(runtimeException);
            }

    private void handleException(RuntimeException runtimeException) {
        final Throwable cause = ExceptionUtils.getRootCause(runtimeException);
        if (cause instanceof TimeoutException ||
                cause instanceof IllegalStateException ||
                cause instanceof NetworkException) {
            log.warn("WorkflowRequest: Error while running {} with cause:", this, runtimeException);
        } else {
            throw runtimeException;
        }
    }
```

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -141,6 +141,9 @@ public void invoke() {
...
+                log.warn("WorkflowRequest: Additional handling of the exception wrapped herein." +
+                        " Running {} on attempt {},  cause:", this, x, e);
```
xnull, line 146: Minor thing. Since you touched this piece of code, would be great if you rename `x` to `retryIteration`

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
achinmishra: Moved the catch block. It was gonna mask other exceptions.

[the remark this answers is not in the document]
achinmishra: 1. The root cause was an instance of TimeoutException, coming from bootstrapManagementServer(). 
2. I think its a good idea to add a method listed as above to handle the exceptions gracefully. I will make those changes.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
achinmishra: Right! Better code readability. Pushing the change.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -119,6 +120,20 @@ private void waitForWorkflow(@NonNull UUID workflow, @NonNull ManagementClient c
         throw new TimeoutException();
     }
 
+    /*
+     * private method to handle exceptions while sendRequest()
+     */
+    private void handleException(RuntimeException runtimeException) {
+        final Throwable cause = ExceptionUtils.getRootCause(runtimeException);
+        if (cause instanceof TimeoutException ||
```
xnull, line 128: TimeoutException can't appear here, because it already handled in `invoke` method

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -119,6 +120,20 @@ private void waitForWorkflow(@NonNull UUID workflow, @NonNull ManagementClient c
...
+                cause instanceof IllegalStateException ||
+                cause instanceof NetworkException) {
+            log.warn("WorkflowRequest: Error while running {} with cause:", this, runtimeException);
```
pankti-m, line 131: another curly bracket needed for runtimeException

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -141,8 +156,10 @@ public void invoke() {
                     throw new IllegalStateException("Orchestrator can not be selected");
                 }
 
-            } catch (NetworkException | TimeoutException | IllegalStateException e) {
-                log.warn("WorkflowRequest: Error while running {} on attempt {}, cause:", this, x, e);
+            } catch (TimeoutException e) {
+                log.warn("WorkflowRequest: Error while running {} on attempt {}, cause:", this, retryIteration, e);
```
pankti-m, line 160: another curly bracket required for exception

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
xnull: Unfortunately, you are not correct, another bracket is not needed

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
xnull: Unfortunately, you are not correct, another bracket is not needed

Maithem: @pankti-m the `log.warn` is overloaded with `void warn(String var1, Object var2, Object var3);`  , so it doesn't need a string specific (i.e. '{}') to log the exception. You can just do `log.warn("msg", new ExceptionType());` and it will log the exception properly.

pankti-m: Good to know.  Thanks @Maithem

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -141,8 +156,10 @@ public void invoke() {
...
+            } catch (RuntimeException e) {
+                handleException(e);
```
xnull, line 162: For your consideration - I personally vote for getting rid of handleException, it reduces the complexity of the code if you make it flat

achinmishra: Let me make a note of your suggestion @xnull

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -140,9 +155,10 @@ public void invoke() {
                 else{
                     throw new IllegalStateException("Orchestrator can not be selected");
                 }
-
-            } catch (NetworkException | TimeoutException | IllegalStateException e) {
-                log.warn("WorkflowRequest: Error while running {} on attempt {}, cause:", this, x, e);
+            } catch (IllegalStateException | TimeoutException e) {
+                log.warn("WorkflowRequest: Error while running {} on attempt {}, cause:", this, retryIteration, e);
+            } catch (RuntimeException e) {
+                handleException(e);
```
WenbinZhu, line 161: This logic here is kind of weird, why not just catch RuntimeException and check "instance of" and cause in side the catch? IllegalStateException is a RuntimeException, and TimeoutException should not be thrown without being wrapped in RuntimeException, because it's a checked exception.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -119,6 +120,20 @@ private void waitForWorkflow(@NonNull UUID workflow, @NonNull ManagementClient c
...
     }
 
+    /*
+     * private method to handle exceptions while sendRequest()
+     */
+    private void handleException(RuntimeException runtimeException) {
+        final Throwable cause = ExceptionUtils.getRootCause(runtimeException);
+        if (cause instanceof TimeoutException ||
+                cause instanceof IllegalStateException ||
+                cause instanceof NetworkException) {
```
WenbinZhu, line 130: This seems incorrect. IllegalStateException and NetworkException are RuntimeException, so the cause returned by ExceptionUtils.getRootCause() is null, so "instance of" will fail, and exception will be thrown without being retried.

Here is a quick example for reference:
```java
@Test
    public void xxx() {
        try {
            throw new NetworkException("s", "t");
        } catch (RuntimeException e) {
            final Throwable cause = ExceptionUtils.getRootCause(e);
            if (cause instanceof NetworkException) {
                System.out.println(1111);
            } else {
                System.out.println(2222);
            }
        }
    }
```
The code prints 222 because `cause` is actually null.

runtime/src/main/java/org/corfudb/runtime/view/workflows/WorkflowRequest.java
```diff
@@ -130,19 +142,20 @@ private void waitForWorkflow(@NonNull UUID workflow, @NonNull ManagementClient c
...
             try {
                 Optional<ManagementClient> orchestrator = getOrchestrator();
                 if(orchestrator.isPresent()){
                     UUID workflowId = sendRequest(orchestrator.get());
                     waitForWorkflow(workflowId, orchestrator.get(), timeout, pollPeriod);
                 }
-                else{
-                    throw new IllegalStateException("Orchestrator can not be selected");
+                else {
+                    log.warn("Orchestrator can not be selected");
```
WenbinZhu, line 153: You may need to add the number of attemp in the log, same as what it does when we catch the exception.

achinmishra: Noted. Thanks

Maithem:
This failed the IF tests twice. Please investigate what's going on.

achinmishra:
> This failed the IF tests twice. Please investigate what's going on.

@Maithem injection framework details server is timing out. Any alternative to reach the reason for the failure?

Maithem:
> > This failed the IF tests twice. Please investigate what's going on.
> 
> @Maithem injection framework details server is timing out. Any alternative to reach the reason for the failure?

I think this means a specific test failed, not a timeout. Why do you think it was timing out?

image

PR #2681, opened 2020-08-05
Sync Status API improvements.

pankti-m:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
@@ -14,7 +14,6 @@ import java.util.UUID;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -23,6 +22,9 @@ public class LogReplicationAckReader {
     private LogReplicationConfig config;
     private CorfuRuntime runtime;
     private String remoteClusterId;
+    // Log tail when the current snapshot sync started.  We do not need to synchronize access to it because it will not
+    // be read(calculateRemainingEntriesToSend) and written(setBaseSnapshot) concurrently.
+    private long baseSnapshotTimestamp;
 
     /*
      * Periodic Thread which reads the last Acked Timestamp and writes it to the metadata table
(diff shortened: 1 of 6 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
```diff
@@ -98,28 +98,51 @@ private long getMaxReplicatedStreamsTail() {
...
      * For Log Entry Sync, this function returns the total number of entries remaining to be sent across all replicated
      * streams.
-     * For Snapshot Sync, each entry sent is a snapshot.  So this function returns the total number of snapshots
-     * remaining to be sent.
-     * If the ack'd timestamp is uninitialized, it returns 100%, which means no replication has been done.
+     *
+     * For Snapshot Sync, each entry sent is a snapshot of the corresponding stream.  The stream could have been
+     * checkpointed and trimmed so we cannot find the remaining number of entries accurately.  In this case, we simply
+     * subtract the acked timestamp from the global log tail when the snapshot sync started.
+     * Note that this method is not accurate because the global tail advances on writes for both replicated and
```
annym, line 105: This is true but a bit misleading as someone could say well don't return global tail but maxTailStreamsToReplicate, and it still holds true that this is not accurate as there is no guarantee that up to maxTailStreamsToReplicate all data is to be replicated. We might rephrase saying something like: Note this method is not accurate as the max tail can reflect the interleaving of replicated and non-replicated streams (or something similar)...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
```diff
@@ -98,28 +98,51 @@ private long getMaxReplicatedStreamsTail() {
...
+            return NO_REPLICATION_REMAINING_ENTRIES;
         }
-        if (ackedTimestamp == Address.NON_ADDRESS) {
-            return FULL_REPLICATION_REMAINING_PERCENT;
+
+        // If doing a snapshot sync and nothing has been acked, all replication is remaining.
+        // So send the stream tail.
+        // NOTE: This is an approximation
+        if (ackedTimestamp == Address.NON_ADDRESS &&
+                lastSyncType == LogReplicationMetadata.ReplicationStatusVal.SyncType.SNAPSHOT) {
```
annym, line 124: I was thinking, we might not need this, right? Also it could be imprecise, in the case deltas are already coming in, so the maxReplicatedStreamTail is moving beyond global tail... we could just calculate all snapshot sync in L145, right?

annym: Yes, that's a good option.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
```diff
@@ -23,6 +22,9 @@
     private LogReplicationConfig config;
     private CorfuRuntime runtime;
     private String remoteClusterId;
+    // Log tail when the current snapshot sync started.  We do not need to synchronize access to it because it will not
+    // be read(calculateRemainingEntriesToSend) and written(setBaseSnapshot) concurrently.
```
annym, line 26: it could be read and written concurrently, right? cause the calculation is scheduled and it could happen to coincide?

PR #2683, opened 2020-08-05
Standby Connection Loss Fix

annym:
## Overview

Description:

1. Fix Issue when Standby connection is lost
2. Fix replication server graceful shutdown
3. Add logging
4. Adjust default and configurable lock values: monitor, renewal, lease duration and lease check

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -25,6 +25,8 @@ import org.corfudb.util.retry.RetryNeededException;
 import org.corfudb.utils.lock.Lock;
 import org.corfudb.utils.lock.LockClient;
 import org.corfudb.utils.lock.LockListener;
+import org.corfudb.utils.lock.states.HasLeaseState;
+import org.corfudb.utils.lock.states.LockState;
 
 import javax.annotation.Nonnull;
 import java.time.Duration;
@@ -51,6 +53,16 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      * Wait interval (in seconds) between consecutive fetch topology attempts to cap exponential back-off.
      */
     private static final int FETCH_THRESHOLD = 300;
+
+    /**
+     * Fraction of Lease Duration for Lease Renewal
+     */
+    private static final int RENEWAL_LEASE_FRACTION = 4;
+
+    /**
+     * Fraction of Lease Duration for Lease Monitoring
+     */
+    private static final int MONITOR_LEASE_FRACTION = 10;
 
     /**
      * Bookkeeping the topologyConfigId, version number and other log replication state information.
@@ -130,6 +142,8 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private volatile AtomicBoolean isLeader;
 
     private LogReplicationServer logReplicationServerHandler;
+
+    private LockClient lockClient;
 
     /**
      * Indicates the server has been started. A server is started once it is determined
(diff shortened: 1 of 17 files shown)


utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -48,9 +48,13 @@
...
     private final ScheduledExecutorService lockMonitorScheduler;
 
+    private final ScheduledExecutorService taskScheduler;
+
+    private final ExecutorService lockListenerExecutor;
+
     // duration between monitoring runs
     @Setter
-    private static int DurationBetweenLockMonitorRuns = 60;
+    private static int DurationBetweenLockMonitorRuns = 30;
```
xiaoqin2012, line 57: What is the original value? The monitoring can be smaller so that when the lease expires, the other waiters can get hold of the lease quickly.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -52,6 +54,16 @@
...
+    /**
+     * Fraction of Lease Duration for Lease Renewal
+     */
+    private static final int RENEWAL_LEASE_FRACTION = 4;
+
+    /**
+     * Fraction of Lease Duration for Lease Monitoring
+     */
+    private static final int MONITOR_LEASE_FRACTION = 2;
+
```
xiaoqin2012, line 66: Can we make the fraction bigger such as 10 or 20?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -131,6 +143,8 @@
 
     private LogReplicationServer logReplicationServerHandler;
 
+    private LockClient lock;
```
pankti-m, line 146: is it better to rename to lockClient?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -704,6 +720,10 @@ public void shutdown() {
         if(clusterManagerAdapter != null) {
             clusterManagerAdapter.shutdown();
         }
+
+        if (lock != null) {
```
xiaoqin2012, line 724: Will it cleaner to call lock.shutdown in the server shutdown call?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
```diff
@@ -413,7 +413,8 @@ public synchronized void onConnectionUp(String endpoint) {
      */
```
xiaoqin2012, line 413: When will this callback be called? Which component detects the connection down?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
annym: The original lease value is 60seconds, to 10 maybe?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: That would mean exposing it. And the lock is really created by the Discovery Service, I think keeping it to the DiscoveryService scope is fine.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
annym: The Transport Plugin is the one detecting connections coming up or down.

xiaoqin2012: I am just wondering what happen to the snapshot sender/log entry sender  when there is a connection loss and what happen to the fsm state machine.

annym: Currently, what happens is that if connection is lost to the remote leader we move to verify remote leader state (if any other connection is still up) or to wait for connections (if no connection is available). But we don't do anything to the snapshot/log entry senders. So good point, this is something we should add. I'm adding that we stop log replication, can you help confirm if that does the correct configuration of the snapshot/log entry sender? or if something else is needed...

xiaoqin2012: I saw that the sender catch all exceptions and called cancel.  There are two options:
1. the sender identify the exception type and generate stop event instead of cancel.
2. when the replication state machine detects the current leader connection loss, put a stop event for FSM.

xiaoqin2012: We can do in another PR ...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -158,6 +159,7 @@ public CorfuLogReplicationRuntime(LogReplicationRuntimeParameters parameters, Lo
         this.router.addClient(new LogReplicationHandler());
         this.sourceManager = new LogReplicationSourceManager(parameters, new LogReplicationClient(router, remoteClusterId),
             metadataManager);
+        this.connectedEndpoints = new HashSet<>();
```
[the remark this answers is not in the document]
annym: I removed that as all the methods that modify or access this object are synchronized.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
annym: I did the option 2.

xiaoqin2012: We need to add try and catch for processReads to catch the exceptions.

PR #2684, opened 2020-08-07
Dedicated Executor for acquired lock monitoring

annym:
## Overview

Description:

Add a dedicated executor for lock acquired monitoring. The reason is that sharing the executor with the lease renewal task can lead to a blocking scenario, where if lease renewal task is stuck, lock client would not expire on the current lock (already lost). 

Why should this be merged: fixes a potential bug when lease renewal is stuck.

### DIFF
--- a/utils/src/main/java/org/corfudb/utils/lock/LockClient.java
+++ b/utils/src/main/java/org/corfudb/utils/lock/LockClient.java
@@ -42,13 +42,18 @@ public class LockClient {
     @Getter
     private final Map<LockId, Lock> locks = new ConcurrentHashMap<>();
 
-    // lock data store
+    // Lock data store
     private final LockStore lockStore;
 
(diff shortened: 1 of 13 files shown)


utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -85,6 +90,14 @@ public LockClient(UUID clientId, CorfuRuntime corfuRuntime) throws NoSuchMethodE
             return t;
         });
 
+        this.leaseMonitorScheduler = Executors.newScheduledThreadPool(1, (r) ->
+        {
```
pankti-m, line 94: where do we define what gets executed when this thread runs?

annym: You can see that logic in the HasLeaseState class L.196

PR #2687, opened 2020-08-10
Use A Faster JSON Implementation

Maithem:
## Overview

Changed the json implementation from GSON to fastjson because of
performance issues. For example, serializing and deserializing
a 4mb byte array takes on ~914.95 ms with GSON, but only ~58.85ms
with fastjson (i.e. 15x faster).

Why should this be merged: Needed to address stream syncing performance problems that is dominated by serialization/deserialization costs. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
@@ -1,8 +1,12 @@
 package org.corfudb.protocols.wireprotocol;
 
-import com.esotericsoftware.kryo.NotNull;
-import com.google.common.reflect.TypeToken;
+import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINTED_STREAM_ID;
+import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINTED_STREAM_START_LOG_ADDRESS;
+import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINT_ID;
+import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINT_TYPE;
 
+
+import com.google.common.reflect.TypeToken;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumMap;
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/org/corfudb/util/serializer/JsonSerializer.java
```diff
@@ -43,35 +35,22 @@ public byte getType() {
...
-                try (InputStreamReader r = new InputStreamReader(bbis)) {
-                    return gson.fromJson(r, Class.forName(className));
-                }
-            } catch (IOException | ClassNotFoundException ie) {
-                log.error("Exception during deserialization!", ie);
-                throw new RuntimeException(ie);
-            }
+        }
+
+        int jsonStringLen = b.readInt();
```
zhangn49, line 47: Should it be numJsonBytes?

Maithem: Done.

runtime/src/main/java/org/corfudb/util/serializer/JsonSerializer.java
```diff
@@ -84,34 +62,19 @@ public Object deserialize(ByteBuf b, CorfuRuntime rt) {
...
-                try (OutputStreamWriter osw = new OutputStreamWriter(bbos)) {
-                    gson.toJson(o, o.getClass(), osw);
-                }
-            } catch (IOException ie) {
-                log.error("Exception during serialization!", ie);
-            }
+        byte[] classNameBytes = className.getBytes();
+        b.writeShort(classNameBytes.length);
+        b.writeBytes(classNameBytes);
+        if (o == null) {
```
annym, line 68: If we're going to return if object is null, would it make sense to do it ever before computing the classNameBytes?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
```diff
@@ -261,13 +258,16 @@ public byte asByte() {
...
-    @Value
-    @AllArgsConstructor
     class DataRank implements Comparable<DataRank> {
-        public long rank;
-        @NotNull
-        public UUID uuid;
 
+        private long rank;
+
+        private UUID uuid;
```
pankti-m, line 265: do we need to non-null annotation?

runtime/src/main/java/org/corfudb/util/serializer/JsonSerializer.java
```diff
@@ -84,34 +62,19 @@ public Object deserialize(ByteBuf b, CorfuRuntime rt) {
...
+            return;
         }
+        b.markWriterIndex();
+        b.writeInt(0);
+        String jsonOutput= JSON.toJSONString(o);
+        int bytesWritten = b.writeCharSequence(jsonOutput, StandardCharsets.UTF_8);
+        int endIdx = b.writerIndex();
+        b.resetWriterIndex();
+        b.writeInt(bytesWritten);
```
pankti-m, line 77: this will overwrite the short for classNameBytes in line 66, correct?  Is it intended?

hisundar:
Is fastjson compatible with gson? i.e. Can a payload serialized via fastjson serializer be deserialized with gson to get back the same object?
This has a lot of implications for upstream consumers.

Maithem:
Will address the problem in a different approach, in a different patch.

PR #2695, opened 2020-08-10
Register System.exit for systemDownHandler for LR CorfuRuntime 

PavelZaytsev:
## Overview

Description:

Why should this be merged: 

Due to the clustering workflow constraints in the production environment, we have to restart the LR once the WCE is hit. This patch registers systemDownHandler on the CorfuRuntime and calls it every time WCE is hit. This should shut down LR JVM that will subsequently get restarted.

### DIFF
--- a/utils/src/main/java/org/corfudb/utils/lock/states/NoLeaseState.java
+++ b/utils/src/main/java/org/corfudb/utils/lock/states/NoLeaseState.java
@@ -90,7 +90,8 @@ public class NoLeaseState extends LockState {
             if (lockStore.acquire(lock.getLockId())) {
                 lock.input(LockEvent.LEASE_ACQUIRED);
             }
-        } catch (Exception e) {
+        }
+        catch (Exception e) {
             log.error("Lock: {} could not acquire lease {}", lock.getLockId(), e);
         }
     }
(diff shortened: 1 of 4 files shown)


utils/src/main/java/org/corfudb/utils/lock/states/NoLeaseState.java
```diff
@@ -90,7 +91,13 @@ private void acquireLease() {
             if (lockStore.acquire(lock.getLockId())) {
                 lock.input(LockEvent.LEASE_ACQUIRED);
             }
-        } catch (Exception e) {
+        }
+        catch (WrongClusterException wce) {
```
annym, line 95: Can we have a test that verifies that LR is correctly restarted and able to resume?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -160,7 +161,12 @@ private void monitorLocks() {
                             log.debug("LockClient: lease revoked for lock {}", lockId.getLockName());
                             locks.get(lockId).input(LockEvent.LEASE_REVOKED);
                         }
-                    } catch (Exception ex) {
+                    }
+                    catch (WrongClusterException wce) {
+                        log.error("Lock client {} is connected to a wrong cluster:", clientId, wce);
+                        throw wce;
+                    }
+                    catch (Exception ex) {
```
pankti-m, line 169: we should log something here also

xnull: Actually, if we re-throw an exception we must not log it.
In the other case, we will log the same thrown exceptions on all levels of the application, which dramatically increases log size and make it difficult to read and understand the application logic.

Would be great if Pavel removes those logs in `catch` block because they will be handled on a higher level of the stack

PR #2696, opened 2020-08-10
Measure Serialize/Deserialize/VLO_SYNC latency

zhangn49:
## Overview

Description:

Measure Serialize/Deserialize/VLO_SYNC latency

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
@@ -1,11 +1,15 @@
 package org.corfudb.protocols.logprotocol;
 
+import com.codahale.metrics.Histogram;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.util.MetricsUtils;
 import org.corfudb.util.serializer.CorfuSerializer;
 import org.corfudb.util.serializer.Serializers;
 
@@ -38,6 +42,10 @@ public class MultiObjectSMREntry extends LogEntry implements ISMRConsumable {
      * This is required to support lazy stream deserialization.
      */
     private final Map<UUID, byte[]> streamBuffers = new ConcurrentHashMap<>();
+
+    private static final MetricRegistry metricRegistry = CorfuRuntime.getDefaultMetrics();
+
+    private static final String METRIC_PREFIX = "MultiObjectSMREntry-";
 
     public MultiObjectSMREntry() {
         this.type = LogEntryType.MULTIOBJSMR;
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -672,25 +674,32 @@ protected void syncStreamUnsafe(ISMRStream stream, long timestamp) {
...
                 ? "Optimistic" : "to " + timestamp);
         long syncTo = (timestamp == Address.OPTIMISTIC) ? Address.MAX : timestamp;
-        stream.streamUpTo(syncTo)
-                .forEachOrdered(entry -> {
-                    try {
-                        Object res = applyUpdateUnsafe(entry, timestamp);
-                        if (timestamp == Address.OPTIMISTIC) {
+        Timer timer = VloMetricsHelper.metrics.timer(VloMetricsHelper.VLO_SYNC_STREAM + stream.getID());
+        Histogram histogram = VloMetricsHelper.metrics.histogram(VloMetricsHelper.VLO_SYNC_COUNT + stream.getID());
+        AtomicLong counter = new AtomicLong();
```
Maithem, line 679: If you move this string concatenation outside, you don't have to do it on each access.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -672,25 +674,32 @@ protected void syncStreamUnsafe(ISMRStream stream, long timestamp) {
...
                         }
-                        entry.setUpcallResult(res);
-                    } catch (Exception e) {
-                        log.error("Sync[{}] Error: Couldn't execute upcall due to {}", this, e);
-                        throw new UnrecoverableCorfuError(e);
-                    }
-                });
+                    });
+        }
+        histogram.update(counter.longValue());
```
Maithem, line 702: This might produce a bimodal distribution. Since the optimistic sync is much faster (or should be) than the materialized stream sync. 

Is it necessary to split them?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
Maithem: VLO is protected under a lock, so you don't need to use an atomic counter.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -164,7 +166,9 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
     public <R> R access(ICorfuSMRAccess<R, T> accessMethod,
                         Object[] conflictObject) {
         try (Timer.Context context = MetricsUtils.getConditionalContext(timerAccess)) {
-            return accessInner(accessMethod, conflictObject);
+            try (Timer.Context streamContext = MetricsUtils.getConditionalContext(timerAccessPerStream)) {
+                return accessInner(accessMethod, conflictObject);
+            }
```
Maithem, line 171: Shouldn't this be enough? Why is it necessary to have another latency histogram inside the VLO ?

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
zhangn49: Moved string concatenation to the class constructor. Replaced atomic counter by a size 1 array.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
zhangn49: Only measure the regular stream now.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -672,25 +678,34 @@ protected void syncStreamUnsafe(ISMRStream stream, long timestamp) {
...
                 ? "Optimistic" : "to " + timestamp);
         long syncTo = (timestamp == Address.OPTIMISTIC) ? Address.MAX : timestamp;
-        stream.streamUpTo(syncTo)
-                .forEachOrdered(entry -> {
-                    try {
-                        Object res = applyUpdateUnsafe(entry, timestamp);
-                        if (timestamp == Address.OPTIMISTIC) {
+
+        Histogram histogram = VloMetricsHelper.metrics.histogram(syncStreamCount);
+        final long[] counter = new long[1];
```
xnull, line 683: why we need an array here? 
It seems that we need AtomicLong, or (I prefer) just use a plain old for cycle

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -672,25 +678,34 @@ protected void syncStreamUnsafe(ISMRStream stream, long timestamp) {
...
+        try (Timer.Context context = getVloStreamSyncContext(timestamp, syncStreamTimer)) {
+            stream.streamUpTo(syncTo)
```
xnull, line 685: BTW, can't we just use this:
```
val stream = stream.streamUpTo(syncTo);
val counter = stream.size();
...
```
?

zhangn49: yes, I can't do that.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
xnull: Oh, I see the motivation. I believe using an array worse idea that using Atomic. 
If you don't like to use atomics for this, just create a class
```
static class Counter {
    int counter;
}
```
For the sake of clear code should work well for you.
Or you can add that to a common module or util package, up to you

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
zhangn49: If I use a plain for, I have to do `stream.streamUpTo(syncTo).collect(Collectors.toList())`.
Will it bring extra copy?

xnull: yep, looks like, so please take a look at my comment about the counter, would be nice if you agree

zhangn49: Yeah, I am using it right now.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -87,34 +95,53 @@ public void mergeInto(MultiObjectSMREntry other) {
...
+                try (Timer.Context streamContext = MetricsUtils.getConditionalContext(streamDeserializeTimer)) {
+                    int start = b.readerIndex();
+                    MultiSMREntry.seekToEnd(b);
+                    int multiSMRLen = b.readerIndex() - start;
+                    b.readerIndex(start);
+                    byte[] streamUpdates = new byte[multiSMRLen];
+                    b.readBytes(streamUpdates);
+                    streamBuffers.put(streamId, streamUpdates);
+                }
+            }
```
pankti-m, line 119: do you need to catch the exception and log it?

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -87,34 +95,53 @@ public void mergeInto(MultiObjectSMREntry other) {
...
+                        int streamStart = b.writerIndex();
+                        try (Timer.Context streamContext = MetricsUtils.getConditionalContext(streamSerializeTimer)) {
+                            b.writeLong(x.getKey().getMostSignificantBits());
+                            b.writeLong(x.getKey().getLeastSignificantBits());
+                            Serializers.CORFU.serialize(x.getValue(), b);
+                        }
+                        streamSizeHistogram.update(b.writerIndex() - streamStart);
+                    });
+        }
+        sizeHistogram.update(b.writerIndex() - startIdx);
```
pankti-m, line 144: catch required?

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
zhangn49: No, this try block utilizes `try-with-resources`
https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html

So we don't need to close this timer explicitly and it does not affect original exception workflow.

PR #2698, opened 2020-08-11
BaseServer Protobuf RPCs

zfrenette:
## Overview

Description: Server-side groundwork for Protobuf RPC calls. Includes new Request/Response Protobuf types, along with BaseServer RequestHandlers for PING/RESET/RESTART/SEAL. 

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/proto/CorfuProtocol.proto
```diff
@@ -0,0 +1,825 @@
...
+  ProtocolVersion version = 1;
+  int64 request_id = 2;
+  Priority priority = 3;
+  MessageType type = 4;
+  int64 epoch = 5;
+  UUID cluster_id = 6;
+  UUID client_id = 7;
+  bool ignore_cluster_id = 8;
+  bool ignore_epoch = 9;
+}
```
xcchang, line 722: Hi Zach, I'm wondering is there any specific reason we use underscore rather than camelCase for the fields of Header? I noticed that the generated Java code has the same format of getter methods though (like getRequestId() for request_id).

zfrenette: Right now, there's a lack of consistency in our style that we need to address and update, but the recommendation is to use CamelCase for message names and using an underscore format for field names: https://confluence.eng.vmware.com/pages/viewpage.action?spaceKey=NSBU&title=NSBU+Protobuf+Conventions+and+Style+Guide and https://developers.google.com/protocol-buffers/docs/style

I haven't had a chance to make the necessary updates, but was thinking of making these changes when we incorporate Vjeko's feedback on the schema.

xcchang: Gotcha, thanks!

runtime/proto/CorfuProtocol.proto
```diff
@@ -0,0 +1,825 @@
...
+
+option java_package = "org.corfudb.runtime.protocol.proto";
+
+/**
+ * Common Definitions
+ **/
+
+enum ProtocolVersion {
+  v0 = 0;
+}
```
vjeko, line 15: I am not really sure what the best strategy is for dealing with message versioning (due to a breaking change).

- One way would be to incorporate the version into the name -- i. e. GetLayoutRequestV1.
- We could also introduce a name-spacing concept. I know this is possible with gRPC, but it might not be possible (or too cumbersome) with plain Protobuf definitions.

runtime/proto/log_unit.proto
```diff
@@ -0,0 +1,150 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
zfrenette: By "all externally facing API messages", do you mean those that correspond directly to an RPC request/response type?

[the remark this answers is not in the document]
vjeko: Yep.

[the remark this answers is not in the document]
chetangudisagar: Hi @vjeko , including all the API based messages into a new file called `service.proto` and all their dependencies into a `helper.proto` file would again make these 2 new files very big (approximately half of the size of `CorfuProtocol.proto` file).

Did you mean clustering/moving the current API layer-based files like `log_unit.proto`, `sequencer.proto`, etc into a new folder `runtime/proto/service` and package `org.corfudb.runtime.proto.service`? This makes a lot of sense but it would leave out only non-API ones out. Should we move them into a `runtime/proto/common` folder or leave them as is outside in `runtime/proto`?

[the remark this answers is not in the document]
vjeko: I was referring to the later approach. I would personally leave them in `runtime/proto`. Generally, concepts like `common` should be avoided: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

runtime/proto/common.proto
```diff
@@ -0,0 +1,77 @@
...
+}
+
+//TODO: Consider alternate representations of a Layout?
+message LayoutMsg {
+  string layout_json = 1;
+}
+
+//TODO: Why do we need NO_ADDRESSES?
+message StreamAddressSpaceMsg {
+  int32 NO_ADDRESSES = 1;
```
zfrenette, line 48: @chetangudisagar Do we need NO_ADDRESSES? It seems like this isn't sent over the wire in the original implementation.

runtime/proto/log_unit.proto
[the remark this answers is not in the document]
zfrenette: Did my best to restructure while keeping this in mind. Let me know if there is any other feedback

runtime/src/main/java/org/corfudb/protocols/API.java
```diff
@@ -0,0 +1,1113 @@
+package org.corfudb.protocols;
```
vjeko, line 1: Can we please split this file as well? It's more than 1000 lines long. We can create a dedicated package for APIs, and have one file for each service.

infrastructure/src/main/java/org/corfudb/infrastructure/NettyRequestRouter.java
```diff
@@ -0,0 +1,164 @@
...
+import java.io.IOException;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+@Slf4j
+@ChannelHandler.Sharable
+public class NettyRequestRouter extends ChannelInboundHandlerAdapter implements IRequestRouter {
```
vjeko, line 31: Do we actually need a separate class for this? Is there a reason why we did not just add the following two functions to `IClientRouter`?

```
    <T> CompletableFuture<T> sendRequestAndGetCompletable(CorfuProtocol.Request.Builder reqBuilder);
    void sendRequest(CorfuProtocol.Request.Builder reqBuilder);
```

runtime/src/main/java/org/corfudb/protocols/API.java
zfrenette: Definitely. That's next on my ToDo list.

infrastructure/src/main/java/org/corfudb/infrastructure/NettyRequestRouter.java
zfrenette: This is for the server. I wanted to do this initially, but we also require a new implementation of channelRead. This became tricky because I have two router instances in my pipeline, as otherwise, we would require it to process both types of messages. Maybe as a workaround, we can combine the functionality of NettyCorfuMsgEncoder and NettyCorfuMsgDecoder into the router, so that we can simply worry about sending and receiving a ByteBuf? I am not sure If this has any other implications... I'll take a look more carefully and see what can be done.

infrastructure/src/main/java/org/corfudb/infrastructure/RequestHandlerMethods.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /** Handle an incoming Corfu request message.
+     *
+     * @param req       The request message to handle.
+     * @param ctx       The channel handler context.
+     * @param r         The server router.
+     */
+    @SuppressWarnings("unchecked")
+    public void handle(Request req, ChannelHandlerContext ctx, IRequestRouter r) {
+        final HandlerMethod handler = handlerMap.get(req.getHeader().getType());
```
vjeko, line 67: Having the type in the header seems redundant. Would it be possible to bootstrap the handlers this way?


```
public @interface RequestHandler {
    /**
     * Returns the message type
     * @return the type of Corfu message request
     */
    CorfuMessage.RequestMsg.PayloadCase type();
}

@RequestHandler(type = CorfuMessage.RequestMsg.PayloadCase.AUTHENTICATE_REQUEST)

final HandlerMethod handler = handlerMap.get(req.getPayloadCase());

File: infrastructure/src/main/java/org/corfudb/infrastructure/NettyRequestRouter.java [hunk 8]
```diff
@@ -0,0 +1,164 @@
...
+            responseOutputStream.writeByte(API.PROTO_CORFU_MSG_MARK);
+            response.writeTo(responseOutputStream);
+            ctx.writeAndFlush(outBuf, ctx.voidPromise());
+        } catch(IOException e) {
+            log.warn("sendResponse[{}]: Exception occurred when sending response {}, caused by {}",
+                    response.getHeader().getRequestId(), response.getHeader(), e.getCause(), e);
+        } finally {
+            IOUtils.closeQuietly(responseOutputStream);
+        }
+    }
```
vjeko, line 94: I would move this functionality into the `AbstractServer`. `ChannelInboundHandlerAdapter` should not concern itself with message serialization.

runtime/src/main/java/org/corfudb/protocols/API.java
```diff
@@ -0,0 +1,1113 @@
...
+
+    public static Response getErrorResponseNoPayload(Header header, ServerError error) {
+        return Response.newBuilder()
+                .setHeader(header)
+                .setError(error)
+                .build();
+    }
+
+    //TODO(Zach): Improve this method
+    public static boolean validateRequestPayloadType(Request request) {
```
zfrenette, line 1037: This is no longer needed with new Protobuf request format.

runtime/proto/service/sequencer.proto
```diff
@@ -6,6 +6,43 @@ option java_package = "org.corfudb.runtime.proto.service";
...
+
+/**
+  proto3 does not support non integer or string keys.
+  This message contains a map of UUID to an array of bytes.
+  The bytes array can be further parsed into a Set of bytes.
+ */
+message UUIDToListOfBytesPairMsg {
+  UuidMsg key = 1;
+  repeated bytes value = 2;
+}
```
zfrenette, line 30: Why did you add this here? This is defined in common.proto

runtime/proto/service/sequencer.proto
```diff
@@ -21,56 +58,93 @@ message TokenRequestMsg {
...
+/**
+  proto3 does not support non integer or string keys.
+  This message contains a map of UUID to an array of bytes.
+  The bytes array can be further parsed into a Set of bytes.
+ */
+message UUIDToLongPairMsg {
+  UuidMsg key = 1;
+  int64 value = 2;
+}
+
```
zfrenette, line 70: Defined in common.proto

runtime/proto/common.proto
```diff
@@ -43,16 +50,37 @@ message LayoutMsg {
...
+ *
+ * A stream's address space is defined by:
+ *       1. The collection of all addresses that belong to this stream.
+ *       2. The trim mark (last trimmed address, i.e., an address that is no longer present and that was subsumed by
+ *       a checkpoint).
+ */
 message StreamAddressSpaceMsg {
-  int32 NO_ADDRESSES = 1;
+  //TODO: Why do we need NO_ADDRESSES?
+  int32 NO_ADDRESSES = 1; // For numeric types, the default value is zero.
```
zfrenette, line 63: Why do we need this?

runtime/proto/common.proto
```diff
@@ -0,0 +1,105 @@
...
+enum ProtocolVersion {
+  PROTOCOL_VERSION_INVALID = 0;
+  V0 = 1;
+}
+
+enum PriorityLevel {
+  PRIORITY_LEVEL_INVALID = 0;
+  NORMAL = 1;
+  HIGH = 2;
+}
```
xcchang, line 18: I noticed that when running `protoc`, the declaration of `NORMAL` and `HIGH` will be conflicted with those declared in `messages.proto`. 

Like `org.corfudb.runtime.HIGH already defined in common.proto`

 I think it's because `common.proto` and `messages.proto` are under the same package, although the generated Java source files belongs to different packages. 

Would this be a problem for us?

runtime/src/main/java/org/corfudb/protocols/API.java
```diff
@@ -0,0 +1,1135 @@
...
+        return Response.newBuilder()
+                .setHeader(header)
+                .setError(error)
+                .setSealResponse(sealResponse)
+                .build();
+    }
+
+    public static CorfuMessage.ResponseMsg getVersionResponse(CorfuMessage.HeaderMsg headerMsg, VersionInfo vi){
+        final Gson parser = new GsonBuilder().create();
+        String payload = parser.toJson(vi);
```
xcchang, line 224: Is there any specific reason that we don't use `final Gson parser = new Gson()` here? I also noticed this patter in the current `JSONPayloadMsg` definition.

PR #2699, opened 2020-08-11
Remove Java Generics From CorfuQueue's API

Maithem:
Since this queue is designed to be used with protobufs, this patch
removes Java generics from the API and uses ProtoBuf's ByteString
instead. Fixing the queue type will allow us to serialize the queues
SMREntries much more efficiently.

The benchmarks below show how different serializers perform
with different payload sizes.

4MB Payload
Benchmark                           Mode  Cnt    Score    Error  Units
JsonBenchmark.GoogleGson           thrpt    5    0.818 ±  0.157  ops/s
JsonBenchmark.ObjectOutputStream   thrpt    5  336.744 ± 54.049  ops/s
JsonBenchmark.QueueSerializer      thrpt    5  358.410 ±  9.058  ops/s

8MB Payload
Benchmark                           Mode  Cnt    Score    Error  Units
JsonBenchmark.kryoSerializer       thrpt    5  116.545 ± 72.152  ops/s
JsonBenchmark.ObjectOutputStream   thrpt    5  135.228 ± 78.692  ops/s
JsonBenchmark.QueueSerializer      thrpt    5  163.246 ± 27.071  ops/s

16MB Payload
Benchmark                           Mode  Cnt    Score    Error  Units
JsonBenchmark.googleGson           thrpt    5   0.211 ±  0.013  ops/s
JsonBenchmark.kryoSerializer       thrpt    5   38.168 ±  8.253  ops/s
JsonBenchmark.ObjectOutputStream   thrpt    5   51.139 ± 60.820  ops/s
JsonBenchmark.QueueSerializer      thrpt    5  108.701 ± 11.039  ops/s

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
@@ -2,9 +2,12 @@ package org.corfudb.runtime.collections;
 
 import static com.google.common.base.Preconditions.checkState;
 
-
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.reflect.TypeToken;
+import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
+
+import io.netty.buffer.ByteBuf;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
@@ -23,6 +26,7 @@ import org.corfudb.runtime.Queue.CorfuQueueIdMsg;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
 import org.corfudb.runtime.object.transactions.TransactionalContext.PreCommitListener;
 import org.corfudb.runtime.view.CorfuGuidGenerator;
+import org.corfudb.util.serializer.ICorfuHashable;
 import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
 
@@ -39,32 +43,30 @@ import org.corfudb.util.serializer.Serializers;
  *
  * Created by hisundar on 5/8/19.
  *
- * @param <E>   Type of the entry to be enqueued into the persisted queue
  */
 @Slf4j
-public class CorfuQueue<E> {
+public class CorfuQueue {
     /**
      * The main CorfuTable which contains the primary key-value mappings.
      */
-    private final CorfuTable<CorfuRecordId, E> corfuTable;
+    private final CorfuTable<CorfuRecordId, ByteString> corfuTable;
     private final CorfuGuidGenerator guidGenerator;
 
-    public CorfuQueue(CorfuRuntime runtime, String streamName, ISerializer serializer,
-                      Index.Registry<CorfuRecordId, E> indices) {
-        final Supplier<StreamingMap<CorfuRecordId, E>> mapSupplier =
-                () -> new StreamingMapDecorator<>(new LinkedHashMap<CorfuRecordId, E>());
+    @VisibleForTesting
+    CorfuQueue(CorfuRuntime runtime, String streamName, ISerializer serializer) {
+        final Supplier<StreamingMap<CorfuRecordId, ByteString>> mapSupplier =
+                () -> new StreamingMapDecorator<>(new LinkedHashMap<CorfuRecordId, ByteString>());
         corfuTable = runtime.getObjectsView().build()
-                .setTypeToken(new TypeToken<CorfuTable<CorfuRecordId, E>>() {})
+                .setTypeToken(new TypeToken<CorfuTable<CorfuRecordId, ByteString>>() {})
                 .setStreamName(streamName)
-                .setArguments(indices, mapSupplier)
+                .setArguments(Index.Registry.empty(), mapSupplier)
                 .setSerializer(serializer)
                 .open();
         guidGenerator = CorfuGuidGenerator.getInstance(runtime);
     }
 
     public CorfuQueue(CorfuRuntime runtime, String streamName) {
-        this(runtime, streamName, Serializers.getDefaultSerializer(), Index.Registry.empty());
-
+        this(runtime, streamName, Serializers.QUEUE_SERIALIZER);
     }
 
     /**
@@ -81,7 +83,7 @@ public class CorfuQueue<E> {
      * were in wrapped a corfu transaction.
      * This class encapsulates these two longs into one Id and add rules on comparability.
      */
-    public static class CorfuRecordId implements Comparable<CorfuRecordId> {
+    public static class CorfuRecordId implements Comparable<CorfuRecordId>, ICorfuHashable {
         @Setter
         @Getter
         private long txSequence;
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -36,31 +37,30 @@
...
+    private final CorfuTable<CorfuRecordId, ByteString> corfuTable;
     private final CorfuGuidGenerator guidGenerator;
 
-    public CorfuQueue(CorfuRuntime runtime, String streamName, ISerializer serializer,
-                      Index.Registry<CorfuRecordId, E> indices) {
-        final Supplier<StreamingMap<CorfuRecordId, E>> mapSupplier =
-                () -> new StreamingMapDecorator<>(new LinkedHashMap<CorfuRecordId, E>());
+    public CorfuQueue(CorfuRuntime runtime, String streamName,
+                      Index.Registry<CorfuRecordId, ByteString> indices) {
+        final Supplier<StreamingMap<CorfuRecordId, ByteString>> mapSupplier =
```
Maithem, line 51: Make private and remove Secondary indexes.

runtime/src/main/java/org/corfudb/util/serializer/CorfuQueueSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+public class CorfuQueueSerializer implements ISerializer {
+
+    private final byte type;
+
+    public CorfuQueueSerializer(byte type) {
+        this.type = type;
+    }
+
+    // Serialization tag for a queue's metadata entry
+    private final byte entryRecordIdMarker = 1;
```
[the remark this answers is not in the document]
Maithem: Truuuu

runtime/src/main/java/org/corfudb/util/serializer/CorfuQueueSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+
+    // Serialization tag for the queue's payload
+    private final byte entryPayloadMarker = 2;
```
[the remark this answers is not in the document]
Maithem: Truuuu

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -36,32 +39,30 @@
...
+                () -> new StreamingMapDecorator<>(new LinkedHashMap<CorfuRecordId, ByteString>());
         corfuTable = runtime.getObjectsView().build()
-                .setTypeToken(new TypeToken<CorfuTable<CorfuRecordId, E>>() {})
+                .setTypeToken(new TypeToken<CorfuTable<CorfuRecordId, ByteString>>() {})
                 .setStreamName(streamName)
-                .setArguments(indices, mapSupplier)
+                .setArguments(Index.Registry.empty(), mapSupplier)
                 .setSerializer(serializer)
                 .open();
```
Maithem, line 60: Add support to add CorfuQueue to the `MultiCheckpointWriter`

Maithem: Done.

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
```diff
@@ -105,9 +117,133 @@ public String toString() {
...
+
+        CorfuRuntime rt2 = getNewRuntime(getDefaultNode()).connect();
+        CorfuQueue oldQueueInstance2 = new CorfuQueue(rt2, "test",
+                Serializers.QUEUE_SERIALIZER);
+
+        // produce items to the queue with two different serializers from two different clients
+        final int numItemsToProduce = 10;
+        IntStream.range(0, numItemsToProduce).forEach(itemIdx -> {
+            if (itemIdx % 2 == 0) {
+                oldQueueInstance1.enqueue(getByteString(String.valueOf(itemIdx)));
```
hisundar, line 137: just wondering for code coverage of the sequence and id part can this test case wrap the enqueue in a transaction?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -135,6 +147,24 @@ public boolean equals(Object o) {
             return getEntryId() == that.getEntryId();
         }
 
+        @Override
+        public byte[] generateCorfuHash() {
+            // Note that this hash is used for transaction conflict resolution and shouldn't
+            // generate collisions for different entries
+            // TODO(Maithem) this probably needs the sequence number to
```
hisundar, line 154: No you are right - the sequence should not be included here. perhaps you can remove the TODO?

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
Maithem: I think there's a bug. When I wrap oldQueueInstance1 with a transaction, but not oldQueueInstance2, then the read order is flipped. If there is no transaction, then how is addPreCommitListener called?

hisundar: it won't be called - we can't support a mixed case like that anyway

Maithem: Why not?

If it's not allowed then the operation should be failed.

Maithem: Also, most of the tests don't use transactions to produce to the queue, so I'm not sure why this isnt allowed.

Maithem: Even if we don't mix transactional and non-transaction. Just having non-transactional enqueues will produce an incorrect order.

hisundar: so you suggest we just throw error if queue is used outside a transaction?

Maithem: Yea, it won't make sense to allow it if its incorrect. #2712

runtime/src/main/java/org/corfudb/util/serializer/CorfuQueueSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+    public byte getType() {
+        return type;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Object deserialize(ByteBuf b, CorfuRuntime rt) {
+        byte type = b.readByte();
```
zhangn49, line 38: maybe rename to tag? It has the same name as the serializer type.

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
```diff
@@ -105,9 +117,136 @@ public String toString() {
...
+
+        // Produce to two different queues using the old and new serializer
+        CorfuQueue queueInstance1 = new CorfuQueue(getDefaultRuntime(), "stream1",
+                Serializers.JSON);
+
+        CorfuQueue queueInstance2 = new CorfuQueue(getDefaultRuntime(), "stream2");
+
+        final int numItemsToProduce = 10;
+        IntStream.range(0, numItemsToProduce).forEach(itemIdx -> {
+            queueInstance1.enqueue(getByteString(String.valueOf(itemIdx)));
```
zhangn49, line 202: Will this test fail after #2712 ?

PR #2701, opened 2020-08-11
CorfuStoreBrowser: variable size records for loading data

hisundar:
Use JsonFormat to display for future editing

## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -3,13 +3,16 @@ package org.corfudb.browser;
 import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.UUID;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
+import com.google.protobuf.InvalidProtocolBufferException;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuStoreMetadata.TableName;
@@ -28,6 +31,8 @@ import org.corfudb.util.serializer.DynamicProtobufSerializer;
 import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
 import org.rocksdb.Options;
+
+import com.google.protobuf.util.JsonFormat;
 
 /**
  * This is the CorfuStore Browser Tool which prints data in a given namespace and table.
@@ -133,11 +138,18 @@ public class CorfuStoreBrowser {
                 Iterables.partition(entryStream::iterator, batchSize);
         for (List<Map.Entry<CorfuDynamicKey, CorfuDynamicRecord>> partition : partitions) {
             for (Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry : partition) {
-                builder = new StringBuilder("\nKey:\n" + entry.getKey().getKey())
-                        .append("\nPayload:\n").append(entry.getValue().getPayload())
-                        .append("\nMetadata:\n").append(entry.getValue().getMetadata())
-                        .append("\n====================\n");
-                log.info(builder.toString());
+                try {
+                    builder = new StringBuilder("\nKey:\n")
+                            .append(JsonFormat.printer().print(entry.getKey().getKey()))
+                            .append("\nPayload:\n")
+                            .append(JsonFormat.printer().print(entry.getValue().getPayload()))
+                            .append("\nMetadata:\n")
+                            .append(JsonFormat.printer().print(entry.getValue().getMetadata()))
+                            .append("\n====================\n");
+                    log.info(builder.toString());
+                } catch (InvalidProtocolBufferException e) {
+                    log.error("invalid protobuf: ", e);
+                }
             }
         }
         return size;
@@ -209,9 +221,10 @@ public class CorfuStoreBrowser {
      * @param tablename - table name without the namespace
      * @param numItems - total number of items to load
      * @param batchSize - number of items in each transaction
+     * @param itemSize - size of each item - a random string array
      * @return - number of entries in the table
      */
-    public int loadTable(String namespace, String tablename, int numItems, int batchSize) {
+    public int loadTable(String namespace, String tablename, int numItems, int batchSize, int itemSize) {
         verifyNamespaceAndTablename(namespace, tablename);
         CorfuStore store = new CorfuStore(runtime);
         try {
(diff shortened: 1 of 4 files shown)


corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -225,9 +238,12 @@ public int loadTable(String namespace, String tablename, int numItems, int batch
                     TableName.getDefaultInstance().getClass(),
                     optionsBuilder.build());
 
-            TableName dummyVal = TableName.newBuilder().setNamespace(namespace).setTableName(tablename).build();
-            log.info("Loading {} items in {} batchSized transactions into {}${}",
-                    numItems, batchSize, namespace, tablename);
+            byte[] array = new byte[itemSize];
+            new Random().nextBytes(array);
```
xnull, line 242: why we need this?

hisundar: if we don't throw in random bytes, the compression reduces the actual payload size making it hard to test things like disk full

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -133,11 +138,18 @@ public int printTable(String namespace, String tablename) {
...
+                    builder = new StringBuilder("\nKey:\n")
+                            .append(JsonFormat.printer().print(entry.getKey().getKey()))
+                            .append("\nPayload:\n")
+                            .append(JsonFormat.printer().print(entry.getValue().getPayload()))
+                            .append("\nMetadata:\n")
+                            .append(JsonFormat.printer().print(entry.getValue().getMetadata()))
+                            .append("\n====================\n");
+                    log.info(builder.toString());
+                } catch (InvalidProtocolBufferException e) {
+                    log.error("invalid protobuf: ", e);
```
pankti-m, line 151: can we print the protobuf type which caused the exception?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: how does adding random bytes solve it?  can we add a comment about this?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -209,9 +221,10 @@ public int dropTable(String namespace, String tablename) {
      * @param tablename - table name without the namespace
      * @param numItems - total number of items to load
      * @param batchSize - number of items in each transaction
+     * @param itemSize - size of each item - a random string array
      * @return - number of entries in the table
      */
-    public int loadTable(String namespace, String tablename, int numItems, int batchSize) {
+    public int loadTable(String namespace, String tablename, int numItems, int batchSize, int itemSize) {
```
pankti-m, line 227: I did not realize earlier, but in this operation we allow data to be loaded only in the schema of dummyKey and dummyVal, but the clients table can have a different schema, right?  Wont the load operation fail?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -225,9 +238,12 @@ public int loadTable(String namespace, String tablename, int numItems, int batch
                     TableName.getDefaultInstance().getClass(),
                     optionsBuilder.build());
 
-            TableName dummyVal = TableName.newBuilder().setNamespace(namespace).setTableName(tablename).build();
-            log.info("Loading {} items in {} batchSized transactions into {}${}",
-                    numItems, batchSize, namespace, tablename);
+            byte[] array = new byte[itemSize];
+            new Random().nextBytes(array);
+            TableName dummyVal = TableName.newBuilder().setNamespace(namespace+tablename)
```
pankti-m, line 243: why namespace+tablename?

hisundar: just to display something meaningful in case someone is silly enough to dump the dummy table :)

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: this is not meant to load data into client tables - if they do that they are intentionally corrupting data. This tool is just to load specific data into corfu.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: in this case there is only one protobuf - the TableName, also I think the error would carry the line where it failed. if it does.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: adding random bytes prevents compression from reducing the size of the written value. random bytes don't compress well.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: sorry, but arent the keys and values the protobufs?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: I see, ok.  what is the use case of this operation?  is it for testing disk full like you mentioned in another comment?  Should we add it in the comments section for this method?  

Also, if accidentally, this is called for a client table, the operation will fail with incompatible schema, right?  If not, is it better to control from Corfu itself which table the data gets loaded into?  It can be a fixed table inside CorfuSystem or something.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: yes same protobuf TableName is being used as key and also as value

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: comments added.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: its for disk full and load testing. added comments to state the same.

pankti-m: dint find the comments...

pankti-m (approved):

PR #2702, opened 2020-08-12
If layoutHelper returns a WrongClusterException best to shutdown

hisundar:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -1140,6 +1140,7 @@ public class CorfuRuntime {
                     } catch (WrongClusterException we) {
                         // It is futile trying to re-connect to the wrong cluster
                         log.warn("Giving up since cluster is incorrect or reconfigured!");
+                        parameters.getSystemDownHandler().run();
                         throw we;
                     } catch (ExecutionException ee){
                         if (ee.getCause() instanceof TimeoutException) {
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -1140,6 +1140,8 @@ public class CorfuRuntime {
                     } catch (WrongClusterException we) {
                         // It is futile trying to re-connect to the wrong cluster
                         log.warn("Giving up since cluster is incorrect or reconfigured!");
+                        log.info("fetchLayout: Invoking the systemDownHandler.");
+                        parameters.getSystemDownHandler().run();
                         throw we;
                     } catch (ExecutionException ee){
                         if (ee.getCause() instanceof TimeoutException) {
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -1140,7 +1140,6 @@ public class CorfuRuntime {
                     } catch (WrongClusterException we) {
                         // It is futile trying to re-connect to the wrong cluster
                         log.warn("Giving up since cluster is incorrect or reconfigured!");
-                        parameters.getSystemDownHandler().run();
                         throw we;
                     } catch (ExecutionException ee){
                         if (ee.getCause() instanceof TimeoutException) {
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -151,6 +151,8 @@ public abstract class AbstractView {
                     log.warn("layoutHelper: System seems unavailable", re);
                 } else if (re instanceof WrongClusterException) {
                     log.warn("layoutHelper: Cluster reconfiguration or incorrect cluster", re);
+                    log.info("layoutHelper: Invoking the systemDownHandler.");
+                    runtime.getParameters().getSystemDownHandler().run();
                     throw re;
                 }else {
                     throw re;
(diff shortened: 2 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -1140,6 +1140,7 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
                     } catch (WrongClusterException we) {
                         // It is futile trying to re-connect to the wrong cluster
                         log.warn("Giving up since cluster is incorrect or reconfigured!");
+                        parameters.getSystemDownHandler().run();
                         throw we;
```
Maithem, line 1144: Should the exception still be thrown after the handler is called ?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -1140,6 +1140,7 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
                     } catch (WrongClusterException we) {
                         // It is futile trying to re-connect to the wrong cluster
                         log.warn("Giving up since cluster is incorrect or reconfigured!");
+                        parameters.getSystemDownHandler().run();
```
WenbinZhu, line 1143: Should we change the layoutHelper instead of here? Here it's fetchLayout, but other operations can still hit WCE

annym: +1

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
hisundar: the reason for throwing the exception is some shutdown handlers may just be no-ops (by intent), in which case we want to still fail the operation.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
hisundar: will do. thanks

hisundar: done. thanks

WenbinZhu: Looks like sometimes this method, fetchLayout, is called inside LayoutHelper, so we will invoke SystemDownHandler right? Will this be a problem?

hisundar: systemDownHandler are supposed to exit the JVM, so not unless there is another bug right?

WenbinZhu: @hisundar Actually I feel this catch block will never be executed. The exceptions should come from layoutFuture.get(), but that will be wrapped in ExecutionException, which we check in the next catch block, right?

PavelZaytsev: @annym @WenbinZhu Since Sundar is out, I fixed it in two places where the issue might happen. 
One way to get WCE is to invalidateLayout() after our local current node has joined another cluster and it had an identity (clusterId) before that. That WCE is hit when we call invalidateLayout() inside a layoutHelper. checkClusterId inside fetchLayout will throw WrongClusterException that will not be handled, so the runtime layout future will complete with ExecutionException with the cause of the WrongClusterException. Then when in layoutHelper we call getLayoutUninterruptibly() and we call get on the this layout future, we will hit ExecutionException with the the cause of WrongClusterException. Here we should call a systemDownHandler and then propagate this exception (since downhandler might be a no-op). 
Another way to get WCE is in function.apply in the layoutHelper. This can happen if the function us an RPC and the client message was stamped with one clusterId but the server router on the remote node has another clusterId. Here we should also call a systemDownHandler and propagate the exception.

WenbinZhu: @PavelZaytsev invalidateLayout() can be called outside of AbstractView, which means if users use runtime.invalidateLayout() directly, then SystemDownHandler will not be trigger, because we don't trigger it in fetchLayout. So I think the  invocation of SystemDownHandler should be put into CorfuRuntime::fetchLayout instead of AbstractView::getLayoutUninterruptibly(), and also we already have SystemDownHandler invocation logic in fetchLayout()

PavelZaytsev: @Ok. I will call it in fetchLayout then every client would have to call system down handler. We still need to call it in the layoutHelper for the RPC case though.  I removed it from getLayoutUninterruptibly, so it will just get propagated as ExecutionException with a RuntimeException cause after the systemDownHandler is called.

WenbinZhu: @PavelZaytsev Thanks, but I think there is another problem as I mentioned in the initial commit.  Looks like `catch (WrongClusterException we)` will never be reached (a bug in the existing code), because execption is wrapped in the ExecutionException. I think this exception handling block should be moved to the `catch (ExecutionException ee) ` block below, otherwise it will not work.

PavelZaytsev: @WenbinZhu catch (WrongClusterException we) is actually getting caught right now and is getting propagated because it is coming from checkClusterId, so it is not wrapped into ExecutionException. You are probably referring to layoutFuture.get() that is being called after getLayout(). But this one can not fail with a WrongClusterException because it is a LAYOUT_REQUEST, and we ignore a clusterId check for it.

```
LAYOUT_REQUEST(10, new TypeToken<CorfuPayloadMsg<Long>>(){}, true, true)
```

WenbinZhu: I see

PavelZaytsev:
@Maithem Not sure I understand. Do you want a separate runtime handler for WCE? How would this help us?

Issue #1949, opened 2019-06-17
LogData payload data is copied 3 times before sending to log unit server.


WenbinZhu:
## Overview

1. First time: `checkMaxWriteSize` opens a `SerializationHandle` that calls `acquireBuffer` and do serialization which copies serialized data into `serializedCache`. After `checkMaxWriteSize` is finished, it closes the `SerializationHandle` which release the `serializedCache` and reassign it to NULL

2. Second time: `ChainReplicationProtocol` write() also opens a `SerializationHandle` and call `acquireBuffer`, since `serializedCache` is assigned to NULL at the end of previous step, it will do serialize again, and copies to `serializedCache`.

3. Third time: `ChainReplicationProtocol` calls `LogUnitClient` write(), which constructs  a `WriteRequest`, when serializing the request, it will call `LogData::doSerialize`. This time the `SerializationHandle` is not released yet, so `serializedCache` is still valid, but still needs to be copied to the Netty buffer.

We should reduce the number of copies as this is a considerable performance overhead.

Maithem:
I don't think the payload is serialized multiple times, you can see this simply by adding this

```java

     void doSerializeInternal(ByteBuf buf) {
+        log.info("Serializing1 at {}", getGlobalAddress());
         ICorfuPayload.serialize(buf, type);
         if (type == DataType.DATA) {
             if (data == null) {
```

On a 3 node cluster, this is the output:

3338 [pool-5-thread-1] INFO  org.corfudb.protocols.wireprotocol.LogData  - Serializing1 at 58
3346 [pool-5-thread-1] INFO  org.corfudb.protocols.wireprotocol.LogData  - Serializing1 at 59
3353 [pool-5-thread-1] INFO  org.corfudb.protocols.wireprotocol.LogData  - Serializing1 at 60
3361 [pool-5-thread-1] INFO  org.corfudb.protocols.wireprotocol.LogData  - Serializing1 at 61
3371 [pool-5-thread-1] INFO  org.corfudb.protocols.wireprotocol.LogData  - Serializing1 at 62
3380 [pool-5-thread-1] INFO  org.corfudb.protocols.wireprotocol.LogData  - Serializing1 at 63

WenbinZhu:
@Maithem I just reproduced with a log inserted in `doSerializeInternal`(same as your example). Based on my test code, it is calling `doSerializeInternal` twice:

```
[...]

PR #2703, opened 2020-08-12
Stream write path performance optimization.

WenbinZhu:
## Overview

Description:

- Reuse serialization handle, which reduces number of serializations.
- Acquire token after payload serialized, which reduces reader wait time.

Related issue(s) (if applicable): #1949 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ILogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/ILogData.java
@@ -1,6 +1,5 @@
 package org.corfudb.protocols.wireprotocol;
 
-import org.corfudb.common.compression.Codec;
 import org.corfudb.protocols.logprotocol.LogEntry;
 import org.corfudb.runtime.CorfuRuntime;
 
@@ -11,7 +10,7 @@ import java.util.UUID;
  * Log data entries represent data stored in the actual log,
  * with convenience methods for software to retrieve the
  * stored information.
- *
+ * <p>
  * Created by mwei on 8/17/16.
  */
 public interface ILogData extends IMetadata, Comparable<ILogData> {
@@ -26,9 +25,8 @@ public interface ILogData extends IMetadata, Comparable<ILogData> {
     }
 
     /**
-     * This class provides a serialization handle, which
-     * manages the lifetime of the serialized copy of this
-     * entry.
+     * This class provides a serialization handle, which manages
+     * the lifetime of the serialized copy of this entry.
      */
     class SerializationHandle implements AutoCloseable {
 
@@ -41,7 +39,7 @@ public interface ILogData extends IMetadata, Comparable<ILogData> {
          * Explicitly request the serialized form of this log data
          * which only exists for the lifetime of this handle.
          *
-         * @return The serialized form of this handle.
+         * @return the serialized form of this handle
          */
         public ILogData getSerialized() {
             return data;
@@ -51,11 +49,12 @@ public interface ILogData extends IMetadata, Comparable<ILogData> {
          * Create a new serialized handle with a reference
          * to the log data.
          *
-         * @param data The log data to manage.
+         * @param data     the log data to manage
+         * @param metadata whether metadata needs to be serialized
          */
-        public SerializationHandle(ILogData data) {
+        public SerializationHandle(ILogData data, boolean metadata) {
             this.data = data;
-            data.acquireBuffer();
+            data.acquireBuffer(metadata);
         }
 
         /**
@@ -67,13 +66,28 @@ public interface ILogData extends IMetadata, Comparable<ILogData> {
         }
     }
 
+    /**
+     * Get the serialization handle of this entry that manages
+     * the lifetime of the serialized copy.
+     *
+     * @param metadata whether metadata needs to be serialized
+     * @return a serialization handle of this entry
+     */
+    default SerializationHandle getSerializedForm(boolean metadata) {
+        return new SerializationHandle(this, metadata);
+    }
+
+    /**
+     * Release the serialization buffer.
+     */
     void releaseBuffer();
 
-    void acquireBuffer();
-
-    default SerializationHandle getSerializedForm() {
-        return new SerializationHandle(this);
-    }
+    /**
+     * Acquire the serialization buffer.
+     *
+     * @param metadata whether metadata needs to be serialized
+     */
+    void acquireBuffer(boolean metadata);
 
     /**
      * Return the payload as a log entry.
@@ -138,20 +152,8 @@ public interface ILogData extends IMetadata, Comparable<ILogData> {
     /**
      * Return true if this LogData contains data
      */
-    default boolean isData() { return  getType() == DataType.DATA; }
-
-    /**
-     * Return the serialized size of an object
-     *
-     * @param obj the entry's payload object
-     * @return size of serialized buffer
-     */
-    static int getSerializedSize(Object obj, Codec.Type codecType) {
-        ILogData ld = new LogData(DataType.DATA, obj, codecType);
-        ld.acquireBuffer();
-        int size = ld.getSizeEstimate();
-        ld.releaseBuffer();
-        return size;
+    default boolean isData() {
+        return getType() == DataType.DATA;
     }
 
     default void setId(UUID clientId) {
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
@@ -2,7 +2,16 @@ package org.corfudb.protocols.wireprotocol;
 
 import com.esotericsoftware.kryo.NotNull;
 import com.google.common.reflect.TypeToken;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+import org.corfudb.common.compression.Codec;
+import org.corfudb.protocols.logprotocol.CheckpointEntry;
+import org.corfudb.runtime.view.Address;
+import org.corfudb.runtime.view.Layout;
 
+import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumMap;
@@ -13,18 +22,6 @@ import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import javax.annotation.Nullable;
-
-import lombok.AllArgsConstructor;
-import lombok.Getter;
-import lombok.RequiredArgsConstructor;
-import lombok.Value;
-
-import org.corfudb.common.compression.Codec;
-import org.corfudb.protocols.logprotocol.CheckpointEntry;
-import org.corfudb.runtime.view.Address;
-import org.corfudb.runtime.view.Layout;
-
 import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINTED_STREAM_ID;
 import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINTED_STREAM_START_LOG_ADDRESS;
 import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.CHECKPOINT_ID;
@@ -34,10 +31,6 @@ import static org.corfudb.protocols.wireprotocol.IMetadata.LogUnitMetadataType.C
  * Created by mwei on 9/18/15.
  */
 public interface IMetadata {
-
-    Map<Byte, LogUnitMetadataType> metadataTypeMap =
-            Arrays.<LogUnitMetadataType>stream(LogUnitMetadataType.values())
-                    .collect(Collectors.toMap(LogUnitMetadataType::asByte, Function.identity()));
 
     EnumMap<IMetadata.LogUnitMetadataType, Object> getMetadataMap();
 
@@ -54,11 +47,12 @@ public interface IMetadata {
 
     /**
      * Get whether or not this entry contains a given stream.
-     * @param stream    The stream to check.
-     * @return          True, if the entry contains the given stream.
+     *
+     * @param stream The stream to check.
+     * @return True, if the entry contains the given stream.
      */
     default boolean containsStream(UUID stream) {
-        return  getBackpointerMap().keySet().contains(stream);
+        return getBackpointerMap().containsKey(stream);
     }
 
     /**
@@ -107,27 +101,29 @@ public interface IMetadata {
         getMetadataMap().put(LogUnitMetadataType.EPOCH, epoch);
     }
 
-    default void setClientId(UUID clientId) {getMetadataMap().put(LogUnitMetadataType.CLIENT_ID, clientId); }
+    default void setClientId(UUID clientId) {
+        getMetadataMap().put(LogUnitMetadataType.CLIENT_ID, clientId);
+    }
 
     @SuppressWarnings("unchecked")
     @Nullable
     default UUID getClientId() {
-            return (UUID) getMetadataMap().getOrDefault(LogUnitMetadataType.CLIENT_ID,
-                    null);
+        return (UUID) getMetadataMap().getOrDefault(LogUnitMetadataType.CLIENT_ID, null);
     }
 
-    default void setThreadId(Long threadId) {getMetadataMap().put(LogUnitMetadataType.THREAD_ID, threadId); }
+    default void setThreadId(Long threadId) {
+        getMetadataMap().put(LogUnitMetadataType.THREAD_ID, threadId);
+    }
 
     @SuppressWarnings("unchecked")
     @Nullable
     default Long getThreadId() {
-        return (Long) getMetadataMap().getOrDefault(LogUnitMetadataType.THREAD_ID,
-                null);
+        return (Long) getMetadataMap().getOrDefault(LogUnitMetadataType.THREAD_ID, null);
     }
-
 
     /**
      * Get Log's global address (global tail).
+     *
      * @return global address
      */
     @SuppressWarnings("unchecked")
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -1,22 +1,23 @@
 package org.corfudb.protocols.wireprotocol;
 
+import com.google.common.base.Preconditions;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
-
-import java.nio.ByteBuffer;
-import java.nio.charset.Charset;
-import java.util.EnumMap;
-import java.util.concurrent.atomic.AtomicReference;
-
 import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.common.compression.Codec;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
 import org.corfudb.protocols.logprotocol.LogEntry;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.WriteSizeException;
+import org.corfudb.runtime.view.Address;
 import org.corfudb.util.serializer.Serializers;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.util.EnumMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Created by mwei on 8/15/16.
@@ -32,13 +33,19 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     @Getter
     byte[] data;
 
-    private ByteBuf serializedCache = null;
+    private SerializedCache serializedCache = null;
 
     private int lastKnownSize = NOT_KNOWN;
 
     private final transient AtomicReference<Object> payload = new AtomicReference<>();
 
     private final EnumMap<LogUnitMetadataType, Object> metadataMap;
+
+    @RequiredArgsConstructor
+    private static class SerializedCache {
+        private final ByteBuf buffer;
+        private final int metadataOffset;
+    }
 
     public static LogData getTrimmed(long address) {
         LogData logData = new LogData(DataType.TRIMMED);
@@ -69,6 +76,17 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
      */
     public Object getPayload(CorfuRuntime runtime) {
         Object value = payload.get();
+
+        // This is only needed for unit test framework to work. Since unit
+        // tests do not serialize payload to byte array, the address will
+        // not be set in the following codes, so doing here instead.
+        if (value instanceof LogEntry) {
+            if (!Address.isAddress(((LogEntry) value).getGlobalAddress())) {
+                ((LogEntry) value).setGlobalAddress(getGlobalAddress());
+            }
+            return value;
+        }
+
         if (value == null) {
             synchronized (this.payload) {
                 value = this.payload.get();
@@ -78,9 +96,9 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
                     } else {
                         ByteBuf serializedBuf = Unpooled.wrappedBuffer(data);
                         if (hasPayloadCodec()) {
-                            // if the payload has a codec we need to decode it before deserialization
+                            // If the payload has a codec we need to decode it before deserialization.
                             ByteBuf compressedBuf = ICorfuPayload.fromBuffer(data, ByteBuf.class);
-                            byte[] compressedArrayBuf= new byte[compressedBuf.readableBytes()];
+                            byte[] compressedArrayBuf = new byte[compressedBuf.readableBytes()];
                             compressedBuf.readBytes(compressedArrayBuf);
                             serializedBuf = Unpooled.wrappedBuffer(getPayloadCodecType()
                                     .getInstance().decompress(ByteBuffer.wrap(compressedArrayBuf)));
@@ -119,20 +137,41 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     @Override
     public synchronized void releaseBuffer() {
         if (serializedCache != null) {
-            serializedCache.release();
-            if (serializedCache.refCnt() == 0) {
+            serializedCache.buffer.release();
+            if (serializedCache.buffer.refCnt() == 0) {
                 serializedCache = null;
             }
         }
     }
 
     @Override
-    public synchronized void acquireBuffer() {
+    public synchronized void acquireBuffer(boolean metadata) {
         if (serializedCache == null) {
-            serializedCache = Unpooled.buffer();
-            doSerializeInternal(serializedCache);
+            acquireBufferInternal(metadata);
         } else {
-            serializedCache.retain();
+            if (metadata) {
+                serializedCache.buffer.resetReaderIndex();
+                serializedCache.buffer.writerIndex(serializedCache.metadataOffset);
+                doSerializeMetadataInternal(serializedCache.buffer);
+            }
+            serializedCache.buffer.retain();
+        }
+    }
+
+    public synchronized void updateAcquiredBuffer(boolean metadata) {
+        Preconditions.checkState(serializedCache != null,
+                "updateAcquiredBuffer requires serialized form");
+        acquireBufferInternal(metadata);
+    }
+
+    private void acquireBufferInternal(boolean metadata) {
+        ByteBuf buf = Unpooled.buffer();
+        if (metadata) {
+            int metadataOffset = doSerializeInternal(buf);
+            serializedCache = new SerializedCache(buf, metadataOffset);
+        } else {
+            doSerializePayloadInternal(buf);
+            serializedCache = new SerializedCache(buf, buf.writerIndex());
         }
     }
 
@@ -166,11 +205,7 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
             data = null;
         }
 
-        if (type.isMetadataAware()) {
-            metadataMap = ICorfuPayload.enumMapFromBuffer(buf, IMetadata.LogUnitMetadataType.class);
-        } else {
-            metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
-        }
+        metadataMap = ICorfuPayload.enumMapFromBuffer(buf, IMetadata.LogUnitMetadataType.class);
     }
 
     /**
@@ -191,7 +226,7 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     /**
      * Constructor for generating LogData.
      *
-     * @param type The type of log data to instantiate.
+     * @param type   The type of log data to instantiate.
      * @param object The actual data/value
      */
     public LogData(DataType type, final Object object) {
@@ -219,8 +254,8 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     /**
      * Constructor for generating LogData.
      *
-     * @param type The type of log data to instantiate.
-     * @param object The actual data/value
+     * @param type      The type of log data to instantiate.
+     * @param object    The actual data/value
      * @param codecType The encoder/decoder type
      */
     public LogData(DataType type, final Object object, final Codec.Type codecType) {
@@ -240,9 +275,6 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
         if (token.getBackpointerMap().size() > 0) {
             setBackpointerMap(token.getBackpointerMap());
         }
-        if (payload.get() instanceof LogEntry) {
-            ((LogEntry) payload.get()).setGlobalAddress(token.getSequence());
-        }
     }
 
     /**
@@ -250,7 +282,7 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
      *
      * @param buf The buffer to read from
      */
-    public byte[] byteArrayFromBuf(final ByteBuf buf) {
+    private byte[] byteArrayFromBuf(final ByteBuf buf) {
         ByteBuf readOnlyCopy = buf.asReadOnly();
         readOnlyCopy.resetReaderIndex();
         byte[] outArray = new byte[readOnlyCopy.readableBytes()];
@@ -261,21 +293,29 @@ public class LogData implements ICorfuPayload<LogData>, IMetadata, ILogData {
     @Override
     public void doSerialize(ByteBuf buf) {
         if (serializedCache != null) {
-            serializedCache.resetReaderIndex();
-            buf.writeBytes(serializedCache);
+            serializedCache.buffer.resetReaderIndex();
+            buf.writeBytes(serializedCache.buffer);
         } else {
             doSerializeInternal(buf);
         }
     }
 
-    void doSerializeInternal(ByteBuf buf) {
+    private int doSerializeInternal(ByteBuf buf) {
+        doSerializePayloadInternal(buf);
+        int metadataOffset = buf.writerIndex();
+        doSerializeMetadataInternal(buf);
+
+        return metadataOffset;
+    }
+
+    private void doSerializePayloadInternal(ByteBuf buf) {
         ICorfuPayload.serialize(buf, type);
         if (type == DataType.DATA) {
             if (data == null) {
                 int lengthIndex = buf.writerIndex();
                 buf.writeInt(0);
                 if (hasPayloadCodec()) {
-                    // if the payload has a codec we need to also compress the payload
+                    // If the payload has a codec we need to also compress the payload
                     ByteBuf serializeBuf = Unpooled.buffer();
                     Serializers.CORFU.serialize(payload.get(), serializeBuf);
                     doCompressInternal(serializeBuf, buf);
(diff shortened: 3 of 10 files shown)


WenbinZhu:
@Maithem the way that IMetadata is used is kind of messy, so I didn't separate it from LogData in this patch, we can have a separate clean up patch later.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ILogData.java
```diff
@@ -51,11 +49,12 @@ public ILogData getSerialized() {
          * Create a new serialized handle with a reference
          * to the log data.
          *
-         * @param data The log data to manage.
+         * @param data     the log data to manage
+         * @param metadata whether metadata needs to be serialized
          */
-        public SerializationHandle(ILogData data) {
+        public SerializationHandle(ILogData data, boolean metadata) {
```
xnull, line 55: Would be great if you replace boolean by 
```
enum SerializationMetaData {
        INCLUDE, EXCLUDE
    }
```

if possible

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
```diff
@@ -54,11 +47,12 @@
...
      * Get whether or not this entry contains a given stream.
-     * @param stream    The stream to check.
-     * @return          True, if the entry contains the given stream.
+     *
+     * @param stream The stream to check.
+     * @return True, if the entry contains the given stream.
      */
     default boolean containsStream(UUID stream) {
-        return  getBackpointerMap().keySet().contains(stream);
+        return getBackpointerMap().keySet().contains(stream);
```
xnull, line 55: a better one `getBackpointerMap().containsKey(stream)`

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -32,14 +31,20 @@
...
+    private SerializedCache serializedCache = null;
 
     private int lastKnownSize = NOT_KNOWN;
 
     private final transient AtomicReference<Object> payload = new AtomicReference<>();
 
     private final EnumMap<LogUnitMetadataType, Object> metadataMap;
 
+    @RequiredArgsConstructor
+    private class SerializedCache {
```
xnull, line 43: it should be static

image

take into account how many instances of LogData we have when Corfu works it should be a huge overhead

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -32,14 +31,20 @@
     @Getter
     byte[] data;
 
-    private ByteBuf serializedCache = null;
+    private SerializedCache serializedCache = null;
```
xnull, line 34: I think `serializedCache` should be Optional<SerializedCache> it will make code more secure and less error-prone

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -119,20 +124,32 @@ public Object getPayload(CorfuRuntime runtime) {
...
         }
     }
 
     @Override
-    public synchronized void acquireBuffer() {
-        if (serializedCache == null) {
-            serializedCache = Unpooled.buffer();
-            doSerializeInternal(serializedCache);
+    public synchronized void acquireBuffer(boolean metadata) {
+        if (serializedCache != null) {
```
xnull, line 136: if you invert the `if` condition it makes code a bit more straightforward
```
    public synchronized void acquireBuffer(boolean metadata) {
        if (serializedCache == null) {
            ByteBuf buf = Unpooled.buffer();
            int metadataOffset;
            
            if (metadata) {
                metadataOffset = doSerializeInternal(buf);
            } else {
                doSerializePayloadInternal(buf);
                metadataOffset = buf.writerIndex();
            }
            serializedCache = new SerializedCache(buf, metadataOffset);
        } else {
            if (metadata) {
                serializedCache.buffer.resetReaderIndex();
                serializedCache.buffer.writerIndex(serializedCache.metadataOffset);
                doSerializeMetadataInternal(serializedCache.buffer);
            }
            serializedCache.buffer.retain();
        }
    }
```

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -119,20 +124,32 @@ public Object getPayload(CorfuRuntime runtime) {
...
+        }
+
+        ByteBuf buf = Unpooled.buffer();
+        if (metadata) {
+            int metadataOffset = doSerializeInternal(buf);
+            serializedCache = new SerializedCache(buf, metadataOffset);
         } else {
-            serializedCache.retain();
+            doSerializePayloadInternal(buf);
+            serializedCache = new SerializedCache(buf, buf.writerIndex());
```
Maithem, line 152: This is a perfect place to use a CompositeByteBuf. The logic of maintaining an offset and overwriting can be greatly simplified. You can also remove the `SerializedCache` class

```java

        CompositeByteBuf serializationCache = Unpooled.compositeBuffer();
        ByteBuf metadata = Unpooled.buffer();
        ByteBuf payload = Unpooled.buffer();
        serializationCache.addComponents(metadata, payload);
```

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ILogData.java
```diff
@@ -67,13 +66,28 @@ public void close() {
...
 
-    default SerializationHandle getSerializedForm() {
-        return new SerializationHandle(this);
-    }
+    /**
+     * Acquire the serialization buffer.
+     *
+     * @param metadata whether metadata needs to be serialized
+     */
+    void acquireBuffer(boolean metadata);
```
Maithem, line 90: I think the metadata serialization is cheap enough to actually do it twice. 
I would just serialize both metadata + payload here. Then after getting a token just call `doSerializeMetadataInternal` with the last ByteBuf of the Composite buffer. It will overwrite it with the new token information. 

Instead of adding the boolean here, you can just introduce a method `updateMetadata` that will rewrite the metadata portion.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -340,12 +359,19 @@ public String toString() {
...
-                throw new WriteSizeException(getSizeEstimate(), limit);
-            }
+    public int checkMaxWriteSize(int limit) {
+        if (serializedCache == null) {
+            throw new IllegalStateException("checkMaxWriteSize requires serialized form");
         }
+
+        int payloadSize = getSizeEstimate();
+
+        if (limit != 0 && payloadSize > limit) {
```
Maithem, line 371: Can remove the first clause if you change maxWriteSize to Integer.MAX_VALUE

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -132,90 +132,98 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
...
+                        // Note that we might invoke the same method multiple times on retries,
+                        // which means the preCommitCallback must be idempotent.
+                        TokenResponse finalTokenResponse = tokenResponse;
+                        log.debug("append: Invoking {} preCommitListeners",
+                                TransactionalContext.getRootContext().getPreCommitListeners().size());
+                        TransactionalContext.getRootContext()
+                                .getPreCommitListeners()
+                                .forEach(e -> e.preCommitCallback(finalTokenResponse));
+                    }
+
```
Maithem, line 189: This doesn't look correct. 

Since the SerializationHandle is acquired before the `for-loop` retries that will end up calling `preCommitCallback` will change the payload, but it will not be re-serialized. Only the token portion will be. This can introduce correctness issues to the queue ordering. 

RIP. I don't remember seeing a PR that introduced this code. It seems like it went in #2498, but it has no reviews.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -132,90 +132,98 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
...
+                    case TX_ABORT_SEQ_OVERFLOW:
+                        abortCause = AbortCause.SEQUENCER_OVERFLOW;
+                        break;
+                    case TX_ABORT_SEQ_TRIM:
+                        abortCause = AbortCause.SEQUENCER_TRIM;
+                        break;
                 }
 
-                // Attempt to write to the log.
-                runtime.getAddressSpaceView().write(tokenResponse, ld, cacheOption);
```
Maithem, line 185: Can you just set the token here instead of propegating it down and setting it later. Ideally, we should just send a LogData object to the write method

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -177,65 +177,72 @@ public long append(Object object,
...
-                    log.debug("Acquisition rejected token={}", tokenResponse);
-                    return -1L;
+        // Opening serialization handle before acquiring token, this way we prevent the
+        // readers to wait for the possibly long serialization time in writer.
+        // The serialization here only serializes the payload because the token is not
+        // acquired yet, thus metadata is incomplete. Once a token is acquired, the
+        // writer will append the serialized metadata to the buffer.
+        try (ILogData.SerializationHandle sh = ld.getSerializedForm(false)) {
+            // Validate if the  size of the log data is under max write size.
+            int payloadSize = ld.checkMaxWriteSize(runtime.getParameters().getMaxWriteSize());
```
Maithem, line 187: Instead of having `checkMaxWriteSize` return the size, you can directly access it from the serialization handle

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: @Maithem Using CompositeByteBuf is exactly same idea that I was initially thinking, but later I changed to simply use one a buffer and offset, because the major purpose of CompositeByteBuf is to provide a single reader/writer index for multiple buffers, but in our case, we mostly want to overwrite the metadata part, which gives two options: 1) using the single reader/writer index that CompositeByteBuf provide, which is same as the current approach, and it's cleaner to just use the normal buffer; 2) maniputate the internal metadata sub-component. This would make confusion on the indexes, and also it's not the standard pattern of using CompositeByteBuf. So I changed to simply use a normal buffer and offset.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
Maithem: What type of overhead ?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ILogData.java
WenbinZhu: The reason I added this flag is that sometimes, metadata is even larger compared to the payload (e.g. logunit cache oom issue we used to fix), and also when the token is not being set, the metadata is in-complete, serializing the metadata makes the meaning of the serialized cache a bit unclear. What do you think?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/IMetadata.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
WenbinZhu: #2498 is the backporting, the PR on master went through reviews.
A quick fix can be putting the serialization in for-loop, but I think it's not the best way of doing it, because we really don't need to serialize payload on every retry just to make it work for the queue case.

@Maithem @hisundar Let's have a discussion to see what is a better solution.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
WenbinZhu: Because when writing the logdata, writer also needs to set the clientId, thereadId, I think it's cleaner to let the writer do that as current impl does, putting that logic into LogData seems a bit unclean and we have to do that all over the places.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
Maithem: I thought that was one of the main use cases of the CompositeByteBuf: allowing multiple buffers to be manipulated independently with the ability to append them to each other cheaply.

https://livebook.manning.com/book/netty-in-action/chapter-5/47

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
Maithem: I see. 

Even pre-this patch this call back can introduce a correctness issue. Imagine getting a commit token for a transaction, but then before replicating the call back changes the payload's contents such that the initial transaction resolution step is invalid.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: So I assume you're referring to the second approahc that manipulates internal metadata sub-component without the offset variable? I that case, if we changed the metadata component, then you also need to change the unified writer index to payload componet writer index + new metadata size. I think this kind of manipulating writer index after write finishes seems error prone and would cause confusion, and it's not clean as having a normal buffer...

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -32,14 +31,20 @@
...
 
     private int lastKnownSize = NOT_KNOWN;
 
     private final transient AtomicReference<Object> payload = new AtomicReference<>();
 
     private final EnumMap<LogUnitMetadataType, Object> metadataMap;
 
+    @RequiredArgsConstructor
+    private static class SerializedCache {
+        private final ByteBuf buffer;
```
xnull, line 44: please add `@NonNull` annotation to guarantee that `buffers` can't be null

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -119,20 +124,42 @@ public Object getPayload(CorfuRuntime runtime) {
     @Override
     public synchronized void releaseBuffer() {
         if (serializedCache != null) {
-            serializedCache.release();
-            if (serializedCache.refCnt() == 0) {
+            serializedCache.buffer.release();
+            if (serializedCache.buffer.refCnt() == 0) {
```
xnull, line 128: Doesn't `release` mean also reset refCnt?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -166,11 +193,7 @@ public LogData(ByteBuf buf) {
             data = null;
         }
 
-        if (type.isMetadataAware()) {
-            metadataMap = ICorfuPayload.enumMapFromBuffer(buf, IMetadata.LogUnitMetadataType.class);
-        } else {
-            metadataMap = new EnumMap<>(IMetadata.LogUnitMetadataType.class);
-        }
+        metadataMap = ICorfuPayload.enumMapFromBuffer(buf, IMetadata.LogUnitMetadataType.class);
```
xnull, line 196: please make metadataMap static, otherwise, it will be a huge overhead

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -119,20 +124,42 @@ public Object getPayload(CorfuRuntime runtime) {
...
+                serializedCache.buffer.writerIndex(serializedCache.metadataOffset);
+                doSerializeMetadataInternal(serializedCache.buffer);
+            }
+            serializedCache.buffer.retain();
+        }
+    }
+
+     public synchronized void updateAcquiredBuffer(boolean metadata) {
+        if (serializedCache == null) {
+            return;
```
xnull, line 150: shouldn't we create an instance of serializedCache?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -340,12 +369,19 @@ public String toString() {
...
      * @param limit Max write limit.
+     * @return the serialized size of the payload
      */
-    public void checkMaxWriteSize(int limit) {
-        try (ILogData.SerializationHandle sh = this.getSerializedForm()) {
-            if (limit != 0 && getSizeEstimate() > limit) {
-                throw new WriteSizeException(getSizeEstimate(), limit);
-            }
+    public int checkMaxWriteSize(int limit) {
+        if (serializedCache == null) {
```
xnull, line 375: when one small method throws two different exceptions it seems too much, can we throw the only WriteSizeException?

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -131,91 +131,105 @@ public void gc(long trimMark) {
...
-                            .forEach(e -> e.preCommitCallback(finalTokenResponse));
+        // Opening serialization handle before acquiring token, this way we prevent the
+        // readers to wait for the possibly long serialization time in writer.
+        // The serialization here only serializes the payload because the token is not
+        // acquired yet, thus metadata is incomplete. Once a token is acquired, the
+        // writer will append the serialized metadata to the buffer.
+        try (ILogData.SerializationHandle sh = ld.getSerializedForm(serializeMetadata)) {
+            int payloadSize = ld.checkMaxWriteSize(runtime.getParameters().getMaxWriteSize());
+
+            for (int x = 0; x < runtime.getParameters().getWriteRetry(); x++) {
```
xnull, line 146: please rename `x`

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: This should be called when serializedCache is already acquired

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: A precondition check is required to throw IllegalStateException

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Release reduces refCnt by one.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -119,20 +124,42 @@ public Object getPayload(CorfuRuntime runtime) {
...
         if (serializedCache == null) {
-            serializedCache = Unpooled.buffer();
-            doSerializeInternal(serializedCache);
+            acquireBufferInternal(metadata);
         } else {
-            serializedCache.retain();
+            if (metadata) {
+                serializedCache.buffer.resetReaderIndex();
+                serializedCache.buffer.writerIndex(serializedCache.metadataOffset);
+                doSerializeMetadataInternal(serializedCache.buffer);
```
Maithem, line 142: You can slice the buffer at the `metadataOffset` offset

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -119,20 +124,42 @@ public Object getPayload(CorfuRuntime runtime) {
...
+    }
+
+     public synchronized void updateAcquiredBuffer(boolean metadata) {
+        if (serializedCache == null) {
+            return;
+        }
+        acquireBufferInternal(metadata);
+    }
+
+    private void acquireBufferInternal(boolean metadata) {
```
hisundar, line 155: acquireBufferInternsl is really serializing data, making the name misleading. Perhaps something like serializeToNewBuffer()

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Slice will share the same buffer and create a new buffer, I think just set the writerIndex might be simpler?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Initially I also wanted to name it serailizeXXX, but we have other methods with this form of name, so to aviod confusion, I changed to use acquireBufferInternal.

Maithem:
@WenbinZhu can you port to corfu-0.3.0.2 ?

WenbinZhu:
> @WenbinZhu can you port to corfu-0.3.0.2 ?

@Maithem Done. https://github.com/CorfuDB/CorfuDB/pull/2724

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -131,91 +131,105 @@ public void gc(long trimMark) {
...
+            int payloadSize = ld.checkMaxWriteSize(runtime.getParameters().getMaxWriteSize());
+
+            for (int retry = 0; retry < runtime.getParameters().getWriteRetry(); retry++) {
+                // Go to the sequencer, grab a token to write.
+                tokenResponse = conflictInfo == null
+                        ? runtime.getSequencerView().next(streamIDs) // Token w/o conflict info
+                        : runtime.getSequencerView().next(conflictInfo, streamIDs); // Token w/ conflict info
+
+                // Is our token a valid type?
+                AbortCause abortCause = null;
```
xnull, line 153: we don't break the cycle here, is it correct? Are we looking for the latest abortCause on the list?
That looks a bit unusual. Could you please explain what expected behavior here?

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -131,91 +131,105 @@ public void gc(long trimMark) {
...
-                            TransactionalContext.getRootContext().getPreCommitListeners().size());
-                    TransactionalContext.getRootContext()
-                            .getPreCommitListeners()
-                            .forEach(e -> e.preCommitCallback(finalTokenResponse));
+        // Opening serialization handle before acquiring token, this way we prevent the
+        // readers to wait for the possibly long serialization time in writer.
+        // The serialization here only serializes the payload because the token is not
+        // acquired yet, thus metadata is incomplete. Once a token is acquired, the
+        // writer will append the serialized metadata to the buffer.
+        try (ILogData.SerializationHandle sh = ld.getSerializedForm(serializeMetadata)) {
```
xnull, line 143: the whole method is quite big, can you split it to a couple of smaller ones?

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -131,91 +131,105 @@ public void gc(long trimMark) {
...
+
+            for (int retry = 0; retry < runtime.getParameters().getWriteRetry(); retry++) {
+                // Go to the sequencer, grab a token to write.
+                tokenResponse = conflictInfo == null
+                        ? runtime.getSequencerView().next(streamIDs) // Token w/o conflict info
+                        : runtime.getSequencerView().next(conflictInfo, streamIDs); // Token w/ conflict info
+
+                // Is our token a valid type?
+                AbortCause abortCause = null;
+                switch (tokenResponse.getRespType()) {
```
xnull, line 154: Alternatively, you can use a map
```
Map<TokenType, AbortCause> dict = new HashMap<>();
                dict.put(TX_ABORT_CONFLICT, AbortCause.CONFLICT);
                dict.put(TX_ABORT_NEWSEQ, AbortCause.NEW_SEQUENCER);
                dict.put(TX_ABORT_SEQ_OVERFLOW, AbortCause.SEQUENCER_OVERFLOW);
                dict.put(TX_ABORT_SEQ_TRIM, AbortCause.TX_ABORT_SEQ_TRIM);

                AbortCause abortCause = dict.get(tokenResponse.getRespType());
...
```

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -131,91 +131,105 @@ public void gc(long trimMark) {
...
-                log.warn("append[{}]: StaleToken, streams {}", tokenResponse.getSequence(),
-                        Arrays.stream(streamIDs).map(Utils::toReadableId).collect(Collectors.toSet()));
-
-                throw new TransactionAbortedException(
-                        conflictInfo,
-                        tokenResponse.getConflictKey(), tokenResponse.getConflictStream(),
-                        tokenResponse.getToken().getSequence(),
-                        AbortCause.NEW_SEQUENCER, // in the future perhaps define a new AbortCause?
-                        TransactionalContext.getCurrentContext());
+                try {
```
xnull, line 177: everything inside for cycle above this point should be extracted into a method, let's say checkAbort();
```
for (int retry = 0; retry < runtime.getParameters().getWriteRetry(); retry++) {
                checkAbort();

                try {
                    if (TransactionalContext.isInTransaction()) {...
```

xnull: the whole for cycle would be good to have like this:
```
for (int retry = 0; retry < runtime.getParameters().getWriteRetry(); retry++) {
                checkAbort();
                write();
```

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -177,65 +177,72 @@ public long append(Object object,
...
+            for (int retry = 0; retry < runtime.getParameters().getWriteRetry(); retry++) {
+                // Next, we call the acquisitionCallback, if present, informing
+                // the client of the token that we acquired.
+                if (acquisitionCallback != null) {
+                    if (!acquisitionCallback.apply(tokenResponse)) {
+                        // The client did not like our token, so we end here.
+                        // We'll leave the hole to be filled by the client or
+                        // someone else.
+                        log.warn("Acquisition rejected token={}", tokenResponse);
+                        return -1L;
```
xnull, line 203: is it Address.NON_ADDRESS ?

xnull:
BTW, I don't if this PR affected performance or others recently merged master, but the time taken by Integration tests increased by a factor of 2. Before it was ~25 min, now it's 45 min.

So, would be great to have benchmarks for the serialization functionality

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
WenbinZhu: Done.

WenbinZhu:
> BTW, I don't if this PR affected performance or others recently merged master, but the time taken by Integration tests increased by a factor of 2. Before it was ~25 min, now it's 45 min.
> 
> So, would be great to have benchmarks for the serialization functionality
> 
> I can write those (maybe in another PR) or help with writing one

It became 45 minutes because LR patch is merged, which introduced a lot of new integration tests.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ILogData.java
Maithem: I see.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -119,20 +124,42 @@ public Object getPayload(CorfuRuntime runtime) {
...
+    public synchronized void acquireBuffer(boolean metadata) {
         if (serializedCache == null) {
-            serializedCache = Unpooled.buffer();
-            doSerializeInternal(serializedCache);
+            acquireBufferInternal(metadata);
         } else {
-            serializedCache.retain();
+            if (metadata) {
+                serializedCache.buffer.resetReaderIndex();
+                serializedCache.buffer.writerIndex(serializedCache.metadataOffset);
```
Maithem, line 141: Why do you need to reset the reader index? shouldn't you assert that its actually zero?

When would you read the payload, then reseralize the metadata portion and then reread ?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
Maithem: If this is only expected to be called if `acquireBuffer` and that would create an instance of `SerializedCache` , then this can never be null, if it is then it should throw an exception.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -340,12 +369,17 @@ public String toString() {
...
-                throw new WriteSizeException(getSizeEstimate(), limit);
-            }
+    public int checkMaxWriteSize(int limit) {
+        Preconditions.checkState(serializedCache != null, "checkMaxWriteSize requires serialized form");
+
+        int payloadSize = getSizeEstimate();
+
+        if (payloadSize > limit) {
+            throw new WriteSizeException(payloadSize, limit);
         }
```
Maithem, line 381: This doesn't seem right. When an object is serialized (not byte array), then this always return 1. Having `serializedCache != null` doesn't imply that `LogData::data` is populated with the serialized data. 

Shouldn't this logic check the size against `serializedCache.writerIndex()` ?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Because in case of retries, the same LogData is read into the netty buffer, which changes readerIndex, in the retries, we need to reset reader index.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: @Maithem getSizeEstimate() will use `lastKnownSize` if that is available, which will be set to the correct payload size when the payload was first serialized, so it will not return 1.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -131,91 +131,75 @@ public void gc(long trimMark) {
...
+        TokenResponse tokenResponse = null;
 
-        ld.checkMaxWriteSize(runtime.getParameters().getMaxWriteSize());
+        // Opening serialization handle before acquiring token, this way we prevent the
+        // readers to wait for the possibly long serialization time in writer.
+        // The serialization here only serializes the payload because the token is not
+        // acquired yet, thus metadata is incomplete. Once a token is acquired, the
+        // writer will append the serialized metadata to the buffer.
+        try (ILogData.SerializationHandle sh = ld.getSerializedForm(serializeMetadata)) {
+            int payloadSize = ld.checkMaxWriteSize(runtime.getParameters().getMaxWriteSize());
```
Maithem, line 144: This doesn't cover the case of writes are issued directly to the `AddressSpaceView::write` method (i.e. the checkpointer writer).

Maithem: It seems like the CheckpointWriter will use this path too.

WenbinZhu: Any  request directly from AddressSpaceView::write will be serialized only once in ChainReplicationProtocol::write, that path should be fine, right?

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -131,91 +131,75 @@ public void gc(long trimMark) {
...
-                if (conflictInfo != null) {
-                    // On retry, check for conflicts only from the previous attempt position,
-                    // otherwise the transaction will always conflict with itself.
-                    conflictInfo.setSnapshotTimestamp(tokenResponse.getToken());
-                }
+                try {
+                    // Run pre-commit listeners if we are in transaction.
+                    runPreCommitListeners(tokenResponse, ld, serializeMetadata);
+                    // Attempt to write to the log.
+                    runtime.getAddressSpaceView().write(tokenResponse, ld, cacheOption);
```
annym, line 167: question, in AddressSpaceView.Write we call useToken which internally sets the global address to the payload, but since, we've already serialized the payload and we'll just serialize the metadata, won't this be lost in the serialized form? 

https://github.com/CorfuDB/CorfuDB/blob/4530118dc80fee2c3a896910bb4fc88ecf6775d3/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java#L243-L245

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -340,12 +369,17 @@ public String toString() {
...
-    public void checkMaxWriteSize(int limit) {
-        try (ILogData.SerializationHandle sh = this.getSerializedForm()) {
-            if (limit != 0 && getSizeEstimate() > limit) {
-                throw new WriteSizeException(getSizeEstimate(), limit);
-            }
+    public int checkMaxWriteSize(int limit) {
+        Preconditions.checkState(serializedCache != null, "checkMaxWriteSize requires serialized form");
+
+        int payloadSize = getSizeEstimate();
+
```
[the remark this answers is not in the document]
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
```diff
@@ -229,6 +213,47 @@ public long append(@Nonnull Object object, @Nullable TxResolutionInfo conflictIn
...
+        if (TransactionalContext.isInTransaction()) {
+            // If this transaction has entries that wish to capture the committed address
+            // invoke its preCommitCallbacks with the tokenResponse from the sequencer.
+            // Note that we might invoke the same method multiple times on retries,
+            // which means the preCommitCallback must be idempotent.
+            log.debug("append: Invoking {} preCommitListeners",
+                    TransactionalContext.getRootContext().getPreCommitListeners().size());
+            List<TransactionalContext.PreCommitListener> listeners =
+                    TransactionalContext.getRootContext().getPreCommitListeners();
+            // If there are pre-commit listeners, the payload will be changed,
```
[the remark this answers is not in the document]
WenbinZhu: Yes, for now we have to assume it changes LogData...

runtime/src/main/java/org/corfudb/runtime/view/StreamsView.java
WenbinZhu: The address in the LogEntry is used when we sync object, for real workloads, this  block is actually redundant, because the address will be set to LogData's address when we call getPayload(), so even if we set here, it will be overwritten after getPayload() is called. However in unit test framework, on the write path, payload is not being serailized to data[], that means when getPayload() is called, it's always in object format, so it won't go to following code to set the address.
This block of code is used to workaround the unit test framework case. I moved this logic to getPayload and added comments to make it more clear.

https://github.com/CorfuDB/CorfuDB/blob/4530118dc80fee2c3a896910bb4fc88ecf6775d3/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java#L94-L95

annym: Great, that makes it cleaner. Thanks!

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
xnull: > What type of overhead ?

This is what I remember from "Effective Java" 
https://stackoverflow.com/questions/43998428/why-to-favor-static-member-classes-over-nonstatic-joshua-bloch-item-22#:~:text=Answer%3A,reference%20to%20its%20enclosing%20instance.

PR #2704, opened 2020-08-12
Fix NPE & Shutdown Server for UnrecoverableErrors

annym:
## Overview

Description:

- Fix a NPE that we hit on stopLogReplication when an invalid topology is provided onProcessNotificationUpdate. 
- Shutdown Corfu Log Replication Service whenever Discovery Service is terminated. 
- Fix some logs to improve current readability/debugging

Why should this be merged: fix some ongoing bugs in LR

- [X] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
@@ -123,9 +123,10 @@ public class LogReplicationServer extends AbstractServer {
 
     @ServerHandler(type = CorfuMsgType.LOG_REPLICATION_QUERY_LEADERSHIP)
     private void handleLogReplicationQueryLeadership(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.info("Log Replication Query Leadership Request received by Server.");
+        log.debug("Log Replication Query Leadership Request received by Server.");
         LogReplicationQueryLeaderShipResponse resp = new LogReplicationQueryLeaderShipResponse(0,
                 isLeader.get(), serverContext.getLocalEndpoint());
+        log.debug("Send Log Replication Leadership Response isLeader={}, endpoint={}", resp.isLeader(), resp.getEndpoint());
         r.sendResponse(msg, CorfuMsgType.LOG_REPLICATION_QUERY_LEADERSHIP_RESPONSE.payloadMsg(resp));
     }
 
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -179,6 +179,9 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
                     DiscoveryServiceEvent event = eventQueue.take();
                     processEvent(event);
                 } catch (Exception e) {
+                    // TODO: We should take care of which exceptions really end up being
+                    //  caught at this level, or we could be stopping LR completely on
+                    //  any exception.
                     log.error("Caught an exception. Stop discovery service.", e);
                     shouldRun = false;
                     stopLogReplication();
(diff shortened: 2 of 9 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -463,7 +465,7 @@ private void fetchTopologyFromClusterManager() {
      * Stop Log Replication
      */
     private void stopLogReplication() {
-        if (localClusterDescriptor.getRole() == ClusterRole.ACTIVE && isLeader.get()) {
+        if (localClusterDescriptor != null && localClusterDescriptor.getRole() == ClusterRole.ACTIVE && isLeader.get()) {
```
zhangn49, line 468: A little confused, If a new topology is not valid, we don't update the localClusterDescriptor, right? How we hit this NPE?

If the localClusterDescriptor is always null, how does the log replication start?... It should hit NPE here.
https://github.com/CorfuDB/CorfuDB/blob/8fc1b8b4671c8783462330247c65ceebfa9bef07/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java#L414

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -123,9 +123,10 @@ private void handleLogReplicationNegotiationRequest(CorfuMsg msg, ChannelHandler
 
     @ServerHandler(type = CorfuMsgType.LOG_REPLICATION_QUERY_LEADERSHIP)
     private void handleLogReplicationQueryLeadership(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.info("Log Replication Query Leadership Request received by Server.");
+        log.debug("Log Replication Query Leadership Request received by Server.");
```
pankti-m, line 126: dint we want to make this trace?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -188,11 +188,13 @@ public void run() {
                 }
             }
         } catch (Exception e) {
-            log.error("Unhandled exception caught during log replication service discovery. Retry,", e);
+            log.error("Unhandled exception caught during log replication service discovery.", e);
```
pankti-m, line 191: is it a good idea to add a todo for revisiting these exceptions handling which we discussed today?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: Yes,  we don't update the localClusterDescriptor and this is called from:

https://github.com/CorfuDB/CorfuDB/blob/8fc1b8b4671c8783462330247c65ceebfa9bef07/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java#L579

At this point we haven't started LR, we wouldn't even be able to acquire the lock because we are still not recognized as part of any cluster...We just. get topology updates from SM, where we don't show up as part of any active or standby cluster...

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
annym: yes, but eventually when the product is a bit more stable. We said to leave it while we're still in early testing stage. But definitely something to downgrade to trace afterwards.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
annym: definitely, so we don't forget it. Let me add it.

Issue #2700, opened 2020-08-11
LockIT.testMultipleLockClientsSameLockFailure


zfrenette:
## Overview
In master, the test testMultipleLockClientsSameLockFailure (LockIT) fails about 50% of the time. 

- (X) **Bug report.** 
In particular, the failures occur on line 318 when asserting that one client has acquired the lock. We expect that one client will acquire the lock, when in actuality, no client appears to do so. This can be observed by running the test many times in a row.

PR #2705, opened 2020-08-12
Fix Lock ITs

annym:
#Overview

Description:

- Intermittent failures due to race condition on FSM transition which affects all other 3 tests in LockIT.

Why should this be merged: stabilize Lock ITs

Related issue(s) (if applicable): #2700

- [X] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/LockIT.java
+++ b/test/src/test/java/org/corfudb/integration/LockIT.java
@@ -153,13 +153,18 @@ public class LockIT extends AbstractIT implements Observer {
             // Since this is the only client, ALL locks should've been acquired, verify, block until condition is met
             waitCondition = WaitConditionType.LOCK_ACQUIRED;
             for (int i=0; i<numLocks; i++) {
-                log.debug("***** Wait until lock " + i + " is acquired");
+                log.debug("***** Wait until lock {} is acquired", i);
                 blockUntilWaitCondition.acquire();
             }
 
             assertThat(lockAcquiredObservables.get(clientId).getValue()).isEqualTo(numLocks);
-            lockIds.forEach(lockId ->
-                assertThat(client.getLocks().get(lockId).getState().getType()).isEqualTo(LockStateType.HAS_LEASE));
+            lockIds.forEach(lockId -> {
+                    while(client.getLocks().get(lockId).getState().getType() != LockStateType.HAS_LEASE) {
+                        // Lock state might take longer to update as the listener is updated before the FSM transitions
+                        // to the new state
+                    }
+                    assertThat(client.getLocks().get(lockId).getState().getType()).isEqualTo(LockStateType.HAS_LEASE);
+            });
         } catch (Exception e) {
             log.debug("Unexpected exception: " + e);
             throw e;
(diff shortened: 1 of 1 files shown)


test/src/test/java/org/corfudb/integration/LockIT.java
```diff
@@ -153,13 +153,18 @@ public void testSingleLockClientMultipleLocks() throws Exception {
...
             }
 
             assertThat(lockAcquiredObservables.get(clientId).getValue()).isEqualTo(numLocks);
-            lockIds.forEach(lockId ->
-                assertThat(client.getLocks().get(lockId).getState().getType()).isEqualTo(LockStateType.HAS_LEASE));
+            lockIds.forEach(lockId -> {
+                    while(client.getLocks().get(lockId).getState().getType() != LockStateType.HAS_LEASE) {
+                        // Lock state might take longer to update as the listener is updated before the FSM transitions
+                        // to the new state
+                    }
```
Maithem, line 165: Shouldn't there be a sleep here?

Maithem: And a max timeout wait before the test fails so that it doesn't keep on spinning indefinitely.

annym: There could be a sleep. The thing is the transition happens in the line right after the listener is notified that the lock was acquired. So technically it shouldn't be spinning for too long. It's just the interleaving between threads...  

Also, regarding the timeout. Isn't it covered by the max time of each test? otherwise it times out.

test/src/test/java/org/corfudb/integration/LockIT.java
```diff
@@ -230,17 +235,25 @@ public void testMultipleLockClientsSameLock() throws Exception {
...
             LockClient clientWithLock = clientsWithLock.get(0);
 
-
             // Verify for 5 cycles that the lock is renewed
             for (int i=0; i < RENEW_CYCLES; i++) {
                 log.debug("***** Wait until lock is renewed");
                 // Wait for the renewal cycle + 1, and verify that the lock is still acquired by the same client
                 Sleep.sleepUninterruptibly(Duration.ofSeconds(LOCK_TIME_CONSTANT + 1));
-                clientsWithLock = getClientsThatAcquiredLock(lockId, clientIdToLockClient);
+                do {
```
pankti-m, line 254: can we add the comment from line 238 here as well?

test/src/test/java/org/corfudb/integration/LockIT.java
```diff
@@ -330,7 +350,10 @@ public void testMultipleLockClientsSameLockFailure() throws Exception {
             }
 
             // Verify that only one client has acquired the lock
-            clientsWithLock = getClientsThatAcquiredLock(lockId, clientIdToLockClient);
+            do {
```
pankti-m, line 353: comment

test/src/test/java/org/corfudb/integration/LockIT.java
annym: Done.

test/src/test/java/org/corfudb/integration/LockIT.java
annym: Done.

PR #2711, opened 2020-08-13
LR Fix Timeout Exceptions

annym:
## Overview

Description:

Catch TimeoutExceptions correctly as these are wrapped by the CompletableFuture in an ExecutionException.

Why should this be merged: functional bug

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
@@ -69,6 +69,8 @@ public abstract class SenderBufferManager {
 
     private DataSender dataSender;
 
+    private long topologyConfigId;
+
     /*
      * The messages sent to the receiver that have not been ACKed yet.
      */
@@ -119,7 +121,6 @@ public abstract class SenderBufferManager {
             errorOnMsgTimeout = Boolean.parseBoolean(props.getProperty("log_reader_error_on_message_timeout",
                     Boolean.toString(errorOnMsgTimeout)));
             reader.close();
-
         } catch (Exception e) {
             log.warn("The config file is not available {} , will use the default values for config.", config_file, e.getCause());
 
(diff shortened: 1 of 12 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
```diff
@@ -175,20 +175,20 @@ public void sendWithBuffering(List<LogReplicationEntry> dataToSend) {
...
     public LogReplicationEntry resend() {
-        //Enforce a resend or not
         LogReplicationEntry ack = null;
         boolean force = false;
         try {
             ack = processAcks();
-        } catch (TimeoutException e) {
-            log.warn("Caught a timeout exception ", e);
-            force = true;
+        } catch (ExecutionException ee) {
```
zhangn49, line 184: I think we still need the catch block for `TimeoutException`.

There are two kinds of `TimeoutException`:
1. From the actual sendMsg's CF, this type will be wrapped with the `ExecutionException `
2. From anyOf(cfs).get(long timeout, TimeUnit unit)

annym: Yes, you're right. Added a catch on TimeoutException as well...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
```diff
@@ -175,36 +174,32 @@ public void sendWithBuffering(List<LogReplicationEntry> dataToSend) {
...
     public LogReplicationEntry resend() {
-        //Enforce a resend or not
         LogReplicationEntry ack = null;
         boolean force = false;
         try {
             ack = processAcks();
-        } catch (TimeoutException e) {
-            log.warn("Caught a timeout exception ", e);
+        } catch (TimeoutException te) {
+            log.warn("Caught a timeout exception while processing ACKs", te);
```
xiaoqin2012, line 184: Can you add comments to explain what can lead this exception?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
```diff
@@ -175,36 +174,32 @@ public void sendWithBuffering(List<LogReplicationEntry> dataToSend) {
...
             force = true;
+        } catch (ExecutionException ee) {
+            log.warn("Caught an execution exception while processing ACKs", ee);
```
xiaoqin2012, line 187: Same as the above add some comments to explain.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/CorfuLogReplicationRuntime.java
```diff
@@ -159,8 +161,8 @@ public CorfuLogReplicationRuntime(LogReplicationRuntimeParameters parameters, Lo
         this.sourceManager = new LogReplicationSourceManager(parameters, new LogReplicationClient(router, remoteClusterId),
             metadataManager);
         this.connectedEndpoints = new HashSet<>();
-        this.communicationFSMWorkers = Executors.newSingleThreadExecutor(new
-                ThreadFactoryBuilder().setNameFormat("runtime-fsm-worker").build());
+        this.communicationFSMWorkers = new ThreadPoolExecutor(1, 1, 0L,
+                TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>());
```
zhangn49, line 165: Do we need to set thread name by a threadFactory?

annym: Yes, I think this is a good idea, so we track it easier. Added.

Issue #2710, opened 2020-08-13
CorfuQueue: Records Are Out of Order When Consumed


Maithem:
**Bug report.**
Calling CorfuQueue::enqueue without a transaction produces out of order elements when consumed.

For example this code

```java

final int numItemsToProduce = 10;
        IntStream.range(0, numItemsToProduce).forEach(itemIdx -> {
            if (itemIdx % 2 == 0) {
                getDefaultRuntime().getObjectsView().TXBegin();
                oldQueueInstance1.enqueue(getByteString(String.valueOf(itemIdx)));
                getDefaultRuntime().getObjectsView().TXEnd();
            } else {
                oldQueueInstance2.enqueue(getByteString(String.valueOf(itemIdx)));
            }
        });

```

will produce 
0
1
3
5
7
9
2
4
6
8

PR #2712, opened 2020-08-13
Fail Non-Transactional Enqueue

Maithem:
## Overview
Since CorfuQueue::enqueue depends on a transactional context to
work correctly, a validation step is required to fail enqueue
calls that are not within a transaction.
Description:

Why should this be merged: Addresses #2710

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
@@ -1,7 +1,19 @@
 package org.corfudb.runtime.collections;
 
+import static com.google.common.base.Preconditions.checkState;
+
+
 import com.google.common.reflect.TypeToken;
 import com.google.protobuf.InvalidProtocolBufferException;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
@@ -13,15 +25,6 @@ import org.corfudb.runtime.object.transactions.TransactionalContext.PreCommitLis
 import org.corfudb.runtime.view.CorfuGuidGenerator;
 import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
-
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
 
 /**
  * Persisted Queue supported by CorfuDB using distributed State Machine Replication.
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -157,40 +160,38 @@ public int hashCode() {
      *         element prevents it from being added to this queue
      */
     public CorfuRecordId enqueue(E e) {
+        checkState(TransactionalContext.isInTransaction(), "must be called within a transaction!");
```
xnull, line 163: I'm concerned, if this change could affect the clients since we changed contract, hope it works

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -157,40 +160,38 @@ public int hashCode() {
...
-
-        corfuTable.put(id, e);
-        return id;
+        /**
+         * This is a callback that is placed into the root transaction's context on
+         * the thread local stack which will be invoked right after this transaction
+         * is deemed successful and has obtained a final sequence number to write.
+         */
+         class QueueEntryAddressGetter implements PreCommitListener {
+             private CorfuRecordId recordId;
```
xnull, line 172: final?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -157,40 +160,38 @@ public int hashCode() {
...
+             private QueueEntryAddressGetter(CorfuRecordId recordId) {
```
xnull, line 173: @AllArgsConstructor ?

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
```diff
@@ -1,38 +1,58 @@
...
  */
+@Slf4j
 public class CorfuQueueTest extends AbstractViewTest {
 
+    @Test
+    public void failNonTxnEnqueue() {
+        CorfuQueue<String>
+                corfuQueue = new CorfuQueue<>(getDefaultRuntime(), "test");
+        Assertions.assertThatThrownBy(() -> corfuQueue.enqueue("c"))
+                .isInstanceOf(IllegalStateException.class)
```
xnull, line 37: Just a question: Are we ok with IllegalStateException? Maybe we have a more specific exception?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
Maithem: Unfortunately there isn't much flexibility here. If clients are enqueuing without a transaction then its incorrect. If we start a transaction internally, then that transaction can abort. They're probably not dealing with the TransacationAbort right now

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
```diff
@@ -157,40 +160,38 @@ public int hashCode() {
...
+             @Override
+             public void preCommitCallback(TokenResponse tokenResponse) {
+                 recordId.setTxSequence(tokenResponse.getSequence());
+                 log.trace("preCommitCallback for Queue: " + recordId.toString());
+             }
+         }
+
+         QueueEntryAddressGetter addressGetter = new QueueEntryAddressGetter(id);
+         log.trace("enqueue: Adding preCommitListener for Queue: " + id.toString());
+         TransactionalContext.getRootContext().addPreCommitListener(addressGetter);
```
xnull, line 191: @hisundar we never clean PreCommitListeners? Is it ok?

Maithem:
@zhangn49 Need to rebase the other PR on this. So yes, some tests will break.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
Maithem: The listener is part of the TransactionalContext which is thread local. After the transaction completes or fails it will be garbage collected.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuQueue.java
Maithem: Done.

test/src/test/java/org/corfudb/runtime/collections/CorfuQueueTest.java
Maithem: I think this should suffice. This exception signals an incorrect usage of the API. There is nothing to handle per-se
```java

 * Signals that a method has been invoked at an illegal or
 * inappropriate time.  In other words, the Java environment or
 * Java application is not in an appropriate state for the requested
 * operation.

```

PR #2715, opened 2020-08-14
Sequencer Reconfiguration Corruption Fixes

Maithem:
## Overview

Prior to this patch, the global tail, stream tails and stream bit
sets were retrieved and computed from the first node in the last
segment. This assumption is incorrect because it can be a node
that is being rebuilt (via state transfer), or in some edge cases
it is not possible to compute the sequencer state from a single node.
This leads to the possibility of computing a view of the log that
is incomplete during reconfiguration of the sequencer. In
consequence, leading to inconsistent reads, data loss and
cluster instability.

This patch introduces multiple fixes to compute the sequencer's
state from all the segments (possibly aggregating views from different
nodes) and forcing the view computation and the reconfiguration to
happen on the same epoch.

Why should this be merged: Fixes data loss/inconsistency/stability issues

Related issue(s) (if applicable): #2678

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
@@ -41,8 +41,34 @@ public class StreamAddressSpace {
         this.addressMap = addressMap;
     }
 
+    public StreamAddressSpace() {
+        this.addressMap = Roaring64NavigableMap.bitmapOf();
+        this.trimMark = Address.NON_ADDRESS;
+    }
+
     public Roaring64NavigableMap getAddressMap() {
         return addressMap;
+    }
+
+    /**
+     * Merges b into a and returns a as the final result.
+     * @param a StreamAddressSpace to merge into
+     * @param b StreamAddressSpace to merge
+     * @return returns a as the merged StreamAddressSpace
+     */
+    public static StreamAddressSpace merge(StreamAddressSpace a, StreamAddressSpace b) {
+        a.mergeFrom(b);
+        return a;
+    }
+
+    // Merge another StreamAddressSpace into this instance
+    private void mergeFrom(StreamAddressSpace other) {
+        this.trimMark = Long.max(trimMark, other.getTrimMark());
+        this.addressMap.or(other.getAddressMap());
+
+        // Because the trim mark can increase after the merge another
+        // trim needs to be issued
+        this.trim(this.trimMark);
     }
 
     /**
@@ -68,7 +94,7 @@ public class StreamAddressSpace {
      */
     public Long getTail() {
         // If no address is present for this stream, the tail is given by the trim mark (last trimmed address)
-        if (addressMap.getLongCardinality() == NO_ADDRESSES) {
+        if (addressMap.isEmpty()) {
             return trimMark;
         }
 
@@ -89,7 +115,7 @@ public class StreamAddressSpace {
      * Remove addresses from the stream's address map
      * and set the new trim mark (to the greatest of all addresses to remove).
      */
-    public void removeAddresses(List<Long> addresses) {
+    private void removeAddresses(List<Long> addresses) {
         addresses.stream().forEach(addressMap::removeLong);
 
         // Recover allocated but unused memory
--- a/runtime/src/main/java/org/corfudb/util/Utils.java
+++ b/runtime/src/main/java/org/corfudb/util/Utils.java
@@ -1,6 +1,21 @@
 package org.corfudb.util;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
+
+import java.text.DecimalFormat;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 import jdk.internal.org.objectweb.asm.util.Printer;
 import jdk.internal.org.objectweb.asm.util.Textifier;
 import jdk.internal.org.objectweb.asm.util.TraceMethodVisitor;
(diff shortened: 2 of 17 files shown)


runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -205,99 +229,78 @@ public static void updateCommittedTail(RuntimeLayout runtimeLayout,
...
 
     /**
-     * Get global log tail.
+     * Find the chain's head node of each segment
+     * @param layout layout to search in
+     * @return returns a set of nodes the represent the first node in all segments
+     */
+    private static Set<String> getChainHeadFromAllSegments(Layout layout) {
+        List<Layout.LayoutSegment> segments = layout.getSegments();
+        validateSegments(layout.getSegments());
```
WenbinZhu, line 238: This can be moved to the first line.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -205,99 +229,78 @@ public static void updateCommittedTail(RuntimeLayout runtimeLayout,
...
-
-                TailsResponse res = CFUtils.getUninterruptibly(runtimeLayout
-                                .getLogUnitClient(stripe.getLogServers().get(DEFAULT_LOGUNIT))
-                                .getAllTails());
-                luResponses.add(res);
-            }
-        } else if (segment.getReplicationMode() == Layout.ReplicationMode.QUORUM_REPLICATION) {
-            throw new UnsupportedOperationException();
-        }
+        long[] globalTail = {Address.NON_EXIST};
```
WenbinZhu, line 282: AtomicLong is cleaner than array.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -307,34 +310,27 @@ static TailsResponse aggregateLogUnitTails(Set<TailsResponse> responses) {
...
-
-        for (StreamsAddressResponse res : responses) {
-            logTail = Math.max(logTail, res.getLogTail());
-            streamAddressSpace = aggregateStreamAddressMap(res.getAddressMap(), streamAddressSpace);
-        }
-        return new StreamsAddressResponse(logTail, streamAddressSpace);
+        // Since a node can exist as a head for multiple segments we need to a set to
+        // coalesce the candidates to unique nodes only
+        Set<String> segmentsHeadNodes = getChainHeadFromAllSegments(runtimeLayout.getLayout());
+        long[] globalTail = {Address.NON_EXIST};
```
WenbinZhu, line 316: AtomicLong is cleaner than array.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -205,99 +229,78 @@ public static void updateCommittedTail(RuntimeLayout runtimeLayout,
...
+                        .getLogUnitClient(node)
+                        .getAllTails())
+                .map(CFUtils::getUninterruptibly)
+                .forEach(resp -> {
+                    // All responses should be computed on the same epoch
+                    checkArgument(resp.getEpoch() == runtimeLayout.getLayout().getEpoch());
+                    // Find the global max global tail and stream tails across all responses
+                    globalTail[0] = Long.max(resp.getLogTail(), globalTail[0]);
+                    resp.getStreamTails().forEach((k, v) -> streamTails.merge(k, v, Long::max));
+                });
```
WenbinZhu, line 296: Because of java stream's lazy evaluation property, if CFUtils.get is put in map(), then the futures are actually executed sequentially. We need to parallelize it to get better performance, by gathering all the futures first and then wait on them, similar to this: https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/util/Utils.java#L152-L162

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -307,34 +310,27 @@ static TailsResponse aggregateLogUnitTails(Set<TailsResponse> responses) {
...
+        // coalesce the candidates to unique nodes only
+        Set<String> segmentsHeadNodes = getChainHeadFromAllSegments(runtimeLayout.getLayout());
+        long[] globalTail = {Address.NON_EXIST};
+        final Map<UUID, StreamAddressSpace> streamsAddressSpace = new HashMap<>();
+        segmentsHeadNodes.stream()
+                .map(node -> runtimeLayout
+                        .getLogUnitClient(node)
+                        .getLogAddressSpace())
+                .map(CFUtils::getUninterruptibly)
+                .forEach(resp -> {
```
WenbinZhu, line 323: Same here.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -307,34 +313,27 @@ static TailsResponse aggregateLogUnitTails(Set<TailsResponse> responses) {
...
+                    // All responses should be computed on the same epoch
+                    //checkArgument(resp.getEpoch() == runtimeLayout.getLayout().getEpoch());
```
PavelZaytsev, line 328: Why did we comment this check out?

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Yea, I thought about it, but didnt think it was worth it.
Will change it anyways.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: That response doesn't actually has an epoch field. I fixed the API, un-commented and added tests.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -41,10 +41,36 @@ public StreamAddressSpace(long trimMark, Roaring64NavigableMap addressMap) {
...
+        this.addressMap = Roaring64NavigableMap.bitmapOf();
+        this.trimMark = Address.NON_ADDRESS;
+    }
+
     public Roaring64NavigableMap getAddressMap() {
         return addressMap;
     }
 
+    /**
+     * Merges b into a and returns b as the final result.
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/util/Utils.java
Maithem: Done.

Maithem:
All comments have been addressed.

runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -307,34 +319,29 @@ static TailsResponse aggregateLogUnitTails(Set<TailsResponse> responses) {
...
 
-        for (StreamsAddressResponse res : responses) {
-            logTail = Math.max(logTail, res.getLogTail());
-            streamAddressSpace = aggregateStreamAddressMap(res.getAddressMap(), streamAddressSpace);
-        }
-        return new StreamsAddressResponse(logTail, streamAddressSpace);
+        cfs.stream().map(CFUtils::getUninterruptibly)
+                .forEach(resp -> {
+                    // All responses should be computed on the same epoch
+                    checkArgument(resp.getEpoch() == runtimeLayout.getLayout().getEpoch());
```
WenbinZhu, line 336: This will throw IllegalArgumentException, but should be throw WrongEpochException to be more clear?

Maithem: good catch.

PR #2732, opened 2020-08-20
API Changes in CorfuStore::openTable and TxBuilder::update

pankti-m:
openTable() currently has two versions, depending on whether a table has been seen by the runtime.  This change consolidates the apis.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
@@ -15,6 +15,7 @@ import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.function.BiFunction;
 
 /**
  * TxBuilder() is a layer that aggregates mutations made to CorfuStore protobuf tables
@@ -82,7 +83,7 @@ public class TxBuilder {
      *
      * @param tableName Table name to perform the update on.
      * @param key       Key.
-     * @param value     Value to update.
+     * @param value     New value to update.
      * @param <K>       Type of Key.
      * @param <V>       Type of Value.
      * @return TxBuilder instance.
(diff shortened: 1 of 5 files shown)


WenbinZhu:
I think instead of removing the second openTable() API, we can rename it to getTable(), because it actually an existing fetch table.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -100,6 +101,37 @@ TxBuilder update(@Nonnull final String tableName,
...
+     * @param key       Key
+     * @param mutation  Function to apply to get the new value
+     * @param arg       Argument to pass to the mutation function
+     * @param <K>       Type of Key.
+     * @param <V>       Type of Value.
+     * @return TxBuilder instance
+     */
+    @Nonnull
+    public <K extends Message, V extends Message, M extends Message>
+    TxBuilder update(@Nonnull final String tableName,
```
hisundar, line 117: What is done here is interesting, it looks like the caller can actually specify a delta update and it is up to the database to apply the update.
Right now you have implemented it as a classic read-modify-write cycle. But in the future we could have corfu defer the application of the 'arg' or delta here and lazily apply them (say at compaction). This is a powerful feature provided by other databases and if the goal is to accomplish this functionality perhaps we can just rename this function to call it 
merge(), the Bifunction as the mergeOperator and the arg as the delta?
Later we can think of optimizing this in corfu to support lazy merge to defer the read in the read-modify-write cycle.

runtime/src/main/java/org/corfudb/runtime/collections/TxBuilder.java
```diff
@@ -100,6 +101,37 @@ TxBuilder update(@Nonnull final String tableName,
...
+                     @Nonnull final K key,
+                     @Nonnull BiFunction<V, V, V> mutation,
+                     @Nonnull final V arg,
+                     @Nullable final M metadata) {
+        Table<K, V, M> table = getTable(tableName);
+        V currentVal = table.get(key).getPayload();
+        if (currentVal == null) {
+            throw new IllegalArgumentException(
```
hisundar, line 125: This has implications on transaction abort - right now you want the transaction to abort on this illegalArgumentException, so perhaps you want to throw a runtime exception instead that wraps an abortCause?

hisundar (changes requested):
please feel free to abandon this change as it has been absorbed into this PR
https://github.com/CorfuDB/CorfuDB/pull/2766

pankti-m:
Closing as the change has been absorbed in PR #2766

PR #2736, opened 2020-08-23
Reset SinkBufferManager on negotiation

zhangn49:
## Overview

Description:

SinkBufferManager has a field `lastProcessedSeq` to apply entry and send ACK. It is initialized when the discovery service starts.

When a non-leader node acquires the lock, `lastProcessedSeq` is very likely outdated, and won't process new entries.

Why should this be merged: 

Fix bug.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
@@ -328,16 +328,22 @@ public class LogReplicationMetadataManager {
         TxBuilder txBuilder = corfuStore.tx(NAMESPACE);
         txBuilder.update(REPLICATION_STATUS_TABLE, key, val, null);
         txBuilder.commit();
+
+        log.trace("setReplicationRemainingEntries: clusterId: {}, remainingEntries: {}, type: {}",
+                clusterId, remainingEntries, type);
     }
 
     public Map<String, ReplicationStatusVal> getReplicationRemainingEntries() {
         Map<String, ReplicationStatusVal> replicationStatusMap = new HashMap<>();
         QueryResult<CorfuStoreEntry<ReplicationStatusKey, ReplicationStatusVal, ReplicationStatusVal>> entries =
-                corfuStore.query(NAMESPACE).executeQuery(REPLICATION_STATUS_TABLE, (x) -> {return true;});
+                corfuStore.query(NAMESPACE).executeQuery(REPLICATION_STATUS_TABLE, record -> true);
 
         for(CorfuStoreEntry<ReplicationStatusKey, ReplicationStatusVal, ReplicationStatusVal>entry : entries.getResult()) {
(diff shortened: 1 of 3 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -436,6 +436,7 @@ private void onLeadershipAcquire() {
                 // Standby Site : the LogReplicationServer (server handler) will initiate the LogReplicationSinkManager
                 log.info("Start as Sink (receiver)");
                 interClusterReplicationService.getLogReplicationServer().setLeadership(true);
+                interClusterReplicationService.getLogReplicationServer().getSinkManager().reset();
```
annym, line 439: should this rather be reset on negotiation? because imagine the case where you lost the lock for a short period or even a long period but writes did not progress, so your lastProcessSequence is still valid as you can continue from that point onwards?

zhangn49: Do you mean lost the lock for a short period and there is no negotiation during this period? So we don't need to reset in this case.

zhangn49: I see your point. Active won't know remote leader loss if there is no write... Moved it to negotiation

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -328,16 +328,22 @@ public void setReplicationRemainingEntries(String clusterId, long remainingEntri
...
         Map<String, ReplicationStatusVal> replicationStatusMap = new HashMap<>();
         QueryResult<CorfuStoreEntry<ReplicationStatusKey, ReplicationStatusVal, ReplicationStatusVal>> entries =
-                corfuStore.query(NAMESPACE).executeQuery(REPLICATION_STATUS_TABLE, (x) -> {return true;});
+                corfuStore.query(NAMESPACE).executeQuery(REPLICATION_STATUS_TABLE, record -> true);
 
         for(CorfuStoreEntry<ReplicationStatusKey, ReplicationStatusVal, ReplicationStatusVal>entry : entries.getResult()) {
             replicationStatusMap.put(entry.getKey().getClusterId(), entry.getPayload());
         }
+
+        log.debug("getReplicationRemainingEntries: replicationStatusMap size: {}", replicationStatusMap.size());
```
pankti-m, line 345: do we also need to print the number of remainingEntries?

zhangn49: I think number of remainingEntries is fine, because the consumer will only check it is 0 or not.

PR #2743, opened 2020-08-29
Performance Client Module + Utils

Maithem:
## Overview
Added a module to facilitate performance testing. 

Why should this be merged: Will be used for regression tests.

- [x] Public API has Javadoc

### DIFF
--- a/perf-simulation/pom.xml
+++ b/perf-simulation/pom.xml
@@ -14,6 +14,7 @@
     <properties>
         <jcommander.version>1.78</jcommander.version>
         <lombok.version>1.18.12</lombok.version>
+        <log4j2.version>2.10.0</log4j2.version>
     </properties>
 
     <dependencies>
@@ -32,6 +33,23 @@
             <artifactId>lombok</artifactId>
             <version>${lombok.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+            <version>${log4j2.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>${log4j2.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j-impl</artifactId>
+            <version>${log4j2.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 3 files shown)


xnull:
What motivation for having this module if we already have a ["universe"+"jmh" benchmark module](https://github.com/CorfuDB/corfudb-tests/blob/master/benchmarks/src/jmh/java/org/corfudb/benchmarks/cluster/ClusterBenchmark.java)

which can deploy and test the performance of a VM, docker and/or process cluster of Corfu

Maithem:
> What motivation for having this module if we already have a ["universe"+"jmh" benchmark module](https://github.com/CorfuDB/corfudb-tests/blob/master/benchmarks/src/jmh/java/org/corfudb/benchmarks/cluster/ClusterBenchmark.java)
> 
> which can deploy and test the performance of a VM, docker and/or process cluster of Corfu

`jmh` is a  `microbenchmark` tool, that's only one aspect of performance testing.  You can read about the objectives of a microbenchmark here https://stackoverflow.com/questions/2842695/what-is-microbenchmarking

You've might have noticed that in jmh you cant specify that number of times a method is executed within an iteration for throughput mode. That's because that number is derived from the iteration time period to improve the sampling quality. For example, one can ask how fast can 100 threads write a billion writes to the system and we wouldn't be able to do it on jmh, or more accurately, it's not designed to provide data for those types of questions.

This module will provide tests that simulate various types of workloads to measure end-to-end performance.

xnull:
>You can read about the objectives of a microbenchmark here 

If you take a look at a [ClusterBenchmark](https://github.com/CorfuDB/corfudb-tests/blob/master/benchmarks/src/jmh/java/org/corfudb/benchmarks/cluster/ClusterBenchmark.java) you will see that JMH successfully used not only for micro benchmarking, it works well, powerful and flexible solution and can be used in many cases despite of "micro" in its name

>You've might have noticed that in jmh you cant specify that number of times a method is executed within an iteration for throughput mode.

why not? you can execute as many methods as many times as you want within an iteration, nothing prevents from that.

> one can ask how fast can 100 threads write a billion writes to the system and we wouldn't be able to do it on jmh
if your 100 threads write 1k operations per second, guess how much time it takes to write 1 billion records?

I personally don't see why JMH is not enough in our case, your arguments not clear to me, all the cases could be easily run and covered.

xnull:
> `jmh` is a `microbenchmarking` tool and that's only one aspect of performance testing.

Btw, https://openjdk.java.net/projects/code-tools/jmh/
>>JMH is a Java harness for building, running, and analysing nano/micro/milli/macro benchmarks written in Java and other languages targetting the JVM.

PR #2746, opened 2020-08-31
StreamAddressSpace Cleanup

Maithem:
## Overview
This patch hides the implementation type of the bitmap used by
StreamAddressSpace. This allows for changing the bitmap implementation
easily.

Why should this be merged: Improves code organization and allows for future optimizations

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -3,35 +3,6 @@ package org.corfudb.infrastructure;
 import com.codahale.metrics.Timer;
 import com.google.common.collect.ImmutableMap;
 import io.netty.channel.ChannelHandlerContext;
-import lombok.Builder;
-import lombok.Builder.Default;
-import lombok.Getter;
-import lombok.Setter;
-import lombok.extern.slf4j.Slf4j;
-
-import org.corfudb.runtime.view.stream.StreamAddressSpace;
-import org.corfudb.protocols.wireprotocol.StreamAddressRange;
-import org.corfudb.protocols.wireprotocol.StreamsAddressRequest;
-import org.corfudb.protocols.wireprotocol.StreamsAddressResponse;
-import org.roaringbitmap.longlong.Roaring64NavigableMap;
-import org.corfudb.infrastructure.SequencerServerCache.ConflictTxStream;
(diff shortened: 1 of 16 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -670,15 +668,14 @@ private void handleStreamsAddressRequest(CorfuPayloadMsg<StreamsAddressRequest>
...
+        for (StreamAddressRange range : addressRanges) {
+            UUID streamId = range.getStreamID();
             // Get all addresses in the requested range
             if (streamsAddressMap.containsKey(streamId)) {
-                addressMap = streamsAddressMap.get(streamId).getAddressesInRange(streamAddressRange);
-                requestedAddressSpaces.put(streamId,
-                        new StreamAddressSpace(streamsAddressMap.get(streamId).getTrimMark(), addressMap));
+                StreamAddressSpace streamSubRange = streamsAddressMap.get(streamId)
+                        .getRange(range.getStart(), range.getEnd());
+                requestedAddressSpaces.put(streamId, streamSubRange);
```
Maithem, line 678: Caller needs to reverse this range.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -61,65 +125,76 @@ public static StreamAddressSpace merge(StreamAddressSpace a, StreamAddressSpace
...
      * Copy this stream's addresses to a set, under a given boundary (inclusive).
      *
      * @param maxGlobal maximum address (inclusive upper bound)
      */
-    public NavigableSet<Long> copyAddressesToSet(final Long maxGlobal) {
-        NavigableSet<Long> queue = new TreeSet<>();
-        this.addressMap.forEach(address -> {
+    public NavigableSet<Long> getAddressesUpTo(final Long maxGlobal) {
+        NavigableSet<Long> subset = new TreeSet<>();
+        bitmap.forEach(address -> {
```
Maithem, line 156: Using `Long` instead of `long` is silly, but required to force compilation errors when code tries add `int`, although that should happen, silent failures can happen due to casting.

Hopefully the compiler can recognize that these `Long` objects are not needed and hences replaces with the primitive type.

PR #2747, opened 2020-08-31
Optimized Bitmap Range Operations

Maithem:
## Overview
Improved StreamAddressSpace range operations by removing iterative implementations and replaced it with logical "native" bitmap operations. 

Why should this be merged: Improves performance + reduces memory pressure

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
@@ -1,17 +1,21 @@
 package org.corfudb.runtime.view.stream;
 
-import lombok.extern.slf4j.Slf4j;
-import org.corfudb.protocols.wireprotocol.StreamAddressRange;
-import org.corfudb.runtime.view.Address;
-import org.corfudb.util.Utils;
-import org.roaringbitmap.longlong.LongIterator;
-import org.roaringbitmap.longlong.Roaring64NavigableMap;
-
+import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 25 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -164,46 +239,61 @@ public void trim(Long trimMark) {
...
     }
 
     public long getHighestAddress() {
-        if (addressMap.isEmpty()) {
+        if (bitmap.isEmpty()) {
             return Address.NON_EXIST;
         }
 
-        return addressMap.getReverseLongIterator().next();
+        return bitmap.getReverseLongIterator().next();
```
Maithem, line 285: Too expensive, replace this with local longs that track the max/min

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -139,14 +214,14 @@ public void trim(Long trimMark) {
         // Note: if a negative value is passed to this API the cardinality
         // of the bitmap is returned, which would be incorrect as we would
         // be removing all addresses upon an invalid trim mark.
-        long numAddressesToTrim = addressMap.rankLong(trimMark);
+        long numAddressesToTrim = bitmap.rankLong(trimMark);
```
Maithem, line 217: rank operations are expensive, replace this with min/max pointers.

PR #2758, opened 2020-09-04
Non-Transactional Access Trim Exception Handling

annym:
## Overview

Description:

   Handle sync of non-transactional access under lock even in the event
of TrimmedExceptions, otherwise, thread interleaving may lead to inconsistent state.

Why should this be merged: existing bug which manifests in intermittent data loss.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import java.lang.reflect.InvocationTargetException;
 import java.time.Duration;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
@@ -184,31 +185,35 @@ public class CorfuCompileProxy<T extends ICorfuSMR<T>> implements ICorfuSMRProxy
             }
         }
 
+        // Linearize this read against a timestamp
+        AtomicLong timestamp = new AtomicLong(rt.getSequencerView().query(getStreamID()));
+
+        log.debug("Access[{}] conflictObj={} version={}", this, conflictObject, timestamp);
+
         // Perform underlying access
-        for (int x = 0; x < rt.getParameters().getTrimRetry(); x++) {
-            // Linearize this read against a timestamp
-            final long timestamp = rt.getSequencerView()
-                            .query(getStreamID());
-            log.debug("Access[{}] conflictObj={} version={}", this, conflictObject, timestamp);
+        return underlyingObject.access(o -> o.getVersionUnsafe() >= timestamp.get()
+                        && !o.isOptimisticallyModifiedUnsafe(),
+                o -> {
+                    for (int x = 0; x < rt.getParameters().getTrimRetry(); x++) {
+                        try {
+                            o.syncObjectUnsafe(timestamp.get());
+                            break;
+                        } catch (TrimmedException te) {
+                            log.info("accessInner: Encountered trimmed address space " +
+                                            "while accessing version {} of stream {} on attempt {}",
+                                    timestamp.get(), getStreamID(), x);
 
-            try {
-                return underlyingObject.access(o -> o.getVersionUnsafe() >= timestamp
-                                && !o.isOptimisticallyModifiedUnsafe(),
-                        o -> o.syncObjectUnsafe(timestamp),
-                        o -> accessMethod.access(o));
-            } catch (TrimmedException te) {
-                log.info("accessInner: Encountered trimmed address space " +
-                    "while accessing version {} of stream {} on attempt {}",
-                    timestamp, getStreamID(), x);
-                // We encountered a TRIM during sync, reset the object
-                underlyingObject.update(o -> {
-                    o.resetUnsafe();
-                    return null;
-                });
-            }
-        }
+                            o.resetUnsafe();
 
-        throw new TrimmedException();
+                            if (x == (rt.getParameters().getTrimRetry() - 1)) {
+                                throw te;
+                            }
+
+                            timestamp.set(rt.getSequencerView().query(getStreamID()));
+                        }
+                    }
+                },
+                o -> accessMethod.access(o));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -184,31 +185,34 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
...
-        throw new TrimmedException();
+        // Perform underlying access
+        return underlyingObject.access(o -> o.getVersionUnsafe() >= timestamp.get()
+                        && !o.isOptimisticallyModifiedUnsafe(),
+                o -> {
+                    for (int x = 0; x < rt.getParameters().getTrimRetry(); x++) {
+                        try {
+                            timestamp.set(rt.getSequencerView()
+                                    .query(getStreamID()));
+                            o.syncObjectUnsafe(timestamp.get());
```
Maithem, line 201: You need a `break` here otherwise it will run 2 times on every access.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -184,31 +185,34 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
...
+        log.debug("Access[{}] conflictObj={} version={}", this, conflictObject, timestamp);
 
-        throw new TrimmedException();
+        // Perform underlying access
+        return underlyingObject.access(o -> o.getVersionUnsafe() >= timestamp.get()
+                        && !o.isOptimisticallyModifiedUnsafe(),
+                o -> {
+                    for (int x = 0; x < rt.getParameters().getTrimRetry(); x++) {
+                        try {
+                            timestamp.set(rt.getSequencerView()
```
Maithem, line 199: Move this to the catch block otherwise you'll have two sequencer queries per access.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -184,31 +185,34 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
...
-                underlyingObject.update(o -> {
-                    o.resetUnsafe();
-                    return null;
-                });
-            }
-        }
+        log.debug("Access[{}] conflictObj={} version={}", this, conflictObject, timestamp);
 
-        throw new TrimmedException();
+        // Perform underlying access
```
Maithem, line 193: Btw, `getUpcallResultInner` has the same problem. The reset in the catch block should be part of the update lambda to be executed under the same lock.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
Maithem: FYI: This is different than like 201. Essentelly each non-transcation access will produce 3 sequence calls.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
annym: Good catch, changed it, please take a look.

Maithem (approved):

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -280,39 +285,39 @@ private long logUpdateInner(String smrUpdateFunction, final boolean keepUpcallRe
...
+                                    + " and we don't have a copy");
+                        } catch (TrimmedException ex) {
+                            log.info("getUpcallResultInner: Encountered trimmed address space " +
+                                            "while accessing version {} of stream {} on attempt {}",
+                                    timestamp, getStreamID(), x);
+                            // We encountered a TRIM during sync, reset the object
+                            underlyingObject.update(obj -> {
+                                obj.resetUnsafe();
+                                return null;
+                            });
```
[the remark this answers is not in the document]
annym: good catch, done.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -280,39 +285,39 @@ private long logUpdateInner(String smrUpdateFunction, final boolean keepUpcallRe
...
-                // We encountered a TRIM during sync, reset the object
-                underlyingObject.update(o -> {
-                    o.resetUnsafe();
-                    return null;
-                });
-            }
-        }
-
-        throw new TrimmedUpcallException(timestamp);
+            throw new TrimmedUpcallException(timestamp);
```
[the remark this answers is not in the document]
annym: Done, addressed all comments.

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -280,39 +285,36 @@ private long logUpdateInner(String smrUpdateFunction, final boolean keepUpcallRe
...
```
pankti-m, line 316: nit - indentation

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -453,7 +453,7 @@ public void setUncommittedChanges(WriteSetSMRStream optimisticStream) {
      *
      * @return Returns the pointer position to the object in the stream.
      */
-    public long getVersionUnsafe() {
+    public long  getVersionUnsafe() {
```
pankti-m, line 456: nit - extra ws

runtime/src/main/java/org/corfudb/runtime/object/CorfuCompileProxy.java
```diff
@@ -184,31 +185,35 @@ public CorfuCompileProxy(CorfuRuntime rt, UUID streamID, Class<T> type, Object[]
...
+                                            "while accessing version {} of stream {} on attempt {}",
+                                    timestamp.get(), getStreamID(), x);
+
+                            o.resetUnsafe();
+
+                            if (x == (rt.getParameters().getTrimRetry() - 1)) {
+                                throw te;
+                            }
+
+                            timestamp.set(rt.getSequencerView().query(getStreamID()));
```
pankti-m, line 212: why do we need to set the timestamp here?  Is it expected to change after it was obtained in line 189?

annym: exactly, the state of the log might have changed and you want to provide the latest state.

WenbinZhu (approved):

pankti-m (approved):

Issue #2697, opened 2020-08-11
 StateTransferTest.restoreRedundancyOldAndNewTrimMarksParallelTransfer


annym:
## Overview

Please take a look at this intermittently failing test: [https://api.travis-ci.org/v3/job/716756834/log.txt](https://api.travis-ci.org/v3/job/716756834/log.txt)

[ERROR]   StateTransferTest.restoreRedundancyOldAndNewTrimMarksParallelTransfer:1475 expected:<1[5]1L> but was:<1[2]1L>

PR #2764, opened 2020-09-08
Fix the trimmed and not transferred segment case in S/T

### LINKED ISSUE #2731: Split segments that are completely trimmed can make state transfer stuck forever
## Overview
s1: [0 - 100]: A
s2: [100 - 200]: A, B, C
s3: [200 - inf]: A, B, C

trim mark: 150

Segment 1 one has node A
Segment 2 and Segment 3 has all the nodes. 
Trim Mark has moved to 150 (past the end of segment 1). 

Node A does not need to call state transfer because it has full redundancy. 
Nodes B and C call state transfer because they are not in the segment s1. However, when they retrieve a trim mark, it turns out they don't need to perform any state transfer because the entire segment s1 was trimmed. They do not update the layout either because they have not performed a state transfer, so the state transfer keeps looping forever.

PavelZaytsev:
## Overview

Description:
- This patch includes the trimmed and not restored segments in the layout restoration workflow (see the linked issue below). 
- Also bug fix in one of the old tests

Closes  #2731, #2697

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/redundancy/RedundancyCalculator.java
```diff
@@ -222,6 +223,38 @@ public static boolean canRestoreRedundancyOrMergeSegments(Layout layout, String
...
+     * TRANSFERRED, eligible for the layout redundancy restoration.
+     *
+     * @param layout   A current layout.
+     * @param trimMark A current global trim mark.
+     * @return A list of transfer segments.
+     */
+    public ImmutableList<TransferSegment> getTrimmedNotRestoredSegments(Layout layout, long trimMark) {
+        return layout.getSegments()
+                .stream()
+                .filter(segment -> segment.getEnd() <= trimMark && !segmentContainsServer(segment, getServer()))
```
xnull, line 240: maybe use two filter methods?
```
.filter(segment -> segment.getEnd() <= trimMark)
.filter(segment ->  !segmentContainsServer(segment, getServer()))
```

xnull (approved):
lgtm

WenbinZhu (approved):

infrastructure/src/test/java/org/corfudb/infrastructure/redundancy/RedundancyCalculatorTest.java
```diff
@@ -556,4 +558,47 @@ public void testPrepareTransferWorkloadCommittedTailSplitsSegment() {
                 s -> s.equals(testSegments.get(1).getStatus()))).isTrue();
     }
 
+    @Test
+    public void testGetTrimmedNotRestoredSegments() {
+        Layout.ReplicationMode replicationMode = CHAIN_REPLICATION;
+        LayoutStripe layoutStripe1 = new LayoutStripe(ImmutableList.of("A"));
+        LayoutSegment segment1 = new LayoutSegment(replicationMode, 0L, 100L,
```
pankti-m, line 565: Even if addresses before the trim mark have been trimmed, they are on disk right?  How is that data restored on B and C?

PavelZaytsev: There is no real reason to transfer trimmed addresses because logically they are trimmed. The starting address has moved and is past the trim mark. So whether they are present on disk or not, we check it either way amd get the same result:
```
private boolean isTrimmed(long address) {
        return address < dataStore.getStartingAddress();
    }
```

infrastructure/src/main/java/org/corfudb/infrastructure/redundancy/RedundancyCalculator.java
PavelZaytsev: Usually, when there are multiple consecutive HOFs I think it's better to merge them into one HOF. For example,
```
.map(x -> x + 1)
.map(x -> x * 2) 
```
I think is better expressed as `.map(x -> (x + 1) * 2)`

PR #2765, opened 2020-09-09
Use A PRNG for ID Generation

Maithem:
## Overview

In CorfuDB UUIDs are mostly used for logging purposes, thus true
randomness is not required. Using UUID.randomUUID for transaction
IDs can be really slow because it uses secureRandom. This patch
uses a PRNG to generate those ids much faster.

Why should this be merged: Reduces the `Transaction::begin` time by up to 50+ milliseconds 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/Utils.java
+++ b/runtime/src/main/java/org/corfudb/util/Utils.java
@@ -14,6 +14,7 @@ import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 import jdk.internal.org.objectweb.asm.util.Printer;
@@ -44,6 +45,19 @@ public class Utils {
     private static final TraceMethodVisitor mp = new TraceMethodVisitor(printer);
 
     private static final char[] hexArray = "0123456789ABCDEF".toCharArray();
+
+
+    /**
+     * When true randomness is not required using UUID.randomUUID() can be really slow.
+     * Blocking for 50+ ms for entropy to build up is not unusual. This method generates
+     * random UUIDs PRNG.
+     * @return A pseudo-random UUID
+     */
+    public static UUID genPseudorandomUUID() {
+        long msb = ThreadLocalRandom.current().nextLong();
+        long lsb = ThreadLocalRandom.current().nextLong();
+        return new UUID(msb, lsb);
+    }
 
     /** Convert a byte array to a hex string.
      * Source:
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/util/Utils.java
```diff
@@ -45,6 +46,19 @@ private Utils() {
...
     private static final char[] hexArray = "0123456789ABCDEF".toCharArray();
 
+
+    /**
+     * When true randomness is not required using UUID.randomUUID() can be really slow.
+     * Blocking for 50+ ms for entropy to build up is not unusual. This method generates
+     * random UUIDs based on a LCG with a thread local seed.
+     * @return A psuedorandom
+     */
+    public static UUID genPseudorandomUUID() {
```
xnull, line 56: >Using UUID.randomUUID for transaction IDs can be really slow because it uses secureRandom.
```
Benchmark                 Mode         Cnt                 Score 
javaUUIDGenerator         thrpt     2_613_099           ops/s
```
UUID.randomUUID() is actually 100k times faster than 50ms

Maithem: Run the benchmark with `-Djava.security.egd=file:/dev/random` I think macos uses urandom and not random as a default. urandom is psuedorandom so it won't block. 

SecureRandom consumes bytes from  `/proc/sys/kernel/random/entropy_avail`. Also, In a real system multiple threads consume bytes from the entropy pool, so having a single thread consume the bytes sequentially is not a good benchmark.

Maithem: It's hard to repro without using the bouncy castle provider

image

zfrenette (commented):
How much of a difference does this make in the grand scheme of things? Although I agree that PRNGs are faster when they don't need to be cryptographically secure, is the improvement significant enough to risk generating duplicate UUIDs? I realize that the transaction ID isn't used in the Sequencer beyond debugging purposes, but I am not so sure about the event ID in LR.

Maithem:
> How much of a difference does this make in the grand scheme of things? 

I've seen profiler snapshots where it can block for ~50ms + 

> is the improvement significant enough to risk generating duplicate UUIDs? 

The difference between true randomness and pseudorandomness is not uniqueness but predictability. For example, a truly random number generator has a higher probability of collision generating two numbers from 1 to 10 as opposed to PRNG that generates numbers sequentially, since the latter has a 0 probability of collision.

zhangn49 (approved):
I still don't think it takes 28ms for a single `getEntropy`... But agree with that no need to compete for the entropy pool

pankti-m (approved):

Maithem:
> I still don't think it takes 28ms for a single `getEntropy`... But agree with that no need to compete for the entropy pool

It's possible that the profiler snapshot is inaccurate. Reproducing it is kind of tricky because there's a timing and deployment configuration aspects that are hard to replicate. Just googling `secureRandom slow on linux` you can see that the possibility of blocking for milliseconds is not uncommon.

Anyways, having true randomness doesn't provide a stronger guarantee of uniqueness see my comment here https://github.com/CorfuDB/CorfuDB/pull/2765#issuecomment-690943554

PR #2768, opened 2020-09-10
Add Test for Single Listener to Multiple Tables in StreamingIT.

pankti-m:
We do not have a test where a single listener subscribes to multiple tables
and verifies that updates made to those tables in a single transaction are
received in the same CorfuStreamEntry.

Why should this be merged:  Test a different usecase of the streaming layer

### DIFF
--- a/test/src/test/java/org/corfudb/integration/StreamingIT.java
+++ b/test/src/test/java/org/corfudb/integration/StreamingIT.java
@@ -17,6 +17,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
@@ -278,7 +279,7 @@ public class StreamingIT extends AbstractIT {
         store.subscribe(s2n2t1, "n2",
                 Collections.singletonList(new TableSchema("t1", Uuid.class, Uuid.class, Uuid.class)), ts1);
 
-        // After a brief wait verify that the listener go all the updates.
+        // After a brief wait verify that the listener gets all the updates.
         TimeUnit.SECONDS.sleep(2);
         LinkedList<CorfuStreamEntries> updates = s1n1t1.getUpdates();
         assertThat(updates.size()).isEqualTo(numUpdates);
@@ -308,7 +309,7 @@ public class StreamingIT extends AbstractIT {
             assertThat(entry.get(0).getMetadata()).isEqualTo(uuid);
         }
 
-        // Now clear a table and make sure that the clear gets propogated
+        // Now clear a table and make sure that the clear gets propagated
         n1t1.clear();
 
         // After a brief pause validate that the listener obtains the clear() table SRM update.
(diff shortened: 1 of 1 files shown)


hisundar (approved):
LGTM

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -17,9 +17,7 @@
 import org.junit.Test;
 
 import java.io.IOException;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
```
WenbinZhu, line 20: Expand.

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -323,4 +321,67 @@ public void testStreamingMultiTableStreams() throws Exception {
...
+
+        // Subscribe to both tables
+        List<TableSchema> tablesSubscribed = new ArrayList<>();
+        tablesSubscribed.add(new TableSchema("t1", Uuid.class, Uuid.class, Uuid.class));
+        tablesSubscribed.add(new TableSchema("t2", Uuid.class, Uuid.class, Uuid.class));
+        StreamListenerImpl listener = new StreamListenerImpl("n1_listener");
+        store.subscribe(listener, "n1", tablesSubscribed, ts1);
+
+        // Verify that both updates come to the subscriber in the same StreamEntry
+        TimeUnit.SECONDS.sleep(2);
```
WenbinZhu, line 380: Why sleeping here? Is the test relying on the sleep to be passed? If not we can remove it, otherwise it's better to make it synchronous instead of sleeping.

WenbinZhu (approved):
LGTM, two comments.

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -323,4 +321,67 @@ public void testStreamingMultiTableStreams() throws Exception {
...
+        LinkedList<CorfuStreamEntries> updates = listener.getUpdates();
+        assertThat(updates.size()).isEqualTo(1);
+        assertThat(updates.getFirst().getEntries().entrySet().size()).isEqualTo(2);
+
```
WenbinZhu, line 384: Maybe add a check for the content of the update?

test/src/test/java/org/corfudb/integration/StreamingIT.java
pankti-m: Done

test/src/test/java/org/corfudb/integration/StreamingIT.java
pankti-m: there is an async thread which delivers the updates every 50ms.  Changing it to be synchronous is a bigger change so keeping the sleep for now.

test/src/test/java/org/corfudb/integration/StreamingIT.java
pankti-m: Done

Issue #2748, opened 2020-08-31
Cluster Id Validation Is Expensive


Maithem:
**Bug report.**
Accessing the local data store on every message to validate the cluster id is really expensive because its serialized.
The RPC path should not participate in any expensive synchronization.

WenbinZhu:
ClusterId is get from currentLayout, which most of the time is cached in memory, rigtht?

Maithem:
The layout itself is cached, but its access with a `synchronized`

```java

@Override
    public synchronized <T> T get(KvRecord<T> key) {
        String path = key.getFullKeyName();
        Object val = cache.get(path, k -> {
            if (!inMem) {
                T loadedVal = load(key.getDataType(), path);
                if (loadedVal != null) {
                    return loadedVal;
                }
            }

            // We need to maintain a path -> null mapping for keys that were loaded, but
            // were empty. This is required to prevent loading an empty key more than once, which is expensive.
            return NullValue.NULL_VALUE;
        });

        return val == NullValue.NULL_VALUE ? null : (T) val;
    }

```

WenbinZhu:
@Maithem I see, can we remove synchronized? cache.get() seems thread safe

Maithem:
@WenbinZhu Yea, I think all of the `synchronized` methods in `DataStore` can be removed.

WenbinZhu:
I have a PR to improve local data store serialization performance, maybe I can resurrect that patch and include this `synchronized` cleanup.

Maithem:
@WenbinZhu yea that would be great. Re-assigned.

Issue #2133, opened 2019-08-28
Local DataStore has redundant byte copies which are not efficient.


WenbinZhu:
## Overview

Local DataStore is doing some redundant byte copies in both read and write path.

https://github.com/CorfuDB/CorfuDB/blob/feature_sparse_trim/infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java#L135

https://github.com/CorfuDB/CorfuDB/blob/feature_sparse_trim/infrastructure/src/main/java/org/corfudb/infrastructure/DataStore.java#L179

PR #2769, opened 2020-09-12
Local datastore read/write performance improvement.

WenbinZhu:
## Overview

- Removed all synchronized keywords.
- Current cache writer has to do two buffer copies before the actual
write syscall happens, first time copy to a heap buffer to include
checksum, then java FileChannle will copy it again to a direct buffer
before performing writes. This patch instead uses a direct buffer for
the first copy to eliminate the second copy.

Related issue(s) (if applicable): #2133, #2748 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
```diff
@@ -93,18 +97,15 @@ public DataStore(@Nonnull Map<String, Object> opts,
         return Caffeine.newBuilder().build(k -> null);
     }
 
-
-    public static int getChecksum(byte[] bytes) {
+    private static int getChecksum(byte[] bytes) {
```
Maithem, line 100: You can use `putBytes(ByteBuffer bytes)` directly

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
```diff
@@ -129,12 +133,19 @@ public synchronized void write(@Nonnull String key, @Nonnull Object value) {
                             String jsonPayload = JsonUtils.parser.toJson(value, value.getClass());
                             byte[] bytes = jsonPayload.getBytes();
```
Maithem, line 134: `String.getBytes` will create a new buffer, you can eliminate it by using the String instead.

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
```diff
@@ -129,12 +133,19 @@ public synchronized void write(@Nonnull String key, @Nonnull Object value) {
...
 
-                            ByteBuffer buffer = ByteBuffer.allocate(bytes.length
-                                    + Integer.BYTES);
-                            buffer.putInt(getChecksum(bytes));
-                            buffer.put(bytes);
-                            Files.write(tmpPath, buffer.array(), StandardOpenOption.CREATE,
+                            buffer = Unpooled.directBuffer(bytes.length + Integer.BYTES);
+                            buffer.writeInt(getChecksum(bytes));
+                            buffer.writeBytes(bytes);
```
Maithem, line 138: To eliminate the other buffer copy you can just use `buffer.writeCharSequence(jsonPayload, Charset.defaultCharset());` , or maybe explicitly set it to UTF-8

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
```diff
@@ -143,27 +154,28 @@ public synchronized void write(@Nonnull String key, @Nonnull Object value) {
...
+                        // Removal of underlying file is done by cleanupTask.
                     }
                 })
-                .maximumSize(dsCacheSize)
                 .build();
     }
 
     private <T> T load(Class<T> tClass, String key) {
         try {
+            System.out.println(1111);
```
Maithem, line 178: This is failing the build.

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
```diff
@@ -223,7 +233,7 @@ public synchronized void delete(@Nonnull String key,
     }
 
     @Override
-    public synchronized <T> void delete(KvRecord<T> key) {
+    public <T> void delete(KvRecord<T> key) {
         cache.invalidate(key.getFullKeyName());
     }
```
Maithem, line 238: Calling delete and not doing anything is a little weird. I would rather just delete the method from the interface. It's not used anyways.

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
WenbinZhu: There are some implementation problems if we do buffer.writeCharSequence. 1) depending on the charset, the final capacity of byte buffer is usually not equal to string.length(), so there might be overheads to scale the the initial capacity. 2) the  checksum needs to be serialized before the payload, so we need to rewind the index and rewrite checksum, and the hasher.putBytes(ByteBuffer bytes) takes a nio buffer instead of a netty bytebuf, which requires more index manipulation, seems a bit unclean and error prone...

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
WenbinZhu: good catch! removed.

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
WenbinZhu: Delete will remove the in-memory cache entry, but not remove the underlying file. This is being used in tests.

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
WenbinZhu: See below comments.

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
WenbinZhu: See below comments.

WenbinZhu:
superseded by #2769

WenbinZhu:
Closed by mistake, reopening.

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
Maithem: It's only used in the test. I think `CacheWriter` is not needed here, you can probably use a loading cache. 
If you really want to keep it then maybe call it `invalidate.` 

```java
/**
   * Deletes the value corresponding to the {@code key} from the external resource. The cache will
   * communicate a delete when the entry is explicitly removed or evicted.
   *
   * @param key the non-null key whose value was removed
   * @param value the value associated with {@code key}, or {@code null} if collected
   * @param cause the reason for which the entry was removed
   * @throws RuntimeException or Error, in which case the mapping is unchanged
   */
  void delete(@Nonnull K key, @Nullable V value, @Nonnull RemovalCause cause);

```

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
Maithem: Its actually not variable for `Layout.java`, it can be fixed to UTF-8. You're already using a direct buffer.

Maithem (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/datastore/DataStore.java
WenbinZhu: @Maithem I think the reason that we need a CacheWriter is because the write to cache and disk needs to be atomic. And since the interface is a generic kv datastore instead of a cache, maybe it's better to still call it delete?

Maithem: I see. You can just use the cache as a concurrent hash map if you need per-key atomicity. My only point is why keep it if you dont need it.

WenbinZhu: Yea, right now we need this interface for tests, and since it's a generic kv store interface, maybe later there will be another impl that actually does delete for purposes other than tests.

Lujie1996 (approved):

zfrenette (approved):

PR #2772, opened 2020-09-17
Store Snapshot Sync Info in Sync Status table

zhangn49:
## Overview

Description:

Added return value for force sync

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/proto/log_replication_metadata.proto
+++ b/infrastructure/proto/log_replication_metadata.proto
@@ -1,5 +1,8 @@
 syntax = "proto3";
+
 package org.corfudb.infrastructure.logreplication.proto;
+
+import "google/protobuf/timestamp.proto";
 
 /**
  * This is used by replication metadata table in corfu store.
@@ -44,17 +47,44 @@ message ReplicationStatusVal {
     SNAPSHOT = 0;
     LOG_ENTRY = 1;
   }
-  SyncType type = 3;
+  SyncType syncType = 3;
+  SyncStatus status = 4;
+  SnapshotSyncInfo snapshotSyncInfo = 5;
 }
 
-enum ReplicationEventType {
-  FORCE_SNAPSHOT_SYNC = 0;
+/*
+ * Snapshot Sync Info
+ *
+ * If replication is in SNAPSHOT sync, this provides details of the
+ * ongoing snapshot sync. If replication is in LOG ENTRY sync (delta).
+ * this provides details of the previous SNAPSHOT sync leading up to
+ * the log entry sync.
+ */
+message SnapshotSyncInfo {
+  enum SnapshotSyncType {
+    DEFAULT = 0;
+    FORCED = 1;
+  }
+  SnapshotSyncType type = 1;
+  SyncStatus status = 2;
+  string snapshotRequestId = 3;
+  google.protobuf.Timestamp completedTime = 4;
+  uint64 baseSnapshot = 5;
+}
+
+enum SyncStatus {
+  ONGOING = 0;
+  COMPLETED = 1;
+  ERROR = 2;
+  NOT_STARTED = 3;
+  STOPPED = 4;
 }
 
 message ReplicationEventKey {
   string key = 1;
 }
+
 message ReplicationEvent {
-  ReplicationEventType type = 1;
-  string clusterId = 2;
+  string clusterId = 1;
+  string eventId = 2;
 }
--- a/infrastructure/proto/log_replication_metadata.proto
+++ b/infrastructure/proto/log_replication_metadata.proto
@@ -72,6 +72,14 @@ message SnapshotSyncInfo {
   uint64 baseSnapshot = 5;
 }
 
+/*
+ * SyncStatus
+ *
+ * COMPLETED:   Used in SnapshotSyncInfo only for the latest snapshot sync
+ * ERROR:       Log Replication is unrecoverable, need to restart
+ * NOT_STARTED: Log Replication might be in the very beginning state
+ * STOPPED:     Log Replication is in initialized state
+ */
 enum SyncStatus {
   ONGOING = 0;
   COMPLETED = 1;
--- a/infrastructure/proto/log_replication_metadata.proto
+++ b/infrastructure/proto/log_replication_metadata.proto
@@ -85,6 +85,10 @@ message ReplicationEventKey {
 }
 
 message ReplicationEvent {
+  enum ReplicationEventType {
+    FORCE_SNAPSHOT_SYNC = 0;
+  }
   string clusterId = 1;
   string eventId = 2;
+  ReplicationEventType type = 3;
 }
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -708,7 +708,8 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      */
     private void processEnforceSnapshotSync(DiscoveryServiceEvent event) {
         if (replicationManager == null || !isLeader.get()) {
-            log.warn("The current node is not the leader will skip doing the snapshot full sync");
+            log.warn("The current node is not the leader, will skip doing the " +
+                    "forced snapshot sync with id {}", event.getEventId());
             return;
         }
 
@@ -759,14 +760,16 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
             throw new LogReplicationDiscoveryServiceException(errorStr);
         }
 
-        log.info("Received the forceSnapshotSync command for standby cluster {}", clusterId);
+        UUID forceSyncId = UUID.randomUUID();
+        log.info("Received the forceSnapshotSync command for standby cluster {}, forced sync id {}",
+                clusterId, forceSyncId);
 
         // Write to the event to the event corfu table
-        UUID forceSyncId = UUID.randomUUID();
         ReplicationEventKey key = ReplicationEventKey.newBuilder().setKey(System.currentTimeMillis() + " "+ clusterId).build();
         ReplicationEvent event = ReplicationEvent.newBuilder()
                 .setClusterId(clusterId)
                 .setEventId(forceSyncId.toString())
+                .setType(ReplicationEvent.ReplicationEventType.FORCE_SNAPSHOT_SYNC)
                 .build();
         getLogReplicationMetadataManager().updateLogReplicationEventTable(key, event);
         return forceSyncId;
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -764,7 +764,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
         log.info("Received the forceSnapshotSync command for standby cluster {}, forced sync id {}",
                 clusterId, forceSyncId);
 
-        // Write to the event to the event corfu table
+        // Write an force sync event to the logReplicationEventTable
         ReplicationEventKey key = ReplicationEventKey.newBuilder().setKey(System.currentTimeMillis() + " "+ clusterId).build();
         ReplicationEvent event = ReplicationEvent.newBuilder()
                 .setClusterId(clusterId)
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -93,7 +93,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      * Adapter for cluster discovery service
      */
     @Getter
-    private CorfuReplicationClusterManagerAdapter clusterManagerAdapter;
+    private final CorfuReplicationClusterManagerAdapter clusterManagerAdapter;
 
     /**
      * Defines the topology, which is discovered through the Cluster Manager
@@ -133,11 +133,11 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     /**
      * Callback to Log Replication Server upon topology discovery
      */
-    private CompletableFuture<CorfuInterClusterReplicationServerNode> serverCallback;
+    private final CompletableFuture<CorfuInterClusterReplicationServerNode> serverCallback;
 
     private CorfuInterClusterReplicationServerNode interClusterReplicationService;
 
-    private ServerContext serverContext;
+    private final ServerContext serverContext;
 
     private String localCorfuEndpoint;
 
@@ -148,7 +148,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private boolean shouldRun = true;
 
     @Getter
-    private volatile AtomicBoolean isLeader;
+    private final AtomicBoolean isLeader;
 
     private LogReplicationServer logReplicationServerHandler;
 
@@ -221,7 +221,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      * Process discovery event
      */
     public synchronized void processEvent(DiscoveryServiceEvent event) {
-        switch (event.type) {
+        switch (event.getType()) {
             case ACQUIRE_LOCK:
                 processLockAcquire();
                 break;
@@ -243,7 +243,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
                 break;
 
             default:
-                log.error("Invalid event type {}", event.type);
+                log.error("Invalid event type {}", event.getType());
                 break;
         }
     }
@@ -752,21 +752,24 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     }
 
     @Override
-    public void forceSnapshotSync(String clusterId) throws LogReplicationDiscoveryServiceException {
+    public UUID forceSnapshotSync(String clusterId) throws LogReplicationDiscoveryServiceException {
         if (localClusterDescriptor.getRole() == ClusterRole.STANDBY) {
             String errorStr = "The forceSnapshotSync command is not supported on standby cluster.";
             log.error(errorStr);
             throw new LogReplicationDiscoveryServiceException(errorStr);
         }
 
-        log.info("Received the forceSnapshotSync command.");
+        log.info("Received the forceSnapshotSync command for standby cluster {}", clusterId);
 
-        /**
-         * write to the event to the event corfu table
-         */
+        // Write to the event to the event corfu table
+        UUID forceSyncId = UUID.randomUUID();
         ReplicationEventKey key = ReplicationEventKey.newBuilder().setKey(System.currentTimeMillis() + " "+ clusterId).build();
-        ReplicationEvent event = ReplicationEvent.newBuilder().setClusterId(clusterId).setType(LogReplicationMetadata.ReplicationEventType.FORCE_SNAPSHOT_SYNC).build();
+        ReplicationEvent event = ReplicationEvent.newBuilder()
+                .setClusterId(clusterId)
+                .setEventId(forceSyncId.toString())
+                .build();
         getLogReplicationMetadataManager().updateLogReplicationEventTable(key, event);
+        return forceSyncId;
     }
 
     @Override
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
@@ -205,7 +205,7 @@ public class CorfuReplicationManager {
     }
 
     /**
-     * Stop the current log replication event and start a full snapshot sync for all standby clusters
+     * Stop the current log replication event and start a full snapshot sync for the given remote cluster.
      */
     public void enforceSnapshotSync(DiscoveryServiceEvent event) {
         CorfuLogReplicationRuntime standbyRuntime = runtimeToRemoteCluster.get(event.getRemoteClusterInfo().getClusterId());
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
@@ -4,7 +4,6 @@ import com.google.common.collect.Sets;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
-
 import org.corfudb.infrastructure.LogReplicationRuntimeParameters;
 import org.corfudb.infrastructure.logreplication.replication.receive.LogReplicationMetadataManager;
 import org.corfudb.infrastructure.logreplication.runtime.CorfuLogReplicationRuntime;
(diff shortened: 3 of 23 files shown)


infrastructure/proto/log_replication_metadata.proto
```diff
@@ -44,17 +47,44 @@ message ReplicationStatusVal {
     SNAPSHOT = 0;
     LOG_ENTRY = 1;
   }
-  SyncType type = 3;
+  SyncType syncType = 3;
+  SyncStatus status = 4;
```
zhangn49, line 51: Planning to cancel the periodic task in `LogReplicationAckReader` explicitly or put a flag.
Otherwise, some special status like error or not started will get overwritten.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/LogReplicationFSM.java
```diff
@@ -253,6 +253,9 @@ public LogReplicationFSM(CorfuRuntime runtime, SnapshotReader snapshotReader, Da
         // Initialize Log Replication 5 FSM states - single instance per state
         initializeStates(snapshotSender, logEntrySender, dataSender);
 
+        // Assign logEntryReader to ackReader
+        this.ackReader.setLogEntryReader(logEntryReader);
```
annym, line 257: I think we might have an issue here, please look at the comment in LogReplicationAckReader with regards to this.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -743,7 +743,7 @@ public void testEnforceSnapshotSync() throws Exception {
         txKey = LogReplicationMetadataKey.newBuilder().setKey(LAST_SNAPSHOT_STARTED.getVal()).build();
         long newSnapshotTimestamp = Long.parseLong(metadataTable.get(txKey).getPayload().getVal());
         assertThat(lastAppliedSnapshot).isLessThan(newSnapshotTimestamp);
-        assertThat(newSnapshotTimestamp).isEqualTo(activeRuntime.getAddressSpaceView().getLogTail());
+//        assertThat(newSnapshotTimestamp).isEqualTo(activeRuntime.getAddressSpaceView().getLogTail());
```
annym, line 746: why is this no longer valid?

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -753,7 +753,7 @@ public void testEnforceSnapshotSync() throws Exception {
         log.info("NewSnapshot Timestamp {} active logTail {} after full snapshot sync!", newSnapshotTimestamp, activeRuntime.getAddressSpaceView().getLogTail());
 
         // The enforced snapshot sync completed.
-        assertThat(newSnapshotTimestamp).isEqualTo(Long.parseLong(metadataTable.get(txKey).getPayload().getVal()));
+//        assertThat(newSnapshotTimestamp).isEqualTo(Long.parseLong(metadataTable.get(txKey).getPayload().getVal()));
```
annym, line 756: same here?

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -699,7 +699,7 @@ public void testEnforceSnapshotSync() throws Exception {
                         "is {}, standby corfu[{}] log tail is {} lastAppliedSnapshot {}",
                 firstBatch, activeClusterCorfuPort, activeRuntime.getAddressSpaceView().getLogTail(),
                 standbyClusterCorfuPort, standbyRuntime.getAddressSpaceView().getLogTail(), lastAppliedSnapshot);
-        assertThat(lastAppliedSnapshot).isEqualTo(activeRuntime.getAddressSpaceView().getLogTail());
+//        assertThat(lastAppliedSnapshot).isEqualTo(activeRuntime.getAddressSpaceView().getLogTail());
```
annym, line 702: Nan, is it possible to include checks on any of our existing ITs, to validate that the replication status is being set correctly?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -378,7 +510,10 @@ public ReplicationStatusVal getReplicationRemainingEntries(String clusterId) {
 
     public void setDataConsistentOnStandby(boolean isConsistent) {
         ReplicationStatusKey key = ReplicationStatusKey.newBuilder().setClusterId(localClusterId).build();
-        ReplicationStatusVal val = ReplicationStatusVal.newBuilder().setDataConsistent(isConsistent).build();
+        ReplicationStatusVal val = ReplicationStatusVal.newBuilder()
+                .setStatus(LogReplicationMetadata.SyncStatus.STOPPED)
```
annym, line 514: this is only valid for a Standby, isn't it more intuitive if they see the status is ongoing? or do we aim to reflect the status of the replication from an active's point of view? In which case should we add a status that reflects that? cause it can be misleading, thinking that data is not consistent cause replication is stopped, when in reality the active is transferring data.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -335,17 +338,146 @@ public void setSnapshotAppliedComplete(LogReplicationEntry entry) {
         log.debug("Commit snapshot apply complete timestamp={}, for topologyConfigId={}", ts, topologyConfigId);
     }
 
-    public void setReplicationRemainingEntries(String clusterId, long remainingEntries,
-                                               ReplicationStatusVal.SyncType type) {
+    /**
+     * Update replication status table's snapshot sync info as completed.
```
annym, line 342: nit -> as ongoing?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -335,17 +338,146 @@ public void setSnapshotAppliedComplete(LogReplicationEntry entry) {
...
+        if (record != null) {
+            ReplicationStatusVal previous = record.getPayload();
+            syncInfo = previous.getSnapshotSyncInfo();
+        }
+
+        if (type == ReplicationStatusVal.SyncType.LOG_ENTRY) {
+            if (syncInfo == null){
+                log.warn("setReplicationStatusTable during LOG_ENTRY sync, " +
+                        "previous status is not present for cluster: {}", clusterId);
+                syncInfo = LogReplicationMetadata.SnapshotSyncInfo.newBuilder().build();
```
annym, line 439: In the case it was not set and we're coming from negotiation saying we can continue with LOGENTRY, on negotiation don't we get info of the last baseSnapshotTimestamp that at least we could populate? Maybe it's not worth the work but something to consider, what do you think? or we can leave a todo..

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -335,17 +338,146 @@ public void setSnapshotAppliedComplete(LogReplicationEntry entry) {
...
+                    clusterId, remainingEntries, type, syncInfo);
+        } else if (type == ReplicationStatusVal.SyncType.SNAPSHOT) {
+            LogReplicationMetadata.SnapshotSyncInfo currentSyncInfo;
+            if (syncInfo == null){
+                log.warn("setReplicationStatusTable during SNAPSHOT sync, " +
+                        "previous status is not present for cluster: {}", clusterId);
+                currentSyncInfo = LogReplicationMetadata.SnapshotSyncInfo.newBuilder().build();
+            } else {
+                currentSyncInfo = syncInfo.toBuilder()
+                        .setStatus(LogReplicationMetadata.SyncStatus.ONGOING)
```
annym, line 463: if we reach this and the snapshotSyncInfo is not present, shouldn't we set it to NOT_STARTED? as the path once it is started will set it to ONGOING... how do we handle potential TAE's from here and from the other async updateSnapshotSyncInfo? cause we could then overwrite a ONGOING, with a NOT_STARTED...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
zhangn49: Yes, ongoing makes more sense than stopped. If the user is able to check the status for each standby from the active side, I think empty status might be better.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
```diff
@@ -89,8 +92,7 @@ public void setSyncType(LogReplicationMetadata.ReplicationStatusVal.SyncType syn
         }
     }
 
-    public void startAckReader(LogEntryReader logEntryReader) {
-        this.logEntryReader = logEntryReader;
+    public void startAckReader() {
```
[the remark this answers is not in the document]
zhangn49: There was a NPE before I made this change. I'll revisit this part when I run some test later

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -335,17 +338,146 @@ public void setSnapshotAppliedComplete(LogReplicationEntry entry) {
...
+                    clusterId, syncInfo);
+    }
+
+    /**
+     * Update replication status table's snapshot sync info as completed.
+     *
+     * @param clusterId standby cluster id
+     */
+    public void updateSnapshotSyncInfo(String clusterId) {
+        Instant time = Instant.now();
```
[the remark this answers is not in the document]
zhangn49: Could you please explain a bit more about NTP? I don't find anything about NTP on google and I see lock store uses Instant too.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: It failed because the replication status is being updated a few more times.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
zhangn49: I don't think syncInfo is null if we come from negotiation. It should have a previous round snapshot sync, and syncInfo should be kept.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
zhangn49: If we reach here, current sync type is snapshot sync. So I have to keep both statuses as ongoing...
All replication status updates are protected by the lock in the ackReader, so TAE should be fine right now.
I added a flag to avoid overwriting by the periodic task.

zhangn49: BTW, I haven't found a way to differentiate NOT_STARTED and STOPPED if the current statue is initialized. All update is based on replication FSM, but negotiation is a runtime state.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: Yes, I will add some validations to a few ITs.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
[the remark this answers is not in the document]
annym: Yes, we were trying to change the lock store one but ran into some issue. Actually there's a PR opened for that change, in which we still need to understand why some tests started failing. 

This is what I recall of the use of Instant (which was an issue that Srinivas ran into). He explained that Instant.now() gives wall clock time, therefore a manual or NTP driven change in the wall clock will lead to the wrong conclusion. But I guess this is in the case of computations in between times, for example to compute a duration. If you're only using it as a point time (no computation), I think we should be fine!

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
annym: ok, I see. Then just delete this line!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
annym: yep, I agree empty status would be better.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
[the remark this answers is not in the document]
annym: ok, cool. Cause yes I think we still might have an issue until the LogEntryReader is set.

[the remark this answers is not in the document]
zhangn49: Brought it back. LogReplicationFSMTest did not start ackReader, so there was a NPE.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/WaitSnapshotApplyState.java
```diff
@@ -73,14 +73,16 @@ public WaitSnapshotApplyState(LogReplicationFSM logReplicationFSM, DataSender da
...
                 LogReplicationState snapshotSyncState = fsm.getStates().get(LogReplicationStateType.IN_SNAPSHOT_SYNC);
-                snapshotSyncState.setTransitionEventId(event.getEventID());
+                snapshotSyncState.setTransitionEventId(event.getEventId());
+                ((InSnapshotSyncState)snapshotSyncState).setForcedSnapshotSync(event.getMetadata().isForcedSnapshotSync());
                 return snapshotSyncState;
             case SYNC_CANCEL:
                 log.debug("Sync has been canceled while waiting for Snapshot Sync {} to complete apply. Restart.", transitionEventId);
                 LogReplicationState inSnapshotSyncState = fsm.getStates().get(LogReplicationStateType.IN_SNAPSHOT_SYNC);
                 inSnapshotSyncState.setTransitionEventId(UUID.randomUUID());
+                ((InSnapshotSyncState)inSnapshotSyncState).setForcedSnapshotSync(false);
```
pankti-m, line 85: is forceSnapshotSync guaranteed to be false?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
```diff
@@ -327,19 +335,61 @@ public void setBaseSnapshot(long baseSnapshotTimestamp) {
         this.baseSnapshotTimestamp = baseSnapshotTimestamp;
     }
 
+    public void markSnapshotSyncInfoCompleted() {
```
pankti-m, line 338: why do we need a special method to mark the sync as complete?  Cant it be inferred from the remainingReplicationEntries?

infrastructure/proto/log_replication_metadata.proto
```diff
@@ -44,17 +47,48 @@ message ReplicationStatusVal {
...
+message SnapshotSyncInfo {
+  enum SnapshotSyncType {
+    DEFAULT = 0;
+    FORCED = 1;
+  }
+  SnapshotSyncType type = 1;
+  SyncStatus status = 2;
+  string snapshotRequestId = 3;
+  google.protobuf.Timestamp completedTime = 4;
+  uint64 baseSnapshot = 5;
```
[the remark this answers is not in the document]
zhangn49: Yes, it is required to represent the base snapshot version of the latest snapshot sync.
Log Entry Sync -> Last snapshot sync's data
Snapshot Sync -> current one's data

infrastructure/proto/log_replication_metadata.proto
```diff
@@ -44,17 +47,48 @@ message ReplicationStatusVal {
...
+}
+
+enum SyncStatus {
+  ONGOING = 0;
+  COMPLETED = 1;
+  ERROR = 2;
```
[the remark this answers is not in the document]
zhangn49: The error means it is unrecoverable and Stopped is used when state is initialized(not in sync or wait for sync completion)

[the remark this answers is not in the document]
zhangn49: Will add comment

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationManager.java
```diff
@@ -216,7 +215,7 @@ public void enforceSnapshotSync(DiscoveryServiceEvent event) {
         } else {
             log.info("EnforceSnapshotSync for cluster {}", standbyRuntime.getRemoteClusterId());
             standbyRuntime.getSourceManager().stopLogReplication();
-            standbyRuntime.getSourceManager().startSnapshotSync();
+            standbyRuntime.getSourceManager().startForcedSnapshotSync(event.getEventId());
```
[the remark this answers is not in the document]
zhangn49: Done. Thanks.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/DiscoveryServiceEvent.java
```diff
@@ -1,35 +1,41 @@
...
 
 import lombok.Getter;
-import lombok.Setter;
-
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo;
 import org.corfudb.infrastructure.logreplication.proto.LogReplicationClusterInfo.TopologyConfigurationMsg;
 
+import java.util.UUID;
+
+@Getter
```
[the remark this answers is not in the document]
zhangn49: It is used get its fields, like getType() or getEventId()

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/DiscoveryServiceEvent.java
```diff
@@ -1,35 +1,41 @@
...
+    private UUID eventId = null;
 
     public DiscoveryServiceEvent(DiscoveryServiceEventType type) {
        this.type = type;
     }
 
     public DiscoveryServiceEvent(DiscoveryServiceEventType type, String clusterId) {
         this.type = type;
-        remoteClusterInfo = new ClusterDescriptor(clusterId, LogReplicationClusterInfo.ClusterRole.STANDBY);
+        this.remoteClusterInfo = new ClusterDescriptor(clusterId, LogReplicationClusterInfo.ClusterRole.STANDBY);
```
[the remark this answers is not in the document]
zhangn49: It is legacy code. Removed role.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/LogReplicationEventListener.java
```diff
@@ -39,21 +38,26 @@ public void onNext(CorfuStreamEntries results) {
...
-             * If the current node is the leader, it generates a discovery event and put it into the discovery service event queue.
-             */
+
+            // If the current node is the leader, it generates a discovery event and put it into the discovery service event queue.
             for (List<CorfuStreamEntry> entryList : results.getEntries().values()) {
                 for (CorfuStreamEntry entry : entryList) {
                     ReplicationEvent event = (ReplicationEvent) entry.getPayload();
-                    log.info("ReplicationEventListener put an event {} to its local discoveryServiceQueue", event);
-                    discoveryService.input(new DiscoveryServiceEvent(DiscoveryServiceEvent.DiscoveryServiceEventType.ENFORCE_SNAPSHOT_SYNC, event.getClusterId()));
+                    log.info("ReplicationEventListener received an event with id {}, type {}, cluster id{}",
```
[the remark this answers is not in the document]
zhangn49: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
```diff
@@ -19,30 +18,38 @@
...
+    private final LogReplicationMetadataManager metadataManager;
+    private final LogReplicationConfig config;
+    private final CorfuRuntime runtime;
+    private final String remoteClusterId;
+
     // Log tail when the current snapshot sync started.  We do not need to synchronize access to it because it will not
     // be read(calculateRemainingEntriesToSend) and written(setBaseSnapshot) concurrently.
     private long baseSnapshotTimestamp;
 
+    // Flag for the periodic task
```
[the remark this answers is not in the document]
zhangn49: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InSnapshotSyncState.java
```diff
@@ -106,12 +113,14 @@ public LogReplicationState processEvent(LogReplicationEvent event) throws Illega
                     UUID newSnapshotSyncId = UUID.randomUUID();
                     log.debug("Starting new snapshot sync after cancellation id={}", newSnapshotSyncId);
                     inSnapshotSyncState.setTransitionEventId(newSnapshotSyncId);
+                    ((InSnapshotSyncState)inSnapshotSyncState).setForcedSnapshotSync(false);
                     snapshotSender.reset();
+                    fsm.getAckReader().markSnapshotSyncInfoOngoing(forcedSnapshotSync, transitionEventId);
```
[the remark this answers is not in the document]
zhangn49: There are a few more fields that ackReader is not aware of, like if the snapshot is forced or not.
And the periodic task is triggered every 15 seconds, we might not able to observe some status.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/WaitSnapshotApplyState.java
zhangn49: In the case of SYNC_CANCEL, it will automatically start a new round of snapshot sync, which means it is not forced.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
zhangn49: remainingReplicationEntries is not 100% accurate and will have 0-15s latency to update status. So I am using the FSM transitions to update status.

PavelZaytsev (approved):

annym (approved):

pankti-m (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -752,21 +753,26 @@ public void updateTopology(LogReplicationClusterInfo.TopologyConfigurationMsg to
...
-         * write to the event to the event corfu table
-         */
+        // Write to the event to the event corfu table
         ReplicationEventKey key = ReplicationEventKey.newBuilder().setKey(System.currentTimeMillis() + " "+ clusterId).build();
-        ReplicationEvent event = ReplicationEvent.newBuilder().setClusterId(clusterId).setType(LogReplicationMetadata.ReplicationEventType.FORCE_SNAPSHOT_SYNC).build();
+        ReplicationEvent event = ReplicationEvent.newBuilder()
+                .setClusterId(clusterId)
+                .setEventId(forceSyncId.toString())
+                .setType(ReplicationEvent.ReplicationEventType.FORCE_SNAPSHOT_SYNC)
+                .build();
```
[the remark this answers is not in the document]
zhangn49: Yes, will address the metadata manager's TAE handling issue in the next PR.

Issue #2575, opened 2020-06-12
Optimization: try avoid new sequencer being elected on tail node.


WenbinZhu:
## Overview

Tail log unit node needs to support read, which has heavier load, so we can try to avoid selecting new sequncer on the tail nodes if we have other options. Of course, in case of multi strips, there could be more than one tail node.

WenbinZhu:
Addressed in #2775

PR #2775, opened 2020-09-24
When sequncer fails over, elect non-tail node as primary if possible.

WenbinZhu:
## Overview

Description:

Log tail nodes need to serve queries, so when sequencer fails over, we
try to elect a non-tail node as the primary sequencer if possible to
have better load distribution.

Related issue(s) (if applicable): #2575 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
@@ -10,8 +10,11 @@ import org.corfudb.runtime.view.Layout.LayoutStripe;
 
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * A builder that allows us to make modifications to a layout and construct
@@ -328,36 +331,49 @@ public class LayoutBuilder {
 
     /**
      * Moves a responsive server to the top of the sequencer server list.
+     * Tries to avoid electing a log tail node if possible to have better load distribution.
      * If all have failed, throws exception.
      *
-     * @param endpoints a non null set of Strings representing failed endpoints
-     *
+     * @param excludedNodes a non null set of Strings representing excluded endpoints
      * @return this builder
-     *
-     * @throws LayoutModificationException is thrown if none of the sequencers in layout
-     *         can be moved to the top of sequencer server list due to being unresponsive
-     *         or a failed endpoint.
+     * @throws LayoutModificationException if none of the sequencers in the layout can be
+     *                                     moved to the top of sequencer server list due
+     *                                     to being unresponsive or a failed endpoint.
      */
-    public LayoutBuilder assignResponsiveSequencerAsPrimary(@NonNull Set<String> endpoints) {
+    public LayoutBuilder assignResponsiveSequencerAsPrimary(@NonNull Set<String> excludedNodes) {
+        LinkedList<String> modifiedSequencers = new LinkedList<>(layout.getSequencers());
+        Optional<String> primarySequencer = Optional.empty();
 
-        List<String> modifiedSequencerServers = new ArrayList<>(layout.getSequencers());
-        for (int i = 0; i < modifiedSequencerServers.size(); i++) {
-            String sequencerServer = modifiedSequencerServers.get(i);
+        // Consider only tail nodes in the last segment.
+        Set<String> logTails = layout.getLastSegment().getStripes().stream()
+                .map(LayoutStripe::getTailEndpoint)
+                .collect(Collectors.toSet());
 
-            // Add a sequencer server given sequencerServer is not already included in
-            // endpoints AND it is not included in the unresponsive nodes of the layout.
-            if (!endpoints.contains(sequencerServer) &&
-                !layout.getUnresponsiveServers().contains(sequencerServer)) {
-                modifiedSequencerServers.remove(sequencerServer);
-                modifiedSequencerServers.add(0, sequencerServer);
-                layout.setSequencers(modifiedSequencerServers);
-                return this;
+        // Elect a sequencer server as primary given that server is not already included
+        // in excludedNodes AND it is not included in the unresponsive nodes of the layout.
+        // Non log tail servers have high priority to be elected for better load distribution.
+        for (String sequencer : modifiedSequencers) {
+            if (excludedNodes.contains(sequencer) ||
+                    layout.getUnresponsiveServers().contains(sequencer)) {
+                continue;
+            }
+            primarySequencer = Optional.of(sequencer);
+            if (!logTails.contains(sequencer)) {
+                break;
             }
         }
 
+        if (primarySequencer.isPresent()) {
+            modifiedSequencers.remove(primarySequencer.get());
+            modifiedSequencers.addFirst(primarySequencer.get());
+            layout.setSequencers(modifiedSequencers);
+            return this;
+        }
+
         log.warn("Failover to a responsive sequencer server failed. Each of the layout " +
-                 "sequencers:{} is either a failed endpoint or an unresponsive server",
-                 modifiedSequencerServers.toString());
+                        "sequencers:{} is either a failed endpoint or an unresponsive server",
+                modifiedSequencers.toString());
+
         throw new LayoutModificationException("All sequencers failed.");
     }
 
--- a/test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
+++ b/test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
@@ -1,5 +1,6 @@
 package org.corfudb.infrastructure;
 
+import com.google.common.collect.Sets;
 import org.assertj.core.api.Assertions;
 import org.corfudb.AbstractCorfuTest;
 import org.corfudb.runtime.exceptions.LayoutModificationException;
@@ -22,9 +23,6 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
  */
 public class LayoutBuilderTest extends AbstractCorfuTest {
 
-    private static final int SERVER_ENTRY_3 = 3;
-    private static final int SERVER_ENTRY_4 = 4;
-
     /**
      * Tests the Layout Builder by removing nodes.
      *
@@ -33,6 +31,9 @@ public class LayoutBuilderTest extends AbstractCorfuTest {
      */
     @Test
     public void checkRemovalOfNodes() throws LayoutModificationException {
+        final int SERVER_ENTRY_3 = 3;
+        final int SERVER_ENTRY_4 = 4;
+
         Layout originalLayout = new TestLayoutBuilder()
                 .setEpoch(1L)
                 .addLayoutServer(SERVERS.PORT_0)
@@ -71,11 +72,9 @@ public class LayoutBuilderTest extends AbstractCorfuTest {
         //Preparing failed nodes set.
         failedNodes.addAll(allNodes);
 
-
         /*
          * Invalid Removals
          */
-
         // Deleting all Layout Servers
         assertThatThrownBy(() -> layoutBuilder.removeLayoutServers(failedNodes).build()
         ).isInstanceOf(LayoutModificationException.class);
@@ -99,7 +98,6 @@ public class LayoutBuilderTest extends AbstractCorfuTest {
         /*
          *  Valid Removal
          */
-
         // Deleting SERVERS.PORT_0 and SERVERS.PORT_4
         // Preparing new layout
         Layout expectedLayout = new TestLayoutBuilder()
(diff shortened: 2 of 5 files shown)


WenbinZhu:
Will add new test cases later.

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -330,35 +332,48 @@ public LayoutBuilder mergePreviousSegment(int segmentIndex) {
...
+                    layout.getUnresponsiveServers().contains(sequencer)) {
+                continue;
+            }
+            if (primarySequencer == null) {
+                primarySequencer = sequencer;
+            }
+            if (!logTails.contains(sequencer)) {
+                primarySequencer = sequencer;
+                break;
             }
```
Lujie1996, line 362: A way to save if-clauses might be:
`primarySequencer = sequencer;
if (!logTails.contains(sequencer)) {
break;
}`
The only difference is that if there is no non-tail servers, your code picks the first sequencer in the loop iteration whereas mine picks the last one. Does this matter?

WenbinZhu: Yea I actually thought about this, but I was afraid there are tests relying on selecting the first node, so I kept it same as before. Looks like it's not the case, just did the change, thanks.

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
```diff
@@ -1,5 +1,6 @@
 package org.corfudb.infrastructure;
 
+import com.google.common.collect.Sets;
```
zhangn49, line 3: Unused

zhangn49 (approved):

annym (approved):

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -328,37 +330,50 @@ public LayoutBuilder mergePreviousSegment(int segmentIndex) {
...
-            // endpoints AND it is not included in the unresponsive nodes of the layout.
-            if (!endpoints.contains(sequencerServer) &&
-                !layout.getUnresponsiveServers().contains(sequencerServer)) {
-                modifiedSequencerServers.remove(sequencerServer);
-                modifiedSequencerServers.add(0, sequencerServer);
-                layout.setSequencers(modifiedSequencerServers);
-                return this;
+    public LayoutBuilder assignResponsiveSequencerAsPrimary(@NonNull Set<String> excludedNodes) {
+        List<String> modifiedSequencers = new LinkedList<>(layout.getSequencers());
+        String primarySequencer = null;
```
xnull, line 344: can you please use Optional instead of null

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
```diff
@@ -328,37 +330,50 @@ public LayoutBuilder mergePreviousSegment(int segmentIndex) {
...
-        throw new LayoutModificationException("All sequencers failed.");
+        if (primarySequencer == null) {
+            log.warn("Failover to a responsive sequencer server failed. Each of the layout " +
+                            "sequencers:{} is either a failed endpoint or an unresponsive server",
+                    modifiedSequencers.toString());
+            throw new LayoutModificationException("All sequencers failed.");
+        }
+
+        modifiedSequencers.remove(primarySequencer);
+        modifiedSequencers.add(0, primarySequencer);
```
xnull, line 373: it can be improved:

```
LinkedList<String> modifiedSequencers = new LinkedList<>(layout.getSequencers());
...
modifiedSequencers.addFirst(primarySequencer);
```

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/view/LayoutBuilder.java
WenbinZhu: Done.

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
```diff
@@ -418,6 +402,61 @@ public void unsuccessfulSequencerFailoverThrowsException() throws Exception {
...
+                .addToSegment()
+                .addToLayout()
+                .build();
+
+        // All servers responsive, server 1 is the only non-tail node.
+        LayoutBuilder layoutBuilder = new LayoutBuilder(originalLayout);
+        Layout newLayout = layoutBuilder
+                .assignResponsiveSequencerAsPrimary(Collections.emptySet())
+                .build();
+        assertThat(newLayout.getPrimarySequencer()).isEqualTo(SERVERS.ENDPOINT_1);
```
xnull, line 444: I have run your test with the previous version and the current one, it seems that the previous version works even better.
It's not clear to me what improvement has been done if your test works the same way for both versions.

Regarding the check above, as I understand it must be ENDPOINT_0, so why you expect to see ENDPOINT_1 here?

xnull (changes requested):

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
WenbinZhu: This is just a unit test not perf test. The purpose of the test is to make sure a non-tail node will be elected for primary sequencer if there is any. It also tests if all nodes are tails or all nodes are unresponsive, the behavior is same as before.

Your understanding is not correct, endpoint 0 is a tail node, so it won't be elected as primary sequencer in the algorithm. Endpoint 1 is a non-tail node, so it will be elected. The test might also work for old implementation because old implementation does not have any preference on which node being elected as long as they are responsive. So endpoint 0,1,2 are all possible. All my codes are documented, please read again.

xnull: Ok, let's start from the beginning, questions, please answer:
 - The layout that you created in the test is not possible in production. We always have only one stripe in one segment. What was the motivation for that?
 
 - making Sequencer election depending on stripes is not the best idea, in my opinion, it makes logic pretty convoluted and unclear, why we need to overcomplicate that logic? 
    The Sequencer is a much more important component of the cluster than stripe and must not be dependent on a stripe. Maybe we can change the FIRST stripe server in the last segment depending on the primary sequencer but not vise versa.

WenbinZhu: > The layout that you created in the test is not possible in production. We always have only one stripe in one segment. What was the motivation for creating multiple stripes in the test?

Striping is a design concept in corfu. We don't have multi-stipe in production does not mean we don't need to test it. You can see a lots of tests have multi-stripe layout. On the other hand, mutli-stripe test is a "super set" of single sripe, it's always better to make sure it works for multi-stripe, which implies it works for single stripe.

> making Sequencer election depending on stripes is not the best idea, in my opinion, it makes logic pretty convoluted and unclear, why we need to overcomplicate that logic?
The Sequencer is a much more important component of the cluster than stripe and must not be dependent on a stripe. Maybe we can change the FIRST stripe server in the last segment depending on the primary sequencer but not vise versa.

This is not depending on stripes, it just give high priority to non tail nodes, you can think of it as an optimization which takes advantage of the log tail information we have in the layout builder, it's not a dependency. Also I don't think the logic is convoluted, it's pretty clear: give high priority to non-tail nodes. You have a bunch of nodes to choose from, just give high priority to those who are not tails.
When you need to do re-election on sequencer, why do you want to change log unit?  You cannot arbitrarily change log units order, there needs to be more layout changes and state transfer.

vjeko:
Can you please provide data that supports you claim? Where was this issue seen? How many times did you observe it?

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
xnull: >it's not a dependency
It is a dependency, the design is - the more important component (sequencer) depends on a less important one (which is an optimization that not even fully used in production). 
I believe that kind of dependency is not acceptable.

WenbinZhu: Please provide reason that makes it not acceptable. If it's because of "convoluted" than I already said it's actually pretty  simple, how could ten lines of change being convoluted. It's just a simple optimization, which will be downgraded to previous approach if all responsive nodes are tails.

> (which is an optimization that not even fully used in production).
This is used in production, which has a single node as tail, the optimization still applies to this case. The unit test verifies on a more general case.

WenbinZhu:
> Can you please provide data that supports you claim? Where was this issue seen? How many times did you observe it?

This can happen any time, the current election has no prefence on backup sequencer nodes. I recently saw a production layout with this pattern, which reminds me that we can do this optimization.

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
xnull: I believe that kind of dependency is not acceptable. 
The most important component of the cluster depends on the least important

pankti-m (approved):

test/src/test/java/org/corfudb/infrastructure/LayoutBuilderTest.java
WenbinZhu: First this is not denpendency. We  are currently deploying everything on a node, but you can even choose to deploy sequencer on a node that does not have log unit server. this is the initial design. This patch does a similar thing.
By the way sequencer already depend on log unit server for fail-over (get persisted address map and stream tails.) Without this dependency, system won't work. Also log unit is not the least important, by design sequencer is just an optimization for log unit. Log unit is the source of truth.

xnull: >We are currently deploying everything on a node, but you can even choose to deploy sequencer on a node that does not have log unit server. this is the initial design.

It has nothing to do with the deployment. We are talking only about layout. To figure out what server will be the next sequencer we need to have information about stripes.

>By the way sequencer already depend on log unit server for fail-over (get persisted address map and stream tails.)

could you please point where it is

>by design sequencer is just an optimization for log unit

Never heard about that, can you please provide a link to the initial design where it claimed?

>Log unit is the source of truth.

It's not true if we are talking about a particular node, that's why we have AutoCommitService and read the latest global address from the entire cluster, right?

WenbinZhu: The deployment is an example. When deploying which node to be sequencer, system admin can decide to deploy a sequencer on non log unit node, which uses stripe information. By the way, you have stripe info in your layout why not use it to do optimization. Sequencer itself is an optmization to separate token issue responsibility from log unit. Log unit is the source of truth for tokens, it can issue tokens instead, we assign this responsibility to sequencer to get better performance. Ideally sequencer should be deployed on a sparate node to achieve most performance. When sequencer fails over it needs to query log unit for the persisted tokens, that means sequencer is an optimization for log unit, that dependency is there from day 1. You can take a look at LayoutManagementView::reconfigureSequencerServers.

xnull: >By the way, you have stripe info in your layout why not use it to do optimization.

because I don't see it as optimization and how it can improve things.

>Sequencer itself is an optmization.

this is just your personal opinion

>When sequencer fails over it needs to query log unit for the persisted tokens
>that means sequencer is an optimization for log unit

Not clear to me how you get to the conclusion that sequencer is an optimization. 
And not clear how we can get from the above statements to the conclusion that the sequencer should be chosen from stripes.

>that dependency is there from day 1

Sequencer needs to know the latest global address, but the election process has nothing to do with the address

>You can take a look at LayoutManagementView::reconfigureSequencerServers.

could you please show the exact point in the code?

WenbinZhu: > this is just your personal opinion

If you ever attended a new team member on board session, you will also hear maithem telling them sequencer is an optimization. Log unit is the persistent storage, that is the ground truth and most important thing.

> And not clear how we can get from the above statements to the conclusion that the sequencer should be chosen from stripes.

This is not to get that conclusion, it's to correct your statement that log unit is the least important.
Also it's not choosing from stripes, it's avoiding from choose a tail node, even there is no log unit server, no stripe, the algorithm still works. That's not a dependecy.

> Sequencer needs to know the latest global address, but the election process has nothing to do with the address

Failover includes election and recorver sequencer state from log unit, which already uses stripe info.

> could you please show the exact point in the code?

https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java#L425-L426

xnull: >it's to correct your statement that log unit is the least important.

One more time, stripe != log unit

>https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java#L425-L426

the change was made 2 months ago
image
looks like also a false assumption

P.S. anyways, let's not spend so much time on an endless discussion. 
It's up to up to address or not my concerns

WenbinZhu: > One more time, stripe != log unit

Every read/write request uses stripe info, I don't know how you come with the statement that stripe is not important. Maybe you think we have only 1 stripe deployment, but to make code work for general case, you need to consider stripping in your implentation.

> looks like also a false assumption

One more time, please read code carefully before commenting. getLogAddressSpace() calls getChainHeadFromAllSegments(), which uses stripe information.

> P.S. anyways, let's not spend so much time on an endless discussion.

I agree, this is kind of wasting time, if you need help in understanding how some certain components work in system, I can help offline.

PR #2777, opened 2020-09-28
Client proto router

xcchang:
## Overview

Put this in draft to make it easier to view my current changes and progress

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
```diff
@@ -0,0 +1,360 @@
...
+        // Write this message out on the channel
+        ByteBuf outBuf = PooledByteBufAllocator.DEFAULT.buffer();
+        ByteBufOutputStream requestOutputStream = new ByteBufOutputStream(outBuf);
+
+        try {
+            // Mark this message as Protobuf message (temporarily)
+            requestOutputStream.writeByte(API.PROTO_CORFU_MSG_MARK);
+            request.writeTo(requestOutputStream);
+            ctx.writeAndFlush(outBuf, ctx.voidPromise());
+            log.trace("Sent one-way request message: {}", request.getHeader());
```
zfrenette, line 175: This is not a one-way message.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
```diff
@@ -0,0 +1,360 @@
...
+        final Timer.Context roundTripMsgContext = MetricsUtils
+                .getConditionalContext(roundTripMsgTimer);
+
+        // Get the next request ID
+        final long thisRequest = requestID.getAndIncrement();
+
+        // Set the base fields for this message.
+        Request.Builder reqBuilder = request.toBuilder();
+        reqBuilder.getHeaderBuilder().setClientId(parameters.getClientId()).setRequestId(thisRequest);
+        request = reqBuilder.build();
```
zfrenette, line 160: At this point, the Request is already built by the client calling this method. Calling toBuilder() involves making a copy of the whole object (I believe including payload), which seems wasteful just to update some fields in the header. I'm wondering if this can be avoided... In a perfect world, we would only build the request once. Perhaps if the Request can't be built entirely by the client, it could be better to pass a builder() and delay calling build()?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
```diff
@@ -0,0 +1,360 @@
...
+     * @param ctx Context of the channel handler.
+     */
+    @Override
+    public void sendRequest(Request request, ChannelHandlerContext ctx) {
+        // Get the next request ID
+        final long thisRequest = requestID.getAndIncrement();
+        // Set the base fields for this message.
+        Request.Builder reqBuilder = request.toBuilder();
+        reqBuilder.getHeaderBuilder().setClientId(parameters.getClientId()).setRequestId(thisRequest);
+        request = reqBuilder.build();
```
zfrenette, line 223: See my above comment. Can we avoid building the Request twice?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
```diff
@@ -0,0 +1,360 @@
...
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        ByteBuf msgBuf = (ByteBuf) msg;
+
+        // Temporary -- If message is a legacy message, forward the message.
+        byte msgMark = msgBuf.getByte(msgBuf.readerIndex());
+        if (msgMark == API.LEGACY_CORFU_MSG_MARK) {
+            // TODO should we also use this method on the client side?
+            ctx.fireChannelRead(msgBuf);
+            return;
```
zfrenette, line 275: This is not needed here. It was needed server-side since both types of routers are in my pipeline. On the client-side, you won't have an older router sitting in the pipeline, so there is no need to forward the message. If the new client receives a legacy message as a Response, we can throw an exception.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
```diff
@@ -0,0 +1,360 @@
...
+        final CompletableFuture<T> cfBenchmarked = cf.thenApply(x -> {
+            MetricsUtils.stopConditionalContext(roundTripMsgContext);
+            return x;
+        });
+
+        // Generate a timeout future, which will complete exceptionally
+        // if the main future is not completed.
+        final CompletableFuture<T> cfTimeout =
+                CFUtils.within(cfBenchmarked, Duration.ofMillis(timeoutResponse));
+        Request finalRequest = request;
```
zfrenette, line 193: Why is this needed?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
xcchang: Yeah that's also what I was thinking about, pass a builder as the parameter could avoid building it twice. I will make the change in next commit.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
xcchang: Got you, thanks!

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
xcchang: This is actually caught by the IDE that 

> Variable used in lambda expression should be final or effectively final

and this line of code is literally inside lambda body. I did some investigation in this problem and found some explanation [here](https://docs.oracle.com/javase/specs/jls/se10/html/jls-15.html#jls-15.27.2).

> The restriction to effectively final variables prohibits access to dynamically-changing local variables, whose capture would likely introduce concurrency problems.

So I change request to an effectively final local variable, not sure is there any better way to address this problem.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
xcchang: I will correct it in next commit. Thanks!

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientProtobufRouter.java
```diff
@@ -0,0 +1,705 @@
...
+    @Override
+    public void sendRequest(Request.Builder reqBuilder) {
+        // Get the next request ID
+        final long thisRequest = requestID.getAndIncrement();
+        // Set the base fields for this message.
+        Header header = reqBuilder.getHeaderBuilder()
+                .setClientId(API.getProtoUUID(parameters.getClientId()))
+                .setRequestId(thisRequest)
+                .build();
+        Request request = reqBuilder.setHeader(header).build();
```
xcchang, line 470: Here I pass the Request.Builder to avoid duplicate creation of the whole request message, but it still recreates the Header by invoking `getHeaderBuilder()`. To totally clear this kind of duplicate recreation, the only method I could think of now is pass two builders `sendRequest(Request.Builder reqBuilder, Header.Builder hdBuilder)`, which will make it too redundant and require lots of changes to upstream interfaces. 
Looking forward to hearing some suggestions here :)

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -0,0 +1,235 @@
...
+                    }
+                    if (handlerMap.containsKey(handler.type())) {
+                        throw new RuntimeException("Handler for " + handler.type()
+                                + " already registered!");
+                    }
+                    // convert the method into a Java8 Lambda for maximum execution speed...
+                    try {
+                        if (Modifier.isStatic(method.getModifiers())) {
+                            MethodHandle mh = caller.unreflect(method);
+                            MethodType mt = mh.type().changeParameterType(0, ResponseMsg.class);
```
zfrenette, line 136: You don't need to change the parameter type here. This what needed in the old implementation to change the instances of `CorfuPayloadMsg` (which extends `CorfuMsg`) to `CorfuMsg`. Here, everything will just be a `ResponseMsg` with nothing extending it.

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -0,0 +1,235 @@
...
+                                            MethodType.methodType(Handler.class),
+                                            mt, mh, mh.type())
+                                    .getTarget().invokeExact());
+                        } else {
+                            // instance method, so we need to capture the type.
+                            MethodType mt = MethodType
+                                    .methodType(method.getReturnType(), method.getParameterTypes());
+                            MethodHandle mh = caller.findVirtual(o.getClass(), method.getName(), mt);
+                            MethodType mtGeneric = mh.type()
+                                    .changeParameterType(1, ResponseMsg.class)
```
zfrenette, line 148: You don't need to change the parameter type here. Everything will just be a `ResponseMsg` with nothing extending it. See my other comment above.

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -0,0 +1,235 @@
...
+                    }
+                    if (errorHandlerMap.containsKey(handler.type())) {
+                        throw new RuntimeException("Handler for " + handler.type()
+                                + " already registered!");
+                    }
+                    // convert the method into a Java8 Lambda for maximum execution speed...
+                    try {
+                        if (Modifier.isStatic(method.getModifiers())) {
+                            MethodHandle mh = caller.unreflect(method);
+                            MethodType mt = mh.type().changeParameterType(0, ResponseMsg.class);
```
zfrenette, line 187: You don't need to change the parameter type here. See above.

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -0,0 +1,235 @@
...
+                            handlerMap.put(handler.type(), (Handler) LambdaMetafactory
+                                    .metafactory(caller, "handle",
+                                            MethodType.methodType(Handler.class),
+                                            mt, mh, mh.type())
```
zfrenette, line 140: Since `mt` is unneeded here, you can just reuse `mh.type()`.

runtime/src/main/java/org/corfudb/runtime/clients/BaseHandler.java
```diff
@@ -159,4 +190,184 @@ private static Object handleWrongClusterId(CorfuPayloadMsg<WrongClusterMsg> msg,
...
+     * @return none, throw a wrong epoch exception instead.
+     */
+    @ServerErrorsHandler(type = ErrorCase.UNKNOWN_ERROR)
+    private static Object handleUnknownError(ResponseMsg msg, ChannelHandlerContext ctx,
+                                             IClientRouter r) throws Throwable {
+        ByteString bs = msg.getPayload().getServerError().getUnknownError().getThrowable();
+        byte[] bytes = new byte[bs.size()];
+        bs.copyTo(bytes, 0);
+        ByteArrayInputStream bis = new ByteArrayInputStream(bytes);
+        ObjectInputStream ois = new ObjectInputStream(bis);
```
zfrenette, line 289: Maybe we can wrap this in a try-with-resources block? We might also need to consider the case where we receive an exception that we do not recognize (can't deserialize)... particularly within the context of rolling upgrades.

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -0,0 +1,235 @@
...
+            }
+        } else if (handlerMap.containsKey(payloadCase)) {
+            try {
+                Object ret = handlerMap.get(payloadCase).handle(response, ctx, router);
+                if (ret != null) {
+                    router.completeRequest(requestId, ret);
+                }
+            } catch (Throwable e) {
+                // TODO: same reason, should we remove this clause?
+                router.completeExceptionally(requestId, e);
```
zfrenette, line 109: I am not sure if this can be removed, since we have to complete the future exceptionally. Does this occur elsewhere if this is removed?

test/src/test/java/org/corfudb/runtime/clients/TestClientRouter.java
```diff
@@ -242,8 +268,56 @@ public IClientRouter addClient(IClient client) {
...
+        final CompletableFuture<T> cf = new CompletableFuture<>();
+        outstandingRequests.put(thisRequestId, cf);
+
+        // TODO: TestRule and TestServerRouter
+        // Evaluate rules.
+        if (rules.stream().allMatch(x -> x.evaluate(request, this))) {
+            // Write the message out to the channel
+            log.trace(Thread.currentThread().getId() + ":Sent request: {}", request);
+            // No need to simulate serialization for Protobuf
+            serverRouter.sendServerMessage(request, channelContext);
```
chetangudisagar, line 304: Why not simulate serialization? Here is my reasoning, feel free to add on.

1. For the time-being, we add request and response marker byte into the buffer, its better to simulate/test that.
2. The serialization is simulated on the server side. Either we can remove there, or add it here just to keep the consistency.
3. Test suites should generally mock the actual behaviours of the main code to cover all the corner cases. The main idea anyway is to catch any possible error during the testing phase itself.

test/src/test/java/org/corfudb/runtime/clients/TestClientRouter.java
```diff
@@ -290,7 +374,23 @@ public void sendMessage(CorfuMsg message) {
...
+        CorfuMessage.RequestMsg request = CorfuMessage.RequestMsg.newBuilder().setHeader(header)
+                .setPayload(payload).build();
+
+        // TODO: TestRule and TestServerRouter
+        // Evaluate rules.
+        if (rules.stream().allMatch(x -> x.evaluate(request, this))) {
+            // Write the message out to the channel
+            log.trace(Thread.currentThread().getId() + ":Sent request: {}", request);
+            // No need to simulate serialization for Protobuf
+            serverRouter.sendServerMessage(request, channelContext);
```
chetangudisagar, line 392: same as above (reg serialization).

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /**
+     * Add a handler to this ClientResponseHandler.
+     *
+     * @param payloadCase Response payload type.
+     * @param handler The handler itself.
+     * @return This handler, to support chaining.
+     */
+    @Deprecated
```
zfrenette, line 67: Why is this deprecated when it's newly added?

test/src/test/java/org/corfudb/runtime/clients/TestRule.java
```diff
@@ -53,6 +58,28 @@ public TestRule matches(Function<CorfuMsg, Boolean> matcher) {
         return this;
     }
 
+    /**
+     * Provide a custom matcher.
+     *
+     * @param requestMatcher A function that takes a CorfuMsg and returns true if the
+     *                       message matches.
```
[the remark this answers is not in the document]
xcchang: Will do

test/src/test/java/org/corfudb/runtime/clients/TestRule.java
```diff
@@ -53,6 +58,28 @@ public TestRule matches(Function<CorfuMsg, Boolean> matcher) {
...
+    public TestRule requestMatches(Function<RequestMsg, Boolean> requestMatcher) {
+        this.requestMatcher = requestMatcher;
+        return this;
+    }
+
+    /**
+     * Provide a custom matcher.
+     *
+     * @param responseMatcher A function that takes a CorfuMsg and returns true if the
+     *                        message matches.
```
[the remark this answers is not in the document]
xcchang: Will do

test/src/test/java/org/corfudb/runtime/clients/TestClientRouter.java
```diff
@@ -265,6 +440,22 @@ private boolean validateClientId(CorfuMsg msg) {
         return true;
     }
 
+    /**
+     * Validate the clientID of a CorfuMsg.
+     *
+     * @param protoClientId The header of incoming message used for validation.
```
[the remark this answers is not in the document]
xcchang: Will do

test/src/test/java/org/corfudb/runtime/clients/TestRule.java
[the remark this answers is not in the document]
zfrenette: Fixed in new branch

test/src/test/java/org/corfudb/runtime/clients/TestRule.java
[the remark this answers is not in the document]
zfrenette: Fixed in new branch

xcchang:
Already merged in another PR

PR #2778, opened 2020-09-29
Lr metrics

PavelZaytsev:
## Overview

Description:
- Introduce micrometer library - facade, over the instrumentation clients that allows to instrument the application without having to lock in on the backend.
- Configure the `LoggingMeterRegistry` - allows metrics to be exportable via a separate logger with a defined frequency (1 min by default).
- Introduce a custom logging sink - `InfluxLineProtocolLoggingSink` - allows exporting metrics in the InfuxDb Line Protocol format to ease the subsequent export of the metrics.
 - Instrument LR. In order to enable metrics a `logback.xml` file needs to be appropriately configured. See `logback-test-lr-metrics.xml` for an example, also a flag `--metrics` needs to be passed on a startup. 

### Further changes:
- Configure `logback.xml` in production repo to enable a separate async logger for the metrics.
- Pass a flag `--metrics` on startup in the production repo.
- Set up and configure infra for the metrics ingestion, analysis and visualization in the infra repo.

- [x] Public API has Javadoc


common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Get the previously configured meter registry.
+     * If the registry has not been previously configured return an
+     * empty option.
+     * @return An optional configured meter registry.
+     */
+    public static Optional<MeterRegistry> getInstance() {
+        return meterRegistry;
```
xnull, line 29: you would have a race condition here in multi-threaded env

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Get the previously configured meter registry.
+     * If the registry has not been previously configured, create a default logging MeterRegistry
+     * and return.
+     * @return An optional configured meter registry.
+     */
+    public static Optional<MeterRegistry> getInstance() {
+        if (!meterRegistry.isPresent()) {
```
xnull, line 31: here is a possible race. Imagine you have two threads
t1->getInstance()
t2->getInstance()

t1-> meterRegistry is not present
at the  same time
t12-> meterRegistry is not present ALSO

t1->creates LoggingMeterRegistry
t2->creates LoggingMeterRegistry

Please take a look https://www.journaldev.com/1377/java-singleton-design-pattern-best-practices-examples

I would prefer "Eager initialization" approach, if possible. 
If you would like to use your method you need to add an additional check, see "4. Thread Safe Singleton"

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Configure the default registry of type LoggingMeterRegistry.
+     */
+    public static void createLoggingMeterRegistry() {
+        Supplier<Optional<MeterRegistry>> supplier = () -> Optional.of(new LoggingMeterRegistry());
+        create(supplier);
+    }
+
+    private static synchronized void create(Supplier<Optional<MeterRegistry>> meterRegistrySupplier) {
+        if (!meterRegistry.isPresent()) {
```
xnull, line 87: It seems to me meterRegistry must be present no matter what.
Is it better not to use Optional but provide guarantees that MeterRegistry always exists?

common/src/main/java/org/corfudb/common/metrics/micrometer/loggingsink/DefaultLoggingSink.java
```diff
@@ -0,0 +1,17 @@
...
+
+import lombok.AllArgsConstructor;
+import org.slf4j.Logger;
+
+/**
+ * A default logging sink. It prints the line in the default format.
+ */
+@AllArgsConstructor
+public class DefaultLoggingSink implements LoggingSink {
+    private final Logger logger;
```
xnull, line 11: please add `@NonNull` lombok annotation

common/src/main/java/org/corfudb/common/metrics/micrometer/loggingsink/InfluxLineProtocolLoggingSink.java
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.Logger;
+
+/**
+ * This sink prints the lines to the log in the InfluxDb line protocol.
+ * https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_tutorial/
+ */
+@AllArgsConstructor
+public class InfluxLineProtocolLoggingSink implements LoggingSink {
+
+    private final Logger logger;
```
xnull, line 20: please add @NonNull lombok annotation

common/src/main/java/org/corfudb/common/metrics/micrometer/protocoltransformer/LineTransformer.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Optional;
+
+/**
+ * Matches and transforms the line with one of the provided matcher transformers.
+ */
+@AllArgsConstructor
+@Slf4j
+public class LineTransformer {
+    private final ImmutableList<MatcherTransformer> matcherTransformers;
```
xnull, line 15: please add `@NonNull` lombok annotation

common/src/main/java/org/corfudb/common/metrics/micrometer/protocoltransformer/influx/InfluxLineProtocolTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+     * Measurement, tagSet and fieldSet.
+     */
+    int NUM_GROUPS = 3;
+
+    /**
+     * A definition of the InfluxDb data point.
+     */
+    @AllArgsConstructor
+    class InfluxLineProtocolPoint {
+        private final String measurement;
```
xnull, line 22: please add `@NonNull` lombok annotation for all String fields

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InLogEntrySyncState.java
```diff
@@ -141,7 +154,13 @@ public void onEntry(LogReplicationState from) {
                 logEntrySender.reset(fsm.getBaseSnapshot(), fsm.getAckedTimestamp());
             }
 
-            logEntrySyncFuture = fsm.getLogReplicationFSMWorkers().submit(() -> logEntrySender.send(transitionEventId));
+            Runnable logEntrySendTask = () -> logEntrySender.send(transitionEventId);
+
+            if (senderTimer.isPresent()) {
```
xnull, line 159: `ifPresent` will work better

xnull (approved):
Please fix Singleton, everything else looks good

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InSnapshotSyncState.java
```diff
@@ -149,9 +171,14 @@ public void onEntry(LogReplicationState from) {
...
                 snapshotSender.reset();
                 fsm.getAckReader().markSnapshotSyncInfoOngoing(forcedSnapshotSync, transitionEventId);
-            }
+                snapshotSyncTimerSample = MeterRegistryProvider.getInstance().map(Timer::start);
 
-            transmitFuture = fsm.getLogReplicationFSMWorkers().submit(() -> snapshotSender.transmit(transitionEventId));
+            }
+            Runnable snapShotSendTask = () -> snapshotSender.transmit(transitionEventId);
+            if (senderTimer.isPresent()) {
+                snapShotSendTask = senderTimer.get().wrap(snapShotSendTask);
```
annym, line 179: I believe we would have the same issue described InLogEntrySyncState here.. the transmit will not reflect accurately the time for sending a snapshot message as it might not send anything at all...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -197,12 +204,17 @@ public LogReplicationEntry read(UUID logEntryRequestId) throws TrimmedException
...
 
                 lastOpaqueEntry = txOpaqueStream.next();
+                deltaCounter.ifPresent(Counter::increment);
+
                 lastOpaqueEntryValid = isValidTransactionEntry(lastOpaqueEntry);
                 currentProcessedEntryMetadata = new StreamIteratorMetadata(txOpaqueStream.txStream.pos(), lastOpaqueEntryValid);
             }
 
             log.trace("Generate LogEntryDataMessage size {} with {} entries for maxDataSizePerMsg {}. lastEntry size {}",
                     currentMsgSize, opaqueEntryList.size(), maxDataSizePerMsg, lastOpaqueEntry == null ? 0 : currentEntrySize);
```
annym, line 214: By looking at this, maybe another good metric to expose is in average how many opaque entries are sent per message... opaqueEntryList.size()

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -197,12 +204,17 @@ public LogReplicationEntry read(UUID logEntryRequestId) throws TrimmedException
                 }
 
                 lastOpaqueEntry = txOpaqueStream.next();
+                deltaCounter.ifPresent(Counter::increment);
```
annym, line 207: I believe this count should be after L.209 only if lastOpaqueEntryValid is true... cause if not, we will be counting transaction entries which are not intended to be replicated. For instance, consider 5 entries in the transaction stream, but only 1 is valid (has streams to replicate).. we would be counting 5 deltas when we should really count 1... or maybe we can count both (in different metrics)... to know which proportion of the transaction stream ends up being required for replication (1 out of 5 in this case).. I think that could be a nice way of analyzing the data available... to see how noisy the tx stream can be with regards to the data to be replicated..

annym (commented):
Could we add some markdown file in the micrometer folder (or as a comment in the main metrics class file, whatever approach you prefer) where we summarize the stats that we are collecting for LR (so far), for instance:

Current metrics collected for LR:

- Number of ACKs received by active cluster, distinguished between snapshot and log entry, tag: logreplication.messages
- …

I think this could be useful for someone trying to understand which stats we have and which might be missing. It is good to keep this around for future reference so this work is not missed or ignored.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -806,4 +814,18 @@ public void shutdown() {
     private String getLocalHost() {
         return NodeLocator.parseString(serverContext.getLocalEndpoint()).getHost();
     }
+
+
+    private Optional<LongTaskTimer.Sample> recordLockAcquire(ClusterRole role) {
+        return MeterRegistryProvider.getInstance()
+                .map(registry -> registry.more()
```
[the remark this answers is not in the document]
PavelZaytsev: Added

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InLogEntrySyncState.java
```diff
@@ -171,4 +191,18 @@ public UUID getTransitionEventId() {
...
     }
+
+    private Optional<AtomicLong> configureAcksCounter() {
+        return MeterRegistryProvider.getInstance()
+                .map(registry -> registry.gauge("logreplication.messages",
+                        ImmutableList.of(Tag.of("replication.type", "logentry")),
+                        new AtomicLong(0)));
+    }
+
+    private Optional<Timer> configureSenderTimer() {
```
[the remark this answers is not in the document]
PavelZaytsev: Moved it to sendWithBuffering for both replication types. The timer now starts when the future starts and stops when the future completes exceptionally or with a result.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InSnapshotSyncState.java
```diff
@@ -103,6 +117,14 @@ public LogReplicationState processEvent(LogReplicationEvent event) throws Illega
                 waitSnapshotApplyState.setBaseSnapshotTimestamp(snapshotSender.getBaseSnapshotTimestamp());
                 fsm.setBaseSnapshot(event.getMetadata().getLastTransferredBaseSnapshot());
                 fsm.setAckedTimestamp(event.getMetadata().getLastLogEntrySyncedTimestamp());
+                snapshotSyncAcksCounter.ifPresent(AtomicLong::getAndIncrement);
```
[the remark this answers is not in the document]
PavelZaytsev: Changed the metric name to reflect the number of types snapshot sync has occurred and also created a metric that counts acks in processAcks.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
PavelZaytsev: Added the one that counts opaqueEntryList size.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
PavelZaytsev: Created the one that counts overall and the one that counts only valid ones.

annym (approved):
Thanks for adding the Readme, that's great.

common/src/main/java/org/corfudb/common/metrics/micrometer/README.md
```diff
@@ -0,0 +1,14 @@
+### Current metrics collected for LR:
+
+  * Message size in bytes (throughput, mean and max), distinguished by replication type (snapshot, logentry): **logreplication.message.size.bytes**.
+  * Duration of holding a leadership lock in seconds, distinguished by role (active, standby): **logreplication.lock.duration**.
+  * Number of times a leadership lock was acquired, distinguised by role (active, standby): **logreplication.lock.acquire.count**.
+  * Duration of sending a logentry in seconds (throughput, mean and max), distinguished by replication type (snapshot, logentry) and status (success, failure): **logreplication.sender.duration.seconds**.
+  * Duration of sending a message overall (throughput, mean and max): **logreplication.rtt.seconds**.
+  * Number of snapshot syncs completed. **logreplication.snapshot.completed.count**.
+  * Duration of completing a snapshot sync in seconds (throughput, mean and max): **logreplication.snapshot.duration**.
+  * Number of acks, distinguished by replication type (snapshot, logentry): **logreplication.acks**.
```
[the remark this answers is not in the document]
PavelZaytsev: you just told me to add two spaces damn it

PR #2783, opened 2020-10-07
Internal Backup/Restore implementation

Lujie1996:
## Overview

Description:
This PR includes the basic implementation of internal B/R and an integration test of it.

Why should this be merged: 
This is fundamental to B/R.

- [x] Public API has Javadoc


zhangn49:
@Lujie1996 You can rebase it on top of the master branch and squash. It will help you generate a jar file.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,146 @@
...
+            if (!backupTable(fileName, streamId)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * If th log is trimmed at timestamp, the backupTable will fail.
```
hisundar, line 81: If the* log is trimmed

Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,180 @@
...
+            while (fileInput.available()> 0) {
+                OpaqueEntry opaqueEntry = OpaqueEntry.read(fileInput);
+                List<SMREntry> smrEntries = opaqueEntry.getEntries().get(streamId);
+                if (smrEntries == null || smrEntries.isEmpty()) {
+                    continue;
+                }
+
+                txn.logUpdate(streamId, smrEntries);
+                numEntries++;
+                log.debug("write uuid {} with {} numEntries", streamId, numEntries);
```
zhangn49, line 120: Should this line be outside of the while loop?

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * Restore a single table
+     * @param fileName   - the name of the temp backup file
+     * @param streamId   - the stream ID of the table which is to be restored
+     * @return true if restore succeeded
+     * @throws IOException when 1) backup file is not found or 2) failed to read file
+     */
+    private boolean restoreTable(String fileName, UUID streamId) throws IOException {
```
zhangn49, line 94: IOException is not thrown by this function.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
```diff
@@ -263,6 +263,8 @@ public void syncWithRetryUnsafe(VersionLockedObject vlo,
 
     public abstract void logUpdate(UUID streamId, SMREntry updateEntry);
 
+    public abstract void logUpdate(UUID streamId, List<SMREntry> updateEntries);
```
zhangn49, line 266: Javadoc comment for public method

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,219 @@
...
+     */
+    private void generateTarFile() throws IOException {
+        File folder = new File(backupTempDirPath);
+        File[] srcFiles = folder.listFiles();
+
+        FileOutputStream fileOutput;
+        TarArchiveOutputStream tarOutput;
+        try {
+            fileOutput = new FileOutputStream(filePath);
+            tarOutput = new TarArchiveOutputStream(fileOutput);
```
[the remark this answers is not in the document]
Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/object/transactions/AbstractTransactionalContext.java
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        if (!restore()) {
+            cleanup();
+            return false;
+        }
+
+        cleanup();
+        return true;
+    }
```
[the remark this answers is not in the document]
Lujie1996: Done

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,236 @@
...
+        File[] srcFiles = folder.listFiles();
+        if (srcFiles == null) {
+            log.debug("no backup file found under directory {}", backupTempDirPath);
+            return;
+        }
+
+        try (FileOutputStream fileOutput = new FileOutputStream(filePath);
+             TarArchiveOutputStream tarOutput = new TarArchiveOutputStream(fileOutput)) {
+            for (File srcFile : srcFiles) {
+                addToTarFile(srcFile, tarOutput);
```
annym, line 167: do we want to preserve the file or delete original after in the TAR? or is this done in the cleanup?

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    private static final String RESTORE_TEMP_DIR_PREFIX = "corfu_restore_";
+
+    /***
+     * @param filePath      - the path of backup tar file
+     * @param runtime       - the runtime which is performing the restore
+     * @throws IOException when failed to create the temp directory
+     */
+    public Restore(String filePath, CorfuRuntime runtime) throws IOException {
+        this.filePath = filePath;
```
annym, line 52: we're restoring whatever is available in that TAR... imagine someone manipulates that TAR and removes some files... should we have a validation similar to the case of Backup, where if a list of streams to restore is given, or tagged streams is specified as preferred option, at the end we consider successful, if and only if we have restored everything requested in the API.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,165 @@
...
+            openTarFile();
+            verify();
+            restore();
+        } catch (IOException e) {
+            log.error("failed to restore from backup tar file {}", filePath);
+            throw e;
+        } finally {
+            cleanup();
+        }
+        log.debug("restore completed");
```
annym, line 74: info?

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    /**
+     * Some verification logic (TBD) such as
+     * - Compare the user provided streamIds and names of table backups under tmp directory, or some metadata file
+     * - Checksum
+     */
+    private void verify() {
+        // TODO: verification logic to be implemented
```
annym, line 156: I think this actually has to do with my comment.. should we solve this now?

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,165 @@
...
+                List<SMREntry> smrEntries = opaqueEntry.getEntries().get(streamId);
+                if (smrEntries == null || smrEntries.isEmpty()) {
+                    continue;
+                }
+
+                txn.logUpdate(streamId, smrEntries);
+                numEntries++;
+            }
+            txn.commit();
+            log.debug("write uuid {} with {} numEntries", streamId, numEntries);
```
annym, line 115: can we add an explicit log.info stating: 'Completed restore of table {}' ... (something in that line)

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
```diff
@@ -0,0 +1,338 @@
...
+
+        // Create Corfu Store to add entries into server
+        CorfuStore srcDataCorfuStore = new CorfuStore(srcDataRuntime);
+        CorfuStore destDataCorfuStore = new CorfuStore(destDataRuntime);
+
+        List<String> tableNames = getTableNames(numTables);
+
+        // Generate random entries and save into sourceServer
+        for (String tableName : tableNames) {
+            generateData(srcDataCorfuStore, tableName, true);
```
annym, line 314: we only have a test case where all tables are tagged as required for backup, can we have a test that combines non-tagged with tagged? so we validate that it is selective.

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
```diff
@@ -0,0 +1,338 @@
...
+
+        // Close servers and runtime before exiting
+        cleanEnv();
+    }
+
+    /**
+     * Backup and restore tables which have requires_backup_support tag set
+     */
+    @Test
+    public void backupRestoreTaggedTables() throws IOException, NoSuchMethodException, IllegalAccessException, InvocationTargetException {
```
annym, line 302: I think we can add a bit of more tests to cover more scenarios:

- A test where the TAR is deleted before restore, to validate exception is thrown.. 
- A test where only a subset of specified streamIds are backed (not all, like in test 1)
- A test where the log is trimmed before backup...

annym: Some other tests I thought would be relevant are:

- Specify a StreamX for backup (through the API where streams for backup are given) which has never been open by any client (not in the registry). 
- Specify a StreamY for backup which is open but empty, so we validate that empty tables are restored with no issues.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Slf4j
+public class Backup {
```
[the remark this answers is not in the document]
Lujie1996: Added class description to both Backup and Restore. Thanks.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,236 @@
...
+        this.filePath = filePath;
+        this.backupTempDirPath = Files.createTempDirectory(BACKUP_TEMP_DIR_PREFIX).toString();
+        this.streamIDs = streamIDs;
+        this.runtime = runtime;
+        this.timestamp = runtime.getAddressSpaceView().getLogTail();
+    }
+
+    /**
+     * The difference from the constructor above is that this one will automatically discover
+     * tables which requires back up from their tags.
```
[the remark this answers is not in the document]
Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,236 @@
...
+    /**
+     * @param filePath      - the filePath where the generated backup tar file will be placed
+     * @param streamIDs     - the stream IDs of tables which are backed up
+     * @param runtime       - the runtime which is performing the back up
+     * @throws IOException  - when failed to create the temp directory
+     */
+    public Backup(String filePath, List<UUID> streamIDs, CorfuRuntime runtime) throws IOException {
+        this.filePath = filePath;
+        this.backupTempDirPath = Files.createTempDirectory(BACKUP_TEMP_DIR_PREFIX).toString();
+        this.streamIDs = streamIDs;
```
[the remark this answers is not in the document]
Lujie1996: Added a check in `backup()`:

```
if (streamIDs.isEmpty()) {
     log.warn("back up is called with empty streamIDs!");
} 
```

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,236 @@
...
+        this.runtime = runtime;
+        this.timestamp = runtime.getAddressSpaceView().getLogTail();
```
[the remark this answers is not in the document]
Lujie1996: Good point. I moved the `this.timestamp = runtime.getAddressSpaceView().getLogTail();` to method `start()`.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,236 @@
...
+                    .build();
+            Stream<OpaqueEntry> stream = (new OpaqueStream(runtime, runtime.getStreamsView().get(uuid, options))).streamUpTo(timestamp);
+
+            writeTableToFile(fileOutput, stream, uuid);
+        } catch (IOException e) {
+            log.error("failed to backup table {} to file {}", uuid, fileName);
+            throw e;
+        }
+
+        log.debug("{} table is backed up and stored to temp file {}", uuid, fileName);
```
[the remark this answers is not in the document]
Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: Those files are under `backupTempDirPath` and are deleted in `cleanup()` at the end of the whole process.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: The requirement for restore is to restore whatever inside the given TAR file. We used to let users input a list of tables they want to restore in the restore API but Graham's team said they don't really need that feature. That's why we remove that parameter and just restore whatever they give us.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: I haven't decided what to check as verification. Also since it's a nice-to-have feature and not currently asked for by Graham's team, I think it might be better to not put too many efforts in this very first backup restore PR.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: Changed to `log.info("completed restore of table {} with {} numEntries", streamId, numEntries);`. Thanks.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param uuid       - the uuid of the table which is being backed up
+     * @throws IOException
+     */
+    private void backupTable(String fileName, UUID uuid) throws IOException {
+        try (FileOutputStream fileOutput = new FileOutputStream(fileName)) {
+            StreamOptions options = StreamOptions.builder()
+                    .ignoreTrimmed(false)
+                    .cacheEntries(false)
+                    .build();
+            Stream<OpaqueEntry> stream = (new OpaqueStream(runtime, runtime.getStreamsView().get(uuid, options))).streamUpTo(timestamp);
```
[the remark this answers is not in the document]
Lujie1996: Good point. I thinking failing the whole backup would be easier for users to implement some retry mechanism. We don't want users to say backup 6 in 10 tables (because the other 4 failed) and backup the remaining 4 tables in a second run, and then merge the two tar files. I understand your point is that if some tables failed at least we don't lose other successful backups. I think we need some comments from Graham to see how they want to handle this exception.

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
Lujie1996: They are all very good test cases. I'll add them soon.

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
Lujie1996: Done. Thanks.

annym (approved):
LGTM, thanks for adding these tests!

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @param fileName   - the name of the temp backup file
+     * @param streamId   - the stream ID of the table which is to be restored
+     * @throws IOException
+     */
+    private void restoreTable(String fileName, UUID streamId) throws IOException {
+        try (FileInputStream fileInput = new FileInputStream(fileName)) {
+            // Clear table before restore
+            TxnContext txn = corfuStore.txn(CORFU_SYSTEM_NAMESPACE);
+            SMREntry entry = new SMREntry("clear", new Array[0], Serializers.PRIMITIVE);
```
pankti-m, line 109: should we also clear the stream and get rid of any ongoing updates?

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
```diff
@@ -0,0 +1,640 @@
...
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.lang.RandomStringUtils;
+import org.corfudb.integration.AbstractIT;
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.collections.*;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
+import org.corfudb.runtime.exceptions.TrimmedException;
+import org.corfudb.runtime.view.TableRegistry;
+import org.corfudb.test.SampleSchema;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
pankti-m, line 15: nit - group all static imports together

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
```diff
@@ -0,0 +1,640 @@
+package org.corfudb.runtime;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.lang.RandomStringUtils;
+import org.corfudb.integration.AbstractIT;
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.runtime.collections.*;
```
pankti-m, line 10: Do we allow wildcard imports in tests?  If not, please do remove them

runtime/pom.xml
```diff
@@ -124,6 +124,17 @@
             <artifactId>commons-lang3</artifactId>
             <version>3.9</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.18</version>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/commons-io/commons-io -->
```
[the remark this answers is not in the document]
Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+    private long timestamp;
+
+    // The Corfu Runtime which is performing the back up
+    private final CorfuRuntime runtime;
+
+    // All tables in Corfu Db
+    private List<UUID> allTablesInDb;
+
+    /**
+     * Backup files of tables are temporarily stored under BACKUP_TEMP_DIR. They are deleted after backup finishes.
```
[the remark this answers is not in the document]
Lujie1996: The backup temporary directory is generated by this line of code:
`this.backupTempDirPath = Files.createTempDirectory(BACKUP_TEMP_DIR_PREFIX).toString();` (https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createTempDirectory-java.lang.String-java.nio.file.attribute.FileAttribute...-)
The directory is created in system default temp path.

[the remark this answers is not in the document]
Lujie1996: as explain above

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * The difference from the constructor above is that this one will automatically discover
+     * tables which require back up from their tags.
+     *
+     * @param filePath      - the filePath where the generated backup tar file will be placed
+     * @param runtime       - the runtime which is performing the back up
+     * @throws IOException  - when failed to create the temp directory
+     */
+    public Backup(String filePath, CorfuRuntime runtime) throws IOException {
+        this.filePath = filePath;
```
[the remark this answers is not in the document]
Lujie1996: Backup() is constructor. Calling `this(filePath, getTaggedTables(), runtime)` will trigger `Error:(83, 24) java: cannot reference this before supertype constructor has been called`

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+        return allTablesInDb.contains(streamId);
+    }
+
+    /**
+     * All temp backupTable files will be placed under BACKUP_DIR_PATH directory.
+     *
+     * @throws IOException
+     */
+    private void backup() throws IOException {
+        if (streamIDs.isEmpty()) {
```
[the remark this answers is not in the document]
Lujie1996: Just added null check in start() and abort backup in case of null streamIDs. Thanks!

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+            backupTable(fileName, streamId);
+        }
+        log.info("successfully backed up {} tables to {} directory", streamIDs.size(), backupTempDirPath);
+    }
+
+    /**
+     * Back up a single table
+     *
+     * If the log is trimmed at timestamp, the backupTable will fail.
+     * If the table has no data to backupTable, it will create a file with empty contents.
```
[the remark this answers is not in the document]
Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+            throw e;
+        }
+    }
+
+    /**
+     * Find tables which have the requires_backup_support tag set
+     *
+     * @return  List<UUID>   - a list of UUIDs for all the tables which require backup support
+     */
+    public List<UUID> getTaggedTables() {
```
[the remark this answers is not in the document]
Lujie1996: Done. Thanks.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+
+        return tables;
+    }
+
+    /**
+     * Cleanup the table backup files under the backupDir directory.
+     */
+    private void cleanup() {
+        try {
+            FileUtils.deleteDirectory(new File(backupTempDirPath));
```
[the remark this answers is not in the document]
Lujie1996: It is somehow possible. But considering the temporary directory is actually created in the middle of Backup process with a prefix in dir name`Files.createTempDirectory(BACKUP_TEMP_DIR_PREFIX)`, I think it's really unlike that someone would give this exact filePath if not being deliberate.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    /**
+     * Open the backup tar file and save the table backups to tableDir directory
+     */
+    private void openTarFile() throws IOException {
+        try (FileInputStream fileInput = new FileInputStream(filePath);
+             TarArchiveInputStream tarInput = new TarArchiveInputStream(fileInput)) {
+            getTablesFromTarFile(tarInput);
```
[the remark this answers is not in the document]
Lujie1996: Interesting. I tried to make it align but line 136 and 137 are actually auto-indented by pressing Enter from the line above them. Line 136 is aligned with `FileInputStream` and line 137 is indented as an inner scope code block.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,174 @@
...
+            long numEntries = 0;
+            while (fileInput.available()> 0) {
+                OpaqueEntry opaqueEntry = OpaqueEntry.read(fileInput);
+                List<SMREntry> smrEntries = opaqueEntry.getEntries().get(streamId);
+                if (smrEntries == null || smrEntries.isEmpty()) {
+                    continue;
+                }
+
+                txn.logUpdate(streamId, smrEntries);
+                numEntries++;
```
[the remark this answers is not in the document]
Lujie1996: Good point. Changed to `numEntries += smrEntries.size();`

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,174 @@
...
+            TxnContext txn = corfuStore.txn(CORFU_SYSTEM_NAMESPACE);
+            SMREntry entry = new SMREntry("clear", new Array[0], Serializers.PRIMITIVE);
+            txn.logUpdate(streamId, entry);
+
+            long numEntries = 0;
+            while (fileInput.available()> 0) {
+                OpaqueEntry opaqueEntry = OpaqueEntry.read(fileInput);
+                List<SMREntry> smrEntries = opaqueEntry.getEntries().get(streamId);
+                if (smrEntries == null || smrEntries.isEmpty()) {
+                    continue;
```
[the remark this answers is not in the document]
Lujie1996: continue and break should have the same effect here. continue will go back to `while(fileInput.available() > 0)` and then exits the loop.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    private void verify() {
+        // TODO: verification logic to be implemented
+    }
+
+    /**
+     * Cleanup the table backup files under the tableDir directory.
+     */
+    private void cleanup() throws IOException {
+        FileUtils.deleteDirectory(new File(restoreTempDirPath));
```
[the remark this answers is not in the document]
Lujie1996: I think this is expected behavior. The tar file is provided by user and I think is better to be users' responsibility to collect or remove. Also deleting such tar file might be unexpected  for some users if they didn't read documents carefully. It's just safer to keep it instead of deleting it.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: When this particular SMREntry gets committed it will clear the stream to empty. Also, the team using this Restore guarantees that all services are shut down during restore so there will be no more ongoing updates after the clear request.

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
Lujie1996: Done. Thanks.

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
Lujie1996: Not sure if it is allowed. Just removed.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+            String fileName = backupTempDirPath + File.separator + streamId;
+            backupTable(fileName, streamId);
+        }
+        log.info("successfully backed up {} tables to {} directory", streamIDs.size(), backupTempDirPath);
+    }
+
+    /**
+     * Back up a single table
+     *
+     * If the log is trimmed at timestamp, the backupTable will fail.
```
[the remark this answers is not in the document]
Lujie1996: As discussed, this happens extremely rarely. In case of such scenario, users can do a retry.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -0,0 +1,273 @@
...
+    private void backupTable(String fileName, UUID uuid) throws IOException {
+        try (FileOutputStream fileOutput = new FileOutputStream(fileName)) {
+            StreamOptions options = StreamOptions.builder()
+                    .ignoreTrimmed(false)
+                    .cacheEntries(false)
+                    .build();
+            Stream<OpaqueEntry> stream = (new OpaqueStream(runtime, runtime.getStreamsView().get(uuid, options))).streamUpTo(timestamp);
+
+            writeTableToFile(fileOutput, stream, uuid);
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
Lujie1996: Done. Thanks.

pankti-m (approved):

PR #2793, opened 2020-10-16
CorfuStore option Tx Logging Disabled

annym:
## Overview

LR does require TX Logging to be disabled on the standby site to avoid
generation of DCNs on replicated data and CorfuStore enforces it to be enabled.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
@@ -58,17 +58,21 @@ public class LogReplicationMetadataManager {
     private final String localClusterId;
 
     public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, String localClusterId) {
+
+        // LR does not require transaction logging enabled as we don't want to trigger subscriber's logic
+        // on replicated data which could eventually lead to overwrites
+        this.runtime = rt;
+        this.corfuStore = new CorfuStore(runtime, false);
+
+        // This special runtime with transaction logging enabled is required for the REPLICATION_EVENT_TABLE
+        // which is a table used to communicate events between lead and non-lead nodes in LR (e.g., notify
+        // force snapshot sync from a non-lead to the lead node)
         CorfuRuntime.CorfuRuntimeParameters params = rt.getParameters();
         this.runtimeTxLogging = CorfuRuntime.fromParameters(params)
                 .setTransactionLogging(true)
                 .parseConfigurationString(rt.getLayoutServers().get(0))
                 .connect();
         this.corfuStoreTxLogging = new CorfuStore(runtimeTxLogging);
-
-        this.runtime = rt;
-        // LR does not require transaction logging as we don't want data change notifications on the replicated data
-        // this runtime is reused by the LogEntryWriter and SnapshotWriter
-        this.corfuStore = new CorfuStore(runtime, false);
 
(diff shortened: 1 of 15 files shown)


zhangn49 (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -59,7 +59,7 @@
 
     public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, String localClusterId) {
         this.runtime = rt;
-        this.corfuStore = new CorfuStore(runtime);
+        this.corfuStore = new CorfuStore(runtime, false);
```
pankti-m, line 62: It will be good to add a comment here that LR simply replicates the streams and does not need it to be added to the transaction log

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -59,7 +59,9 @@
 
     public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, String localClusterId) {
         this.runtime = rt;
-        this.corfuStore = new CorfuStore(runtime);
+        // LR does not require transaction logging as we don't want data change notifications on the replicated data
```
pankti-m, line 62: is the relation between data change notification and tx stream general to corfu?

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
```diff
@@ -179,6 +196,88 @@ public void testSnapshotSyncApplyInterrupted() throws Exception {
...
+
+    private Future<Boolean>  subscribeTransactionStream() {
+
+        ExecutorService consumer = Executors.newSingleThreadExecutor();
+        List<CorfuRuntime> consumerRts = new ArrayList<>();
+
+        // A thread that starts and consumes transaction updates via the Transaction Stream.
+        return consumer.submit(() -> {
+
+            CorfuRuntime consumerRt = CorfuRuntime.fromParameters(CorfuRuntime.CorfuRuntimeParameters
```
pankti-m, line 233: slightly unrelated to this change, but do we need to create multiple runtimes?

pankti-m (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
annym: not really, it's one use case (out of Corfu) but it can be used for any purpose. Just didn't know how to make it clear.. I've rephrased it, take a look and let me know if you think it still is confusing.

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
annym: no, changing to only one and shutting it down...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
pankti-m: maybe we can say - 'Replicated entries from LR need not be logged to the transaction stream'... or something like that?

PR #2794, opened 2020-10-20
ProtobufSerializer: Reuse existing classMap

hisundar:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
@@ -31,6 +31,7 @@ import org.corfudb.runtime.collections.StreamingMap;
 import org.corfudb.runtime.collections.StreamingMapDecorator;
 import org.corfudb.runtime.collections.Table;
 import org.corfudb.runtime.collections.TableOptions;
+import org.corfudb.runtime.exceptions.SerializerException;
 import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.object.ICorfuVersionPolicy;
 import org.corfudb.runtime.object.transactions.TransactionType;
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -96,11 +97,24 @@
...
+            ProtobufSerializer registeredSerializer = (ProtobufSerializer)Serializers.getSerializer(ProtobufSerializer.PROTOBUF_SERIALIZER_CODE);
+            classMap1 = registeredSerializer.getClassMap();
+            protoSerializer = registeredSerializer;
+        } catch (SerializerException se) {
+            // This means the protobuf serializer had not been registered yet
+            classMap1 = new ConcurrentHashMap<>();
+            protoSerializer = new ProtobufSerializer(classMap1);
+            Serializers.registerSerializer(protoSerializer);
+        }
+        this.classMap = classMap1;
```
annym, line 115: I got rid of this classMap class variable as it is not used.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -283,9 +297,8 @@ public static String getFullyQualifiedTableName(String namespace, String tableNa
     private <T extends Message> void addTypeToClassMap(T msg) {
         String typeUrl = getTypeUrl(msg.getDescriptorForType());
         // Register the schemas to schema table.
-        if (!classMap.containsKey(typeUrl)) {
-            classMap.put(typeUrl, msg.getClass());
-        }
+        ((ProtobufSerializer)Serializers.getSerializer(ProtobufSerializer.PROTOBUF_SERIALIZER_CODE))
+                .getClassMap().put(typeUrl, msg.getClass());
```
WenbinZhu, line 301: Do you still need this change? Since the classMap is already pointing to the old classMap in the constructor, so you can directly operate on the classMap, right?

WenbinZhu, line 301: Do you still need this change? Since the classMap is already pointing to the old classMap in the constructor, so you can directly operate on the classMap, right?

WenbinZhu (commented):
@hisundar Should we start to branch out, and push this workaround patch to HL and do the better fix (per runtime SerialierMap) for master?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: sure.

hisundar:
change included in https://github.com/CorfuDB/CorfuDB/pull/2793

PR #2807, opened 2020-10-30
Client side corfu metrics

PavelZaytsev:
## Overview

Description:

Introduce client side metrics.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -242,6 +246,12 @@ public static CorfuRuntimeParametersBuilder builder() {
          */
         private MetricRegistry metricRegistry;
 
+        private boolean microMeterRegistryEnabled;
+
+        private Duration microMeterLoggingInterval;
+
+        private String configuredLogger;
```
xnull, line 253: can we extract these 3 variables into a MetricsConfig class?

xnull (approved):
looks good

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
PavelZaytsev: done

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
```diff
@@ -25,6 +26,26 @@ private MeterRegistryProvider() {
...
+
+        /**
+         * Create a new instance of MeterRegistry with the given logger, loggingInterval
+         * and clientId.
+         * @param logger A configured logger.
+         * @param loggingInterval A duration between log appends for every metric.
+         * @param clientId An id of a client for this metric.
+         * @return A new meter registry.
+         */
+        public static MeterRegistry newInstance(Logger logger, Duration loggingInterval,
```
Maithem, line 38: Can you remove usages to `getInstance` and just instantiate a new instance for LR (just like you do it for the client/server). Having a singleton implemented with a `synchronized` on the getter is not efficient.

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
```diff
@@ -41,6 +62,22 @@ public static synchronized void init(Logger logger, Duration loggingInterval,
...
 
+        /**
+         * Configure the meter registry of type LoggingMeterRegistry. All the metrics registered
+         * with this meter registry will be exported in the InfluxDB line protocol format
+         * (https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_tutorial/)
+         * with  the provided loggingInterval frequency.
+         * @param logger A configured logger.
+         * @param loggingInterval A duration between log appends for every metric.
+         */
+        public static synchronized void init(Logger logger, Duration loggingInterval) {
```
Maithem, line 73: This `init` doesn't seem like its being used.

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
```diff
@@ -63,6 +100,25 @@ public static synchronized void init(Duration loggingInterval,
...
         }
 
+        /**
+         * Configure the meter registry of type LoggingMeterRegistry. All the metrics registered
+         * with this meter registry will be exported via provided logging sink with
+         * the provided loggingInterval frequency.
+         * @param sink A configured logging sink.
+         * @param loggingInterval A duration between log appends for every metric.
+         */
+        public static synchronized void init(Duration loggingInterval,
```
Maithem, line 110: This `init` doesn't seem like its being used.

common/src/main/java/org/corfudb/common/metrics/micrometer/loggingsink/InfluxLineProtocolLoggingSink.java
```diff
@@ -20,6 +21,7 @@
 
     @NonNull
     private final Logger logger;
+    @Getter
```
Maithem, line 24: Why does this sink log to debug?

Maithem, line 24: Also, in `LineTransformer` instead of returning return an `Optional.empty()`  we should fail. Right now we don't have any metrics that we don't expect to transform. Returning an empty optional can hide incorrect regexes which makes it harder to debug (i.e., https://github.com/CorfuDB/CorfuDB/pull/2825#discussion_r523245312 )

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -825,6 +849,17 @@ private CorfuRuntime(@Nonnull CorfuRuntimeParameters parameters) {
             MetricsUtils.metricsReportingSetup(
                     defaultMetrics, parameters.getPrometheusMetricsPort());
         }
+        CorfuRuntimeParameters.MicroMeterRuntimeConfig microMeterRuntimeConfig =
+                this.parameters.getMicroMeterRuntimeConfig();
+        if (microMeterRuntimeConfig.metricsEnabled) {
+            LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
+            registry = Optional.ofNullable(loggerContext.exists(microMeterRuntimeConfig.configuredLoggerName))
+                    .map(logger -> MeterRegistryProvider.MeterRegistryInitializer.newInstance(logger,
```
Maithem, line 857: If `metricsEnabled = true` but no logger was found then we should print a warning message.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -1118,6 +1154,9 @@ private void pruneRemovedRouters(@Nonnull Layout layout) {
                         // Prune away removed node routers from the nodeRouterPool.
                         pruneRemovedRouters(l);
 
+                        getRegistry().ifPresent(reg ->
+                                fetchSample.ifPresent(sample ->
+                                        sample.stop(reg.timer("runtime.fetch_layout.timer"))));
```
Maithem, line 1159: What is the advantage of using Java's Optional over the No-op metric types that micrometer provides?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -269,6 +284,15 @@ public static CorfuRuntimeParametersBuilder builder() {
             private PriorityLevel priorityLevel = PriorityLevel.NORMAL;
             private Codec.Type codecType = Codec.Type.ZSTD;
             private MetricRegistry metricRegistry = null;
+            private MicroMeterRuntimeConfig microMeterRuntimeConfig =
+                    new MicroMeterRuntimeConfig(true,
+                    "CorfuMetrics", Duration.ofMinutes(1));
```
Maithem, line 289: Can you use the existing logger name https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/util/MetricsUtils.java#L212  ?

We have put out documents that use that logger's name, using a different one might be more confusing.

Maithem: It would be nice to have a read me and explicitly spell it out too, that is, the configuration of metrics.

runtime/pom.xml
```diff
@@ -167,6 +167,11 @@
             <version>6.0.53</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.2.3</version>
+        </dependency>
```
Maithem, line 174: Wouldn't this cause an issue if the class path already has slf4j bindings ?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -112,14 +120,26 @@ public AddressSpaceView(@Nonnull final CorfuRuntime runtime) {
...
-                this.hashCode());
-        metrics.register(pfx + "cache-size", (Gauge<Long>) readCache::size);
-        metrics.register(pfx + "evictions", (Gauge<Long>) () -> readCache.stats().evictionCount());
-        metrics.register(pfx + "hit-rate", (Gauge<Double>) () -> readCache.stats().hitRate());
-        metrics.register(pfx + "hits", (Gauge<Long>) () -> readCache.stats().hitCount());
-        metrics.register(pfx + "misses", (Gauge<Long>) () -> readCache.stats().missCount());
+        Optional<MeterRegistry> metricsRegistry = runtime.getRegistry();
+
+        metricsRegistry.ifPresent(registry ->
+                GuavaCacheMetrics.monitor(registry, readCache, "address_space.read_cache"));
```
Maithem, line 126: Can you add a gauge on the miss ratio and some metrics for the load time stats.

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
```diff
@@ -122,27 +124,24 @@ public StreamAddressSpace getStreamAddressSpace(StreamAddressRange streamsAddres
...
-        try (Timer.Context context = MetricsUtils.getConditionalContext(sequencerNextOneStream)) {
-            StreamsAddressResponse streamsAddressResponse = layoutHelper(e ->
-                    CFUtils.getUninterruptibly(e.getPrimarySequencerClient()
-                            .getStreamsAddressSpace(streamsAddressesRange)));
-            return streamsAddressResponse.getAddressMap();
-        }
+        StreamsAddressResponse streamsAddressResponse = layoutHelper(e ->
+                CFUtils.getUninterruptibly(e.getPrimarySequencerClient()
+                        .getStreamsAddressSpace(streamsAddressesRange)));
+        return streamsAddressResponse.getAddressMap();
```
Maithem, line 130: This is a stream range query (scans the bitsets on the sequencer), we should re-add that metric. Actually, there's a bug right now related to bitset queries #2747

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
PavelZaytsev: On the server-side, this would entail passing the registry as a parameter to a lot of classes that require instrumentation. This would make code more difficult to read, this would also require a lot of modifications. I would prefer to leave the registry as a singleton but I need to guarantee that the instance can be configured only once.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
PavelZaytsev: I did not know about no-op times. But I guess this also depends on the order the registries and the meters are instantiated. As far as I understand, the meters are no-oped unless they are registered. Since we configure the registry on startup first, depending on the passed arguments the registry is optionally getting created first, hence by composition all of the meters that are produced from it are also optional. 
Also having an optional registry allows for a more fluent code when it comes to lookup by key.

Maithem (approved):

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
PavelZaytsev: removed

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
PavelZaytsev: done

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
PavelZaytsev: done

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
PavelZaytsev: done

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
PavelZaytsev: done

runtime/pom.xml
PavelZaytsev: Removed for now

common/src/main/java/org/corfudb/common/metrics/micrometer/loggingsink/InfluxLineProtocolLoggingSink.java
PavelZaytsev: Removed

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
PavelZaytsev: Removed

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -679,32 +714,31 @@ protected void syncStreamUnsafe(ISMRStream stream, long timestamp) {
...
-                            log.error("Sync[{}] Error: Couldn't execute upcall due to {}", this, e);
-                            throw new UnrecoverableCorfuError(e);
-                        }
-                    });
-        }
-        if (timestamp != Address.OPTIMISTIC) {
-            histogram.update(counter.count);
+                        });
+        if (syncStreamTimer.isPresent()) {
+            syncStreamTimer.get().record(syncStreamRunnable);
```
zhangn49, line 739: Not sure if we need to use 2 different timers to record optimistic sync and regular sync

zhangn49 (approved):

PR #2809, opened 2020-10-30
DRAFT: CICD workflow: setup timeouts

xnull:
## Overview

Description:
Since cicd looks good after fixing a flaky unit test, let's postpone this change unless it's really needed. 
Please review https://github.com/CorfuDB/CorfuDB/pull/2810

Why should this be merged: 

Related issue(s) (if applicable): #2810 

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/CorfuTestParameters.java
+++ b/test/src/test/java/org/corfudb/CorfuTestParameters.java
@@ -86,22 +86,22 @@ public class CorfuTestParameters {
     public CorfuTestParameters(){
 
         // Timeouts
-        TIMEOUT_VERY_SHORT = Duration.of(100, MILLIS);
-        TIMEOUT_SHORT = Duration.of(1, SECONDS);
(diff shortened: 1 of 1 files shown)


WenbinZhu (approved):

WenbinZhu:
@xnull Why this is closed? Last time we said we need to restore the timeouts. The test you fixed in another patch is just just one, there are other tests randomly failing because of reduced timeout

xnull:
It's not entirely true, we were talking about if we can fix flaky tests it'd be better than increasing timeouts. But because we are limited in time let's bring timeouts back, so I had time to fix that test quickly. Also, I have restarted builds several times, they were stable. 
There are just a couple of tests that are failing, they will fail anyways even with increased timeouts, so it had better to fix them than increasing timeouts

xnull:
Just let me know if you find any other tests which fail, I'll try to fix them quickly, otherwise, we bring timeouts back

PR #2811, opened 2020-11-03
CICD workflow: fix flaky CorfuStoreShimTest#checkRevisionValidation

xnull:
## Overview

Description:
fix flaky CorfuStoreShimTest#checkRevisionValidation

Why should this be merged: 

- [ ] Public API has Javadoc


test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -248,7 +249,7 @@ public void checkRevisionValidation() throws Exception {
         }
         assertThat(entry.getMetadata().getRevision()).isEqualTo(2L);
         assertThat(entry.getMetadata().getCreateUser()).isEqualTo("user_1");
-        assertThat(entry.getMetadata().getLastModifiedTime()).isLessThan(System.currentTimeMillis());
+        assertThat(entry.getMetadata().getLastModifiedTime()).isLessThan(System.currentTimeMillis() + 1);
```
zfrenette, line 252: Maybe consider using `.isLessThanOrEqualTo()` instead?

zfrenette (approved):

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
hisundar, line 252: Please add a comment on why you are comparing with a future timestamp. I believe you did it to avoid having to sleep() which I think is a great idea, but would be good to document.

hisundar (commented):
Thanks, just a minor comment

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
xnull: Yeah, probably then @zfrenette 's idea of using `isLessThanOrEqualTo` would be self-describing, now it seems that we compare with the future

xnull: fone

PavelZaytsev (approved):

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
WenbinZhu: @xnull The change to isLessThanOrEqualTo() did not take effect in the patch

WenbinZhu (approved):

PR #2823, opened 2020-11-11
Protobuf Layout RPCs

xcchang:
## Overview

Description: Layout RPCs using Protobuf

Why should this be merged: Provides an implementation of Layout RPCs using Protobuf. Done as part of an effort to support rolling upgrades.

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -13,18 +15,29 @@
...
-import org.corfudb.protocols.wireprotocol.LayoutCommittedRequest;
-import org.corfudb.protocols.wireprotocol.LayoutMsg;
-import org.corfudb.protocols.wireprotocol.LayoutPrepareRequest;
-import org.corfudb.protocols.wireprotocol.LayoutPrepareResponse;
-import org.corfudb.protocols.wireprotocol.LayoutProposeRequest;
-import org.corfudb.protocols.wireprotocol.LayoutProposeResponse;
+import org.corfudb.runtime.proto.service.CorfuMessage.HeaderMsg;
 import org.corfudb.runtime.proto.service.CorfuMessage.RequestMsg;
+import org.corfudb.runtime.proto.service.CorfuMessage.RequestPayloadMsg;
+import org.corfudb.runtime.proto.service.CorfuMessage.ResponseMsg;
```
Maithem, line 21: Reformat imports.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -66,6 +79,7 @@
 
     /**
      * HandlerMethod for this server.
+     * [RM] Remove this after Protobuf for RPC Completion
      */
     @Getter
     private final HandlerMethods handler =
```
Maithem, line 85: Why can't you remove this in this patch? It's not needed. I think we should eagerly remove dead code paths because its hard to track them and try to delete them later.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -86,6 +100,7 @@
 
     @Override
     public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
+        //[RM] Remove this after Protobuf for RPC Completion
         return getState() == ServerState.READY;
```
Maithem: Throw an `UnsupportedOperationException` instead.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -118,6 +133,7 @@ public LayoutServer(@Nonnull ServerContext serverContext) {
 
     @Override
     protected void processRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+        //[RM] Remove this after Protobuf for RPC Completion
```
Maithem, line 136: Throw an `UnsupportedOperationException` instead.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
zfrenette: Note: It was made abstract in the other PR, so a getHandler method is required.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -132,6 +148,7 @@ public void shutdown() {
         executor.shutdown();
     }
 
+    //[RM] Remove this after Protobuf for RPC Completion
     private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
```
Maithem, line 152: Dead code, delete method.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
         return true;
     }
 
+    private boolean isBootstrapped(HeaderMsg reqHeader, ChannelHandlerContext ctx, IServerRouter r) {
```
Maithem, line 161: I would pass the `RequestMsg` and extract the header just before it is needed. Passing header around seems a little strange.

Maithem, line 161: The old implementation also logs a message. Can you put that back, but as a debug log?

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
-        if (!isBootstrapped(msg, ctx, r)) {
-            return;
-        }
+    // TODO this can work under a separate lock for this step as it does not change the global components
+    @RequestHandler(type = RequestPayloadMsg.PayloadCase.PREPARE_LAYOUT_REQUEST)
+    public synchronized void handlePrepareLayoutRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        final HeaderMsg requestHeader = req.getHeader();
+        final PrepareLayoutRequestMsg payload = req.getPayload().getPrepareLayoutRequest();
+
+        if(!isBootstrapped(requestHeader, ctx, r)) return;
```
Maithem, line 255: I no you didn't write this, but I think its cleaner to make the `isBootstrapped` such that it doesn't have a side effect. Maybe somethign like this:

```java
if(!isBootstrapped(requestHeader, ctx, r)) {
   r.sendNoBootstrapError(reqHeader, ctx);
}
```

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
+            HeaderMsg responseHeader = getHeaderMsg(req.getHeader(), false, true);
+            ResponseMsg response = getResponseMsg(responseHeader, getLayoutResponseMsg(getCurrentLayout()));
+            r.sendResponse(response, ctx);
         } else {
-            // else the client is somehow ahead of the server.
-            //TODO figure out a strategy to deal with this situation
-            r.sendResponse(ctx, msg, new CorfuPayloadMsg<>(CorfuMsgType.WRONG_EPOCH, serverEpoch));
-            log.warn("handleMessageLayoutRequest: Message Epoch {} ahead of Server epoch {}",
-                    msgEpoch, serverEpoch);
+            //TODO: Client is ahead of the server. Is any other handling required?
```
Maithem, line 191: You can remove the TODO this scenario can naturally happen. Imagine you run Paxos Layout at epoch 2. Its possible that out of the 3 nodes, only 2 nodes accepted epoch two layout. So, when a client requests a layout from the third node (i.e., the one that didn't participate in the paxos round, then it should just return the older layout. If you return a wrong epoch exception its fine, but superfluous because the clients need to check that they never regress the epoch the layout that they are tracking.

Maithem (commented):
@xcchang @zfrenette wouldn't it be easier if you negate the header boolean variables? for example instead of ignoreEpoch -> checkEpoch and ignoreClusterId -> checkClusterId.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
-            log.info("handleMessageLayoutBootstrap: Bootstrap with new layout={}, {}",
-                    layout, msg);
-
-            if (layout.getClusterId() == null) {
-                log.warn("handleMessageLayoutBootstrap: The layout does {} not have a clusterId",
-                        layout);
-                r.sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.NACK));
-            }
-            else{
+    @RequestHandler(type = RequestPayloadMsg.PayloadCase.BOOTSTRAP_LAYOUT_REQUEST)
```
Maithem, line 206: Can you drop the "REQUEST" suffix naming convention, it seems redundant. Maybe use `RequestPayloadMsg.PayloadCase.BOOTSTRAP_LAYOUT`

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
+    public synchronized void handleBootstrapLayoutRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        final HeaderMsg requestHeader = req.getHeader();
+        HeaderMsg responseHeader;
+        ResponseMsg response;
+
+        if(getCurrentLayout() == null) {
+            final Layout layout = getLayout(req.getPayload().getBootstrapLayoutRequest().getLayout());
+            log.info("handleBootstrapLayoutRequest[{}]: Bootstrap with new layout={}", requestHeader.getRequestId(), layout);
+
```
Maithem, line 215: You don't need to log the request id since its per-channel and only used to timeout/complete requests.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
-                r.sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.NACK));
-            }
-            else{
+    @RequestHandler(type = RequestPayloadMsg.PayloadCase.BOOTSTRAP_LAYOUT_REQUEST)
+    public synchronized void handleBootstrapLayoutRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        final HeaderMsg requestHeader = req.getHeader();
+        HeaderMsg responseHeader;
+        ResponseMsg response;
+
+        if(getCurrentLayout() == null) {
```
Maithem, line 212: Can you reverse the logic and get rid of the if-else here?

You can check `if(getCurrentLayout() != null)` you can send the error message and return right away. Otherwise continue to save the new bootstrap layout.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
+            final Layout layout = getLayout(req.getPayload().getBootstrapLayoutRequest().getLayout());
+            log.info("handleBootstrapLayoutRequest[{}]: Bootstrap with new layout={}", requestHeader.getRequestId(), layout);
+
+            if(layout.getClusterId() == null) {
+                log.warn("handleBootstrapLayoutRequest[{}]: The layout={} does not have a clusterId",
+                        requestHeader.getRequestId(), layout);
+
+                responseHeader = getHeaderMsg(requestHeader, false, false);
+                response = getResponseMsg(responseHeader,
+                        getBootstrapLayoutResponseMsg(BootstrapLayoutResponseMsg.Type.NACK));
```
Maithem, line 222: Un-nest this, move the validation to the top (after the `layout.getClusterId() == null` check)

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
+            } else {
                 setCurrentLayout(layout);
                 serverContext.setServerEpoch(layout.getEpoch(), r);
-                //send a response that the bootstrap was successful.
-                r.sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.ACK));
+
```
Maithem, line 226: Log the newly bootstrapped layout as `info`

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
Maithem, line 161: Instead of calling this from each handler, you can just move it to `processRequest`. Note that it needs to be called on all messages except for `LAYOUT_NOBOOTSTRAP`

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,280 +158,315 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
+            Rank highestProposedRank = proposedLayout == null ?
+                    new Rank(-1L, getUUID(requestHeader.getClientId())) : getPhase2Rank(payloadEpoch);
+
             setPhase1Rank(prepareRank, payloadEpoch);
-            log.debug("handleMessageLayoutPrepare: New phase 1 rank={}", prepareRank);
-            r.sendResponse(ctx, msg, CorfuMsgType.LAYOUT_PREPARE_ACK.payloadMsg(new
-                    LayoutPrepareResponse(highestProposedRank.getRank(), proposedLayout)));
+            log.debug("handlePrepareLayoutRequest[{}]: New phase 1 rank={}", requestHeader.getRequestId(), prepareRank);
+
+            responseHeader = getHeaderMsg(requestHeader, false, true);
```
Maithem, line 288: Why is this `responseHeader` different than the one in line 277

The response header should always copy its metadata from the requestHeader.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLayout.java
```diff
@@ -0,0 +1,116 @@
...
+                .setPrepareLayoutResponse(PrepareLayoutResponseMsg.newBuilder()
+                        .setRespType(type)
+                        .setRank(rank)
+                        .setLayout(getLayoutMsg(layout))
+                        .build())
+                .build();
+    }
+
+    public static LayoutPrepareResponse getLayoutPrepareResponse(PrepareLayoutResponseMsg msg) {
+        return new LayoutPrepareResponse(msg.getRank(), getLayout(msg.getLayout()));
```
Maithem, line 61: Not used?

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLayout.java
```diff
@@ -0,0 +1,116 @@
...
+import org.corfudb.runtime.proto.service.Layout.PrepareLayoutResponseMsg;
+import org.corfudb.runtime.proto.service.Layout.ProposeLayoutRequestMsg;
+import org.corfudb.runtime.proto.service.Layout.ProposeLayoutResponseMsg;
+import org.corfudb.runtime.view.Layout;
+
+import static org.corfudb.protocols.CorfuProtocolCommon.getLayout;
+import static org.corfudb.protocols.CorfuProtocolCommon.getLayoutMsg;
+
+@Slf4j
+public class CorfuProtocolLayout {
```
Maithem, line 23: Wouldn't it be cleaner to pass the requestHeader to these methods and have these methods construct and return the final message. Instead of doing this construction all over the place it can be in one place, here.

runtime/src/main/java/org/corfudb/runtime/clients/LayoutClient.java
```diff
@@ -27,13 +28,15 @@ public LayoutClient(IClientRouter router, long epoch, UUID clusterID) {
         super(router, epoch, clusterID);
     }
 
+    // Protobuf region
+
```
Maithem, line 32: Remove. All of this uses protobuf messages now. Also, don't forget to remove Layout requests/responses from CorfuMsgType.

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
```diff
@@ -94,4 +105,116 @@ private static Object handleAlreadyBootstrap(CorfuMsg msg,
             throws Exception {
         throw new AlreadyBootstrappedException();
     }
+
+    // Protobuf region
+
+    /**
```
Maithem, line 111: Remove all usages of `CorfuMsgType` with these changes those paths become inactive. Its better to remove them than assume that they are not functional.

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
```diff
@@ -94,4 +105,116 @@ private static Object handleAlreadyBootstrap(CorfuMsg msg,
...
+                                                     IClientRouter r) {
+        CommitLayoutResponseMsg commitLayoutMsg = msg.getPayload().getCommitLayoutResponse();
+        CommitLayoutResponseMsg.Type type = commitLayoutMsg.getRespType();
+
+        switch (type) {
+            case ACK: return true;
+            case NACK: return false;
+            // TODO for INVALID
+            default: throw new UnsupportedOperationException("Response handler not provided");
+        }
```
Maithem, line 194: Why don't you just use a boolean to represent this.

```proto
message CommitLayoutResponseMsg {
      bool committed = 1;
}

```

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
```diff
@@ -94,4 +105,116 @@ private static Object handleAlreadyBootstrap(CorfuMsg msg,
...
+    @ResponseHandler(type = PayloadCase.BOOTSTRAP_LAYOUT_RESPONSE)
+    private static Object handleBootstrapLayoutResponse(ResponseMsg msg, ChannelHandlerContext ctx,
+                                                        IClientRouter r) {
+        BootstrapLayoutResponseMsg bootstrapLayoutMsg =  msg.getPayload().getBootstrapLayoutResponse();
+        BootstrapLayoutResponseMsg.Type type = bootstrapLayoutMsg.getRespType();
+
+        switch (type) {
+            case ACK: return true;
+            case NACK: return false;
+            // TODO for INVALID
```
Maithem, line 214: Cleaner to use a boolean in the protobuf message. See comment above.

Maithem (commented):
Looks good, but needs more clean up. 

All the new handlers (and each branch) needs to be tested with a proper unit tests (i.e. by mocking the channel context).

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
zfrenette: Note that checking this in `processRequest `means we are performing the check before the task is submitted to the executor service.  Is it possible that the status changes by the time the new thread is actually executing the handler?

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
zfrenette: All of the fields are the same, except possibly the `ignoreEpoch `or `ignoreClusterId ` fields. These differ because the message in 277 corresponds to a NACK/REJECT while here it corresponds to an ACK. Note that in the original `CorfuMsgType ` implementation, an ACK will ignore the epoch check while a NACK does not.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLayout.java
zfrenette: Good catch. It should be used in the LayoutHandler when returning the response back to the caller.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLayout.java
zfrenette: The reason why we don't pass the header at this stage is because the client class (i.e. LayoutClient) is currently responsible for building the message payload, but it doesn't have all of the information required to build the header, most notably the request id, which is only available once we reach the router. So if the client builds a RequestMsg directly (consisting of HeaderMsg and RequestPayloadMsg), then the router needs to rebuild the object in order to stamp the request id. Therefore, the client builds the payload and passes its information to the router, and its the router that then builds the header and wraps everything together into a `RequestMsg`.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLayout.java
```diff
@@ -0,0 +1,116 @@
...
+    public static RequestPayloadMsg getLayoutRequestMsg(long epoch) {
```
zfrenette, line 24: Can we add the Javadocs for the methods in this class?

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
```diff
@@ -94,4 +105,116 @@ private static Object handleAlreadyBootstrap(CorfuMsg msg,
...
+    @ResponseHandler(type = PayloadCase.PREPARE_LAYOUT_RESPONSE)
+    private static Object handlePrepareLayoutResponse(ResponseMsg msg, ChannelHandlerContext ctx,
+                                                      IClientRouter r) {
+        PrepareLayoutResponseMsg prepareLayoutMsg = msg.getPayload().getPrepareLayoutResponse();
+        PrepareLayoutResponseMsg.Type type = prepareLayoutMsg.getRespType();
+        long rank = prepareLayoutMsg.getRank();
+        Layout layout = CorfuProtocolCommon.getLayout(prepareLayoutMsg.getLayout());
+
+        switch (type) {
+            case ACK: return new LayoutPrepareResponse(rank, layout);
```
zfrenette, line 145: `LayoutPrepareResponse` no longer needs to implement `ICorfuPayoad`. It can be removed, along with the related serialization methods.

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
xcchang: I think when @zfrenette design these message types, enum types were used to make it easier for extension. For example when the message type needs to change sometime in the future, to model something like ACK, NACK-FAILURE-REASON-A, NACK-FAILURE-REASON-B, enum might provide more flexibility and easier for backward compatibility. But definitely using a boolean would make it cleaner at this time. We might need to make a tradeoff here.

Maithem: I would just go with a boolean, what do you think @zfrenette

test/src/test/java/org/corfudb/runtime/view/AbstractViewTest.java
```diff
@@ -469,6 +474,12 @@ public String getEndpoint(int port) {
 
     // Private
 
+    private RequestMsg getRequest(CorfuMessage.RequestPayloadMsg payloadMsg) {
+        CorfuMessage.HeaderMsg header = getHeaderMsg(0, CorfuMessage.PriorityLevel.NORMAL, 0,
+                DEFAULT_UUID, DEFAULT_UUID, true, true);
+        return getRequestMsg(header, payloadMsg);
+    }
+
```
chetangudisagar, line 482: We need the default header for the next set of handleMessage() calls of `MANAGEMENT_BOOTSTRAP_REQUEST` and `BOOTSTRAP_SEQUENCER`.
Therefore, Can we move this to `CorfuProtocolMessage`?

We can add a new method to get default header and call `CorfuProtocolMessage.getRequestMsg()` with that header and the new `payloadMsg`.

```
public static HeaderMsg getDefaultHeaderMsg(boolean ignoreClusterId, boolean ignoreEpoch) {
    return getHeaderMsg(0, CorfuMessage.PriorityLevel.NORMAL, 0,
        DEFAULT_UUID, DEFAULT_UUID, ignoreClusterId, ignoreEpoch);
}
```
**Note:** Unfortunately, we cannot use `HeaderMsg.getDefaultInstance()` provided by Protobuf as it sets the values to empty instead of the default intended values. Its output is -
```
headerMsg.getRequestId() - 0
headerMsg.getClientId() - 
headerMsg.getClusterId() - 
headerMsg.getEpoch() - 0
headerMsg.getIgnoreEpoch() - false
headerMsg.getIgnoreClusterId() - false
headerMsg.getPriority() - PRIORITY_LEVEL_INVALID
headerMsg.getVersion() - 
```

chetangudisagar: **Update:** After an internal discussion, we decided to add a temporary variable within the method to be reused by all 3 request messges -
```
public void bootstrapAllServers(Layout l) {
    CorfuMessage.HeaderMsg headerMsg = getHeaderMsg(0, CorfuMessage.PriorityLevel.NORMAL, 0,
            DEFAULT_UUID, DEFAULT_UUID, true, true);
```

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
xcchang: I had a discuss with @zfrenette offline and we thought it might be okay to keep the suffix, since sometimes we directly import the `PayloadCase` rather than `RequestPayloadMsg.PayloadCase`, the suffix might help to tell the type of the message. (if we remove the suffix, both the `RequestPayloadMsg` and `ResponsePayloadMsg` will have a `PayloadCase.PING` for example)

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuLayoutServerTest.java
```diff
@@ -0,0 +1,302 @@
...
+        // Initialize with newDirectExecutorService to execute the server RPC
+        // handler methods on the calling thread
+        when(mockServerContext.getExecutorService(anyInt(), anyString()))
+                .thenReturn(MoreExecutors.newDirectExecutorService());
+
+        // Initialize an in memory DataStore for LayoutServer
+        when(mockServerContext.getDataStore()).thenReturn(
+                new DataStore(new ImmutableMap.Builder<String, Object>()
+                .put("--memory", true)
+                .build(), fn -> { }));
```
zfrenette, line 111: Would it be better to use a mock of the DataStore instead?

xcchang: This DataStore is used to initialize the paxosDataStore field of LayoutServer, which is needed to get phase1rank when handling PROPOSE_LAYOUT_REQUEST for example, and it's not exposed. So some changes may be necessary if we do want to use mocking for paxosDataStore, like moving it to serverContext?

runtime/proto/service/layout.proto
```diff
@@ -46,13 +40,7 @@ message ProposeLayoutRequestMsg {
...
 
 message ProposeLayoutResponseMsg {
-  enum Type {
-    INVALID = 0;
-    ACK = 1;
-    REJECT = 2;
-  }
-
-  Type resp_type = 1;
+  bool proposed = 1;
```
xnull, line 43: please bring back enums. Boolean variables much harder to understand - the value doesn't explain domain meaning, and that brings additional complexity and makes code more indirect

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
```diff
@@ -52,46 +59,116 @@ public LayoutClient getClient(long epoch, UUID clusterID) {
...
+     * @param r A reference to the router.
+     * @return {@link LayoutPrepareResponse} if ACK, throw an {@link OutrankedException} if REJECT.
+     */
+    @ResponseHandler(type = PayloadCase.PREPARE_LAYOUT_RESPONSE)
+    private static Object handlePrepareLayoutResponse(ResponseMsg msg, ChannelHandlerContext ctx,
+                                                      IClientRouter r) {
+        PrepareLayoutResponseMsg prepareLayoutMsg = msg.getPayload().getPrepareLayoutResponse();
+        boolean prepared = prepareLayoutMsg.getPrepared();
+        long rank = prepareLayoutMsg.getRank();
+        Layout layout = CorfuProtocolCommon.getLayout(prepareLayoutMsg.getLayout());
```
xnull, line 114: static methods make code less maintainable/testable

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuLayoutServerTest.java
zfrenette: It is already available from the server-context:
```
        this.paxosDataStore = PaxosDataStore.builder()
                .dataStore(serverContext.getDataStore())
                .build();
```
You can continue to use the PaxosDataStore wrapper, but the data store itself can already be mocked easily since we are mocking the server context.

runtime/src/test/java/org/corfudb/runtime/clients/CorfuLayoutHandlerTest.java
```diff
@@ -0,0 +1,235 @@
...
+        verify(mockClientRouter).completeRequest(responseNACK.getHeader().getRequestId(), false);
+    }
+
+    /**
+     * Test that the LayoutHandler correctly handles a PREPARE_LAYOUT_RESPONSE.
+     */
+    @Test
+    public void testPrepare() throws IOException {
+        Layout defaultLayout = getDefaultLayout();
+        long defaultRank = 0L;
```
zfrenette, line 138: Maybe use a non-zero value here? It could make some failures more noticeable.

runtime/src/test/java/org/corfudb/runtime/clients/CorfuLayoutHandlerTest.java
```diff
@@ -0,0 +1,235 @@
...
+        assertThat(retLayout.getSequencers()).containsExactly("localhost:9000");
+        assertThat(retLayout.getAllLogServers()).containsExactly("localhost:9002", "localhost:9001", "localhost:9000");
+    }
+
+    /**
+     * Test that the LayoutHandler correctly handles a PROPOSE_LAYOUT_RESPONSE.
+     */
+    @Test
+    public void testPropose() {
+        long defaultRank = 0L;
```
zfrenette, line 179: Maybe use a non-zero value here too.

runtime/src/test/java/org/corfudb/runtime/clients/CorfuLayoutHandlerTest.java
```diff
@@ -0,0 +1,235 @@
...
+    public void setup() {
+        mockClientRouter = mock(IClientRouter.class);
+        mockChannelHandlerContext = mock(ChannelHandlerContext.class);
+        layoutHandler = new LayoutHandler();
+        layoutHandler.setRouter(mockClientRouter);
+    }
+
+    /**
+     * Test that the LayoutHandler correctly handles a LAYOUT_RESPONSE.
+     */
```
zfrenette, line 80: Would it be worth considering the case where deserializing the Layout throws an exception, and seeing that we complete exceptionally?

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuLayoutServerTest.java
```diff
@@ -0,0 +1,710 @@
...
+        ResponseMsg response = responseCaptor.getValue();
+        // Assert that the payload has a BOOTSTRAP_LAYOUT_RESPONSE and that the base
+        // header fields have remained the same
+        assertTrue(compareBaseHeaderFields(request.getHeader(), response.getHeader()));
+        assertTrue(response.getPayload().hasBootstrapLayoutResponse());
+        assertTrue(response.getPayload().getBootstrapLayoutResponse().getBootstrapped());
+    }
+
+    /**
+     * Test that the LayoutServer correctly handles a BOOTSTRAP_LAYOUT_REQUEST with null clusterId.
```
zfrenette, line 146: Maybe consider adding a cause of a null clusterId in the comment description (i.e. ... correctly handles a BOOTSTRAP_LAYOUT_REQUEST with a legacy layout having a null clusterId)

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuLayoutServerTest.java
```diff
@@ -0,0 +1,710 @@
...
+        verify(mockServerRouter).sendResponse(responseCaptor.capture(), any(ChannelHandlerContext.class));
+
+        ResponseMsg response = responseCaptor.getValue();
+        // Assert that the payload has a LAYOUT_RESPONSE and that the base
+        // header fields have remained the same
+        assertTrue(compareBaseHeaderFields(request.getHeader(), response.getHeader()));
+        assertTrue(response.getPayload().hasServerError());
+        // Check that Layout Server handler sends a WrongEpochErrorMsg
+        ServerErrorMsg serverErrorMsg = response.getPayload().getServerError();
+        assertTrue(serverErrorMsg.hasWrongEpochError());
```
zfrenette, line 576: Can you also verify that the correct epoch is what you expect?

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuLayoutServerTest.java
```diff
@@ -0,0 +1,710 @@
...
+        Layout l = getDefaultLayout();
+        long payloadEpoch = 0L;
+        RequestMsg request = getRequestMsg(
+                getBasicHeader(false, true),
+                getCommitLayoutRequestMsg(false, payloadEpoch, l)
+        );
+
+        ArgumentCaptor<ResponseMsg> responseCaptor = ArgumentCaptor.forClass(ResponseMsg.class);
+        ArgumentCaptor<Layout> layoutCaptor = ArgumentCaptor.forClass(Layout.class);
+        ArgumentCaptor<Long> serverEpochCaptor = ArgumentCaptor.forClass(Long.class);
```
zfrenette, line 525: Are these argument captors necessary? I think calling verify with the expected object should suffice here. In most cases, if you know the result ahead of time, you don't need an argument captor.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuLayoutServerTest.java
```diff
@@ -0,0 +1,710 @@
...
+        Layout l = getDefaultLayout();
+        long payloadEpoch = 0L;
+        RequestMsg request = getRequestMsg(
+                getBasicHeader(false, true),
+                getCommitLayoutRequestMsg(true, payloadEpoch, l)
+        );
+
+        ArgumentCaptor<ResponseMsg> responseCaptor = ArgumentCaptor.forClass(ResponseMsg.class);
+        ArgumentCaptor<Layout> layoutCaptor = ArgumentCaptor.forClass(Layout.class);
+        ArgumentCaptor<Long> serverEpochCaptor = ArgumentCaptor.forClass(Long.class);
```
zfrenette, line 593: Same here.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -127,10 +121,9 @@ public void shutdown() {
         executor.shutdown();
     }
 
-    private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
+    private boolean isBootstrapped(RequestMsg requestMsg) {
```
Maithem, line 124: Notice that this check is called on all the handlers except for the actual bootstrap call. I would recommend that this check is executed in the entry point, i.e, processRequest.

```java
@Override
    protected void processRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
        executor.submit(() -> {


            if (!req.getPayload().getPayloadCase().equals(RequestPayloadMsg.PayloadCase.BOOTSTRAP_LAYOUT_REQUEST)
            && !isBootstrapped(req)) {
                r.sendNoBootstrapError(req.getHeader(), ctx);
                return;
            }

            // Don't process layout server requests unless the server is bootstrapped 
            getHandlerMethods().handle(req, ctx, r);
        });
    }
```

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,275 +134,317 @@ private boolean isBootstrapped(CorfuMsg msg, ChannelHandlerContext ctx, IServerR
...
+     * @param req              corfu message containing LAYOUT_REQUEST
      * @param ctx              netty ChannelHandlerContext
      * @param r                server router
      */
-    @ServerHandler(type = CorfuMsgType.LAYOUT_REQUEST)
-    public synchronized void handleMessageLayoutRequest(CorfuPayloadMsg<Long> msg,
-                                                    ChannelHandlerContext ctx, IServerRouter r) {
-        if (!isBootstrapped(msg, ctx, r)) {
+    @RequestHandler(type = RequestPayloadMsg.PayloadCase.LAYOUT_REQUEST)
+    public synchronized void handleLayoutRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
```
Maithem, line 142: I know this predates your patch, but honestly this doesn't make sense at all: having a multi-threaded executor to server the request handlers and marking all the handlers as synchronized is just lol. 

Can you just change `this.executor`  to be single threaded and drop all the synchronized ?

Maithem: With Biased locking having synchronized without contention should be cheap, so you can keep the synchronized methods, but change the executor to a single thread.

I would just remove the synchronized though, it seems super unnecessary.

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
```diff
@@ -52,46 +59,116 @@ public LayoutClient getClient(long epoch, UUID clusterID) {
...
+     * @param ctx The context the message was sent under.
+     * @param r A reference to the router.
+     * @return True if ACK, false if NACK.
+     */
+    @ResponseHandler(type = PayloadCase.BOOTSTRAP_LAYOUT_RESPONSE)
+    private static Object handleBootstrapLayoutResponse(ResponseMsg msg, ChannelHandlerContext ctx,
+                                                        IClientRouter r) {
+        BootstrapLayoutResponseMsg bootstrapLayoutMsg =  msg.getPayload().getBootstrapLayoutResponse();
+
+        return bootstrapLayoutMsg.getBootstrapped();
```
Maithem, line 172: This is a bug. The clients are expecting an exception to be thrown when this response is sent back, but this will return a value.

Maithem: If you look here: https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/view/LayoutView.java#L328

A lot of code paths that rely on the exception to be thrown won't execute, because this changes the semantics of the method, but code that consumes this wasn't changed.

Maithem: Ops, I'm wrong. 

There's a test for this and it is handled in the base handler. Kind of confusing when some errors return a false bool and some throw an exception.

runtime/src/test/java/org/corfudb/runtime/clients/CorfuLayoutHandlerTest.java
```diff
@@ -0,0 +1,263 @@
...
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+@Slf4j
+@RunWith(MockitoJUnitRunner.class)
+public class CorfuLayoutHandlerTest {
```
Maithem, line 44: Can you remove the tests in LayoutHandlerTest and replace them with these tests?

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
```diff
@@ -52,46 +59,116 @@ public LayoutClient getClient(long epoch, UUID clusterID) {
...
 
+    /**
+     * TODO: Remove this method when Management RPC get landed.
+     */
+    @Deprecated
+    @ClientHandler(type = CorfuMsgType.LAYOUT_NOBOOTSTRAP)
+    private static Object handleNoBootstrap(CorfuMsg msg,
+                                            ChannelHandlerContext ctx, IClientRouter r)
+            throws Exception {
+        throw new NoBootstrapException();
```
Maithem, line 70: Can you delete these instead of deprecating ?

They are not exposed publicly and we dont support backward compatibility.

Maithem (approved):
Minor comments and questions Otherwise looks great!

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuLayoutServerTest.java
```diff
@@ -0,0 +1,710 @@
...
+        assertTrue(compareBaseHeaderFields(request.getHeader(), response.getHeader()));
+        assertTrue(response.getPayload().hasServerError());
+        // Check that Layout Server handler sends a WrongEpochErrorMsg
+        ServerErrorMsg serverErrorMsg = response.getPayload().getServerError();
+        assertTrue(serverErrorMsg.hasWrongEpochError());
+    }
+
+    /**
+     * Test that the LayoutServer correctly handles a PREPARE_LAYOUT_REQUEST.
+     */
```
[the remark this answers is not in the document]
xcchang: Added, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
xcchang: Yeah it's called for most handlers. But @zfrenette formerly brought up a concern that, by this way we are performing the check before the task is submitted to the executor service, and the status could change by the time the new thread is actually executing the handler. (https://github.com/CorfuDB/CorfuDB/pull/2823#discussion_r523764871) I think it's reasonable, but would like to know your opinion for this.

xcchang: but hmm followed by your next comment, those handlers are literally single threaded, then it should be fine to move this check to `processRequest`

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
xcchang: Yeah they will be deleted after the following PRs get merged. Currently they are still used by some other CorfuMsg RPCs even if all the Layout RPCs has been replaced by Protobuf.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LayoutPrepareResponse.java
```diff
@@ -19,30 +19,7 @@
...
 @AllArgsConstructor
-public class LayoutPrepareResponse implements ICorfuPayload<LayoutPrepareResponse> {
+public class LayoutPrepareResponse {
     private long rank;
     private Layout layout;
-
-    /**
-     * Constructor for layout server response in first phase of Paxos.
-     */
-    public LayoutPrepareResponse(ByteBuf buf) {
```
zfrenette, line 29: You can also remove the unused ByteBuf import here.

runtime/src/test/java/org/corfudb/runtime/clients/CorfuLayoutHandlerTest.java
xcchang: How about just delete the old `LayoutHanlderTest`?

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
xcchang: Removed the synchronized, thanks!

zfrenette (approved):

Maithem:
Are you still planning to push this PR ?

xcchang:
> Are you still planning to push this PR ?

xcchang:
> Are you still planning to push this PR ?

No I won't push to this PR. The latest commits were from rebasing it onto master.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
Maithem: The status cannot change if only one thread is executing all handlers

runtime/src/main/java/org/corfudb/runtime/clients/LayoutHandler.java
Maithem: Sure, but in this case its a It's a stateless helper method. Anyways, if you still want this to change, the please propose something more actionable.

runtime/proto/service/layout.proto
Maithem: Please address this.

PR #2824, opened 2020-11-11
CorfuStore: Allow re-entrant Transactions, full Queue over protobuf API support

hisundar:
## Overview

Description:

Why should this be merged: 

Without re-entrant transaction support, all callers have to rewrite very large sections of their layers to pass around the TxnContext. Also in many cases the CorfuStore Transaction needs to work within code that already has a transaction management.

CorfuQueue will be deprecated and its functionality is to be accessed from the CorfuStore layer directly.

- [x] Public API has Javadoc


WenbinZhu:
@hisundar Could you tell what is the ask here and what is the uses cases that requires this capability?

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -494,4 +505,57 @@ public void onCommit(Map<String, List<CorfuStreamEntry>> mutations) {
...
+
+    /**
+     * Validate that nested transactions do not throw exception if txnWithNesting is used.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void checkNestedTransaction() throws Exception {
+// Get a Corfu Runtime instance.
+        CorfuRuntime corfuRuntime = getTestRuntime();
```
chetangudisagar, line 517: **Nit:** Can we indent the comment with the parameter?

hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/TxnContextShim.java
```diff
@@ -7,46 +7,59 @@
...
- * Thin shim layer around CorfuStore's TxnContext for providing metadata management
- * capabilities.
+ * Shim layer around CorfuStore's TxnContext for providing metadata management
+ * capabilities, nested transaction support and more!
  *
  * Created by hisundar on 2020-09-16
  */
 @Slf4j
-public class TxnContextShim extends TxnContext implements AutoCloseable {
+public class TxnContextShim implements AutoCloseable {
```
medhavidhawan, line 32: The Shim prefix for the name is not that great. Can we change it to a better name.

runtime/src/main/java/org/corfudb/runtime/collections/TxnContextShim.java
```diff
@@ -230,6 +608,92 @@ void putRecord(@Nonnull Table<K, V, M> table,
...
+            for (Descriptors.FieldDescriptor fieldDescriptor : deltaMetadata.getDescriptorForType().getFields()) {
+                if ("revision".equals(fieldDescriptor.getName())) {
+                    if (oldRecord == null || oldRecord.getMetadata() == null) {
+                        return null;
+                    } else {
+                        Long prevRevision = (Long) oldRecord.getMetadata().getField(fieldDescriptor);
+                        Long givenRevision = (Long) deltaMetadata.getField(fieldDescriptor);
+                        if ((givenRevision > 0 && // Validate revision only if set
+                                prevRevision.longValue() == givenRevision.longValue())
+                                || givenRevision == 0) { // Do not validate revision if field isn't set
```
medhavidhawan, line 685: nit:since you use short circuit operator || the faster check could be first.

medhavidhawan (approved):
Minor comments: Please make sure to delete the deprecated API txnWithNesting ASAP

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -607,6 +633,24 @@ boolean isExists(@Nonnull String tableName, @Nonnull final K key) {
         return table.scanAndFilterByEntry(record -> true);
     }
 
+    /**
+     * @return The the thread local's TxnContext, null if not in a transaction.
+     */
+    public static TxnContext getMyTxnContext() {
```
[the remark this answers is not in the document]
hisundar: done, thanks

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -652,10 +710,16 @@ private void txBeginInternal() {
...
-        if (!TransactionalContext.isInTransaction()) {
+        if (!isInMyTransaction()) {
             throw new IllegalStateException("commit() called without a transaction!");
         }
+
+        // Apply any buffered up operations.
         operations.forEach(Runnable::run);
+
+        // Regardless of transaction outcome remove any TxnContext association from ThreadLocal.
+        TransactionalContext.getRootContext().setTxnContext(null);
```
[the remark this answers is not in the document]
hisundar: yes, then the auto-closeable will catch it and abort the transaction

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -623,13 +667,26 @@ void applyWritesForReadOnTable(Table<K, V, M> tableBeingRead) {
...
+            TxnContext txnContext = TransactionalContext.getRootContext().getTxnContext();
+            if (!allowNestedTransactions) {
+                log.error("Current thread already has a transaction started and nested transactions are disabled!");
+                throw new TransactionAlreadyStartedException(TransactionalContext.getRootContext().toString());
+            }
+            if (txnContext != null) {
+                log.error("Cannot start new transaction in this thread without ending previous one");
+                throw new TransactionAlreadyStartedException(TransactionalContext.getRootContext().toString());
+            }
+            this.txnStartTime = System.nanoTime();
```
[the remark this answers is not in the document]
hisundar: it's best effort given that we did not actually start the transaction

runtime/src/main/java/org/corfudb/runtime/collections/TxnContextShim.java
hisundar: done. thanks

runtime/src/main/java/org/corfudb/runtime/collections/TxnContextShim.java
hisundar: ok, changed to ManagedTxnContext

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -164,21 +184,27 @@ void putRecord(@Nonnull Table<K, V, M> table,
...
             try {
-                mergedRecord = mergeCallback.doMerge(table, oldRecord, recordDelta);
+                mergedRecord = mergeCallback.doMerge(table, key, oldRecord, recordDelta);
             } catch (Exception ex) {
                 txAbort(); // explicitly abort this transaction and then throw the abort manually
                 log.error("TX Abort merge: {}", table.getFullyQualifiedTableName(), ex);
                 throw ex;
             }
-            table.put(key, mergedRecord.getPayload(), mergedRecord.getMetadata());
+            if (mergedRecord == null) {
```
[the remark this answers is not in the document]
hisundar: this was to actually fix a bug where delete with metadata wasn't being handled. I am open to suggestions though.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -206,7 +205,8 @@ public TxnContext txn(@Nonnull final String namespace, IsolationLevel isolationL
                 this.runtime.getObjectsView(),
                 this.runtime.getTableRegistry(),
                 namespace,
-                isolationLevel);
+                isolationLevel,
+                false);
```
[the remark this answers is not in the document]
hisundar: done

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -86,13 +87,15 @@ public TxnContext(@Nonnull final ObjectsView objectsView,
...
+     * @param table         - table being written to
+     * @param key           - key used in the transaction to check for null.
+     * @param validateKey   - should key be validated for null.
+     * @param <K>           - type of the key
+     * @param <V>           - type of the payload/value
+     * @param <M>           - type of the metadata
      */
     private <K extends Message, V extends Message, M extends Message>
-    void validateTableWrittenIsInNamespace(@Nonnull Table<K, V, M> table) {
+    void validateWrite(@Nonnull Table<K, V, M> table, K key, boolean validateKey) {
```
Maithem, line 140: Probably should be validateNameSpace and not write. Kind of confusing when invoked for deletes. Also, for writes the Key can never be null, so its not even optional for a map.

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -102,10 +105,24 @@ void validateTableWrittenIsInNamespace(@Nonnull Table<K, V, M> table) {
                     "TxnContext cannot be used after a transaction has ended on "+
                     table.getFullyQualifiedTableName());
         }
+        if (validateKey && key == null) {
+            throw new IllegalArgumentException("Key cannot be null on "
+                    + table.getFullyQualifiedTableName() + " in transaction on namespace " + namespace);
+        }
```
Maithem, line 153: Seems a little weird to manipulate the bitmask in a validate method.

Maithem (commented):
The OO modeling is a little complicated and there's a lot of code duplication and lack of unit tests for the "wrapper" classes.

Consider having having one transaction type and two implementations of that type. Exposing a new (ManagedTxnContext) external transaction type that is almost identical to the existing one (TxnContext) just for the shim layer can be confusing. If you unify the transaction type then you can git rid of the shim layer. Later on instead of deprecating the shim and nested transactions, you can just deprecate one type of implementation. Which also improves code coverage by reusing the CorfuStore unit tests for both Transaction implementations.

I would also suggest improving the naming, we currently have TransactionalContext, TxnContext and ManagedTxnContext.

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
hisundar: You'd be surprised then :)
Key can never be null but the way it fails when the table.put() is invoked with the key being null is not pretty..
`ObjectsView 9496 TXEnd[TX[9672]]: Unexpected exception
java.lang.NullPointerException: null
	at org.corfudb.util.serializer.ISerializer.hash(ISerializer.java:123) ~[runtime-3.2.20201109031413.7872.1.jar:?]
	at org.corfudb.runtime.object.transactions.ConflictSetInfo.generateHashFromObject(ConflictSetInfo.java:28) ~[runtime-3.2.20201109031413.7872.1.jar:?]
	at org.corfudb.runtime.object.transactions.ConflictSetInfo.lambda$null$1(ConflictSetInfo.java:41) ~[runtime-3.2.20201109031413.7872.1.jar:?]
	at`
we do need to validate for deletes as well because sometimes buggy code causes callers to send in null into the key.
We need to catch the error much early on for debug-ability.

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
hisundar: sure, open to any alternate suggestions for a better name.
How about `preProcessMutation()` ?

hisundar:
> The OO modeling is a little complicated and there's a lot of code duplication and lack of unit tests for the "wrapper" classes.
> 

There is API duplication because ManagedTxnContext uses composition over inheritance for supporting nested transactions with the auto closable pattern while still re-using the thread local TxnContext. There shouldn't be code duplication. 
Also almost every single api and pattern has a unit test covering it, please check out CorfuStoreShimTest.java for code coverage tests for the wrapper classes.

> Consider having having one transaction type and two implementations of that type. Exposing a new (ManagedTxnContext) external transaction type that is almost identical to the existing one (TxnContext) just for the shim layer can be confusing. If you unify the transaction type then you can git rid of the shim layer. Later on instead of deprecating the shim and nested transactions, you can just deprecate one type of implementation. Which also improves code coverage by reusing the CorfuStore unit tests for both Transaction implementations.
> 

The reason ManagedTxnContext still needs to be separate is because verticals will be enhancing that layer by adding context specific validations and processing that are outside the scope of the public corfudb repository.

> I would also suggest improving the naming, we currently have TransactionalContext, TxnContext and ManagedTxnContext.
Naming is hard, so I am open to suggestions. Please see @medhavidhawan 's comment based on which I changed it to ManagedTxnContext.
thanks

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -473,25 +482,125 @@ public void onCommit(Map<String, List<CorfuStreamEntry>> mutations) {
...
         }
 
-        try (TxnContextShim txn = shimStore.txn(someNamespace)) {
-            txn.delete(tableName, key);
-            txn.commit((mutations) -> {
-                 mutations.values().forEach(mutation -> {
-                     CorfuStreamEntry.OperationType op = mutation.get(0).getOperation();
-                     assertThat(op).isEqualTo(CorfuStreamEntry.OperationType.DELETE);
-                 });
+        CommitCallbackImpl commitCallback = new CommitCallbackImpl();
```
Lujie1996, line 485: Where is this commitCallback used?

Lujie1996 (approved):

achinmishra (approved):

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
hisundar: it is used by verticals to extract the modified metadata. It can also serve as a way to do fast table notifications.

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -649,21 +745,35 @@ private void txBeginInternal() {
...
-            commitAddress = this.objectsView.TXEnd();
-        } catch (Exception ex) {
-            tablesInTxn.values().forEach(t -> t.getMetrics().incNumTxnAborts());
-            tablesInTxn.clear();
-            throw ex;
+
+        // CorfuStore should have only one transactional context since nesting is prohibited.
+        AbstractTransactionalContext rootContext = TransactionalContext.getRootContext();
+        // Regardless of transaction outcome remove any TxnContext association from ThreadLocal.
+        TransactionalContext.getRootContext().setTxnContext(null);
```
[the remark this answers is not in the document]
hisundar: done. thanks

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
Lujie1996: But I couldn't find the usage of this `commitCallback` variable in this test. The following `txn.addCommitCallback()` adds its own callback lambda function right?

PR #2826, opened 2020-11-13
CorfuStore streaming redesign part 3 - implementation and integration.

WenbinZhu:
## Overview

Description:

Implements the new subscription API and new threading model for
streaming and integrate into CorfuStore.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
@@ -19,12 +19,12 @@ import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * CorfuStore is a protobuf API layer that provides all the features of CorfuDB.
- *
+ * <p>
  * Key APIs exposed are:
- *   o-> TxnContext() for CRUD operations
- *   o-> getTimestamp() for database snapshots
- *   o-> table lifecycle management
- *
+ * o-> TxnContext() for CRUD operations
+ * o-> getTimestamp() for database snapshots
+ * o-> table lifecycle management
+ * <p>
  * Created by zlokhandwala on 2019-08-02.
  */
 public class CorfuStore {
@@ -33,6 +33,7 @@ public class CorfuStore {
     private final CorfuRuntime runtime;
 
     private final Optional<AtomicLong> openTableCounter;
+
     /**
      * Creates a new CorfuStore.
      *
@@ -46,7 +47,7 @@ public class CorfuStore {
     /**
      * Creates a new CorfuStore.
      *
-     * @param runtime Connected instance of the Corfu Runtime.
+     * @param runtime         Connected instance of the Corfu Runtime.
      * @param enableTxLogging
      */
     @Nonnull
@@ -250,9 +251,10 @@ public class CorfuStore {
      * @param streamListener   callback context
      * @param namespace        the CorfuStore namespace to subscribe to
      * @param tablesOfInterest only updates from these tables of interest will be sent to listener
-     * @param timestamp        if specified, all stream updates from this timestamp will be returned
+     * @param timestamp        if specified, all stream updates after this timestamp will be returned
      *                         if null, only future updates will be returned
      */
+    @Deprecated
     public <K extends Message, V extends Message, M extends Message>
     void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
                    @Nonnull List<TableSchema<K, V, M>> tablesOfInterest,
@@ -265,20 +267,43 @@ public class CorfuStore {
     /**
      * Subscribe to transaction updates on specific tables with the streamTag in the namespace.
      * Objects returned will honor transactional boundaries.
+     * <p>
+     * Note: if memory is a consideration consider use the other version of subscribe that is
+     * able to specify the size of buffered transactions entries.
      *
      * @param streamListener   callback context
      * @param namespace        the CorfuStore namespace to subscribe to
      * @param streamTag        only updates of tables with the stream tag will be polled
      * @param tablesOfInterest only updates from these tables of interest will be sent to listener
-     * @param timestamp        if specified, all stream updates from this timestamp will be returned
+     * @param timestamp        if specified, all stream updates from this timestamp will be returned,
      *                         if null, only future updates will be returned
      */
-    public void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
-                   @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest,
-                   @Nullable Timestamp timestamp) {
-        runtime.getTableRegistry().getStreamManager()
+    public void subscribeListener(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                                  @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest,
+                                  @Nullable Timestamp timestamp) {
+        runtime.getTableRegistry().getStreamingManager()
                 .subscribe(streamListener, namespace, streamTag, tablesOfInterest,
                         (timestamp == null) ? getTimestamp().getSequence() : timestamp.getSequence());
+    }
+
+    /**
+     * Subscribe to transaction updates on specific tables with the streamTag in the namespace.
+     * Objects returned will honor transactional boundaries.
+     *
+     * @param streamListener   client listener for callback
+     * @param namespace        the CorfuStore namespace to subscribe to
+     * @param streamTag        only updates of tables with the stream tag will be polled
+     * @param tablesOfInterest only updates from these tables of interest will be sent to listener
+     * @param timestamp        if specified, all stream updates after this timestamp will be returned,
+     *                         if null, only future updates will be returned
+     * @param bufferSize       maximum size of buffered transaction entries
+     */
+    public void subscribeListener(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                                  @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest,
+                                  @Nullable Timestamp timestamp, int bufferSize) {
+        runtime.getTableRegistry().getStreamingManager()
+                .subscribe(streamListener, namespace, streamTag, tablesOfInterest,
+                        (timestamp == null) ? getTimestamp().getSequence() : timestamp.getSequence(), bufferSize);
     }
 
     /**
@@ -287,8 +312,18 @@ public class CorfuStore {
      *
      * @param streamListener - callback context.
      */
+    @Deprecated
     public void unsubscribe(@Nonnull StreamListener streamListener) {
-        runtime.getTableRegistry().getStreamManager()
-                .unsubscribe(streamListener);
+        runtime.getTableRegistry().getStreamManager().unsubscribe(streamListener);
+    }
+
+    /**
+     * Gracefully shutdown a streamer.
+     * Once this call returns no further stream updates will be returned.
+     *
+     * @param streamListener - callback context.
+     */
+    public void unsubscribeListener(@Nonnull StreamListener streamListener) {
+        runtime.getTableRegistry().getStreamingManager().unsubscribe(streamListener);
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
@@ -1,26 +1,25 @@
 package org.corfudb.runtime.collections;
 
 import com.google.protobuf.Message;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-
 import org.corfudb.protocols.logprotocol.SMREntry;
 
+import javax.annotation.Nonnull;
+
 /**
- * Entry returned by CorfuStore's StreamListener interface
+ * Entry returned by CorfuStore's StreamListener interface.
+ * NOTE: Ensure that the below protobuf generated classes are accessible to your JVM!
+ * <p>
+ * Created by hisundar on 2019-10-18
  *
  * @param <K> - type of the protobuf KeySchema defined while the table was created.
  * @param <V> - type of the protobuf PayloadSchema defined by the table creator.
  * @param <M> - type of the protobuf metadata schema defined by table creation.
- *
- *  NOTE: Ensure that the above protobuf generated classes are accessible to your JVM!
- *
- *  Created by hisundar on 2019-10-18
  */
 @Slf4j
+@EqualsAndHashCode
 public class CorfuStreamEntry<K extends Message, V extends Message, M extends Message> {
     /**
      * Key of the UFO stream entry
@@ -46,16 +45,18 @@ public class CorfuStreamEntry<K extends Message, V extends Message, M extends Me
     public enum OperationType {
         UPDATE,
         DELETE,
-        CLEAR;
-    };
+        CLEAR
+    }
 
     @Getter
     private final OperationType operation;
+
     /**
      * Version number of the layout at the time of this entry.
      */
     @Getter
     private final long epoch;
+
     /**
      * Stream address of this entry
      */
@@ -75,22 +76,12 @@ public class CorfuStreamEntry<K extends Message, V extends Message, M extends Me
      * Convert a given SMREntry to CorfuStreamEntry.
      */
     public static <K extends Message, V extends Message, M extends Message>
-        CorfuStreamEntry<K, V, M> fromSMREntry(SMREntry entry, @Nonnull final long epoch,
-                                               @Nonnull final Class<K> keyClass,
-                                               @Nonnull final Class<V> payloadClass,
-                                               @Nullable final Class<M> metadataClass) {
-
+    CorfuStreamEntry<K, V, M> fromSMREntry(SMREntry entry, final long epoch) {
         long address = entry.getGlobalAddress();
 
         OperationType operationType = getOperationType(entry);
-        log.trace("fromSRMEntry: Table {} streamer got SMR {}", keyClass.getName(), entry.getSMRMethod());
-        // TODO[sneginhal]: Need a way to differentiate between update and create.
+        // TODO(sneginhal): Need a way to differentiate between update and create.
         Object[] args = entry.getSMRArguments();
-
-        log.trace("Converting SMREntry at address {} to CorfuStreamEntry: Operation: {} Length of arguments: {}. " +
-                "({} -> {}, {})",
-                address, operationType, args.length, keyClass.getCanonicalName(), payloadClass.getCanonicalName(),
-                (metadataClass != null) ? metadataClass.getCanonicalName() : "null");
 
         K key = null;
         V payload = null;
@@ -104,7 +95,7 @@ public class CorfuStreamEntry<K extends Message, V extends Message, M extends Me
             }
         }
 
-        return new CorfuStreamEntry<K, V, M>(key, payload, metadata, epoch, address, operationType);
+        return new CorfuStreamEntry<>(key, payload, metadata, epoch, address, operationType);
     }
 
     private static OperationType getOperationType(@Nonnull SMREntry entry) {
@@ -121,36 +112,10 @@ public class CorfuStreamEntry<K extends Message, V extends Message, M extends Me
                 operationType = OperationType.DELETE;
                 break;
             default:
-                throw new RuntimeException("SMRMethod "+entry.getSMRMethod()
+                throw new RuntimeException("SMRMethod " + entry.getSMRMethod()
                         + " cannot be translated to any known operation type");
         }
+
         return operationType;
-    }
-
-    /**
-     *  Convert an SMREntry into a CorfuStreamEntry when the SMR arguments have clear types
-     *  (like if called from a commit callback)
-     * @param entry - the entry to convert from.
-     * @param epoch - the epoch this entry was written to/will be written to.
-     */
-    public static
-    CorfuStreamEntry fromSMREntry(@Nonnull SMREntry entry, final long epoch) {
-
-        long address = entry.getGlobalAddress();
-
-        OperationType operationType = getOperationType(entry);
-        Object[] args = entry.getSMRArguments();
-        if (args.length > 0) {
-            if (args.length > 1) {
-                CorfuRecord record = (CorfuRecord) args[1];
-                return new CorfuStreamEntry((Message)args[0],
-                        record.getPayload(), record.getMetadata(), 0, address, operationType);
-            }
-            // this would likely be the case with "remove" SMR method
-            return new CorfuStreamEntry((Message)args[0], null, null, 0, address, operationType);
-        }
-
-        // this would be the case for the "clear" SMR method
-        return new CorfuStreamEntry(null, null, null, 0, address, operationType);
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
@@ -2,12 +2,21 @@ package org.corfudb.runtime.collections;
 
 import com.google.common.collect.Sets;
 import com.google.protobuf.Message;
-
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.logprotocol.MultiObjectSMREntry;
+import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.StreamSubscriptionException;
+import org.corfudb.runtime.view.Address;
+import org.corfudb.runtime.view.ObjectsView;
+import org.corfudb.runtime.view.StreamOptions;
+import org.corfudb.runtime.view.TableRegistry;
+import org.corfudb.runtime.view.stream.IStreamView;
 
+import javax.annotation.Nonnull;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -20,18 +29,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
-import javax.annotation.Nonnull;
-
-import org.corfudb.protocols.logprotocol.MultiObjectSMREntry;
-import org.corfudb.protocols.wireprotocol.ILogData;
-import org.corfudb.runtime.CorfuRuntime;
-import org.corfudb.runtime.exceptions.StreamSubscriptionException;
-import org.corfudb.runtime.view.Address;
-import org.corfudb.runtime.view.ObjectsView;
-import org.corfudb.runtime.view.StreamOptions;
-import org.corfudb.runtime.view.TableRegistry;
-import org.corfudb.runtime.view.stream.IStreamView;
-
 /**
  * A simple thread based subscription engine where each subscriber or listener gets
  * a thread which will listen on the tables of interest to it and moves at a rate
@@ -40,6 +37,7 @@ import org.corfudb.runtime.view.stream.IStreamView;
  * Created by hisundar on 04/28/2020.
  */
 @Slf4j
+@Deprecated
 public class StreamManager {
 
     private static final int DEFAULT_NUM_SUBSCRIBERS = 6; // Max number of threads.
@@ -93,11 +91,11 @@ public class StreamManager {
      * @param streamListener   client listener
      * @param namespace        namespace of interest
      * @param tablesOfInterest only updates from these tables will be returned
-     * @param startAddress     address to start the notifications from
+     * @param lastAddress      last processed address, new notifications start from lastAddress + 1
      */
     synchronized <K extends Message, V extends Message, M extends Message>
     void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
-                   @Nonnull List<TableSchema<K, V, M>> tablesOfInterest, long startAddress) {
+                   @Nonnull List<TableSchema<K, V, M>> tablesOfInterest, long lastAddress) {
         if (subscriptions.containsKey(streamListener)) {
             // Multiple subscribers subscribing to same namespace and table is allowed.
             // Which is why we use the caller's hashcode() and equals() are used to validate re-subscription.
@@ -119,10 +117,10 @@ public class StreamManager {
         tablesOfInterest.forEach(t -> {
             log.info("table name = {} ID {}", t.getTableName(), CorfuRuntime.getStreamID(t.getTableName()));
         });
-        log.info("StreamManager::subscribe {}, startAddress {}, namespace {}, tables {}",
-                streamListener, startAddress, namespace, tablesOfInterest.toString());
+        log.info("StreamManager::subscribe {}, lastAddress {}, namespace {}, tables {}",
+                streamListener, lastAddress, namespace, tablesOfInterest.toString());
         SubscriberTask task = new SubscriberTask(this,
-                streamListener, namespace, tablesOfInterest, startAddress);
+                streamListener, namespace, tablesOfInterest, lastAddress);
         final ScheduledFuture<?> scheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(
                 task, 0, DEFAULT_SLEEP_MILLIS, TimeUnit.MILLISECONDS);
         subscriptions.put(streamListener, new StreamSubscriber(task, scheduledFuture));
@@ -135,14 +133,14 @@ public class StreamManager {
      * @param namespace        namespace of interest
      * @param streamTag        only updates of tables with the stream tag will be polled
      * @param tablesOfInterest only updates from these tables will be returned
-     * @param startAddress     address to start the notifications from
+     * @param lastAddress      last processed address, new notifications start from lastAddress + 1
      * @throws NoSuchElementException   if any table of interest is never registered
      * @throws IllegalArgumentException if any table of interest is not opened before subscription
      */
     synchronized <K extends Message, V extends Message, M extends Message>
     void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
                    @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest,
-                   long startAddress) {
+                   long lastAddress) {
         TableRegistry registry = runtime.getTableRegistry();
         List<TableSchema<K, V, M>> tableSchemas = tablesOfInterest
                 .stream()
@@ -154,7 +152,7 @@ public class StreamManager {
                 })
                 .collect(Collectors.toList());
 
-        subscribe(streamListener, namespace, tableSchemas, startAddress);
+        subscribe(streamListener, namespace, tableSchemas, lastAddress);
     }
 
     /**
@@ -229,7 +227,7 @@ public class StreamManager {
                        @Nonnull StreamListener listener,
                        @Nonnull String namespace,
                        @Nonnull List<TableSchema<K, V, M>> tableSchemas,
-                       long startAddress) {
+                       long lastAddress) {
             this.streamManager = streamManager;
             this.listener = listener;
             this.namespace = namespace;
@@ -247,7 +245,7 @@ public class StreamManager {
 
             this.txnStream = runtime.getStreamsView()
                     .get(ObjectsView.TRANSACTION_STREAM_ID, options);
-            this.txnStream.seek(startAddress + 1);
+            this.txnStream.seek(lastAddress + 1);
         }
 
         public void run() {
@@ -270,11 +268,7 @@ public class StreamManager {
                             .forEach(streamId -> entries.put(tablesOfInterest.get(streamId),
                                     // Only extract the list of updates per stream as a list
                                     multiObjSMREntry.getSMRUpdates(streamId).stream().map(smrEntry ->
-                                            CorfuStreamEntry.fromSMREntry(smrEntry,
-                                                    epoch,
-                                                    tablesOfInterest.get(streamId).getKeyClass(),
-                                                    tablesOfInterest.get(streamId).getPayloadClass(),
-                                                    tablesOfInterest.get(streamId).getMetadataClass())
+                                            CorfuStreamEntry.fromSMREntry(smrEntry, epoch)
                                     ).collect(Collectors.toList())));
 
                     if (!entries.isEmpty()) {
--- a/runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
@@ -1,0 +1,98 @@
+package org.corfudb.runtime.collections;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.time.Duration;
+import java.util.concurrent.ExecutorService;
+
+/**
+ * A runnable task for a subscription to retrieve the previously
+ * polled transaction updates from the buffer and send notifications
+ * to client via the registered call backs. This task is executed
+ * by the thread pool continuously until an error occurs which would
+ * stop the subscription.
+ * <p>
+ * Created by WenbinZhu on 11/9/20.
+ */
+@Slf4j
+class StreamNotificationTask implements Runnable {
+
+    // Number of transaction updates to send notification in each run.
+    private static final int NOTIFICATION_BATCH_SIZE = 10;
+
+    // Total amount of time to wait for retrieving the data changes from buffer if it is empty.
+    private static final Duration QUEUE_EMPTY_BLOCK_TIME_MS = Duration.ofMillis(1_000);
+
+    // A warning will be raised if client call back takes longer than this time threshold.
+    private static final Duration SLOW_NOTIFICATION_TIME_MS = Duration.ofMillis(1_500);
+
+    // The streaming manager that is in charge of listener subscriptions.
+    private final StreamingManager streamingManager;
+
+    // The subscription context associated with this task.
+    private final StreamSubscription subscription;
+
+    // The Thread pool for executing client notification tasks.
+    private final ExecutorService notificationExecutor;
+
+    StreamNotificationTask(StreamingManager streamingManager,
+                           StreamSubscription subscription,
+                           ExecutorService notificationExecutor) {
+        this.streamingManager = streamingManager;
+        this.subscription = subscription;
+        this.notificationExecutor = notificationExecutor;
+    }
+
+    @Override
+    public void run() {
+        try {
+            sendNotifications();
+        } catch (Throwable throwable) {
+            StreamListener listener = subscription.getListener();
+            log.error("Encountered exception {} during client notification callback, " +
+                    "listener: {}, namespace: {}", throwable, listener, subscription.getNamespace());
+            streamingManager.unsubscribe(listener, false);
+            listener.onError(throwable);
+        }
+    }
+
+    /**
+     * Retrieve the first data change from the buffer and send notification
+     * to the client via the pre-registered call back.
+     */
+    private void sendNotifications() throws Exception {
+        // Total amount of time for the batch waiting for buffer being not empty.
+        long remainingBlockTime = QUEUE_EMPTY_BLOCK_TIME_MS.toNanos();
+        StreamListener listener = subscription.getListener();
+
+        for (int iter = 0; iter < NOTIFICATION_BATCH_SIZE; iter++) {
+            // If listener already unsubscribed, do not process or schedule again.
+            if (subscription.isStopped()) {
+                return;
+            }
+
+            long startTime = System.nanoTime();
+            CorfuStreamEntries nextUpdate = subscription.dequeueStreamEntry(remainingBlockTime);
+
+            // No new updates after max waiting time elapses, break and re-schedule.
+            if (nextUpdate == null) {
+                break;
+            }
+
+            long endTime = System.nanoTime();
+            remainingBlockTime -= endTime - startTime;
+
+            // Send notification to client with the pre-registered callback.
+            startTime = endTime;
+            subscription.getStreamingMetrics().recordDeliveryDuration(() -> listener.onNext(nextUpdate));
+            endTime = System.nanoTime();
+
+            Duration onNextElapse = Duration.ofNanos(endTime - startTime);
+            if (onNextElapse.compareTo(SLOW_NOTIFICATION_TIME_MS) > 0) {
+                log.warn("Stream listener: {} onNext() took to long: {} ms.", listener, onNextElapse.toMillis());
+            }
+        }
+
+        notificationExecutor.submit(this);
+    }
+}
--- a/runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
@@ -1,0 +1,114 @@
+package org.corfudb.runtime.collections;
+
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.runtime.view.stream.IStreamView;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A runnable task for a subscription to poll data change from the
+ * transaction stream and put into the subscription's buffer. This
+ * task is executed by the thread pool continuously until an error
+ * occurs which would stop the subscription.
+ * <p>
+ * Created by WenbinZhu on 11/9/20.
+ */
+@Slf4j
+class StreamPollingTask implements Runnable {
+
+    // A period of time in ms to sleep before next cycle when poller gets no new data changes.
+    private static final long IDLE_WAIT_TIME_MS = 50;
+
+    // Total amount of time to wait for putting the polled data changes into buffer if it is full.
+    private static final Duration QUEUE_FULL_BLOCK_TIME_MS = Duration.ofMillis(1_000);
+
+    // The streaming manager that is in charge of listener subscriptions.
+    private final StreamingManager streamingManager;
+
+    // The corfu transaction stream to poll data changes from.
+    private final IStreamView txnStream;
+
+    // The subscription context associated with this task.
+    private final StreamSubscription subscription;
+
+    // The Thread pool for executing stream polling tasks.
+    private final ScheduledExecutorService pollingExecutor;
+
+    // Last address of the data successfully processed by the buffer.
+    private long lastReadAddress;
+
+    StreamPollingTask(StreamingManager streamingManager, long lastAddress,
+                      StreamSubscription subscription, ScheduledExecutorService executor) {
+        this.streamingManager = streamingManager;
+        this.subscription = subscription;
+        this.pollingExecutor = executor;
+        this.lastReadAddress = lastAddress;
+        this.txnStream = subscription.getTxnStream();
+    }
+
+    @Override
+    public void run() {
+        try {
+            pollTxnStream();
+        } catch (Throwable throwable) {
+            StreamListener listener = subscription.getListener();
+            log.error("Encountered exception {} during txn stream polling, listener: {}, " +
+                    "namespace: {}", throwable, listener, subscription.getNamespace());
+            streamingManager.unsubscribe(listener, false);
+            listener.onError(throwable);
+        }
+    }
+
+    /**
+     * Poll new data changes from the transaction stream and put into
+     * the subscription's buffer one by one.
+     */
+    private void pollTxnStream() throws Exception {
+        // If listener already unsubscribed, do not process or schedule again.
+        if (subscription.isStopped()) {
+            return;
+        }
+
+        // Seek to next address and poll transaction updates.
+        txnStream.seek(lastReadAddress + 1L);
+        List<ILogData> updates = subscription.getStreamingMetrics().recordPollingDuration(
+                () -> txnStream.remainingAtMost(subscription.getStreamBufferSize()));
+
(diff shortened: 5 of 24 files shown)


WenbinZhu:
More tests on the way.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -121,36 +112,10 @@ private static OperationType getOperationType(@Nonnull SMREntry entry) {
...
-
-        OperationType operationType = getOperationType(entry);
-        Object[] args = entry.getSMRArguments();
-        if (args.length > 0) {
-            if (args.length > 1) {
-                CorfuRecord record = (CorfuRecord) args[1];
-                return new CorfuStreamEntry((Message)args[0],
-                        record.getPayload(), record.getMetadata(), 0, address, operationType);
-            }
-            // this would likely be the case with "remove" SMR method
```
hisundar, line 149: Consolidating the functions is ok, but would be good to retain these comments.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    private void pollTxnStream() throws Exception {
+        // If listener already unsubscribed, do not process or schedule again.
+        if (subscription.isStopped()) {
+            return;
+        }
+
+        // Seek to next address and poll transaction updates.
+        txnStream.seek(lastReadAddress + 1L);
+        List<ILogData> updates = txnStream.remaining();
```
hisundar, line 78: use of remaining() is exactly what got us into the OOM scenario. Even with more selectivity, this will result in OOM if there is a lot of distance between the lastAddress and the tail.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+        if (updates.isEmpty()) {
+            TimeUnit.MILLISECONDS.sleep(IDLE_WAIT_TIME_MS.toMillis());
+            pollingExecutor.submit(this);
+            return;
+        }
+
+        // Insert polled updates to the subscription buffer, with a shared
+        // fixed amount of time waiting for buffer being not full.
+        long remainingBlockTime = QUEUE_FULL_BLOCK_TIME_MS.toNanos();
+        for (ILogData update : updates) {
```
hisundar, line 90: If the goal of using fixed size queue and retry was to avoid starvation of other streams when the number of subscribers exceeds the number of threads. Then won't this still bring back the starvation if the number of updates are large?

hisundar (changes requested):
I see some concerns on this implementation specifically with OOM and starvation.
We should probably sync up on this

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: Initially I tried to add a parameter in remaining() to limit the number of entries returned, but the current implementation makes it complicated and error-prone to make this change. However, with the try catch block and the tunable buffer size, when OOM happens it will be captured and user be able to handle it by re-process the initial snapshot and re-subscribe. Let's sync up this and other concern you might have on Monday.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: This is why the QUEUE_FULL_BLOCK_TIME_MS is being used. It is the total amount of time to wait, which is set to 2 seconds right now. It means it will wait for up to 2 seconds for all the updates polled in this cycle (only the wait time when inserting into queue). Also it's not typical that all call backs are slow, so as long as one thread is available, they won't be blocked.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
WenbinZhu: Will do.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
hisundar: OOM exceptions cannot always be caught in a try-catch block - the actual OOM error can happen in a stack trace not controlled by us.
Use of remaining() the way it is very buggy, if there are any infra issues then the stream address space can grow quickly which means remaining() will pull the entire stream history into memory.
Maybe this loop can be change to do a next() up to the NOTIFICATION_BATCH_SIZE and then stop?

WenbinZhu: In initial impl, OOM can happen in a stack trace not controlled by us because it has large buffer size, but now we can fine tine it. Using next() is too slow and not a generic approach. And if you remember even if we use next(), they still have OOM because their jvm is too small, that problem is not one our side, we don't want to do our impl for a specific case that slows down other common use cases right? The right solution is too enhance remaining()/remainingUpto() API so that it limits the number of entries returned, the current maxGlobal parameter in streaming API is not the number of entries returned for this stream, but the last GLOBAL address to look for, which might return 0 stream entries up to that address, so need another parameter to control the number of actual entries returned.

WenbinZhu:
@hisundar @Maithem Instead of using next() or remainingUpTo(), which is not efficient, I added a new API called remainingAtMost(), which retrieves at most N entries, while not changing the existing remaining()/remaingUpTo() API to avoid complication. 

@Everyone You can now start reviewing this patch as well as the part 2 patch #2820 (this patch is stacked on top of #2820).

hisundar (changes requested):
The changes mostly look good to me. One point though since this is really a performance improvement, shuouldnt this change add streaming metrics into the Table's metrics so it can be captured and analyzed.
The most useful metric is the time since transaction creation.
If the Metadata schema has last_modified_time, then we can capture the time since mutation as the difference between current time and last_modified_time.
Without metrics we really have no way to justify if the change is regressing anything or improving right?

WenbinZhu:
> The changes mostly look good to me. One point though since this is really a performance improvement, shuouldnt this change add streaming metrics into the Table's metrics so it can be captured and analyzed.
> The most useful metric is the time since transaction creation.
> If the Metadata schema has last_modified_time, then we can capture the time since mutation as the difference between current time and last_modified_time.
> Without metrics we really have no way to justify if the change is regressing anything or improving right?

last_modified_time is per entry, but the metrics is per table right? Also table metrics is in-memory per jvm, where last_modified_time time is stored on server, per entry.  It's not efficient to calculate these info right?

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,145 @@
...
+ * streamlined and clients can get notifications via the registered
+ * call backs.
+ * <p>
+ * Created by WenbinZhu on 11/5/20.
+ */
+@Slf4j
+public class StreamingManager {
+
+    // Number of thread in polling and notification pool.
+    private static final int NUM_THREAD_PER_POOL = 4;
```
medhavidhawan, line 28: How did we arrive at this number ?

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,145 @@
...
+
+    // Default buffer size for each subscription.
+    private static final int DEFAULT_BUFFER_SIZE = 50;
```
medhavidhawan, line 31: do we need parameterize some of these numbers in case we need to tweak them to find the right values.

medhavidhawan (commented):
Overall looks ok at first review. 
Some comments

WenbinZhu:
@hisundar Per our discussion, instead of using last_modified_time, I ceated a StreamingMetrics to capture the notification delivery duration metric. Please review.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+        // No new updates, take a short break and poll again.
+        if (updates.isEmpty()) {
+            TimeUnit.MILLISECONDS.sleep(IDLE_WAIT_TIME_MS.toMillis());
+            pollingExecutor.submit(this);
+            return;
+        }
+
+        // Insert polled updates to the subscription buffer, with a shared
+        // fixed amount of time waiting for buffer being not full.
+        long remainingBlockTime = QUEUE_FULL_BLOCK_TIME_MS.toNanos();
```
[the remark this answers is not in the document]
WenbinZhu: This is used when inserting into the queue: `subscription.enqueueStreamEntry(update, remainingBlockTime)`

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+            return;
+        }
+
+        // Seek to next address and poll transaction updates.
+        txnStream.seek(lastReadAddress + 1L);
+        List<ILogData> updates = txnStream.remainingAtMost(subscription.getStreamBufferSize());
+
+        // No new updates, take a short break and poll again.
+        if (updates.isEmpty()) {
+            TimeUnit.MILLISECONDS.sleep(IDLE_WAIT_TIME_MS.toMillis());
```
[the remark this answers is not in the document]
WenbinZhu: In normal cases, the updates are not empty. Here by sleeping a tiny period (50 ms), we prevent cpu busy polling in rare cases that no new entries were produced. The current master impl uses scheduling, but it has extras delays for normal cases and not scalable. The sleep here is actually same as current impl, but the new model solves the aforementioned two problems.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
WenbinZhu: This is based on an estimation of subscribers and and limits of threads.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
WenbinZhu: Actually the buffer size is already configurable. User can specify the buffer size on subscribe, but if it's not provided, this default value will be used.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    /**
+     * Poll new data changes from the transaction stream and put into
+     * the subscription's buffer one by one.
+     */
+    private void pollTxnStream() throws Exception {
+        // If listener already unsubscribed, do not process or schedule again.
+        if (subscription.isStopped()) {
```
pankti-m, line 72: log a message?

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+            TimeUnit.MILLISECONDS.sleep(IDLE_WAIT_TIME_MS.toMillis());
+            pollingExecutor.submit(this);
+            return;
+        }
+
+        // Insert polled updates to the subscription buffer, with a shared
+        // fixed amount of time waiting for buffer being not full.
+        long remainingBlockTime = QUEUE_FULL_BLOCK_TIME_MS.toNanos();
+        for (ILogData update : updates) {
+            if (subscription.isStopped()) {
```
pankti-m, line 91: already checked this on L72

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStreamEntry.java
```diff
@@ -1,26 +1,25 @@
...
 
 import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-
-import org.corfudb.protocols.logprotocol.SMREntry;
 
 /**
- * Entry returned by CorfuStore's StreamListener interface
+ * Entry returned by CorfuStore's StreamListener interface.
+ * NOTE: Ensure that the above protobuf generated classes are accessible to your JVM!
```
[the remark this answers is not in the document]
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -1,12 +1,21 @@
...
+import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.StreamSubscriptionException;
+import org.corfudb.runtime.view.Address;
+import org.corfudb.runtime.view.ObjectsView;
+import org.corfudb.runtime.view.StreamOptions;
+import org.corfudb.runtime.view.TableRegistry;
+import org.corfudb.runtime.view.stream.IStreamView;
 
+import javax.annotation.Nonnull;
```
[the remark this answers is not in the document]
WenbinZhu: These are automatically ordered by IDEA, I hit the optimize hot key and it's still in this order, I didn't change the default ordering config, and I think we can just observe the default ordering.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
```diff
@@ -0,0 +1,100 @@
+package org.corfudb.runtime.collections;
+
+import lombok.extern.slf4j.Slf4j;
```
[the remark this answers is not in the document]
WenbinZhu: Same reason as above

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
```diff
@@ -0,0 +1,100 @@
...
+     * to the client via the pre-registered call back.
+     */
+    private void sendNotifications() throws Exception {
+        // Total amount of time for the batch waiting for buffer being not empty.
+        long remainingBlockTime = QUEUE_EMPTY_BLOCK_TIME_MS.toNanos();
+        StreamListener listener = subscription.getListener();
+
+        for (int iter = 0; iter < NOTIFICATION_BATCH_SIZE; iter++) {
+            // If listener already unsubscribed, do not process or schedule again.
+            if (subscription.isStopped()) {
```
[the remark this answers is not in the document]
WenbinZhu: This could happen because of a timing race, but this race does not matter a lot, so we don't need to log here.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: Same reason as above.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: Yea, just check again here because we synchronize it, so the stop reqeust could happen after the previous check.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
+package org.corfudb.runtime.collections;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.exceptions.StreamSubscriptionException;
+
+import javax.annotation.Nonnull;
```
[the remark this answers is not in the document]
WenbinZhu: Same reason as above.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
...
+        StreamingMetrics metrics = new StreamingMetrics(
+                streamListener, namespace, streamTag, runtime.getParameters().getMetricRegistry());
+        StreamSubscription subscription = new StreamSubscription(
+                runtime, streamListener, namespace, streamTag, tablesOfInterest, bufferSize, metrics);
+        subscriptions.put(streamListener, subscription);
+
+        pollingExecutor.submit(new StreamPollingTask(this, lastAddress, subscription, pollingExecutor));
+        notificationExecutor.submit(new StreamNotificationTask(this, subscription, notificationExecutor));
+
+        log.info("Subscribed stream listener {}, numSubscribers: {}, streamTag: {}, lastAddress: {}, " +
```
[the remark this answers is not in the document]
WenbinZhu: Currently if user overloads toString(), then we have a meaningful identifier, and that's what we currently do in testing. I'm trying to avoid changing interfaces in this patch, but sure later we could change the interface to have better debuggability.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Unsubscribe a prior subscription.
+     *
+     * @param streamListener            client listener to unsubscribe
+     * @param warnIfNotSubscribedBefore whether to log warning if listener is not
+     *                                  subscribed or already unsubscribed
+     */
+    synchronized void unsubscribe(@Nonnull StreamListener streamListener,
```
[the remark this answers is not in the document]
WenbinZhu: No, this is also used in StreamNotificationTask and StreamPollingTask.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
...
+     * Subscribe to transaction updates.
+     *
+     * @param streamListener   client listener for callback
+     * @param namespace        namespace of interested tables
+     * @param streamTag        only updates of tables with the stream tag will be polled
+     * @param tablesOfInterest only updates from these tables will be returned
+     * @param lastAddress      last processed address, new notifications start from lastAddress + 1
+     * @param bufferSize       maximum size of buffered transaction entries
+     */
+    synchronized void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
```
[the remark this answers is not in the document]
WenbinZhu: No, this is also used in CorfuStore.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Create the stream manager, initialize the tasks pools.
+     *
+     * @param runtime Corfu runtime to use for streaming
+     */
+    public StreamingManager(@Nonnull CorfuRuntime runtime) {
+        this.runtime = runtime;
+        this.subscriptions = new HashMap<>();
+
+        this.pollingExecutor = Executors.newFixedThreadPool(NUM_THREAD_PER_POOL,
```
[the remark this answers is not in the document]
WenbinZhu: This is actually on client side. However this is a good suggestion, we do need to shut down the executors even on client  side, but the problem is there is no shutdown functionality in CorfuStore, so I don't know when to shut down the executors. Once we decide to add shutdown to CorfuStore, I can also shutdown the executors here.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -42,6 +32,19 @@
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.lang.reflect.InvocationTargetException;
```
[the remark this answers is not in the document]
WenbinZhu: Same reason as above.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -461,6 +486,21 @@ public synchronized StreamManager getStreamManager() {
...
     }
 
+    /**
+     * Register a streaming subscription manager as a singleton.
+     */
+    public StreamingManager getStreamingManager() {
+        if (streamingManager == null) {
+            synchronized (StreamingManager.class) {
+                if (streamingManager == null) {
+                    streamingManager = new StreamingManager(runtime);
```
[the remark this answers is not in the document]
WenbinZhu: This is lazy initialization, for clients that do not need streaming functionality, we don't need to create StreamingManager.

WenbinZhu:
@pankti-m I addressed your comments, thanks.

pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -274,21 +274,40 @@ void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace
      *                         if null, only future updates will be returned
      */
     public void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace,
-                   @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest,
-                   @Nullable Timestamp timestamp) {
-        runtime.getTableRegistry().getStreamManager()
+                          @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest,
```
Maithem, line 277: Deprecate the `subscribe` method that uses the old stream manager too ?

https://github.com/CorfuDB/CorfuDB/pull/2826/files#diff-f44ea07bfcd37e9243f12f8e2e3666056f2a25d4943965e359e61a43d4e64135R257

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -274,21 +274,40 @@ void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace
...
     /**
      * Gracefully shutdown a streamer.
      * Once this call returns no further stream updates will be returned.
      *
      * @param streamListener - callback context.
      */
     public void unsubscribe(@Nonnull StreamListener streamListener) {
-        runtime.getTableRegistry().getStreamManager()
-                .unsubscribe(streamListener);
+        runtime.getTableRegistry().getStreamingManager().unsubscribe(streamListener);
```
Maithem, line 311: This looks like a leak. You are allowing subscribe to be called from `StreamManager` and `StreamingManager` but `unsubscribe` only invokes `StreamingManager::unsubscribe`

Maithem: There is not a lot of dependencies on `StreamManager`, I would suggest removing it instead of deprecating it. Its cleaner than just assuming that API won't be called.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
WenbinZhu: Done, thanks for finding out.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
WenbinZhu: Done, thanks for finding out, I renamed the new subscribe/unscribe to subscribeListener/unsubscribeListener, which is clearer name and deperated the old unscribe API. 

The reason is we don't want to break the interface, some verticals are using the old subscribe api, and to convert to new API we need them to clean up their code, which is better to be in a separate effort to unblock merging process.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -273,22 +277,52 @@ void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace
...
+     *
+     * @param streamListener   client listener for callback
+     * @param namespace        the CorfuStore namespace to subscribe to
+     * @param streamTag        only updates of tables with the stream tag will be polled
+     * @param tablesOfInterest only updates from these tables of interest will be sent to listener
+     * @param timestamp        if specified, all stream updates after this timestamp will be returned
+     *                         if null, only future updates will be returned
+     * @param bufferSize       maximum size of buffered transaction entries
+     */
+    public void subscribeListener(@Nonnull StreamListener streamListener, @Nonnull String namespace,
```
Maithem, line 300: Do you need to expose this to the shim `CorfuStoreShim` ?

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * A streaming subscription manager that allows clients to listen on
+ * the transaction updates of interested tables. The updates will be
+ * streamlined and clients can get notifications via the registered
+ * call backs.
+ * <p>
+ * Created by WenbinZhu on 11/5/20.
+ */
+@Slf4j
+public class StreamingManager {
```
Maithem, line 25: How do you stop the `StreamingManager` ?

This class is creating non-daemon threads and ignoring interrupts. A shutdown/close method should be added an invoked from `CorfuStore`.

Also, why is `StreamManager::shutdown` invoked from `TableRegistry`, it seems very strange.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
...
+
+    // Number of thread in polling and notification pool.
+    private static final int NUM_THREAD_PER_POOL = 4;
+
+    // Default buffer size for each subscription.
+    private static final int DEFAULT_BUFFER_SIZE = 50;
```
Maithem, line 31: These have to be configurable and exposed. We have already seen bugs where we need to be able to tune a bunch of parameters. Unfortunately, most of our configurations are not configurable and require source code changes.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -480,6 +505,21 @@ public synchronized StreamManager getStreamManager() {
         return this.streamManager;
     }
 
+    /**
+     * Register a streaming subscription manager as a singleton.
+     */
+    public StreamingManager getStreamingManager() {
```
Maithem, line 511: You can improve the read performance by introducing a local variable. For example, 

```java
public StreamingManager getStreamingManager() {
        StreamingManager localRef = streamingManager;
        if (localRef == null) {
            synchronized (this) {
                localRef = streamingManager;
                if (localRef == null) {
                    localRef = streamingManager = new StreamingManager(runtime);
                }
            }
        }
        return localRef;
    }
```

Maithem: Alternately, you can just use Lomboks lazy annotation and it would emit the code for double checked locking.

```java
    @Getter(lazy = true)
    private final StreamingManager streamingManager = create();
    private StreamingManager create() {
        return new StreamingManager(runtime);
    }
```

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -393,19 +419,18 @@ public static String getFullyQualifiedTableName(TableName tableName) {
...
-                + " this existing table " + tableName + " in namespace " + namespace);
+                        + " this existing table " + tableName + " in namespace " + namespace);
             } else {
-                // If the table is completely unheard of return NoSuchElementException
-                throw new NoSuchElementException(
-                        String.format("No such table found: namespace: %s, tableName: %s",
-                        namespace, tableName));
+                // If the table is completely unheard of return NoSuchElementException.
+                throw new NoSuchElementException(String.format(
+                        "No such table found: namespace: %s, tableName: %s", namespace, tableName));
```
Maithem, line 433: Not related to this PR per-se but should these types be validated when `registryTable` is populated ? So `TableRegistry` should try to load the types with the class loader and should fail with ClassNotFoundException if the type is not there, which is more appropriate than NoSuchElementException

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -383,7 +382,34 @@ private boolean isTrimCoveredByCheckpoint(long trimMark) {
...
+            return streamAddressSpace.getHighestAddress();
+        }
+
+        LongIterator it = streamAddressSpace.getAddressMap().getLongIterator();
+        while (--maxEntries > 0) {
+            it.next();
+        }
+
+        return it.next();
     }
```
Maithem, line 413: I don't think its a good idea to spin just to get the largest address. I would suggest removing the loop and doing something like:

```java
return streamAddressSpace.getAddressMap().getReverseLongIterator().next();
```

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -383,7 +382,34 @@ private boolean isTrimCoveredByCheckpoint(long trimMark) {
...
+        if (size <= maxEntries) {
+            return streamAddressSpace.getHighestAddress();
+        }
+
+        LongIterator it = streamAddressSpace.getAddressMap().getLongIterator();
+        while (--maxEntries > 0) {
+            it.next();
+        }
+
+        return it.next();
```
Maithem, line 412: You can simplyf this logic to 

```java
        long lastAddress = 0;
        while (iterator.hasNext() && --maxEntries > 0) {
            lastAddress = iterator.next();
        }

        return lastAddress;
```

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
```diff
@@ -397,6 +401,34 @@ private void processTrimmedException(TrimmedException te) {
...
+        // of entries returned instead of calling it.
+        if (maxEntries < 1) {
+            throw new IllegalArgumentException(
+                    "remainingAtMost(): maxEntries cannot be less than 1.");
+        }
+
+        long maxGlobal = getMaxGlobalFromMaxEntries(maxEntries);
+        if (Address.nonAddress(maxGlobal)) {
+            return Collections.emptyList();
+        }
```
Maithem, line 419: This will emit all of of unnecessary sequencer queries

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -293,6 +302,16 @@ public static String getFullyQualifiedTableName(TableName tableName) {
...
 
+    /**
+     * Return the stream Id for the provided stream tag.
+     *
+     * @param streamTag stream tag in string
+     * @return stream Id in UUID
+     */
+    public static UUID getStreamIdForStreamTag(String streamTag) {
+        return CorfuRuntime.getStreamID(STREAM_TAG_PREFIX + streamTag);
+    }
```
Maithem, line 313: Shouldn't this be scoped to a namespace (i.e., STREAM_TAG_PREFIX + nameSpace + streamTag) ?

There could be an issue were subscribers can read data from a different namespace.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
```diff
@@ -178,5 +185,10 @@ protected boolean discoverAddressSpace(final UUID streamId,
 
         return !queue.isEmpty();
     }
+
+    @Override
+    protected long getMaxGlobalFromMaxEntries(int maxEntries) {
+        return Address.MAX;
+    }
```
Maithem, line 192: Instead of returning an incorrect value here and relying on remainingAtMost to throw an exception I think its cleaner to just move UnsupportedOperationException to a default implementation in the IStreamView interface.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMetrics.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * List of metrics captured for a stream listener.
+ * <p>
+ * Created by WenbinZhu on 11/23/20.
+ */
+public class StreamingMetrics {
+
+    private final String listenerId;
+
+    private final TimingHistogram deliveryDuration;
```
Maithem, line 15: Is this class only required for logging purposes? If the these stats can't be accessed externally and only used for logging, then I would suggest converting it to a proper Metric so that we can collect it and log it as within our metrics performance library. 

This metric can be kept registered and kept in the `StreamSubscription` instance.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMetrics.java
```diff
@@ -0,0 +1,34 @@
...
+
+    StreamingMetrics(StreamListener listener, String namespace, String streamTag, MetricRegistry registry) {
+        this.listenerId = String.format("StreamListener_%s_%s_%s", listener, namespace, streamTag);
+        this.deliveryDuration = new TimingHistogram(registry.histogram(listenerId + "_deliveryDuration"));
+    }
+
+    public void setDeliveryDuration(long elapsedTime) {
+        deliveryDuration.update(elapsedTime);
+    }
```
Maithem, line 24: You can just nuke this and use a histogram from Micrometer. I also think that `record` is a better verb than `set` here.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMetrics.java
Maithem: Also, note that `TimingHistogram` uses a EDR histogram (histogram) underneath which is not really accurate when reporting these stats in a short interval.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMetrics.java
```diff
@@ -0,0 +1,34 @@
...
+ * Created by WenbinZhu on 11/23/20.
+ */
+public class StreamingMetrics {
+
+    private final String listenerId;
+
+    private final TimingHistogram deliveryDuration;
+
+    StreamingMetrics(StreamListener listener, String namespace, String streamTag, MetricRegistry registry) {
+        this.listenerId = String.format("StreamListener_%s_%s_%s", listener, namespace, streamTag);
```
Maithem, line 18: This won't produce any metrics. We moved away from using dropwizard to micrometer, make sure the correct registry is passed.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
Maithem, line 82: This is not the only way to prevent busy polling and can cause high latencies.

If you have 1 thread running the notification tasks and 8 subscriptions and only one that is producing data then, the listener of that active stream will take 7 * 50 milliseconds to deliver the notification. You can try to hide this by increasing the number of threads in the notification task, but then the pathological case can be really bad and it doesn't seem like its unlikely, or rather, its not clear that at most a 50ms penalty will be inured. 

Depending on the producer rate it seems really hard to reason about the worst case. It's better to just use a scheduled executor and push that sleep to the executor scheduling instead of doing it before the submit.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            // Buffer is full after max waiting time elapses, break and re-schedule.
+            long startTime = System.nanoTime();
+            if (!subscription.enqueueStreamEntry(update, remainingBlockTime)) {
+                break;
+            }
+
+            remainingBlockTime -= System.nanoTime() - startTime;
+            lastReadAddress = update.getGlobalAddress();
```
Maithem, line 102: Can you just put a sanity check here that the lastReadAddress is not regressing ?

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -0,0 +1,108 @@
...
+    private void pollTxnStream() throws Exception {
+        // If listener already unsubscribed, do not process or schedule again.
+        if (subscription.isStopped()) {
+            return;
+        }
+
+        // Seek to next address and poll transaction updates.
+        txnStream.seek(lastReadAddress + 1L);
+        List<ILogData> updates = txnStream.remainingAtMost(subscription.getStreamBufferSize());
+
```
Maithem, line 79: The logic can be simplified while also improving graceful degradation by actually trying to check if `StreamSubscription::streamBuffer` has enough space before issuing the read to the server. Otherwise if the system is in a degraded state and the buffers are not being consumed fast enough, then you will be issuing reads and computing all the logic in `StreamSubscription::enqueueStreamEntry`  and then returning false on `streamBuffer.offer`. 

When throttling is actually required its probably better assume the pessimistic case vs the optimistic case (i.e., current logic): where you hope the the queue clears up while you're doing your processing.

Maithem:
Final thogut

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
```diff
@@ -0,0 +1,147 @@
...
+    private final CorfuRuntime runtime;
+
+    // A map of all stream listeners and their subscription contexts.
+    private final Map<StreamListener, StreamSubscription> subscriptions;
+
+    // Thread pool for executing stream polling tasks.
+    private final ExecutorService pollingExecutor;
+
+    // Thread pool for executing client call back tasks.
+    private final ExecutorService notificationExecutor;
```
Maithem, line 43: There are so many use cases where the client would want to control their own threading. I think we should just allow them to pass in an executor when calling subscribe. For example, they might need to prioritize some subscriptions over other subscriptions. 

Furthermore, it makes it harder to test. For example, the application can't write a deterministic unit test if they can't control the executors in StreamingManager. If the application is able to passs their own exeuctors, then they can deterministcly test call backs using something like  `MoreExecutors.newDirectExecutorService`.

Finally, if they own their threading then they have to manage the exeuctors lifecycle, which means we don't need to make a decision on their behalf on whether threads should be daemon or not, and its less code for us.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
WenbinZhu: Done, added shutdown logic.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
WenbinZhu: The buffer size is already exposed in the subscribe() API, clients can specify a customed buffer size, or not to use this default setting.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
WenbinZhu: Since this is not the hot path, I switched to use the previous way that leverages synchronizes the method itself, to make it simpler as we have another shutdown() method that needs to check if it's null or not.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
WenbinZhu: So the contract of CorfuStore when they designed it is that clients have to first explicitly register this table by calling openTable() and later other clients use getTable() to use the opened table.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
WenbinZhu: The first block is an optimization to avoid iteration if maxEntries is greater than the size of the bitmap, so we can directly return the highest address.

runtime/src/main/java/org/corfudb/runtime/view/stream/AbstractQueuedStreamView.java
WenbinZhu: I think our maxResolution is broken as of now.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
WenbinZhu: Done, initially I expect streamTags to be unique, but yes having a namespace is better.

runtime/src/main/java/org/corfudb/runtime/view/stream/BackpointerStreamView.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: Done, added at the place when lastReadAddress is assigned.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: I think the calculation is not right here. It's won't wait for 7 * 50 ms. As long as one thread is available, it can pick up the next task. If you have 7 subscriptions which are not producing data, and you have 4 threads in total, the latency of the  next subscription is just 50 ms, because all threads wait in parallel, not sequentially. And if you have 8 threads, there will be no latecy at all. Actually I chose this model because the schduled exector has an intrinsic latency, which is the scheduling interval. This interval causes latecy on every poll, no matter there is data or not, while this model has no latency when there are data produced, which is the common case. Sundar tried to reduce the schduled exector's latecy by using an infinite loop in the task, but that cause scalability issues as mentioned in my Conflucence deisign doc. So I intentionally abandoned the scheduled executor when making the design decision.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: StreamSubscription::enqueueStreamEntry actually waits for a short period for the notification thread to consume the queue. This approach is trying to balance between differnce scenarios and consumption speed.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
WenbinZhu: I think the reason that we don't want users to pass the executor is that the executor is shared by all subscribers. Usually differnent subscribers are from different verticals, the executor serves them all, so it's not so appropriate for a vertical to specify an executor when subscribing. If any subscriber can specify any an executor, the resource usage can be a problem and we need to maintain a lot of exectutors from different subscribers which is kind of messy.

Maithem: Yes, that make sense, but the centralization shouldn't be hidden inside the library. The entity (i.e., application code) that is creating the StreamingManager should manage that.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMetrics.java
WenbinZhu: Done, switched to micrometer.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMetrics.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingMetrics.java
WenbinZhu: Done.

runtime/src/main/java/org/corfudb/runtime/collections/StreamingManager.java
WenbinZhu: The entity that is creating CorfuStore is usually different from the veticals that are using it, so it's no better than we doing it. Also if we change that it will break a lot of application code, maybe later we can discuss with verticals to see how they are using it.

Maithem: That is my point. There needs to be a centralization point that figures out configuration before exposing it to verticals. That centralization point of figuring configuration shouldn't be managed by the library itself, but in upper layers. 

I guess there is two points here 1) what's a better design approach 2) what are the implications of this change (i.e., breaking existing code). I don't think that its a good idea to forgo #1 because of #2.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
Maithem: The max blocking time for a stream that has unread data is a function of the number of threads and number of streams.

If you have 1 threads running the notification task and 7 subscribers with only 1 active stream then the executor queue will look like this | Active | Inactive | Inactive | Inactive | Inactive | Inactive | Inactive (processing thread). If the queue will be consumed from right to left, then the processing thread will sleep for 6 * 50ms before the active thread is discovered. If you use a scheduled executes then the tasks will incur the sleep while waiting in the min-heap of the executor, notice that cost will not be incurred on the processing thread. This means that it can pick up active streams that can be read right away. 

The fundamental difference is do you incur the cost of the sleep on the thread (which stalls the thread), or inside a waiting queue (which allows the worker thread to do useful work for active threads). 

Is that a correct characterization of a possible pattern ?

WenbinZhu: This is a pathological case where your subscriber number is 8x than your polling thread, and most of the polled data is empty. I think the optimization is not related to using a scheduled executor or not, because scheduled executor has an intrintic delay which is the schedule interval. The right optimization, as you and Medhavi both kind of mentioned, is to use a customed delay queue for the executor and we still let the thread itself submit its tasks after the previous one is done, and the tasks will be a DelayedTask in case of empty data. However for now I'm not sure if that is worthy because our workloads are not large enough to have these pathological cases.

Maithem: > However for now I'm not sure if that is worthy because our workloads are not large enough to have these pathological cases.

The max blocking time for a stream that has unread data is a function of the number of threads and number of streams. Therefore in order to determine the likelihood of the worst case happening we need to know the numbers. We currently have ~300 tables in the system (as of ~6 months ago), if we assume that 20% of the tables will be tagged then each thread will on average handle (300 * .2) / 4 = 15 streams. Since most of the data is config data (i.e. static) then its clearly not a pathological case. In the future, we can only assume that there will be more tables.

> executor has an intrintic delay which is the schedule interval

I don't understand where this intrinsic delay comes from and where it is incurred. Please explain. 

I also don't see this as an optimization per-se it just seems like a natural polling strategy, that is, to let the task sleep in the queue and not block the worker.

WenbinZhu: Sure I will explore the delayedQueue approach, but let me clarify the calculation: the blocking time is not related to number of streams, because for each tag we have only one poller. Second, the blocking time is strongly related to number of subscribers, if number of subscribers is less than or equal to the thread number (which is the common case), than there will be no delay at all. An example is If you have 8 subscribers and 4 threads, the maximum possible delay if every poller gets empty data is 8/4*50 = 100ms, not 8 * 50. The reason I'm saying it's an optimization is that our current workload is not even saturating the thread number, i.e. number of subscribers <= thread number.

> I don't understand where this intrinsic delay comes from and where it is incurred. Please explain.

If the scheduled executors you mean is the scheduledExecutorService, than I mean it has an scheduling interval, which is the intrinsic delay, that will be casted on the poller every time, not matter it gets data or not. You can set that interval to be 0, then it's no different than the current approach. I think what you mean is not related to scheduled executor, rather it is the queue inside the executor, that needs to be delayed queue, and that's when the min-heap come into play.

Maithem: > Sure I will explore the delayedQueue approach, but let me clarify the calculation

Yep. sorry my point was Subscriber/tag vs Notification threads. 

> 8/4*50 = 100ms, not 8 * 50.

We are talking about two different configs. 4 threads are currently being used, but I used 1 thread in my example to highlight the relation between notification threads and tags

>If the scheduled executors you mean is the scheduledExecutorService,

What I was referring to was [ScheduledExecutorService::schedule](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#schedule(java.lang.Runnable,%20long,%20java.util.concurrent.TimeUnit)) which creates a one-shot execution that will run after a certain delay. I'm not saying that you should schedule the tasks periodically with same delay. Using the current scheme depending on whether you read yields some entries or not, you can call `schedule(task, 0)` or `schedule(task, Delay)`. This way the task delay will be incurred in the queue waiting and not on the worker thread. `ScheduledExecutorService` internally maintains a heap ordered by the tasks delay time. Now if all the streams are active then most of your schedules will be submitted with a 0 delay and that seems to be be fine when it comes to fairness. Based on code inspection it seems like they will be in FIFO order.

WenbinZhu: > What I was referring to was ScheduledExecutorService::schedule which creates a one-shot execution that will run after a certain delay.

Ok then we are talking about the same idea, but slightly different implementation.

WenbinZhu:
@Maithem Comments addressed.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
Maithem: It seems like you ended up using the same method. Just out of curiosity, what was the different implementation that you were referring to, explicitly managing the scheduler's min-heap?

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscriptionMetrics.java
```diff
@@ -0,0 +1,35 @@
...
+import org.corfudb.runtime.CorfuRuntime;
+
+import java.util.Optional;
+
+/**
+ * List of metrics captured for a stream listener.
+ * <p>
+ * Created by WenbinZhu on 11/23/20.
+ */
+public class StreamSubscriptionMetrics {
```
Maithem, line 13: Can you also add metrics on the consumer side. The kvstore already has issues where the consumer callbacks consume too much time, so I'm expecting we will need to track this as well.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
```diff
@@ -0,0 +1,98 @@
...
+                return;
+            }
+
+            long startTime = System.nanoTime();
+            CorfuStreamEntries nextUpdate = subscription.dequeueStreamEntry(remainingBlockTime);
+
+            // No new updates after max waiting time elapses, break and re-schedule.
+            if (nextUpdate == null) {
+                break;
```
Maithem, line 79: Can you also add 

```java 

if (iter == 0 && nextUpdate == null) {

 notificationExecutor.schedule(this, IDLE_WAIT_TIME_MS, TimeUnit.MILLISECONDS);
return;
     }
```

This will prevent unnecessary waits on the consumer side when the stream is always empty. This improves the CPU usage for empty streams, but the ideal solution is to only submit the task if the producer enqueued entries. Going from 50ms polling to seconds should be ok for our scale, I think.

Maithem (approved):
Great improvement. Added some more comments about unnecessary cpu usage and debuggability.

Although I think this PR is good to go, I want to make a point about `StreamingManager` managing the threading. I don't think its necessary and I think it's an anti-pattern for a library to manage threads out of convenience and not necessity. 

I know that the current code (i.e., `StreamManager`) also does this and that this patch improves the existing implementation, but I think we still need to address that issue. Possibly in a subsequent PR.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
WenbinZhu: Yes, I used ScheduledExecutorService. The alternative is still using a normal executor, but give it a customed delayed queue. Initially I thought DelayedWorkQueue is a public utility class that I can use, but later I found it's internal to ScheduledThreadPoolExecutor.

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscriptionMetrics.java
WenbinZhu: Yes we already have this metrics that records customer callback duration.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
WenbinZhu: Actually I think we do not need to wait in this case. If we hit this case, that means we already waited for some time, so cpu won't busy loop.

WenbinZhu:
> Great improvement. Added some more comments about unnecessary cpu usage and debuggability.
> 
> Although I think this PR is good to go, I want to make a point about `StreamingManager` managing the threading. I don't think its necessary and I think it's an anti-pattern for a library to manage threads out of convenience and not necessity.
> 
> I know that the current code (i.e., `StreamManager`) also does this and that this patch improves the existing implementation, but I think we still need to address that issue. Possibly in a subsequent PR.

So I think the problem is the streaming API is used by different verticals, and there is no centralized point in application that does this, and even there is, that centralized point is probably provided by another vertical that is different from the subscribers.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
Maithem: I think it makes sense for the producer to wait one the enqueue to happen (i.e., on offer) because if the enqueue cannot happen right away it means means that the queue is full and the consumer is backed up and still processing. But when the consumer waits on poll, it never knows if the producer will ever enqueue or not, so it will keep waiting on the worker thread over and over again even if the stream is always empty.

 I feel like it makes to have `dequeueStreamEntry` return right away without waiting and then submitting the task with a delay in `notificationExecutor.schedule(this, delay)`

PR #2827, opened 2020-11-22
Protobuf Management RPCs

zfrenette:
## Overview

Description: Provides an implementation of Management RPCs using Protobuf messages.
Done as part of an effort to support rolling upgrades.

Why should this be merged: 

- [X] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
```diff
@@ -144,11 +144,15 @@ public IClientRouter addClient(IClient client) {
...
-                });
+        try {
+            client.getHandledTypes().stream()
+                    .forEach(x -> {
+                        handlerMap.put(x, client);
+                        log.info("Registered client to handle messages of type {}", x);
+                    });
+        } catch (Exception ex) {
+            // No Legacy getHandledTypes - Ignore
+        }
```
xcchang, line 155: Is the exception expected to be thrown when getHandledTypes() returns an empty set? Will it be better to explicitly check if it's empty and log the warning/info? I thought it won't throw a NPE if the streaming collection is empty.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -203,109 +172,117 @@ private CorfuRuntime getNewCorfuRuntime() {
...
-    @ServerHandler(type = CorfuMsgType.ORCHESTRATOR_REQUEST)
-    public synchronized void handleOrchestratorMsg(@Nonnull CorfuPayloadMsg<OrchestratorMsg> msg,
-                                                   @Nonnull ChannelHandlerContext ctx,
-                                                   @Nonnull IServerRouter r) {
-        log.debug("Received an orchestrator message {}", msg);
-        orchestrator.handle(msg, ctx, r);
+    @RequestHandler(type = PayloadCase.ORCHESTRATOR_REQUEST)
+    public synchronized void handleOrchestratorMsg(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        log.debug("handleOrchestratorMsg: message:{}", req.getPayload().getOrchestratorRequest());
+        orchestrator.handle(req, ctx, r);
```
xcchang, line 194: I just noticed some old RPC handlers have Nonnull constraint of their arguments (but some don't, also for other servers...), will it be better if we add Nonnull consistently to all the arguments? 
Currently I cannot image any case where the argument is null but this might be helpful for debugging sometime in the future?

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -203,109 +172,117 @@ private CorfuRuntime getNewCorfuRuntime() {
...
-    @ServerHandler(type = CorfuMsgType.MANAGEMENT_BOOTSTRAP_REQUEST)
-    public synchronized void handleManagementBootstrap(CorfuPayloadMsg<Layout> msg,
-                                                       ChannelHandlerContext ctx, IServerRouter r) {
-        if (serverContext.getManagementLayout() != null) {
-            // We are already bootstrapped, bootstrap again is not allowed.
-            log.warn("handleManagementBootstrap: Got a request to bootstrap a server " +
-                    "with {} which is already bootstrapped, rejecting!", msg.getPayload());
-            r.sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.MANAGEMENT_ALREADY_BOOTSTRAP_ERROR));
+    @RequestHandler(type = PayloadCase.BOOTSTRAP_MANAGEMENT_REQUEST)
+    public synchronized void handleBootstrapManagement(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
```
xcchang, line 206: Is there any specific criteria that some handlers are set to `synchronized` but some are not? It looks to me that the operations upon shared resources have been synchronized in a lower level, (i.e. methods in servercontext, some concurrenthashmap, etc), and all the synchronized signature could be removed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
xcchang, line 155: Is the exception expected to be thrown when getHandledTypes() returns empty set? I thought it might be better if we explicitly check `if (client.getHandledTypes().isEmpty())` and log the warn/info. I'm not 100% sure that will it throw an exception or just pass silently if the streaming collection is empty

zfrenette: Not an empty set, but when it is simply called with no implementation.  If the clients don't provide a legacy message handler explicitly, then the underlying `getMsgHandler` call throws an `UnsupportedOperation` exception. A set of size zero would cause no exception.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
zfrenette: I am not sure if it's correct, but this is my reasoning. It does not always suffice to consider that the shared resources are themselves synchronized, but one also needs to look at the interaction between them. Consider the following scenario. The management server receives a `BOOTSTRAP_MANAGEMENT` request with `Layout` A. Since it isn't bootstrapped yet, the call to `getManagementLayout` will return `null` and the thread will enter the `else` clause. Now, before the thread has a chance to call `setManagementLayout`, it goes to sleep. Now the ManagementServer receives another `BOOTSTRAP_MANAGEMENT` request with `Layout` B. Since the first request did not complete, this thread will also get `null` from the call to `getManagementLayout` and enter the `else` clause. Now we are in the state where there are two bootstrap requests with different layouts calling `setManagementLayout`. If my understanding it correct, then this is problematic and leaving the `synchronized` keyword here will prevent such an issue.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
Maithem: Those annotation can improve code inspection for inferring nullity﻿

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
Maithem: Good point, this was actually a production bug at point.

Sometimes you need a bigger lock to protect against interleaving with other "synchronized APIs". For example, think about the local data store where put/get are individually synchronized, but there are some code paths that do something like:

```java

val = datastore.get(key)
// some computation on val
val2 = someFunction(val)
datastore.put(key, val2)
````

This is a read-modify-write cycle where interleaving with another thread between the get and put can cause some inconsistency. Notice that this consistency can happen despite each API being thread-safe individually.

Although this style of synchronization can be safe when implemented carefully, its usually fragile. Ideally, the datastore should expose an interface for some notion of a transaction, where multiple operations can be executed as a single atomic unit.

File: infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java [hunk 4]
```diff
@@ -203,109 +172,117 @@ private CorfuRuntime getNewCorfuRuntime() {
...
+        final Layout layout = getLayout(req.getPayload().getBootstrapManagementRequest().getLayout());
+        HeaderMsg responseHeader;
+        ResponseMsg response;
+
+        if (layout == null || layout.getClusterId() == null) {
+            log.warn("handleBootstrapManagement[{}]: Got a request to bootstrap a server with " +
```
Maithem, line 212: Can we improve the logging message here, maybe just "handleBootstrapManagement[{}]: incomplete layout {}" would suffice.

Maithem: Also, the log level should be `log.error`

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -203,109 +172,117 @@ private CorfuRuntime getNewCorfuRuntime() {
...
+                    "incomplete layout {}", req.getHeader().getRequestId(), layout);
+
+            responseHeader = getHeaderMsg(req.getHeader(), false, false);
+            response = getResponseMsg(responseHeader, getBootstrapManagementResponseMsg(false));
+        } else if (serverContext.getManagementLayout() != null) {
+            log.warn("handleBootstrapManagement[{}]: Got a request to bootstrap a server, with layout " +
```
Maithem, line 218: "handleBootstrapManagement[{}]: already bootstrapped with {}, rejecting {}"

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -203,109 +172,117 @@ private CorfuRuntime getNewCorfuRuntime() {
...
-                                         ChannelHandlerContext ctx, IServerRouter r) {
-
-        // This server has not been bootstrapped yet, ignore all requests.
-        if (!isBootstrapped(msg)) {
-            r.sendResponse(ctx, msg, new CorfuMsg(CorfuMsgType.MANAGEMENT_NOBOOTSTRAP_ERROR));
+    @RequestHandler(type = PayloadCase.REPORT_FAILURE_REQUEST)
+    public void handleReportFailure(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        // If the server isn't bootstrapped yet, ignore the request
+        if (!isBootstrapped(req)) {
+            r.sendNoBootstrapError(req.getHeader(), ctx);
```
Maithem, line 247: Do we need to differentiate between a management server no bootstrap error and a non-management server bootstrap error? it seems like we used to have two different errors, but they were essentially the same.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
```diff
@@ -411,23 +399,19 @@ public void handleHealingDetectedMsg(CorfuPayloadMsg<DetectorMsg> msg,
...
                 .getNode(serverContext.getLocalEndpoint())
                 .orElseGet(this::buildDefaultNodeState);
 
-        r.sendResponse(ctx, msg, CorfuMsgType.NODE_STATE_RESPONSE.payloadMsg(nodeState));
-    }
-
-    @ServerHandler(type = CorfuMsgType.FAILURE_DETECTOR_METRICS_REQUEST)
-    public void handleFailureDetectorMetricsRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
-        FailureDetectorMetrics metrics = serverContext.getFailureDetectorMetrics();
-        r.sendResponse(ctx, msg, CorfuMsgType.FAILURE_DETECTOR_METRICS_RESPONSE.payloadMsg(metrics));
```
Maithem, line 430: Why is this missing from the PR?

zfrenette: It isn't currently being used anywhere and hasn't been active in a year. I spoke with @xnull and it seems fine to remove.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
zfrenette: I haven't seen a need to. The same exception was thrown in both cases. The only difference now is that this is handled by BaseHandler.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
```diff
@@ -144,11 +144,15 @@ public IClientRouter addClient(IClient client) {
...
-                    log.info("Registered client to handle messages of type {}", x);
-                });
+        try {
+            client.getHandledTypes().stream()
+                    .forEach(x -> {
+                        handlerMap.put(x, client);
+                        log.info("Registered client to handle messages of type {}", x);
+                    });
+        } catch (Exception ex) {
+            // No Legacy getHandledTypes - Ignore
```
Maithem, line 154: Why can't we explicitly fail this, or log?

Also, what if the thread was interrupted, swallowing all exceptions might cause problems, like "hanging".

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
```diff
@@ -69,70 +75,73 @@
...
+                        Thread thread = new Thread(r);
+                        thread.setDaemon(true);
+                        String threadName = serverContext.getThreadPrefix() + "orchestrator-"
+                                + threadNumber.getAndIncrement();
+                        thread.setName(threadName);
+                        thread.setUncaughtExceptionHandler(this::handleUncaughtException);
+                        return thread;
+                    }
+
+                    void handleUncaughtException(Thread t, @Nonnull Throwable e) {
```
Maithem, line 103: This can be replaced with the getExecutor helper methods in server context, right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
zfrenette: We don't want to fail the creation of the router just because a particular client (or server in the case of a server-side router) doesn't implement a legacy message handler. If it doesn't, we simply want to continue on since it isn't a sign that something has gone wrong. I can add a log message saying that a particular client or server doesn't support the handling of old messages.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
```diff
@@ -142,24 +151,22 @@ public void handle(@Nonnull CorfuPayloadMsg<OrchestratorMsg> msg,
...
 
-        Response resp;
-        if (activeWorkflows.containsKey(req.getId())) {
-            resp = new QueryResponse(true);
-            log.trace("handleQuery: returning active for id {}", req.getId());
-        } else {
-            resp = new QueryResponse(false);
-            log.trace("handleQuery: returning not active for id {}", req.getId());
-        }
+        if (activeWorkflows.containsKey(workflowId)) isActive = true;
```
Maithem, line 162: Danger. Always use curly braces. 

https://www.imperialviolet.org/2014/02/22/applebug.html

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
zfrenette: Can you clarify? The `getExecutor` helper method is being called on line 87.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuManagementServerTest.java
```diff
@@ -0,0 +1,645 @@
+package org.corfudb.infrastructure;
+
+import com.google.common.util.concurrent.MoreExecutors;
```
Maithem, line 3: Can we drop the "Corfu" prefix from the naming convention here, I feel like its superfluous.  Why don't you just replace the tests in ManagementServerTest?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/orchestrator/HealNodeRequest.java
```diff
@@ -1,14 +1,8 @@
...
-import static org.corfudb.protocols.wireprotocol.orchestrator.OrchestratorRequestType.HEAL_NODE;
-
-import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
-import java.util.Map;
-
 import lombok.Getter;
 
-import org.corfudb.util.JsonUtils;
+import static org.corfudb.protocols.wireprotocol.orchestrator.OrchestratorRequestType.HEAL_NODE;
```
Maithem, line 5: Is it possible to collapse the `CreateRequest` and `Request` interfaces?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -284,11 +284,15 @@ public IClientRouter addClient(IClient client) {
...
-                });
+        try {
+            client.getHandledTypes()
+                    .forEach(x -> {
+                        handlerMap.put(x, client);
+                        log.trace("Registered {} to handle messages of type {}", client, x);
+                    });
+        } catch (Exception ex) {
+            // No Legacy getHandledTypes - Ignore
+        }
```
Maithem, line 295: instead catching all exceptions can you just ignore unsupported operation and log it?

Also, why was this not an issue when we pushed in the new BaseServer changes?

Maithem (commented):
Look good. Just added some minor questions and comments

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
zfrenette: Will log and catch `UnsupportedOperation`. It was an issue for the `BaseServer`, but not for the `BaseHandler` because there were still some old `CorfuMsg` types that needed to be handled by it. Since it wasn't removed, no exception was thrown.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
zfrenette: Done. Added the @Nonnull annotations to the handler methods.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
zfrenette: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementServer.java
zfrenette: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
zfrenette: Added the appropriate logging statement.

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Orchestrator.java
zfrenette: Fixed.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuManagementServerTest.java
zfrenette: I removed the "Corfu" prefix from our tests, but prefer to also leave the old tests there for the time being.

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
zfrenette: Fixed. Only catching `UnsupportedOperation` and logging accordingly.

Maithem (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
```diff
@@ -144,11 +144,15 @@ public IClientRouter addClient(IClient client) {
...
-                    log.info("Registered client to handle messages of type {}", x);
-                });
+        try {
+            client.getHandledTypes().stream()
+                    .forEach(x -> {
+                        handlerMap.put(x, client);
+                        log.info("Registered client to handle messages of type {}", x);
+                    });
+        } catch (UnsupportedOperationException ex) {
+            log.warn("No registered CorfuMsg handler for client {}", client);
```
hisundar, line 154: log.error() with the exception too? This might be one of the things we would need to handle for rolling upgrade in the future, if a client message arrives that does not have any registered handler right, so perhaps a comment or TODO for the same?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationServerRouter.java
```diff
@@ -71,7 +71,7 @@ public LogReplicationServerRouter(List<AbstractServer> servers) {
             try {
                 server.getHandler().getHandledTypes().forEach(x -> handlerMap.put(x, server));
             } catch (UnsupportedOperationException ex) {
-                // Ignore
+                log.warn("No registered CorfuMsg handler for server {}", server);
```
hisundar, line 74: log.error with the exception perhaps?

hisundar (approved):
I think the test coverage with the Mockito framework is really impressive!
My comments are minor.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
zfrenette: I have adjusted the logging as suggested, but I want to point out that this is for legacy `CorfuMsg` handlers. The default behavior of these legacy methods is to throw an exception when invoked but not implemented, and will be cleaned up in the last PR once all RPCs are done over Protobuf.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationServerRouter.java
zfrenette: Done!

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
hisundar: sounds good

PR #2829, opened 2020-11-23
Add the implementation of Sequencer RPCs using Protobuf

chetangudisagar:
Provides the implementation of Sequencer RPCs using Protobuf. Done as part of an effort to support rolling upgrades.

## Overview

Description:

Why should this be merged: Enables the Sequencer RPCs to communicate using Protobuf objects instead of Java objects.

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
```diff
@@ -23,49 +23,70 @@
...
-    STREAMS_ADDRESS_REQUEST(26, new TypeToken<CorfuPayloadMsg<StreamsAddressRequest>>(){}),
-    STREAMS_ADDRESS_RESPONSE(27, new TypeToken<CorfuPayloadMsg<StreamsAddressResponse>>(){}),
+//    TOKEN_REQ(20, new TypeToken<CorfuPayloadMsg<TokenRequest>>(){}),
+//    TOKEN_RES(21, new TypeToken<CorfuPayloadMsg<TokenResponse>>(){}),
+//    BOOTSTRAP_SEQUENCER(22, new TypeToken<CorfuPayloadMsg<SequencerRecoveryMsg>>(){}),
+//    SEQUENCER_TRIM_REQ(23, new TypeToken<CorfuPayloadMsg<Long>>() {}),
+//    SEQUENCER_METRICS_REQUEST(24, TypeToken.of(CorfuMsg.class), true, false),
+//    SEQUENCER_METRICS_RESPONSE(25, new TypeToken<CorfuPayloadMsg<SequencerMetrics>>(){}, true, false),
+//    STREAMS_ADDRESS_REQUEST(26, new TypeToken<CorfuPayloadMsg<StreamsAddressRequest>>(){}),
+//    STREAMS_ADDRESS_RESPONSE(27, new TypeToken<CorfuPayloadMsg<StreamsAddressResponse>>(){}),
```
zfrenette, line 61: Can these types be removed, along with the wire protocol objects that are no longer needed (like `TokenRequest`)? Also, for the wire protocol objects that are still needed, we can still remove the `implements ICorfuPayload` as well as the serialization methods that are no longer needed.

runtime/src/main/java/org/corfudb/runtime/clients/SequencerHandler.java
```diff
@@ -37,33 +40,95 @@ public SequencerClient getClient(long epoch, UUID clusterID) {
...
+
+        switch (type) {
+            case ACK:
+                return true;
+            case NACK:
+                return false;
+            // TODO INVALID
+            default:
+                throw new UnsupportedOperationException("Response handler not provided");
+        }
```
zfrenette, line 85: Maybe this can become a boolean so that we can just directly return the result? I think this is what's being done in the layout PR.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -110,12 +122,7 @@
...
-
-    /**
-     * Per streams map and their corresponding address space (an address space is defined by the stream's addresses
-     *  and its latest trim mark)
-     */
-    private Map<UUID, StreamAddressSpace> streamsAddressMap = new HashMap<>();
+    private final FactoryHelper factoryHelper;
 
     /**
      * HandlerMethod for this server.
```
zfrenette, line 128: The old handler methods can be removed as well.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@Slf4j
+// TODO (Chetan): Remove this after completion
+@Ignore("Still under development.")
+public class CorfuSequencerServerTest {
```
zfrenette, line 61: Consider removing the 'Corfu' prefix in order to match the other tests.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+
+        // Assert that the payload has a getSequencerMetricsResponseMsg and that the base
+        // header fields have remained the same
+        assertTrue(compareBaseHeaderFields(request.getHeader(), response.getHeader()));
+        // Assert that response has a sequencerTrimResponse object (Note: SequencerTrimResponse will not have any body)
+        assertTrue(response.getPayload().hasSequencerTrimResponse());
+        // Verify that sequencerServerCache was invalidated up to the trimMark
+        ArgumentCaptor<Long> trimMarkCaptor = ArgumentCaptor.forClass(Long.class);
+        verify(cache).invalidateUpTo(trimMarkCaptor.capture());
+        assertThat(trimMarkCaptor.getValue()).isEqualTo(trimMark);
```
zfrenette, line 520: Is this argument captor necessary? I think calling verify with the expected object should suffice here. In most cases, if you know the result ahead of time, you don't need an argument captor.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+        when(mockFactoryHelper.getStreamAddressSpaceMap()).thenReturn(tailMap);
+        // The global log tail is set to 0 when sequencer is bootstrapped with the above tailMap
+        when(mockFactoryHelper.getGlobalLogTail()).thenReturn(globalTail);
+        when(mockFactoryHelper.getSequencerServerCache(anyInt(), anyLong())).thenReturn(cache);
+        sequencerServer = new SequencerServer(mockServerContext, mockFactoryHelper);
+
+        // Send a request with SequencerTrimRequestMsg
+        long trimMark = 0;
+        RequestMsg request = getRequestMsg(
+                getBasicHeader(true, true),
```
zfrenette, line 496: Should these be equal to false? That's how the message is sent in the client.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+        // Assert that response has a sequencerTrimResponse object (Note: SequencerTrimResponse will not have any body)
+        assertTrue(response.getPayload().hasSequencerTrimResponse());
+        // Verify that sequencerServerCache was invalidated up to the trimMark
+        ArgumentCaptor<Long> trimMarkCaptor = ArgumentCaptor.forClass(Long.class);
+        verify(cache).invalidateUpTo(trimMarkCaptor.capture());
+        assertThat(trimMarkCaptor.getValue()).isEqualTo(trimMark);
+
+        // Verify that the streamAddressSpace.trim() was called and verify the argument passed
+        verify(streamAddressSpace).trim(trimMarkCaptor.capture());
+        assertThat(trimMarkCaptor.getValue()).isEqualTo(trimMark);
```
zfrenette, line 524: Here too.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+                requestHeader.getClientId().equals(responseHeader.getClientId()) &&
+                requestHeader.getClusterId().equals(responseHeader.getClusterId());
+    }
+
+    /**
+     * Perform the required preparation before running individual tests.
+     * This includes preparing the mocks and initializing the DirectExecutorService.
+     */
+    @Before
+    public void setup() {
```
zfrenette, line 113: I notice that you have some repeated initialization code, such as the initialization of the server, in a handful of tests. Would it be simpler to move such snippets here? This would also help you break the tests down into smaller units without having to repeat the initialization work.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+        Map<UUID, StreamAddressSpace> tailMap = new HashMap<>();
+        long globalTail = 1;
+        // StreamAddressSpace streamAddressSpace = new StreamAddressSpace(Address.NON_ADDRESS,
+        //         Roaring64NavigableMap.bitmapOf(globalTail));
+        StreamAddressSpace streamAddressSpace = mock(StreamAddressSpace.class);
+        tailMap.put(streamA, streamAddressSpace);
+        long bootstrapMsgEpoch = 1;
+
+        SequencerServerCache cache = mock(SequencerServerCache.class);
+        SequencerServer.FactoryHelper mockFactoryHelper = mock(SequencerServer.FactoryHelper.class);
```
zfrenette, line 486: Why do you have multiple factories? You have already used the spy annotation above.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * Test that when client sends the SequencerMetricsRequestMsg, the server responds with
+     * SequencerMetricsResponseMsg with SequencerStatus.READY as its content.
+     */
+    @Test
+    public void sequencerMetricsRequest() {
+        sequencerServer = new SequencerServer(mockServerContext, spyFactoryHelper);
+        sequencerServer.setSequencerEpoch(0L);
+        RequestMsg request = getRequestMsg(
+                getBasicHeader(true, true),
```
zfrenette, line 446: Why does these values differ from the ones set when the client sends the message?

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+        // Assert that the payload has a BOOTSTRAP_SEQUENCER_RESPONSE and that the base
+        // header fields have remained the same
+        assertThat(compareBaseHeaderFields(request.getHeader(), response.getHeader())).isTrue();
+        assertThat(response.getPayload().hasBootstrapSequencerResponse()).isTrue();
+        assertThat(response.getPayload().getBootstrapSequencerResponse().getRespType())
+                .isEqualTo(Sequencer.BootstrapSequencerResponseMsg.Type.ACK);
+
+        // Verify that the mockServerContext.setSequencerEpoch() was called and inspect the argument
+        ArgumentCaptor<Long> longArgumentCaptor = ArgumentCaptor.forClass(Long.class);
+        verify(mockServerContext).setSequencerEpoch(longArgumentCaptor.capture());
```
zfrenette, line 273: I don't think this argument captor is needed. See similar comments.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+     * the address map of a particular stream requested by the client.
+     * <p>
+     * If the request type is StreamsAddressRequestMsg.ALL_STREAMS then the server should respond with
+     * the address map of all the streams.
+     * <p>
+     * If the request type is StreamsAddressRequestMsg.INVALID then the server should throw a new
+     * {@link IllegalArgumentException}.
+     */
+    @Test
+    public void testStreamAddressRequest() {
```
zfrenette, line 310: Can we consider breaking up these larger tests into smaller units, instead of testing all the data paths for an RPC in a single unit test?

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * Test that the SequencerServer correctly handles a BootstrapSequencerRequestMsg. If the
+     * epoch sent by the client is greater than or equal to the current server epoch, and bootstrapWithoutTailsUpdate
+     * is set as false, a BootstrapSequencerResponseMsg.ACK message should be received.
+     * <p>
+     * Further, we request the server with a StreamsAddressRequestMsg and verify that
+     * the globalLogTail, trimMark and the streamsAddressMap were correctly updated on the server side.
+     */
+    @Test
+    public void testBootStrapSequencer() {
```
zfrenette, line 235: Should we also have a test that validates that a bootstrap request with an older epoch is rejected?

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+
+
+    /**
+     * Test that if sequencer server's epoch is Layout.INVALID_EPOCH (startup) OR
+     * if epoch sent by the client is not the consecutive epoch of the sequencerEpoch
+     * then the sequencer should not accept bootstrapWithoutTailsUpdate BootstrapSequencerRequestMsg,
+     * and return a Sequencer.BootstrapSequencerResponseMsg.NACK message.
+     */
+    @Test
+    public void testRequireFullBootstrap() {
```
zfrenette, line 159: Can we also verify that some of the state was not updated? Methods like `setSequencerEpoch` should not have been invoked in this case.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+                .thenReturn(MoreExecutors.newDirectExecutorService());
+    }
+
+    /**
+     * Test that the SequencerServer sends NotReadyErrorMsg when client sends a TokenRequestMsg
+     * without bootstrapping it before.
+     * Essentially tests the {@link SequencerServer#isServerReadyToHandleMsg(RequestMsg)} method.
+     */
+    @Test
+    public void testSequencerServerNotReady() {
```
zfrenette, line 126: Can we also verify that some of the state was not updated? Methods like `setSequencerEpoch` should not have been invoked in this case.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
```diff
@@ -0,0 +1,528 @@
...
+                .isEqualTo(Sequencer.BootstrapSequencerResponseMsg.Type.NACK);
+    }
+
+    /**
+     * Test that stale bootstrap requests are discarded.
+     * If epoch sent by the client is not greater than the sequencerEpoch then the sequencer should discard the
+     * BootstrapSequencerRequestMsg, and return a Sequencer.BootstrapSequencerResponseMsg.NACK message.
+     */
+    @Test
+    public void testStaleBootstrapRequest() {
```
zfrenette, line 195: Can we also verify that some of the state was not updated? Methods like `setSequencerEpoch` and others should not have been invoked when the bootstrap request is stale.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
zfrenette, line 159: Same here. Consider also verifying that methods like `setSequencerEpoch` were not called.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -147,25 +154,47 @@
...
         this.serverContext = serverContext;
+        this.factoryHelper = factoryHelper;
         Config config = Config.parse(serverContext.getServerConfig());
 
         // Sequencer server is single threaded by current design
-        this.executor = Executors.newSingleThreadExecutor(
-                new ServerThreadFactory("sequencer-", new ServerThreadFactory.ExceptionHandler()));
+        this.executor = serverContext.getExecutorService(1, "sequencer-");
+
+        this.globalLogTail = factoryHelper.getGlobalLogTail();
```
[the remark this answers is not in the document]
chetangudisagar: It was initially present as a parameter initializer along with its declaration and I wanted it to be overridden by the test cases to initialize with a new custom value. So I kept the default behaviour in this part. I understand the `FactoryHelper` should always return the default objects as we will never be testing it. But I didn't see any other alternative. Can you suggest one if you have one in mind?

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolSequencer.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import static org.corfudb.protocols.CorfuProtocolCommon.*;
+import static org.corfudb.protocols.CorfuProtocolTxResolution.*;
+
+@Slf4j
+public class CorfuProtocolSequencer {
```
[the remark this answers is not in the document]
chetangudisagar: Done, Thank you.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolTxResolution.java
```diff
@@ -0,0 +1,81 @@
...
+import org.corfudb.runtime.proto.TxResolution.TxResolutionInfoMsg;
+import org.corfudb.runtime.proto.TxResolution.UuidToListOfBytesPairMsg;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static org.corfudb.protocols.CorfuProtocolCommon.*;
+
+@Slf4j
+public class CorfuProtocolTxResolution {
```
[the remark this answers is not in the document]
chetangudisagar: Done, Thank you.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -174,8 +203,9 @@ public SequencerServer(ServerContext serverContext) {
     }
 
     @Override
+    @Deprecated
     protected void processRequest(CorfuMsg msg, ChannelHandlerContext ctx, IServerRouter r) {
-        executor.submit(() -> getHandler().handle(msg, ctx, r));
+        throw new UnsupportedOperationException("This operation is not supported");
     }
```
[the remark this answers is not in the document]
chetangudisagar: Done, Thank you.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -190,26 +220,25 @@ public void shutdown() {
...
 
     @Override
+    @Deprecated
     public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
-        if (getState() != ServerState.READY){
-            return false;
-        }
+        throw new UnsupportedOperationException("This operation is not supported");
+    }
```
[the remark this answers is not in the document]
chetangudisagar: Done, Thank you.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -11,31 +11,32 @@
...
-
 import org.corfudb.common.metrics.micrometer.MeterRegistryProvider;
-import org.corfudb.runtime.view.stream.StreamAddressSpace;
-import org.corfudb.protocols.wireprotocol.StreamAddressRange;
-import org.corfudb.protocols.wireprotocol.StreamsAddressRequest;
-import org.corfudb.protocols.wireprotocol.StreamsAddressResponse;
-import org.roaringbitmap.longlong.Roaring64NavigableMap;
 import org.corfudb.infrastructure.SequencerServerCache.ConflictTxStream;
+import org.corfudb.protocols.CorfuProtocolCommon;
 import org.corfudb.protocols.wireprotocol.CorfuMsg;
```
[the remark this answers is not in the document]
chetangudisagar: Done, Thank you.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -707,11 +790,25 @@ private void handleStreamsAddressRequest(CorfuPayloadMsg<StreamsAddressRequest>
...
-            Integer.parseInt((String)opts.get("--sequencer-cache-size")) : DEFAULT_CACHE_SIZE);
+            int cacheSize = opts.containsKey("--sequencer-cache-size") ?
+                    Integer.parseInt((String) opts.get("--sequencer-cache-size")) : DEFAULT_CACHE_SIZE;
             return Config.builder()
                     .cacheSize(cacheSize)
                     .build();
         }
     }
+
+    static class FactoryHelper {
```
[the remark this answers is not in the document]
chetangudisagar: refactored into `SequencerFactoryHelper`. Thank you.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -617,56 +693,62 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
...
+
+        if (streamsAddressRequest.getReqType() == StreamsAddressRequestMsg.Type.STREAMS) {
+            respStreamsAddressMap = getStreamsAddressesMap(streamsAddressRequest.getStreamRangeList());
+        } else if (streamsAddressRequest.getReqType() == StreamsAddressRequestMsg.Type.ALL_STREAMS) {
+            // Retrieve address space for all streams
+            respStreamsAddressMap = new HashMap<>(this.streamsAddressMap);
+        } else {
+            throw new IllegalArgumentException("handleStreamsAddressRequest: " +
+                    "Received an INVALID type of streamsAddressRequestMsg.");
         }
```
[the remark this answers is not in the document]
chetangudisagar: Can be done but it may not be future proof in this use case, like, what if a new type of stream request comes up other than streams and all_streams? for example, streams meet a certain criteria/behaviour -- `active_streams`, `inactive_streams`. The chances are less but there is a possibility. So I didn't bother changing it. I thought that was one reason why it was initially an Enum instead of a boolean value.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -707,11 +790,25 @@ private void handleStreamsAddressRequest(CorfuPayloadMsg<StreamsAddressRequest>
...
+            int cacheSize = opts.containsKey("--sequencer-cache-size") ?
+                    Integer.parseInt((String) opts.get("--sequencer-cache-size")) : DEFAULT_CACHE_SIZE;
             return Config.builder()
                     .cacheSize(cacheSize)
                     .build();
         }
     }
+
+    static class FactoryHelper {
+        public Map<UUID, StreamAddressSpace> getStreamAddressSpaceMap() {
```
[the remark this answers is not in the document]
chetangudisagar: Done, Thank you.

runtime/src/main/java/org/corfudb/runtime/clients/SequencerHandler.java
chetangudisagar: Done, Thank you.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
chetangudisagar: Done, Thank you.

Removed the following files.
```
- TokenRequest.java
- SequencerRecoveryMsg.java
- StreamsAddressRequest.java
```

Removed `implements ICorfuPayload` and serialization and deserialization constructors for the following files. (I couldn't remove them as they were used beyond the RPC layer.)
```
- TokenResponse.java
- SequencerMetrics.java
- StreamsAddressResponse.java
```

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Done, thank you.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Done, Thank you.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Done, Thank you.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Done, Thank you.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Thanks for pointing it out. I was still in the middle of some changes. I will keep in mind to move the common code here.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Good catch.
`when(mockSequencerFactoryHelper.getSequencerServerCache(anyInt(), anyLong())).thenReturn(cache);`
This line was failing with `IllegalArgumentException`  from `SequencerServerCache` class (or above levels) with spy one as it was actually invoking the method. Hence, I had to convert it into mock. I didn't want to change any of that code. I even tried some default values(ex - `DEFAULT_CACHE_SIZE = 250_000`) but to no avail.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Fixed them. Thank you.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: Done, Thank you.

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
chetangudisagar: They all operate on the same `streamsAddressSpaceMap` data on the sequencer server. The only difference between each one is the way they fetch the address space data (specific stream vs all stream). The third one is an invalid one. Hence I clubbed them. Splitting would slightly increase the time and space required for tests.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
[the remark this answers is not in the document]
zfrenette: Hmm. Would spying on the `SequencerServer` instance work here? There's a getter annotation for the `globalLogTail`, so you could inject your desired value through this getter method?

infrastructure/src/test/java/org/corfudb/infrastructure/CorfuSequencerServerTest.java
zfrenette: I'm not sure I understand where the exception comes from. Something like `doReturn(cache).when(mockSequencerFactoryHelper).getSequencerServerCache(anyInt(), anyLong())` should work. Alternatively, maybe we can always use a mock of the factory instead of a spy?

Note: https://stackoverflow.com/questions/11620103/mockito-trying-to-spy-on-method-is-calling-the-original-method provides a few possible explanations for the issue.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolCommon.java
```diff
@@ -258,4 +273,50 @@ public static StreamsAddressResponse getStreamsAddressResponse(long tail, List<U
...
+     * parameters.
+     *
+     * @param logTail    the logTail to be set on the {@link Sequencer.StreamsAddressResponseMsg} object
+     * @param epoch    the epoch to be set on the {@link Sequencer.StreamsAddressResponseMsg} object
+     * @param addressMap addressMap of the {@link Sequencer.StreamsAddressResponseMsg} object
+     * @return a new {@link CorfuMessage.ResponsePayloadMsg} Protobuf object
+     */
+    public static CorfuMessage.ResponsePayloadMsg getStreamsAddressResponseMsg(
+            long logTail,
+            long epoch,
```
xcchang, line 289: Could we use on-demand import for ResponsePayloadMsg?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -342,72 +367,104 @@ private void handleTokenQuery(CorfuPayloadMsg<TokenRequest> msg,
...
+        );
+
+        r.sendResponse(response, ctx);
     }
 
-    @ServerHandler(type = CorfuMsgType.SEQUENCER_TRIM_REQ)
-    public void trimCache(CorfuPayloadMsg<Long> msg, ChannelHandlerContext ctx, IServerRouter r) {
+    @RequestHandler(type = PayloadCase.SEQUENCER_TRIM_REQUEST)
+    public void trimCache(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
         log.info("trimCache: Starting cache eviction");
```
xcchang, line 392: Add @Nonnull annotation for the request handler methods could improve code inspection for inferring nullity﻿. Please check this comment
https://github.com/CorfuDB/CorfuDB/pull/2827#discussion_r535539159

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
```diff
@@ -70,6 +60,8 @@
     ERROR_VALUE_ADOPTED(59,new TypeToken<CorfuPayloadMsg<ReadResponse>>() {}),
 
     // EXTRA CODES
+    LAYOUT_ALREADY_BOOTSTRAP(60, TypeToken.of(CorfuMsg.class), true, false),
+    LAYOUT_PREPARE_ACK(61, new TypeToken<CorfuPayloadMsg<LayoutPrepareResponse>>(){}, true, false),
```
WenbinZhu, line 64: Where are these two msg types used? Looks like these are accidently added.

WenbinZhu (commented):
Frist batch of comments.

WenbinZhu (approved):
LGTM, have some comments.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
xcchang: Yeah these two msg types were removed in the PR for Layout PRC, they might be accidentally added when solving the conflicts.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolCommon.java
```diff
@@ -243,19 +256,48 @@ public static StreamAddressRangeMsg getStreamAddressRangeMsg(StreamAddressRange
...
-     * represented in Protobuf.
+     * Returns a new {@link ResponsePayloadMsg} Protobuf object consisting of a
+     * {@link StreamsAddressResponseMsg} object with the logTail and addressMap set from the
+     * parameters.
      *
-     * @param tail   the log tail
-     * @param map    a list of address map entries represented in Protobuf
-     * @return       an equivalent StreamsAddressResponse object
+     * @param logTail    the logTail to be set on the {@link StreamsAddressResponseMsg} object
+     * @param epoch    the epoch to be set on the {@link StreamsAddressResponseMsg} object
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -538,62 +638,69 @@ private void handleRawToken(CorfuPayloadMsg<TokenRequest> msg, ChannelHandlerCon
...
         //   1. obtain the last back-pointer for this stream, if exists; -1L otherwise.
         //   2. record the new global tail as back-pointer for this stream.
         //   3. Add the allocated addresses to each stream's address map.
         ImmutableMap.Builder<UUID, Long> backPointerMap = ImmutableMap.builder();
-        for (UUID id : req.getStreams()) {
+        for (UuidMsg id : tokenRequest.getStreamsList()) {
 
             // step 1. and 2. (comment above)
-            streamTailToGlobalTailMap.compute(id, (k, v) -> {
+            streamTailToGlobalTailMap.compute(getUUID(id), (k, v) -> {
```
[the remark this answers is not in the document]
xcchang: Added, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -190,26 +220,21 @@ public void shutdown() {
     }
 
     @Override
-    public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
-        if (getState() != ServerState.READY){
-            return false;
-        }
+    public boolean isServerReadyToHandleMsg(RequestMsg request) {
+        if (getState() != ServerState.READY) return false;
```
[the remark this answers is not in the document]
xcchang: Added, thanks!

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -0,0 +1,898 @@
...
+ * The RequestMsg type messages that are tested are as follows -
+ * - TokenRequestMsg
+ * - BootstrapSequencerRequestMsg
+ * - SequencerTrimRequestMsg
+ * - SequencerMetricsRequestMsg
+ * - StreamsAddressRequestMsg
+ * See test methods for their functionality descriptions.
+ */
+@Slf4j
+public class SequencerServerTest {
```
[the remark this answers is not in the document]
xcchang: Hmm, I thought it was Maithem's suggestion to keep this name (we originally name them like CorfuSequencerServerTest, etc). They might not be able to be combined since these two test sets have quite different setup, and we intended to replace the old ones with the new Mockito unit tests later. Maybe @zfrenette and @Maithem could add more detail on this?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/StreamsAddressResponse.java
```diff
@@ -33,6 +33,9 @@ public StreamsAddressResponse(long logTail, Map<UUID, StreamAddressSpace> stream
     /**
      * Deserialization Constructor from Bytebuf to StreamsAddressResponse.
      *
+     * Note: Removing this method fails the
+     * AutoCommitServiceTest.linktestSequencerChangeDuringAutoCommit() test case}
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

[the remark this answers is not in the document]
xcchang: I tried to remove the inheritance and the `testSequencerChangeDuringAutoCommit` failed as stated in the javadoc. Could you share what you found about this? @chetangudisagar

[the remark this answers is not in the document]
zfrenette: This is also used by the `LogUnitServer`, so it cannot be removed until both servers are running on Protobuf.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -419,114 +492,141 @@ public void resetServer(CorfuPayloadMsg<SequencerRecoveryMsg> msg,
...
                     Tags.empty(), Meter.Type.GAUGE);
-            cache = new SequencerServerCache(cache.getCacheSize(), globalLogTail - 1);
+            cache = sequencerFactoryHelper.getSequencerServerCache(
+                    cache.getCacheSize(),
+                    globalLogTail - 1
+            );
+
             // Clear the existing map as it could have been populated by an earlier reset.
-            streamTailToGlobalTailMap = new HashMap<>();
+            streamTailToGlobalTailMap.clear();
```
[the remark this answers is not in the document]
xcchang: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -419,114 +492,141 @@ public void resetServer(CorfuPayloadMsg<SequencerRecoveryMsg> msg,
...
+                if (log.isTraceEnabled()) {
+                    log.trace("On Sequencer reset, tail for stream {} set to {}",
+                            streamAddressSpace.getKey(), streamTail);
+                }
                 streamTailToGlobalTailMap.put(streamAddressSpace.getKey(), streamTail);
             }
 
             // Reset streams address map
-            this.streamsAddressMap = new HashMap<>();
+            this.streamsAddressMap.clear();
```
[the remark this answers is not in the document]
xcchang: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -190,26 +220,23 @@ public void shutdown() {
...
-        if ((sequencerEpoch != serverContext.getServerEpoch())
-                && (!msg.getMsgType().equals(CorfuMsgType.BOOTSTRAP_SEQUENCER))) {
+        if ((sequencerEpoch != serverContext.getServerEpoch()) &&
+                (!request.getPayload().getPayloadCase()
+                        .equals(PayloadCase.BOOTSTRAP_SEQUENCER_REQUEST))) {
+
             log.warn("Rejecting msg at sequencer : sequencerStateEpoch:{}, serverEpoch:{}, "
-                    + "msg:{}", sequencerEpoch, serverContext.getServerEpoch(), msg);
+                            + "header:{}", sequencerEpoch, serverContext.getServerEpoch(),
+                    request.getHeader());
```
zfrenette, line 234: Can we use `TextFormat.shortDebugString` here and in other places where we log a Protobuf message?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -419,114 +494,141 @@ public void resetServer(CorfuPayloadMsg<SequencerRecoveryMsg> msg,
...
-    @ServerHandler(type = CorfuMsgType.TOKEN_REQ)
-    public void tokenRequest(CorfuPayloadMsg<TokenRequest> msg,
-                                          ChannelHandlerContext ctx, IServerRouter r) {
-        log.trace("Token request. Msg: {}", msg);
+    @RequestHandler(type = PayloadCase.TOKEN_REQUEST)
+    public void tokenRequest(@Nonnull RequestMsg req,
+                             @Nonnull ChannelHandlerContext ctx,
+                             @Nonnull IServerRouter r) {
+        if (log.isTraceEnabled()) {
+            log.trace("tokenRequest: Token request msg: {}", req);
```
zfrenette, line 588: Use `TextFormat.shortDebugString`?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -617,56 +731,78 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
...
+        if (streamsAddressRequest.getReqType() == StreamsAddressRequestMsg.Type.STREAMS) {
+            respStreamsAddressMap =
+                    getStreamsAddressesMap(streamsAddressRequest.getStreamRangeList());
+        } else if (streamsAddressRequest.getReqType() ==
+                StreamsAddressRequestMsg.Type.ALL_STREAMS) {
+            // Retrieve address space for all streams
+            respStreamsAddressMap = new HashMap<>(streamsAddressMap);
+        } else {
+            throw new IllegalArgumentException("handleStreamsAddressRequest: " +
+                    "Received an INVALID type of streamsAddressRequestMsg.");
```
zfrenette, line 785: Maybe include the type received as part of the exception message?

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -0,0 +1,986 @@
...
+                requestHeader.getClientId().equals(responseHeader.getClientId()) &&
+                requestHeader.getClusterId().equals(responseHeader.getClusterId());
+    }
+
+    /**
+     * Perform the required preparation before running individual tests.
+     * This includes preparing the mocks and initializing the DirectExecutorService.
+     */
+    @Before
+    public void setup() {
```
zfrenette, line 145: Maybe consider factoring some of the common preparation code here? If nothing can be cleanly factored, consider updating the comment.

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -0,0 +1,986 @@
...
+        assertThat(compareBaseHeaderFields(request.getHeader(), response.getHeader())).isTrue();
+        assertThat(response.getPayload().hasBootstrapSequencerResponse()).isTrue();
+        assertThat(response.getPayload().getBootstrapSequencerResponse().getIsBootstrapped())
+                .isFalse();
+
+        // Assert that the sequencerServer's and serverContext's Epoch has not changed
+        assertEquals(Layout.INVALID_EPOCH, sequencerServer.getSequencerEpoch());
+        verify(mockServerContext, never()).setSequencerEpoch(anyLong());
+
+        // CASE 1: the sequencer server's epoch != Layout.INVALID_EPOCH
```
zfrenette, line 233: CASE 2:

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -0,0 +1,986 @@
...
+    @Mock
+    private ServerContext mockServerContext;
+    @Mock
+    private IServerRouter mockServerRouter;
+    @Mock
+    private ChannelHandlerContext mockChannelHandlerContext;
+    @Mock
+    private Map<UUID, StreamAddressSpace> mockStreamAddressSpaceMap;
+    @Mock
+    private Map<UUID, Long> mockStreamTailToGlobalTailMap;
```
zfrenette, line 105: Out of curiosity, why does this need to be mocked?

zfrenette (commented):
Looks pretty good overall. Most of my comments/concerns are minor.

runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
```diff
@@ -54,9 +57,10 @@ public SequencerClient(IClientRouter router, long epoch, UUID clusterID) {
      * @param streamsAddressesRange requested streams and ranges.
      * @return streams address maps in the given range.
      */
-    public CompletableFuture<StreamsAddressResponse> getStreamsAddressSpace(List<StreamAddressRange> streamsAddressesRange) {
-        return sendMessageWithFuture(CorfuMsgType.STREAMS_ADDRESS_REQUEST.payloadMsg(
-                new StreamsAddressRequest(streamsAddressesRange)));
+    public CompletableFuture<StreamsAddressResponse> getStreamsAddressSpace(
+            List<StreamAddressRange> streamsAddressesRange) {
+        return sendRequestWithFuture(getStreamsAddressRequestMsg(streamsAddressesRange),
+                false, false);
```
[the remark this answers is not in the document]
zfrenette: Actually, looking at the test cases in SequencerServerTest suggests that this is not an issue.

runtime/src/main/java/org/corfudb/runtime/clients/SequencerHandler.java
```diff
@@ -37,33 +41,89 @@ public SequencerClient getClient(long epoch, UUID clusterID) {
...
+    /**
+     * Handle a token response from the server.
+     *
+     * @param msg The token response message.
+     * @param ctx The context the message was sent under.
+     * @param r   A reference to the router.
+     * @return {@link TokenResponseMsg} sent back from server.
+     */
+    @ResponseHandler(type = PayloadCase.TOKEN_RESPONSE)
+    private static Object handleTokenResponse(CorfuMessage.ResponseMsg msg,
```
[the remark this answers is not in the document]
xcchang: Done.

runtime/src/main/java/org/corfudb/runtime/clients/SequencerHandler.java
```diff
@@ -37,33 +41,89 @@ public SequencerClient getClient(long epoch, UUID clusterID) {
...
+     * @param r   A reference to the router.
+     * @return True if ACK, false if NACK.
+     */
+    @ResponseHandler(type = PayloadCase.BOOTSTRAP_SEQUENCER_RESPONSE)
+    private static Object handleBootstrapSequencerResponse(ResponseMsg msg,
+                                                           ChannelHandlerContext ctx,
+                                                           IClientRouter r) {
+        BootstrapSequencerResponseMsg responseMsg =
+                msg.getPayload().getBootstrapSequencerResponse();
+        return responseMsg.getIsBootstrapped();
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

runtime/src/test/java/org/corfudb/runtime/clients/SequencerHandlerTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+    /**
+     * A helper method that creates a basic message header populated
+     * with default values.
+     *
+     * @param ignoreClusterId   indicates if the message is clusterId aware
+     * @param ignoreEpoch       indicates if the message is epoch aware
+     * @return                  the corresponding HeaderMsg
+     */
+    private CorfuMessage.HeaderMsg getBasicHeader(boolean ignoreClusterId, boolean ignoreEpoch) {
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolCommon.java
```diff
@@ -243,19 +256,48 @@ public static StreamAddressRangeMsg getStreamAddressRangeMsg(StreamAddressRange
     }
 
     /**
-     * Returns a StreamAddressResponse object from its log tail and List
-     * of address map entries, each consisting of a UUID and a StreamAddressSpace,
-     * represented in Protobuf.
+     * Returns a new {@link ResponsePayloadMsg} Protobuf object consisting of a
+     * {@link StreamsAddressResponseMsg} object with the logTail and addressMap set from the
+     * parameters.
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolTxResolution.java
```diff
@@ -0,0 +1,111 @@
+package org.corfudb.protocols;
+
+import com.google.protobuf.ByteString;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.wireprotocol.Token;
+import org.corfudb.protocols.wireprotocol.TxResolutionInfo;
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolSequencer.java
```diff
@@ -0,0 +1,369 @@
...
+                                                         Map<UUID, Long> backPointerMap) {
+        return getTokenResponseMsg(TokenType.NORMAL, TokenResponse.NO_CONFLICT_KEY,
+                TokenResponse.NO_CONFLICT_STREAM, token, backPointerMap, Collections.emptyMap());
+    }
+
+    /**
+     * Returns the Java {@link TokenResponse} object from the given {link TokenResponseMsg}
+     * Protobuf object.
+     *
+     * @param msg the {@link TokenResponseMsg} Protobuff object
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

test/src/test/java/org/corfudb/runtime/view/AbstractViewTest.java
```diff
@@ -317,9 +313,16 @@ public void bootstrapAllServers(Layout l) {
                 });
         TestServer primarySequencerNode = testServerMap.get(l.getSequencers().get(0));
         primarySequencerNode.sequencerServer
-                .handleMessage(CorfuMsgType.BOOTSTRAP_SEQUENCER.payloadMsg(new SequencerRecoveryMsg(0L,
-                        Collections.emptyMap(), l.getEpoch(), false)), null,
-                        primarySequencerNode.serverRouter);
+                .handleMessage(
+                        CorfuProtocolMessage.getRequestMsg(
+                                getBasicHeader(true, true),
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

runtime/src/test/java/org/corfudb/runtime/clients/SequencerHandlerTest.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    /**
+     * Test that the SequencerHandler correctly handles a SEQUENCER_METRICS_RESPONSE.
+     */
+    @Test
+    public void testSequencerMetricsResponseNormal() {
+        SequencerMetrics sequencerMetricsReady = SequencerMetrics.READY;
+        SequencerMetrics sequencerMetricsNotReady = SequencerMetrics.READY;
+        SequencerMetrics sequencerMetricsUnknown = SequencerMetrics.READY;
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

runtime/src/test/java/org/corfudb/runtime/clients/SequencerHandlerTest.java
```diff
@@ -0,0 +1,322 @@
...
+        verify(mockClientRouter, never()).completeExceptionally(anyLong(), any(Throwable.class));
+        verify(mockClientRouter).completeRequest(eq(response.getHeader().getRequestId()), captor.capture());
+
+        StreamsAddressResponse streamsAddressResponse = captor.getValue();
+        assertEquals(defaultLogTail, streamsAddressResponse.getLogTail());
+        assertEquals(defaultEpoch, streamsAddressResponse.getEpoch());
+        Map<UUID, StreamAddressSpace> retMap = streamsAddressResponse.getAddressMap();
+        for (UUID id : defaultMap.keySet()) {
+            assertEquals(defaultMap.get(id).toString(), retMap.get(id).toString());
+        }
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xcchang: Done, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xcchang: Done, thanks!

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
xcchang: Updated the comment, thanks!

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
xcchang: Done, thanks!

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
xcchang: These two maps were exposed to have some verifications in the test cases, but we now use `new HashMap<>()` pattern in the `SequencerServer` and don't want to go into such details in tests, so they are deleted.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -707,11 +846,41 @@ private void handleStreamsAddressRequest(CorfuPayloadMsg<StreamsAddressRequest>
...
+     * Used by the unit tests to inject a custom value for the required parameters through the
+     * constructor
+     *
+     * The default implementations are listed here which will be overloaded with
+     * the custom return values in unit tests.
+     *
+     * Note: This class should always return the default/initial objects and should not have
+     *       any logic as it will never be tested.
+     */
+    static class SequencerFactoryHelper {
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xcchang: Added, thanks!

runtime/src/main/java/org/corfudb/protocols/wireprotocol/StreamsAddressResponse.java
```diff
@@ -41,6 +44,14 @@ public StreamsAddressResponse(ByteBuf buf) {
         this.addressMap = ICorfuPayload.mapFromBuffer(buf, UUID.class, StreamAddressSpace.class);
     }
 
+    /**
+     * Serialize a the object and append to the {@link ByteBuf}  parameter passed.
+     *
+     * Note: Removing this method fails the
+     * AutoCommitServiceTest.testSequencerChangeDuringAutoCommit() test case
+     *
```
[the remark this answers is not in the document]
xcchang: Added, thanks!

test/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -199,23 +214,25 @@ public void SequencerWillResetTails() throws Exception {
...
+                false, false);
         long tailB = future.join().getToken().getSequence();
 
-        future = sendRequest(new CorfuPayloadMsg<>(CorfuMsgType.TOKEN_REQ,
-                new TokenRequest(1L, Collections.singletonList(streamC))));
-        future = sendRequest(new CorfuPayloadMsg<>(CorfuMsgType.TOKEN_REQ,
-                new TokenRequest(1L, Collections.singletonList(streamC))));
+        future = sendRequest(
+                getTokenRequestMsg(1L, Collections.singletonList(streamC)),
+                false, false);
```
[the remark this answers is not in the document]
xcchang: Just checked that sending two requests won't change the verification logic later, so I'd prefer to keep one request since  having two identical sendRequest blocks here looks a little weird.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -431,8 +435,9 @@ private void forceLayout(@Nonnull RequestMsg req, @Nonnull ChannelHandlerContext
     // TODO as this message is not set to ignore EPOCH.
     // TODO How do we handle holes in history if we let in layout commit message. Maybe we have a
     // TODO hole filling process
+    @VisibleForTesting
     @RequestHandler(type = RequestPayloadMsg.PayloadCase.COMMIT_LAYOUT_REQUEST)
-    public void handleCommitLayoutRequest(@Nonnull RequestMsg req, @Nonnull ChannelHandlerContext ctx,
+    void handleCommitLayoutRequest(@Nonnull RequestMsg req, @Nonnull ChannelHandlerContext ctx,
```
Maithem, line 440: These changes to the LayoutServer are addressing comments from the LayoutServer PR ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -419,114 +495,141 @@ public void resetServer(CorfuPayloadMsg<SequencerRecoveryMsg> msg,
...
         // Update epochRangeLowerBound if the bootstrap epoch is not consecutive.
-        if (epochRangeLowerBound == Layout.INVALID_EPOCH || bootstrapMsgEpoch != sequencerEpoch + 1) {
+        if (epochRangeLowerBound == Layout.INVALID_EPOCH
+                || bootstrapMsgEpoch != sequencerEpoch + 1) {
             epochRangeLowerBound = bootstrapMsgEpoch;
         }
 
         // Mark the sequencer as ready after the tails have been populated.
-        sequencerEpoch = bootstrapMsgEpoch;
+        setSequencerEpoch(bootstrapMsgEpoch);
```
Maithem, line 551: sequencerEpoch should probably be a private volatile.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -617,56 +731,78 @@ private void handleAllocation(CorfuPayloadMsg<TokenRequest> msg,
...
+                    "Received an " + reqType + "type of streamsAddressRequestMsg.");
         }
-
-        log.trace("handleStreamsAddressRequest: return address space for streams [{}]",
-                streamsAddressMap.keySet());
-        r.sendResponse(ctx, msg, CorfuMsgType.STREAMS_ADDRESS_RESPONSE.payloadMsg(
-                new StreamsAddressResponse(getGlobalLogTail(), streamsAddressMap)));
+        if (log.isTraceEnabled()) {
+            log.trace("handleStreamsAddressRequest: return address space for streams [{}]",
+                    respStreamsAddressMap.keySet());
```
Maithem, line 789: This changes the default behavior from the switch statement.

runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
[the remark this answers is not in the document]
Maithem: Java doesn't have unsigned longs. When the system is first bootstrap the trim mark should be less than 0. Otherwise address 0 will be incorrectly considered trimmed. Why not use int64?

Maithem (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
xcchang: Yes they are

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
xcchang: Done,  thanks!

runtime/src/main/java/org/corfudb/runtime/clients/SequencerClient.java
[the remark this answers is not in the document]
zfrenette: Thanks for confirming. This change has been made.

PR #2830, opened 2020-11-24
Enable Tx Logging Back on Standby

corfudb-performance:
## Overview

Description: enable TX logging back on LR components

Why should this be merged: UI relies on TX log updates to refresh its content. Currently, replicated data does not show up on the UI.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
@@ -51,9 +51,6 @@ public class LogReplicationMetadataManager {
 
     private final Table<ReplicationStatusKey, ReplicationStatusVal, ReplicationStatusVal> replicationStatusTable;
 
-    private final CorfuRuntime runtimeTxLogging;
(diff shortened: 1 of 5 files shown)


zhangn49 (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -708,9 +696,7 @@ public void unsubscribeReplicationEventTable(StreamListener listener) {
     }
 
     public void shutdown() {
-        if (runtimeTxLogging != null) {
-            runtimeTxLogging.shutdown();
-        }
```
achinmishra, line 713: shutdown() handler's  a no-op? may I ask why?

annym: Because the runtime used here is shared so it is shutdown at the upper layer.. it was to used before to shutdown the specific Tx Log enabled runtime, but now that we've removed it, there is not any shutdown specific logic to trigger...

achinmishra (approved):

Lujie1996 (approved):

PR #2835, opened 2020-11-30
Optimize StreamAddressSpace::getAddressesInRange

Maithem:
## Overview
This patch replaces a linear scan with a reverse iterator to
efficiently query a stream's suffix.

Why should this be merged: Optimizes stream tail queries which improves the stream syncing performance.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/protocols/wireprotocol/ICorfuPayload.java
```diff
@@ -397,8 +397,6 @@
             buffer.writeLong(streamRange.getEnd());
         } else if (payload instanceof Roaring64NavigableMap) {
             Roaring64NavigableMap mrb = (Roaring64NavigableMap) payload;
-            // Improve compression
-            mrb.runOptimize();
```
zhangn49, line 401: Could you please help explain why we are removing runOptimize()?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -111,6 +111,16 @@ public void addAddress(long address) {
         addressMap.addLong(address);
     }
 
+    /**
+     * Add an array of addresses to the stream address space
+     * @param addresses add addresses to the underlying bitmap
+     */
+    public void addAddresses(long ... addresses) {
```
zhangn49, line 118: Looks like it is used by a unit test only, do we need it? If so, the underlying bitmap has API to add an array. `addressMap.add(addresses)`

zhangn49 (approved):

zfrenette (approved):

PavelZaytsev (approved):

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ICorfuPayload.java
Maithem: This code path is called when the sequencer replies to a clients stream query, which is usually a small message. I don't think optimizing the bitset just for transfer is worth the latency cost because the bitset is relatively small to begin with.

Maithem:
superseded by #2871

PR #2836, opened 2020-11-30
Protobuf LogUnit RPCs

xcchang:
Provides routing support for Protobuf messages and an implementation of LogUnit
RPCs using Protobuf messages. Done as part of an effort to support rolling upgrades

## Overview

Description:

Why should this be merged: 

- [X] Public API has Javadoc


infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
```diff
@@ -0,0 +1,308 @@
...
+    private StreamLog mockStreamLog;
+    private final AtomicInteger requestCounter = new AtomicInteger();
+    private final long DEFAULT_SEAL_EPOCH = 1L;
+    private final long LARGER_SEAL_EPOCH = 5L;
+
+
+    /**
+     * A helper method that creates a basic message header populated
+     * with default values. Note that the sealEpoch in BatchProcessor
+     * should equal to the epoch in request header.
```
hisundar, line 67: nit: should be equal to the epoch in request header

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
```diff
@@ -0,0 +1,308 @@
...
+     */
+    private HeaderMsg getBasicHeader(boolean ignoreClusterId, boolean ignoreEpoch) {
+        return getHeaderMsg(requestCounter.incrementAndGet(), PriorityLevel.NORMAL, DEFAULT_SEAL_EPOCH,
+                getUuidMsg(DEFAULT_UUID), getUuidMsg(DEFAULT_UUID), ignoreClusterId, ignoreEpoch);
+    }
+
+    /**
+     * A helper method that creates a basic message header populated
+     * with default values. Note that the sealEpoch in BatchProcessor
+     * should equal to the epoch in request header.
```
hisundar, line 79: nit: should be equal to the epoch in request header

test/src/test/resources/logback-test.xml
```diff
@@ -62,9 +62,9 @@
     <logger name="org.corfudb.runtime.object" level="INFO"/>
     <logger name="org.corfudb.runtime.clients" level="INFO"/>
     <logger name="org.corfudb.infrastructure" level="INFO"/>
-    <logger name="io.netty.util" level="INFO"/>
-    <logger name="io.netty.util.internal" level="INFO"/>
-    <logger name="io.netty.buffer" level="INFO"/>
+    <logger name="io.netty.util" level="WARN"/>
+    <logger name="io.netty.util.internal" level="WARN"/>
+    <logger name="io.netty.buffer" level="WARN"/>
```
hisundar, line 67: Can the test log level still be left as INFO?

test/src/test/resources/logback-test.xml
```diff
@@ -77,7 +77,7 @@
 
     <root level="INFO">
         <!--<appender-ref ref="FILE" />-->
-        <!--<appender-ref ref="STDOUT" />-->
+        <appender-ref ref="STDOUT" />
```
hisundar, line 80: Do you want to check this in with the default appender set to STDOUT?

hisundar (changes requested):
Looks good to me overall. Impressive set of tests added!

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
xcchang: Added, thanks!

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
xcchang: Added, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +184,256 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
-    @ServerHandler(type = CorfuMsgType.WRITE)
-    public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        LogData logData = (LogData) msg.getPayload().getData();
-        log.debug("log write: type: {}, address: {}, streams: {}", logData.getType(),
-                logData.getToken(), logData.getBackpointerMap());
+    @RequestHandler(type = PayloadCase.WRITE_LOG_REQUEST)
+    public void write(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        LogData logData = getLogData(req.getPayload().getWriteLogRequest().getLogData());
+        log.debug("log write: type: {}, address: {}, streams: {}",
+                logData.getType(), logData.getToken(), logData.getBackpointerMap());
```
xcchang, line 284: Will it be better to keep the naming style for those handlers consistent? like handleWrite, handleRangeWrite, etc

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +185,256 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
         // because the requirement is to allow reads, but only block writes once the quota is exhausted
         if (logData.isHole()) {
-            msg.setPriorityLevel(PriorityLevel.HIGH);
+            req = getRequestMsg(getHighPriorityHeaderMsg(req.getHeader()), req.getPayload());
         }
 
-        batchWriter
-                .addTask(WRITE, msg)
+        final RequestMsg fReq = req;
+        batchWriter.addTask(BatchWriterOperation.Type.WRITE, req)
```
zfrenette, line 295: Note: this should be fReq.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +185,256 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
-      log.trace("read: {}, cacheable: {}", msg.getPayload().getAddresses(), cacheable);
-    }
+    @RequestHandler(type = PayloadCase.READ_LOG_REQUEST)
+    public void read(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        final boolean cacheable = req.getPayload().getReadLogRequest().getCacheResults();
+        final List<Long> addressList = req.getPayload().getReadLogRequest().getAddressList();
 
-    ReadResponse rr = new ReadResponse();
+        if (log.isTraceEnabled()) {
+            log.trace("read: {}, cacheable: {}", cacheable, addressList);
```
zfrenette, line 353: Note: fix the order.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zfrenette: Fixed.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +184,256 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
-        batchWriter.<StreamsAddressResponse>addTask(LOG_ADDRESS_SPACE_QUERY, payloadMsg)
-                .thenAccept(tailsResp -> r.sendResponse(ctx, msg,
-                        CorfuMsgType.LOG_ADDRESS_SPACE_RESPONSE.payloadMsg(tailsResp))
-                )
-                .exceptionally(ex -> {
-                    handleException(ex, ctx, payloadMsg, r);
+    @RequestHandler(type = PayloadCase.LOG_ADDRESS_SPACE_REQUEST)
+    public void handleLogAddressSpaceRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        log.trace("handleLogAddressSpaceRequest[{}]: received a log " +
+                "address space request {}", req.getHeader().getRequestId(), TextFormat.shortDebugString(req));
```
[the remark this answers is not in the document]
zfrenette: It isn't clear to me that all RPCs should have the same logging level. In any case, I've wrapped these with `log.isTraceEnabled()`.

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
```diff
@@ -0,0 +1,308 @@
...
+     * Test that the BatchProcessor successfully handles a WRITE request.
+     */
+    @Test
+    public void testWrite() {
+        LogData logData = getDefaultLogData(0L);
+        RequestMsg request = getRequestMsg(getBasicHeader(false, false),
+                getWriteLogRequestMsg(logData));
+
+        batchProcessor.addTask(BatchWriterOperation.Type.WRITE, request).join();
+        verify(mockStreamLog).append(0L, logData);
```
zfrenette, line 151: I don't think these work as expected for `LogData` types, as they are considered equal if the `clientId` and `threadId` are equal.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +185,270 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
-                    dataCache.put(msg.getPayload().getGlobalAddress(), logData);
-                    r.sendResponse(ctx, msg, CorfuMsgType.WRITE_OK.msg());
-                }, executor)
-                .exceptionally(ex -> {
-                    handleException(ex, ctx, msg, r);
-                    return null;
-                });
+                    dataCache.put(logData.getGlobalAddress(), logData);
+                    r.sendResponse(getResponseMsg(fReq.getHeader(), getWriteLogResponseMsg()), ctx);
+                }, executor).exceptionally(ex -> {
```
WenbinZhu, line 309: NIT: Format the .exceptionally() block to align with .thenRunAsync() as it was.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +185,270 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
+
+        if (log.isTraceEnabled()) {
+            log.trace("handleInspectAddressesRequest[{}]: " +
+                    "addresses {}", req.getHeader().getRequestId(), addresses);
+        }
+
+
+        for (long address : addresses) {
+            try {
+                if (!streamLog.contains(address)) emptyAddresses.add(address);
```
WenbinZhu, line 399: Format the if block to wrap in curly brackets on new lines.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -550,10 +598,39 @@ public static LogUnitServerConfig parse(Map<String, Object> opts) {
...
                     .noVerify((Boolean) opts.get("--no-verify"))
                     .noSync((Boolean) opts.get("--no-sync"))
                     .build();
         }
     }
+
+    /**
+     * Utility class used by the LogUnit server.
+     */
+    public static class LogUnitServerInitializer {
```
WenbinZhu, line 611: I guess this class is for testing purpose only? If true, please mention this in the Javadoc.

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
```diff
@@ -0,0 +1,308 @@
...
+                .getTailRequest()
+                .getStreamList()
+                .stream()
+                .map(CorfuProtocolCommon::getUUID)
+                .collect(Collectors.toList());
+
+        when(mockStreamLog.getTails(any())).thenReturn(new TailsResponse(0L));
+        Object ret = batchProcessor.addTask(BatchWriterOperation.Type.TAILS_QUERY, request).join();
+        verify(mockStreamLog).getTails(list);
+        assertTrue(ret instanceof TailsResponse);
```
WenbinZhu, line 214: I think we usually use this in unit tests:  `assertThat(ret).isInstanceOf(TailsResponse.class);`

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -0,0 +1,347 @@
...
+import org.corfudb.runtime.proto.LogData.ReadResponseMsg;
+import org.corfudb.util.serializer.Serializers;
+
+import java.nio.ByteBuffer;
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import static org.corfudb.protocols.CorfuProtocolCommon.*;
```
WenbinZhu, line 35: Expand imports.

WenbinZhu (commented):
First batch of comments.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    /**
+     * Returns the Protobuf representation of a LogData's
+     * metadata map.
+     */
+    private static LogUnitMetadataMsg getLogUnitMetadataMsg(EnumMap<LogUnitMetadataType, Object> metadata) {
+        LogUnitMetadataMsg.Builder metadataMsgBuilder = LogUnitMetadataMsg.newBuilder();
+
+        metadata.forEach((type, obj)  -> {
```
WenbinZhu, line 71: NIT: extra space before "->"

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLogUnit.java
```diff
@@ -0,0 +1,539 @@
...
+                .build();
+    }
+
+    /**
+     * Returns a TAILS request of type STREAM_TAILS that can be sent by the client.
+     *
+     * @param streamIds  the list of stream IDs
+     * @return           a RequestPayloadMsg containing the TAILS request
+     */
+    public static RequestPayloadMsg getTailRequestMsg(List<UUID> streamIds) {
```
WenbinZhu, line 292: Where is this method being used? Looks like only the one above this one, i.e. getTailRequestMsg with the type as parameter is being used. I think you can remove this one.

WenbinZhu (commented):
Please take a look at the TODOs in CorfuProtocolLogData.java and see if they need to be addressed now. Also please take a look at my comments in this file, I think there are some problems with LogData serialization.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
chetangudisagar: Have a look at the Javadoc of `SequencerInitializer` (previously `SequencerFactoryHelper`) in `SequencerServer.java`. You can copy the relevant info if required.

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
```diff
@@ -0,0 +1,308 @@
...
+import static org.corfudb.protocols.service.CorfuProtocolBase.getSealRequestMsg;
+import static org.corfudb.protocols.service.CorfuProtocolLogUnit.getLogAddressSpaceRequestMsg;
+import static org.corfudb.protocols.service.CorfuProtocolLogUnit.getRangeWriteLogRequestMsg;
+import static org.corfudb.protocols.service.CorfuProtocolLogUnit.getResetLogUnitRequestMsg;
+import static org.corfudb.protocols.service.CorfuProtocolLogUnit.getTailRequestMsg;
+import static org.corfudb.protocols.service.CorfuProtocolLogUnit.getTrimLogRequestMsg;
+import static org.corfudb.protocols.service.CorfuProtocolLogUnit.getWriteLogRequestMsg;
+import static org.corfudb.protocols.service.CorfuProtocolMessage.getHeaderMsg;
+import static org.corfudb.protocols.service.CorfuProtocolMessage.getRequestMsg;
+import static org.corfudb.runtime.proto.service.CorfuMessage.*;
```
chetangudisagar, line 44: This unused import can be removed I think.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +101,149 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
-                } else if (currOp.getMsg().getEpoch() != sealEpoch) {
-                    log.warn("batchWriteProcessor: wrong epoch on {} msg, seal epoch is {}, and msg epoch is {}",
-                            currOp.getType(), sealEpoch, currOp.getMsg().getEpoch());
-                    currOp.getFutureResult().completeExceptionally(new WrongEpochException(sealEpoch));
-                    res.add(currOp);
-                    processed++;
-                    lastOp = currOp;
+                } else if (streamLog.quotaExceeded() &&
+                        (currentOp.getRequest().getHeader().getPriority() != CorfuMessage.PriorityLevel.HIGH)) {
+                    currentOp.getFutureResult().completeExceptionally(
```
chetangudisagar, line 161: NIT: Can we add an on-demand import statement for `CorfuMessage.PriorityLevel`?

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
xcchang: Which one was unused?

chetangudisagar: `import static org.corfudb.runtime.proto.service.CorfuMessage.*;`
this one?

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -0,0 +1,347 @@
...
+     */
+    private static LogDataMsg getLogDataEntryMsg(LogData logData) {
+        if (!logData.isData()) {
+            throw new IllegalArgumentException("Incorrect LogData type provided: " + logData.getType().toString());
+        }
+
+        LogDataEntryMsg.Builder entryMsgBuilder = LogDataEntryMsg.newBuilder();
+
+        //TODO(Zach): Optimize and cleanup -- Concurrency bug?
+        //TODO(Zach): Reduce number of buffer copies
```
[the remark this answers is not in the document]
zfrenette: Yes they are. This whole class is still under adjustments.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -111,20 +117,27 @@
...
      */
     public LogUnitServer(ServerContext serverContext) {
+        this(serverContext, new LogUnitServerInitializer());
+    }
+
+    /**
+     * Returns a new LogUnitServer.
+     *
+     * @param serverContext      context object providing settings and objects
+     * @param serverInitializer  a LogUnitServerInitializer object
```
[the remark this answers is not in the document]
zfrenette: Added additional details.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zfrenette: Expanded the Javadoc to include this information.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +101,149 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
 
         try {
             BatchWriterOperation lastOp = null;
-            int processed = 0;
             List<BatchWriterOperation> res = new LinkedList<>();
+            int numProcessed = 0;
 
             while (true) {
-                BatchWriterOperation currOp;
-                queueSizeDist.ifPresent(dist -> dist.record(operationsQueue.size()));
```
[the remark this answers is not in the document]
zfrenette: Good catch. This was accidental.

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
zfrenette: This has been resolved in an offline discussion with @xcchang

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -12,50 +12,48 @@
...
 
     private ExecutorService processorService = Executors
             .newSingleThreadExecutor(new ThreadFactoryBuilder()
                     .setDaemon(false)
                     .setNameFormat("LogUnit-BatchProcessor-%d")
                     .build());
+
     private final Optional<Timer> writeRecordTimer;
     private final Optional<Timer> writeRecordsTimer;
     private final Optional<DistributionSummary> queueSizeDist;
```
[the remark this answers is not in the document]
zfrenette: These appeared unused because the metrics were accidentally removed. See https://github.com/CorfuDB/CorfuDB/pull/2836/files#r547489298.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -0,0 +1,347 @@
...
+
+        if (logData.getData() != null) {
+            entryMsgBuilder.setData(ByteString.copyFrom(logData.getData()));
+        } else {
+            ByteBuf serializedBuf = Unpooled.buffer();
+            Serializers.CORFU.serialize(logData.getPayload(null), serializedBuf);
```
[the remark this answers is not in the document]
Maithem: LogData objects are usually long lived and should probably be lightweight, using a protobuf representation will probably increase memory overhead. I think a protobuf representation would be much more "bloated", so I suggest keeping the current representation.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -111,20 +117,27 @@
...
+     */
+    public LogUnitServer(ServerContext serverContext, LogUnitServerInitializer serverInitializer) {
```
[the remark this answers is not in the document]
xcchang: Changed in the other PR, thanks!

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
zfrenette: Done.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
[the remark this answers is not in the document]
zfrenette: Thank you both for your comments and suggestions. I've gone with the latter approach, and no longer have separate definitions for the metadata. The entire entry is now obtained from LogData's serialization methods.

infrastructure/src/test/java/org/corfudb/infrastructure/batchprocessor/BatchProcessorTest.java
zfrenette: Done.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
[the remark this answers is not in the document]
zfrenette: These have all been updated.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLogUnit.java
zfrenette: It's only being used in some of the unit tests.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogDataSerializerUtils.java
```diff
@@ -0,0 +1,197 @@
...
+
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Utility class used by LogData to help serialize its contents.
+ */
+public class LogDataSerializerUtils {
```
zfrenette, line 18: Note: since `ICorfuPayload` is being deprecated, this class provides the helper methods necessary to serialize and serialize `LogData` components.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java
```diff
@@ -45,233 +48,265 @@ public LogUnitClient getClient(long epoch, UUID clusterID) {
...
+     * @param r A reference to the router.
+     * @return Always True, since the reset log unit was successful.
      */
-    @ClientHandler(type = CorfuMsgType.LOG_ADDRESS_SPACE_RESPONSE)
-    private static Object handleStreamsAddressResponse(CorfuPayloadMsg<TailsResponse> msg,
-                                             ChannelHandlerContext ctx, IClientRouter r) {
-        return msg.getPayload();
+    @ResponseHandler(type = PayloadCase.RESET_LOG_UNIT_RESPONSE)
+    private static Object handleResetLogUnitResponse(ResponseMsg msg, ChannelHandlerContext ctx, IClientRouter r) {
+        return true;
```
xcchang, line 239: nit: Please align the indent for `r` here and the methods above in the javadoc.

infrastructure/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -0,0 +1,814 @@
...
+     * with default values.
+     * @param ignoreClusterId   indicates if the message is clusterId aware
+     * @param ignoreEpoch       indicates if the message is epoch aware
+     * @return                  the corresponding HeaderMsg
+     */
+    private HeaderMsg getBasicHeader(boolean ignoreClusterId, boolean ignoreEpoch) {
+        return getHeaderMsg(requestCounter.incrementAndGet(), PriorityLevel.NORMAL, 1L,
+                getUuidMsg(DEFAULT_UUID), getUuidMsg(DEFAULT_UUID), ignoreClusterId, ignoreEpoch);
+    }
+
```
xcchang, line 114: nit: Adding an empty line between the description and params? Same for the other methods in this file I thought.

xcchang (commented):
Looks great! Just a few comments for minor changes.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -40,22 +41,21 @@
...
-    private ExecutorService processorService = Executors
+    private final ExecutorService processorService = Executors
             .newSingleThreadExecutor(new ThreadFactoryBuilder()
                     .setDaemon(false)
                     .setNameFormat("LogUnit-BatchProcessor-%d")
                     .build());
+
     private final Optional<Timer> writeRecordTimer;
     private final Optional<Timer> writeRecordsTimer;
     private final Optional<DistributionSummary> queueSizeDist;
```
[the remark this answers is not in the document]
zfrenette: The first timer is meant to be used for regular writes, while the second for range writes. @PavelZaytsev is in the process of addressing this in another patch.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +102,152 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
+                        for (BatchWriterOperation op : res) {
+                            if (!op.getFutureResult().isCompletedExceptionally()
+                                    && !op.getFutureResult().isCancelled()) {
+                                op.getFutureResult().complete(op.getResultValue());
                             }
                         }
+
                         res.clear();
-                        processed = 0;
+                        numProcessed = 0;
```
Maithem, line 152: This loop is really messy and probably should be improved in another PR. 

The performance of this loop can be improved by using `ArrayList` instead of `LinkedList`. It produces less garbage and should reduce gc pressure.

Maithem: `numProcessed` can also be removed and replaced with `list.size()`

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +102,152 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
-                    processed++;
-                    lastOp = currOp;
+                } else if (streamLog.quotaExceeded() &&
+                        (currentOp.getRequest().getHeader().getPriority() != PriorityLevel.HIGH)) {
+                    currentOp.getFutureResult().completeExceptionally(
+                            new QuotaExceededException("Quota of " + streamLog.quotaLimitInBytes() + " bytes"));
+
+                    log.warn("batchWriteProcessor: quota exceeded, dropping request {}",
+                            TextFormat.shortDebugString(currentOp.getRequest()));
+                } else if (currentOp.getType() == BatchWriterOperation.Type.SEAL &&
```
Maithem, line 169: I think we can remove `BatchWriterOperation.Type` since all the message types map to an operation in this processor, except for shutdown, which can be encoded as a variable instead of a message in the work queue.

Maithem: Then you can check and map the message type directly. For example, `if (currentOp.getRequest().getPayload().hasSealRequest())`

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +102,152 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
+                    numProcessed++;
+                    lastOp = currentOp;
+                } else if (currentOp.getRequest().getHeader().getEpoch() != sealEpoch) {
+                    log.warn("batchWriteProcessor: wrong epoch on {} request, seal epoch is {}, and request epoch is {}",
+                            currentOp.getType(), sealEpoch, currentOp.getRequest().getHeader().getEpoch());
+
+                    currentOp.getFutureResult().completeExceptionally(new WrongEpochException(sealEpoch));
+                    res.add(currentOp);
+                    numProcessed++;
+                    lastOp = currentOp;
```
Maithem, line 185: The point of batch writing is to amortize the cost of the fsync, but in this case the request is rejected and doesn't produce any io and thus shouldn't contribute to the batch size.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +102,152 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
-                                streamLog.prefixTrim(prefixTrim.getAddress().getSequence());
+                                final long addr = payload.getTrimLogRequest().getAddress().getSequence();
+                                streamLog.prefixTrim(addr);
                                 break;
                             case WRITE:
-                                WriteRequest write = (WriteRequest) currOp.getMsg().getPayload();
-                                Runnable append =
-                                        () -> streamLog.append(write.getGlobalAddress(), (LogData) write.getData());
+                                LogData logData = getLogData(payload.getWriteLogRequest().getLogData());
+                                Runnable append = () -> streamLog.append(logData.getGlobalAddress(), logData);
```
Maithem, line 196: Can WriteLogRequestMsg and RangeWriteLogRequestMsg be collapsed ?

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -90,7 +94,9 @@ public Integer getPort() {
         wr.setRank(rank);
         wr.setBackpointerMap(backpointerMap);
         wr.setGlobalAddress(address);
-        CompletableFuture<Boolean> cf = sendMessageWithFuture(CorfuMsgType.WRITE.payloadMsg(wr));
+        CompletableFuture<Boolean> cf = sendRequestWithFuture(
+                getWriteLogRequestMsg((LogData) wr.getData()), false, false);
+
```
Maithem, line 99: A `WriteRequest` is constructed, but its not being used when the message is sent. I would recommend eliminating that type and directly using the LogData object.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +102,152 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
+
+                        currentOp.getFutureResult().completeExceptionally(e);
                     }
-                    res.add(currOp);
 
-                    processed++;
-                    lastOp = currOp;
+                    res.add(currentOp);
+                    numProcessed++;
+                    lastOp = currentOp;
```
Maithem, line 250: Error handling logic here is not correct. If an exception happens within a batch then all requests should be errored out because the write/fsync won't guarantee any boundaries.

Maithem: Actually, IO errors are fatal in this cause and the whole LogUnit should be re-initialized.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/CorfuMsgType.java
```diff
@@ -31,42 +31,6 @@
     // Layout Messages
     LAYOUT_NOBOOTSTRAP(19, TypeToken.of(CorfuMsg.class), true, false),
```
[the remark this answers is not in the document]
xcchang: It might not worth the effort to remove this type at this time, might be better to remove when the LR is supported in Protobuf (i.e. at the time we don't need CorfuMsg at all)

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -550,10 +596,41 @@ public static LogUnitServerConfig parse(Map<String, Object> opts) {
...
         }
     }
+
+    /**
+     * Utility class used by the LogUnit server to initialize its components,
+     * including the StreamLog, LogUnitServerCache, BatchProcessor and StreamLogCompaction.
+     * This facilitates the injection of mocked objects during unit tests.
+     */
+    public static class LogUnitServerInitializer {
+        StreamLog buildStreamLog() {
```
Maithem, line 612: the naming can be improved here, consider renaming it to `buildInMemoryStreamLog`

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -158,15 +165,19 @@ protected void processRequest(RequestMsg req, ChannelHandlerContext ctx, IServer
...
-                })
+    @RequestHandler(type = PayloadCase.TAIL_REQUEST)
+    public void handleTailRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        log.debug("handleTailRequest[{}]: received a tail request {}",
+                req.getHeader().getRequestId(), TextFormat.shortDebugString(req));
+
+        batchWriter.<TailsResponse>addTask(BatchWriterOperation.Type.TAILS_QUERY, req)
+                .thenAccept(tailsResp ->
+                    // Note: we reuse the request header as the ignore_cluster_id and
+                    // ignore_epoch fields are the same in both cases.
```
Maithem, line 176: This is incorrect. You can get receive a message on epoch 1, then by the time you reply the server is on epoch 2. The client should be able to detect that. The message can be queued internally while the epoch of the router changes.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +186,272 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_OVERWRITE
-                    .payloadMsg(owe.getOverWriteCause().getId()));
-        } else if (ex.getCause() instanceof DataOutrankedException) {
-            r.sendResponse(ctx, msg, CorfuMsgType.ERROR_DATA_OUTRANKED.msg());
-        } else if (ex.getCause() instanceof ValueAdoptedException) {
+            responseHeader = getHeaderMsg(req.getHeader(), false, true);
+            r.sendResponse(getResponseMsg(responseHeader, getOverwriteErrorMsg(owe.getOverWriteCause().getId())), ctx);
+        } else if (ex.getCause() instanceof  DataOutrankedException) {
+            responseHeader = getHeaderMsg(req.getHeader(), false, false);
+            r.sendResponse(getResponseMsg(responseHeader, getDataOutrankedErrorMsg()), ctx);
```
Maithem, line 274: Since the same request header is used for the response header these checks won't make a  lot of sense, right?

Imagine the client is on epoch 1 and sends a message to the server and the server transitions to epoch 2 just before it sends out the response. When the client receives the response it will be on epoch 1, but ignoreEpoch=false would mean that the client should have raised an exception. In this case it wont.

Maithem: It's a semantic change from the current code base. There is an difference between sending back the wrong epoch and saying that the return epoch value has no effect on the final result. That difference should be explicitly encoded.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +186,272 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
-    @ServerHandler(type = CorfuMsgType.WRITE)
-    public void write(CorfuPayloadMsg<WriteRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        LogData logData = (LogData) msg.getPayload().getData();
-        log.debug("log write: type: {}, address: {}, streams: {}", logData.getType(),
-                logData.getToken(), logData.getBackpointerMap());
+    @RequestHandler(type = PayloadCase.WRITE_LOG_REQUEST)
+    public void handleWrite(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        LogData logData = getLogData(req.getPayload().getWriteLogRequest().getLogData());
+        log.debug("handleWrite: type: {}, address: {}, streams: {}",
+                logData.getType(), logData.getToken(), logData.getBackpointerMap());
```
Maithem, line 296: Is it better to name it `LogWriteRequest` vs `WriteLogRequest?` . Not sure if there is a convention about verb/noun vs noun/verb

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +186,272 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
         // some reads will block until a hole is filled this is required (i.e. bypass quota checks)
         // because the requirement is to allow reads, but only block writes once the quota is exhausted
         if (logData.isHole()) {
-            msg.setPriorityLevel(PriorityLevel.HIGH);
+            req = getRequestMsg(getHighPriorityHeaderMsg(req.getHeader()), req.getPayload());
         }
 
-        batchWriter
-                .addTask(WRITE, msg)
+        final RequestMsg fReq = req;
```
Maithem, line 305: This is strange, why is it needed?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +186,272 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
 
-    for (long address : msg.getPayload().getAddresses()) {
-      try {
-        ILogData logData = dataCache.get(address, cacheable);
-        if (logData == null) {
-          rr.put(address, LogData.getEmpty(address));
-        } else {
-          rr.put(address, (LogData) logData);
+        ReadResponse rr = new ReadResponse();
+
```
Maithem, line 369: This is not needed, you just unwrap it to `List<LogData> addresses`  and then construct a protobuf from that. Since a lot of quorum replication relies on that type, it would be more work to remove the `ReadResponse` class .

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -175,231 +186,272 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
      * Handles requests for known entries in specified range.
      * This is used by state transfer to catch up only the remainder of the segment.
      */
-    @ServerHandler(type = CorfuMsgType.KNOWN_ADDRESS_REQUEST)
-    private void getKnownAddressesInRange(CorfuPayloadMsg<KnownAddressRequest> msg,
-                                          ChannelHandlerContext ctx, IServerRouter r) {
-
-        KnownAddressRequest request = msg.getPayload();
+    @RequestHandler(type = PayloadCase.KNOWN_ADDRESS_REQUEST)
+    private void handleKnownAddressRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
```
Maithem, line 422: This handler and handleInspectAddressesRequest seem to be redundant. We should remove one of them.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -411,23 +463,26 @@ private void handleFlushCacheRequest(CorfuMsg msg, ChannelHandlerContext ctx, IS
...
-        msg.setEpoch(epoch);
-        msg.setPriorityLevel(PriorityLevel.HIGH);
+        RequestMsg batchProcessorMsg = getRequestMsg(
+                getHighPriorityHeaderMsg(HeaderMsg.newBuilder().setEpoch(epoch).build()),
+                getSealRequestMsg(epoch)
+        );
+
         try {
-            batchWriter.addTask(SEAL, msg).join();
+            batchWriter.addTask(BatchWriterOperation.Type.SEAL, batchProcessorMsg).join();
```
Maithem, line 472: We should have a uniform way to propagate these requests. For example `sealServerWithEpoch` should be invoked with a seal message instead of only a long.

infrastructure/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
xcchang: Done

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java
xcchang: Done

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -40,22 +41,21 @@
...
-    final private StreamLog streamLog;
+    private final int BATCH_SIZE = 50;
+    private final boolean sync;
+    private final StreamLog streamLog;
+    private final BlockingQueue<BatchWriterOperation> operationsQueue;
 
-    final private BlockingQueue<BatchWriterOperation> operationsQueue;
-
-    private ExecutorService processorService = Executors
+    private final ExecutorService processorService = Executors
```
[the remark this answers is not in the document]
xcchang: Done

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLogUnit.java
```diff
@@ -0,0 +1,539 @@
...
+     *
+     * @param msg  the desired Protobuf ReadLogResponse message
+     * @return     an equivalent ReadResponse object
+     */
+    public static ReadResponse getReadResponse(ReadLogResponseMsg msg) {
+        ReadResponse rr = new ReadResponse();
+        msg.getResponseList().forEach(e -> rr.put(e.getAddress(), getLogData(e.getLogData())));
+
+        return rr;
+    }
```
Maithem, line 115: Why does the response return another response? It should unwrap the resulted LogData and return that. No point in propagated the wrappers.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java
```diff
@@ -45,233 +48,265 @@ public LogUnitClient getClient(long epoch, UUID clusterID) {
...
+     * @return {@link InspectAddressesResponse} sent back from server.
      */
-    @ClientHandler(type = CorfuMsgType.ERROR_DATA_OUTRANKED)
-    private static Object handleDataOutranked(CorfuMsg msg,
-                                              ChannelHandlerContext ctx, IClientRouter r)
-            throws Exception {
-        throw new DataOutrankedException();
+    @ResponseHandler(type = PayloadCase.INSPECT_ADDRESSES_RESPONSE)
+    private static Object handleInspectResponse(ResponseMsg msg, ChannelHandlerContext ctx, IClientRouter r) {
+        return getInspectAddressesResponse(msg.getPayload().getInspectAddressesResponse());
```
Maithem, line 104: Most of these handlers should return the actual value, not the wrapped values. That is, LogData and not ReadResponse (which wraps LogData) etc.,

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
```diff
@@ -114,127 +102,152 @@ private void recordRunnable(Runnable fsyncRunnable, Optional<Timer> fsyncTimer)
...
-                                    case TailsRequest.STREAMS_TAILS:
-                                        tails = streamLog.getTails(tailsRequest.getStreams());
+                                    case STREAMS_TAILS:
+                                        tails = streamLog.getTails(currentOp.getRequest()
+                                                .getPayload()
+                                                .getTailRequest()
+                                                .getStreamList()
+                                                .stream()
+                                                .map(CorfuProtocolCommon::getUUID)
+                                                .collect(Collectors.toList()));
```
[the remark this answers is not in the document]
Maithem: Good catch. It seems like this should be `ALL_STREAMS_TAIL`. This should be a bug, not sure how the tests are not failing. On reconfiguration, the sequencer won't be configured unless `LogUnitClient::getAllTails` is called, and that won't construct a STREAMS_TAILS type.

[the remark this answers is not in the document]
Maithem: I see. Its not failing because its hitting the default cause which returns all stream tails. I would recommend throwing an exception in the default case and explicitly specifying `ALL_STREAMS_TAIL`

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * This class provides methods for creating and converting between the Protobuf
+ * objects defined in log_data.proto and their Java counterparts. Used by the
+ * LogUnit RPCs.
+ */
+@Slf4j
+public class CorfuProtocolLogData {
+    // Prevent class from being instantiated
+    private CorfuProtocolLogData() {}
+
```
Maithem, line 20: You can mark the class as final to prevent it from being inherited as well since its just a utils class.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -0,0 +1,59 @@
...
+     * @return         an equivalent Protobuf LogData message
+     */
+    public static LogDataMsg getLogDataMsg(LogData logData) {
+        ByteBuf buf = Unpooled.buffer();
+        logData.doSerialize(buf);
+
+        return LogDataMsg.newBuilder()
+                .setEntry(ByteString.copyFrom(buf.resetReaderIndex().array()))
+                .build();
+    }
```
Maithem, line 34: Too many array copies :p. Maybe we should use pooling. I think we need to address some possible performance issues in subsequent patches.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Returns a LogData object from its Protobuf representation.
+     *
+     * @param msg  the desired Protobuf LogData message
+     * @return     a equivalent LogData object
+     */
+    public static LogData getLogData(LogDataMsg msg) {
+        return new LogData(Unpooled.wrappedBuffer(msg.getEntry().toByteArray()));
+    }
```
Maithem, line 44: I think `new LogData(Unpooled.wrappedBuffer(msg.getEntry().asReadOnlyByteBuffer()));` is cheaper than `toByteArray` since it will eliminate one allocation and one copy.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogDataSerializerUtils.java
```diff
@@ -0,0 +1,211 @@
...
+
+    @FunctionalInterface
+    interface Serializer {
+        void serialize(ByteBuf buffer, Object payload);
+    }
+
+    static ConcurrentHashMap<Class<?>, PayloadConstructor<?>> constructorMap = new ConcurrentHashMap<>(
+            ImmutableMap.<Class<?>, PayloadConstructor<?>>builder()
+                    .put(Byte.class, ByteBuf::readByte)
+                    .put(Integer.class, ByteBuf::readInt)
```
Maithem, line 35: Shouldn't this be an immutable map that is created once? Why do we add more types at runtime?

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogDataSerializerUtils.java
```diff
@@ -0,0 +1,211 @@
...
+            return (T) constructorMap.get(cls).construct(buf);
+        } else if (cls.isEnum()) {
+            // We only know how to deal with enums with a type map
+            try {
+                Map<Byte, T> typeMap = (Map<Byte, T>) cls.getDeclaredField("typeMap").get(null);
+                constructorMap.put(cls, x -> typeMap.get(x.readByte()));
+                return (T) constructorMap.get(cls).construct(buf);
+            } catch (NoSuchFieldException e) {
+                throw new RuntimeException("Only enums with a typeMap are supported!");
+            } catch (IllegalAccessException e) {
```
Maithem, line 128: All this reflection is expensive to do per LogData. I think we need to revisit this later and just inline the metadata into the protobuf type.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WriteRequest.java
```diff
@@ -12,25 +12,15 @@
  */
 @Builder
 @AllArgsConstructor
-public class WriteRequest implements ICorfuPayload<WriteRequest>, IMetadata {
+public class WriteRequest implements IMetadata {
 
     @Getter
```
Maithem, line 17: You can just eliminate this class and Just use the LogData object directly.

Maithem (approved):
Looks great! Just some minor comments and questions.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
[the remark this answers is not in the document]
zfrenette: I think that this means we can remove the `STREAMS_TAILS` case entirely, since this is the only place where it's referenced, and the client never sends this type of message.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogDataSerializerUtils.java
```diff
@@ -0,0 +1,197 @@
...
+            buffer.writeByte((Byte) payload);
+        } else if (payload instanceof Long) {
+            buffer.writeLong((Long) payload);
+        } else if (payload instanceof byte[]) {
+            buffer.writeInt(((byte[]) payload).length);
+            buffer.writeBytes((byte[]) payload);
+        } else if (payload instanceof UUID) {
+            serialize(buffer, ((UUID) payload).getMostSignificantBits());
+            serialize(buffer, ((UUID) payload).getLeastSignificantBits());
+        } else if (payload instanceof EnumMap) {
```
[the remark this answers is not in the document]
xcchang: We had an offline discussion and decide to keep the current implementation.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xcchang: I thought it's the effectively final constraint enforced in the lambda expressions.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
xcchang: Looks like we should mark all the CorfuProtocol*.java files to be final?

Maithem:
Can you rebase on https://github.com/CorfuDB/CorfuDB/pull/2855 ? I removed a lot of unfunctional quorum replication code. This also means you can drop those definitions from the proto schemas.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
zfrenette: Done.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/WriteRequest.java
zfrenette: Done.

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
xcchang: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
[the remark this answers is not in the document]
xcchang: Removed STREAMS_TAILS and related codes, thanks!

test/src/test/java/org/corfudb/infrastructure/LogUnitServerTest.java
```diff
@@ -524,77 +524,6 @@ public void testVerifyWithNoVerifyLog() throws Exception {
...
-        ServerContext sc = new ServerContextBuilder()
-                .setLogPath(serviceDir)
-                .setSingle(true)
-                .setMemory(false)
-                .build();
-
-        sc.installSingleNodeLayoutIfAbsent();
-        sc.setServerRouter(router);
-        sc.setServerEpoch(sc.getCurrentLayout().getEpoch(), router);
-
```
xcchang: Should we just remove this test since we don't have Rank now?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zfrenette: With PR [2855](https://github.com/CorfuDB/CorfuDB/pull/2855), this message type is no longer needed and has been removed.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLogUnit.java
zfrenette: Note: this has been [removed](https://github.com/CorfuDB/CorfuDB/pull/2836#discussion_r556715660).

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
zfrenette: I've replaced the `LinkedList` by an `ArrayList` and have removed `numProcessed`.

infrastructure/src/main/java/org/corfudb/infrastructure/BatchProcessor.java
zfrenette: Done.

Maithem:
@xcchang Were you able to fix the test that was failing intermittently?

xcchang:
> @xcchang Were you able to fix the test that was failing intermittently?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -173,226 +182,265 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
      */
-    @ServerHandler(type = CorfuMsgType.UPDATE_COMMITTED_TAIL)
-    public void updateCommittedTail(CorfuPayloadMsg<Long> msg,
-                                    ChannelHandlerContext ctx, IServerRouter r) {
-        log.trace("updateCommittedTail: received request to update committed tail {}", msg);
-        streamLog.updateCommittedTail(msg.getPayload());
-        r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+    @RequestHandler(type = PayloadCase.UPDATE_COMMITTED_TAIL_REQUEST)
+    public void handleUpdateCommittedTailRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        if (log.isTraceEnabled()) {
```
vjeko, line 239: Instead of introducing an if statement, you can simply emulate lazy-evaluation:

```
    interface LazyString {
        String toString();

        static LazyString create(MessageOrBuilder message) {
            return new LazyString() {
                @Override
                public String toString() {
                    return TextFormat.shortDebugString(message);
                }
            };
        }
    }
```
Which will enable you to do:

```
log.trace("handleLogAddressSpaceRequest[{}]: received a log " +
        "address space request {}", req.getHeader().getRequestId(), LazyString.create(req));
```

Alternatively, we can upgrade `slf4j` to include the fluent API, which would allow you to pass a `Supplier`. However, that is a riskier change.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -156,15 +161,19 @@ protected void processRequest(RequestMsg req, ChannelHandlerContext ctx, IServer
...
-    public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.debug("handleTailRequest: received a tail request {}", msg);
-        batchWriter.<TailsResponse>addTask(TAILS_QUERY, msg)
-                .thenAccept(tailsResp -> {
-                    r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp));
-                })
+    @RequestHandler(type = PayloadCase.TAIL_REQUEST)
+    public void handleTailRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        log.debug("handleTailRequest[{}]: received a tail request {}",
+                req.getHeader().getRequestId(), TextFormat.shortDebugString(req));
```
vjeko, line 167: I assume you wanted to guard this statement as well.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -173,226 +182,265 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
-        r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
+    @RequestHandler(type = PayloadCase.UPDATE_COMMITTED_TAIL_REQUEST)
+    public void handleUpdateCommittedTailRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        if (log.isTraceEnabled()) {
+            log.trace("handleUpdateCommittedTailRequest: received request to "
+                    + "update committed tail {}", TextFormat.shortDebugString(req));
+        }
+
+        streamLog.updateCommittedTail(req.getPayload().getUpdateCommittedTailRequest().getCommittedTail());
+        HeaderMsg responseHeader = getHeaderMsg(req.getHeader(), false, true);
```
vjeko, line 245: I actually think that this method does nor provide much value. The problem I have with it is that arguments two and three are not type safe, so it is pretty easy to mess us the ordering. When using the `Builder` directly, it is much harder, since the parameter names are referenced explicitly.

vjeko: Alternatively, you can defines types for each of the paramets:

```
    public enum ClusterId {
        CHECK,
        IGNORE
    }
    
    public enum Epoch {
        CHECK,
        IGNORE
    }
```
And change you method implementation:
```
    public static HeaderMsg getHeaderMsg(HeaderMsg header, ClusterId ignoreClusterId, Epoch ignoreEpoch) {
        return HeaderMsg.newBuilder()
                .mergeFrom(header)
                .setIgnoreClusterId(ignoreClusterId == ClusterId.IGNORE)
                .setIgnoreEpoch(ignoreEpoch == Epoch.IGNORE)
                .build();
    }
```
So a call to that function would become: `getHeaderMsg(req.getHeader(), ClusterId.IGNORE, Epoch.IGNORE);`

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -173,226 +182,265 @@ public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerC
...
+        LogData logData = getLogData(req.getPayload().getWriteLogRequest().getLogData());
+        log.debug("handleWrite: type: {}, address: {}, streams: {}",
+                logData.getType(), logData.getToken(), logData.getBackpointerMap());
 
         // Its not clear that making all holes high priority is the right thing to do, but since
         // some reads will block until a hole is filled this is required (i.e. bypass quota checks)
         // because the requirement is to allow reads, but only block writes once the quota is exhausted
         if (logData.isHole()) {
-            msg.setPriorityLevel(PriorityLevel.HIGH);
+            req = getRequestMsg(getHighPriorityHeaderMsg(req.getHeader()), req.getPayload());
```
vjeko, line 291: I think it would be much cleaner to user the `Builder directly. Also, you should avoid shadowing `req` argument. How about this?

```        
        HeaderMsg.Builder responseHeader = HeaderMsg.newBuilder().mergeFrom(req.getHeader());

        // Its not clear that making all holes high priority is the right thing to do, but since
        // some reads will block until a hole is filled this is required (i.e. bypass quota checks)
        // because the requirement is to allow reads, but only block writes once the quota is exhausted
        if (logData.isHole()) {
            responseHeader.setPriority(CorfuMessage.PriorityLevel.HIGH);
        }

        batchWriter.addTask(BatchWriterOperation.Type.WRITE, req)
                .thenRunAsync(() -> {
                    dataCache.put(logData.getGlobalAddress(), logData);
                    r.sendResponse(getResponseMsg(responseHeader.build(), getWriteLogResponseMsg()), ctx);
                }, executor)
                .exceptionally(ex -> {
                    handleException(ex, ctx, req, r);
                    return null;
                });

File: runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java [hunk 60]
```diff
@@ -40,204 +46,237 @@ public LogUnitClient getClient(long epoch, UUID clusterID) {
...
+     * @return Always True, since the trim log was successful.
      */
-    @ClientHandler(type = CorfuMsgType.ERROR_RANK)
-    private static Object handleOutranked(CorfuMsg msg, ChannelHandlerContext ctx, IClientRouter r)
-            throws Exception {
-        throw new Exception("rank");
+    @ResponseHandler(type = PayloadCase.TRIM_LOG_RESPONSE)
+    private static Object handleTrimLogResponse(ResponseMsg msg, ChannelHandlerContext ctx, IClientRouter r) {
+        return true;
     }
```
vjeko, line 116: Can't the handler be of return type `void`?

runtime/src/main/java/org/corfudb/runtime/clients/NettyClientRouter.java
```diff
@@ -292,7 +292,9 @@ public IClientRouter addClient(IClient client) {
                         log.trace("Registered {} to handle messages of type {}", client, x);
                     });
         } catch (UnsupportedOperationException ex) {
-            log.error("No registered CorfuMsg handler for client {}", client, ex);
+            if (log.isTraceEnabled()) {
```
vjeko, line 295: No need for it.

test/src/test/java/org/corfudb/runtime/clients/TestClientRouter.java
```diff
@@ -203,10 +203,11 @@ public IClientRouter addClient(IClient client) {
                         log.trace("Registered {} to handle messages of type {}", client, x);
                     });
         } catch (UnsupportedOperationException ex) {
-            log.error("No registered CorfuMsg handler for client {}", client, ex);
+            if (log.isTraceEnabled()) {
```
vjeko, line 206: Not needed.

test/src/test/java/org/corfudb/infrastructure/TestServerRouter.java
```diff
@@ -127,10 +134,11 @@ public void addServer(AbstractServer server) {
                 log.trace("Registered {} to handle messages of type {}", server, x);
             });
         } catch (UnsupportedOperationException ex) {
-            log.error("No registered CorfuMsg handler for server {}", server, ex);
+            if (log.isTraceEnabled()) {
```
vjeko, line 137: Not needed.

vjeko (approved):
LGTM. Please try to address the comments.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitHandler.java
xcchang: No, they are supposed to match the FunctionalInterface specified in `ClientResponseHandler`

infrastructure/src/main/java/org/corfudb/infrastructure/NettyServerRouter.java
```diff
@@ -76,7 +76,9 @@ public NettyServerRouter(ImmutableList<AbstractServer> servers, ServerContext se
             try {
                 server.getHandler().getHandledTypes().forEach(handledType -> handlerMap.put(handledType, server));
             } catch (UnsupportedOperationException ex) {
-                log.error("No registered CorfuMsg handler for server {}", server, ex);
+                if (log.isTraceEnabled()) {
```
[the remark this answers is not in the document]
xcchang: This was originally from one of Maithem's comments, (https://github.com/CorfuDB/CorfuDB/pull/2838#discussion_r542945405). I'm a little inclined to keep current way to make it consistent, and since this happens in a lot of places, it might be worth adding this check?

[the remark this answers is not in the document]
zfrenette: @xcchang In Maithem's comment, I think the concern was the computation of `TextFormat.shortDebugString`. In this particular instance, we don't have this problem.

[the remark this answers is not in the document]
xcchang: Oh I see now, thanks!

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolCommon.java
```diff
@@ -255,6 +257,28 @@ public static StreamAddressRangeMsg getStreamAddressRangeMsg(StreamAddressRange
...
+        StreamsAddressResponse response = new StreamsAddressResponse(tail,
+                map.stream().collect(Collectors.<UuidToStreamAddressSpacePairMsg, UUID, StreamAddressSpace>toMap(
+                        entry -> getUUID(entry.getStreamUuid()),
+                        entry -> getStreamAddressSpace(entry.getAddressSpace())
+                )));
+
+        response.setEpoch(epoch);
+        return response;
+    }
+
```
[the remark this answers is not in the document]
zfrenette: There's actually an identical method in `CorfuProtocolSequencer`. My original intent was to place this method in `CorfuProtocolCommon` since it is used by both the `Sequencer` and `LogUnit`, but I must have missed this. Given that this can be used by both servers, I think it would be better to leave the implementation in `CorfuProtocolCommon` and remove the references to the one in `CorfuProtocolSequencer`? This would avoid duplication.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zfrenette: I like this approach, but as discussed, changing the `getHeaderMsg` method signatures requires updating all of the servers, clients and tests. I'll address this in a different PR.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zfrenette: The issue here is that the message given to `BatchProcessor` when adding the task is expected to have `PriorityLevel.HIGH` when the corresponding `logData` is a hole. In this snippet, we are changing the priority of the response instead. Although this is fine, it still seems like the request needs to be updated in order to use this new header.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zfrenette: Although this emulates lazy-evaluation, it does not guard against the allocation of an additional object like the if statement would when `trace`/`debug` is disabled. I thinking keeping the if statement is slightly better since it provides both benefits.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -404,23 +452,26 @@ private void handleFlushCacheRequest(CorfuMsg msg, ChannelHandlerContext ctx, IS
...
                 throw (WrongEpochException) ce.getCause();
             }
         }
+
         log.info("LogUnit sealServerWithEpoch: sealed and flushed with epoch {}", epoch);
     }
 
     @Override
-    public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
+    public boolean isServerReadyToHandleMsg(RequestMsg request) {
```
zfrenette, line 474: Note: this method can be removed since this is the default behaviour of `AbstractServer`.

runtime/src/main/java/org/corfudb/runtime/clients/LogUnitClient.java
```diff
@@ -79,10 +86,12 @@ public Integer getPort() {
...
         Serializers.CORFU.serialize(writeObject, payload);
-        WriteRequest wr = new WriteRequest(DataType.DATA, payload);
-        wr.setBackpointerMap(backpointerMap);
-        wr.setGlobalAddress(address);
-        CompletableFuture<Boolean> cf = sendMessageWithFuture(CorfuMsgType.WRITE.payloadMsg(wr));
+        LogData ld = new LogData(DataType.DATA, payload);
+        ld.setBackpointerMap(backpointerMap);
+        ld.setGlobalAddress(address);
+        CompletableFuture<Boolean> cf = sendRequestWithFuture(
+                getWriteLogRequestMsg(ld), false, false);
```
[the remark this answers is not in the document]
zfrenette: As mentioned here https://github.com/CorfuDB/CorfuDB/pull/2836#discussion_r564756308, I'll address this in another PR.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -433,36 +484,33 @@ public boolean isServerReadyToHandleMsg(CorfuMsg msg) {
...
         // Check if the reset request is with an epoch greater than the last reset epoch seen to
-        // prevent multiple reset in the same epoch and should be equal to the current router
+        // prevent multiple reset in the same epoch. and should be equal to the current router
         // epoch to prevent stale reset requests from wiping out the data.
-        if (msg.getPayload() <= serverContext.getLogUnitEpochWaterMark()
-                || msg.getPayload() != serverContext.getServerEpoch()) {
-            log.info("LogUnit Server Reset request received but reset already done.");
-            r.sendResponse(ctx, msg, CorfuMsgType.ACK.msg());
-            return;
-        }
```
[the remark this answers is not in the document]
zfrenette: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
zfrenette: I have used the `Builder` directly to set `PriorityLevel.HIGH`, but still modify the request message to use this new header.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -156,15 +161,19 @@ protected void processRequest(RequestMsg req, ChannelHandlerContext ctx, IServer
...
      */
-    @ServerHandler(type = CorfuMsgType.TAIL_REQUEST)
-    public void handleTailRequest(CorfuPayloadMsg<TailsRequest> msg, ChannelHandlerContext ctx, IServerRouter r) {
-        log.debug("handleTailRequest: received a tail request {}", msg);
-        batchWriter.<TailsResponse>addTask(TAILS_QUERY, msg)
-                .thenAccept(tailsResp -> {
-                    r.sendResponse(ctx, msg, CorfuMsgType.TAIL_RESPONSE.payloadMsg(tailsResp));
-                })
+    @RequestHandler(type = PayloadCase.TAIL_REQUEST)
+    public void handleTailRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
```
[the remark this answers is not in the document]
zfrenette: I have updated this here and in other related locations.

PR #2837, opened 2020-12-01
UFO nested secondary index

annym:
## Overview

Description: support secondary indexes on complex inner paths by using a new synthetic field option of the type string which can specify multiple secondary keys by comma separated values which reflect the full path to the inner object that needs to be indexed.

Support secondary indexes on arrays (repeated fields)

Why should this be merged: required for UFO completion

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -23,43 +26,136 @@
...
+                        (key, val) -> {
+                            // Get all nested fields, '.' separated
+                            String[] nestedFields = indexName.split("\\.");
+
+                            if (nestedFields.length > 0) {
+                                Message subMessage = val.getPayload();
+
+                                for (int i =0; i < nestedFields.length; i++) {
+                                    try {
+                                        subMessage = (Message) subMessage.getField(
```
hisundar, line 58: So this is meant to go 2 levels right, not N levels? Or am I missing something?
2 levels is actually fine too!

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -23,43 +26,136 @@
...
         if (fieldDescriptor.getOptions().getExtension(CorfuOptions.schema).getSecondaryKey()) {
             final String indexName = fieldDescriptor.getName();
             if (fieldDescriptor.getType() == FieldDescriptor.Type.GROUP) {
                 throw new IllegalArgumentException("group is a deprecated, unsupported type");
             }
             indices.put(indexName, getIndex(indexName, fieldDescriptor));
+        } else if (fieldDescriptor.getOptions().getExtension(CorfuOptions.schema).hasNestedSecondaryKeys()) {
+            if (fieldDescriptor.getType() == FieldDescriptor.Type.GROUP) {
+                throw new IllegalArgumentException("group is a deprecated, unsupported type");
+            }
```
hisundar, line 88: Shouldn't we also handle else if (fieldDescriptor.isRepeated()) { for indexes on collections?

hisundar (commented):
Thanks Anny, this change looks good. Couple of quick observations/questions.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
annym: It is bounded to nestedFields.length, so it can actually go to N levels. It will return outside of the for loop, unless a cast exception is encountered and it's in the last level.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
annym: I didn't consider collections, let me work on adding it.

annym: Done, now it supports collections and nested secondary indexes.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -199,6 +201,701 @@ public void testSecondaryIndexes() throws Exception {
...
+        Table<Uuid, ExampleSchemas.ClassRoom, ManagedMetadata> table = shimStore.openTable(
+                someNamespace,
+                tableName,
+                Uuid.class,
+                ExampleSchemas.ClassRoom.class,
+                ManagedMetadata.class,
+                // TableOptions includes option to choose - Memory/Disk based corfu table.
+                TableOptions.builder().build());
+
+        // Create records for 40 students
```
zhangn49, line 332: For 40 Class rooms, right?

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -199,6 +201,701 @@ public void testSecondaryIndexes() throws Exception {
...
+                        user);
+                txn.commit();
+            }
+        }
+
+        // Get by secondary index, retrieve from database all even entries
+        try (ManagedTxnContext readWriteTxn = shimStore.txn(someNamespace)) {
+            List<CorfuStoreEntry<Uuid, ExampleSchemas.ClassRoom, ManagedMetadata>> entries = readWriteTxn
+                    .getByIndex(table, "students.age", youngStudent);
+            assertThat(entries.size()).isEqualTo(totalStudents/2);
```
zhangn49, line 362: totalRooms/2

zhangn49 (approved):

runtime/proto/example_schemas.proto
```diff
@@ -28,4 +28,105 @@ message ExampleValue {
...
+                                                   "nonPrimitiveFieldLevel0.key1Level1, nonPrimitiveFieldLevel0.key2Level1.key1Level2"];
+}
+
+message NonPrimitiveValue {
+    fixed64 key1Level1 = 1;
+    NonPrimitiveNestedValue key2Level1 = 2;
+}
+
+message NonPrimitiveNestedValue {
+    string key1Level2 = 1;
```
hisundar, line 42: can we try a oneOf here based on our meeting with the consumers?

hisundar (approved):
I think this is really great work!

Lujie1996 (approved):

runtime/proto/example_schemas.proto
```diff
@@ -28,4 +28,101 @@ message ExampleValue {
     string payload = 1;
     fixed64 anotherKey = 2 [(org.corfudb.runtime.schema).secondary_key = true];
     Uuid uuid = 3 [(org.corfudb.runtime.schema).secondary_key = true];
+    fixed64 entryIndex = 4;
+    NonPrimitiveValue nonPrimitiveFieldLevel0 = 5 [(org.corfudb.runtime.schema).nested_secondary_keys =
```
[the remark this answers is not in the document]
annym: Good observation, required one change to make it work. But it's not really needed in all fields, right? in any case by changing 3 of them we now have samples of both formats.

runtime/proto/corfu_options.proto
```diff
@@ -17,6 +17,8 @@ message SchemaOptions {
     optional bool is_federated = 4;
     // Tag tables with unique stream listener tags for selectivity in receiving change notifications.
     repeated string stream_tag = 5;
+    // String of comma-separated nested secondary keys
+    optional string nested_secondary_keys = 6;
```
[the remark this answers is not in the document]
annym: I don't think so. I was not part of the design, I was already told of how it would look like, but one reason I could think of is simplification, as copying "[(org.corfudb.runtime.schema).nested_secondary_keys = " multiple times could be annoying. But I don't think there is any other special reason.

PR #2839, opened 2020-12-05
Collect Metrics for JVM Safe Point Stats

Maithem:
## Overview
Excessive JVM pauses can adversely affect the fault detection system
rendering the cluster instable because of unnecessary reconfigurations.
Safepoint metrics are crucial for debugging such problems, so this patch
adds safe point stats to the standard collection process.

Why should this be merged: Improves debugging cluster instability and performance issues. 

- [x] Public API has Javadoc

### DIFF
--- a/common/src/main/java/org/corfudb/common/metrics/micrometer/JVMMetrics.java
+++ b/common/src/main/java/org/corfudb/common/metrics/micrometer/JVMMetrics.java
@@ -1,0 +1,44 @@
+package org.corfudb.common.metrics.micrometer;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.util.Optional;
+import lombok.Data;
+import lombok.Getter;
+import sun.management.HotspotRuntimeMBean;
+import sun.management.ManagementFactoryHelper;
+
+/** Provides JVM level metrics. */
+
+public final class JVMMetrics {
+
+    @Getter(lazy=true)
+    private final static HotspotRuntimeMBean runtimeMBean = ManagementFactoryHelper.getHotspotRuntimeMBean();
+
+    @Data
+    private static class SafePointStats {
+        long safepointTime;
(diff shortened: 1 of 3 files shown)


PavelZaytsev (approved):

common/src/main/java/org/corfudb/common/metrics/micrometer/JVMMetrics.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Data
+    private static class SafePointStats {
+        long safepointTime;
+        long safepointCount;
+    }
+
+    public static void register(Optional<MeterRegistry> metricsRegistry) {
+
+        if (metricsRegistry.isPresent()) {
```
xnull, line 26: maybe we should use micrometer's JVM metrics? https://micrometer.io/docs/ref/jvm

xnull (approved):
looks ok

common/src/main/java/org/corfudb/common/metrics/micrometer/JVMMetrics.java
Maithem: I checked micrometer's jvm metrics, but they don't collect TotalSafepointTime. Maybe because its a "private" API.

infrastructure/src/main/resources/logback.prod.xml
```diff
@@ -50,7 +50,7 @@
     <root level="info">
         <appender-ref ref="async_file"/>
     </root>
-    <logger additivity="false" level="info" name="org.corfudb.metricsdata">
```
Maithem, line 53: @PavelZaytsev this is a bug right? PR #2841 will disable the server metrics.

PavelZaytsev: This line should be 'debug' because in MeterRegistryInitializer we pass logger consumer as logger.debug. It needs to be whatever the consumer we are passing to create a registry

PR #2841, opened 2020-12-07
Remove deps on logback

PavelZaytsev:
## Overview

Description:

Fixes log binding issues. 

- [x] Public API has Javadoc

### DIFF
--- a/common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
+++ b/common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
@@ -5,6 +5,7 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.logging.LoggingRegistryConfig;
+import lombok.extern.slf4j.Slf4j;
 import org.corfudb.common.metrics.micrometer.registries.LoggingMeterRegistryWithHistogramSupport;
 import org.slf4j.Logger;
 
@@ -18,6 +19,7 @@ import static org.corfudb.common.metrics.micrometer.registries.LoggingMeterRegis
 /**
  * A configuration class for a meter (metrics) registry.
  */
+@Slf4j
 public class MeterRegistryProvider {
     private static Optional<MeterRegistry> meterRegistry = Optional.empty();
     private static Optional<String> endpoint = Optional.empty();
@@ -79,7 +81,7 @@ public class MeterRegistryProvider {
     /**
      * Register timer if needed.
      * @param name Name of a timer.
-     * @param tags Tags for a timer. 
+     * @param tags Tags for a timer.
      */
     public static void timer(String name, String... tags) {
         MeterRegistryProvider.getInstance().ifPresent(registry -> registry.timer(name, tags));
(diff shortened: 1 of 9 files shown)


hisundar (approved):
Lgtm thanks!

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -847,13 +848,9 @@ private CorfuRuntime(@Nonnull CorfuRuntimeParameters parameters) {
...
         if (microMeterRuntimeConfig.metricsEnabled) {
-            LoggerContext context =  (LoggerContext) LoggerFactory.getILoggerFactory();
-            registry = Optional.ofNullable(context.exists(microMeterRuntimeConfig.configuredLoggerName))
-                    .map(logger -> MeterRegistryProvider.MeterRegistryInitializer.newInstance(logger,
+            Logger logger = LoggerFactory.getLogger(microMeterRuntimeConfig.configuredLoggerName);
+            registry = Optional.of(MeterRegistryInitializer.newInstance(logger,
                             microMeterRuntimeConfig.loggingInterval, parameters.clientId));
-            if (!registry.isPresent()) {
-                log.warn("Logger was not found to enable metrics.");
-            }
```
Maithem, line 856: You can probably check if a logger was defined or not with `Logger::isInfoEnabled`

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
```diff
@@ -70,7 +72,8 @@ public static synchronized void init(Logger logger, Duration loggingInterval, St
 
         private static void init(Supplier<Optional<MeterRegistry>> meterRegistrySupplier) {
             if (meterRegistry.isPresent()) {
-                throw new IllegalStateException("Registry has already been initialized.");
+                log.warn("Registry has already been initialized.");
+                return;
```
Maithem, line 76: What if two different suppliers are passed? 

I don't think its a good idea to just silently print a warning message.

xnull (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/HandlerMethods.java
```diff
@@ -207,10 +207,14 @@ private void registerMethod(@Nonnull final MethodHandles.Lookup caller,
...
+                aType -> (CorfuComponent.INFRA_MSG_HANDLER +
+                        aType.name().toLowerCase()));
+        double[] percentiles = new double[]{0.50, 0.95, 0.99};
 
         return MeterRegistryProvider.getInstance()
-                .map(registry -> Timer.builder(timerNameCache.get(type)).register(registry));
+                .map(registry -> Timer.builder(timerNameCache.get(type))
+                        .publishPercentiles(percentiles)
+                        .publishPercentileHistogram(true)
+                        .register(registry));
```
zfrenette, line 218: Should these updates also be made to the `RequestHandlerMethods` class? This class is only responsible for `CorfuMsg` types and does not take into account the server handler methods that handle Protobuf messages.

PavelZaytsev: Did not know about RequestHandlerMethods. I'll address in a separate PR, because this one is mostly to remove dependencies on logback.

common/src/main/java/org/corfudb/common/metrics/micrometer/MeterRegistryProvider.java
PavelZaytsev: I threw exception here and caught it at the places where we initialize a registry. We do need to ignore it for Corfu and Replication servers because otherwise, it would fail on restarts.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
PavelZaytsev: done

hisundar:
Hi Pavel, as we discussed, please help address comments from @zfrenette and @Maithem in a subsequent patch. merged this to unblock consumers

PR #2843, opened 2020-12-16
UFO Metadata merge for repeated fields

annym:
## Overview

Description: support merge of metadata for repeated fields

Why should this be merged: requirement applications

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -584,6 +584,12 @@ private M mergeOldAndNewMetadata(@Nonnull M previousMetadata,
                         Optional.ofNullable(previousMetadata.getField(fieldDescriptor))
                                 .map(previousVersion -> (isCreate ? 0L : ((Long) previousVersion) + 1))
                                 .orElse(0L));
+            } else if (fieldDescriptor.isRepeated()) {
+                int count = previousMetadata.getRepeatedFieldCount(fieldDescriptor);
+                for (int i = 0; i < count; i++) {
+                    builder.setRepeatedField(fieldDescriptor, i,
+                            previousMetadata.getRepeatedField(fieldDescriptor, i));
```
pankti-m, line 591: should the repeated field always be set with the old metadata's values?  What if the new metadata has this field set?  Shouldnt we use the new values in that case?

hisundar (approved):
Thanks Anny! LGTM!

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
annym: Done, please take a look at the change!

PR #2844, opened 2020-12-17
LayoutServer isBootstrapped checking cleanup

xcchang:
## Overview

Description: 

The LayoutServer is now single-threaded since all the handlers need to be synchronized. The checking of whether the LayoutServer has been bootstrapped could be moved to processRequest which happends at the entry point of each handler call.

Why should this be merged: 

It's a comment left in #2823 after it's approved, so addressed separately here.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
@@ -112,6 +112,13 @@ public class LayoutServer extends AbstractServer {
 
     @Override
     protected void processRequest(RequestMsg req, ChannelHandlerContext ctx, IServerRouter r) {
+        // Don't process layout server requests unless the server is bootstrapped
+        if (!req.getPayload().getPayloadCase().equals(RequestPayloadMsg.PayloadCase.BOOTSTRAP_LAYOUT_REQUEST)
+                && !isBootstrapped(req)) {
+            r.sendNoBootstrapError(req.getHeader(), ctx);
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


zfrenette (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -441,11 +433,6 @@ public void handleCommitLayoutRequest(@Nonnull RequestMsg req, @Nonnull ChannelH
             return;
         }
 
-        if (!isBootstrapped(req)) {
-            r.sendNoBootstrapError(req.getHeader(), ctx);
-            return;
-        }
-
```
Maithem, line 448: I think this is a bug. See how this check happens after `if (payload.getForced())` , but you are reversing the order. So if it's not bootstrapped, you cant force a bootstrap.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -141,11 +148,6 @@ private boolean isBootstrapped(RequestMsg requestMsg) {
     @RequestHandler(type = RequestPayloadMsg.PayloadCase.LAYOUT_REQUEST)
     public void handleLayoutRequest(@Nonnull RequestMsg req, @Nonnull ChannelHandlerContext ctx,
```
Maithem, line 149: Since the class is not thread safe its better to disallow access to the handlers. You can mark them as private and annotate with `VisibibleForTesting`

PR #2850, opened 2021-01-05
Fix NPE thrown by browser tool when value and/or metadata are null.

pankti-m:
## Overview
If the metadata schema is null, trying to read the corresponding entry throws a NPE.  This PR fixes the issue.

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -142,9 +142,11 @@ public class CorfuStoreBrowser {
                     builder = new StringBuilder("\nKey:\n")
                             .append(JsonFormat.printer().print(entry.getKey().getKey()))
                             .append("\nPayload:\n")
-                            .append(JsonFormat.printer().print(entry.getValue().getPayload()))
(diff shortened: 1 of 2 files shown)


corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -142,9 +142,11 @@ public int printTable(String namespace, String tablename) {
                     builder = new StringBuilder("\nKey:\n")
                             .append(JsonFormat.printer().print(entry.getKey().getKey()))
                             .append("\nPayload:\n")
-                            .append(JsonFormat.printer().print(entry.getValue().getPayload()))
+                            .append(entry.getValue() != null && entry.getValue().getPayload() != null ?
```
WenbinZhu, line 145: Is it possible that value is null?

WenbinZhu (approved):
LGTM, just have a question.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: I checked the code and it is not possible that value is null.  But this is how it has been changed in Impactor(check for value == null) so I have reused the same approach.

annym (approved):

PR #2851, opened 2021-01-06
Nested Secondary Index Alias Support

annym:
## Overview

Description: 

Support access to secondary indexes through a given alias instead
of secondary index full path.

A custom alias can be specified upon protoBuf definition, otherwise the alias
defaults to the name of the last attribute in the multi-level secondary key path.

Why should this be merged: client request

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
@@ -80,6 +80,7 @@ public class CorfuTable<K, V> implements
     private final ContextAwareMap<K, V> mainMap;
     private final Set<Index.Spec<K, V, ?>> indexSpec;
     private final Map<String, Map<Object, Map<K, V>>> secondaryIndexes;
+    private final Map<String, String> secondaryIndexesAliasToPath;
     private final CorfuTable<K, V> optimisticTable;
     private final VersionPolicy versionPolicy;
 
@@ -90,6 +91,7 @@ public class CorfuTable<K, V> implements
         this.mainMap = mainMap;
         this.indexSpec = indexSpec;
         this.secondaryIndexes = secondaryIndexes;
+        this.secondaryIndexesAliasToPath = new HashMap<>();
         this.optimisticTable = optimisticTable;
         this.versionPolicy = ICorfuVersionPolicy.DEFAULT;
     }
@@ -104,6 +106,7 @@ public class CorfuTable<K, V> implements
                       VersionPolicy versionPolicy) {
         this.indexSpec = new HashSet<>();
         this.secondaryIndexes = new HashMap<>();
+        this.secondaryIndexesAliasToPath = new HashMap<>();
         this.mainMap = streamingMapSupplier.get();
         this.versionPolicy = versionPolicy;
         this.optimisticTable = new CorfuTable<>(this.mainMap.getOptimisticMap(), this.indexSpec,
@@ -111,6 +114,7 @@ public class CorfuTable<K, V> implements
 
         indices.forEach(index -> {
             secondaryIndexes.put(index.getName().get(), new HashMap<>());
+            secondaryIndexesAliasToPath.put(index.getAlias().get(), index.getName().get());
             indexSpec.add(index);
         });
 
@@ -233,8 +237,9 @@ public class CorfuTable<K, V> implements
     Collection<Entry<K, V>> getByIndex(@Nonnull Index.Name indexName, I indexKey) {
         String secondaryIndex = indexName.get();
         Map<Object, Map<K, V>> secondaryMap;
-        if (secondaryIndexes.containsKey(secondaryIndex) &&
-                ((secondaryMap = secondaryIndexes.get(secondaryIndex)) != null)) {
+        if ((secondaryIndexes.containsKey(secondaryIndex) &&
+                ((secondaryMap = secondaryIndexes.get(secondaryIndex)) != null)) || secondaryIndexesAliasToPath.containsKey(secondaryIndex)
+                && ((secondaryMap = secondaryIndexes.get(secondaryIndexesAliasToPath.get(secondaryIndex))) != null)) {
             // If secondary index exists and function for this index is not null
             Map<K, V> res = secondaryMap.get(indexKey);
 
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -233,8 +237,9 @@ public V get(@ConflictParameter Object key) {
     Collection<Entry<K, V>> getByIndex(@Nonnull Index.Name indexName, I indexKey) {
         String secondaryIndex = indexName.get();
         Map<Object, Map<K, V>> secondaryMap;
-        if (secondaryIndexes.containsKey(secondaryIndex) &&
-                ((secondaryMap = secondaryIndexes.get(secondaryIndex)) != null)) {
+        if ((secondaryIndexes.containsKey(secondaryIndex) &&
```
pankti-m, line 240: if using an alias, "secondaryIndex" will be the alias, correct?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
```diff
@@ -233,8 +237,9 @@ public V get(@ConflictParameter Object key) {
...
+                ((secondaryMap = secondaryIndexes.get(secondaryIndex)) != null)) || secondaryIndexesAliasToPath.containsKey(secondaryIndex)
+                && ((secondaryMap = secondaryIndexes.get(secondaryIndexesAliasToPath.get(secondaryIndex))) != null)) {
```
pankti-m, line 242: dont we need a similar change in getByIndexAndFilter(), un/mapSecondaryIndexes()?

runtime/src/main/java/org/corfudb/runtime/collections/Index.java
```diff
@@ -46,28 +46,42 @@
      */
     public static class Spec<K, V, I> {
         private final Name name;
+        private final Name alias;
         private final MultiValueFunction<K, V, I> indexFunction;
 
         public Spec(Name name, Function<K, V, I> indexFunction) {
+            this(name, name, indexFunction);
```
pankti-m, line 53: name will have the full path, correct?  dint we say the default will be only the last part of the path if alias is not specified?

runtime/src/main/java/org/corfudb/runtime/collections/Index.java
```diff
@@ -46,28 +46,42 @@
...
 
-        public Spec(Name name, MultiValueFunction<K, V, I> indexFunction) {
+        public Spec(Name name, Name alias, MultiValueFunction<K, V, I> indexFunction) {
             this.name = name;
+            this.alias = alias;
             this.indexFunction = indexFunction;
         }
 
+        public Spec(Name name, MultiValueFunction<K, V, I> indexFunction) {
+            this(name, name, indexFunction);
```
pankti-m, line 70: same here

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -31,22 +32,27 @@
             Index.Spec<Message, CorfuRecord<Message, Message>, ?>>
             indices = new HashMap<>();
 
+    // Map from secondary index alias to nested full path
+    private final HashMap<String, String> secondaryIndexAlias = new HashMap<>();
```
pankti-m, line 36: nit - can we rename it to secondaryIndexAliasToFullPath?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
annym: yes, "secondaryIndex" is either a fully qualified name or an alias, whatever was provided in the call (both are valid)

runtime/src/main/java/org/corfudb/runtime/collections/CorfuTable.java
annym: good catch, yes those paths require this as well, will add..

runtime/src/main/java/org/corfudb/runtime/collections/Index.java
annym: that is tru for protoBuf... however this can be extended by other Indexers, so this code is there to avoid a NPE in the case of other indexers for which the notion of alias is not present...

zhangn49 (approved):

Maithem (commented):
I agree that addressing a secondary index by the fully qualified path is verbose, but that issue only exists because `TxnContext::getByIndex` changes the interfaces from a type to a string, which is weird. If you look at `CorfuTable::getByIndex` it actually takes in a `Index.Name` type. Having a type is much cleaner than having an arbitrary string. Also, if that is fixed, then there is no point of aliasing since the fully qualified index name wont be necessary to do a secondary index query.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -187,30 +193,61 @@ private void registerIndices(final Descriptors.FieldDescriptor fieldDescriptor)
...
+                indices.put(secondaryKey.getValue(), getNestedIndex(secondaryKey.getValue(), secondaryKey.getKey()));
+                // For nested secondary indexes, aliases are supported (it can be user-defined or defaults to the last attributes name if not specified)
+                // e.g., full-path secondary key 'adult.children.child.age' defaults to alias 'age'
+                // We keep a map of alias to full path, as secondary indexes can ba accessed on either one of them
+                secondaryIndexAliasToFullPath.put(secondaryKey.getKey(), secondaryKey.getValue());
             }
         }
     }
 
+    private NestedSecondaryIndex parseSecondaryKey(String secondaryKey) {
```
WenbinZhu, line 216: Are there other ways to define the alias instead of manually parsing the string, which seems not very neat and might be error prone? For example, can the `nested_secondary_keys` be a custom option, which is some structure, instead of a string? Or is it possible to define multiple field options, one is `nested_secondary_keys`, and another alias option?

WenbinZhu (approved):
LGTM, just have one ask.

pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
annym: Sure, this sounds reasonable. 

I've made the change such that the new syntax is:

`message Adult {
    Person person = 1 [(org.corfudb.runtime.schema).nested_secondary_key = { index_path: "person.age" index_name: "age" },
                       (org.corfudb.runtime.schema).nested_secondary_key = { index_path: "person.children.child.age" index_name: "kidsAge" }];
    Company work = 2;
}`

annym:
> I agree that addressing a secondary index by the fully qualified path is verbose, but that issue only exists because `TxnContext::getByIndex` changes the interfaces from a type to a string, which is weird. If you look at `CorfuTable::getByIndex` it actually takes in a `Index.Name` type. Having a type is much cleaner than having an arbitrary string. Also, if that is fixed, then there is no point of aliasing since the fully qualified index name wont be necessary to do a secondary index query.

Discussed this offline! :)

PR #2857, opened 2021-01-15
Fix package name in queue.proto

Lujie1996:
## Overview

Description:

Why should this be merged: 
1. queue.proto does not set package correctly.
2. The message name should be generated correctly without a package name.

In `DynamicProtobufSerializer`, the key in `messagesFdProtoNameMap` is `.CorfuGuidMsg` but `getFullMessageName` returns `CorfuGuidMsg` (without prefix dot)

- [x] Public API has Javadoc

### DIFF
--- a/runtime/proto/queue.proto
+++ b/runtime/proto/queue.proto
@@ -1,16 +1,17 @@
-syntax = "proto2";
+syntax = "proto3";
 
+package org.corfudb.runtime;
 option java_package = "org.corfudb.runtime";
 
 message CorfuQueueIdMsg {
-    optional int64 txSequence = 1;
-    optional int64 entryId = 2;
+    int64 txSequence = 1;
+    int64 entryId = 2;
 }
 
 message CorfuGuidMsg {
-    optional int64 instanceId = 1;
+    int64 instanceId = 1;
 }
 
 message CorfuQueueMetadataMsg {
-    optional int64 txSequence = 1;
+    int64 txSequence = 1;
 }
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -191,7 +196,13 @@ private FileDescriptor getDescriptor(String name) throws DescriptorValidationExc
      */
     private String getMessageName(Any message) {
         String typeUrl = message.getTypeUrl();
-        return typeUrl.substring(typeUrl.lastIndexOf('.') + 1);
+        String messageName = typeUrl.substring(typeUrl.lastIndexOf('.') + 1);
+        if (messageName.contains("/")) {
+            // In case the message lacks package name
+            // E.g. typeUrl: type.googleapis.com/TableName
+            messageName = messageName.substring(messageName.lastIndexOf("/") + 1);
```
achinmishra, line 203: Did you print out the names of packages that contained a "/"? In my testing, all I could find were with "."

Lujie1996: Yes, inside RegistryTable there are some tables look like this `type.googleapis.com/TableName`

runtime/proto/queue.proto
```diff
@@ -1,16 +1,17 @@
-syntax = "proto2";
+syntax = "proto3";
 
+package org.corfudb.runtime;
 option java_package = "org.corfudb.runtime";
 
 message CorfuQueueIdMsg {
-    optional int64 txSequence = 1;
-    optional int64 entryId = 2;
+    int64 txSequence = 1;
```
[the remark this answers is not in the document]
Lujie1996: I think it's the proto3 syntax that fields don't need `optional` any more.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
Lujie1996: Actually I just reviewed the RegistryTable in the setup, and there isn't this pattern. But I remember I did encounter an issue when I didn't handle this pattern in another setup I created using newer impactor build (which is no longer on Jarvis). Let's keep it as it is in case any such pattern appears.

Maithem (approved):

PavelZaytsev (approved):

pankti-m (approved):

PR #2861, opened 2021-01-20
Metrics format modification

PavelZaytsev:
## Overview

Description:

- Fix the metrics format to facilitate easier data ingestion.
- Fix timer for range and single writes.


common/src/main/java/org/corfudb/common/metrics/micrometer/registries/LoggingMeterRegistryWithHistogramSupport.java
```diff
@@ -83,130 +157,22 @@ protected void publish() {
...
-                                    double count = timer.count();
-                                    if (!config.logInactive() && count == 0) return;
-                                    loggingSink.accept(print.id() + " throughput=" + print.rate(count) +
-                                            " mean=" + print.time(timer.mean(getBaseTimeUnit())));
-                                },
-                                meter -> loggingSink.accept(writeMeter(meter, print))
-                        );
-                    });
+                    .flatMap(meter -> meter.match(
+                            gauge -> writeGauge(gauge.getId(), gauge.value()),
```
Maithem, line 161: Does `writeGauge` also write the histogram percentiles?

xnull (approved):
Looks good

common/src/main/java/org/corfudb/common/metrics/micrometer/registries/LoggingMeterRegistryWithHistogramSupport.java
PavelZaytsev: Yes, I've included tests on percentiles to verify it.

Maithem: Histogram objects usually also include count. Is there an easy way to print the count per histogram as well?

Instead of tracking throughput independently they can be derived from the histgorams.

Maithem (approved):

PR #2862, opened 2021-01-22
Remove timestamp from infra metrics logs

PavelZaytsev:
## Overview

Description:

The timestamp should be removed for each line because now it is a part of a message (in milliseconds). This notation follows the line protocol.

### DIFF
--- a/infrastructure/src/main/resources/logback.prod.xml
+++ b/infrastructure/src/main/resources/logback.prod.xml
@@ -31,7 +31,7 @@
         </triggeringPolicy>
         <encoder>
             <pattern>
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/resources/logback.prod.xml
```diff
@@ -31,7 +31,7 @@
         </triggeringPolicy>
         <encoder>
             <pattern>
-                %date{yyyy-MM-dd'T'HH:mm:ss.SSSXXX, UTC} | %msg%n%xException
+                %msg%n%xException
```
Maithem, line 34: This will make it hard to inspect the metrics files manually. I suggest you just post process the logs and truncate these timestamps.

PR #2863, opened 2021-01-25
[backport] LR supports FQDN 

zhangn49:
## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -124,7 +126,7 @@
     /**
      * Current node's id
      */
-    private Optional<String> localNodeId = Optional.empty();
+    private String localNodeId;
```
xnull, line 129: instead of String, it should be URL. Would be great to change it to URL if possible

zhangn49: The node id's format is more like a uuid, so I used String here.

PR #2866, opened 2021-01-30
Hide Stream Bitmap Implementation

Maithem:
## Overview
This refactor hides the stream's bitmap implementation from the
consumers. It also improves code robustness and fixes some bugs
related to directly consuming the bitmap. Furthermore, this change will
allow for easily changing the implementation to optimize encoding
and certain APIs.

Why should this be merged: Improves code quality and allows for future optimizations. 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -37,7 +37,6 @@ import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.Layout;
 import org.corfudb.runtime.view.stream.StreamAddressSpace;
 import org.corfudb.util.Utils;
-import org.roaringbitmap.longlong.Roaring64NavigableMap;
 
 import javax.annotation.Nonnull;
 import java.lang.invoke.MethodHandles;
@@ -537,17 +536,13 @@ public class SequencerServer extends AbstractServer {
 
             for (Map.Entry<UUID, StreamAddressSpace> streamAddressSpace :
                     streamsAddressMap.entrySet()) {
-                log.info("Stream[{}] set to last trimmed address {} and {} addresses in the range" +
-                                " [{}-{}], on sequencer reset.",
+                log.info("Stream[{}]={} on sequencer reset.",
                         Utils.toReadableId(streamAddressSpace.getKey()),
-                        streamAddressSpace.getValue().getTrimMark(),
-                        streamAddressSpace.getValue().getAddressMap().getLongCardinality(),
-                        streamAddressSpace.getValue().getLowestAddress(),
-                        streamAddressSpace.getValue().getHighestAddress());
+                        streamAddressSpace.getValue());
                 if (log.isTraceEnabled()) {
-                    log.trace("Stream[{}] address map on sequencer reset: {}",
+                    log.trace("Stream[{}] addresses: {}",
                             Utils.toReadableId(streamAddressSpace.getKey()),
-                            streamAddressSpace.getValue().getAddressMap());
+                            streamAddressSpace.getValue().toArray());
                 }
             }
         }
@@ -728,10 +723,7 @@ public class SequencerServer extends AbstractServer {
             // (to keep track of all updates to this stream)
             streamsAddressMap.compute(uuid, (streamId, addressMap) -> {
                 if (addressMap == null) {
-                    addressMap = new StreamAddressSpace(
-                            Address.NON_ADDRESS,
-                            new Roaring64NavigableMap()
-                    );
+                    addressMap = new StreamAddressSpace();
                 }
 
                 for (long i = globalLogTail; i < newTail; i++) {
@@ -825,17 +817,14 @@ public class SequencerServer extends AbstractServer {
     private Map<UUID, StreamAddressSpace> getStreamsAddressesMap(
             List<StreamAddressRangeMsg> addressRanges) {
         Map<UUID, StreamAddressSpace> requestedAddressSpaces = new HashMap<>();
-        Roaring64NavigableMap addressMap;
 
         for (StreamAddressRangeMsg streamAddressRange : addressRanges) {
             UUID streamId = getUUID(streamAddressRange.getStreamId());
             // Get all addresses in the requested range
             if (streamsAddressMap.containsKey(streamId)) {
-                addressMap = streamsAddressMap.get(streamId)
+                StreamAddressSpace addressesInRange = streamsAddressMap.get(streamId)
                         .getAddressesInRange(getStreamAddressRange(streamAddressRange));
-
-                requestedAddressSpaces.put(streamId, new StreamAddressSpace(
-                        streamsAddressMap.get(streamId).getTrimMark(), addressMap));
+                requestedAddressSpaces.put(streamId, addressesInRange);
             } else {
                 log.warn("getStreamsAddressesMap: address space map is not present for stream {}." +
                         " Verify this is a valid stream.", streamId);
--- a/infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
+++ b/infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
@@ -29,7 +29,6 @@ import org.mockito.Mock;
 import org.mockito.Spy;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
-import org.roaringbitmap.longlong.Roaring64NavigableMap;
 
 import java.io.ObjectInputStream;
 import java.util.ArrayList;
@@ -325,9 +324,9 @@ public class SequencerServerTest {
         long globalTailA = 0;
         long globalTailB = 1;
         StreamAddressSpace streamAddressSpaceA = new StreamAddressSpace(Address.NON_ADDRESS,
-                Roaring64NavigableMap.bitmapOf(globalTailA));
-        StreamAddressSpace streamAddressSpaceB = new StreamAddressSpace(Address.ABORTED,
-                Roaring64NavigableMap.bitmapOf(globalTailB));
+                Collections.singleton(globalTailA));
+        StreamAddressSpace streamAddressSpaceB = new StreamAddressSpace(Address.NON_ADDRESS,
+                Collections.singleton(globalTailB));
         tailMap.put(streamA, streamAddressSpaceA);
         tailMap.put(streamB, streamAddressSpaceB);
 
@@ -390,9 +389,9 @@ public class SequencerServerTest {
         long globalTailA = 0;
         long globalTailB = 1;
         StreamAddressSpace streamAddressSpaceA = new StreamAddressSpace(Address.NON_ADDRESS,
-                Roaring64NavigableMap.bitmapOf(globalTailA));
-        StreamAddressSpace streamAddressSpaceB = new StreamAddressSpace(Address.ABORTED,
-                Roaring64NavigableMap.bitmapOf(globalTailB));
+                Collections.singleton(globalTailA));
+        StreamAddressSpace streamAddressSpaceB = new StreamAddressSpace(Address.NON_ADDRESS,
+                Collections.singleton(globalTailB));
         tailMap.put(streamA, streamAddressSpaceA);
         tailMap.put(streamB, streamAddressSpaceB);
         long bootstrapMsgEpoch = 1;
@@ -434,8 +433,7 @@ public class SequencerServerTest {
 
         // Verify the values. Note that global tail is of streamB as it was greater than streamA.
         assertEquals(globalTailB, response.getPayload().getStreamsAddressResponse().getLogTail());
-        assertEquals(Roaring64NavigableMap.bitmapOf(globalTailA),
-                responseStreamAddressSpace.getAddressMap());
+        assertEquals(new StreamAddressSpace(Collections.singleton(globalTailA)), responseStreamAddressSpace);
         assertEquals(Address.NON_ADDRESS, responseStreamAddressSpace.getTrimMark());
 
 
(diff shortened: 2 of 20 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -94,32 +115,32 @@ private void mergeFrom(StreamAddressSpace other) {
...
     }
 
     /**
      * Add an address to this address space.
      *
      * @param address address to add.
      */
-    public void addAddress(long address) {
-        addressMap.addLong(address);
+    public void addAddress(Long address) {
```
Maithem, line 131: int to long implicit conversion should be safe. Revert type to a primitive long to prevent auto-boxing

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -323,9 +323,9 @@ public void testBootStrapSequencer() {
         long globalTailA = 0;
         long globalTailB = 1;
         StreamAddressSpace streamAddressSpaceA = new StreamAddressSpace(Address.NON_ADDRESS,
-                Roaring64NavigableMap.bitmapOf(globalTailA));
-        StreamAddressSpace streamAddressSpaceB = new StreamAddressSpace(Address.ABORTED,
```
Maithem, line 327: @zfrenette This is a bug in the unit test, right? (Aborted shouldnt be used here)

zfrenette: Yeah, `ABORTED` should not have been used... `NON_ADDRESS` is more appropriate here.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -395,21 +393,17 @@ protected long getMaxGlobalFromMaxEntries(int maxEntries) {
...
             return streamAddressSpace.getHighestAddress();
         }
 
-        LongIterator it = streamAddressSpace.getAddressMap().getLongIterator();
-        while (--maxEntries > 0) {
-            it.next();
-        }
-
-        return it.next();
+        return streamAddressSpace.select(maxEntries);
```
zfrenette, line 406: Is there an off by one error here?

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
```diff
@@ -40,7 +55,122 @@ public void testStreamAddressSpaceMerge() {
...
+    @Test
+    public void testNettySerialization() throws Exception {
+        StreamAddressSpace obj1 = new StreamAddressSpace(3, ImmutableSet.of(1L, 2L, 3L, 4L, 5L));
+        ByteBuf buf = Unpooled.buffer();
+        obj1.serialize(buf);
+        StreamAddressSpace deserialized = StreamAddressSpace.deserialize(buf);
+        assertThat(obj1.getTrimMark()).isEqualTo(deserialized.getTrimMark());
+        assertThat(obj1.size()).isEqualTo(deserialized.size());
+        assertThat(obj1.toArray()).isEqualTo(deserialized.toArray());
+    }
```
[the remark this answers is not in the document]
Maithem: It is tested in `testStreamSerialization`. The backing implementation of DataOutput/DataInputStream isn't what is being tested here. The use of ByteBuf is just for convenience to create DataOutput/DataInputStream , so all four serialization/deserialization methods are being tested.

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
```diff
@@ -40,7 +49,124 @@ public void testStreamAddressSpaceMerge() {
...
+        assertThat(obj2.contains(6L)).isTrue();
+
+        StreamAddressSpace obj3 = new StreamAddressSpace(Collections.singleton(6L));
+        assertThat(obj3.getTrimMark()).isEqualTo(Address.NON_ADDRESS);
+        assertThat(obj3.size()).isEqualTo(1L);
+        assertThat(obj3.contains(6L)).isTrue();
+    }
+
+    @Test
+    @Ignore
```
annym, line 74: should this ignore be removed? I'm guessing it has to do with the commented code you had in the constructors...

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
```diff
@@ -40,7 +49,124 @@ public void testStreamAddressSpaceMerge() {
...
+        assertThat(obj1.contains(4L)).isFalse();
+
+    }
+
+    @Test
+    public void testCopy() {
+        StreamAddressSpace obj1 = new StreamAddressSpace();
+        StreamAddressSpace obj1Copy = obj1.copy();
+        assertThat(obj1Copy.getTrimMark()).isEqualTo(obj1.getTrimMark());
+        assertThat(obj1Copy.size()).isEqualTo(obj1.size());
```
annym, line 112: probably (for completeness) we can add an assertion to verify they're not the same reference..

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -432,8 +431,7 @@ public void testStreamAddressRequest() {
 
         // Verify the values. Note that global tail is of streamB as it was greater than streamA.
         assertEquals(globalTailB, response.getPayload().getStreamsAddressResponse().getLogTail());
-        assertEquals(Roaring64NavigableMap.bitmapOf(globalTailA),
-                responseStreamAddressSpace.getAddressMap());
+        assertEquals(new StreamAddressSpace(Collections.singleton(globalTailA)), responseStreamAddressSpace);
         assertEquals(Address.NON_ADDRESS, responseStreamAddressSpace.getTrimMark());
```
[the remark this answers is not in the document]
annym: +1

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
```diff
@@ -470,17 +468,15 @@ public void testStreamAddressRequest() {
...
             } else {
                 StreamAddressSpace responseStreamAddressSpaceB =
                         getStreamAddressSpace(uuidToStreamAddressSpacePairMsg.getAddressSpace());
                 assertEquals(globalTailB,
                         finalResponse.getPayload().getStreamsAddressResponse().getLogTail());
-                assertEquals(Roaring64NavigableMap.bitmapOf(globalTailB),
-                        responseStreamAddressSpaceB.getAddressMap());
-                assertEquals(Address.ABORTED, responseStreamAddressSpaceB.getTrimMark());
+                assertEquals(new StreamAddressSpace(Collections.singleton(globalTailB)), responseStreamAddressSpaceB);
+                assertEquals(Address.NON_ADDRESS, responseStreamAddressSpaceB.getTrimMark());
```
annym, line 479: same here, 479 can be removed

runtime/proto/rpc_common.proto
```diff
@@ -49,13 +49,7 @@ message LayoutMsg {
...
  */
 message StreamAddressSpaceMsg {
-  // Holds the last trimmed address for this stream.
-  // Note: keeping the last trimmed address is required in order to properly set the stream tail on sequencer resets
-  // when a stream has been checkpointed and trimmed and there are no further updates to this stream.
-  int64 trim_mark = 1;
-
-  // Holds the complete map of addresses for this stream.
-  // Note: currently of type Roaring64NavigableMap.
+  // Holds the last trimmed address for this stream and complete map of addresses for this stream.
```
annym, line 52: why do we remove the trim mark? we wouldn't be covering the case of a fully checkpointed stream whose last address is no longer present, but we need this to set the stream tail... I'm thinking we probably don't have a test that covers this scenario which we have seen in production..

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: +1:

    @Test
    public void xyz() {
        StreamAddressSpace space = new StreamAddressSpace(Arrays.asList(1L, 2L, 6L, 8L, 12L, 15L));
        System.out.println(">>>> " + space.select(space.size()));
    }

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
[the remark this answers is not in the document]
Maithem: If someone changes the Default trim mark from for example -1L to (NOT_FOUND = -3L) or (NON_EXIST = -6L) the assertion in like 434 would pass, but it shouldn't. They are both negative, but they are interpreted different in different areas of the code base.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
Maithem: Good catch. Will add more tests.

runtime/proto/rpc_common.proto
Maithem: It's just removed from the proto model, but it is still serialized/deserialized

In `StreamAddressSpace`
```java
    public void serialize(DataOutput out) throws IOException {
        out.writeLong(trimMark);
        bitmap.serialize(out);
    }

    public static StreamAddressSpace deserialize(DataInputStream in) throws IOException {
        long trimMark = in.readLong();
        Roaring64NavigableMap map = new Roaring64NavigableMap();
        map.deserialize(in);
        return new StreamAddressSpace(trimMark, map);
    }
```

The reason I removed it is because the bitmap and the trimmark shouldn't be accessed outside of StreamAddressSpace object.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -1,57 +1,89 @@
...
+        addresses.forEach(bitmap::addLong);
+        //this.trim(trimMark);
+    }
+
+    public StreamAddressSpace(Collection<Long> addresses) {
+        this(Address.NON_ADDRESS, addresses);
+    }
+
+    public boolean contains(long address) {
+        // TODO(Maithem): < or <= ?
```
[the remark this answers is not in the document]
Maithem: If the bitmap = [1 , 2, 3] and setTrimMark = 2 then contains would return true for 1 and 2, but that is incorrect. setTrimMark should truncate the bitmap, but there is existing code that depends on this incorrect behavior.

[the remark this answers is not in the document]
Maithem: since the bitmap implementation accepts negative numbers and interprets them as greater than positive numbers, we need to return false for negative numbers. For example, the bitmap implementation has this order 0, 1 , ..., Long.Max_value, Long.min_value, ... -2, -1

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
Maithem: Done.

infrastructure/src/test/java/org/corfudb/infrastructure/SequencerServerTest.java
Maithem: We have to check that both StreamAddressSpace objects are the same AND that the trim mark is NON_ADDRESS.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -1,57 +1,89 @@
...
+     * The internal bitmap container shouldn't be exposed to external consumers.
+     * @param trimMark Stream's trim mark
+     * @param bitmap Stream's bitmap
+     */
+    public StreamAddressSpace(long trimMark, Roaring64NavigableMap bitmap) {
         this.trimMark = trimMark;
-        this.addressMap = addressMap;
+        this.bitmap = bitmap;
+        //TODO(Maithem): re-enable this. cluster re-configuration tests are failing because
+        // they relay on incorrect behavior.
```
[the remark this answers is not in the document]
Maithem: Yes.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -1,57 +1,89 @@
...
-        this.trimMark = Address.NON_ADDRESS;
+        this(Address.NON_ADDRESS, Roaring64NavigableMap.bitmapOf());
     }
 
-    public Roaring64NavigableMap getAddressMap() {
-        return addressMap;
+    public StreamAddressSpace(long trimMark, Collection<Long> addresses) {
+        this(trimMark, Roaring64NavigableMap.bitmapOf());
+        addresses.forEach(bitmap::addLong);
+        //this.trim(trimMark);
```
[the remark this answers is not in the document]
Maithem: Fixed.

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
Maithem: Fixed.

Maithem:
All comments have been addressed. Please take a look.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -537,17 +536,13 @@ public void resetServer(@Nonnull RequestMsg req,
...
-                        streamAddressSpace.getValue().getTrimMark(),
-                        streamAddressSpace.getValue().getAddressMap().getLongCardinality(),
-                        streamAddressSpace.getValue().getLowestAddress(),
-                        streamAddressSpace.getValue().getHighestAddress());
+                        streamAddressSpace.getValue());
                 if (log.isTraceEnabled()) {
                     log.trace("Stream[{}] address map on sequencer reset: {}",
                             Utils.toReadableId(streamAddressSpace.getKey()),
-                            streamAddressSpace.getValue().getAddressMap());
+                            streamAddressSpace.getValue().toString());
```
Maithem, line 545: this should be `streamAddressSpace.getValue().toArray()`

zfrenette: That's fine, but it requires making `toArray()` public.

annym (approved):

zfrenette (approved):

PR #2871, opened 2021-02-02
Optimize StreamAddressSpace Range Operations

Maithem:
## Overview
Optimize StreamAddressSpace Range Operations 

This patch introduces space/runtime optimizations and makes the
following changes:
2. On trim create a suffix copy to avoid needing to call runOptimize
   and trim on the bitmap
3. Replace linear scan range queries with bitwise operations

Why should this be merged: Fixes a performance bug caused by linear time range queries and improves sanity checking
prevent silent in memory corruption.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
@@ -13,8 +13,10 @@ import org.corfudb.runtime.view.StreamOptions;
 import javax.annotation.Nonnull;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.NavigableSet;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.UUID;
@@ -171,7 +173,9 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
         // Transfer discovered addresses to queue. We must limit to maxGlobal,
         // as startAddress could be ahead of maxGlobal---in case it reflects
         // the tail of the stream.
-        queue.addAll(streamAddressSpace.copyAddressesToSet(maxGlobal));
+        Set<Long> prefix = new HashSet<>();
+        streamAddressSpace.forEachUpTo(maxGlobal, prefix::add);
+        queue.addAll(prefix);
 
         final long trimMark = streamAddressSpace.getTrimMark();
 
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
@@ -1,5 +1,6 @@
 package org.corfudb.runtime.view.stream;
 
+import com.google.common.primitives.Longs;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.StreamAddressRange;
 import org.corfudb.runtime.view.Address;
@@ -10,12 +11,8 @@ import org.roaringbitmap.longlong.Roaring64NavigableMap;
 import java.io.DataInputStream;
 import java.io.DataOutput;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.NavigableSet;
 import java.util.Set;
-import java.util.TreeSet;
+import java.util.function.LongConsumer;
 
 /**
  * This class represents the space of all addresses belonging to a stream.
@@ -36,7 +33,7 @@ final public class StreamAddressSpace {
     private long trimMark;
 
     // Holds the complete map of addresses for this stream.
-    private final Roaring64NavigableMap bitmap;
+    private Roaring64NavigableMap bitmap;
 
     /**
      * This constructor is required to facilitate deserialization, keep it private.
@@ -47,7 +44,6 @@ final public class StreamAddressSpace {
     private StreamAddressSpace(long trimMark, Roaring64NavigableMap bitmap) {
         this.trimMark = trimMark;
         this.bitmap = bitmap;
-        this.trim(trimMark);
     }
 
     public StreamAddressSpace() {
@@ -55,9 +51,7 @@ final public class StreamAddressSpace {
     }
 
     public StreamAddressSpace(long trimMark, Set<Long> addresses) {
-        this(trimMark, Roaring64NavigableMap.bitmapOf());
-        addresses.forEach(bitmap::addLong);
-        this.trim(trimMark);
+        this(trimMark, Roaring64NavigableMap.bitmapOf(Longs.toArray(addresses)));
     }
 
     public StreamAddressSpace(Set<Long> addresses) {
@@ -84,34 +78,34 @@ final public class StreamAddressSpace {
      * @return returns a as the merged StreamAddressSpace
      */
     public static StreamAddressSpace merge(StreamAddressSpace a, StreamAddressSpace b) {
-        a.mergeFrom(b);
-        return a;
-    }
-
-    // Merge another StreamAddressSpace into this instance
-    private void mergeFrom(StreamAddressSpace other) {
-        this.trimMark = Long.max(trimMark, other.getTrimMark());
-        this.bitmap.or(other.bitmap);
+        long newTrimMark = Long.max(a.getTrimMark(), b.getTrimMark());
+        a.bitmap.or(b.bitmap);
 
         // Because the trim mark can increase after the merge another
         // trim needs to be issued
-        this.trim(this.trimMark);
+        a.trim(newTrimMark);
+        // Since the trim method only sets the trimMark if an address was trimmed, we need to force
+        // the new trim mark for cases where the trim mark increases, but there isn't any addresses
+        // to trim.
+        a.trimMark = newTrimMark;
+        return a;
     }
 
     /**
-     * Copy this stream's addresses to a set, under a given boundary (inclusive).
+     * Call a long consumer on range [trimMark + 1, max].
      *
-     * @param maxGlobal maximum address (inclusive upper bound)
+     * @param max maximum address (inclusive upper bound)
      */
-    public NavigableSet<Long> copyAddressesToSet(final long maxGlobal) {
-        NavigableSet<Long> queue = new TreeSet<>();
-        this.bitmap.forEach(address -> {
-            if (address <= maxGlobal) {
-                queue.add(address);
+    public void forEachUpTo(final long max, LongConsumer consumer) {
+        LongIterator iterator = this.bitmap.getLongIterator();
+        while (iterator.hasNext()) {
+            long current = iterator.next();
+            if (current > trimMark && current <= max) {
+                consumer.accept(current);
+            } else {
+                break;
             }
-        });
-
-        return queue;
+        }
     }
 
     /**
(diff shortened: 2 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -195,16 +195,20 @@ public void trim(long trimMark) {
...
+        Roaring64Bitmap addressesInRange = new Roaring64Bitmap();
+
+        if (!(range.getStart() > range.getEnd()) || this.bitmap.isEmpty()) {
+            return  new StreamAddressSpace(this.trimMark, addressesInRange);
         }
 
+        long leftRank = this.bitmap.rankLong(range.getEnd() + 1);
+        long rightRank = this.bitmap.rankLong(range.getStart());
+        long leftEndpoint = this.bitmap.select(leftRank);
+        long rightEndpoint = this.bitmap.select(rightRank);
```
Maithem, line 207: These indexes need to be adjusted for the case of -1 and Long.MAX_VALUE

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -84,34 +79,30 @@ public boolean contains(long address) {
...
-    public NavigableSet<Long> copyAddressesToSet(final long maxGlobal) {
-        NavigableSet<Long> queue = new TreeSet<>();
-        this.bitmap.forEach(address -> {
-            if (address <= maxGlobal) {
-                queue.add(address);
+    public void forEachUpTo(final long max, LongConsumer consumer) {
+        LongIterator iterator = this.bitmap.getLongIterator();
+        while (iterator.hasNext()) {
+            long current = iterator.next();
+            if (current <= max) {
```
Maithem, line 100: setTrim can break this contract.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +120,137 @@ public long getTail() {
...
 
     /**
      * Get addresses in range (end, start], where start > end.
      *
      * @return Bitmap with addresses in this range.
      */
-    public StreamAddressSpace getAddressesInRange(StreamAddressRange range) {
-        Roaring64NavigableMap addressesInRange = new Roaring64NavigableMap();
+    public StreamAddressSpace getAddressesInRange2(StreamAddressRange range) {
+        Roaring64Bitmap addressesInRange = new Roaring64Bitmap();
```
Maithem, line 218: Remove this.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +120,137 @@ public long getTail() {
...
                     addressesInRange.add(address);
                 }
             });
         }
+        StreamAddressSpace ret = new StreamAddressSpace(this.trimMark, addressesInRange);
+        StreamAddressSpace other = getAddressesInRange(range);
+
+        if (ret.getTrimMark() != other.getTrimMark() && !Arrays.equals(ret.toArray(), other.toArray())) {
+            throw new AssertionError();
+        }
```
Maithem, line 231: Remove this.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +120,137 @@ public long getTail() {
...
+        return ret;
+    }
+    public StreamAddressSpace getAddressesInRange(StreamAddressRange range) {
+        Roaring64Bitmap addressesInRange = new Roaring64Bitmap();
+
+        if (range.getStart() <= range.getEnd() ||
```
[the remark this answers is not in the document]
annym: should we instead throw an IllegalArgumentException? as the range is an invalid range?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -216,6 +263,8 @@ public StreamAddressSpace getAddressesInRange(StreamAddressRange range) {
 
     public void setTrimMark(long trimMark) {
         this.trimMark = trimMark;
+        // TODO(Maithem): This method should be removed. When the trimMark is set here, then trim is called
```
annym, line 266: But before removing this, changes need to be done on the trim method. The way it is right now, it would return when no addresses are in the bitmap without setting the trim mark (and this is needed as it is used when rebuilding the stream address space on the log unit)

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
```diff
@@ -175,6 +181,47 @@ public void testStreamSerialization() throws Exception {
         assertThat(obj1.toArray()).isEqualTo(deserialized.toArray());
     }
 
+    @Test
```
annym, line 184: Should we add a test for the forEachUpTo?

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -171,7 +171,9 @@ private void moveToReadQueue(final StreamAddressSpace streamAddressSpace,
         // Transfer discovered addresses to queue. We must limit to maxGlobal,
         // as startAddress could be ahead of maxGlobal---in case it reflects
         // the tail of the stream.
-        queue.addAll(streamAddressSpace.copyAddressesToSet(maxGlobal));
+        NavigableSet<Long> prefix = new TreeSet<>();
```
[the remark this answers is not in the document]
Maithem: Yes.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -84,34 +79,30 @@ public boolean contains(long address) {
...
-    private void mergeFrom(StreamAddressSpace other) {
-        this.trimMark = Long.max(trimMark, other.getTrimMark());
-        this.bitmap.or(other.bitmap);
+        a.trimMark = Long.max(a.trimMark, b.getTrimMark());
+        a.bitmap.or(b.bitmap);
 
         // Because the trim mark can increase after the merge another
         // trim needs to be issued
-        this.trim(this.trimMark);
+        a.trim(a.trimMark);
```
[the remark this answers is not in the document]
Maithem: It is. `a.trimMark = Long.max(a.trimMark, b.getTrimMark());` , but I think its better to set the new trim mark through by calling trim instead of setting the field directly.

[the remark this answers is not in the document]
Maithem: Sometimes we need to force the trim mark, because the trim method doesn't update the trimMark if there are no addresses to trim. I added a comment.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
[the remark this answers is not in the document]
Maithem: It is commented `Get addresses in range (end, start], where start > end` just need to remove `getAddressesInRange2`

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
Maithem: Tried to do it in this PR, but a bunch of tests failed. I will address that in a subsequent PR.

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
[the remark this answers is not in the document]
Maithem: Added a test `testMergeDifferentTrimMarks`

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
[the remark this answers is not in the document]
Maithem: added test `testGetAddressesInRange` for the new exception 

```java 

StreamAddressRange invalidRange = new StreamAddressRange(UUID.randomUUID(), 1,  2);
        assertThrows(IllegalArgumentException.class,
                () -> obj1.getAddressesInRange(invalidRange),
                "Invalid range (2, 1]");

```

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -36,28 +34,25 @@
     private long trimMark;
 
     // Holds the complete map of addresses for this stream.
-    private final Roaring64NavigableMap bitmap;
+    private Roaring64Bitmap bitmap;
```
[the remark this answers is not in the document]
Maithem: There is a memory leak in the RoaringBitmap library where removes don't de-allocate unused space. As a work around, I'm copying the subset and creating a new bitmap from it. Look here : https://github.com/CorfuDB/CorfuDB/pull/2871/files#diff-2316e4c0466f66804dc3be7894fc4a9263114393d28bafc7be324c59fb71c3f7R192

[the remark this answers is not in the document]
Maithem: I will change it back to final once they release these fixes
https://github.com/RoaringBitmap/RoaringBitmap/pull/453
https://github.com/RoaringBitmap/RoaringBitmap/pull/451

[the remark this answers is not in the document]
Maithem: This method can also be improved with a range remove, but that feature hasn't been implemented yet.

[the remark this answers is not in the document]
pankti-m: got it

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
annym: are we going to handle this?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
annym: getAddressesInRange2 can be removed, right?

annym (approved):
Just 2 minor comments.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
Maithem: Yes. Added `(current > trimMark && current <= max)`

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +124,136 @@ public long getTail() {
...
+     * @param newTrimMark upper limit of addresses to trim
      */
-    public void trim(long trimMark) {
-        if (!Address.isAddress(trimMark)) {
+    public void trim(long newTrimMark) {
+        if (!Address.isAddress(newTrimMark)) {
             // If not valid address return and do not attempt to trim.
-            log.warn("trim: attempting to trim non-valid address {}", trimMark);
+            log.warn("trim: attempting to trim non-valid address {}", newTrimMark);
+            return;
```
WenbinZhu, line 160: Should this throw exception instead of return?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +124,136 @@ public long getTail() {
...
-        // be removing all addresses upon an invalid trim mark.
-        long numAddressesToTrim = bitmap.rankLong(trimMark);
+        // At this point point the trim mark is positive and we need to compute
+        // the rank. The rank will indicate how many addresses need to be trimmed
+        long rank = bitmap.rankLong(newTrimMark);
 
-        if (numAddressesToTrim <= 0) {
+        if (rank <= 0) {
             return;
         }
```
WenbinZhu, line 176: Do you need to update trimMark to the new trimMark even though the rank is 0? Since at this point newTrimMark is greater than the current one.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +124,136 @@ public long getTail() {
...
     }
 
     /**
      * Get addresses in range (end, start], where start > end.
      *
      * @return Bitmap with addresses in this range.
      */
-    public StreamAddressSpace getAddressesInRange(StreamAddressRange range) {
-        Roaring64NavigableMap addressesInRange = new Roaring64NavigableMap();
+    public StreamAddressSpace getAddressesInRange2(StreamAddressRange range) {
```
WenbinZhu, line 221: Remove?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +124,136 @@ public long getTail() {
...
+        StreamAddressSpace ret = new StreamAddressSpace(this.trimMark, addressesInRange);
+        StreamAddressSpace other = getAddressesInRange(range);
+
+        if (ret.getTrimMark() != other.getTrimMark() && !Arrays.equals(ret.toArray(), other.toArray())) {
+            throw new AssertionError();
+        }
+        return ret;
+    }
+    public StreamAddressSpace getAddressesInRange(StreamAddressRange range) {
+        Roaring64Bitmap addressesInRange = new Roaring64Bitmap();
```
WenbinZhu, line 239: This can be moved down after the trival case check to avoid overheads.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,81 +124,136 @@ public long getTail() {
...
+        if (range.getStart() <= range.getEnd()) {
+            throw new IllegalArgumentException("Invalid range (" + range.getEnd() + ", " + range.getStart() + "]");
+        }
+
+        long first = getFirst();
+        long last = getTail();
+
+        if (this.bitmap.isEmpty() || range.getEnd() >= last || range.getStart() < first) {
+            // Return empty address space if the range doesn't intersect, or if the bitmap is empty
+            return  new StreamAddressSpace(this.trimMark, addressesInRange);
```
WenbinZhu, line 250: NIT: extra space

WenbinZhu (approved):
One comment might be noteworthy: in trim(), might need to update trimMark when rank is <= 0, others lgtm.

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
Maithem: This behavior is very subtle I added a unit test for it `testTrimMarkUpdate`

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -129,82 +123,109 @@ public long getTail() {
...
+                log.info("trim: stream trim mark moved from {} to {}", this.trimMark, current);
+                this.trimMark = current;
+                break;
+            }
+
+            trimmedBitmap.addLong(current);
         }
 
-        log.trace("trim: Remove {} addresses for trim mark {}", addressesToTrim.size(), trimMark);
+        this.bitmap = trimmedBitmap;
```
zhangn49, line 190: Let's say before the trim, address is [trimMark + 1, max]. Do we assume that [trimMark + 1, new trimMark] always has more addresses than [new trimMark + 1, max]?

If not, is it possible to calculate which part has less addresses? If [trimMark + 1, new trimMark] is smaller, is removing address better in this case?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
zhangn49: Do those tests fail if you call trim here?

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
Maithem: I thought about this when I first implemented it and I couldn't find a good general solution. The problem is to find that point where it's cheaper to removes VS when to copy is an expensive series of operations. For example, you need to do a couple of rank operations and then do the copy/remove, when I tested it a copy of the suffix turned out to be faster. Trim is called infrequently too, I think just paying the cost of the copy is reasonable and doesn't seem to require optimizing. 

If you have any suggestions, I can try them too.

zhangn49 (approved):

zfrenette (approved):
Looks good!

PR #2875, opened 2021-02-03
Removing batch size on smr entries and adding a limit on smr entry serialized size

achinmishra:
## Overview
Removing batch size operation when adding SMR entries of a table into a checkpoint log entry during compaction.

Description:
The issue was seen when compactor tried adding 3 entries amounting to more than the standard write size of 25 MB of any type of log entry. Each log entry were less than 25MB but during checkpointing, the checkpointer was considering a batch operation on a fixed batch size of <50 SMR entries, leading to a write entry with >25MB of serialized data.

This fix would add each SMR entry until the limit is reached. And then, create a checkpoint log entry and proceed to add the next SMR entry, leading to a dynamic sized batches of SMR entries with a hard check on the size of each checkpointed log entry.

Why should this be merged: 
A scenario where in a customer of the DB encounters write size exception during compaction, however, with no entries in the table violating the size rule of the serialized data. Its a bug and the batch operation should be removed during checkpointing. 

None

- [ YES] There are no TODOs left in the code
- [ YES] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [ NOT ] Change is covered by automated tests
- [ YES] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,31 +251,33 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+                    new Object[]{keyMutator.apply(entry.getKey()),
+                            valueMutator.apply(entry.getValue())},
+                    serializer);
+
+            each.serialize(b);
+            numBytesPerCheckpointEntry += b.writerIndex();
+
+            if(numBytesPerCheckpointEntry > rt.getParameters().getMaxWriteSize()) {
+                convertToCheckpointEntry(smrEntries);
+                numBytesPerCheckpointEntry = b.writerIndex();
```
Lujie1996, line 270: Is this `numBytesPerCheckpointEntry = 0;`?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -207,6 +207,22 @@ private long nonCachedAppend(Object object, UUID ... streamIDs) {
         return sv.append(object, null, CacheOption.WRITE_AROUND, streamIDs);
     }
 
+    /*
+     * Pack smr entries into one checkpoint entry
+     */
+    private void convertToCheckpointEntry(MultiSMREntry smrEntries) {
```
Lujie1996, line 213: At the first glimpse I saw the name `convertToCheckpointEntry`, I thought it should have a return value. I have to look inside the function to realize that it performs both conversion and appending of a CheckpointEntry. So perhaps rename it to `convertAndAppendCheckpointEntry` or just `appendCheckpointEntry` since append should already include the step of conversion. Also update the comment to reflect this.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,31 +251,33 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
-                smrEntries.addTo(new SMREntry("put",
-                        new Object[]{keyMutator.apply(entry.getKey()),
-                                valueMutator.apply(entry.getValue())},
-                        serializer));
-                entryCount++;
+        int numBytesPerCheckpointEntry = 0;
+        ByteBuf b = Unpooled.buffer();
+        MultiSMREntry smrEntries = new MultiSMREntry();
+
+        for (Map.Entry entry : (Iterable<Map.Entry>) entryStream::iterator) {
```
Lujie1996, line 259: We removed the 50 entries limit here. We need to confirm with Maithem or someone else that this removal does not possibly affect anywhere else in Corfu.

Lujie1996 (commented):
Can we reproduce this bug and try using this PR to fix it?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,31 +251,33 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+            SMREntry each = new SMREntry("put",
+                    new Object[]{keyMutator.apply(entry.getKey()),
+                            valueMutator.apply(entry.getValue())},
+                    serializer);
+
+            each.serialize(b);
+            numBytesPerCheckpointEntry += b.writerIndex();
+
+            if(numBytesPerCheckpointEntry > rt.getParameters().getMaxWriteSize()) {
```
Maithem, line 268: This is not accurate. You need to check for `b.writerIndex() + numBytesPerCheckpointEntry > rt.getParameters().getMaxWriteSize()`, if that is true that you write it out, otherwise you add it to  smrEntries and adjust your accumulator.

Maithem: I would just write out the batch inside the loop body and use `continue`, it simplifies the control-flow logic.

Maithem (commented):
Good start. Please address comments and add tests.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Lujie1996: You're right. Got it.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: +1 to Maithem's suggestion that would ease readability..

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
annym: With regards to this, there are some tests that rely on this to verify an entry is of type continuation, those tests need to accordingly be changed, like for example: checkpointWriterInterleavedTest

achinmishra: It did. The test"checkpointWriterInterleavedTest" was failing. After discussion with Maithem on the zoom call, I am going to maintain a local batch size counter, specially for the validity by the test.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: In order to overlap the resetting of the bookkeeping data structures and utilizing the same loop to consider the SMR entry that is read from the stream (once its read from the stream, the entry cannot be re-read after a "continue" in a loop, I have come up with this control flow. I can add more comments into it for clarification around the control flow.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: @annym Thanks for pointing it out. I am going to take care of it in the next update I am going to make to this pull request

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,31 +251,33 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
-            numEntries++;
-            // CheckpointEntry::serialize() has a side-effect we use
-            // for an accurate count of serialized bytes of SRMEntries.
-            numBytes += cp.getSmrEntriesBytes();
+            smrEntries.addTo(each);
+            entryCount++;
+            b.clear();
         }
 
+        if(!smrEntries.getUpdates().isEmpty())
```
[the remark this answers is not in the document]
achinmishra: Thanks for pointing it out @annym

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: Good suggestion. Will make that change

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: @annym please see my above comment.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
pankti-m: b.writerIndex() will not have the correct value as the last entry 'each' has already been serialized inside it, right?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,72 @@ public void checkpointWriterInterleavedTest() throws Exception {
         }
     }
 
+    /** Test the CheckpointWriter write size limit.
+     *
+     * CheckpointWriter aggregates a batch of 50 SMREntries into one
```
pankti-m, line 467: should this comment be updated to reflect the new behavior, i.e., 50 entries or a limit of 25MB?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,72 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+        // each entry is 1MB
+        byte[] buf = new byte[base * base];
+        ThreadLocalRandom.current().nextBytes(buf);
+        String payload = new String(buf);
+
+        for (int i = 0; i < numKeys; i++) {
+            m.put(keyPrefix + i, payload);
+        }
+
+        // Set up CP writer.  Add fudgeFactor to all CP data,
```
pankti-m, line 494: what does this comment mean about the fudgefactor?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
achinmishra: I will make that change

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
achinmishra: @Lujie1996 can better answer this one. George?

Lujie1996: Removed.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
pankti-m: clarified offline

pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +255,49 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
-            long pos = nonCachedAppend(cp, checkpointStreamID);
-            postAppendFunc.accept(cp, pos);
-            numEntries++;
-            // CheckpointEntry::serialize() has a side-effect we use
-            // for an accurate count of serialized bytes of SRMEntries.
-            numBytes += cp.getSmrEntriesBytes();
         }
 
-        return entryCount;
+        //the entries which are left behind for a final flush
```
vjeko, line 294: Improper comment style: https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +255,49 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+                            valueMutator.apply(entry.getValue())},
+                    serializer);
+
+            each.serialize(b);
+            numBytesPerCheckpointEntry += b.writerIndex();
+
+            // consider a 2% less than writesize limit check on smr entries. Checkpoint entry
+            // adds more metadata to the actual size of SMR entries. Its a safeguard against the
+            // smr entries amounting to the actual boundary limit.
+            if(currentBatchSize == batchSize || numBytesPerCheckpointEntry >
```
vjeko, line 274: Space after `if`.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +255,49 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+                log.trace("Batched size of checkpoint log entry consists {} smr entries",
+                        smrEntries.getUpdates().size());
+                //reset the num of bytes and the new batch size entries below
+                //also, reset the smr entry to add the newly read SMR:each from the stream
+                numBytesPerCheckpointEntry = b.writerIndex();
+                smrEntries = new MultiSMREntry();
+                currentBatchSize = 0;
             }
+            smrEntries.addTo(each);
+            //maintain current batch size only for test purposes
```
vjeko, line 287: Improper comment style: https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+            Map<String, String> m2 = instantiateStringMap(streamName);
+            for (int i = 0; i < numKeys; i++) {
+                assertThat(m2.get(keyPrefix + i)).describedAs("get " + i)
+                        .isEqualTo(m.get(keyPrefix + i) + mutationSuffix);
+            }
+        } finally {
+            r.getObjectsView().TXEnd();
+        }
+    }
+
```
vjeko, line 522: Missing test description.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+            System.out.println(String.format("After putting entry %s, numBytesPerCheckpointEntry is: %s",
+                    (keyPrefix + i), numBytesPerCheckpointEntry));
+            System.out.println("nextSmrEntrySize: " + nextSmrEntrySize);
+            System.out.println("---------------------------------");
+            i++;
+        }
+
+        System.out.println("maxWriteSize: " + getRuntime().getParameters().getMaxWriteSize());
+        System.out.println("Total numBytesPerCheckpointEntry is: " + numBytesPerCheckpointEntry);
+        System.out.println("(1 - marginRatio) * maxWriteSize is: " +
```
vjeko, line 564: I don't think we use `System.out.println` in testing either. Use a logger, and if you are not want to do debugging, modify the appropriate XML file (I think it's `logback-classic.xml`)

vjeko (commented):
Looks good -- just make sure you follow appropriate code style guidelines. Same goes for your commit message.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: Added. Thanks

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+            String payload = getRandomStringOfSize(1 << 20);
+            m.put(keyPrefix + i, payload);
+        }
+
+        // set the max write size to 25 MB
+        getRuntime().getParameters().setMaxWriteSize(maxWriteSize);
+        CheckpointWriter cpw = new CheckpointWriter(getRuntime(), streamId, author, (CorfuTable) m);
+        cpw.setSerializer(serializer);
+        cpw.setValueMutator((l) -> (String) l + mutationSuffix);
+        // use default batch size which is 50
```
annym, line 496: is this comment out of place?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: We actually use `System.out.println` in other unit tests here (`consumeRawStreamCheckpointTrim` and `getCheckpointId`).

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+            // each entry is 1 MB
+            String payload = getRandomStringOfSize(1 << 20);
+            m.put(keyPrefix + i, payload);
+        }
+
+        // set the max write size to 25 MB
+        getRuntime().getParameters().setMaxWriteSize(maxWriteSize);
+        CheckpointWriter cpw = new CheckpointWriter(getRuntime(), streamId, author, (CorfuTable) m);
+        cpw.setSerializer(serializer);
+        cpw.setValueMutator((l) -> (String) l + mutationSuffix);
```
annym, line 495: Can we add a comment on why this mutator is placed, for readability...

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+                .type(TransactionType.SNAPSHOT)
+                .build()
+                .begin();
+        Token snapshot = TransactionalContext
+                .getCurrentContext()
+                .getSnapshotTimestamp();
+        try {
+            cpw.startCheckpoint(snapshot);
+            cpw.appendObjectState(m.entryStream());
+            cpw.finishCheckpoint();
```
annym, line 509: in line with my previous comment, once we are sure we have compressed data > 25MB, we should confirm we have more than one CONTINUATION record... that way we can validate our approach better...

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+        final int maxWriteSize = 25 << 20;
+        final double marginRatio = 0.02;
+
+        StreamingMap<String, String> m = instantiateStringMap(streamName);
+
+        // set the max write size to 25 MB
+        getRuntime().getParameters().setMaxWriteSize(maxWriteSize);
+        CheckpointWriter cpw = new CheckpointWriter(getRuntime(), streamId, author, (CorfuTable) m);
+        cpw.setSerializer(serializer);
+        // use default batch size which is 50
```
annym, line 546: same here, out of place comment?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+        final String author = "Me, myself, and I";
+        final String mutationSuffix = "_mutation_suffix";
+        // max write size set to 25MB
+        final int maxWriteSize = 25 << 20;
+
+        StreamingMap<String, String> m = instantiateStringMap(streamName);
+
+        for (int i = 0; i < numKeys; i++) {
+            // each entry is 1 MB
+            String payload = getRandomStringOfSize(1 << 20);
```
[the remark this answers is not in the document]
Maithem: Probably need to test both code paths, where compression is enabled and disabled. For the case where compression is enabled, you can either assert that the compressed size is not less than the uncompressed size, or keep generating random strings until that condition is met.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -47,8 +47,12 @@
     private LocalDateTime startTime;
     private long startAddress;
     private long endAddress;
+    @Getter
     private long numEntries = 0;
     private long numBytes = 0;
+    //SMR entries total size falling on EXACTLY on the write size limit set will create problems
+    //Making sure there is 2% room for the check to create two long check point entries instead
```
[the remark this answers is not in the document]
Maithem: I know that I mentioned `2%` overhead at some point but you shouldn't depend on this number. I would just opt for simplicity and not account for the overhead at all. Since the payload size can be tuned I feel like this is unnecessary (because its not accurate).

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -457,6 +462,213 @@ public void checkpointWriterInterleavedTest() throws Exception {
         }
     }
 
+    /** Test the CheckpointWriter write size limit.
+     *
```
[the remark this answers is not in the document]
Lujie1996: Hi Anny. I wrote the test and I might miss some comments you had with Achin. Removed 50 here.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: This comment is not for any line of code. Cpw  will set batch size to 50 by default. I added this comment just to highlight that the batch size is 50 here. But now seems like this information is not very related here. Removed.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: Sure. I reused the other checkpointWriter test code and it has this mutator. I think it's good to have it in this test as well since it verifies that checkpointWriter did touch SMREntries.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: Removed

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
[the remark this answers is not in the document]
Lujie1996: I'm not sure if I understand correctly. With the new code which guarantees the uncompressed size of a batch is always smaller than 25 MB, it is impossible to make compressed size > 25 MB in `StreamView::append`. We are testing the old code here since the implementation of `CheckpointWriter::appendObjectState` has changed and it does not blindly combine 50 entries into one. We can't reproduce or test the bug with this PR.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: The number of `CheckpointEntry.CONTINUATION` is not determined by the compressed data size. It is determined by the size of serialized SMR objects. The way I generated test data is based on `CheckpointWriter::appendObjectState`.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -47,8 +47,12 @@
     private LocalDateTime startTime;
     private long startAddress;
     private long endAddress;
+    @Getter
     private long numEntries = 0;
     private long numBytes = 0;
+    //SMR entries total size falling on EXACTLY on the write size limit set will create problems
```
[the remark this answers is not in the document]
achinmishra: done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -207,6 +211,22 @@ private long nonCachedAppend(Object object, UUID ... streamIDs) {
...
+        postAppendFunc.accept(cp, pos);
+        numEntries++;
+        // CheckpointEntry::serialize() has a side-effect we use
+        // for an accurate count of serialized bytes of SRMEntries.
+        numBytes += cp.getSmrEntriesBytes();
+    }
+
     /** Append zero or more CONTINUATION records to this
      *  object's stream.  Each will contain a fraction of
      *  the state of the object that we're checkpointing
```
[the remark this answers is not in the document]
achinmishra: done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +255,49 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
-                        new Object[]{keyMutator.apply(entry.getKey()),
-                                valueMutator.apply(entry.getValue())},
-                        serializer));
-                entryCount++;
+        int totalEntryCount = 0, currentBatchSize = 0, numBytesPerCheckpointEntry = 0;
+        ByteBuf b = Unpooled.buffer();
+        MultiSMREntry smrEntries = new MultiSMREntry();
+
+        for (Map.Entry entry : (Iterable<Map.Entry>) entryStream::iterator) {
+            SMREntry each = new SMREntry("put",
```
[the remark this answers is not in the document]
achinmishra: @vjeko I did not touch the "put" entry in the SMR object. I do not understand which variable here you are talking about.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +255,49 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+
+            each.serialize(b);
+            numBytesPerCheckpointEntry += b.writerIndex();
+
+            // consider a 2% less than writesize limit check on smr entries. Checkpoint entry
+            // adds more metadata to the actual size of SMR entries. Its a safeguard against the
+            // smr entries amounting to the actual boundary limit.
+            if(currentBatchSize == batchSize || numBytesPerCheckpointEntry >
+                    (rt.getParameters().getMaxWriteSize() -
+                            (BATCH_THRESHOLD_PERCENTAGE * rt.getParameters().getMaxWriteSize()))) {
```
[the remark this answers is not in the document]
achinmishra: 1. BATCH_THRESHOLD_PERCENTAGE: making this configurable would create issues where in the clients would end up shooting in the foot. The idea is to keep this internal to Checkpoint Writer keep use the maximum of the 'maxWriteSize' while writing the checkpoint entry.
2. Good suggestion. I made BATCH_THRESHOLD_PERCENTAGE = 0.98 and that will simply the if condition check.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +255,49 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
-
-        for (List<Map.Entry> partition : partitions) {
-            MultiSMREntry smrEntries = new MultiSMREntry();
-            for (Map.Entry entry : partition) {
-                smrEntries.addTo(new SMREntry("put",
-                        new Object[]{keyMutator.apply(entry.getKey()),
-                                valueMutator.apply(entry.getValue())},
-                        serializer));
-                entryCount++;
+        int totalEntryCount = 0, currentBatchSize = 0, numBytesPerCheckpointEntry = 0;
```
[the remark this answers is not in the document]
achinmishra: done!

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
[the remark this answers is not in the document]
achinmishra: correct. done

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
achinmishra: done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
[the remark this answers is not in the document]
achinmishra: @Maithem wouldnt a safety guard here better? Hypothetically speaking, there are chances of hitting the boundary even when the payload size is tunable and although, the chances of hitting this exception again after the payload is tuned and re-tried is slim, it still exists. Adding this measure would remove that slim chance altogether in my opinion.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +255,49 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+                    serializer);
+
+            each.serialize(b);
+            numBytesPerCheckpointEntry += b.writerIndex();
+
+            // consider a 2% less than writesize limit check on smr entries. Checkpoint entry
+            // adds more metadata to the actual size of SMR entries. Its a safeguard against the
+            // smr entries amounting to the actual boundary limit.
+            if(currentBatchSize == batchSize || numBytesPerCheckpointEntry >
+                    (rt.getParameters().getMaxWriteSize() -
```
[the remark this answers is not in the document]
achinmishra: done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +258,50 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+                            valueMutator.apply(entry.getValue())},
+                    serializer);
+
+            each.serialize(b);
+            numBytesPerCheckpointEntry += b.writerIndex();
+
+            /* consider a 2% less than writesize limit check on smr entries. Checkpoint entry
+             * adds more metadata to the actual size of SMR entries. Its a safeguard against the
+             * smr entries amounting to the actual boundary limit. */
+            if (currentBatchSize == batchSize || numBytesPerCheckpointEntry >= maxWriteSizeLimit) {
```
annym, line 280: Q. If the first comparison is there only for 'testing' purposes (like suggested in the comments) aren't we then limiting production code to this? cause if that condition is met before the size limit we won't be taking advantage of this change... I don't see any issue as that was how it was working before, I'm just asking if we have thought it through and we'd prefer to adjust our tests cause it's not that relevant or if its ok..

annym: It would be only >? cause if it's equal to we could still add it to this batch, right?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -456,6 +461,218 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+
+    /**
+     * CheckpointWriter has BATCH_THRESHOLD_PERCENTAGE which decreases the
+     * writeSizeLimit to (1-BATCH_THRESHOLD_PERCENTAGE) * writeSizeLimit.
+     * This test checks that a new checkpointEntry is created when the
+     * total size of SMREntries exceeds the decreased writeSizeLimit.
+     */
+    @Test
+    @SuppressWarnings("checkstyle:magicnumber")
+    public void checkpointWriterSizeLimitMarginTest() throws Exception {
```
annym, line 530: Just to confirm, I'm guessing this test has compression disabled, right? otherwise there is no way to make sure we're triggering this scenario...

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +258,50 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+        for (Map.Entry entry : (Iterable<Map.Entry>) entryStream::iterator) {
+            SMREntry each = new SMREntry("put",
+                    new Object[]{keyMutator.apply(entry.getKey()),
+                            valueMutator.apply(entry.getValue())},
+                    serializer);
+
+            each.serialize(b);
+            numBytesPerCheckpointEntry += b.writerIndex();
+
+            /* consider a 2% less than writesize limit check on smr entries. Checkpoint entry
```
annym, line 277: I discussed this offline with @Lujie1996 I think it's important you guys add a comment indicating that this is a conservative approach and that size limit is calculated on uncompressed data, meaning that the actual final write size (in the case that compression is enabled, could significantly fall below this size limit) ... that's why we had the confusion in the test, cause I had the impression this estimation was on compressed data..

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: The compression is enabled. This test is to verify that the new batching strategy works, and specifically verify that uncompressed size of a batch, which is going to be later compressed, is smaller than 98% * writeSizeLimit. Everything is uncompressed at this stage. We are not testing the size check that StreamView has.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Lujie1996: I thought that keeping the old check (number of entries) and adding a new check (numBytes) is safer than replacing the old one. I'm not sure if removing the batchSize limit would harm / impose side-effects on somewhere else in Corfu (just from my guess). I do totally agree that removing batchSize check will improve performance.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
annym: Oh when I discussed with Achin I thought the idea of this test was to test that when the size of the data + overhead exceeded the size limit we were fine... So we need to ensure that the data is exactly 98% such that when adding the overhead we don't get a WriteSizeLimitException, and the only way to control this is to have compression disabled, otherwise, 98% could turn in 60% + overhead is never near the boundary..

Lujie1996: Oh I finally see your point. So we generate some entries whose size add up to somewhere very close to 98% (still smaller) and verify that it doesn't trigger WriteSizeLimitException right?

annym: Correct. Well I would say exactly 98%, because we're assuming the 2% accounts for the overhead, so we want to test the border case, that is, the worst case when data + overhead could exceed the write size limit. Actually if you write this test and then change 0.98 to 1.0 and make your data 25 MB (or whichever your limit is) it should fail cause 100% + overhead > write size limit...

apavan:
@achinmishra description only talk about adding log entries up to a max threshold say 25MB or what ever configured limit is but it does not talk about the limit on number of entries. Did you consider if we should go the route of either max number or entries or max batch size which ever is less?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: Hi Anny, a new test case is added [here](https://github.com/CorfuDB/CorfuDB/pull/2875/files#diff-bd6de4f220c80c5a84d6c60f308074aac0a096efcefe915f124f15a932652cabR660). In this test, I disabled compression and changed batchThresholdPercentage to 100%, then generate entries whose total size is in range [maxWriteSize-100bytes, maxWriteSize) so that it could pass the first check and fail at the second check and triggers WriteSizeException due to the overhead of CheckpointEntry. Note that it's almost impossible to have total size of SMR entries exactly == 100% because we can't precisely decide the SMR entry size. Although we can provide a buf as `byte[] buf = new byte[size]`, the SMR entry that encapsulates this buf is unknown.

achinmishra:
@apavan With configurable entries, the entire computation would still come down to the calculation based off of maxWriteSize limit with current implementation. However, I can think of an additional runtime configurable parameter, as you have suggested, that could potentially hold the number of entries per checkpoint entry and provide the aforementioned logic. 
The downsides I could envision, based on the knowledge I have so far: 
a. different size of checkpoint entries might lead to different performances while serializing and de-serializing these entries as those would be size dependent. 
b. under utilizing of the LogUnit for a checkpoint entry.

We would need to explore more if going by your suggested route could/might have other implications and how this could be useful in the future, provided currently this is a corner case.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: I was under the same impression that the test had compression disabled. let me add the comment.

annym (approved):

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Lujie1996: @Maithem As per Anny's suggestion, removing the limit of max number of entries would improve the checkpoint performance, could you give your thoughts on this? From the fact that WriteSizeException has not been triggered until very recently, the total numBytes must be much lower than 25MB under the old 50 batch size limit, which means it has been underusing the stream capacity.  If having only one limit on numBytes does not harm anywhere else in Corfu, I think it would be an optimization for the slow checkpointing process today.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -47,8 +47,15 @@
...
     private long endAddress;
+    @Getter
     private long numEntries = 0;
     private long numBytes = 0;
+    // SMR entries total size falling on EXACTLY on the write size limit set will create problems
+    // Making sure there is 2% room, a safety guard to account for checkpoint entry overhead
+    // (metadata)
+    @Setter
+    @Getter
+    private double BATCH_THRESHOLD_PERCENTAGE = 0.98;
```
Maithem, line 58: This seems arbitrary. I would just suggest using `sizeOf` on an empty checkpoint to calculate it exactly.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem: I think it can improve performance, but the current state is probably going to decrease throughput. Imagine the case where you have a batch of 50 entries and that each entry is 2500000 bytes, then this algorithm will generate a batch size of 10, but the existing algorithm would generate a batch size of 50. Assuming a compression rate of 5x, this change will decrease throughput in comparison to what's on master. 

What is missing here is that the serialized size is accounted for before compression and that is not accurate.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: Resolving this after our discussion. For now, we agree to go with 0.95

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +261,65 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+                    rt.getParameters()
+                            .getCodecType()
+                            .getInstance().compress(ByteBuffer.wrap(inputBuffer.array(),
+                            inputBuffer.readerIndex(), inputBuffer.writerIndex()));
+
+            //System.out.println("ACHIN  numBytes: " + numBytesPerCheckpointEntry);
+            //System.out.println(" compressed buffer position: " + compressedBuffer.position());
+            //System.out.println("original size: " + inputBuffer.writerIndex());
+
+            numBytesPerCheckpointEntry += compressedBuffer.position();
```
Maithem, line 292: Since the compress method flips the ByteBuffer after it has been compressed the size is the `limit` of the buffer and not the `position`. Should be `numBytesPerCheckpointEntry += compressedBuffer.limit()`

After `ByteBuffer::flip` the limit is set to the position and the position is set to 0 to allow a read to consume the buffer.

achinmishra: Understood. Thanks a ton!

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +261,60 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+            smrPutEntry.serialize(inputBuffer);
+            ByteBuffer compressedBuffer =
+                    rt.getParameters()
+                            .getCodecType()
+                            .getInstance().compress(ByteBuffer.wrap(inputBuffer.array(),
+                            inputBuffer.readerIndex(), inputBuffer.writerIndex()));
+
+            numBytesPerCheckpointEntry += compressedBuffer.limit();
+
+            /* this is a conservative approach, size limit is calculated on uncompressed data
```
annym, line 290: this comment is outdated, correct? Now it seems the calculation is made on compressed data...

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +261,60 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+             * consider a 5% less than writesize limit check on smr entries. Checkpoint entry
```
annym, line 291: I thought one of the points discussed was not to assume a 5% but actually calculate the checkpoint entry overhead? (which I believe is fixed)

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +261,60 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
+             * adds more metadata to the actual size of SMR entries. Its a safeguard against the
+             * smr entries amounting to the actual boundary limit. */
+            if (currentBatchSize == batchSize || numBytesPerCheckpointEntry >= maxWriteSizeLimit) {
```
annym, line 294: I think > is enough? instead of >=

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -47,8 +48,15 @@
     private LocalDateTime startTime;
     private long startAddress;
     private long endAddress;
+    @Getter
     private long numEntries = 0;
     private long numBytes = 0;
+    // SMR entries total size falling on EXACTLY on the write size limit set will create problems
+    // Making sure there is 5% room, a safety guard to account for checkpoint entry overhead
+    // (metadata)
```
Maithem, line 56: Consider rewarding this. Maybe "Creates a batch that is 95% of the the max write size to accommodate for metadata overhead"

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -207,10 +215,28 @@ private long nonCachedAppend(Object object, UUID ... streamIDs) {
         return sv.append(object, null, CacheOption.WRITE_AROUND, streamIDs);
     }
 
+    /*
+     * Pack smr entries into one checkpoint entry
+     */
```
Maithem, line 220: "Create a checkpoint continuation record from a MultiSMREntry"

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +261,60 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
         ImmutableMap<CheckpointEntry.CheckpointDictKey, String> mdkv =
```
Maithem, line 261: On master `mdkv` was copied and passed to `appendObjectState` , but its not used in this patch.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +261,60 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
-                entryCount++;
+        int totalEntryCount = 0;
+        int currentBatchSize = 0;
+        int numBytesPerCheckpointEntry = 0;
+        int maxWriteSizeLimit = (int) (BATCH_THRESHOLD_PERCENTAGE * rt.getParameters().getMaxWriteSize());
+
+        ByteBuf inputBuffer = Unpooled.buffer();
+        MultiSMREntry smrEntries = new MultiSMREntry();
+
+        for (Map.Entry entry : (Iterable<Map.Entry>) entryStream::iterator) {
```
Maithem, line 272: Casting to `Iterable` just to use an enhanced-for loop seems unnecessary here.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -235,32 +261,60 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
...
-                smrEntries.addTo(new SMREntry("put",
-                        new Object[]{keyMutator.apply(entry.getKey()),
-                                valueMutator.apply(entry.getValue())},
-                        serializer));
-                entryCount++;
+        int totalEntryCount = 0;
+        int currentBatchSize = 0;
+        int numBytesPerCheckpointEntry = 0;
+        int maxWriteSizeLimit = (int) (BATCH_THRESHOLD_PERCENTAGE * rt.getParameters().getMaxWriteSize());
+
```
Maithem, line 268: what if getMaxWriteSize is not set, that is, it will be `Integer.MAX_SIZE` applying the threshhold seems a little weird here.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem, line 294: Can you just go with one batching strategy, having both seems a little complicated and unnecessary.

Maithem: Yes, >= will emit an extra CONT record if numBytesPerCheckpointEntry == maxWriteSizeLimit, but it shouldn't

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -456,6 +462,293 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+        final String streamName = "mystream5";
+        final UUID streamId = CorfuRuntime.getStreamID(streamName);
+        final String keyPrefix = "a-prefix";
+        final int numKeys = 50;
+        final String author = "Me, myself, and I";
+        final String mutationSuffix = "_mutation_suffix";
+        // max write size set to 25MB
+        final int maxWriteSize = 25 << 20;
+
+        StreamingMap<String, String> m = instantiateStringMap(streamName);
```
Maithem, line 483: I would prefer to use mocks instead of this.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -456,6 +462,293 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+            cpw.finishCheckpoint();
+
+            // Instantiate new runtime & map.
+            setRuntime();
+            Map<String, String> m2 = instantiateStringMap(streamName);
+            for (int i = 0; i < numKeys; i++) {
+                assertThat(m2.get(keyPrefix + i)).describedAs("get " + i)
+                        .isEqualTo(m.get(keyPrefix + i) + mutationSuffix);
+            }
+        } finally {
```
Maithem, line 518: This test is incorrect. `instantiateStringMap` will return the same reference for m2 and m

Maithem: Probably need to add a trim after the checkpoint too.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -456,6 +462,293 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+        while (numBytesPerCheckpointEntry + nextSmrEntrySize < batchThresholdPercentage * getRuntime().getParameters().getMaxWriteSize()) {
+            m.put(keyPrefix + i, payload);
+            payload = getRandomStringOfSize(1 << 20);
+            numBytesPerCheckpointEntry += nextSmrEntrySize;
+            nextSmrEntrySize = getSerializedSMREntrySize(keyPrefix + i, payload,
+                    cpw.getKeyMutator(), cpw.getValueMutator(), cpw.getSerializer());
+            System.out.println(String.format("After putting entry %s, numBytesPerCheckpointEntry is: %s",
+                    (keyPrefix + i), numBytesPerCheckpointEntry));
+            System.out.println("nextSmrEntrySize: " + nextSmrEntrySize);
+            System.out.println("---------------------------------");
```
Maithem, line 564: Don;t use logging statements. Replace them with assertions.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -456,6 +462,293 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+                .getCurrentContext()
+                .getSnapshotTimestamp();
+        try {
+            cpw.startCheckpoint(snapshot);
+            cpw.appendObjectState(m.entryStream());
+            cpw.finishCheckpoint();
+
+            // Instantiate new runtime & map.
+            setRuntime();
+            Map<String, String> m2 = instantiateStringMap(streamName);
```
Maithem, line 644: This will return the original map reference, its not a new map.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
```diff
@@ -456,6 +462,293 @@ public void checkpointWriterInterleavedTest() throws Exception {
...
+                payload = getRandomStringOfSize(payloadLength);
+                nextSmrEntrySize = getSerializedSMREntrySize(keyPrefix + i, payload,
+                        cpw.getKeyMutator(), cpw.getValueMutator(), cpw.getSerializer());
+                continue;
+            }
+            m.put(keyPrefix + i, payload);
+            numBytesPerCheckpointEntry += nextSmrEntrySize;
+            System.out.println(String.format("After putting entry %s (payload len is %s, size is %s), numBytesPerCheckpointEntry is: %s",
+                    (keyPrefix + i), payloadLength, nextSmrEntrySize, numBytesPerCheckpointEntry));
+            System.out.println("-------------------------------------------------------------");
```
Maithem, line 710: Logging shouldn't be used in tests, and especially `System.out.`. Instead, these should be asserts.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: We need a real corfu map so that we can perform checkpointing later.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: You are right. Comparing `m` and `m2` is not meaningful. I changed to compare `m2` with a java hashmap object which holds the entries we originally put into `m`. As for trimming after checkpoint, following how other tests in this test suite are written (e.g. checkpointWriterTest) I didn't trim log here. I think these tests are purely for checkpointing.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: Done. Please see the reply above.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: These prints are for debugging only. I'll just remove them.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
achinmishra: Resolving this conversation.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
achinmishra: Thanks @Lujie1996

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: @Maithem The other batchsize strategy is here only for the test checkpointWriterInterleavedTest. I had a chat session with you over this. We either change the test or keep this around for only this test. What would be your suggestion here?

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
achinmishra: @Lujie1996 should I remove this out?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: @Maithem Do you have a better suggestion here? I got rid of one Iterable.partitions call from the original change and this is something I came up with that seemed a better version and suitably aligned to this change. Can you help me understand what might be better here with an example?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: @Maithem This is missed out. If you check the method "convertAndAppendCheckpointEntry()", it is using the class variable mkdv successfully.  I do not understand the purpose of creating a shallow copy of mdkv into a local variable for checkpoint entry, when the class variable mdkv could have been used instead. Am I missing something here? For cleaning purpose I should remove this statement.

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Lujie1996: I think it's already removed in my most recent commit https://github.com/CorfuDB/CorfuDB/pull/2875/commits/d02a4512ea51298b833370593ee9c23578171705

test/src/test/java/org/corfudb/runtime/checkpoint/CheckpointSmokeTest.java
Maithem: I think you need to compare the hashmap with a new instance of CorfuTable. The CorfuTable refered by `m2` is the same as `m`

Lujie1996: Hi Maithem. `m2` is instantiated using a new corfu runtime (https://github.com/CorfuDB/CorfuDB/pull/2875/files#diff-bd6de4f220c80c5a84d6c60f308074aac0a096efcefe915f124f15a932652cabR514). Therefore, `m2` should be a new instance of CorfuTable, right?

Maithem: Didn't notice that `setRuntime();` switches the references. I think you're right it is a new map.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: Done.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: Done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: Done

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: I think so, but seems they had a discussion and agreed to have the 95%.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: yes, the calculation is made on compressed data

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: Modified, now each continuation chunk will use the copy of mdkv.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: Changed to use an iterator.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
zhangn49: Added a helper function to have a default 25MB write size if the MaxWriteSize is not set

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
achinmishra: @Maithem I have removed the batchSize and currentBatchSize objects altogether and its dependencies. Made minor changes to the tests and everything looks good.

zhangn49 (approved):

achinmishra:
@Maithem I have verified the changes. I will do a rebase and squash all the commits into a single one. Kindly merge it after.

Maithem:
@achinmishra codacy is failing

achinmishra:
@Maithem Codacy is failing due to "Avoid long methods" in the CheckpointSmokeTest. Its a test thats too long for Codacy.

PR #2877, opened 2021-02-09
Reduce Protobuf LogData Payload Size

zfrenette:
Since the backing array of a ByteBuf can be larger than the number of bytes written, this PR reduces unnecessary copying of data when serializing LogData into Protobuf by using appropriate bounds.

## Overview

Description:

Why should this be merged: 

- [X] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
@@ -29,7 +29,7 @@ public final class CorfuProtocolLogData {
         logData.doSerialize(buf);
 
         return LogDataMsg.newBuilder()
-                .setEntry(ByteString.copyFrom(buf.resetReaderIndex().array()))
+                .setEntry(ByteString.copyFrom(buf.array(), buf.readerIndex(), buf.readableBytes()))
                 .build();
     }
 


runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -29,7 +29,7 @@ public static LogDataMsg getLogDataMsg(LogData logData) {
         logData.doSerialize(buf);
 
         return LogDataMsg.newBuilder()
-                .setEntry(ByteString.copyFrom(buf.resetReaderIndex().array()))
+                .setEntry(ByteString.copyFrom(buf.array(), buf.readerIndex(), buf.writerIndex()))
                 .build();
     }
```
xcchang, line 34: Do we want to copy from readerIndex to writerIndex? I thought the third argument of `copyFrom` is the number of bytes to copy.

zfrenette: Fixed, Thanks!

xcchang (approved):
LGTM!

runtime/src/main/java/org/corfudb/protocols/CorfuProtocolLogData.java
```diff
@@ -29,7 +29,8 @@ public static LogDataMsg getLogDataMsg(LogData logData) {
         logData.doSerialize(buf);
 
         return LogDataMsg.newBuilder()
-                .setEntry(ByteString.copyFrom(buf.resetReaderIndex().array()))
+                .setEntry(ByteString.copyFrom(buf.array(), buf.readerIndex(),
+                        buf.writerIndex() - buf.readerIndex()))
```
Maithem, line 33: You can just do this `ByteString.copyFrom(buf.array(), buf.readerIndex(), buf.readableBytes()`

Maithem (approved):

Issue #2740, opened 2020-08-28
verifyStateTransferWithChainHeadFailure Intermittent Failure


annym:
## Overview

Observed intermittent failure of test verifyStateTransferWithChainHeadFailure. 

Logs:  [https://api.travis-ci.org/v3/job/722121113/log.txt](https://api.travis-ci.org/v3/job/722121113/log.txt)

[ERROR] Failures: 
1471[ERROR]   StateTransferIT.verifyStateTransferWithChainHeadFailure:84->verifyStateTransferWithNodeFailure:188 
1472Expecting code not to raise a throwable but caught a
1473  <java.util.concurrent.CompletionException>
1474with message :
1475  "org.corfudb.runtime.exceptions.OverwriteException"

annym:
4 tests in StateTransferIT are intermittently failing:

[Link to Failures](https://github.com/CorfuDB/CorfuDB/pull/2941/checks?check_run_id=2463292715)

Error:  Errors: 
Error:    StateTransferIT.verifyStateTransferWithChainHeadFailure » Completion java.util...
Error:    StateTransferIT.verifyStateTransferWithChainHeadRestart » Completion java.util...
Error:    StateTransferIT.verifyStateTransferWithChainTailFailure » Completion java.util...
Error:    StateTransferIT.verifyStateTransferWithChainTailRestart » Completion java.util...

PR #2878, opened 2021-02-09
Fix flaky ST test

PavelZaytsev:
## Overview

Description:

Fix flaky integration test.


xcchang (approved):
LGTM, thanks Pavel!

test/src/test/java/org/corfudb/integration/StateTransferIT.java
```diff
@@ -281,9 +282,14 @@ private void verifyStateTransferWithNodeRestart(int restartNode) throws Exceptio
...
-                TokenResponse token = writerRuntime.getSequencerView().next();
-                writerRuntime.getAddressSpaceView().write(token, "Test Payload".getBytes());
-                Sleep.sleepUninterruptibly(Duration.ofMillis(timeOut));
+                try {
+                    TokenResponse token = writerRuntime.getSequencerView().next();
+                    writerRuntime.getAddressSpaceView().write(token, "Test Payload".getBytes());
+                    Sleep.sleepUninterruptibly(Duration.ofMillis(timeOut));
+                }
+                catch (OverwriteException oe) {
+                    // Continue writing
```
annym, line 291: Just curious, with which writer does it collide such that we get Overwrite Exceptions? cause the two tests where startWriter is being used I don't see any other writer, perhaps I missed it...

Maithem:
Can we push this PR ? The tests keeps failing 

image

test/src/test/java/org/corfudb/integration/StateTransferIT.java
xnull, line 291: Looks like, this exception shouldn't appear, looks like we broke streamlog recently

xnull (approved):
ST test looks good

test/src/test/java/org/corfudb/integration/StateTransferIT.java
Maithem: If this exception is not expected and this test ignores it, then why approve?

Maithem: I think we should answer this question before merging.

Maithem: Also, how recent is "recently" ? This bug was reported back in august #2740

xcchang: Previously I had a meeting with Wenbin about this issue, and one possibility he suggested was, during the completion of `WriterFuture`, the Sequencer issued a new token, before the writer actually complete the write the Sequencer failed, and this token is issued again next time. So the writer might be collided with itself. Not sure if this is something @PavelZaytsev thinks of.

xcchang: But if that scenario is the case, the test could also fail when writing records to 2 node cluster (like line 151-154), currently I only noticed its failures from the completion of `writerFuture` (line 193). So hoping to learn more about this too...

PavelZaytsev: I was not able to repro it on my setup but in this particular case, I think the sequencer has issued the same token for the same address twice. In this test we kill the head node while writing to the cluster, the head node contains a sequencer. Sequencer internal state, the global tail was not updated before it got killed and brought back again, as a result, the writer could write to the same address twice. 
I don't remember seeing this issue with other tests - for example, verifyStateTransferWithChainTailFailure, where the tail is getting killed.

annym: Ok, I think I understand. Based on @PavelZaytsev comment, the scenario here is that lets say Token1 is issued (0, 5)->(epoch, address) to Client1... and precisely at that moment, the head node is killed (where the sequencer was running), but the write has not gone through yet... the new sequencer comes up, reconstructs his new maps and finds the tail is at 4, so he issues 5 to the same writer for its next write, in the meanwhile, Client1 (0,5) goes through, hence second time issued (1,5) now hits the OverwriteException...

Pavel, can you please add an elegant explanation of this in the catch.. just for future reference, cause it got us all thinking...

PR #2879, opened 2021-02-16
Improve the log for client-side Protobuf RPCs

xcchang:
## Overview

Description: Currently the log on client-side does not display request/response
types for outbound and inbound messages. This change can be helpful
for debug.

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -84,11 +85,18 @@ public boolean handle(ResponseMsg response, ChannelHandlerContext ctx) {
                     // For errors, we always want to completeExceptionally
                     errorHandlerMap.get(errorCase).handle(response, ctx, router);
                 } catch (Throwable e) {
+                    log.warn("Server threw exception for {} with request_id: {}",
+                            response.getPayload().getPayloadCase(),
+                            response.getHeader().getRequestId(),
+                            e);
```
zfrenette, line 91: Is this usage typically reserved for `log.error`?

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
```diff
@@ -84,11 +85,18 @@ public boolean handle(ResponseMsg response, ChannelHandlerContext ctx) {
...
                     router.completeExceptionally(requestId, e);
                 }
             }
         } else if (handlerMap.containsKey(payloadCase)) {
             try {
+                log.debug("Received {} from the server - {}",
+                        response.getPayload().getPayloadCase(),
+                        TextFormat.shortDebugString(response));
```
zfrenette, line 99: Note, if the whole message is logged, then we probably don't need to explicitly add the payload case. We can also guard the logging statements where `shortDebugString` is used.

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
zfrenette, line 91: I'm wondering if it might be worth logging the message a bit differently. Right now, if a message type is not contained in either handlerMap, then no message is logged whatsoever. Maybe instead we can just have one guarded logging statement outside of the if?

xcchang: I set to `warn` since previously in the `BaseHandler` there was:
```
    /**
     * Generic handler for a server exception.
     */
    @ClientHandler(type = CorfuMsgType.ERROR_SERVER_EXCEPTION)
    @Deprecated
    private static Object handleServerException(CorfuPayloadMsg<ExceptionMsg> msg,
                                                ChannelHandlerContext ctx, IClientRouter r) throws Throwable {
        log.warn("Server threw exception for request {}", msg.getRequestID(),
                msg.getPayload().getThrowable());
        throw msg.getPayload().getThrowable();
    }
```
I'm a little inclined to use `warn` since sometimes the server throws an exception doesn't necessarily mean there is an error?

zfrenette: Sorry, i was referring to the use of the parameter `e`.

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
xcchang: Added, thanks!

runtime/src/main/java/org/corfudb/runtime/clients/ClientResponseHandler.java
xcchang: We had an offline discussion and decide to keep the warning log here and move the guarded log out of the if clause.

xcchang: Removed, thanks!

zfrenette (approved):

Maithem (approved):

PavelZaytsev (approved):

PR #2884, opened 2021-02-19
Use Native Epoll Transport

Maithem:
## Overview
This patch allows for selecting an efficient transport implementation
during runtime. For example, when running on linux, EpollSocketChannel
will be used instead of NioServerSocketChannel, this allows for
better performance and reduced GC pressure.

Why should this be merged: Reduces GC pressure 

- [x] Public API has Javadoc


xcchang (approved):

pankti-m (approved):
it will be useful to add to the commit message why epoll is better in some cases and nio in others

Maithem:
> it will be useful to add to the commit message why epoll is better in some cases and nio in others

NIO uses epoll. It's just that one implementation generates more gc pressure than the other.

This is the git comment:

```
This patch allows for selecting an efficient transport implementation
during runtime. For example, when running on linux, EpollSocketChannel
will be used instead of NioServerSocketChannel, this allows for
better performance and reduced GC pressure.
```

annym (approved):

PR #2892, opened 2021-02-25
Miscellaneous cleanups with respect to static code analysis

xcchang:
## Overview

Description: We bring back Codacy for static code analysis. There are more than 200 issues marked with error-prone by Codacy. In this PR we fixed 32 of them, which are small miscellaneous changes. Some other issues might be non-trivial and will be solved in separate PRs.

Why should this be merged: We now have an A grade from Codacy's static code analysis and we can add the badge to our repo after this PR is merged :) 

- [x] Public API has Javadoc


it/src/main/java/org/corfudb/universe/UniverseManager.java
```diff
@@ -310,7 +310,7 @@ public void shutdown() {
         }
 
         private void checkMode(UniverseMode expected) {
-            if (universeMode != expected) {
+            if (!universeMode.equals(expected)) {
```
xnull, line 313: please don't use "equals" for enums, so, please change it back

xcchang: Can you clarify it? I went through some discussion about this topic https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals
I'm a little inclined to use equals here though, as if one of the variable is null, we may expect a NPE instead of IllegalStateException?

xnull: equals is too much, it's overkill. The same way you can check numbers: `Integer.of(1).equals(1)`. 
Anytime when you use algebraic data types it's better to use algebraic operators.

The same works for boolean, imagine this: `if(table.isEmpty().equals(true))` the code looks awkward

zhangn49 (approved):
Some nested if conditions were used to be more readable and clear, but it is good to clean up those static analysis issues.

test/src/test/java/org/corfudb/runtime/collections/SMRMultiLogunitTest.java
```diff
@@ -187,15 +187,14 @@ public void transactionalManyWritesThenRead() throws TransactionAbortedException
...
-                    System.out.println("base is " + base);
-                }
+            if (val != null && i == 0) {
+                int underscore = val.indexOf("_");
+                assertNotEquals(-1, underscore);
+                base = val.substring(0, underscore);
+                System.out.println("base is " + base);
             }
-            assertEquals(true, val.contains(base));
+            assert val != null;
```
zfrenette, line 196: I think if this assertion is moved before the if condition, then it doesn't need to be checked a second time and the condition can be simplified. You can also use `assertNotNull`.

test/src/test/java/org/corfudb/runtime/collections/SMRMultiLogunitTest.java
```diff
@@ -187,15 +187,14 @@ public void transactionalManyWritesThenRead() throws TransactionAbortedException
         for (int i=0; i < numKeys; i++) {
             String key = "key" + String.valueOf(i);
```
zfrenette, line 188: is `valueOf` necessary here?

xcchang: Done, thanks!

test/src/test/java/org/corfudb/runtime/collections/SMRMultiLogunitTest.java
xcchang: Done, thanks!

test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
```diff
@@ -148,7 +148,6 @@ public void testToArray() {
 
     @Test
     public void testEquality() {
-        assertThat(new StreamAddressSpace().equals(null)).isFalse();
```
zfrenette, line 151: Can you explain why this is being removed? It seems like this was explicitly testing the first condition in the `equals` method of `StreamAddressSpace`.

xcchang: The result of 'new StreamAddressSpace().equals(null)' is always 'false', since when we do `new StreamAddressSpace()` the object is always non-null.

xcchang: We had an offline discussion and decide to put this back, as it's a test for the first condition in the `equals` method of `StreamAddressSpace`.

zfrenette (approved):
Looks good overall, though I agree that some of the nested if conditions are more readable in their original forms.

infrastructure/Dockerfile
```diff
@@ -5,6 +5,6 @@ WORKDIR /app
 
 RUN apk add --update iptables sudo
 
-ADD target/${CORFU_JAR} /app/corfu.jar
+COPY target/${CORFU_JAR} /app/corfu.jar
```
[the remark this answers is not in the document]
xcchang: I thought it was from the new docker doc: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

> For other items (files, directories) that do not require ADD’s tar auto-extraction capability, you should always use COPY.

runtime/src/main/java/org/corfudb/runtime/object/VersionLockedObject.java
```diff
@@ -546,7 +546,7 @@ public String toString() {
      * @return an appropriate context
      */
     private ICorfuExecutionContext.Context getContext(ISMRStream stream) {
-        if (stream == optimisticStream) {
+        if (stream.equals(optimisticStream)) {
```
Maithem, line 549: This conversion is not always correct. One is a reference equality check and the other is a logical equality check. For example, this logic would return different results if there are two different objects (i.e, different references), but they are logically equal.

xcchang: Yeah I agree, but here `equals` could have the same  impact as `==` since we didn't override the `equals` method in either of `ISMRStream` implementation. It should be fine to keep this conversion here.

PR #2894, opened 2021-02-25
Fix getChildPIDs method in AbstractIT

xcchang:
## Overview

Description: This commit fixes the getChildPIDs method in AbstactIT. Formerly the return statement in finally clause conceals all the exceptions, and the recursive call for child PIDs does not work since it always produces ConcurrentModificationException.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/AbstractIT.java
+++ b/test/src/test/java/org/corfudb/integration/AbstractIT.java
@@ -54,8 +54,8 @@ public class AbstractIT extends AbstractCorfuTest {
     private static final String KILL_COMMAND = "pkill -9 -P ";
     private static final String FORCE_KILL_ALL_CORFU_COMMAND = "jps | grep -e CorfuServer -e CorfuInterClusterReplicationServer|awk '{print $1}'| xargs kill -9";
 
-    private static final int SHUTDOWN_RETRIES = 10;
-    private static final long SHUTDOWN_RETRY_WAIT = 500;
+    private static final int SHUTDOWN_RETRIES = 20;
+    private static final long SHUTDOWN_RETRY_WAIT = 1000;
 
     // Config the msg size for log replication data
     // sent from active cluster to the standby cluster.
--- a/test/src/test/java/org/corfudb/integration/AbstractIT.java
+++ b/test/src/test/java/org/corfudb/integration/AbstractIT.java
@@ -26,8 +26,6 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -225,41 +225,37 @@ public static void waitFor(Supplier<Boolean> supplier) throws InterruptedExcepti
      *
      * @param pid parent process identifier
      * @return list of children process identifiers
-     *
-     * @throws IOException
+     * @throws IOException {@inheritDoc}
+     * @throws InterruptedException {@inheritDoc}
      */
-    private static List<Long> getChildPIDs (long pid) {
+    private static List<Long> getChildPIDs (long pid) throws IOException, InterruptedException {
```
achinmishra, line 231: The method now throws these exceptions to the caller. I do not the caller of this method in this file meaning its probably being called outside of this file. Did you make sure all the callers are handling this exception and logging correctly, as a cautionary measure?

xcchang: I think there isn't any semantic changes from this change, since in the previous implementation the catch clause was just re-throwing the exceptions (concealed by the return in finally so we never got them). But yeah I will check the callers to make sure it, thanks!

test/src/test/java/org/corfudb/integration/AbstractIT.java
```diff
@@ -225,41 +225,37 @@ public static void waitFor(Supplier<Boolean> supplier) throws InterruptedExcepti
...
-                    previous = line;
-                    long childPID = Long.parseLong(line);
-                    childPIDs.add(childPID);
-                }
-            }
+        // Get child pid(s)
+        ProcessBuilder builder = new ProcessBuilder();
+        builder.command("sh", "-c", "pgrep -P " + pid);
+        Process p = builder.start();
+        p.waitFor();
```
Maithem, line 237: Can we get rid of this recursion?

`pkill -9 -P {pid}` would match all child process with a parent pid, why isn't that sufficient ?

Maithem (commented):
`runServer` doesn't track the process that it creates, it assumes that the test will track those pids and then later calls shutdown on them. Can we do this regardless, so everytime a process is created, we keep track of it and then after a test completes, we call kill on each of those pids. 
Junit has an after-test annotation that can facilitate this flow.

test/src/test/java/org/corfudb/integration/AbstractIT.java
xcchang: Sorry what does the `-9` mean? I didn't found this option.
For `-P` I thought only processes created by another process are considered to be the child process, so we may need to traverse the tree to get all child/grandchild processes.

xcchang: Oh it's the SIGKILL, I got you know. Thanks!

Maithem: `AbstractIT` launches the server by invoking the bash script `bin/corfu_server`. This script will start a process with the JVM. 

```
ps aux | grep corfu
maithem   243794  0.0  0.0   9632  3652 pts/1    S+   11:02   0:00 bash ./bin/corfu_server -ms 9000
maithem   243802 16.0  0.4 5863460 155444 pts/1  Sl+  11:02   0:01 /usr/lib/jvm/java-8-openjdk-amd64/bin/java -cp /home/maithem/CorfuDB/scripts/../infrastructure/target/infrastructure-0.3.1-SNAPSHOT-shaded.jar -Xmx2000m org.corfudb.infrastructure.CorfuServer -ms 9000

```
`pkill -9 -P {pid}` will kill the parent bash script and all child processes that have the same parent pid, so you don't need to implement the recursion here.

annym: Can you just confirm that for LR it kills all processes? I remember I had to add this cause there was no way the server got killed by killing just the parent process (but it could be that there was something else interfering at that time)

PR #2896, opened 2021-02-27
UFO Nested Secondary Index Bug Fix

annym:
## Overview

Bug fix: current code throws an Exception whenever a new record to add does not contain a value for a secondary index field (which is optional). This PR considers this case and includes a test to validate it. 

Description:

- Support the case when a field marked as secondary index is not set + test
- Move secondary index tests to its own file
- Add UFO Streaming API Unit Tests

Why should this be merged: existing bug affecting verticals

- [x] Public API has Javadoc


Lujie1996 (approved):

pankti-m (approved):

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -77,7 +77,7 @@
             lastNestedField = (i == (nestedFields.length - 1));
 
             if (nestedFieldDescriptor == null) {
-                throw new IllegalArgumentException(String.format("Secondary key %s, invalid field %s", indexPath, nestedFields[i]));
+                return Collections.EMPTY_LIST;
```
Maithem, line 80: Not sure I understand this. This means that a registered secondary index doesn't produce a mapping. When would that be the case? and why would it be useful to keep such index specs in the registry. 

```java

  // Map entry into secondary indexes
            for (Index.Spec<K, V, ?> index : indexSpec) {
                String indexName = index.getName().get();
                Map<Object, Map<K, V>> secondaryIndex = secondaryIndexes.get(indexName);
                for (Object indexKey : index.getMultiValueIndexFunction().apply(key, value)) {
                    Map<K, V> slot = secondaryIndex.computeIfAbsent(indexKey, k -> new HashMap<>());
                    slot.put(key, value);
                }
            }

```

Maithem: I see a case where the secondary index produces a non-valid value (i.e., null) and that is ignored and not indexed, but why would the descriptor be null ?

annym: The descriptor would be null if the 'indexed field' is not set on the recently added value... in this case there is simply nothing to add to the index..

Maithem: When would this happen? It might be confusing to half the table partially indexed.

annym: Consider in the file CorfuStoreSecondaryIndexTest, the test **testUnsetSecondaryKeyField**, where you have a table of Adult instances (type of the value) and there is a secondary index on the field 'age'... but an adult is added with no 'age' specified. In this case the new record has nothing to contribute to the nestedSecondaryIndex table... so we don't need to throw an error but instead return no actual entry to be added to the secondary index map, i.e, Collections.EMPTY_LIST...

Maithem: I see. Now I understand the fix, but what I'm not sure about is if this is a semantic that we should support. Especially, when some databases allow for nulls in the secondary index. I feel that the semantics of the secondary indexing APIs haven't been strictly defined yet. Another issue that is related is, should these secondary indexes also support projections?

Maithem (approved):

PR #2897, opened 2021-03-02
Streaming API Bug Fix

annym:
## Overview

Description:

With current code, subscribing to an empty table will cause a NPE and furthermore the subscriber
will be automatically unsubscribed, therefore, future updates are lost. Fix to this bug + test.

Why should this be merged: bug currently encountered by clients

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/StreamingIT.java
+++ b/test/src/test/java/org/corfudb/integration/StreamingIT.java
@@ -458,6 +458,70 @@ public class StreamingIT extends AbstractIT {
     }
 
     /**
+     * Test the case where a table is empty at the time of subscription, and that it is able
+     * to receive deltas once updates to the table happen.
+     */
+    @Test
+    @SuppressWarnings("checkstyle:magicnumber")
+    public void testSubscriberOnEmptyTables() throws Exception {
+        // Run a corfu server.
+        Process corfuServer = runSinglePersistentServer(corfuSingleNodeHost, corfuStringNodePort);
+
+        // Start a Corfu runtime.
+        runtime = createRuntime(singleNodeEndpoint);
+        CorfuStore store = new CorfuStore(runtime);
+
+        // Record the initial timestamp.
+        Timestamp ts1 = store.getTimestamp();
+        final String namespace = "test_namespace";
+        final String tableName = "table_test";
+
+        // Create 2 tables.
+        Table<Uuid, SampleTableAMsg, Uuid> tableA = store.openTable(
+                namespace, tableName,
+                Uuid.class, SampleTableAMsg.class, Uuid.class,
+                TableOptions.builder().build()
+        );
+
+        // Subscribe to streaming updates, while table has not been yet updated
+        StreamListenerImpl listener1 = new StreamListenerImpl("stream_listener_1");
+        store.subscribeListener(listener1, namespace, "sample_streamer_1",
+                Collections.singletonList(tableName), ts1);
+
+        // Wait for a while, so we are sure the subscriber poller has run
+        TimeUnit.SECONDS.sleep(2);
+
+        // Make some updates to the table
+        final int numUpdates = 5;
+        for (int index = 0; index < numUpdates; index++) {
+            try (TxnContext tx = store.txn(namespace)) {
+                Uuid uuid = Uuid.newBuilder().setMsb(index).setLsb(index).build();
+                SampleTableAMsg msgA = SampleTableAMsg.newBuilder().setPayload(String.valueOf(index)).build();
+                tx.putRecord(tableA, uuid, msgA, uuid);
+                tx.commit();
+            }
+        }
+
+        // Wait for a while, so we are sure the subscriber poller has run
+        TimeUnit.SECONDS.sleep(2);
+
+        // Verify updates
+        LinkedList<CorfuStreamEntries> updates1 = listener1.getUpdates();
+        assertThat(updates1).hasSize(numUpdates);
+
+        for (int index = 0; index < numUpdates; index++) {
+            assertThat(updates1.get(index).getEntries()).hasSize(1);
+            List<CorfuStreamEntry> entries = updates1.get(index).getEntries().values().stream().findFirst().get();
+            assertThat(entries).hasSize(1);
+            assertThat(((Uuid)entries.get(0).getKey()).getMsb()).isEqualTo(index);
+            assertThat(((SampleTableAMsg)entries.get(0).getPayload()).getPayload()).isEqualTo(String.valueOf(index));
+            assertThat(((Uuid)entries.get(0).getMetadata()).getMsb()).isEqualTo(index);
+        }
+
+        assertThat(shutdownCorfuServer(corfuServer)).isTrue();
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


achinmishra (commented):
I did a first pass and ran the test. Seems ok, although, in the added test I could not find a check if the listener was getting unsubscribed automatically as the bug description says. Also, 2 seconds seem less to test if the subscriber to the empty table got unsubscribed.

annym:
> I did a first pass and ran the test. Seems ok, although, in the added test I could not find a check if the listener was getting unsubscribed automatically as the bug description says. Also, 2 seconds seem less to test if the subscriber to the empty table got unsubscribed.

@achinmishra If you run the test without the fix, you'll get to see the updates after subscription are not received (because it unsubscribes upon the NPE exception), so verifying that it did not unsubscribe is to see if updates are still being received. Currently there is no API to check if you are subscribed (you could attempt to re-subscribe, but then with the fix this would throw an error). I think 2 seconds should be enough because it is scheduled to run every 50ms.

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -457,6 +457,70 @@ public void testMoreSubscribersAndSlowSubscribers() throws Exception {
...
+     * to receive deltas once updates to the table happen.
+     */
+    @Test
+    @SuppressWarnings("checkstyle:magicnumber")
+    public void testSubscriberOnEmptyTables() throws Exception {
+        // Run a corfu server.
+        Process corfuServer = runSinglePersistentServer(corfuSingleNodeHost, corfuStringNodePort);
+
+        // Start a Corfu runtime.
+        runtime = createRuntime(singleNodeEndpoint);
```
Maithem, line 471: I don't feel like this case deserves a separate unit test, can it be merged with `testStreamingSingleTable` ?

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -394,6 +394,13 @@ protected long getMaxGlobalFromMaxEntries(int maxEntries) {
         StreamAddressSpace streamAddressSpace = runtime.getSequencerView()
                 .getStreamAddressSpace(new StreamAddressRange(getId(), Address.MAX, getCurrentGlobalPosition()));
 
+        // If sequencer does not return an address map for this stream, it means it is an empty table, we can
+        // safely return because CorfuStore APIs have validated the table in question is present in the RegistryTable
+        // before hitting this API
+        if (streamAddressSpace == null) {
+            return Address.NON_ADDRESS;
+        }
+
```
Maithem, line 403: I think this is a bug in `SequencerServer::getStreamsAddressesMap`, it shouldn't return nulls for streams that are empty. It should return an empty `StreamAddressSpace` with the trimMark set to `Address.NON_EXIST`. Then part will take care of the empty table issue:

```java

long size = streamAddressSpace.size();
        if (size == 0L) {
            return Address.NON_ADDRESS;
        }

```

Maithem (commented):
This is really tricky to debug, it seems like the thread dies with an NPE without logging anything. Can you also change submit calls to the `StreamManager` 's `pollingExecutor` and `notificationExecutor` with `LambdaUtils::runSansThrow`

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: I would prefer to leave it as a separate test cause when a test is overloaded we tend to think the case is not covered.  Typically we (or at least me) just read the test names or quickly do a ctrl+find on key names...

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: I agree, the only concern I have here is wether if this would affect the cases where a stream really does not exist (other code paths), could it? if not I can move it there...

annym:
> This is really tricky to debug, it seems like the thread dies with an NPE without logging anything. Can you also change submit calls to the `StreamManager` 's `pollingExecutor` and `notificationExecutor` with `LambdaUtils::runSansThrow`

The error is reported on the streamListener onError method, there is actually where consumers should re-subscribe.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Done

runtime/src/main/java/org/corfudb/runtime/view/SequencerView.java
```diff
@@ -130,7 +130,8 @@ public TokenResponse next(UUID... streamIds) {
      * @return address space composed of the trim mark and collection of all addresses belonging to this stream.
      */
     public StreamAddressSpace getStreamAddressSpace(StreamAddressRange streamsAddressesRange) {
-        return getStreamsAddressSpace(Arrays.asList(streamsAddressesRange)).get(streamsAddressesRange.getStreamID());
+        StreamAddressSpace streamAddressSpace = getStreamsAddressSpace(Arrays.asList(streamsAddressesRange)).get(streamsAddressesRange.getStreamID());
+        return streamAddressSpace == null ? new StreamAddressSpace() : streamAddressSpace;
```
Maithem, line 134: I think, its better to set it in https://github.com/CorfuDB/CorfuDB/blob/86b7aae3ff1f6025b76cae1a54032d5322e04e83/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java#L829 because then you can set the trimMark to `Address.NON_EXIST` . Only the sequencer has that information, so it would make sense to construct it in the `SequencerServer` and not the `SequencerView`

Maithem (approved):

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -457,6 +457,70 @@ public void testMoreSubscribersAndSlowSubscribers() throws Exception {
...
+        );
+
+        // Subscribe to streaming updates, while table has not been yet updated
+        StreamListenerImpl listener1 = new StreamListenerImpl("stream_listener_1");
+        store.subscribeListener(listener1, namespace, "sample_streamer_1",
+                Collections.singletonList(tableName), ts1);
+
+        // Wait for a while, so we are sure the subscriber poller has run
+        TimeUnit.SECONDS.sleep(2);
+
```
zhangn49, line 493: Might be good to add an assertion here if PR is not urgent. 

`assertThatThrownBy(() -> store.subscribeListener(listener1, namespace, "sample_streamer_1",
                Collections.singletonList(tableName), ts1)).isExactlyInstanceOf(StreamSubscriptionException.class);`

zhangn49 (approved):
LGTM

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: I'll add it in PR 2896 which anyways I'll need to rebase, thanks!

PR #2907, opened 2021-03-10
StreamListener expose Timestamp for Re-Subscription

annym:
## Overview

Description: subscribers need a means to know the last synced timestamp. If an error occurs (for instance, a poison pill) they can attempt to re-subscribe from this point (instead of having to sync the complete space or use a timestamp which could incur in data loss).  

Why should this be merged: requirement from clients. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
@@ -9,6 +9,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.logprotocol.MultiObjectSMREntry;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
 import org.corfudb.runtime.exceptions.StreamSubscriptionException;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.ObjectsView;
@@ -272,7 +273,11 @@ public class StreamManager {
                                     ).collect(Collectors.toList())));
 
                     if (!entries.isEmpty()) {
-                        CorfuStreamEntries callbackResult = new CorfuStreamEntries(entries);
+                        Timestamp timestamp = Timestamp.newBuilder()
+                                .setSequence(logData.getGlobalAddress())
+                                .setEpoch(epoch)
+                                .build();
+                        CorfuStreamEntries callbackResult = new CorfuStreamEntries(entries, timestamp);
                         log.trace("{}::onNext with {} updates", listener.toString(), entries.size());
                         long onNextStart = System.nanoTime();
                         listener.onNext(callbackResult);
--- a/runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
@@ -6,6 +6,7 @@ import lombok.Getter;
 import org.corfudb.protocols.logprotocol.MultiObjectSMREntry;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
 import org.corfudb.runtime.view.StreamOptions;
 import org.corfudb.runtime.view.TableRegistry;
 import org.corfudb.runtime.view.stream.IStreamView;
@@ -137,7 +138,12 @@ public class StreamSubscription<K extends Message, V extends Message, M extends 
             return true;
         }
 
-        return streamBuffer.offer(new CorfuStreamEntries(streamEntries), maxBlockTime, TimeUnit.NANOSECONDS);
+        Timestamp timestamp = Timestamp.newBuilder()
+                .setSequence(logData.getGlobalAddress())
+                .setEpoch(epoch)
+                .build();
+
+        return streamBuffer.offer(new CorfuStreamEntries(streamEntries, timestamp), maxBlockTime, TimeUnit.NANOSECONDS);
     }
 
     /**
--- a/test/src/test/java/org/corfudb/integration/StreamingIT.java
+++ b/test/src/test/java/org/corfudb/integration/StreamingIT.java
@@ -73,27 +73,46 @@ public class StreamingIT extends AbstractIT {
      * A StreamListener implementation to be used in the tests.
      * This listener accumulates all updates streamed to it into a linked list that
      * can then be used to verify.
+     * This listener can throw an exception on 'n' number of updates, in order to
+     * trigger the onError(Throwable t) and consequently, automatically unsubscribe.
      */
     private class StreamListenerImpl implements StreamListener {
 
         @Getter
         private final String name;
 
+        private int updatesToError = 0;
+
+        @Getter
+        private Timestamp timestamp;
+
         @Getter
         private final LinkedList<CorfuStreamEntries> updates = new LinkedList<>();
+
+        private CountDownLatch latch = null;
 
         StreamListenerImpl(String name) {
             this.name = name;
         }
 
+        StreamListenerImpl(String name, int updatesToError, CountDownLatch latch) {
+            this(name);
+            this.updatesToError = updatesToError;
+            this.latch = latch;
+        }
+
         @Override
         public void onNext(CorfuStreamEntries results) {
             updates.add(results);
+            timestamp = results.getTimestamp();
+            if (updates.size() == updatesToError) {
+                throw new StreamSubscriptionException("Artificial exception to trigger onError");
+            }
         }
 
         @Override
         public void onError(Throwable throwable) {
-
+            latch.countDown();
         }
 
         @Override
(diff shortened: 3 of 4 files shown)


test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -73,27 +74,47 @@ public void loadProperties() {
...
         }
 
+        StreamListenerImpl(String name, int updatesToError, CountDownLatch latch) {
+            this(name);
+            this.updatesToError = updatesToError;
+            this.latch = latch;
+        }
+
         @Override
+        @SneakyThrows
```
Maithem, line 106: `StreamSubscriptionException` is an unchecked exception, is `SneakyThrows` required ?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -272,7 +273,11 @@ public void run() {
                                     ).collect(Collectors.toList())));
 
                     if (!entries.isEmpty()) {
-                        CorfuStreamEntries callbackResult = new CorfuStreamEntries(entries);
+                        Timestamp timestamp = Timestamp.newBuilder()
+                                .setSequence(logData.getGlobalAddress())
+                                .setEpoch(epoch)
+                                .build();
+                        CorfuStreamEntries callbackResult = new CorfuStreamEntries(entries, timestamp);
```
Maithem, line 280: Can the old streaming manager be removed ?

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
```diff
@@ -137,7 +138,12 @@ boolean enqueueStreamEntry(ILogData logData, long maxBlockTime) throws Interrupt
             return true;
         }
 
-        return streamBuffer.offer(new CorfuStreamEntries(streamEntries), maxBlockTime, TimeUnit.NANOSECONDS);
+        Timestamp timestamp = Timestamp.newBuilder()
+                .setSequence(smrEntries.getGlobalAddress())
```
Maithem, line 142: Since the epoch is derived from `logData` can you also use it for `gobalAddress`. The order of when `setGlobalAddress` is called on the payload is not clear.

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -73,27 +74,47 @@ public void loadProperties() {
      * A StreamListener implementation to be used in the tests.
      * This listener accumulates all updates streamed to it into a linked list that
      * can then be used to verify.
+     * This listener can throw an exception on 'n' number of updates, in order to
+     * trigger the onError(Throwable t) and consequently, automatically unsubscribe.
```
Maithem, line 78: What type of errors would cause the listener to briefly de-register then register again ?

`onError` should be used to notify the listen of corfu related errors, usually shutdown/trim exceptions. In those two cases, the listener will not resume from a know sequence number, but it will re-register from the current time. 

It's not clear how exposing this sequence number helps reconcile issues for the aforementioned type of errors.

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -610,6 +631,113 @@ public void testSubscriberOnEmptyTables() throws Exception {
...
+
+        // Subscribe to streaming for one 'tag' corresponding to TableA
+        // This listener will artificially throw an Exception when numUpdates/2 are received, enforcing unsubscription
+        // and testing the last provided timestamp truly tracks the last received update and can resume from that point
+        // onwards (no data loss or need to sync a full snapshot)
+        CountDownLatch latch = new CountDownLatch(1);
+        StreamListenerImpl listener1 = new StreamListenerImpl("stream_listener_1", numUpdates/2, latch);
+        store.subscribeListener(listener1, namespace, "sample_streamer_1",
+                Collections.singletonList(tableNameA), ts1);
+
```
Maithem, line 677: This timestamp is not reliable. In the case of a sequence regression, some deltas can be missed.

Maithem (commented):
Since the tracked streams are are continuously being checkpointed/trimmed I don't see a use case where a client wants to briefly register a listener, de-register and then register again at the last consumed address. Can you please elaborate on the use case ?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
annym: There is a pending effort to remove all deprecated APIs and old code but that will come in another PR.

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
annym: It should be the same, but yes no problem. Done.

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: yeah, not needed. Initially I was throwing a raw exception and that's why I used this. Removed.

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: not really, right? Cause this timestamp is obtained from the streaming API which is only called after a write is committed, so a sequencer regression cannot regress below this materialized point.

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: The scenario that they were discussing initially was the case of any Corfu failure from which they'd be able to recover (I'm trying to think what this could be, I can only think of some race condition or network connection, cause they also talked about a poison pill, but in any case we won't be able to recover from that). The case that I do see more relevant is any sort of processing exception (business logic in the onNext, which we can't prevent from occurring) would also cause onError to be triggered and consequently unsubscribe. They want a means to be able to subscribe from the last synced point, rather than having to do a full sync.

Maithem: I think the intention of SubscriptionException was to signal corfu related exceptions, most of which won't allow the consumer to resume from the last successful point. For example, TrimmedException. Network exceptions and other related corfu exceptions shouldn't be passed to the clients since they can't really handle them (ex NetworkException).

Business logic should handle their own exceptions. I don't see the benefit of them not handling an exception, then we catch it and return it to them wrapped in another exception.

One possible use case is producers and consumers trying to coordinate via these subscriptions. Where a producer might write a command and the consumers would unsubscribe once they receive that command via onNext. I'm just trying to see if its possible to hide the sequence/epoch from clients.

test/src/test/java/org/corfudb/integration/StreamingIT.java
Maithem: `ts1` is just a sequencer call and it's not gaurenteed to be materialized `Timestamp ts1 = store.getTimestamp();` will call `runtime.getSequencerView().query().getToken();`

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: Yes, a TrimmedException is undoubtedly an unrecoverable error. 

As per business logic handling their own exceptions, I agree, but that is not the case now, we still need to make changes to our code to be resilient to that. We currently catch those exceptions and enter onError state (that is actually how I simulated a failure in the test). 

Ok, I see but why do we want to hide it if anyways we ask for it in the subscription API?

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: I thought you were referring to the second one which comes from the LogData, that's the one I'm saying can't regress as it was captured on the onNext, when reading an update which is already persisted in the datastore.

This one, yes I agree, in this case sequence regression is a possibility which would incur in data loss. But this was designed earlier not part of this change... And actually this is what clients are doing, so probably we'll need a solution to this as you're pointing out, but that can be part of another PR.

Maithem (approved):

test/src/test/java/org/corfudb/integration/StreamingIT.java
hisundar: Reattaching a stream onError() actually leads to a chain failure scenario as we had seen in one of the early clients. Hope the timestamp is not meant to be used purely for onError() based re-subscription since the right thing as @Maithem mentioned is for the client to open a snapshot before continuing.
Exposing the timestamp is fine, but perhaps we need to clarify the intention in the ask?

hisundar (approved):
Looks good but might be good to understand the reasoning behind this ask a little more just for us.

zhangn49 (approved):
Just curious, what is the solution to the onError with a TrimmedException today? Resubscribe without assigning a timestamp?

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -272,7 +273,11 @@ public void run() {
                                     ).collect(Collectors.toList())));
 
                     if (!entries.isEmpty()) {
-                        CorfuStreamEntries callbackResult = new CorfuStreamEntries(entries);
+                        Timestamp timestamp = Timestamp.newBuilder()
+                                .setSequence(logData.getGlobalAddress())
+                                .setEpoch(epoch)
```
pankti-m, line 278: is it required to expose the epoch too to the client?  Isnt globalAddress enough?

Lujie1996 (approved):
LGTM

PR #2908, opened 2021-03-11
Measure the time it takes to serialize the checkpoint entry log data …

PavelZaytsev:
Measure the checkpointing serialization, with percentages

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
@@ -1,5 +1,6 @@
 package org.corfudb.runtime.object;
 
+import io.micrometer.core.instrument.Timer;
 import org.corfudb.protocols.logprotocol.CheckpointEntry;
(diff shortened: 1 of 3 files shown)


xnull (approved):
looks good

runtime/src/main/java/org/corfudb/runtime/object/StreamViewSMRAdapter.java
```diff
@@ -81,8 +93,13 @@ public void gc(long trimMark) {
     public List<SMREntry> remainingUpTo(long maxGlobal) {
         return streamView.remainingUpTo(maxGlobal).stream()
                 .filter(m -> m.getType() == DataType.DATA)
-                .filter(m -> m.getPayload(runtime) instanceof ISMRConsumable
-                        || m.hasCheckpointMetadata())
+                .filter(m -> {
+                    Supplier<Object> getPayload = () -> m.getPayload(runtime);
+                    Object object = streamUpToDeserializationTimer.map(timer -> timer.record(getPayload))
```
zhangn49, line 98: Is it ok to use the same streamUpToDeserializationTimer in remainingUpTo?

PR #2909, opened 2021-03-17
UFO Streaming Bug Fix & Exception Handling

annym:
## Overview

Description:
- TrimmedExceptions were not being handled, therefore, if a client subscribed below a trim mark it would not fail incurring in data loss (bug). 
- Hierarchy of exceptions, any Corfu-specific exception during streaming is wrapped in UnrecoverableStreamingException, while client processing exceptions are thrown as received (so clients can decide to treat these differently).
- Tests to support all changes.

Why should this be merged: bug fix + API cleanup 

- [x] Public API has Javadoc


test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -812,4 +854,166 @@ public void testStreamingMultiTableSingleListener() throws Exception {
...
+        // Confirm we are unsubscribed (subscribing again does not throw an error) -> subscribing beyond trim point
+        Timestamp trimTimestamp = Timestamp.newBuilder()
+                .setEpoch(trimPoint.getEpoch())
+                .setSequence(trimPoint.getSequence())
+                .build();
+        store.subscribeListener(listener, namespace, "sample_streamer_1",
+                Collections.singletonList(defaultTableName), trimTimestamp);
+
+        // Verify number of updates received by listener is zero (empty) as the space was trimmed.
+        assertThat(listener.getUpdates()).isEmpty();
```
Lujie1996, line 899: Also verify that CountDownLatch `errorListener` is decremented to 0?

annym: Not really needed right, as in L. 888 we blocked until the latch count down reached zero.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
```diff
@@ -84,7 +92,13 @@ private void sendNotifications() throws Exception {
 
             // Send notification to client with the pre-registered callback.
             startTime = endTime;
-            subscription.getStreamingMetrics().recordDeliveryDuration(() -> listener.onNext(nextUpdate));
+            try {
+                subscription.getStreamingMetrics().recordDeliveryDuration(() -> listener.onNext(nextUpdate));
+            } catch (Throwable t) {
```
Maithem, line 97: The only Throwable expected from Corfu's POV is a TrimmedException. If Throwable is not an instance of TrimmedException then it's user exception.

runtime/src/main/java/org/corfudb/runtime/exceptions/UnrecoverableStreamingException.java
```diff
@@ -0,0 +1,15 @@
+package org.corfudb.runtime.exceptions;
+
+/**
+ * CorfuStore Streaming Exceptions are wrapped in this Exception type, so clients
+ * can distinguish from streaming exceptions caused by their own processing logic
+ * and those internal to Corfu, which will require a snapshot in order to recover.
+ *
+ * @author annym
+ */
+public class UnrecoverableStreamingException extends RuntimeException {
```
Maithem, line 10: Can you just use `StreamSubscriptionException`, just to keep the API's interface minimal.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -394,6 +394,15 @@ protected long getMaxGlobalFromMaxEntries(int maxEntries) {
...
 
+        // Validate the streaming start address (seeked address) is beyond the trim mark, otherwise, updates would be missed
+        // (as there is no further resolution for this space).
+        // A TrimmedException indicates a snapshot needs to be performed to sync the object from its checkpoint, followed
+        // by a re-subscription from a point beyond the trimmed address (i.e., the snapshot address)
+        if (getCurrentGlobalPosition() < streamAddressSpace.getTrimMark()) {
+            throw new TrimmedException(String.format("Stream seeked to %s, but address space is trimmed at %s", getCurrentGlobalPosition(),
+                    streamAddressSpace.getTrimMark()));
+        }
+
```
Maithem, line 405: Why isn't this exception discovered during `remainingUpTo` ?

Maithem: The stream being polled is opened with

```java

        StreamOptions options = StreamOptions
                .builder()
                .cacheEntries(false)
                .isCheckpointCapable(false)
                .build();

```

This should be sufficient to detect and throw a trimmed exception. If it's not, then there might be a bug.

runtime/src/main/java/org/corfudb/runtime/exceptions/UnrecoverableStreamingException.java
annym: I used that one in my first version, but then it was sort of confusing because this has nothing to do with an exception upon subscriptio. Probably we can rename that one to UnrecoverableStreamingException? so its more generic..

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Good catch, actually the bug is really returning Address.NON_ADDRESS when the address space size == 0. Because my test case happened to fall in this case it revealed the bug. If the log would've been partially trimmed we would've observed the TrimmedException. I did the correct fix and added a test case such that we cover both scenarios (partial trim and full trim)

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
annym: I was not 100% sure this is true. I agree in the theory it should, but I felt I might be missing on something. For instance, I just experimented with the test and if the runtime is shutdown after the subscription is done we'll get on the onError the following:

org.corfudb.runtime.exceptions.UnrecoverableStreamingException: java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.CompletableFuture$AsyncSupply@5326d2e3 rejected from java.util.concurrent.ThreadPoolExecutor@2c0da708[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 1]

Which is a Corfu exception, of course the issue here is major, as the runtime is disconnected but then it kind of breaks the assumption and like this I'm not sure if we could be missing something else.

Lujie1996 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
Maithem: Those are generic Java exceptions and I think its fine to pass them to the client. Not only can `RejectedExecutionException`, but also `InterruptedException` both of which are hard failures, wrapping them doesn't add any value, or facilitate for processing downstream. 

The point of `StreamSubscriptionException` is to signal to the client a `soft error` that is retriable, but this patch clubs two different classes of errors (soft/hard) in a single exception. 

Furthermore, those exceptions happen on daemon threads that we control, so if they happen they will not prevent other client logic from processing. There is a use case where clients would want to do post processing after shutting down the StreamingManager and those APIs are already available.

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -105,14 +105,14 @@ void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace
...
-            throw new StreamSubscriptionException(
+            throw new UnrecoverableStreamingException(
                     "StreamManager:subscriber already registered "
                             + streamListener + ". Maybe call unregister first?");
         }
         if (subscriptions.size() >= MAX_SUBSCRIBERS) {
             log.error("StreamManager::subscribe {} has too many {} subscribers",
                     streamListener, MAX_SUBSCRIBERS);
-            throw new StreamSubscriptionException(
+            throw new UnrecoverableStreamingException(
```
Maithem, line 115: `IllegalStateException` can be used here.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
Maithem: StreamSubscriptionException essentially should only wrap `TrimmedException`, that can be thrown directly too. But later on down the road there might be multiple soft errors that are retriable and StreamSubscriptionException can be an indirection to wrap those exceptions as well.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -396,7 +396,8 @@ protected long getMaxGlobalFromMaxEntries(int maxEntries) {
 
         long size = streamAddressSpace.size();
         if (size == 0L) {
-            return Address.NON_ADDRESS;
+            // The trim mark will allow to detect trimmed exceptions if the seeked address falls behind it
```
Maithem, line 399: Can you also add a unit test that targets  `IStreamView` since this is a separate bug altogether.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
annym: Just to be sure we're on the same page (as you wrote it in the opposite way) exceptions wrapped as 'StreamSubscriptionException' are actually <NOT> retriable. The message to consumers will be, if you receive a 'StreamSubscriptionException' do not retry, get a snapshot and then re-attach. Any other type of exception you can attempt to re-attach at your own risk.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Done, added 4 test cases in StreamViewTest.

BTW, the IStreamView APIs that are used to directly consume the stream (not the ones used through VLO) do not work for checkpointable streams. I remember Wenbin was going to work on this some time back but then he got deviated to something else. But this is something important that we should address any time soon. I'll add it as an issue.

runtime/src/main/java/org/corfudb/runtime/collections/StreamNotificationTask.java
annym: @Maithem addressed all comments, please re-review and let me know if you see it this way too (previous comment)... I've already changed it such that TrimmedException is the only Corfu expected exception in the NotificationTask, though I don't think we should ever hit this, cause TrimmedException could be hit only on the pollingTask but not on the NotificationTask (as by the time of notification the data has already been read and its enqueued)

annym: I actually removed it, the entry has already been read and is enqueued... there is no way we can hit a TrimmedException...

runtime/src/main/java/org/corfudb/runtime/collections/StreamManager.java
```diff
@@ -42,7 +42,7 @@
 public class StreamManager {
 
     private static final int DEFAULT_NUM_SUBSCRIBERS = 6; // Max number of threads.
-    private static final int MAX_SUBSCRIBERS = 32; // Arbitrary sanity check value.
+    public static final int MAX_SUBSCRIBERS = 32; // Arbitrary sanity check value.
```
Maithem, line 45: Why is this public?

annym: So we can access it in a test. There was no test for this edge case (when more than MAX_SUBSCRIBERS subscribe)..

runtime/src/main/java/org/corfudb/runtime/exceptions/UnrecoverableStreamingException.java
Maithem: `Unrecoverable` convey that there was a hard fault that cannot be recovered, which is not true. Perhaps, `StreamingException`.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -59,7 +60,7 @@ public void run() {
             log.error("Encountered exception {} during txn stream polling, listener: {}, " +
                     "namespace: {}", throwable, listener, subscription.getNamespace());
             streamingManager.unsubscribe(listener, false);
-            listener.onError(throwable);
+            listener.onError(new StreamingException(throwable));
```
Maithem, line 63: This will also propagate `RejectedExecutionException` and `InterruptedException` to the listener, but it shouldn't right? If those happen the client cannot react to those exceptions and we manage those threads. What is the benefit of signaling an exception on our threads to the consumer listeners?

annym: But then how do they know they're no longer subscribed, and deltas won't be coming in any more? the case of the RejectedExecutionException, deltas won't be coming in...

Maithem (approved):

PR #2910, opened 2021-03-18
Refactor + metrics

PavelZaytsev:
Make runtime registry static and add metrics to MOSMREntry de/serializations

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
@@ -1,10 +1,13 @@
 package org.corfudb.protocols.logprotocol;
 
 import com.google.common.annotations.VisibleForTesting;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.Timer;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.common.metrics.micrometer.MeterRegistryProvider;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.util.serializer.CorfuSerializer;
 import org.corfudb.util.serializer.Serializers;
@@ -13,6 +16,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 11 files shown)


runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -93,39 +103,115 @@ public void mergeInto(MultiObjectSMREntry other) {
      */
     @Override
     public void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
-        super.deserializeBuffer(b, rt);
+        Optional<Timer> deserializeTimer = MeterRegistryProvider.getInstance()
```
zhangn49, line 106: Looks like this part will only deserialize the ByteBuffer to a mapping of <StreamID, byte array>, do we also need the part that deserialize the byte array to the actual MultiSMREntry?

zhangn49 (commented):
Looks good, added one question

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
PavelZaytsev: Oh you mean - the time it takes to do the lazy deserialization? I can include that

zhangn49 (approved):

PavelZaytsev:
There is a lot of boilerplate code..
I will refactor all the occurrences of histograms and timers and create the proper functions for it in the MicroMeterUtils in some other PR.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -93,39 +103,115 @@ public void mergeInto(MultiObjectSMREntry other) {
...
+                int multiSMRLen = 0;
+                try {
+                    int start = b.readerIndex();
+                    MultiSMREntry.seekToEnd(b);
+                    multiSMRLen = b.readerIndex() - start;
+                    b.readerIndex(start);
+                    byte[] streamUpdates = new byte[multiSMRLen];
+                    b.readBytes(streamUpdates);
+                    streamBuffers.put(streamId, streamUpdates);
+                } finally {
```
Maithem, line 148: Why do you need a finally block? If there is a failure you don't want to add that data point. 

If it fails the variable will go out of scope and get GC'd.

runtime/src/main/java/org/corfudb/protocols/logprotocol/MultiObjectSMREntry.java
```diff
@@ -93,39 +103,115 @@ public void mergeInto(MultiObjectSMREntry other) {
...
+                            Serializers.CORFU.serialize(x.getValue(), b);
+                        } finally {
+                            serializeStreamSample.ifPresent(sample -> serializeStreamTimer.ifPresent(sample::stop));
+                            streamDistributionSummary.ifPresent(summary -> summary.record(b.writerIndex() - streamStart));
+                        }
+                    });
+        } finally {
+            serializeSample.ifPresent(sample -> serializeTimer.ifPresent(sample::stop));
+            distributionSummary.ifPresent(summary -> summary.record(b.writerIndex() - startIdx));
+            distributionSummaryEntries.ifPresent(summary -> summary.record(streamUpdates.size()));
```
Maithem, line 213: This will record the number of streams in a MultiOBjectSMR, you probably also want to record the number of updates within a stream, that is, ` x.getValue().getUpdates().size()`

Maithem (approved):
Looks good. Do you also want to add timers to `compress` and `decompress`  for `Codec` ?

xnull (approved):
looks good

PR #2914, opened 2021-03-19
Experiment: Parallelize SMREntry Deserialization

zhangn49:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
+++ b/runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
@@ -4,6 +4,8 @@ import static com.google.common.base.Preconditions.checkState;
 
 
 import io.netty.buffer.ByteBuf;
+
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -13,6 +15,8 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
 import lombok.ToString;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.util.serializer.CorfuSerializer;
 import org.corfudb.util.serializer.ISerializer;
@@ -120,7 +124,7 @@ public class SMREntry extends LogEntry implements ISMRConsumable {
         SMRMethod = new String(methodBytes);
         byte serializerId = b.readByte();
         byte numArguments = b.readByte();
-        Object[] arguments = new Object[numArguments];
+        List<Object> arguments = new ArrayList<>();
 
         if (!opaque) {
             serializerType = Serializers.getSerializer(serializerId);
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
```diff
@@ -120,27 +124,33 @@ void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
...
 
+        List<Pair<ByteBuf, Integer>> objBufPairs= new ArrayList<>();
         for (byte arg = 0; arg < numArguments; arg++) {
             int len = b.readInt();
             ByteBuf objBuf = b.slice(b.readerIndex(), len);
+            objBufPairs.add(new ImmutablePair<>(objBuf, len));
+            b.skipBytes(len);
+        }
+
+        objBufPairs.parallelStream().forEachOrdered(objBufPair -> {
```
Maithem, line 143: The issue with this, is that it will give you parallelisim within a single transaction, but not across transactions. For write heavy patterns, this solution won't improve the sync performance. I suggest you try this at a higher level, for example:

```java

    @Override
    public Stream<SMREntry> streamUpTo(long maxGlobal) {
        return streamView.streamUpTo(maxGlobal)
                .filter(m -> m.getPayload(runtime) instanceof ISMRConsumable
                        || m.hasCheckpointMetadata())
                .map(this::dataAndCheckpointMapper)
                .flatMap(List::stream);
    }

```

apavan: @Maithem this solution works for customers who are batching their writes in a single transaction and this experiment was also conducted to see if this approach can help address the current read performance problems we are facing in that context. I would recommend us taking a 2 phase approach here i.e. push this CL to address batch writes customer which is a performance blocker now and then later as you suggested we can take this up to a higher level. We are also running really short on time as we have less than week to bake this change.

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
```diff
@@ -120,27 +124,33 @@ void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
...
             serializerType = Serializers.getSerializer(serializerId);
         } else {
             this.serializerId = serializerId;
         }
 
+        List<Pair<ByteBuf, Integer>> objBufPairs= new ArrayList<>();
         for (byte arg = 0; arg < numArguments; arg++) {
             int len = b.readInt();
             ByteBuf objBuf = b.slice(b.readerIndex(), len);
+            objBufPairs.add(new ImmutablePair<>(objBuf, len));
```
zfrenette, line 139: Can you explain why `len` needs to be stored? Can it be inferred from `readableBytes()`?

zhangn49: I didn't look into the ByteBuf class and simply wrap them to a Pair. Thanks for pointing it out, just wrote a test to verify it and will remove it.

```
@Test
    public void testByteBuffer() {
        ByteBuf byteBuf = Unpooled.buffer();
        byte[] bytes = "testByteBuffer".getBytes();
        Serializers.CORFU.serialize(bytes, byteBuf);
        int len = 5;
        ByteBuf slice = byteBuf.slice(byteBuf.readerIndex(), len);
        assertThat(slice.readableBytes()).isEqualTo(len);
    }
```

annym (approved):

Lujie1996 (approved):

PavelZaytsev (approved):

runtime/src/main/java/org/corfudb/protocols/logprotocol/SMREntry.java
```diff
@@ -120,27 +122,34 @@ void deserializeBuffer(ByteBuf b, CorfuRuntime rt) {
...
-                arguments[arg] = argBytes;
+                arguments.add(argBytes);
             } else {
-                arguments[arg] = serializerType.deserialize(objBuf, rt);
+                arguments.add(serializerType.deserialize(objBuf, rt));
             }
-            b.skipBytes(len);
-        }
-        SMRArguments = arguments;
+        });
```
Maithem, line 150: I don't understand how this optimization works. I have a couple of questions. From the oracle docs:

> Note that you may lose the benefits of parallelism if you use operations like forEachOrdered with parallel streams

Since `forEachOrdered` is the last operation in the pipeline what is the point of using parallel stream, wouldn't `forEachOrdered` serialize the processing?

Most of the operations only have two arguments, so it's not clear this parallelization improves performance unless there are arguments that are collectios (i.e., putAll).

Also, `new ArrayList`, is not thread safe, see https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html

> Note that this implementation is not synchronized. If multiple threads access an ArrayList instance concurrently, and at least one of the threads modifies the list structurally, it must be synchronized externally.

xnull: It looks like we lose parallelism if we use forEachOrdered https://stackoverflow.com/questions/47336825/benefit-of-using-foreachordered-with-parallel-streams

zhangn49:
SMREntry level parallelism does not work, a higher level parallelism is here #2917

PR #2917, opened 2021-03-23
Parallelize the deserialization of LogData at stream view layer

zhangn49:
## Overview

Description:

Some intermediate actions of a stream like 'filter' and 'map' can benefit from parallelism.

In this case, utilize parallelism to speed up ILogData deserialization. To avoid materializing all items and loading them to memory, this PR introduced a new customized stream spliterator, (i.e., FixedBatchSpliterator), which can split the stream into a few small chunks. 

Why should this be merged: 

performance enhancement

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/stream/FixedBatchSpliterator.java
```diff
@@ -0,0 +1,70 @@
...
+    public long estimateSize() {
+        return est;
+    }
+
+    @Override
+    public int characteristics() {
+        return characteristics;
+    }
+
+    private static final class HoldingConsumer<T> implements Consumer<T> {
```
xnull, line 62: What is the point of having HoldingConsumer? 
1. why not just use the object?
2. HoldingConsumer brings complexity, developers in the future will have a hard time figuring out what type of an object it holds. Please don't give up on typing, otherwise, we all will struggle with the code

zhangn49: > What is the point of having HoldingConsumer?
> 
> 1. why not just use the object?
> 2. HoldingConsumer brings complexity, developers in the future will have a hard time figuring out what type of an object it holds. Please don't give up on typing, otherwise, we all will struggle with the code

It is used to get the object from `tryAdvance`, which does not return the object. But we need those objects to split batch.
And it is borrowed from 'java.util.Spliterators'

xnull: Why not use Consumer<T> then?
Or at least please use `T value` instead of `Object value`

PR #2918, opened 2021-03-25
Metadata Revision Field fix

annym:
## Overview

Description:

Revision field should be a non-primitive type such that we can differentiate
an unset field from a field that is explicitly set to the default value (0L).

**Bug**: if we can't differentiate when the revision field was unset (where its value defaults to 0) from the case when the revision field is explicitly set to 0 (case of primitives), we could erroneously bump a revision even though it's stale. Consider the following example, revision is currently at 5, someone setting it explicitly to 0, would currently not get a StaleRevisionException but instead bump the version to 6 and commit successfully, though a stale revision was specified. This PR fixes this issue.

Why should this be merged: current bug in the revision field.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/proto/example_schemas.proto
+++ b/runtime/proto/example_schemas.proto
@@ -4,10 +4,10 @@ package org.corfudb.runtime;
 option java_package = "org.corfudb.runtime";
 
 import "corfu_options.proto";
-import "google/protobuf/descriptor.proto";
+import "google/protobuf/wrappers.proto";
 
 message ManagedMetadata {
-    int64 revision = 1;
+    .google.protobuf.Int64Value revision = 1;
     int64 create_time = 2;
     string create_user = 3;
     int64 last_modified_time = 4;
--- a/runtime/src/main/java/org/corfudb/runtime/collections/ManagedTxnContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/ManagedTxnContext.java
@@ -1,6 +1,7 @@
 package org.corfudb.runtime.collections;
 
 import com.google.protobuf.Descriptors;
+import com.google.protobuf.Int64Value;
 import com.google.protobuf.Message;
 import lombok.Getter;
 import org.corfudb.runtime.exceptions.StaleRevisionUpdateException;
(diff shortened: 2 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/ManagedTxnContext.java
```diff
@@ -675,16 +683,18 @@ void deleteRecord(@Nonnull String tableName,
...
-                        Long givenRevision = (Long) deltaMetadata.getField(fieldDescriptor);
-                        if (givenRevision == 0 || // Do not validate revision if field isn't set
-                                (givenRevision > 0 && // Validate revision only if set
+                        Long prevRevision = ((Int64Value) oldRecord.getMetadata().getField(fieldDescriptor)).getValue();
+                        Long givenRevision = null;
+                        if (deltaMetadata.hasField(fieldDescriptor)) {
+                            givenRevision = ((Int64Value) deltaMetadata.getField(fieldDescriptor)).getValue();
+                        }
+                        if (givenRevision == null || // Do not validate revision if field isn't set
+                                (givenRevision >= 0 && // Validate revision only if set
```
zhangn49, line 697: is ">=0" still needed here?

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/ManagedTxnContext.java
annym: I was thinking about ignoring negative numbers, but you're actually right its not needed. Removed!

runtime/proto/example_schemas.proto
```diff
@@ -4,10 +4,10 @@ package org.corfudb.runtime;
 option java_package = "org.corfudb.runtime";
 
 import "corfu_options.proto";
-import "google/protobuf/descriptor.proto";
+import "google/protobuf/wrappers.proto";
 
 message ManagedMetadata {
-    int64 revision = 1;
+    .google.protobuf.Int64Value revision = 1;
```
xnull, line 10: is it correct?

hisundar (commented):
This change by itself is ok but won't affect anything as ManagedMetadata isn't consumed by verticals. I am concerned about the amount of changes required to migrate all the existing metadata protobufs already deployed from a primitive type to a non-primitive type just to support a 0.
Also this behavior was requested by some early CorfuStore consumers. Just wondering if we have checked with them on the new changes

annym:
> This change by itself is ok but won't affect anything as ManagedMetadata isn't consumed by verticals. I am concerned about the amount of changes required to migrate all the existing metadata protobufs already deployed from a primitive type to a non-primitive type just to support a 0.
> Also this behavior was requested by some early CorfuStore consumers. Just wondering if we have checked with them on the new changes

Yes, this is an initial effort to start from Corfu side, but we need to do these exact changes on the shim layer. We did have a meeting with the CorfuStore consumers and we agreed we could not change the initial version from 0 to 1 but this was an option. I plan to do these same changes on that side and directly change all the usages of the primitives to non-primitives.

hisundar (approved):
Let's coordinate on the migration tasks once this PR lands.

PR #2919, opened 2021-03-25
Fix skipping of Universe Tests in GitHub Actions

chetangudisagar:
## Overview

Description: The Universe Tests were getting skipped earlier unintentionally even after mentioning the maven build profile. Adding execution id and execution goals ensured that they are not skipped.

Why should this be merged: This resumes the universal tests in the GitHub Actions.

- [x] Public API has Javadoc

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -93,4 +93,4 @@ jobs:
         run: ./mvnw clean install -DskipTests -Pdocker -Dmaven.javadoc.skip=true -T 1C
 
       - name: Run universe tests
-        run: ./mvnw -pl :universe verify -Puniverse -DskipTests -Dmaven.javadoc.skip=true
+        run: ./mvnw -pl :universe verify -Puniverse -Dmaven.javadoc.skip=true
--- a/it/pom.xml
+++ b/it/pom.xml
@@ -100,6 +100,15 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-failsafe-plugin</artifactId>
                         <version>3.0.0-M5</version>
+                        <executions>
+                            <execution>
+                                <id>integration-test</id>
+                                <goals>
+                                    <goal>integration-test</goal>
+                                    <goal>verify</goal>
+                                </goals>
+                            </execution>
+                        </executions>
                         <configuration>
                             <skipITs>false</skipITs>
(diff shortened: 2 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -93,4 +93,4 @@ jobs:
         run: ./mvnw clean install -DskipTests -Pdocker -Dmaven.javadoc.skip=true -T 1C
 
       - name: Run universe tests
-        run: ./mvnw -pl :universe verify -Puniverse -DskipTests -Dmaven.javadoc.skip=true
+        run: ./mvnw -pl :universe verify -Puniverse -Dmaven.javadoc.skip=true
```
xnull, line 96: If you enable unit tests, it will run unit tests along with integration tests, which is not desirable

xnull (changes requested):

.github/workflows/pull_request.yml
xnull: sorry, looks like you are right

it/pom.xml
```diff
@@ -100,6 +100,15 @@
...
                         <version>3.0.0-M5</version>
+                        <executions>
+                            <execution>
+                                <id>integration-test</id>
+                                <goals>
+                                    <goal>integration-test</goal>
+                                    <goal>verify</goal>
+                                </goals>
+                            </execution>
+                        </executions>
```
xcchang, line 111: Could you explain why is this needed? for me it looks like the change in `pull_request.yml` will bring back the universe test?

xnull (approved):
-DskipTests should affect only the surefire plugin, for some reason it also offs failsafe plugin.
Surefire and failsafe plugins are quite messed in maven. 
It would be great to migrate to Gradle, that will improve everything in our build.

good catch, thank you

it/pom.xml
xnull: This adds integration-test task into the verification task, it's a necessary change
P.S. as I remember universe tests worked even without this change though, but it's good to have it here

zhangn49 (approved):

it/pom.xml
chetangudisagar: Without this "integration-test" goal, the integration-tests were getting skipped with the "universe" profile id. It used to work before with the "it" profile id. Hence I had to add this goal specifically.

PavelZaytsev (approved):

PR #2921, opened 2021-03-30
Log Replication: Bug fix + More Lock logging

zhangn49:
## Overview

Description:

Why should this be merged: 

Fixed a bug that log replication can not stop when the distributed lock is released.
Added an IT to test lock release scenario.
Added more logging around lock state transition.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -135,12 +135,13 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private NodeDescriptor localNodeDescriptor;
 
     /**
-     * Unique node identifier
+     * Unique node identifier of lock
      */
     // Note: not to be confused with NodeDescriptor's NodeId, which is a unique
     // identifier for the node as reported by the Cluster/Topology Manager
     // This node Id is internal to Corfu Log Replication and used for the lock acquisition
-    private final UUID logReplicationNodeId;
+    @Getter
+    private final UUID logReplicationLockId;
 
     /**
      * A queue of Discovery Service events
@@ -205,7 +206,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
                                             @Nonnull CorfuReplicationClusterManagerAdapter clusterManagerAdapter,
                                             @Nonnull CompletableFuture<CorfuInterClusterReplicationServerNode> serverCallback) {
         this.clusterManagerAdapter = clusterManagerAdapter;
-        this.logReplicationNodeId = serverContext.getNodeId();
+        this.logReplicationLockId = serverContext.getNodeId();
         this.serverContext = serverContext;
         this.localEndpoint = serverContext.getLocalEndpoint();
         this.serverCallback = serverCallback;
@@ -445,7 +446,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
 
             IRetry.build(IntervalRetry.class, () -> {
                 try {
-                    lockClient = new LockClient(logReplicationNodeId, getCorfuRuntime());
+                    lockClient = new LockClient(logReplicationLockId, getCorfuRuntime());
                     // Callback on lock acquisition or revoke
                     LockListener logReplicationLockListener = new LogReplicationLockListener(this);
                     // Register Interest on the shared Log Replication Lock
@@ -455,8 +456,8 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
                     throw new RetryNeededException();
                 }
 
-                log.debug("Registered to lock, client msb={}, lsb={}", logReplicationNodeId.getMostSignificantBits(),
-                        logReplicationNodeId.getLeastSignificantBits());
+                log.debug("Registered to lock, client msb={}, lsb={}", logReplicationLockId.getMostSignificantBits(),
+                        logReplicationLockId.getLeastSignificantBits());
                 return null;
             }).run();
         } catch (InterruptedException e) {
@@ -564,8 +565,9 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
      */
     public void processLockRelease() {
         log.debug("Lock released");
-        isLeader.set(false);
+        // Unset isLeader flag after stopping log replication
         stopLogReplication();
+        isLeader.set(false);
         // Signal Log Replication Server/Sink to stop receiving messages, leadership loss
         interClusterReplicationService.getLogReplicationServer().setLeadership(false);
         recordLockRelease();
--- a/runtime/src/main/java/org/corfudb/runtime/view/replication/ReadWaitHoleFillPolicy.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/replication/ReadWaitHoleFillPolicy.java
@@ -74,8 +74,8 @@ public class ReadWaitHoleFillPolicy implements IHoleFillPolicy {
                 }
 
                 // Otherwise try again.
-                log.debug("peekUntilHoleFillRequired: Attempted read at address {}, "
-                        + "but data absent. Retrying.", address);
+                log.trace("peekUntilHoleFillRequired: Attempted read at address {}, "
+                            + "but data absent. Retrying.", address);
                 throw new RetryNeededException();
             }).setOptions(x -> x.setMaxRetryThreshold(retryWaitThreshold)).run();
         } catch (InterruptedException ie) {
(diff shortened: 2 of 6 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -564,8 +564,9 @@ private void updateTopologyConfigId(long configId) {
      */
     public void processLockRelease() {
         log.debug("Lock released");
-        isLeader.set(false);
+        // Unset isLeader flag after stopping log replication
```
annym, line 567: Nan can you explain this a bit more, how changing the order fixes a bug, can we write a test?

zhangn49: > Nan can you explain this a bit more, how changing the order fixes a bug, can we write a test?

function `stopLogReplication()` has an if condition about isLeader flag. 


```
    /**
     * Stop Log Replication
     */
        if (localClusterDescriptor != null && localClusterDescriptor.getRole() == ClusterRole.ACTIVE && isLeader.get()) {
            log.info("Stopping log replication.");
            replicationManager.stop();
        }
    }
```


However, `isLeader` flag was unset before calling it, which means it will never be true in this lock release case.

runtime/src/main/java/org/corfudb/runtime/view/replication/ReadWaitHoleFillPolicy.java
```diff
@@ -74,8 +74,10 @@ public ILogData peekUntilHoleFillRequired(long address, Function<Long, ILogData>
                 }
 
                 // Otherwise try again.
-                log.debug("peekUntilHoleFillRequired: Attempted read at address {}, "
-                        + "but data absent. Retrying.", address);
+                if (log.isTraceEnabled()) {
```
xnull, line 77: please don't use `if` condition http://www.slf4j.org/faq.html#logging_performance

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -82,31 +82,31 @@
     //TODO need to determine if the application should provide a clientId or should it be internally generated.
     public LockClient(UUID clientId, CorfuRuntime corfuRuntime) throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
 
-        this.taskScheduler = Executors.newScheduledThreadPool(1, (r) ->
+        this.taskScheduler = Executors.newScheduledThreadPool(1, r ->
```
xnull, line 85: please rename `r` to something meaningful

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
```diff
@@ -82,31 +82,31 @@
...
         {
             Thread t = Executors.defaultThreadFactory().newThread(r);
             t.setName("LockTaskThread");
             t.setDaemon(true);
             return t;
         });
 
-        this.leaseMonitorScheduler = Executors.newScheduledThreadPool(1, (r) ->
+        this.leaseMonitorScheduler = Executors.newScheduledThreadPool(1, r ->
```
xnull, line 93: what does `1` mean? please provide a constant with a decent name

utils/src/main/java/org/corfudb/utils/lock/persistence/LockStore.java
```diff
@@ -51,7 +51,7 @@
     /**
      * Cache of all the observed locks/leases. Contains the last timestamp at which the lock was last observed.
      */
-    private Map<LockId, ObservedLock> observedLocks = new ConcurrentHashMap<>();
+    private final Map<LockId, ObservedLock> observedLocks = new ConcurrentHashMap<>();
```
xnull, line 54: would be great to use `ConcurrentMap` which tells the user that the map is concurrent

utils/src/main/java/org/corfudb/utils/lock/states/HasLeaseState.java
```diff
@@ -103,6 +103,7 @@ public HasLeaseState(Lock lock) {
      */
     @Override
     public void onEntry(LockState from) {
+        log.debug("Lock Transition from {} to HasLeaseState. Start lease renewal and monitor!", from.getType());
```
xnull, line 106: are you sure that we won't pollute logs with this message?

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
zhangn49: done

runtime/src/main/java/org/corfudb/runtime/view/replication/ReadWaitHoleFillPolicy.java
zhangn49: done, good to know! thanks

utils/src/main/java/org/corfudb/utils/lock/LockClient.java
zhangn49: done

utils/src/main/java/org/corfudb/utils/lock/persistence/LockStore.java
zhangn49: agree, done

utils/src/main/java/org/corfudb/utils/lock/states/HasLeaseState.java
zhangn49: Yes, the transition won't happen frequently.

zhangn49:
@annym Added a new IT to test active lock release

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -891,6 +906,92 @@ public void testEnforceSnapshotSync() throws Exception {
...
+
+        // Release Active's lock
+        TxnContext txnContext = activeCorfuStore.txn(CORFU_SYSTEM_NAMESPACE);
+        txnContext.clear(activeLockTable);
+        txnContext.commit();
+        log.info("Active's lock is released!");
+        TimeUnit.SECONDS.sleep(lockInterval);
+
+        // Release Active's lock again
+        txnContext = activeCorfuStore.txn(CORFU_SYSTEM_NAMESPACE);
```
annym, line 978: Curious, why do you need to release it again? is it because eventually it will take it again? If that's the case, you could also change the topology like we had talked, which would also cause log replication to stop (even though the lock is still held)

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
```diff
@@ -891,6 +906,92 @@ public void testEnforceSnapshotSync() throws Exception {
...
+        for (int i = secondBatch; i < thirdBatch; i++) {
+            activeRuntime.getObjectsView().TXBegin();
+            mapActive.put(String.valueOf(i), i);
+            activeRuntime.getObjectsView().TXEnd();
+        }
+        assertThat(mapActive.size()).isEqualTo(thirdBatch);
+        log.info("Active map has {} entries now!", thirdBatch);
+
+        // Standby map should still have secondBatch size
+        log.info("Standby map should still have {} size", secondBatch);
```
annym, line 991: should you add some time to emulate transfer time and confirm it really doesn't occur? otherwise, we could be querying before it was even initiated.

annym (approved):
Check the comment in the test to see if this should be adjusted or not.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: Yes, it will acquire the lock again. The fix is only for lock release case, so I think change topology does not help too much here.

test/src/test/java/org/corfudb/integration/CorfuReplicationClusterConfigIT.java
zhangn49: If the lock release does not stop the replication, it will remain the log entry sync type and should replicate pretty fast.
I removed sleep time and eventual verification because it will take the lock in the next monitor cycle...

utils/src/main/java/org/corfudb/utils/lock/states/HasLeaseState.java
```diff
@@ -124,6 +125,7 @@ public void onEntry(LockState from) {
      */
     @Override
     public void onExit(LockState to) {
+        log.debug("Lock Transition to {} from HasLeaseState. Stop lease renewal and monitor!", to.getType());
```
Lujie1996, line 128: We can make it uniform by saying from A_state to B_state.

Lujie1996 (approved):

PR #2924, opened 2021-04-03
Redefine Streaming API

annym:
## Overview

Description:

Shim layer should not expose timestamp, instead a snapshot timestamp should come from the commit of read-only transactions.

Why should this be merged: potential data loss bug (user believes he's subscribing to streaming updates from version 'X', but 'X' regresses to 'R',  and updates between 'R' and 'X' are not sent/reported to the subscriber).  Also, timestamp between snapshot and delta sync point with current API are not easily correlated and subject to dev usage error.

- [ ] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -265,6 +265,25 @@ void subscribe(@Nonnull StreamListener streamListener, @Nonnull String namespace
...
+     * Note: if memory is a consideration consider use the other version of subscribe that is
+     * able to specify the size of buffered transactions entries.
+     *
+     * @param streamListener   callback context
+     * @param namespace        the CorfuStore namespace to subscribe to
+     * @param streamTag        only updates of tables with the stream tag will be polled
+     * @param tablesOfInterest only updates from these tables of interest will be sent to listener
+     */
+    public void subscribeListener(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                                  @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest) {
```
annym, line 283: Add tests for this new API (pending)

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
```diff
@@ -258,7 +259,11 @@ public void testStandbyTransactionLogging() throws Exception {
     private void subscribe(String mapName) {
         // Subscribe to mapName and upon changes stop standby LR
         CorfuStore corfuStore = new CorfuStore(standbyRuntime);
-        CorfuStoreMetadata.Timestamp ts = corfuStore.getTimestamp();
+        Token token = standbyRuntime.getSequencerView().query().getToken();
+        CorfuStoreMetadata.Timestamp ts = CorfuStoreMetadata.Timestamp.newBuilder()
+                .setEpoch(token.getEpoch())
+                .setSequence(token.getSequence())
+                .build();
```
zhangn49, line 266: is queryToken necessary here? NON_ADDRESS seems to be ok.

test/src/test/java/org/corfudb/integration/SourceForwardingDataSender.java
```diff
@@ -104,6 +104,7 @@ public SourceForwardingDataSender(String destinationEndpoint, LogReplicationConf
         final CompletableFuture<LogReplicationEntryMsg> cf = new CompletableFuture<>();
 
         // Emulate Channel by directly accepting from the destination, whatever is sent by the source manager
+        System.out.println(">>> Receive message :: entryType :: " + message.getMetadata().getEntryType() + " ts::" + message.getMetadata().getTimestamp());
```
zhangn49, line 107: s.out -> log

utils/src/main/java/org/corfudb/utils/lock/persistence/LockStore.java
```diff
@@ -90,7 +93,11 @@ public LockStore(CorfuRuntime runtime, UUID clientUuid) throws NoSuchMethodExcep
      * @throws LockStoreException
      */
     public boolean acquire(LockId lockId) throws LockStoreException {
-        CorfuStoreMetadata.Timestamp timestamp = corfuStore.getTimestamp();
+        Token token = runtime.getSequencerView().query().getToken();
+        CorfuStoreMetadata.Timestamp timestamp = CorfuStoreMetadata.Timestamp.newBuilder()
+                .setEpoch(token.getEpoch())
+                .setSequence(token.getSequence())
+                .build();
```
zhangn49, line 100: LockStore still use the deprecated query API, so TS is still here. Do we plan to remove it?

Just wanted to make sure my understand is correct - Every corfu store user should get timestamp from read txn commit, rather than querying token to build this TS. And looks like user should not have its runtime?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -203,41 +211,43 @@ private void processOpaqueEntry(List<SMREntry> smrEntries, UUID shadowStreamUuid
...
+
+            try {
+                if (!snapshotSyncStartMarker.isPresent()) {
+                    logReplicationMetadataManager.setSnapshotSyncStartMarker(txnContext, snapshotSyncId, timestamp);
+                    snapshotSyncStartMarker = Optional.of(new SnapshotSyncStartMarker(snapshotSyncId, timestamp.getSequence()));
+                }
+                txnContext.commit();
+            } catch (Exception e) {
+                log.warn("Caught an exception ", e);
+                throw e;
```
zhangn49, line 225: same here

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
```diff
@@ -185,12 +186,11 @@ public LogReplicationEntryMsg processAcks() throws InterruptedException, Executi
...
     }
 
-    public CompletableFuture<LogReplicationEntryMsg> sendWithBuffering(LogReplicationEntryMsg message,
-                                                                    String metricName,
-                                                                    Tag replicationTag) {
+    public CompletableFuture<LogReplicationEntryMsg> sendWithBuffering(LogReplicationEntryMsg message, String metricName, Tag replicationTag) {
         LogReplicationEntryMetadataMsg metadata = overrideSyncSeqNum(
                 message.getMetadata(), snapshotSyncSequenceNumber++);
         LogReplicationEntryMsg newMessage = overrideMetadata(message, metadata);
+        pendingMessages.append(newMessage);
```
zhangn49, line 193: This missing line could have caused many bugs..

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -99,71 +104,111 @@ public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, Str
     }
 
     public CorfuStoreMetadata.Timestamp getTimestamp() {
-        return corfuStore.getTimestamp();
+        Token token = runtime.getSequencerView().query().getToken();
+        return CorfuStoreMetadata.Timestamp.newBuilder()
+                .setEpoch(token.getEpoch())
+                .setSequence(token.getSequence())
+                .build();
```
zhangn49, line 111: Is this function still needed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -99,71 +104,111 @@ public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, Str
...
+
+            metadataMap.put(keyType, stringValue != null ? Long.parseLong(stringValue) : -1L);
         }
 
-        return val;
+        return metadataMap;
     }
 
-    public long query(CorfuStoreMetadata.Timestamp timestamp, LogReplicationMetadataType key) {
+    public long queryMetadata(CorfuStoreMetadata.Timestamp ts, LogReplicationMetadataType key) {
```
zhangn49, line 173: looks like all caller use null? should we remove it? and modify queryString?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/SenderBufferManager.java
annym: right, this took me hours to realize! And luckily because I had disabled the metrics to ease logs readability I was able to hit this. This was an error doing copy/paste with the RPC change but it would've been critical for LR if metrics were disabled.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
annym: So this is only needed for the snapshot sync start marker timestamp. I think we might be able to get this timestamp from somewhere else (for instance from the tx.commit of the first shadow stream commit, but got lost in the code with many other changes).. let me know your thoughts..

zhangn49: Yes, from tx.commit makes more sense. As we don't want users to build their TS.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -182,14 +189,15 @@ public void reset(long topologyConfigId, long snapshot) {
...
-            throw e;
+        try (TxnContext txnContext = logReplicationMetadataManager.getTxnContext()) {
+            processOpaqueEntry(txnContext, smrEntries, shadowStreamUuid);
+
+            try {
+                txnContext.commit();
+            } catch (Exception e) {
+                log.warn("Caught an exception ", e);
+                throw e;
+            }
```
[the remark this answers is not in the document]
hisundar: +1

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/utils/LogReplicationStreamNameTableManager.java
```diff
@@ -148,64 +155,58 @@ private void deleteExistingStreamNameAndVersionTables() {
...
-        Set<LogReplicationStreams.TableInfo> tables =
-            q.keySet(LOG_REPLICATION_STREAMS_NAME_TABLE, null);
         Set<String> tableNames = new HashSet<>();
-        tables.forEach(table -> {
-            tableNames.add(table.getName());
-        });
+        try (TxnContext txn = corfuStore.txn(CORFU_SYSTEM_NAMESPACE)) {
+
+            Set<TableInfo> tables = txn.keySet(LOG_REPLICATION_STREAMS_NAME_TABLE);
+            tables.forEach(table -> tableNames.add(table.getName()));
```
hisundar, line 208: perhaps add a txn.commit() here too?

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -149,6 +150,38 @@ private void txBeginInternal(boolean allowNestedTransactions) {
...
+     * @param <V>         - type of the payload/value
+     * @param <M>         - type of the metadata
+     */
+    private <K extends Message, V extends Message, M extends Message>
+    void validateWrite(@Nonnull Table<K, V, M> table, K key, V value, M metadata, boolean validateKey) {
+        baseValidateWrite(table, key, validateKey);
+        if (table.getMetadataClass() == null && metadata != null) {
+            throw new IllegalArgumentException("Metadata schema for table " + table.getFullyQualifiedTableName() + " is defined as NULL, non-null metadata is not allowed.");
+        }
+        if (table.getValueClass() == null && value != null) {
```
hisundar, line 175: I don't think we support null payloads, perhaps this check can be moved into openTable itself?

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
hisundar: Also given that we have at least one use case where we need to subscribe from the latest, should we overload the subscribeListener() api to not have a timestamp, where we will under the hood subscribe to the latest timestamp?

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -130,6 +148,156 @@ public void checkDirtyReads() throws Exception {
...
+     *
+     * @throws Exception
+     */
+    @Test
+    public void checkOpenRetriesTXN() throws Exception {
+        CorfuRuntime corfuRuntime = getDefaultRuntime();
+        CorfuStoreShim corfuStore = new CorfuStoreShim(corfuRuntime);
+        final String nsxManager = "nsx-manager"; // namespace for the table
+        final String tableName = "EventInfo"; // table name
+        final int numThreads = 10;
```
hisundar, line 288: not related to your change but this seems to be a source of flakiness in tests, because in GitHub actions the number of cores may be far less, executing 10 threads can sometimes timeout, perhaps we can reduce this to 5?

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
```diff
@@ -464,45 +466,53 @@ void dataStoreIntegration(
...
+            Set<EventInfo> scannedValues = queryResult.stream()
+                    .map(CorfuStoreEntry::getPayload).collect(Collectors.toSet());
+
+            assertThat(filteredEvents.size()).isGreaterThan(0).isLessThan(SAMPLE_SIZE);
+            assertThat(scannedValues.size()).isEqualTo(filteredEvents.size());
+            assertThat(tx.count(tableName)).isEqualTo(SAMPLE_SIZE + 1);
+        }
+
+        try (TxnContext tx = corfuStore.txn(namespace, IsolationLevel.snapshot(timestamp))) {
+            assertThat(tx.count(tableName)).isEqualTo(1);
```
hisundar, line 514: can we add a tx.commit() here too please?

test/src/test/java/org/corfudb/runtime/collections/DiskBackedCorfuClientTest.java
```diff
@@ -464,45 +466,53 @@ void dataStoreIntegration(
...
+                    .collect(Collectors.toSet());
+            List<CorfuStoreEntry<Uuid, EventInfo, SampleSchema.ManagedResources>> queryResult =
+                    tx.executeQuery(tableName,
+                            record -> ((EventInfo) record.getPayload()).getEventTime() > medianEventTime);
+            Set<EventInfo> scannedValues = queryResult.stream()
+                    .map(CorfuStoreEntry::getPayload).collect(Collectors.toSet());
+
+            assertThat(filteredEvents.size()).isGreaterThan(0).isLessThan(SAMPLE_SIZE);
+            assertThat(scannedValues.size()).isEqualTo(filteredEvents.size());
+            assertThat(tx.count(tableName)).isEqualTo(SAMPLE_SIZE + 1);
```
hisundar, line 510: can we add a tx.commit() here please?

utils/src/main/java/org/corfudb/utils/lock/persistence/LockStore.java
```diff
@@ -253,9 +263,9 @@ private void update(LockId lockId, LockData lockMetaData, CorfuStoreMetadata.Tim
      */
     @VisibleForTesting
     public Optional<LockData> get(LockId lockId) throws LockStoreException {
-        try {
-            CorfuRecord record = corfuStore.query(NAMESPACE).getRecord(TABLE_NAME, lockId);
-            if (record != null) {
+        try (TxnContext txn = corfuStore.txn(NAMESPACE)) {
+            CorfuStoreEntry record = txn.getRecord(TABLE_NAME, lockId);
+            if (record.getPayload() != null) {
```
hisundar, line 268: txn.commit() please

hisundar (changes requested):
Thanks so much Anny for the huge cleanup!

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
annym: It should cause currently nothing prevents value or key being null. The annotation in place is only informational.

annym: I've added the check on openTable

hisundar (approved):
LGTM, thanks

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
annym: I already added that API and actually added one more where the table names are also not needed. Changed this code to subscribe to latest timestamp instead of using queryToken.

utils/src/main/java/org/corfudb/utils/lock/persistence/LockStore.java
annym: correct, actually I'm not sure why this TS was needed to be passed on the update and create if it does not need a snapshot. I completely removed the dependency from timestamp, please take a look in case I missed something.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
annym: Ok, made this change.. this is the only one I'm a bit concerned we don't see the implications of this change. If you can give it some extra thoughts, I'm doing the same.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -203,41 +206,42 @@ private void processOpaqueEntry(List<SMREntry> smrEntries, UUID shadowStreamUuid
...
-         */
-    private CorfuStoreMetadata.Timestamp processOpaqueEntry(TxBuilder txBuilder, List<SMREntry> smrEntries, UUID shadowStreamUuid) {
+    /**
+     * Write a list of SMR entries to the specified stream log.
+     *
+     * @param smrEntries
+     * @param shadowStreamUuid
+     */
+    private CorfuStoreMetadata.Timestamp processOpaqueEntry(TxnContext txnContext, List<SMREntry> smrEntries, UUID shadowStreamUuid) {
         CorfuStoreMetadata.Timestamp timestamp = logReplicationMetadataManager.getTimestamp();
```
zhangn49, line 238: The return value seems to be unnecessary right now, we can remove it.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -99,71 +101,103 @@ public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, Str
     }
 
     public CorfuStoreMetadata.Timestamp getTimestamp() {
-        return corfuStore.getTimestamp();
+        Token token = runtime.getSequencerView().query().getToken();
+        return CorfuStoreMetadata.Timestamp.newBuilder()
+                .setEpoch(token.getEpoch())
+                .setSequence(token.getSequence())
+                .build();
```
zhangn49, line 108: The only caller is `processOpaqueEntry`, so we can remove this function completely.

zhangn49 (approved):
LGTM, 2 comments about minor cleanup

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
annym: yes, good catch. Forgot to revert it.

PR #2926, opened 2021-04-06
Instrument Streaming API

annym:
## Overview

Description:

- Add metric to hold time spent by entries in streaming queue
- Add trace logging to ease performance/functional debugging
- Move streaming to runtime parameters to ease experimentation (tuning)

Why should this be merged:  help with performance issues in the client. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -234,37 +234,80 @@ public class CorfuRuntime {
          */
         private Codec.Type codecType = Codec.Type.ZSTD;
 
-        /**
+        /*
          * Enable runtime metrics.
          */
         private boolean metricsEnabled = true;
 
+        /*
+         * Total time in milliseconds for polling task to block until buffer space is available.
+         */
+        private long streamingPollingBlockingTimeMs = 5;
+
+        /*
+         * Period of time in ms to sleep before next cycle, when poller gets no new data changes.
+         */
+        private int streamingPollingIdleWaitTimeMs = 5;
+
+        /*
+         * Capacity of queue shared by by streaming polling and notification tasks.
+         */
+        private int streamingQueueSize = 100;
+
+        /*
+         * Total number of threads in Polling Executor Pool (shared across all listeners)
+         */
+        private int streamingPollingThreadPoolSize = 2;
+
+        /*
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscriptionMetrics.java
```diff
@@ -39,6 +41,12 @@
                         .publishPercentiles(percentiles)
                         .publishPercentileHistogram(true)
                         .register(registry));
+        this.queueDurationTimer = MeterRegistryProvider.getInstance()
```
hisundar, line 44: Wondering if the MetricRegistryProvider can be obtained from the corfu runtime so all ufo metrics go to the same place?
but this is ok too.

hisundar (approved):
Looks good to me. Comment is minor

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscriptionMetrics.java
annym: Pavel made this change recently, he can comment on this @PavelZaytsev

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -234,37 +234,74 @@ public static CorfuRuntimeParametersBuilder builder() {
...
+            private int invalidateRetry = 5;
             private PriorityLevel priorityLevel = PriorityLevel.NORMAL;
             private Codec.Type codecType = Codec.Type.ZSTD;
             private boolean metricsEnabled = true;
+            private long streamingPollingBlockingTimeMs = 200;
+            private int streamingQueueSize = 60;
+            private int streamingPollingThreadPoolSize = 4;
+            private int streamingNotificationThreadPoolSize = 4;
+            private long streamingNotificationBlockingTimeMs = 200;
+            private int streamingNotificationBatchSize = 20;
```
zhangn49, line 304: Would be great if these default values here are same as corresponding fields of CorfuRuntime

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
annym: Done.

PR #2927, opened 2021-04-06
Full backup restore and more logging

Lujie1996:
## Overview

Description:
Updated APIs to allow backup and restore all UFO tables.
Added extra logging such as elapsed time for backing up / restoring each table, and table names.

Why should this be merged: 
Rolling upgrade requires Corfu to back up all tables before upgrade so that they are able to roll back Corfu to a previous state if something failed after the upgrade.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/Backup.java
+++ b/runtime/src/main/java/org/corfudb/runtime/Backup.java
@@ -15,6 +15,8 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -73,18 +75,22 @@ public class Backup {
     }
 
     /**
-     * The difference from the constructor above is that this one will automatically discover
-     * tables which require back up from their tags.
+     * Discover and back up all tables, or tables with requires_backup_support tag
      *
-     * @param filePath      - the filePath where the generated backup tar file will be placed
-     * @param runtime       - the runtime which is performing the back up
-     * @throws IOException  - when failed to create the temp directory
+     * @param filePath          - the filePath where the generated backup tar file will be placed
+     * @param runtime           - the runtime which is performing the back up
+     * @param taggedTablesOnly  - if true, back up tables which has requires_backup_support tag set;
+     *                            if false, back up all UFO tables
      */
-    public Backup(String filePath, CorfuRuntime runtime) throws IOException {
+    public Backup(String filePath, CorfuRuntime runtime, boolean taggedTablesOnly) throws IOException {
         this.filePath = filePath;
         this.backupTempDirPath = Files.createTempDirectory(BACKUP_TEMP_DIR_PREFIX).toString();
         this.runtime = runtime;
-        this.streamIDs = getTaggedTables();
+        if (taggedTablesOnly) {
+            this.streamIDs = getTaggedTables();
+        } else {
+            this.streamIDs = getAllTables();
+        }
     }
 
     /**
@@ -139,15 +145,24 @@ public class Backup {
             return;
         }
 
+        long startTime = System.currentTimeMillis();
+        Map<UUID, String> streamIdToTableNameMap = getStreamIdToTableNameMap();
         for (UUID streamId : streamIDs) {
             if (!tableExists(streamId)) {
-                log.warn("cannot back up a non-existent table: {}", streamId);
+                log.warn("cannot back up a non-existent table stream id {} table name {}",
+                        streamId, streamIdToTableNameMap.get(streamId));
                 continue;
             }
-            String fileName = backupTempDirPath + File.separator + streamId;
-            backupTable(fileName, streamId);
+
+            // temporary backup file's name format: uuid.namespace$tableName
+            Path filePath = Paths.get(backupTempDirPath)
+                    .resolve(streamId + "." + streamIdToTableNameMap.get(streamId));
+            backupTable(filePath, streamId);
         }
-        log.info("successfully backed up {} tables to {} directory", streamIDs.size(), backupTempDirPath);
+        long elapsedTime = System.currentTimeMillis() - startTime;
+
+        log.info("successfully backed up {} tables to {} directory, elapsed time {}ms",
+                streamIDs.size(), backupTempDirPath, elapsedTime);
     }
 
     /**
@@ -156,43 +171,53 @@ public class Backup {
      * If the log is trimmed at timestamp, the backupTable will fail.
      * If the table has no data to be backed up, it will create a file with empty contents.
      *
-     * @param fileName   - the name of the backup file
+     * @param filePath   - the path of the backup file
      * @param uuid       - the uuid of the table which is being backed up
      * @throws IOException
      */
-    private void backupTable(String fileName, UUID uuid) throws IOException {
-        try (FileOutputStream fileOutput = new FileOutputStream(fileName)) {
+    private void backupTable(Path filePath, UUID uuid) throws IOException {
+        long startTime = System.currentTimeMillis();
+        BackupTableStats backupTableStats;
+
+        try (FileOutputStream fileOutput = new FileOutputStream(filePath.toString())) {
             StreamOptions options = StreamOptions.builder()
                     .ignoreTrimmed(false)
                     .cacheEntries(false)
                     .build();
             Stream<OpaqueEntry> stream = (new OpaqueStream(runtime.getStreamsView().get(uuid, options))).streamUpTo(timestamp);
 
-            writeTableToFile(fileOutput, stream, uuid);
+            backupTableStats = writeTableToFile(fileOutput, stream, uuid);
         } catch (IOException e) {
-            log.error("failed to back up table {} to file {}", uuid, fileName);
+            log.error("failed to back up table {} to file {}", uuid, filePath);
             throw e;
         } catch (TrimmedException e) {
             log.error("failed to back up tables as log was trimmed after back up starts.");
             throw e;
         }
 
-        log.info("{} table is backed up and stored to temp file {}", uuid, fileName);
+        long elapsedTime = System.currentTimeMillis() - startTime;
+
+        log.info("{} entries (size: {} bytes, elapsed time: {} ms) saved to temp file {}",
+                backupTableStats.getNumOfEntries(), backupTableStats.getTableSize(), elapsedTime, filePath);
     }
 
-    private void writeTableToFile(FileOutputStream fileOutput, Stream<OpaqueEntry> stream, UUID uuid) throws IOException {
+    private BackupTableStats writeTableToFile(FileOutputStream fileOutput, Stream<OpaqueEntry> stream, UUID uuid) throws IOException {
         Iterator<OpaqueEntry> iterator = stream.iterator();
+        int numOfEntries = 0;
+        int tableSize = 0;
         while (iterator.hasNext()) {
+            numOfEntries++;
             OpaqueEntry lastEntry = iterator.next();
             List<SMREntry> smrEntries = lastEntry.getEntries().get(uuid);
             if (smrEntries != null) {
                 Map<UUID, List<SMREntry>> map = new HashMap<>();
                 map.put(uuid, smrEntries);
                 OpaqueEntry newOpaqueEntry = new OpaqueEntry(lastEntry.getVersion(), map);
-                OpaqueEntry.write(fileOutput, newOpaqueEntry);
+                tableSize = OpaqueEntry.write(fileOutput, newOpaqueEntry);
             }
         }
         fileOutput.flush();
+        return new BackupTableStats(numOfEntries, tableSize);
     }
 
     /**
@@ -210,6 +235,8 @@ public class Backup {
 
         try (FileOutputStream fileOutput = new FileOutputStream(filePath);
              TarArchiveOutputStream tarOutput = new TarArchiveOutputStream(fileOutput)) {
+            // truncate file names if too long
+            tarOutput.setLongFileMode(TarArchiveOutputStream.LONGFILE_TRUNCATE);
             for (File srcFile : srcFiles) {
                 addToTarFile(srcFile, tarOutput);
             }
@@ -247,7 +274,7 @@ public class Backup {
     }
 
     /**
-     * Find tables which have the requires_backup_support tag set
+     * Get UUIDs of tables which have the requires_backup_support tag set
      *
      * @return  List<UUID>   - a list of UUIDs for all the tables which require backup support
      */
@@ -265,11 +292,36 @@ public class Backup {
                 .map(tableName -> CorfuRuntime
                         .getStreamID(TableRegistry.getFullyQualifiedTableName(tableName)))
                 .collect(Collectors.toList());
-        log.info("{} tables need to be backed up: {}", tables.size(), tables);
+        log.info("{} tables need to be backed up.", tables.size());
 
         return tables;
     }
 
+    /**
+     * Get UUIDs of all registered UFO tables
+     *
+     * @return  List<UUID>   - a list of UUIDs for all registered UFO tables
+     */
+    private List<UUID> getAllTables() {
+        return runtime.getTableRegistry().listTables().stream()
+                .map(table -> CorfuRuntime.getStreamID(
+                        TableRegistry.getFullyQualifiedTableName(table.getNamespace(), table.getTableName())))
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Get a map which maps stream ID to TableName
+     *
+     * @return  Map<UUID, String>   - a map: streamId -> fully qualified table name
+     */
+    private Map<UUID, String> getStreamIdToTableNameMap() {
+        Map<UUID, String> streamIdToTableNameMap = new HashMap<>();
+        runtime.getTableRegistry().listTables().forEach(tableName -> {
+                String name = TableRegistry.getFullyQualifiedTableName(tableName);
+                streamIdToTableNameMap.put(CorfuRuntime.getStreamID(name), name);
+        });
+        return streamIdToTableNameMap;
+    }
     /**
      * Cleanup the table backup files under the backupDir directory.
      */
@@ -278,6 +330,25 @@ public class Backup {
             FileUtils.deleteDirectory(new File(backupTempDirPath));
         } catch (IOException e) {
             log.error("failed to clean up the temporary backup directory {}", backupTempDirPath);
+        }
+    }
+
+    private static class BackupTableStats {
+
+        private final int numOfEntries;
+        private final int tableSize;
+
+        BackupTableStats(int numOfEntries, int tableSize) {
+            this.numOfEntries = numOfEntries;
+            this.tableSize = tableSize;
+        }
+
+        public int getNumOfEntries() {
+            return numOfEntries;
+        }
+
+        public int getTableSize() {
+            return tableSize;
         }
     }
 }
--- a/runtime/src/main/java/org/corfudb/runtime/Restore.java
+++ b/runtime/src/main/java/org/corfudb/runtime/Restore.java
@@ -8,6 +8,7 @@ import org.corfudb.protocols.logprotocol.OpaqueEntry;
 import org.corfudb.protocols.logprotocol.SMREntry;
 import org.corfudb.runtime.collections.CorfuStore;
 import org.corfudb.runtime.collections.TxnContext;
+import org.corfudb.runtime.view.TableRegistry;
 import org.corfudb.util.serializer.Serializers;
 
 import java.io.File;
@@ -17,6 +18,8 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.lang.reflect.Array;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
@@ -41,11 +44,15 @@ public class Restore {
     // The path of a temporary directory under which the unpacked table's backup files are stored
     private final String restoreTempDirPath;
 
-    // The stream IDs of tables which are restored
-    private List<UUID> streamIDs;
+    // The filename of each table's backup file, format: uuid.namespace$tableName.
+    private final List<String> tableBackups;
 
     // The Corfu Store associated with the runtime
     private CorfuStore corfuStore;
+
+    // If FULL, will clean up all tables before restore;
+    // If PARTIAL, will only clean up tables that are to be restored
+    private RestoreMode restoreMode;
 
     /**
      * Unpacked files from backup tar file are stored under RESTORE_TEMP_DIR. They are deleted after restore finishes.
@@ -57,11 +64,12 @@ public class Restore {
      * @param runtime       - the runtime which is performing the restore
      * @throws IOException when failed to create the temp directory
      */
-    public Restore(String filePath, CorfuRuntime runtime) throws IOException {
+    public Restore(String filePath, CorfuRuntime runtime, RestoreMode restoreMode) throws IOException {
         this.filePath = filePath;
         this.restoreTempDirPath = Files.createTempDirectory(RESTORE_TEMP_DIR_PREFIX).toString();
-        this.streamIDs = new ArrayList<>();
+        this.tableBackups = new ArrayList<>();
         this.corfuStore = new CorfuStore(runtime);
+        this.restoreMode = restoreMode;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -139,12 +142,17 @@ private void backup() throws IOException {
...
             if (!tableExists(streamId)) {
-                log.warn("cannot back up a non-existent table: {}", streamId);
+                log.warn("cannot back up a non-existent table stream id {} table name {}",
+                        streamId, streamIdToTableNameMap.get(streamId));
                 continue;
             }
-            String fileName = backupTempDirPath + File.separator + streamId;
+
+            // temporary backup file's name format: uuid.namespace$tableName
+            String fileName = backupTempDirPath + File.separator + streamId +
```
xnull, line 154: would be great to use Path instead
```
Path fileName = Paths.get(backupTempDirPath).resolve(streamId +"." + streamIdToTableNameMap.get(streamId));
```

achinmishra (commented):
Is there a possibility someone could mess up the tables written to the file? like delete the backed up file from system? Should we add like a checksum to these files and try to identify if these backed up files were tempered with when restoring?

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -42,11 +44,15 @@
...
-    private List<UUID> streamIDs;
+    private List<String> tableBackups;
 
     // The Corfu Store associated with the runtime
     private CorfuStore corfuStore;
 
+    // If set true, will clean up all tables before restore;
+    // If set false, will only clean up tables that are to be restored
+    private boolean clearAllTablesBeforeRestore = false;
+
```
xcchang, line 55: nit: is this initialization needed here? also, I thought `tableBackups`, `corfuStore` and `clearAllTablesBeforeRestore` could be set to final

Lujie1996:
> Is there a possibility someone could mess up the tables written to the file? like delete the backed up file from system? Should we add like a checksum to these files and try to identify if these backed up files were tempered with when restoring?
> What if the restore is expecting certain set of tagged tables from backup and restore fails to restore a certain table, say may be the way the data was written to the table? There were issues we have seen where the entries of the tables get messed up and there is no way to read that table any more.

@achinmishra  From the current consumer's perspective (Graham), they will make sure that the backup file is safely stored. It's a good point though to have our own mechanism to protect the file integrity and actually they have asked us to provide backup file encryption in the future. As for restore, the current design is to restore whichever tables that are provided in the tar file. We don't expect to restore anything other than that, such as all tagged tables.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: You are right. Removed the initial value.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -265,11 +282,36 @@ private void addToTarFile(File tableFile,TarArchiveOutputStream tarOutput) throw
...
+                .map(table -> CorfuRuntime.getStreamID(
+                        TableRegistry.getFullyQualifiedTableName(table.getNamespace(), table.getTableName())))
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Get a map which maps stream ID to TableName
+     *
+     * @return  Map<>UUID, String</>   - a map: streamId -> fully qualified table name
+     */
```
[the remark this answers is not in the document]
Lujie1996: Yes. Done.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -265,11 +282,36 @@ private void addToTarFile(File tableFile,TarArchiveOutputStream tarOutput) throw
...
+    private Map<UUID, String> getStreamIdToTableNameMap() {
+        Map<UUID, String> streamIdToTableNameMap = new HashMap<>();
+        runtime.getTableRegistry().listTables().forEach(tableName -> {
+                String name = TableRegistry.getFullyQualifiedTableName(tableName);
+                streamIdToTableNameMap.put(CorfuRuntime.getStreamID(name), name);
+        });
+        return streamIdToTableNameMap;
+    }
```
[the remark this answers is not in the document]
Lujie1996: I think `putIfAbsent ()` inside forEach needs to be synchronized, so introducing parallelism is not that helpful right?

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -84,12 +93,17 @@ public void start() throws IOException {
     }
 
     private void restore() throws IOException {
-        for (UUID streamId : streamIDs) {
-            String fileName = restoreTempDirPath + File.separator + streamId;
+        if (clearAllTablesBeforeRestore) {
+            clearAllTables();
+        }
+
```
zhangn49, line 99: total time?

runtime/src/main/java/org/corfudb/runtime/Backup.java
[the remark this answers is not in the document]
zhangn49: @xcchang suggested to use a ConcurrentHashMap. It can be parallelized here, but I think it won't benefit too much from the parallelism - the task it not cpu intensive.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -84,12 +93,17 @@ public void start() throws IOException {
...
             try {
-                restoreTable(fileName, streamId);
+                Path tableBackupPath = Paths.get(restoreTempDirPath).resolve(tableBackup);
+                restoreTable(tableBackupPath, streamId);
             } catch (IOException e) {
-                log.error("failed to restore table {} from temp file {}", streamId, fileName);
+                log.error("failed to restore table {} from temp file {}", streamId, tableBackup);
                 throw e;
             }
         }
```
zhangn49, line 109: Any performance requirement we have? Is it possible to restore tables in parallel?

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -73,18 +75,21 @@ public Backup(String filePath, List<UUID> streamIDs, CorfuRuntime runtime) throw
...
-     * The difference from the constructor above is that this one will automatically discover
-     * tables which require back up from their tags.
+     * Discover and back up all tables, or tables with requires_backup_support tag
      *
-     * @param filePath      - the filePath where the generated backup tar file will be placed
-     * @param runtime       - the runtime which is performing the back up
-     * @throws IOException  - when failed to create the temp directory
+     * @param filePath          - the filePath where the generated backup tar file will be placed
+     * @param runtime           - the runtime which is performing the back up
+     * @param backupAllTables   - if true, back up all UFO tables; if false, back up all tagged tables
```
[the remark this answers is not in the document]
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -156,33 +166,39 @@ private void backup() throws IOException {
...
-     * @param fileName   - the name of the backup file
+     * @param filePath   - the path of the backup file
      * @param uuid       - the uuid of the table which is being backed up
      * @throws IOException
      */
-    private void backupTable(String fileName, UUID uuid) throws IOException {
-        try (FileOutputStream fileOutput = new FileOutputStream(fileName)) {
+    private void backupTable(Path filePath, UUID uuid) throws IOException {
+        int numberOfEntries = 0;
+        long startTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -139,13 +144,18 @@ private void backup() throws IOException {
...
             }
-            String fileName = backupTempDirPath + File.separator + streamId;
-            backupTable(fileName, streamId);
+
+            // temporary backup file's name format: uuid.namespace$tableName
+            Path filePath = Paths.get(backupTempDirPath)
+                    .resolve(streamId + "." + streamIdToTableNameMap.get(streamId));
+            backupTable(filePath, streamId);
         }
         log.info("successfully backed up {} tables to {} directory", streamIDs.size(), backupTempDirPath);
```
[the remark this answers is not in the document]
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: For now we don't have requirements for performance. It's a good idea to explore parallelism to boost speed. One thing worth mentioning is that since both backup and restore use OpaqueEntry, they don't involve the compute-intensive step of serialize/deserialize of SMR objects, so I suppose exploiting CPU parallelism will not benefit too much in this case.

zhangn49 (approved):
LGTM.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -54,6 +59,17 @@
     // All tables in Corfu Db
     private List<UUID> allTablesInDb;
 
+    // Metrics
+    private final Optional<DistributionSummary> backupTimeDistributionSummary =
+            MeterRegistryProvider.getInstance().map(registry ->
+                    DistributionSummary.builder("backup.table.time")
```
PavelZaytsev, line 65: Use timer for measuring times/latencies, also don't forget to include percentiles.
publishPercentiles(0.50, 0.99)
.publishPercentileHistogram()

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -54,6 +59,17 @@
...
+                            .baseUnit("ms").register(registry));
+
+    private final Optional<DistributionSummary> backupEntryCountDistributionSummary =
+            MeterRegistryProvider.getInstance().map(registry ->
+                    DistributionSummary.builder("backup.table.entryCount")
```
PavelZaytsev, line 70: Add 
publishPercentiles(0.50, 0.99)
.publishPercentileHistogram()

Distributions usually have very long tails, which are better captured with percentiles. Otherwise, you would just see mean and max, and you will not be able to accurately judge what's going on.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -156,33 +185,42 @@ private void backup() throws IOException {
...
+            log.error("failed to back up table {} to file {}", uuid, filePath.toString());
             throw e;
         } catch (TrimmedException e) {
             log.error("failed to back up tables as log was trimmed after back up starts.");
             throw e;
         }
 
-        log.info("{} table is backed up and stored to temp file {}", uuid, fileName);
+        long elapsedTime = System.currentTimeMillis() - startTime;
+        backupTimeDistributionSummary.ifPresent(summary -> summary.record(elapsedTime));
```
PavelZaytsev, line 213: Better wrap the code block from line 194 to 212 in runnable and just record the runnable with timer.

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -42,11 +49,26 @@
...
     // The Corfu Store associated with the runtime
     private CorfuStore corfuStore;
 
+    // If set true, will clean up all tables before restore;
+    // If set false, will only clean up tables that are to be restored
+    private boolean clearAllTablesBeforeRestore;
+
+    // Metrics
+    private final Optional<DistributionSummary> restoreTimeDistributionSummary =
+            MeterRegistryProvider.getInstance().map(registry ->
```
PavelZaytsev, line 63: Timer + percentiles

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -42,11 +49,26 @@
...
+                    DistributionSummary.builder("restore.table.time")
+                            .baseUnit("ms").register(registry));
+
+    private final Optional<DistributionSummary> restoreEntryCountDistributionSummary =
```
PavelZaytsev, line 67: Percentiles

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -121,9 +158,16 @@ private void restoreTable(String fileName, UUID streamId) throws IOException {
                 numEntries += smrEntries.size();
             }
             txn.commit();
-            log.info("completed restore of table {} with {} numEntries", streamId, numEntries);
+
+            long elapsedTime = System.currentTimeMillis() - startTime;
+            restoreTimeDistributionSummary.ifPresent(summary -> summary.record(elapsedTime));
```
PavelZaytsev, line 163: Record runnable with timer

PavelZaytsev (commented):
Just some changes

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: The appliance backup&restore team asks to log the times outside metrics log as well, so I have to explicitly take the timer and add it to both backup&restore log and metrics log (DistributionSummary).

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: Added. Thanks.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: After discussion, I think using Metrics to record backup/restore time per table does not give more insight comparing to using simple log messages. Backup/restore is one-time action and does not fit the aggregation and periodic report fashion of Metrics. However, the corfu runtime Metrics will still be used to monitor the low level statistics such as read/write latency.

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -43,7 +45,7 @@
     private final String backupTempDirPath;
 
     // The stream IDs of tables which are backed up
-    private final List<UUID> streamIDs;
+    private List<UUID> streamIDs;
```
xnull, line 48: why not final?

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -156,33 +171,40 @@ private void backup() throws IOException {
...
      * If the table has no data to be backed up, it will create a file with empty contents.
      *
-     * @param fileName   - the name of the backup file
+     * @param filePath   - the path of the backup file
      * @param uuid       - the uuid of the table which is being backed up
      * @throws IOException
      */
-    private void backupTable(String fileName, UUID uuid) throws IOException {
-        try (FileOutputStream fileOutput = new FileOutputStream(fileName)) {
+    private void backupTable(Path filePath, UUID uuid) throws IOException {
```
xnull, line 178: Why not throw something more application-specific? Like BackupException?

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -156,33 +171,40 @@ private void backup() throws IOException {
...
                     .ignoreTrimmed(false)
                     .cacheEntries(false)
                     .build();
             Stream<OpaqueEntry> stream = (new OpaqueStream(runtime.getStreamsView().get(uuid, options))).streamUpTo(timestamp);
 
-            writeTableToFile(fileOutput, stream, uuid);
+            numberOfEntries = writeTableToFile(fileOutput, stream, uuid);
         } catch (IOException e) {
-            log.error("failed to back up table {} to file {}", uuid, fileName);
+            log.error("failed to back up table {} to file {}", uuid, filePath.toString());
```
xnull, line 191: don't need to use .toString(), you can use as is

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -156,33 +171,40 @@ private void backup() throws IOException {
...
         }
 
-        log.info("{} table is backed up and stored to temp file {}", uuid, fileName);
+        long elapsedTime = System.currentTimeMillis() - startTime;
+
+        log.info("{} entries (elapsed time {}ms) saved to temp file {}", numberOfEntries, elapsedTime, filePath.toString());
     }
 
-    private void writeTableToFile(FileOutputStream fileOutput, Stream<OpaqueEntry> stream, UUID uuid) throws IOException {
+    private int writeTableToFile(FileOutputStream fileOutput, Stream<OpaqueEntry> stream, UUID uuid) throws IOException {
```
xnull, line 203: Would be great to create a class with some stats and return it instead of just a number of tables
For instance
```
class WriteTableStats {
    int tablesNum;
    long totalBytesWritten;
    List<UUID> tableNames
}
```

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -247,11 +272,11 @@ private void addToTarFile(File tableFile,TarArchiveOutputStream tarOutput) throw
     }
 
     /**
-     * Find tables which have the requires_backup_support tag set
+     * Get UUIDs of tables which have the requires_backup_support tag set
      *
      * @return  List<UUID>   - a list of UUIDs for all the tables which require backup support
      */
-    private List<UUID> getTaggedTables() {
+    private List<UUID> getTaggedTablesUUID() {
```
xnull, line 279: previous name was better, since you use strongly typed language it's not recommended to use a type descriptor in the variable name

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -265,11 +290,36 @@ private void addToTarFile(File tableFile,TarArchiveOutputStream tarOutput) throw
                 .map(tableName -> CorfuRuntime
                         .getStreamID(TableRegistry.getFullyQualifiedTableName(tableName)))
                 .collect(Collectors.toList());
-        log.info("{} tables need to be backed up: {}", tables.size(), tables);
+        log.info("{} tables need to be backed up.", tables.size());
```
xnull, line 293: prev version is a bit better in terms of logs parsing

runtime/src/main/java/org/corfudb/runtime/Backup.java
```diff
@@ -265,11 +290,36 @@ private void addToTarFile(File tableFile,TarArchiveOutputStream tarOutput) throw
...
 
+    /**
+     * Get UUIDs of all registered UFO tables
+     *
+     * @return  List<UUID>   - a list of UUIDs for all registered UFO tables
+     */
+    private List<UUID> getAllTablesUUID() {
+        return runtime.getTableRegistry().listTables().stream()
+                .map(table -> CorfuRuntime.getStreamID(
+                        TableRegistry.getFullyQualifiedTableName(table.getNamespace(), table.getTableName())))
```
xnull, line 306: can we add `getFullyQualifiedTableName` method into `table`? then it will be much cleaner: `table.getFullyQualifiedTableName()` and more logically consistent

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -42,11 +46,15 @@
     private final String restoreTempDirPath;
 
     // The stream IDs of tables which are restored
-    private List<UUID> streamIDs;
+    private List<String> tableBackups;
```
xnull, line 49: why not final?

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -57,11 +65,12 @@
      * @param runtime       - the runtime which is performing the restore
      * @throws IOException when failed to create the temp directory
      */
-    public Restore(String filePath, CorfuRuntime runtime) throws IOException {
+    public Restore(String filePath, CorfuRuntime runtime, boolean clearAllTablesBeforeRestore) throws IOException {
```
xnull, line 68: would be great to have an enum instead of boolean, like:
`enum PostRestore{
  CLEAN, EMPTY_ACTION
}
`

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -84,30 +93,44 @@ public void start() throws IOException {
...
     private void restore() throws IOException {
-        for (UUID streamId : streamIDs) {
-            String fileName = restoreTempDirPath + File.separator + streamId;
+        if (clearAllTablesBeforeRestore) {
+            clearAllTables();
+        }
+
+        long startTime = System.currentTimeMillis();
+        for (String tableBackup : tableBackups) {
+            UUID streamId = UUID.fromString(tableBackup.substring(0, tableBackup.indexOf(".")));
```
xnull, line 102: would be great to move this into `tableBackup`, like: 
```
UUID streamId = tableBackup.getStreamID()
```

runtime/src/main/java/org/corfudb/runtime/Restore.java
```diff
@@ -171,4 +199,13 @@ private void verify() {
     private void cleanup() throws IOException {
         FileUtils.deleteDirectory(new File(restoreTempDirPath));
     }
+
+    private void clearAllTables() {
+        TxnContext txn = corfuStore.txn(CORFU_SYSTEM_NAMESPACE);
+        Collection<TableName> tables = corfuStore.listTables(null);
```
xnull, line 205: let's pass an empty list, instead of null

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
```diff
@@ -271,7 +271,8 @@ private void compareCorfuStoreTables(CorfuStore corfuStore1, String tableName1,
         TarArchiveInputStream tarInput = new TarArchiveInputStream(fileInput);
         TarArchiveEntry entry;
         while ((entry = tarInput.getNextTarEntry()) != null) {
-            streamIDs.add(UUID.fromString(entry.getName()));
+            String streamId = entry.getName().substring(0, entry.getName().indexOf("."));
```
xnull, line 274: `entry.getStreamId()`

xnull (approved):
looks good

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: Done. I made some temporary changes and forgot to bring it back.

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: This function only writes one table, so `tableNums` and `tableNames` are not really meaningful here. It's a good idea to wrap `numOfEntries` and `totablBytesWritten` in a class. For the current requirements we have received it is enough.

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: `tables` is usually a very large set and considering there is already logging for each table, it could be cleaner to not log them again here.
`log.info("{} entries (elapsed time {}ms) saved to temp file {}", numberOfEntries, elapsedTime, filePath);`
(temp file has both table UUID and table name info)

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: Done.

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: You are right that it can make code looks cleaner. 
`listTables()` returns a collection of TableNames (protobuf message) whose only fields are `namespace` and `table_name`, and `getFullyQualifiedTableName()` is application logic that determines how we present the full name. If we add a new field `fully_qualified_table_name` in `TableName` that would introduce data duplication and stronger coupling of data and logic.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: We can't do that right? `lisTables(String namespace)` when passing null it will return all tables. From the current code if I pass empty string it will only search for empty namespace.
`.filter(tableName -> namespace == null || tableName.getNamespace().equals(namespace))`

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: Good idea! Done.

test/src/test/java/org/corfudb/runtime/BackupRestoreIT.java
Lujie1996: Note that entry is an instance of `TarArchiveEntry`

xnull: sorry, didn't notice

runtime/src/main/java/org/corfudb/runtime/Backup.java
xnull: So, it will return 1 always? Stil why not return more info instead of meaningless int number? Next time when someone will need that info, that developer will need to add that info anyways. And this info is useful even for backup right now, why wouldn't like to provide that info?

runtime/src/main/java/org/corfudb/runtime/Backup.java
xnull: there is no coupling. Actually, the current code creates more boilerplate code than it should be. 
The table always knows its namespace and its own name, so do you really believe `$` symbol makes code more fragile?
Actually, it's vise versa - in the table class we define a contact - what is a full name of the table. Otherwise how a developer should guess that there is a Utility class somewhere, which keeps that information about the contract for the naming?

image

runtime/src/main/java/org/corfudb/runtime/Restore.java
xnull: sorry, `tableBackup` is a string, then question - why we keep strings instead of UUID-s?

runtime/src/main/java/org/corfudb/runtime/Restore.java
xnull: that looks weird, I'd have two different methods then, first gets a list of tables, a second one has no parameters and returns all the tables, the current implementation looks a bit convoluted

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: Just to make sure that we are on the same page -- if we want to do `table.getFullyQualifiedTableName()`, we have to add a field in TableName (table is a protobuf message).
```
// Table name.
message TableName {
    string namespace = 1;
    string table_name = 2;
    string fully_qualified_name = 3;
}
```
If one day we need to change the connector from `$` to `&`, or add support for another kind of tables besides UFO tables which uses whichever other symbol, we will have to do DB migration and rewrite the `fully_qualified_name` field. In other words, we are hardcoding the rule/logic of generating fully_qualified_name (`TableRegistry. getFullyQualifiedTableName`) into actual DB tables.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: `tableBackups` is the file names of what comes out from unpacking the backup tar file, which contains the data of each table that was previously backed up. A tableBackup is more than just UUID, it's actually `uuid.namespace$tableName`. I noticed that I didn't update the comment above this variable so that caused your confusion.

runtime/src/main/java/org/corfudb/runtime/Backup.java
xnull: oh, ok, didn't know that Table is a protobuf class

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: I added a `BackupTableStats` class which reports `numOfEntries` and `tableSize` in bytes.

runtime/src/main/java/org/corfudb/runtime/Restore.java
Lujie1996: This flag is used to indicate if we clean up tables BEFORE restore. I added a enum `RestoreMode` and it has `FULL` (clean up all tables) and `PARTIAL` (only clean up tables that needs to be restored). This makes the API easier to understand.

runtime/src/main/java/org/corfudb/runtime/Backup.java
Lujie1996: Good point. I will have discussion with Pavan and Graham about exception handling.

PR #2932, opened 2021-04-09
Move version checking to handshake and deprecate Version RPC

xcchang:
## Overview

Description: In this PR we move the version checking to the RPC layer (the handshake handler), and added version info to RPC header such that we have the capability to detect when client and server are running different versions of corfu. The Version RPC has been removed since it was previously only used upon connection establish in CorfuRuntime. Unit test coverage for ClientHandshakeHandler and ServerHandshakeHandler is also added.

Why should this be merged: Done as part of the effort for rolling upgrade.

- [x] Public API has Javadoc


common/src/main/java/org/corfudb/common/CorfuSourceCodeVersion.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Getter for the long value which will be populated into our protobuf header. We only take the
+     * first {@link #CUTOFF} characters of {@link #CORFU_SOURCE_VERSION_STRING} as they
+     * are more than enough to uniquely identify a commit in our codebase.
+     *
+     * @return Long value converted from the first 12 characters of {@link #CORFU_SOURCE_VERSION_STRING}
+     */
+    public static long getCorfuSourceCodeVersion() {
+        return Long.parseLong(CORFU_SOURCE_VERSION_STRING.substring(0, CUTOFF), HEX_OPT);
```
xnull, line 33: GitRepositoryState should be able to provide all needed information, please check this out https://github.com/CorfuDB/CorfuDB/blob/d42d17b360acfd2eccac4862b8f1fd1a79f75586/infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java#L433

xcchang: Yeah we also explored that option, but we want a separate static definition because instead of updating this "version" every time the master branch get updated, we may just do it for every release or artifact deployment. This PR is still in draft mode and we are continue exploring automatic way to update the commit id here like using scripting, and it will be populated to other places like the handshake and layout.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
```diff
@@ -45,6 +42,18 @@ private CorfuProtocolMessage() {}
         IGNORE
     }
 
+    /**
+     * Returns a ProtocolVersionMsg with default capability_vector and static
+     * corfu_source_code_version.
```
chetangudisagar, line 47: As we are not setting any default value, change `default capability_vector` to `empty capability_vector list` maybe?

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
```diff
@@ -45,6 +42,18 @@ private CorfuProtocolMessage() {}
...
+     *
+     * @return A ProtocolVersionMsg with default capability_vector.
```
chetangudisagar, line 49: same here.

common/src/main/java/org/corfudb/common/CorfuSourceCodeVersion.java
hisundar: GitRepositoryState isn't a library, it is a class written to extract certain generated files like "git.branch" etc and then populating it. These property files need to be correctly filled up for it to work.
What @xcchang is doing is essentially short-circuiting the process and making it directly available so it works in all the cases.
Automation to fill this can be added later.

xnull: BTW, can we get corfu version from corfu app? Do we need anything else than the version? 
I mean we have "0.3.2.1" we know everything about the build, and release number?

I'm trying to understand the motivation here, it seems that ($major.$minor.$fix) semver works well in our case?

hisundar: good question - it's just a safeguard to help folks with build to build breakages, if we retain major.minor then it becomes hard to test build-to-build upgrade failures, since these versions are not going to be modified in build-to-build setups. 
also having a git commit id in the version field protects us from the case where the version wasn't updated.

common/src/main/java/org/corfudb/common/CorfuSourceCodeVersion.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Static class for representing the corfu source code version using git commit id.
+ */
+public final class CorfuSourceCodeVersion {
+
+    /**
+     * Original git commit id of the head of master branch, which should be updated upon every
+     * corfu artifact deployment.
+     */
+    private static final String CORFU_SOURCE_VERSION_STRING = "d42d17b360acfd2eccac4862b8f1fd1a79f75586";
```
chetangudisagar, line 13: nit: can we rename it to `CORFU_SOURCE_CODE_VERSION_STRING` to keep it consistent with the class name?

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
```diff
@@ -168,6 +169,12 @@ default boolean validateRequest(RequestMsg req, ChannelHandlerContext ctx) {
             return true;
         }
 
+        // TODO: In some cases this needs to be handled specially in the future
```
chetangudisagar, line 172: Could you please help me understand more as to in which cases we need to handle this especially?  Do you mean specific handling(throw error/print and return something) in the case of a version mismatch?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -106,14 +118,13 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
         // Store clientID as a channel attribute.
         ctx.channel().attr(clientIdAttrKey).set(clientId);
         log.info("channelRead: Handshake validated by Server.");
-        log.debug("channelRead: Sending handshake response: Node Id: {} Corfu Version: {}",
-                this.nodeId, this.corfuVersion);
+        log.debug("channelRead: Sending handshake response: Node Id: {}", this.nodeId);
```
chetangudisagar, line 121: Can we print a debug message of the server version like before? Might help in bug triaging as this is the only place where we are doing main version checks.

```
log.debug("channelRead: Sending handshake response: Node Id: {}, ServerVersion: {}, ClientVersion: {}", this.nodeId, serverVersion.toString(), clientVersion.toString());
````

File: infrastructure/src/test/java/org/corfudb/infrastructure/BaseServerTest.java [hunk 7]
```diff
@@ -182,37 +177,6 @@ public void testSealWrongEpoch() {
...
-        assertTrue(compareBaseHeaderFields(request.getHeader(), response.getHeader()));
-        assertTrue(response.getPayload().hasVersionResponse());
-
-        // Assert that the contents of the versionInfo are what we expect
-        VersionInfo versionInfo = getVersionInfo(response.getPayload().getVersionResponse());
-        assertEquals("localhost", versionInfo.getNodeId());
-        assertEquals(1, versionInfo.getOptionsMap().size());
-        assertEquals("localhost", versionInfo.getOptionsMap().get("--address"));
-    }
-
```
chetangudisagar, line 215: If required, can we write some test cases to check that the warning message was printed when there is a version mismatch and was not printed when there is no mismatch? Later this can be extended to checking the thrown exceptions or printing some other error/warn messages.

runtime/src/main/java/org/corfudb/util/GitRepositoryState.java
```diff
@@ -76,4 +76,19 @@ public static GitRepositoryState getRepositoryState() {
...
+     * @return Long value converted from the first 12 characters of parent git commit id.
+     */
+    public static long getCorfuSourceCodeVersion() {
+        if (_gitRepositoryState == null) {
+            getRepositoryState();
+        }
+        final int CUTOFF = 12;
+        final int HEX_OPT = 16;
+        return Long.parseLong(_gitRepositoryState.commitId.substring(0, CUTOFF), HEX_OPT);
+    }
```
chetangudisagar, line 93: Maybe we can write some unit tests for this method too.

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
xcchang: I thought currently this is mainly used for logging and as a placeholder. From our discussion before, I remember we want to forward requestMsg with different versions to different versions of handlers/servers (for example, `Sequencer ServerV2` or `handleTokenRequestV2`, this part of design is still pending to be determined though)

zfrenette (commented):
First batch of comments -- Can we add tests in this PR for these versioning changes?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
xcchang: Added, thanks!

test/src/test/java/org/corfudb/runtime/clients/BaseHandlerTest.java
```diff
@@ -32,9 +32,4 @@
                 .add(baseHandler)
                 .build();
     }
-
-    @Test
-    public void canGetVersionInfo() {
-        CFUtils.getUninterruptibly(client.getVersionInfo());
-    }
```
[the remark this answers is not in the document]
xcchang: Removed, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
```diff
@@ -168,6 +169,12 @@ default boolean validateRequest(RequestMsg req, ChannelHandlerContext ctx) {
...
+        if (requestHeader.getVersion().getCorfuSourceCodeVersion() !=
+                GitRepositoryState.getCorfuSourceCodeVersion()) {
+            log.warn("validateRequest: Version mismatch.");
```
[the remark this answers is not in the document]
xcchang: Added, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
```diff
@@ -324,8 +322,6 @@ protected void initChannel(@Nonnull Channel ch) throws Exception {
                 ch.pipeline().addLast(new NettyCorfuMessageDecoder());
                 ch.pipeline().addLast(new NettyCorfuMessageEncoder());
                 ch.pipeline().addLast(new ServerHandshakeHandler(context.getNodeId(),
-                        Version.getVersionString() + "("
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -67,16 +65,30 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
...
+            log.warn("channelRead: Dropping the message as the received msg is not an instance" +
+                    " of RequestMsg. Message - {}", m);
+            return;
+        }
+
+        RequestMsg requestMsg = (RequestMsg) m;
+        long clientVersion = requestMsg.getHeader().getVersion().getCorfuSourceCodeVersion();
+        long serverVersion = GitRepositoryState.getCorfuSourceCodeVersion();
+        if (clientVersion != serverVersion) {
+            log.warn("channelRead: Version mismatch. Corfu client version: {}, Corfu server version: {}",
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -129,7 +130,11 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
             return;
         }
 
-        log.info("channelRead: Handshake succeeded. Server Corfu Version: [{}]", corfuVersion);
+        log.info("channelRead: Handshake succeeded. Corfu Server Version: [{}]", corfuServerVersion);
+        if (corfuServerVersion != GitRepositoryState.getCorfuSourceCodeVersion()) {
+            log.warn("channelRead: Version mismatch. Corfu client version: {}, Corfu server version: {}",
+                    GitRepositoryState.getCorfuSourceCodeVersion(), corfuServerVersion);
```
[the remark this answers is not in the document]
xcchang: Done, thanks!

infrastructure/src/test/java/org/corfudb/infrastructure/BaseServerTest.java
xcchang: Unit tests added, thanks!

common/src/main/java/org/corfudb/common/CorfuSourceCodeVersion.java
xcchang: Note: We had off-line discussions and found the `git.properties` file is actually carried into `runtime.jar` and `infrastructure.jar`, so it could be good to use the `GitRepositoryState`.

infrastructure/src/main/java/org/corfudb/infrastructure/IServerRouter.java
xcchang: Note: We finally decide to remove this logging here to avoid flooding the logs.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
```diff
@@ -45,6 +42,18 @@ private CorfuProtocolMessage() {}
...
 
+    /**
+     * Returns a ProtocolVersionMsg with empty capability_vector and static
+     * corfu_source_code_version.
+     *
+     * @return A ProtocolVersionMsg with empty capability_vector.
+     */
+    public static ProtocolVersionMsg getDefaultProtocolVersionMsg() {
+        return ProtocolVersionMsg.newBuilder()
+                .setCorfuSourceCodeVersion(GitRepositoryState.getCorfuSourceCodeVersion())
```
zfrenette, line 53: Would it be better to initialize the compatibility vector as well? It seems like right now we are using an empty list to denote the default capabilities instead of a vector of all zeros.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -1294,8 +1244,6 @@ public synchronized CorfuRuntime connect() {
             }
         }
 
-        checkVersion();
```
zfrenette, line 1297: How will the runtime become aware of the version of the nodes in the cluster? Will it need to?

runtime/pom.xml
```diff
@@ -173,6 +173,12 @@
             <version>6.0.53</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>${logback.classic.version}</version>
+            <scope>test</scope>
+        </dependency>
```
zfrenette, line 181: Why is this needed? I believe it was removed because of conflicts in NSX:  https://github.com/CorfuDB/CorfuDB/pull/2841/.

xcchang: It is for inspecting the logs in unit test of `ClientHandshakeHandler`. I set the scope to be test hoping to avoid the issue, but I will double check.

runtime/src/test/java/org/corfudb/runtime/ClientHandshakeHandlerTest.java
```diff
@@ -0,0 +1,190 @@
...
+     * with default values.
+     *
+     * @param ignoreClusterId   indicates if the message is clusterId aware
+     * @param ignoreEpoch       indicates if the message is epoch aware
+     * @return                  the corresponding HeaderMsg
+     */
+    private HeaderMsg getBasicHeader(ClusterIdCheck ignoreClusterId, EpochCheck ignoreEpoch) {
+        return getHeaderMsg(requestCounter.incrementAndGet(), PriorityLevel.NORMAL, 0L,
+                getUuidMsg(DEFAULT_UUID), getUuidMsg(DEFAULT_UUID), ignoreClusterId, ignoreEpoch);
+    }
```
zfrenette, line 64: We should consider factoring/restructuring common RPC unit test code in a subsequent PR.

runtime/src/test/java/org/corfudb/runtime/ClientHandshakeHandlerTest.java
```diff
@@ -0,0 +1,190 @@
...
+                SERVER_NODEID, DEFAULT_HANDSHAKE_TIMEOUT);
+        embeddedChannel = new EmbeddedChannel(clientHandshakeHandler);
+
+        // Setting up the logger such that we could inspect the logged messages in ClientHandshakeHandler.
+        Logger logger = (Logger) LoggerFactory.getLogger(ClientHandshakeHandler.class);
+        logger.setLevel(Level.DEBUG);
+
+        listAppender = new ListAppender<>();
+        listAppender.start();
+        logger.addAppender(listAppender);
```
zfrenette, line 78: I think relying on logging for unit tests makes them fragile. Handshake success/failure can be tested by detecting the presence or absence of another message sent down the pipeline. In the case of a version mismatch, maybe there isn't enough substance for a unit test, since it has no impact on the behaviour yet.

infrastructure/src/test/java/org/corfudb/infrastructure/ServerHandshakeHandlerTest.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    private HeaderMsg getBasicHeader(ClusterIdCheck ignoreClusterId, EpochCheck ignoreEpoch) {
+        return getHeaderMsg(requestCounter.incrementAndGet(), PriorityLevel.NORMAL, 0L,
+                getUuidMsg(DEFAULT_UUID), getUuidMsg(DEFAULT_UUID), ignoreClusterId, ignoreEpoch);
+    }
+
+    @Before
+    public void setup() {
+        ServerHandshakeHandler serverHandshakeHandler = new ServerHandshakeHandler(SERVER_NODEID,
+                GitRepositoryState.getCorfuSourceCodeVersion(), DEFAULT_HANDSHAKE_TIMEOUT);
```
zfrenette, line 67: Since this is a unit test, maybe the server version can be mocked as well so it doesn't have to rely on GitRepositoryState.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
```diff
@@ -145,6 +165,14 @@ public static ResponseMsg getResponseMsg(HeaderMsg header, ResponsePayloadMsg re
      * @return        a ResponseMsg with the provided header and error
      */
     public static ResponseMsg getResponseMsg(HeaderMsg header, ServerErrorMsg error) {
+        if (header.getVersion().getCorfuSourceCodeVersion() !=
+                GitRepositoryState.getCorfuSourceCodeVersion()) {
```
zfrenette, line 169: Is this the best place for such a check? The assumption here was that the response header was already built.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
```diff
@@ -131,6 +143,14 @@ public static RequestMsg getRequestMsg(HeaderMsg header, RequestPayloadMsg reque
      * @return           a ResponseMsg with the provided header and payload
      */
     public static ResponseMsg getResponseMsg(HeaderMsg header, ResponsePayloadMsg response) {
+        if (header.getVersion().getCorfuSourceCodeVersion() !=
+                GitRepositoryState.getCorfuSourceCodeVersion()) {
```
zfrenette, line 147: Maybe this can be a private constant in the API class.

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
```diff
@@ -95,6 +96,13 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
         }
 
         ResponseMsg response = ((ResponseMsg) m);
+        long corfuServerVersion = response.getHeader().getVersion().getCorfuSourceCodeVersion();
+        long corfuClientVersion = GitRepositoryState.getCorfuSourceCodeVersion();
+
+        if (corfuServerVersion != GitRepositoryState.getCorfuSourceCodeVersion()) {
```
zfrenette, line 102: `corfuServerVersion != corfuClientVersion`?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
xcchang: Currently the server version is contained in the response header, and for now only the ClientHandshakeHandler will be aware of it.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
xcchang: From our discussion this part would be handled in a separate PR for capability vector implementation.

infrastructure/src/test/java/org/corfudb/infrastructure/ServerHandshakeHandlerTest.java
xcchang: Changed, thanks!

runtime/src/main/java/org/corfudb/protocols/wireprotocol/ClientHandshakeHandler.java
xcchang: Done, thanks!

runtime/pom.xml
xcchang: As we no longer inspect log message in test cases, it is now removed, thanks!

runtime/src/test/java/org/corfudb/runtime/ClientHandshakeHandlerTest.java
xcchang: Changed, thanks!

runtime/src/test/java/org/corfudb/runtime/ClientHandshakeHandlerTest.java
xcchang: Will do this in a following cleanup/refactor PR.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
xcchang: Will do this in a following cleanup/refactor PR.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
xcchang: Will do this in a following cleanup/refactor PR.

runtime/proto/service/corfu_message.proto
```diff
@@ -22,7 +22,8 @@ enum PriorityLevel {
 }
 
 message ProtocolVersionMsg {
-  fixed64 capability_vector = 1;
+  repeated fixed64 capability_vector = 1;
+  int64 corfu_source_code_version = 2;
```
hisundar, line 26: maybe a comment explaining how this is populated?

test/src/test/resources/logback-test.xml
```diff
@@ -61,7 +61,7 @@
     <!-- Control logging levels for individual components here. -->
     <logger name="org.corfudb.runtime.object" level="INFO"/>
     <logger name="org.corfudb.runtime.clients" level="INFO"/>
-    <logger name="org.corfudb.infrastructure" level="INFO"/>
+    <logger name="org.corfudb.infrastructure" level="WARN"/>
```
hisundar, line 64: why did we change the test logging to WARN?

test/src/test/resources/logback-test.xml
```diff
@@ -78,7 +78,7 @@
 
     <root level="INFO">
         <!--<appender-ref ref="FILE" />-->
-        <!--<appender-ref ref="STDOUT" />-->
+        <appender-ref ref="STDOUT" />
```
hisundar, line 81: Can you please remove these changes to the logback since it would create a lot of noise in the GitHub action verifications

hisundar (commented):
Looks good @xcchang just a few minor comments.

test/src/test/resources/logback-test.xml
xcchang: Sorry for the confusion, those from the second commit are just temporary for debugging the flaky test.

hisundar (approved):
LGTM, please remove the test changes once the tests run fine

runtime/src/main/java/org/corfudb/util/GitRepositoryState.java
Maithem, line 93: This has a race condition. Writing to `_gitRepositoryState` and accessing it can happen from multiple threads but it's not synchronized. 

`_gitRepositoryState` can be made a singleton.

test/src/test/java/org/corfudb/CorfuTestParameters.java
```diff
@@ -87,8 +87,8 @@ public CorfuTestParameters(){
 
         // Timeouts
         TIMEOUT_VERY_SHORT = Duration.of(500, MILLIS);
-        TIMEOUT_SHORT = Duration.of(1, SECONDS);
-        TIMEOUT_NORMAL = Duration.of(10, SECONDS);
+        TIMEOUT_SHORT = Duration.of(5, SECONDS);
+        TIMEOUT_NORMAL = Duration.of(60, SECONDS);
         TIMEOUT_LONG = Duration.of(2, MINUTES);
```
Maithem, line 92: This should be part of a separate PR.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
Maithem: +1 @zfrenette  comment. I think it should be addressed in this PR.

test/src/test/java/org/corfudb/CorfuTestParameters.java
xcchang: Yeah this will be removed, it's temporary for debugging a flaky test, sorry for the confusion

runtime/proto/service/corfu_message.proto
Maithem: Please address this.

Maithem (approved):

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
xcchang: Done, thanks!

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolMessage.java
xcchang: Done, thanks!

runtime/src/main/java/org/corfudb/util/GitRepositoryState.java
xcchang: Revised, thanks a lot for pointing out this!

PR #2933, opened 2021-04-09
Exclude CP Entries from AddressSpaceView Cache

zhangn49:
## Overview

Description:

New checkpoint entry batch policy introduced large checkpoint entry, which can be more than 25MB
Exclude all checkpoint entry from the addressSpaceView cache (read cache) to avoid OOM

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -446,6 +446,13 @@ public class AddressSpaceView extends AbstractView {
             return loadedValue;
         }
 
+        // Exclude all checkpoint entries, which might be 25MB and cause OOM.
+        // These entries don't need to be cached because they are not used when
+        // rolling forward/backwards to materialize a snapshot.
+        if (loadedValue.hasCheckpointMetadata()) {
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -446,6 +446,11 @@ private ILogData cacheLoadAndGet(@NonNull Cache<Long, ILogData> cache, long addr
             return loadedValue;
         }
 
+        // Exclude all checkpoint entries, which might be 25MB and cause OOM
```
Maithem, line 449: You can point out that these entries don't need to be cached because they are not used when rolling forward/backwards to materialize a snapshot.

Maithem (approved):

xnull (approved):
looks good

annym (approved):
Thanks for adding tests with the change! LGTM

apavan:
Great work Nan! It was too easy to just parameterize the batch size and change the batch size to 0.1MB to solve the issue but you took a wholistic look and proposed/implemented the perfect fix which brings in much needed memory efficiency.

PR #2941, opened 2021-04-29
Streaming API TrimmedExceptions fix due to runtimeGC

annym:
## Overview

Description: correct corner case for streaming API when th number of updates is zero for a listener and the runtimeGC kicks in modifying boundary conditions (gcTrimMark). 

Why should this be merged: customer bug

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
@@ -396,8 +396,18 @@ public class AddressMapStreamView extends AbstractQueuedStreamView {
 
         long size = streamAddressSpace.size();
         if (size == 0L) {
-            // The trim mark will allow to detect trimmed exceptions if the seeked address falls behind it
-            return streamAddressSpace.getTrimMark();
+            // We want to optimize and directly return when no updates are found in advance (i.e., no need to do remainingUpTo)
+            // but we need to consider the case when the 'syncUpTo' address falls below the trim mark, so we correctly
+            // report the TrimmedException allowing consumers to know that data is no longer available from this point.
+            // We could throw the TrimmedException directly or in this case let remainingUpTo discover it for us.
+            if (getCurrentGlobalPosition() <= streamAddressSpace.getTrimMark()) {
+                // If pointer is -1L we must return the known trim mark or the trimmed exception will be lost
+                return getCurrentGlobalPosition() == Address.NON_ADDRESS ? streamAddressSpace.getTrimMark() : getCurrentGlobalPosition();
+            } else  {
+                // Case: valid sync up to address
+                // Do not sync (remainingUpTo) as we know in advance that there are no updates available
+                return Address.NON_ADDRESS;
+            }
         }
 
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -396,8 +396,18 @@ protected long getMaxGlobalFromMaxEntries(int maxEntries) {
...
         if (size == 0L) {
-            // The trim mark will allow to detect trimmed exceptions if the seeked address falls behind it
-            return streamAddressSpace.getTrimMark();
+            // We want to optimize and directly return when no updates are found in advance (i.e., no need to do remainingUpTo)
+            // but we need to consider the case when the 'syncUpTo' address falls below the trim mark, so we correctly
+            // report the TrimmedException allowing consumers to know that data is no longer available from this point.
+            // We could throw the TrimmedException directly or in this case let remainingUpTo discover it for us.
+            if (getCurrentGlobalPosition() <= streamAddressSpace.getTrimMark()) {
+                // If pointer is -1L we must return the known trim mark or the trimmed exception will be lost
+                return getCurrentGlobalPosition() == Address.NON_ADDRESS ? streamAddressSpace.getTrimMark() : getCurrentGlobalPosition();
```
hisundar, line 405: This -1 pass through logic is tied to how other functions set the GC trim mark and we should consider refactoring this code since it's very non-obvious.

hisundar (approved):
Non-specific comment. looks good to me.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: Not really, this -1 does not have to do with how other functions set the GC trim mark but instead is tied to what we discussed offline. 

We could directly throw a TrimmedException here (because we know for sure the position to which we want to sync up to falls behind the known trim mark). But, Maithem had suggested this method to limit its functions to return an address up to which to sync up to, and let remainingUpTo discover the trim exception. 

With that in mind, if the currentGlobalPosition is -1 (let's say the case where they want to sync the stream from the beginning of time), if we know that it falls behind the trim mark, we can't just return the current global position, otherwise syncing up to -1 is doing nothing, so we'll miss the known TrimmedException. So in this case we force it to sync up to the trim mark (which is above -1 and therefore the remainingUpTo will detect the TrimmedException).

pankti-m, line 405: does getCurrentGlobalPointer() return the max global seen by the sequencer?  or is it tracking the max address for this stream known so far(before querying the sequencer)?

annym: It is tracking the position at which the pointer is currently on for this stream (not necessarily the max)... In the case of subscription API (where this call is coming from) it is up to the position where we want to start syncing or we have synced up to...

pankti-m (approved):

PR #2949, opened 2021-04-30
Sleep between retries

PavelZaytsev:
## Overview

Description:
When we retry actions during clustering workflows, we do not wait between retries. It's unlikely that intermittent server or router issues are resolved within the microseconds. Sometimes the complex clustering workflows like restore fail in production due to this.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Action.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Action.java
@@ -2,15 +2,13 @@ package org.corfudb.infrastructure.orchestrator;
 
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
-import org.corfudb.util.Sleep;
 
 import javax.annotation.Nonnull;
-import java.time.Duration;
+import java.util.concurrent.TimeUnit;
 
 /**
- *
  * A workflow action. All workflow actions must extend this class.
- *
+ * <p>
  * Created by Maithem on 10/25/17.
  */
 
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Action.java
```diff
@@ -47,6 +51,7 @@ public void execute(@Nonnull CorfuRuntime runtime, int numRetry) {
                         getName(), x, e);
                 changeStatus(ActionStatus.ERROR);
                 runtime.invalidateLayout();
+                Sleep.sleepUninterruptibly(retryInterval);
```
zhangn49, line 54: Sleep class is deprecated. 

> @deprecated Sleep doesn't have any advantages over TimeUnit. Please always use TimeUnit.

PavelZaytsev: Fixed

zhangn49 (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/orchestrator/Action.java
```diff
@@ -47,12 +51,14 @@ public void execute(@Nonnull CorfuRuntime runtime, int numRetry) {
                         getName(), x, e);
                 changeStatus(ActionStatus.ERROR);
                 runtime.invalidateLayout();
+                TimeUnit.MILLISECONDS.sleep(retryInterval);
```
Maithem, line 54: Invalidate will also wait if a layout server cannot be reached. There is a retry parameter in the api the ends up calling this. Why is it necessary to have the sleep here? Shouldn't the calling API just increase its number of retries?

PR #2950, opened 2021-04-30
GetHighestSequenceNumber API filter HOLE entries

annym:
## Overview

Description: current getHighestSequenceNumber returns highest sequence number regardless of it being a DATA or HOLE entry. This patch allows this API to distinguish and retrieve highest sequence number of a valid DATA entry. 

Why should this be merged: client request

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
@@ -1,22 +1,29 @@
 package org.corfudb.runtime.collections;
 
+import com.google.common.collect.Iterables;
 import com.google.protobuf.Message;
 import lombok.Getter;
-import org.corfudb.common.metrics.micrometer.MeterRegistryProvider;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.protocols.wireprotocol.ILogData;
+import org.corfudb.protocols.wireprotocol.StreamAddressRange;
 import org.corfudb.protocols.wireprotocol.Token;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuStoreMetadata.TableName;
 import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
 import org.corfudb.runtime.Queue;
+import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.TableRegistry;
+import org.corfudb.runtime.view.stream.StreamAddressSpace;
+import org.corfudb.util.Utils;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.List;
-import java.util.Optional;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.Map;
+import java.util.NavigableSet;
+import java.util.UUID;
 
 /**
  * CorfuStore is a protobuf API layer that provides all the features of CorfuDB.
@@ -28,12 +35,13 @@ import java.util.concurrent.atomic.AtomicLong;
  * <p>
  * Created by zlokhandwala on 2019-08-02.
  */
+@Slf4j
 public class CorfuStore {
 
     @Getter
     private final CorfuRuntime runtime;
 
-    private final Optional<AtomicLong> openTableCounter;
+    private final CorfuStoreMetrics corfuStoreMetrics;
 
     /**
      * Creates a new CorfuStore.
@@ -55,9 +63,7 @@ public class CorfuStore {
     public CorfuStore(@Nonnull final CorfuRuntime runtime, boolean enableTxLogging) {
         runtime.setTransactionLogging(enableTxLogging);
         this.runtime = runtime;
-        openTableCounter = MeterRegistryProvider.getInstance()
-                .map(registry ->
-                        registry.gauge("open_tables.count", new AtomicLong(0L)));
+        this.corfuStoreMetrics = new CorfuStoreMetrics();
     }
 
     /**
@@ -104,7 +110,7 @@ public class CorfuStore {
         long startTime = System.nanoTime();
         Table table =
                 runtime.getTableRegistry().openTable(namespace, tableName, kClass, vClass, mClass, tableOptions);
-        openTableCounter.ifPresent(count -> count.getAndIncrement());
+        corfuStoreMetrics.recordTableCount();
         long elapsedTime = System.nanoTime() - startTime;
         table.getMetrics().recordTableOpenTime(elapsedTime);
         return table;
@@ -218,7 +224,11 @@ public class CorfuStore {
     }
 
     /**
-     * Return the address of the latest updated made in this table.
+     * Return the address of the latest update made in this table.
+     *
+     * Note: we can't deliberately return the tail of the stream map
+     * as this entry might be a HOLE, we need to filter and return only that
+     * corresponding to the last DATA entry.
      *
      * @param namespace - namespace that this table belongs to.
      * @param tableName - table name of this table without the namespace prefixed in.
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -218,19 +220,55 @@ public TxnContext txn(@Nonnull final String namespace, IsolationLevel isolationL
...
+        // Overall strategy:
+        // (1) Retrieve streams full address map from sequencer
+        // (2) Read entries in reverse to confirm it's a DATA type entry (for optimization
+        // read in batches). A batch size of 4 is guaranteed to suit one edge case case scenario: i.e.,
+        // consecutive checkpoints enforcing holes (3 rounds of CP accumulate before trim happens == 3 holes)
+
+        long startTime = System.nanoTime();
+        StreamAddressSpace streamAddressSpace = this.runtime.getSequencerView()
+                .getStreamAddressSpace(new StreamAddressRange(CorfuRuntime.getStreamID(
+                        TableRegistry.getFullyQualifiedTableName(namespace, tableName)), Address.MAX, Address.NON_ADDRESS));
```
hisundar, line 245: Just wondering if it would suffice to just the last getHighestSequenceNumberBatchSize number of entries from Address.MAX, and fetch more if the first batch is depleted, that way we avoid sending large bitmaps over?
(But since this is just the bitmap, perhaps its ok too)

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -218,19 +220,55 @@ public TxnContext txn(@Nonnull final String namespace, IsolationLevel isolationL
...
+        int numBatches = 0;
+
+        if (streamAddressSpace.size() != 0) {
+            NavigableSet<Long> addresses = streamAddressSpace.copyAddressesToSet(Address.MAX).descendingSet();
+            Iterable<List<Long>> batches = Iterables.partition(addresses,
+                    runtime.getParameters().getHighestSequenceNumberBatchSize());
+
+            for (List<Long> batch : batches) {
+                numBatches++;
+                Map<Long, ILogData> entries = runtime.getAddressSpaceView().read(batch);
```
hisundar, line 256: Maybe a comment or TODO to indicate that we can optimize this later to not transfer actual data since we are only really interested in testing for holes?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -218,19 +220,55 @@ public TxnContext txn(@Nonnull final String namespace, IsolationLevel isolationL
...
+                .getStreamAddressSpace(new StreamAddressRange(CorfuRuntime.getStreamID(
+                        TableRegistry.getFullyQualifiedTableName(namespace, tableName)), Address.MAX, Address.NON_ADDRESS));
+
+        int numBatches = 0;
+
+        if (streamAddressSpace.size() != 0) {
+            NavigableSet<Long> addresses = streamAddressSpace.copyAddressesToSet(Address.MAX).descendingSet();
+            Iterable<List<Long>> batches = Iterables.partition(addresses,
+                    runtime.getParameters().getHighestSequenceNumberBatchSize());
+
```
hisundar, line 253: Do you want to short circuit this for empty streams?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -218,19 +220,55 @@ public TxnContext txn(@Nonnull final String namespace, IsolationLevel isolationL
...
+                        return address;
+                    }
+                }
+            }
+        }
+
+        corfuStoreMetrics.recordBatchReads(numBatches);
+        corfuStoreMetrics.recordNumberReads(numBatches*runtime.getParameters().getHighestSequenceNumberBatchSize());
+        corfuStoreMetrics.recordHighestSequenceNumberDuration(System.nanoTime() - startTime);
+        return streamAddressSpace.getTrimMark();
```
hisundar, line 271: can we add a log line to indicate that we are returning the trim mark? Hopefully this should only happen for totally empty streams.

hisundar (commented):
Looks good mostly, just minor comments. thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
annym: It's a valid point, but I honestly doubt this would be necessary as we have never seen these bitmaps being a problem on transfer (and on restart they are completely transferred from log unit to sequencer), also if checkpointing is happening bitmaps are trimmed regularly.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
annym: what do you mean? this is only if streamAddressSpace.size() != 0 (not empty) or are you referring to something different?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
annym: sure, I'm adding this at DEBUG level as I'm guessing this is only for debugging purposes. Otherwise, not sure if for the case of search it could end up being called too many times, cause I'm guessing that on each read they call this API. If you think it should be a different level let me know...

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: I guess batches would be 1 if the number of addresses are less than the batch size.

hisundar (approved):
LGTM

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStoreMetrics.java
```diff
@@ -0,0 +1,72 @@
...
+import io.micrometer.core.instrument.MeterRegistry;
+import org.corfudb.common.metrics.micrometer.MeterRegistryProvider;
+
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CorfuStoreMetrics {
+    // Keep track of the number of read entries to compute a stream's highestSequenceNumber
+    private final Optional<AtomicInteger> numberBatchReads;
+    private final Optional<AtomicInteger> numberReads;
```
zhangn49, line 14: The type of these two metrics is Integer, which means they will be reset every time when getHighestSequenceNumber gets called. They won't have statistical data, is this what we want?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -218,19 +224,61 @@ public TxnContext txn(@Nonnull final String namespace, IsolationLevel isolationL
...
+        if (streamAddressSpace.size() != 0) {
+            NavigableSet<Long> addresses = streamAddressSpace.copyAddressesToSet(Address.MAX).descendingSet();
+            Iterable<List<Long>> batches = Iterables.partition(addresses,
+                    runtime.getParameters().getHighestSequenceNumberBatchSize());
+
+            for (List<Long> batch : batches) {
+                numBatches++;
+                // TODO: We can optimize this such that actual data is not transferred back to the client,
+                //  we require an API that inspects the data at the remote and indicates whether its an actual DATA
+                //  entry or a HOLE, e.g.: Map<Long, DataType> entryTypes = runtime.getAddressSpaceView().getDataType(batch)
```
zhangn49, line 263: Is it possible that we have an API to get stream's non-HOLE tail sequence number? It will have only one RPC

zhangn49 (approved):
LGTM, one question about metrics.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: It is, we discussed the same, just want to do this in phases, first phase we do the basic client only fix. Next phase either an rpc that returns the first non-hole address or an rpc that does not transfer data.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStoreMetrics.java
annym: I'm not on expert on this framework. But for these two metrics I actually used a gauge instead of a counter as according to the documentation:
"While a counter must always go up, gauges have no such constraint: a gauge outputs a value, unrelated from one call to the next." which is the case of these variable which are not increasing but are independent one from the other. Still I think this will give statistical data or are you sure it won't?

zhangn49: I am pretty sure it won't give statistical data. 

> Gauge. A gauge is a metric that represents a single numerical value that can arbitrarily go up and down. Gauges are typically used for measured values like temperatures or current memory usage

Previously we use gauge to capture cache hit ratio or miss ratio, which is a single value when you access it. In this case, I think you can use histogram

annym: Yep you're correct, I looked into the metrics we collect and effectively there is no statistical data on these. In that case I think I'll use a DistributionSummary.

PR #2951, opened 2021-05-05
Streaming API Trim Exceptions Bug Fix - 2

annym:
## Overview

Description:
Fix an incorrect TrimmedException which derives from the case of 'tagged' streams which do not receive updates over a long period of time. This leads to a state in which the last synced point (globalPointer) will fall below the log trim mark. In this case, we should not throw a TrimmedException as it is up to date and the pointer has not moved because there are no actual updates. 

- Also move LR in the direction of its own dedicated transaction stream, as this will disappear soon. 

Why should this be merged: bug fix

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
@@ -27,6 +27,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import static org.corfudb.runtime.view.ObjectsView.LOG_REPLICATOR_STREAM_ID;
 import static org.corfudb.runtime.view.ObjectsView.TRANSACTION_STREAM_ID;
 
 @Slf4j
@@ -169,11 +170,11 @@ public class LogReplicationAckReader {
     }
 
     private long getTxStreamTail(Map<UUID, Long> tailMap) {
-        if (tailMap.containsKey(TRANSACTION_STREAM_ID)) {
-            return tailMap.get(TRANSACTION_STREAM_ID);
+        if (tailMap.containsKey(LOG_REPLICATOR_STREAM_ID)) {
+            return tailMap.get(LOG_REPLICATOR_STREAM_ID);
         }
 
-        log.warn("Tx Stream tail not present in sequencer, id={}", TRANSACTION_STREAM_ID);
+        log.warn("Tx Stream tail not present in sequencer, id={}", LOG_REPLICATOR_STREAM_ID);
         return Address.NON_ADDRESS;
     }
 
@@ -321,7 +322,7 @@ public class LogReplicationAckReader {
         long totalEntries = 0;
 
         if (upperBoundary > lowerBoundary) {
-            StreamAddressRange range = new StreamAddressRange(TRANSACTION_STREAM_ID, upperBoundary, lowerBoundary);
+            StreamAddressRange range = new StreamAddressRange(LOG_REPLICATOR_STREAM_ID, upperBoundary, lowerBoundary);
             StreamAddressSpace txStreamAddressSpace = runtime.getSequencerView().getStreamAddressSpace(range);
             // Count how many entries are present in the Tx Stream (this can include holes,
             // valid entries and invalid entries), but we count them all (equal weight).
--- a/runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
@@ -38,7 +38,13 @@ public class ObjectsView extends AbstractView {
      * The Transaction stream is used to log/write successful transactions from different clients.
      * Transaction data and meta data can be obtained by reading this stream.
      */
+    @Deprecated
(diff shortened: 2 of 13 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/LogReplicationAckReader.java
```diff
@@ -169,11 +170,11 @@ private long calculateRemainingEntriesToSend(long ackedTimestamp) {
     }
 
     private long getTxStreamTail(Map<UUID, Long> tailMap) {
-        if (tailMap.containsKey(TRANSACTION_STREAM_ID)) {
-            return tailMap.get(TRANSACTION_STREAM_ID);
+        if (tailMap.containsKey(LOG_REPLICATOR_STREAM_ID)) {
```
xnull, line 173: Simpler version:
`return tailMap.getOrDefault(LOG_REPLICATOR_STREAM_ID, Address.NON_ADDRESS)`

runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
```diff
@@ -38,8 +38,13 @@
      * The Transaction stream is used to log/write successful transactions from different clients.
      * Transaction data and meta data can be obtained by reading this stream.
      */
+    @Deprecated
     public static final UUID TRANSACTION_STREAM_ID = CorfuRuntime.getStreamID("Transaction_Stream");
 
+    // We are temporarily naming this stream with the same name as TRANSACTION_STREAM_ID to avoid breaking LR code
+    // while transition to UFO is completed (once migration is complete to UFO this stream can be renamed)
+    public static final UUID LOG_REPLICATOR_STREAM_ID = CorfuRuntime.getStreamID("Transaction_Stream");
```
xnull, line 46: I think the probability that one day someone creates a transaction_stream table is pretty high, let's add some prefix, for instance: `org.corfudb.logreplication.transactionstream`

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -381,8 +381,8 @@ private boolean isTrimCoveredByCheckpoint(long trimMark) {
      * @return whether this stream is capable of being checkpointed
      */
     private boolean isCheckpointCapable() {
-        return !getId().equals(ObjectsView.TRANSACTION_STREAM_ID)
-                && getStreamOptions().isCheckpointCapable();
+        // Subscription tag streams are not checkpoint(able) (used for streaming) this is set on the StreamSubscription
+        return getStreamOptions().isCheckpointCapable();
```
xnull, line 385: isCheckpointCapable of what? it would be great to rename to something better

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
```diff
@@ -400,7 +400,7 @@ protected long getMaxGlobalFromMaxEntries(int maxEntries) {
             // but we need to consider the case when the 'syncUpTo' address falls below the trim mark, so we correctly
             // report the TrimmedException allowing consumers to know that data is no longer available from this point.
             // We could throw the TrimmedException directly or in this case let remainingUpTo discover it for us.
-            if (getCurrentGlobalPosition() <= streamAddressSpace.getTrimMark()) {
+            if (getCurrentGlobalPosition() < streamAddressSpace.getTrimMark()) {
```
zhangn49, line 403: Tried to understand the equal case but failed. Could you please explain a little more here.

annym: Sure, so the equal case is only valid on initial subscription, i.e., the exact moment when you do .subscribeListener(ts); For example, if someone requests to stream from the exact address that corresponds to the stream's trim mark, you are sure that address is no longer available and you can throw a TrimmedException (that is why the validateSyncPoint was added right on the subscribe method). 

But if you were allowed to subscribe and polling tasks are issued every Xms, you could reach a valid point where currentGlobalPosition = streamTrimMark, e.g., you are up to address = 10 (nothing has been trimmed)... you keep polling every Xms but no new updates come in (this happens for long time), then the log is trimmed (let's say at address 200), so now the streamTrimMark = 10... in this case they're equal but you didn't miss it, you had already processed 10 (but there were no further updates). 

For this reason we separated the case where you need the "=" on initial subscription (and moved that logic as a pre-validation as soon as you subscribe) and here removed it, as the pointer being at that position means it has already been resolved.

runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
annym: I get your suggestion and we can definitely do it on a second round, I've added it to the TODO, the reason why I don't want to do it now is because this most likely is used by DAO and if we change it here, any code still not migrated to UFO will break.

runtime/src/main/java/org/corfudb/runtime/view/stream/AddressMapStreamView.java
annym: if the stream is checkpoint capable. Renaming this would break NSX code as this is used by the DCN layer. Once DCNs are removed we can rename as this will only be used within Corfu code.

zhangn49 (approved):

hisundar (approved):
Looks good to me, thanks

PR #2958, opened 2021-05-13
Reduce metrics noise + fix some metrics reporting  bugs

PavelZaytsev:
## Overview

* Remove 95 percentile from logs.
* Don't log zero values.
* Fix some timing bugs.

### DIFF
--- a/common/src/main/java/org/corfudb/common/metrics/micrometer/registries/LoggingMeterRegistryWithHistogramSupport.java
+++ b/common/src/main/java/org/corfudb/common/metrics/micrometer/registries/LoggingMeterRegistryWithHistogramSupport.java
@@ -72,7 +72,7 @@ public class LoggingMeterRegistryWithHistogramSupport extends StepMeterRegistry 
     }
 
     Stream<String> writeGauge(Meter.Id id, Double value) {
-        if (Double.isFinite(value) && value != 0) {
+        if (Double.isFinite(value)) {
             return Stream.of(influxLineProtocol(id, "gauge", Stream.of(new Field("value", value))));
         }
         return Stream.empty();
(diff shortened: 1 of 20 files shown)


Maithem (approved):

tomato-coding (approved):
looks good

common/src/main/java/org/corfudb/common/metrics/micrometer/registries/LoggingMeterRegistryWithHistogramSupport.java
```diff
@@ -65,22 +65,22 @@ public LoggingMeterRegistryWithHistogramSupport(LoggingRegistryConfig config,
...
-        if (Double.isFinite(count)) {
+        if (Double.isFinite(count) && count != 0) {
             return Stream.of(influxLineProtocol(id, "counter", Stream.of(new Field("value", count))));
         }
         return Stream.empty();
     }
 
     Stream<String> writeGauge(Meter.Id id, Double value) {
-        if (Double.isFinite(value)) {
+        if (Double.isFinite(value) && value != 0) {
```
hisundar, line 75: Guage can go up and down right, so is this ok for printing only non-zero value?

hisundar (approved):
Just a minor question about Guages, but otherwise this change looks good to me, thanks

common/src/main/java/org/corfudb/common/metrics/micrometer/registries/LoggingMeterRegistryWithHistogramSupport.java
PavelZaytsev: I agree, unlike every other meter (timer, distribution), gauges report current values and not rates. 0 is a valid current value to be reported.

Issue #2959, opened 2021-05-17
Wild NTP fluctuations degrades CorfuQueue performance


hisundar:
If NTP moves ahead by too much, the corrections cause a lot of transaction retries which results in degraded performance in parallel operations of the CorfuQueue.

PR #2960, opened 2021-05-17
CorfuGuid: Retry if NTP is moving time ahead by crazy amounts

hisundar:
## Overview

Description:

Why should this be merged: 
Some systems the NTP is so misconfigured that time moves ahead by huge leaps only to get recorrected.
This causes some instability in the guid generator which relies on a sane NTP for performance.

Related issue(s) (if applicable): #2959 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
@@ -18,6 +18,7 @@ import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Globally Unique Identity generator that returns ids
@@ -233,7 +234,17 @@ public class CorfuGuidGenerator implements OrderedGuidGenerator {
      */
     private synchronized CorfuGuid getCorrectedTimestamp() {
         long currentTimestamp = System.currentTimeMillis();
-        if (currentTimestamp == previousTimestamp) {
+        if (currentTimestamp > previousTimestamp) {
+            final long oneYearInMillis = TimeUnit.DAYS.toMillis(365);
+            int maxCrazyNtpValueRetries = 1000;
+            while (currentTimestamp - previousTimestamp > oneYearInMillis) {
+                currentTimestamp = System.currentTimeMillis();
+                if (maxCrazyNtpValueRetries-- <= 0) {
+                    log.warn("Time moved forward by too much, NTP misbehaving?");
+                    break;
+                }
+            }
+        } else if (currentTimestamp == previousTimestamp) { // No change in wall clock time
             resolutionCorrection++;
             // If more than 16 concurrent threads are calling this api, we can hit this.
             // Simple solution is to make the last thread wait a few microseconds.
@@ -244,12 +255,9 @@ public class CorfuGuidGenerator implements OrderedGuidGenerator {
                 }
                 resolutionCorrection = 0;
             }
-        } else {
-            resolutionCorrection = 0;
-        }
-
-        if (currentTimestamp < previousTimestamp) {
+        } else if (currentTimestamp < previousTimestamp) { // NTP corrects time backward
             driftCorrection++;
+            resolutionCorrection = 0;
             if (driftCorrection > CorfuGuid.MAX_CORRECTION) {
                 log.warn("updateInstanceId: Time went backward too many times "+
                                 " timestamp={} previousTimestamp={} correction={} instanceId={}",


runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -244,12 +255,9 @@ private synchronized CorfuGuid getCorrectedTimestamp() {
                 }
                 resolutionCorrection = 0;
             }
-        } else {
-            resolutionCorrection = 0;
-        }
-
-        if (currentTimestamp < previousTimestamp) {
+        } else if (currentTimestamp < previousTimestamp) { // NTP corrects time backward
```
annym, line 258: why do you need the condition? we already checked >, and =...

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
```diff
@@ -233,7 +234,17 @@ public long nextLong() {
      */
     private synchronized CorfuGuid getCorrectedTimestamp() {
         long currentTimestamp = System.currentTimeMillis();
-        if (currentTimestamp == previousTimestamp) {
+        if (currentTimestamp > previousTimestamp) {
+            final long oneYearInMillis = TimeUnit.DAYS.toMillis(365);
+            int maxCrazyNtpValueRetries = 1000;
+            while (currentTimestamp - previousTimestamp > oneYearInMillis) {
```
annym, line 240: should this sleep in between retries? how do we know this is "enough" cycles/time to typically correct itself?

hisundar: just experimental observations at this point - when ntp throws time around, it ends up correcting itself in subsequent calls. I just thought adding a sleep might slow down a problem that corrects itself quickly?

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: they are different right - currentTime > previousTimestamp means normal time moving forward, we only need to check for crazy jumps,
currentTime == previousTime means this call was invoked within the same millisecond resolution
But if currentTime < previousTime means time has moved backward and monotonic growth is broken so we fix up by adding drift correction.

annym: correct, but if it is not greater and it is not equal, it necessarily is lower, so an else would be enough, right? (without the if)

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
annym: ok. I was just wondering based on what we knew this was guaranteed to be enough time to recover... and one more question, what are the implications of this happening? is a warning enough?

Maithem: This loop will most likely execute way faster than any clock correction. I don't see how it addresses the time jump issue.

Maithem (commented):
Is this GUID generator still necessary for the queue?

Since the queue relies on the sequencer clock to generate ids can this generator be removed?
With the write pre-hook (using the sequencer result to populate the write payload) I don't see the importance of this GUID generator. The sequencer clock is already monotonic.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: oh, got it, sure, I can remove the redundant if

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
hisundar: logs had showed that in one call to currentTimeMillis() the value was something like INT_MAX and in the next call it had returned back to normal. I don't understand NTP to know why that happens, I can add a sleep if you think that helps.

hisundar:
> Is this GUID generator still necessary for the queue?
> 
> Since the queue relies on the sequencer clock to generate ids can this generator be removed?
> With the write pre-hook (using the sequencer result to populate the write payload) I don't see the importance of this GUID generator. The sequencer clock is already monotonic.

GUID is still needed for the key, The pre-hook value is not used as key, it is used in the metadata section for storing the order.

runtime/src/main/java/org/corfudb/runtime/view/CorfuGuidGenerator.java
Maithem: I don't feel like spinning on currentTimeMillis is robust. Why isn't it sufficient to just retry the whole transaction from a higher layer? All transactions are susceptible to non-conflict related aborts and need to retry anyways.

hisundar: we don't have control over that aspect right - I agree in general with the fact that they are supposed to retry the transaction, but according to them this part of the code should never throw any exception

hisundar:
This isn't the right fix for the NTP issue, abandoning after discussion.

PR #2902, opened 2021-03-05
Publishing corfu docker images

xnull:
## Overview

Description:
Publishing Corfu docker images

Why should this be merged: 

Related issue(s) (if applicable): #<number>

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #2962, opened 2021-05-19
Version Compatibility Test Automation - Phase 1

xcchang:
## Overview

Description: 
Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -94,3 +94,49 @@ jobs:
 
       - name: Run universe tests
         run: ./mvnw -pl :universe verify -Puniverse -Dmaven.javadoc.skip=true
+
+  compatibility_test:
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on
+    runs-on: ubuntu-18.04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.6.0
+        with:
+          access_token: ${{ github.token }}
+
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - uses: satackey/action-docker-layer-caching@v0.0.8
+        continue-on-error: true
+
+      - name: Prepare top of trunk docker image
+        run: ./mvnw clean install -DskipTests -Pcompatibility -Dmaven.javadoc.skip=true -T 1C
+
+      - name: Pull docker image built from master parent commit
+        run: docker pull corfudb/corfu-server:0.3.1-SNAPSHOT
+
+      - uses: actions/checkout@v2
+        with:
+          repository: 'CorfuDB/corfudb-cloud'
+
+      - name: Build universe-core
+        working-directory: ./universe
+        run: rm -r ~/.m2/repository/org/corfudb && ./gradlew clean publishToMavenLocal
+
+      - name: Run version compatibility tests
+        working-directory: ./tests
+        run: ./gradlew clean test -Dtags=version-compatibility
--- a/.mvn/settings.xml
+++ b/.mvn/settings.xml
@@ -1,16 +1,16 @@
-<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
-          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
+<?xml version="1.0" encoding="UTF-8"?>
+<settings xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0 http://maven.apache.org/xsd/settings-1.1.0.xsd" xmlns="http://maven.apache.org/SETTINGS/1.1.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <servers>
         <server>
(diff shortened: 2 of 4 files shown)


hisundar (approved):
Please feel free to merge if this configuration works

.mvn/settings.xml
```diff
@@ -1,16 +1,16 @@
...
-          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
+<?xml version="1.0" encoding="UTF-8"?>
+<settings xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0 http://maven.apache.org/xsd/settings-1.1.0.xsd" xmlns="http://maven.apache.org/SETTINGS/1.1.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
     <servers>
         <server>
-            <id>bintray-corfudb-maven</id>
-            <username>corfudb</username>
-            <password>${secrets.jfrog_oss_password}</password>
+            <username>xchang@vmware.com</username>
```
xnull, line 6: please use jfrog_oss_password as it was in the previous version, please just update user/password in github

pom.xml
```diff
@@ -196,16 +196,16 @@
...
-            <id>corfudb-snapshots</id>
-            <name>oss-jfrog-artifactory-snapshots</name>
-            <url>https://oss.jfrog.org/artifactory/oss-snapshot-local</url>
-        </snapshotRepository>
         <repository>
-            <id>bintray-corfudb-maven</id>
-            <name>CorfuDB repository</name>
-            <url>https://api.bintray.com/maven/corfudb/maven/org.corfudb/;publish=1;override=1</url>
+            <id>central</id>
+            <name>a0w7ezmoyzis4-artifactory-primary-0-releases</name>
```
xnull, line 201: please give it a nice name

.mvn/settings.xml
xnull: I mean we don't need to share outside, so far. If it would be needed we can do that in the future

xcchang: Yeah I will request for putting this password in our repo's secrets, and this password is already encrypted by the JFrog though, it's not the real password :)

xnull (approved):
Looks great

.github/workflows/pull_request.yml
```diff
@@ -94,3 +94,49 @@ jobs:
...
+          java-version: 1.8
+
+      - uses: satackey/action-docker-layer-caching@v0.0.8
+        continue-on-error: true
+
+      - name: Prepare top of trunk docker image
+        run: ./mvnw clean install -DskipTests -Pdocker -Dmaven.javadoc.skip=true -T 1C
+
+      - name: Pull base docker image for cross version env
+        run: docker pull corfudb/corfu-server:BASE
```
xnull, line 130: we need to double-check that 'BASE' can be used, for now you need to deploy docker container locally and the image name will be `corfudb/corfu-server:${corfu_version}`

xcchang: tag BASE now is built temporarily for the branch, under the `compatibility` profile in pom, which represents the top of trunk Corfu server version. tag `0.3.1-SNAPSHOT` is pulled down from dockerhub, which represents parent version and will be updated automatically after https://github.com/CorfuDB/CorfuDB/pull/2902 is merged. Also, by doing this we keep the `0.3.1-SNAPSHOT` consistently refers to parent version in corfudb-cloud repo in version compatibility use case.

.github/workflows/pull_request.yml
```diff
@@ -94,3 +94,49 @@ jobs:
...
+          java-version: 1.8
+
+      - uses: satackey/action-docker-layer-caching@v0.0.8
+        continue-on-error: true
+
+      - name: Prepare top of trunk docker image
+        run: ./mvnw clean install -DskipTests -Pcompatibility -Dmaven.javadoc.skip=true -T 1C
+
+      - name: Pull docker image built from master parent commit
+        run: docker pull corfudb/corfu-server:0.3.1-SNAPSHOT
```
xnull, line 130: It seems to me you don't need to do docker pull, universe doing it actomatically

.github/workflows/pull_request.yml
```diff
@@ -94,3 +94,49 @@ jobs:
...
+
+      - uses: actions/checkout@v2
+        with:
+          repository: 'CorfuDB/corfudb-cloud'
+
+      - name: Build universe-core
+        working-directory: ./universe
+        run: rm -r ~/.m2/repository/org/corfudb && ./gradlew clean publishToMavenLocal
```
xnull, line 138: in this case you delete all the artifacts you built, in a previous step

.github/workflows/pull_request.yml
xcchang: yeah -Pdocker will build a docker image  with `0.3.1-SNAPSHOT` tagged, but that will be the top of trunk version. Here we want `0.3.1-SNAPSHOT` to be the version which is already hosted in docker hub, i.e. the parent version.
But yeah this is a little confusing, we may come up with better versioning strategy later..

.github/workflows/pull_request.yml
xcchang: it is supposed to do so, as the previous install step will build latest artifacts to maven local, we want the corfudb-cloud to get them from JFrog

xnull: yeah, but you remove everything in corfu dir when you need only remove universe-core and universe-infrastructure

xcchang: I think the `universe-core` and `universe-infrastructure` are built here. The command is trying to remove runtime, infrastructure, etc., which are supposed to get from jfrog repo

.github/workflows/pull_request.yml
xnull: as discussed offline, we will use 'latest' tag

.github/workflows/pull_request.yml
xnull: oh ok, got it

infrastructure/pom.xml
```diff
@@ -90,6 +90,43 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>compatibility</id>
```
xnull, line 94: we could just add a tag into docker profile, no need for another profile. If you'd like to merge these two profiles then please create another pr

PR #2963, opened 2021-05-19
Introduce Compatibility Vector for RPCs

chetangudisagar:
## Overview

Description: We introduce the compatibility vectors for communicating the information about features that are supported by the sender and receiver. We represent each feature as an enum value in a Java file and set its corresponding enum number's index to 1 in the BitSet and store it in Protobuf's ByteString format. The BitSet values will be the same across both the client and the server of a particular version.

Why should this be merged: Enables features compatibility checks for rolling upgrades.

- [x] Public API has Javadoc

### DIFF
--- a/common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
+++ b/common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
@@ -1,0 +1,130 @@
+package org.corfudb.common.util;
+
+import com.google.protobuf.ByteString;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Arrays;
+import java.util.BitSet;
+
+/**
+ * A collection of {@link BitSet}s representing the Corfu features enabled so far.
+ * <p>
+ * Creates the Compatibility Vectors that contain the information about the Corfu features enabled
+ * so far as per the current stage of the code during the compile time of the class.
+ * The vectors are of type {@link ByteString} in which each bit will represent if the corresponding
+ * feature is enabled or not based on its value (0 or 1).
+ * <p>
+ * Client-Server Transmission:
+ * The BitSet values are already converted to the ByteString format and are accessed through
+ * {@link #getCompatibilityVectors()} method. These will be set in the Protobuf header
+ * HeaderMsg.ProtocolVersionMsg value (of both RequestMsg and ResponseMsg),
+ * and sent to the receiver. The receiver will then check whether the specific features are set or
+ * not using {@link CompatibilityVectorUtils#isFeatureEnabled(Feature, ByteString)} method.
+ * <p>
+ * For adding new Features:
+ * - <b>Introduce</b> them by creating an entry in this class's {@link Feature} enum structure.
+ * <p>
+ * Created by cgudisagar on 5/3/21.
+ */
+@Slf4j
+public class CompatibilityVectorUtils {
+
+    // Prevent class from being instantiated
+    private CompatibilityVectorUtils() {
+    }
+
+    /**
+     * @return a {@link ByteString} of vectors representing the Corfu features enabled so far.
+     */
+    public static ByteString getCompatibilityVectors() {
+        return Feature.vectors;
+    }
+
+    /**
+     * Checks if the feature is enabled in the given byte array or not.
+     *
+     * @param feature the {@link Feature} enum value to be checked
+     * @param vectors the {@link ByteString} that contains the bits of the Corfu features
+     * @return a boolean value that denotes if the feature is enabled or not.
+     */
+    @SuppressWarnings("unused") // it will be used in future
+    public static boolean isFeatureEnabled(Feature feature, ByteString vectors) {
+        int featureNumber = feature.getNumber();
+
+        // If the feature number (0-indexed) is greater than or equal the bytes length,
+        // or if it is a negative value, then the feature is not enabled.
+        if (featureNumber >= 8 * vectors.size() || featureNumber < 0)
+            return false;
+
+        // Check if the bit is set in the vector
+        // = ((corresponding byte) AND (1 only on the index of the bit) !=0)
+        // & 0xff to prevent a possible int promotion error while using bytes with
+        // shifts and bitwise operators
+        return (((vectors.byteAt((featureNumber) / 8) & 0xff) & (1 << ((featureNumber) % 8))) != 0);
+    }
+
+    /**
+     * This enum loads when {@link #getCompatibilityVectors()} is called
+     * for the first time. It gives a thread-safe lazy-initialization because the class loader
+     * guarantees that all static initialization is complete before getting access to the class.
+     */
+    public enum Feature {
+        // Comma separated enum values indicating each feature along with their sequence number.
+        // FEATURE_A(0),
+        ;
+
+        /**
+         * Contains the vectors representing the Corfu features enabled so far.
+         */
+        private static final ByteString vectors = generateVectors();
+
+        private final int number;
+
+        /**
+         * A constructor to initiate the enum value along with its sequence number
+         *
+         * @param number the sequence number of the enum
+         */
+        @SuppressWarnings("unused")
+        // it will be used in future
+        Feature(int number) {
+            if (number < 0)
+                throw new IllegalArgumentException("Enum number should be a non negative integer." +
+                        " The current value is " + number + ".");
+            this.number = number;
+        }
+
+        private static ByteString generateVectors() {
+            Feature[] features = Feature.values();
+
+            // Though initializing with the length is optional, it improves performance.
+            BitSet bitSet = new BitSet(features.length);
+
+            for (Feature feature : features) {
+                bitSet.set(feature.getNumber());
+            }
+
+            if (features.length > 0) {
+                log.info("generateVectors: Enabled features are {}", Arrays.toString(features));
+                log.info("generateVectors: Compatibility vectors are {}", bitSet);
+            }
+
+
+            return ByteString.copyFrom(bitSet.toByteArray());
+        }
+
+        /**
+         * Get the sequence number of the enum entry
+         *
+         * @return the sequence number of the enum entry
+         */
+        private int getNumber() {
+            return this.number;
+        }
+
+        @Override
+        public String toString() {
+            return this.number + ": " + super.toString();
+        }
+    }
+}
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
@@ -44,6 +44,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 
 import static org.corfudb.protocols.CorfuProtocolCommon.getUuidMsg;
+import static org.corfudb.protocols.service.CorfuProtocolMessage.getDefaultProtocolVersionMsg;
 import static org.corfudb.protocols.service.CorfuProtocolMessage.getRequestMsg;
 
 /**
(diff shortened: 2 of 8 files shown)


common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
```diff
@@ -0,0 +1,118 @@
...
+        /**
+         * Contains the vectors representing the Corfu features enabled so far.
+         */
+        private static final BitSet vectors = generateVectors();
+
+        private static BitSet generateVectors(){
+            // length-1 to get rid of the default UNRECOGNIZED enum value, else we get an error
+            BitSet bitSet = new BitSet(Feature.values().length-1);
+            Feature[] features = Feature.values();
+            for(int feature_index=1; feature_index < features.length-1; ++feature_index){
```
hisundar, line 50: Should this be 
for (int feature_index=1; feature_index < features.length; ++feature_index) {
Since you already have a < check should it less than length instead of length - 1 ?

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
```diff
@@ -0,0 +1,118 @@
...
+                if(features[feature_index]!= Feature.UNRECOGNIZED ||
```
hisundar, line 51: Where is Feature.UNRECOGNIZED defined? I only see one enum that is INVALID in the Feature enum. Am I looking at a wrong commit?

hisundar, line 51: Minor nit: I think some convention requires adding space after primitive language constructs like if while for etc.

common/proto/corfu_features.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package org.corfudb.common.util.proto;
+
+enum Feature {
+  FEATURE_INVALID = 0;
```
hisundar, line 6: Can we not have an INVALID that way each feature just represents a bit position starting with index 0?
First feature = 0 - > takes first bit position at index 0
Second feature = 1 -> takes second bit at index 1

hisundar (changes requested):
Hi Chetan, please let me know if my comments make sense?

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: Thank you, Done.

common/proto/corfu_features.proto
chetangudisagar: I was following the practice - "Always Start Enums with INVALID = 0;" for backward compatibility. I will send you the links for further information on this.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: > The protocol buffer compiler will generate a Java enum type called Foo with the same set of values. If you are using proto3, it also adds the special value UNRECOGNIZED to the enum type.

[Docs Link](https://developers.google.com/protocol-buffers/docs/reference/java-generated#enum)

zfrenette: It seems like UNRECOGNIZED and INVALID are serving the same purpose?

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: Sorry for the confusion, Hope this helps -

For a given enum - 
```
enum Feature {
  FEATURE_INVALID = 0;
  TEST = 1;
  ABC = 2;
}
```

Feature.values() returns `["FEATURE_INVALID", "TEST", "ABC", "UNRECOGNIZED"]` (a list of enum values, not exactly strings)
Note: `length = 4`

We want the result bitset to contain the first three elements i.e. length-1 elements.

The `for` loop **should** start from index 1, else it will set the INVALID feature to 1. We want it to be 0 as it's an INVALID feature by convention.

Further, the loop can end at index 2 (`< length -1`), otherwise, it will try to access the UNRECOGNIZED feature at index 3, which gives an error at runtime. I had added the `if` checks for safety. I will remove them.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: We had an offline discussion and changed the enum implementations from Protobuf to Java.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+            // Though initializing with the length is optional, it improves performance.
+            BitSet bitSet = new BitSet(features.length);
+
+            for (Feature feature : features) {
+                bitSet.set(feature.getNumber());
+            }
+
+            log.info("generateVectors: Compatibility vectors are successfully initiated with {}",
+                    bitSet);
```
xcchang, line 119: Since this log will only be done once for a node, would it be better to log out the features also? such that we could have a comparison

common/src/test/java/org/corfudb/common/util/CompatibilityVectorUtilsTest.java
```diff
@@ -0,0 +1,50 @@
+package org.corfudb.common.util;
+
+import com.google.protobuf.ByteString;
+import org.corfudb.common.util.CompatibilityVectorUtils.Feature;
+import org.junit.Ignore;
+import org.junit.Test;
```
xcchang, line 6: Import of Ignore could be removed I think

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
```diff
@@ -0,0 +1,149 @@
...
+ * not using {@link CompatibilityVectorUtils#isFeatureEnabled(int, ByteString)} method.
+ * <p>
+ * For adding new Features:
+ * - <b>Introduce</b> them by creating an entry in this class's {@link Feature} enum structure.
+ * <p>
+ * Created by cgudisagar on 5/3/21.
+ */
+@Slf4j
+public class CompatibilityVectorUtils {
+
```
xcchang, line 34: Should we add stubs to the places where Compatibility Vector will be used? like have a log of "No feature needs to be specially handled" or something like that. I think we want to be clear about where it will checked at this time.

xcchang (commented):
Looks nice and left some minor comments. By the way, please wrap the extended git commit message into multiple lines when rebase :)

common/src/test/java/org/corfudb/common/util/CompatibilityVectorUtilsTest.java
chetangudisagar: Done. Thank you.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: For now, these bitset vectors will be used only in the Handler methods on the server-side and the client-side as they are specific to the features. The only other place was in the handshake handler part but we already have version checks there.
So, in my opinion, the checks can be placed once we have the first feature from the next release.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: Will do, thanks!

zfrenette:
Does version information need to be added to the `LogReplicationClientRouter`'s `sendRequestAndGetCompletable` method as well: https://github.com/CorfuDB/CorfuDB/blob/0aa9eb5f886448ac4a282be3209b598f70f4c0ad/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java#L181

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        /**
+         * Get the sequence number of the enum entry
+         *
+         * @return the sequence number of the enum entry
+         */
+        public int getNumber() {
+            return number;
+        }
```
zfrenette, line 152: Can this be made private? Or does the numerical representation need to be known outside of this class?

common/src/test/java/org/corfudb/common/util/CompatibilityVectorUtilsTest.java
```diff
@@ -0,0 +1,49 @@
...
+    private final ByteString vectors = CompatibilityVectorUtils.getCompatibilityVectors();
+
+    /**
+     * Tests {@link Feature#isFeatureEnabled(int, ByteString)} with invalid parameters.
+     */
+    @Test
+    public void testFeatureInvalidValues() {
+        // Assert that negative feature's bit returns False as features start from 0 onwards.
+        assertThat(CompatibilityVectorUtils.isFeatureEnabled(-1, vectors)
+        ).isFalse();
```
zfrenette, line 26: nit: Can you address the formatting here?

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Checks if the feature is enabled in the given {@link ByteString} or not.
+     *
+     * @param feature int value of the feature to be checked
+     * @param vectors the {@link ByteString} that contains the bits of the Corfu features
+     * @return a boolean value that denotes if the feature is enabled or not.
+     */
+    public static boolean isFeatureEnabled(int feature, ByteString vectors) {
```
zfrenette, line 63: Here as well.

zfrenette (approved):
Looks good. Just some minor comments/questions.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: Makes sense. Thank you.

common/src/test/java/org/corfudb/common/util/CompatibilityVectorUtilsTest.java
chetangudisagar: Done, Thank you.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
chetangudisagar: Done, Thank you.

common/src/main/java/org/corfudb/common/util/CompatibilityVectorUtils.java
```diff
@@ -0,0 +1,127 @@
...
+            Feature[] features = Feature.values();
+
+            // Though initializing with the length is optional, it improves performance.
+            BitSet bitSet = new BitSet(features.length);
+
+            for (Feature feature : features) {
+                bitSet.set(feature.getNumber());
+            }
+
+            log.info("generateVectors: Enabled features are {}", Arrays.toString(features));
```
hisundar, line 106: Can this log line be inside the bitSet.set() the reason is if there are no features enabled then we don't want to log anything?

hisundar (approved):
LGTM, minor comment about logging.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/runtime/LogReplicationClientRouter.java
```diff
@@ -179,6 +180,7 @@ public IClientRouter addClient(IClient client) {
             @Nonnull String nodeId) {
 
         HeaderMsg.Builder header = HeaderMsg.newBuilder()
+                .setVersion(getDefaultProtocolVersionMsg())
```
zfrenette, line 183: Is there anything server-side that needs to be adjusted similarly?

chetangudisagar: No, I couldn't find it in any other place. In the server, we merge from the client's header and set the server version explicitly.

Issue #2965, opened 2021-05-20
CorfuStore's TableRegistry duplicates several table descriptors


hisundar:
TableRegistry needs to hold on to every table's protobuf file descriptors so that it checkpointing and browsing can work.
But if the same set of protobufs are included in a lot of tables then this results in a lot of duplication.
This can also overwhelm a JVM which has small memory footprint.

hisundar:
fix is merged

PR #2966, opened 2021-05-20
CorfuStore: Optimize TableRegistry by De-duplicating Descriptors

hisundar:
## Overview

Description:

Before this change, every TableRegistry entry carried it's own list of protobuf file descriptors which wastes memory.
De-duplicate all protobuf file descriptors into a single table instead so each protobuf file is store exactly once.

Note that we are not de-duplicating internals of descriptors that is beyond scope of this PR.

Why should this be merged: 
Some JVMs have a small memory footprint and if the same set of protobuf files are included in many of them, these protobuf file descriptors get repeated in each of the table's metadata section. Since metadata is memory resident, this can overwhelm JVMs that have a small memory footprint.

Related issue(s) (if applicable): #2965 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -203,25 +232,36 @@ void registerTable(@Nonnull String namespace,
...
+                                for (StackTraceElement st : Thread.currentThread().getStackTrace()) {
+                                    log.warn("{}", st);
+                                }
+                                log.debug("registerTable: old schema: {}", oldRecord.getPayload().getFileDescriptorsMap());
+                                log.debug("registerTable: new schema: {}", tableDescriptors.getFileDescriptorsMap());
+                                this.protobufDescriptorTable.put(entry.getKey(), entry.getValue());
+                                this.registryTable.put(tableNameKey,
+                                        new CorfuRecord<>(tableDescriptors, metadataBuilder.build()));
+                            }
+                        } else { // this is the first time we are seeing this new protobuf file descriptor
```
Lujie1996, line 255: When it reaches this else clause, oldRecord != null && prevDescriptor == null, which means it wants to update table schema using a new descriptor that we have never seen, right? If that's correct, then we can also add log messages here to log the schema change. BTW I found the nested if clauses kind of hard to understand.

hisundar: thanks - let me see if I can refactor this code section

hisundar: done refactoring, can you please review once more? thanks

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -110,11 +111,15 @@ public Timestamp getTimestamp() {
...
             throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        long startTime = System.nanoTime();
         Optional<Timer.Sample> sample = MeterRegistryProvider.getInstance().map(Timer::start);
         Table table =
                 runtime.getTableRegistry().openTable(namespace, tableName, kClass, vClass, mClass, tableOptions);
         corfuStoreMetrics.recordTableCount();
+        int tableSize = table.count();
         table.getMetrics().recordTableOpenTime(sample);
+        log.info("openTable {}${} took {}ms to load {} entries", namespace, tableName,
+                TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startTime), tableSize);
```
Maithem, line 122: If you are logging in milliseconds why do you need to account for time in nanoseconds? converting from nanoseconds to milliseconds will lose any values less than 1ms, so the conversion is superfluous. 

I suggest you just use milliseconds, or instead use microseconds.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -110,11 +111,15 @@ public Timestamp getTimestamp() {
                              @Nullable final Class<M> mClass,
                              @Nonnull final TableOptions tableOptions)
             throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        long startTime = System.nanoTime();
         Optional<Timer.Sample> sample = MeterRegistryProvider.getInstance().map(Timer::start);
         Table table =
                 runtime.getTableRegistry().openTable(namespace, tableName, kClass, vClass, mClass, tableOptions);
         corfuStoreMetrics.recordTableCount();
+        int tableSize = table.count();
```
Maithem, line 119: Can you document that this count is to force a sync on the table, or maybe just introduce a sync method and underneath you can call size.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -178,17 +205,18 @@ void registerTable(@Nonnull String namespace,
         K defaultKeyMessage = (K) keyClass.getMethod("getDefaultInstance").invoke(null);
         V defaultValueMessage = (V) payloadClass.getMethod("getDefaultInstance").invoke(null);
 
+        Map<ProtobufFileName, CorfuRecord<ProtobufFileDescriptor, TableMetadata>> allDescriptors = new HashMap<>();
```
Maithem, line 208: This will only de-duplicate the descriptors within a single table, right? But wouldn't you also want to de-duplicate across tables?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -262,7 +342,13 @@ private void insertAllDependingFileDescriptorProtos(TableDescriptors.Builder tab
...
 
             // Add the fileDescriptorProto into the tableDescriptor map.
-            tableDescriptorsBuilder.putFileDescriptors(fileDescriptorProto.getName(), fileDescriptorProto);
+            tableDescriptorsBuilder.putFileDescriptors(fileDescriptorProto.getName(),
+                    FileDescriptorProto.getDefaultInstance());
+
+            // Add the actual descriptor into a common pool of descriptors to avoid duplication
+            ProtobufFileName protoFileName = ProtobufFileName.newBuilder().setFileName(fileDescriptorProto.getName()).build();
+            ProtobufFileDescriptor protoFd = ProtobufFileDescriptor.newBuilder().setFileDescriptor(fileDescriptorProto).build();
+            allDescriptors.putIfAbsent(protoFileName, new CorfuRecord<>(protoFd, null));
```
Maithem, line 351: Can you do a sanity check before you insert where the protoFileName actually maps to the same proto file descriptor value? otherwise this could fail silently and over write different descriptors.

Maithem, line 351: How can this generate duplicates if the condition in line 340 is 

```java

// If the fileDescriptorProto has already been added then continue.
            if (tableDescriptorsBuilder.getFileDescriptorsMap().containsKey(fileDescriptorProto.getName())) {
                continue;
            }
```

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -240,6 +282,42 @@ void registerTable(@Nonnull String namespace,
...
+     * For each protobuf filename -> descriptor map, validate if the previous descriptor matches
+     * If there is a new schema or an update, record the new update in the table.
+     *
+     * @param protoName - name of the protobuf file
+     * @param newProtoFd - descriptor of the protobuf file
+     * @param namespace - namespace of the table
+     * @param tableName - tablename without the namespace
+     * @return - true if schema change is detected, false otherwise
+     */
+    private boolean recordNewSchema(ProtobufFileName protoName,
```
Maithem, line 295: Why does this return a boolean if its never used?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -240,6 +282,42 @@ void registerTable(@Nonnull String namespace,
...
+                log.warn("registerTable: Schema update detected for table {}${} in schema file {}",
+                        namespace, tableName, protoName);
+
+                for (StackTraceElement st : Thread.currentThread().getStackTrace()) {
+                    log.debug("{}", st);
+                }
+                log.debug("registerTable: old schema: {}", oldProto.getPayload().getFileDescriptor());
+                log.debug("registerTable: new schema: {}", newProtoFd.getPayload().getFileDescriptor());
+            }
+            this.protobufDescriptorTable.put(protoName, newProtoFd);
```
Maithem, line 315: Should this fail instead of emitting a warning and just overriding the old schema?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -90,31 +92,28 @@ public DynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
...
-            });
+        // Cache the FileDescriptorProtos from the protobuf descriptor table.
+        descriptorTable.forEach((fdName, fileDescriptorProto) -> {
+            String protoFileName = fileDescriptorProto.getPayload().getFileDescriptor().getName();
+            // Looks like protobuf file descriptors are randomly truncating the path.
+            // This causes dynamicProtobufSerializer to fail since the full path is needed.
+            if (protoFileName.equals("corfu_options.proto")) {
+                fdProtoMap.putIfAbsent(protoFileName, fileDescriptorProto.getPayload().getFileDescriptor());
+                // Until the truncating issue can be addressed, manually add both paths.
+                protoFileName = "corfudb/runtime/corfu_options.proto";
```
Maithem, line 113: What does this mean, what is the truncating issue?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -90,31 +92,28 @@ public DynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
...
-        // Open the Registry Table.
-        CorfuTable<TableName, CorfuRecord<TableDescriptors, TableMetadata>> corfuTable = corfuRuntime.getObjectsView()
+        // Open the Protobuf Descriptor Table.
+        CorfuTable<ProtobufFileName, CorfuRecord<ProtobufFileDescriptor, TableMetadata>> descriptorTable = corfuRuntime.getObjectsView()
                 .build()
-                .setTypeToken(new TypeToken<CorfuTable<TableName, CorfuRecord<TableDescriptors, TableMetadata>>>() {
+                .setTypeToken(new TypeToken<CorfuTable<ProtobufFileName, CorfuRecord<ProtobufFileDescriptor, TableMetadata>>>() {
                 })
-                .setStreamName(getFullyQualifiedTableName(CORFU_SYSTEM_NAMESPACE, REGISTRY_TABLE_NAME))
+                .setStreamName(getFullyQualifiedTableName(CORFU_SYSTEM_NAMESPACE, PROTOBUF_DESCRIPTOR_TABLE_NAME))
```
Maithem, line 100: This map won't be garbage collected, so you will have two copies of the descriptor map. Why do you need both, can't you just keep one?

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
```diff
@@ -381,7 +408,8 @@ public void browserRegistryTableTest() throws
         runtime = createRuntime(singleNodeEndpoint);
         CorfuStoreBrowser browser = new CorfuStoreBrowser(runtime);
         // Invoke listTables and verify table count
-        Assert.assertEquals(2, browser.printTableInfo(TableRegistry.CORFU_SYSTEM_NAMESPACE,
+        final int three = 3; // keep nutty code check from crying about magic number
```
Maithem, line 411: Remove comment, or escape the check with an ignore annotation.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
Maithem, line 100: We need an explicit unit test that targets this logic.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -202,36 +230,50 @@ void registerTable(@Nonnull String namespace,
...
+             *
+             * 2. Entry exists in TableRegistry -> Existing table re-opened:
+             *    2.1: Existing protobuf file list matches, newly inserted protobuf list
+             *    2.1.1: Each of Existing protobuf file descriptors match that in new file descriptor list
+             *    2.1.2: Existing protobuf file descriptor does not match that in new file descriptor![ SchemaChange!]
+             *
+             *    2.2: Existing protobuf file list does not match, newly opened protobuf list [big Schema Change]
+             *    2.2: Existing protobuf file list matches existing proto files
+             *    2.3: Existing protobuf file's descriptor mismatches with the incoming descriptor
+             */
```
Maithem, line 248: We need an explicit unit test that targets this logic.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: This is just a helper routine to convert the units, nanoseconds tends to be produce long lines, so just picked milliseconds.
The api I use to measure time is System.nanoTime() is there another alternative I should be using?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: Actually, we had a discussion and looks they rather want to retain existing behavior, so I'll remove this count().
If someone wants to force a sync they can call count() or another similar getter api on their own.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: we do deduplicate across tables, please see the protobufDescriptorTable where the key is protobuf file name and value is the descriptor for that file and that is shared across all tables in the table registry.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
Maithem: You can use `System.currentTimeMillis()` , if the load time is less than 1ms and you still want to see how long it takes you can use `System.nanoTime()` and convert to microseconds.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: yes we do that check, please see the method recordNewSchema() this is just placing the descriptor in a stack local map.

hisundar: That's correct, It should not generate duplicates. The de-duplication is really only across tables in the registry not within a given schema's set of protobufs.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: This was from @Lujie1996 's comments above, was trying to refactor the code, let me change the return type back to void.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: We do want to always override the schema that is how developers perform build-to-build upgrades.
Also overwriting schema is normal workflow during upgrades.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: since we are within the corfu runtime, the Any module produces the corfu_options without the path prefix (truncation), but later in another module we require the entire prefix to decode causing deserialization errors, hence this fix-up for corfu_options.proto

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: This was just an optimization to avoid going to the corfu table on every access, which seemed to slow down checkpointing for some reason. Putting a map to cache the corfu table sped up the checkpointing. What might be good is if the corfu table itself can be closed and garbage collected with a new api.

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
hisundar: Can you please share the ignore annotation so I can remove this comment?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: We do, please check CorfuBrowserIT

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Schema update is tested in CorfuStoreShimTest.java
Don't know of an easy way to produce different protobuf files with the same filename, but different schemas.

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
hisundar: Btw this magic number check does not make much sense in test code, I think the magic number checking should not be done for test files. What do you say?

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
hisundar: Done.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Done added unit tests

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: Added more unit tests.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -128,10 +139,19 @@ public TableRegistry(CorfuRuntime runtime) {
                 .setSerializer(this.protobufSerializer)
                 .open();
 
+        this.protobufDescriptorTable = this.runtime.getObjectsView().build()
```
annym, line 142: I'm a bit confused, if the problem is memory limitations, aren't we opening a new table which contains the data already in registryTable? so it's like duplication between them? shouldn't the table registry only contain the type and not the definitions, maybe a mapping to the file name.

runtime/proto/corfu_store_metadata.proto
```diff
@@ -45,3 +45,11 @@ message Record {
     google.protobuf.Any payload = 1;
     google.protobuf.Any metadata = 2;
 }
+
+message ProtobufFileName {
```
pankti-m, line 49: information provided from both these protobufs is already available in TableDescriptors message.

  map<string, google.protobuf.FileDescriptorProto> fileDescriptors = 1;

Why do we need these new protobufs?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -215,6 +216,34 @@ public int printTableInfo(String namespace, String tablename) {
         return tableSize;
     }
 
+    /**
+     * Helper to analyze all the protobufs used in this cluster
+     */
+    public int printAllProtoDescriptors() {
+        int numProtoFiles = -1;
+        log.info("=========KEYS===========");
```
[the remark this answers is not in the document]
hisundar: Done

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -202,36 +230,50 @@ void registerTable(@Nonnull String namespace,
         while (numRetries-- > 0) {
             // Schema validation to ensure that there is either proper modification of the schema across open calls.
             // Or no modification to the protobuf files.
+            /**
+             * Caller is opening a new table with a map of protobufFilename -> protobufFileDescriptor pairs
+             * Here are the cases that can happen:
+             * 1. No entry in TableRegistry -> New table
+             *  1.1: When inserting new protobufs into descriptor table, we find existing entries and they all match
+             *  1.2: When inserting new protobufs into descriptor table, we find mismatching [Schema Change!]
```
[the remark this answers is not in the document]
hisundar: Case 1 is where a table is getting registered for the first time and its sub-cases are what happens when registering its protobuf definitions. So the case where protobuf don't match is same as a schema mismatch 1.2 right?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -178,17 +205,18 @@ void registerTable(@Nonnull String namespace,
...
 
+        Map<ProtobufFileName, CorfuRecord<ProtobufFileDescriptor, TableMetadata>> allDescriptors = new HashMap<>();
         TableDescriptors.Builder tableDescriptorsBuilder = TableDescriptors.newBuilder();
 
         FileDescriptor keyFileDescriptor = defaultKeyMessage.getDescriptorForType().getFile();
-        insertAllDependingFileDescriptorProtos(tableDescriptorsBuilder, keyFileDescriptor);
+        insertAllDependingFileDescriptorProtos(tableDescriptorsBuilder, keyFileDescriptor, allDescriptors);
         FileDescriptor valueFileDescriptor = defaultValueMessage.getDescriptorForType().getFile();
-        insertAllDependingFileDescriptorProtos(tableDescriptorsBuilder, valueFileDescriptor);
+        insertAllDependingFileDescriptorProtos(tableDescriptorsBuilder, valueFileDescriptor, allDescriptors);
```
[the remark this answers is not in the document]
hisundar: We can't right - we need all the definitions to re-construct a DynamicMessage even if it is not part of key, value, metadata directly but get's included indirectly through one of them.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -240,6 +282,42 @@ void registerTable(@Nonnull String namespace,
...
+    private boolean recordNewSchema(ProtobufFileName protoName,
+                                    CorfuRecord<ProtobufFileDescriptor, TableMetadata> newProtoFd,
+                                    String namespace,
+                                    String tableName) {
+        final CorfuRecord<ProtobufFileDescriptor, TableMetadata> oldProto = this.protobufDescriptorTable.get(protoName);
+        if (oldProto == null) {
+            this.protobufDescriptorTable.put(protoName, newProtoFd);
+            return false;
+        }
+        if (!oldProto.getPayload().getFileDescriptor().equals(newProtoFd.getPayload().getFileDescriptor())) {
```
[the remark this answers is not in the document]
hisundar: Yes, I thought about this, but this requires some form of safe garbage collection - a sweep of the table registry and descriptor table to delete non-referenced stale entries. This would need to be done as part of compaction probably for correctness.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: That's exactly what it is - We don't add the descriptors in the table registry, only the file names. the actual descriptors which are the memory hogs are de-duplicated in the protobuf descriptor table.

runtime/proto/corfu_store_metadata.proto
hisundar: This is a new type not a protobuf map right?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
Maithem: You might have not rebased since this diff doesn't look refactored and the return value still hasn't changed.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
Maithem: Also please indicate in a comment that `recordNewSchema` should be execute in a transaction, otherwise an old schema and override a new one in certain race conditions. Also, how do you know what is old and what is new? Is there a monotonic counter that tracks this?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
Maithem: Please capture this in comments.

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
Maithem: You can use `@SuppressWarnings("checkstyle:magicnumber")` .

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -90,31 +92,28 @@ public DynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
         ISerializer protobufSerializer = createProtobufSerializer();
         Serializers.registerSerializer(protobufSerializer);
 
-        // Open the Registry Table.
-        CorfuTable<TableName, CorfuRecord<TableDescriptors, TableMetadata>> corfuTable = corfuRuntime.getObjectsView()
+        // Open the Protobuf Descriptor Table.
+        CorfuTable<ProtobufFileName, CorfuRecord<ProtobufFileDescriptor, TableMetadata>> descriptorTable = corfuRuntime.getObjectsView()
```
Maithem, line 96: This will load the complete `fdProtoMap` map, but only lazily load `fileDescriptorMap`. 

Also, I don't think the serializer itself should do the db access. A higher level of the abstraction should query the DB and initialize the Serializer with the schemas.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -683,4 +683,45 @@ public void queueOrderInTxnContext() throws Exception {
...
+                    someNamespace,
+                    tableNamePrefix+i,
+                    UuidMsg.class,
+                    org.corfudb.runtime.proto.LogData.LogDataMsg.class, // this brings in 1 new protobuf file
+                    ManagedMetadata.class,
+                    // TableOptions includes option to choose - Memory/Disk based corfu table.
+                    TableOptions.builder().build());
+        }
+        final int numProtoFiles2 = corfuRuntime.getTableRegistry().getProtobufDescriptorTable().size();
+        assertThat(numProtoFiles2).isEqualTo(numProtoFiles + 1);
```
Maithem, line 725: This assertion doesn't really check for duplicates. Can you build a `CorfuTable<ProtobufFileName, CorfuRecord<ProtobufFileDescriptor, TableMetadata>>` and make sure its equal to `corfuRuntime.getTableRegistry().getProtobufDescriptorTable()`

Maithem (commented):
Overtime this code has evolved and I think there's enough technical debt to make it worthwhile to clean up the life cycle of the proto schema and the interaction between the DB Access/TableRegistry and the ProtobufSerializer. It's hard to review and follow the logic.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: done

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: done. existing means old, incoming means new.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Done, logged to one line.
It was in info level and they did not like the noise. So they asked us to move it to debug since it's only really needed for debugging and not for correctness.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
hisundar: DynamicProtobufSerializer is only used for tooling and checkpointing. If we aggressively build the complete dependency tree even before opening the first table it would slow down tools like browser which only really want to open one table.
That is the reason for the lazy initialization of fileDescriptorMap.
The serializer's DB access is limited to the constructor. Moving it outside will only complicate the caller's logic who does not need to know how DynamicProtobufSerialization works.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
hisundar: The test updates the schema in 10 tables but the number of descriptors updated is only 1. How is that not de-duplication?
What is a new CorfuTable going to demonstrate that the above test does not?

hisundar:
> Overtime this code has evolved and I think there's enough technical debt to make it worthwhile to clean up the life cycle of the proto schema and the interaction between the DB Access/TableRegistry and the ProtobufSerializer. It's hard to review and follow the logic.

Sure. We can do a follow up PR as soon as we have concrete idea on what the cleaned up logic would be.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
Maithem: Meant to say another map with the expected mapping and not a `CorfuTable`. 

> assertThat(numProtoFiles2).isEqualTo(numProtoFiles + 1);

Is not a sufficient test, what if two keys are being mapped to the same value ?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -215,6 +216,34 @@ public int printTableInfo(String namespace, String tablename) {
         return tableSize;
     }
 
+    /**
+     * Helper to analyze all the protobufs used in this cluster
+     */
+    public int printAllProtoDescriptors() {
+        int numProtoFiles = -1;
```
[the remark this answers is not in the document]
hisundar: actually as @zfrenette had pointed out returning 0, can hide a case where there should be some data but the test passes since 0 is the default value. Since this is for testing I would like to leave it at -1. please let me know if that's ok

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
hisundar: De-duplicating actual FileDescriptorProto contents is not trivial and beyond the scope of this change.
This PR only seeks to store one FileDescriptorProto per filename.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -262,7 +347,13 @@ private void insertAllDependingFileDescriptorProtos(TableDescriptors.Builder tab
             }
 
             // Add the fileDescriptorProto into the tableDescriptor map.
-            tableDescriptorsBuilder.putFileDescriptors(fileDescriptorProto.getName(), fileDescriptorProto);
+            tableDescriptorsBuilder.putFileDescriptors(fileDescriptorProto.getName(),
+                    FileDescriptorProto.getDefaultInstance());
```
pankti-m, line 351: this is the key part which does the de-duplication.  Earlier we had the same file descriptor proto for every protobuf in the same file.  Now this is replacing it with a generic protobuf with default values.  Is that right?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -202,36 +230,53 @@ void registerTable(@Nonnull String namespace,
         while (numRetries-- > 0) {
             // Schema validation to ensure that there is either proper modification of the schema across open calls.
             // Or no modification to the protobuf files.
+            /**
+             * Caller is opening a new table with a map of protobufFilename -> protobufFileDescriptor pairs
+             * Here are the cases that can happen:
+             * 1. No entry in TableRegistry -> New table
```
pankti-m, line 236: If a table is new(never seen before), how(between what) can we detect a mismatch?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -202,36 +230,53 @@ void registerTable(@Nonnull String namespace,
...
+             *  1.1: When inserting new protobufs into descriptor table, we find existing entries and they all match
+             *  1.2: When inserting new protobufs into descriptor table, we find mismatching [Schema Change!]
+             *
+             * 2. Entry exists in TableRegistry -> Existing table re-opened:
+             *    2.1: Existing protobuf file list matches, newly inserted protobuf list
+             *    2.1.1: Each of Existing protobuf file descriptors match that in new file descriptor list
+             *    2.1.2: Existing protobuf file descriptor does not match that in new file descriptor![ SchemaChange!]
+             *
+             *    2.2: Existing protobuf file list does not match, newly opened protobuf list [big Schema Change]
+             *    2.2: Existing protobuf file list matches existing proto files
```
pankti-m, line 246: duplicate 2.2?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -202,36 +230,53 @@ void registerTable(@Nonnull String namespace,
...
+             *    2.3: Existing protobuf file's descriptor mismatches with the incoming descriptor
```
pankti-m, line 247: 2.2.3?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -202,36 +230,53 @@ void registerTable(@Nonnull String namespace,
...
+                } else { // Case 2 above, re-open of existing table
+                    if (!oldRecord.getPayload().getFileDescriptorsMap().keySet()
+                            .equals(tableDescriptors.getFileDescriptorsMap().keySet())) {
+                        log.warn("Protobuf file list changed for {} table {}. Prev list {}, new list {}",
+                                namespace, tableName, oldRecord.getPayload().getFileDescriptorsMap().keySet(),
+                                tableDescriptors.getFileDescriptorsMap().keySet());
+                        for (StackTraceElement st : Thread.currentThread().getStackTrace()) {
+                            log.debug("{}", st);
+                        }
+                        this.registryTable.put(tableNameKey,
```
pankti-m, line 267: should this be treated as an error instead?  Is it right to continue in this case?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -240,15 +285,55 @@ void registerTable(@Nonnull String namespace,
...
+     * @param namespace - namespace of the table
+     * @param tableName - tablename without the namespace
+     */
+    private void recordNewSchema(ProtobufFileName protoName,
+                                    CorfuRecord<ProtobufFileDescriptor, TableMetadata> newProtoFd,
+                                    String namespace,
+                                    String tableName) {
+        final CorfuRecord<ProtobufFileDescriptor, TableMetadata> oldProto = this.protobufDescriptorTable.get(protoName);
+        if (oldProto == null) {
+            this.protobufDescriptorTable.put(protoName, newProtoFd);
```
pankti-m, line 305: wouldnt this have already been populated in insertDependingFileDescriptorProtos()?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -240,15 +285,55 @@ void registerTable(@Nonnull String namespace,
...
+            return;
+        }
+        if (!oldProto.getPayload().getFileDescriptor().equals(newProtoFd.getPayload().getFileDescriptor())) {
+            if (!protoName.getFileName().startsWith("google/protobuf")) {
```
pankti-m, line 309: why are we ignoring the files in the google protobuf library?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -240,15 +285,55 @@ void registerTable(@Nonnull String namespace,
...
+                StringBuilder stackTraceStr = new StringBuilder();
+                for (StackTraceElement st : Thread.currentThread().getStackTrace()) {
+                    stackTraceStr.append(st);
+                    stackTraceStr.append("\n");
+                }
+                log.debug("{}", stackTraceStr);
+                log.debug("registerTable: old schema: {}", oldProto.getPayload().getFileDescriptor());
+                log.debug("registerTable: new schema: {}", newProtoFd.getPayload().getFileDescriptor());
+            }
+            this.protobufDescriptorTable.put(protoName, newProtoFd);
```
pankti-m, line 322: shouldnt this be treated as an error or unexpected case?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: exactly default values take up almost nothing compared to the real thing and that's where the space savings come from.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: thanks

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: thanks

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: In data-migration this is the normal process, also in build-to-build experiments that different developers do from now until the release, this path is also expected and should not be treated as an error.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: No, in `insertAllDependingFileDescriptorProtos()` we do not touch the actual protobuf Table because we want to do it transactionally.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: 2 reasons:
1. they are default library files, even if there is a schema change there nothing we can or should do about it.
2. They are very huge with complex nested types, comparing these itself is a CPU intensive process.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: No, because in data-migration and build-to-build experiments done by developers this is the normal path and we should allow it to happen.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
Maithem: Instead of constantly trying to explain what the code "should" do in the comments section its a better idea to enforce it in code with a **properly** written **unit test**. In this particular test case merely checking the size of the map doesn't assert your expectations of the map's contents. Iterating the mapping one-by-one and assert that the contents map the right file name (ProtobufFileName) to the correct descriptor.

hisundar: Sure, will do. But please note that it's really not possible for the map to have incorrect contents and still have all the CorfuStoreIT, StreamingIT and CorfuStoreBrowserIT pass, because the new map is critical to build the dynamic message.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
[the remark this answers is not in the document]
hisundar: yes that is correct

annym (approved):

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: We can still detect a mismatch if a new unseen table tries to open and existing protobuf file which has now changed.
For example if someone change the corfu_store_metadata.proto and opened a new table, we can detect that there is a schema change with this PR. Before it was not possible.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: done. Removal is a whole other conversation. We need to provide future releases with a removeTable() functionality so we can safely delete old stale protobufs associated with a deprecated table.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -247,19 +247,6 @@ public void checkMetadataTransactions() throws Exception {
                         .setNestedType(SampleSchema.NestedTypeA.newBuilder().build())
                         .setVersion(expectedVersion).build());
 
-        // Validate that if Metadata schema is specified, a null metadata is not acceptable
```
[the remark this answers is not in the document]
hisundar: these checks were testing a deprecated api `table.update()` which I removed as part of the cleanup.

Maithem (approved):

pankti-m (approved):

PR #2968, opened 2021-05-25
StreamingIT: Example for getPreviousValue

hisundar:
## Overview

Description:

Why should this be merged: Shares a simple example on how to get the previous value from a stream entry.

- [x] Public API has Javadoc


test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -441,6 +443,130 @@ public void testStreamingMultiTableSameListener() throws Exception {
...
+         * @param epoch
+         * @param sequence
+         * @return
+         */
+        public CorfuStoreEntry<K, V, M> getPreviousRecord(String namespace, String tableName,
+                                                          CorfuStreamEntry entry,
+                                                          long epoch,
+                                                          long sequence) {
+            Timestamp prevTimestamp = Timestamp.newBuilder()
+                    .setEpoch(epoch)
```
zhangn49, line 505: For the actual user, is it possible that the previous timestamp has a lower epoch? If so, will it get previous record or a null value?

zhangn49 (approved):
LGTM, one question.

test/src/test/java/org/corfudb/integration/StreamingIT.java
hisundar: this entry carries the exact epoch and address, so if the previous timestamp epoch is lower, then it should get the previous record. I haven't tested that case though to be sure. thanks

PR #2969, opened 2021-05-26
Add Support for Editing Records in CorfuStore Tables.

pankti-m:
Add Support for Editing Records in CorfuStore Tables.

Description:
Allow editing the record corresponding to a given key in a CorfuStore table.

CorfuStore tables contain records in protobuf format.  To construct a protobuf from the JSON string inputs, use type information stored in TableRegistry.

Why should this be merged: Editing an existing record is a key requirement for debugging and troubleshooting.

- [ ] Public API has Javadoc

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
@@ -18,8 +18,11 @@ import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import com.google.protobuf.Any;
+import com.google.protobuf.DynamicMessage;
 import com.google.protobuf.InvalidProtocolBufferException;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
@@ -43,11 +46,12 @@ import org.corfudb.runtime.collections.StreamingMap;
 import org.corfudb.runtime.collections.Table;
 import org.corfudb.runtime.collections.TableOptions;
 import org.corfudb.runtime.collections.ManagedTxnContext;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.object.ICorfuVersionPolicy;
+import org.corfudb.runtime.object.transactions.TransactionalContext;
 import org.corfudb.runtime.view.SMRObject;
 import org.corfudb.runtime.view.TableRegistry;
 import org.corfudb.util.serializer.DynamicProtobufSerializer;
-import org.corfudb.util.serializer.ISerializer;
 import org.corfudb.util.serializer.Serializers;
 import org.rocksdb.Options;
 
@@ -64,14 +68,14 @@ import javax.annotation.Nonnull;
 public class CorfuStoreBrowser {
     private final CorfuRuntime runtime;
     private final String diskPath;
+    private final DynamicProtobufSerializer dynamicProtobufSerializer;
 
     /**
      * Creates a CorfuBrowser which connects a runtime to the server.
      * @param runtime CorfuRuntime which has connected to the server
      */
     public CorfuStoreBrowser(CorfuRuntime runtime) {
-        this.runtime = runtime;
-        this.diskPath = null;
+        this(runtime, null);
     }
 
     /**
@@ -83,28 +87,9 @@ public class CorfuStoreBrowser {
     public CorfuStoreBrowser(CorfuRuntime runtime, String diskPath) {
         this.runtime = runtime;
         this.diskPath = diskPath;
-    }
-
-    /**
-     * Validate that the namespace is not null
-     * @param namespace
-     */
-    private static void verifyNamespace(String namespace) {
-        if (namespace == null) {
-            throw new IllegalArgumentException("Please specify --namespace");
-        }
-    }
-
-    /**
-     * Validate that both namespace and tablename are present
-     * @param namespace - the namespace where the table belongs
-     * @param tableName - table name without the namespace
-     */
-    private static void verifyNamespaceAndTablename(String namespace, String tableName) {
-        verifyNamespace(namespace);
-        if (tableName == null) {
-            throw new IllegalArgumentException("Please specify --tablename");
-        }
+        dynamicProtobufSerializer =
+            new DynamicProtobufSerializer(runtime);
+        Serializers.registerSerializer(dynamicProtobufSerializer);
     }
 
     /**
@@ -118,10 +103,8 @@ public class CorfuStoreBrowser {
         log.info("Namespace: {}", namespace);
         log.info("TableName: {}", tableName);
 
-        ISerializer dynamicProtobufSerializer =
-                new DynamicProtobufSerializer(runtime);
-        Serializers.registerSerializer(dynamicProtobufSerializer);
         String fullTableName = TableRegistry.getFullyQualifiedTableName(namespace, tableName);
+
         SMRObject.Builder<CorfuTable<CorfuDynamicKey, CorfuDynamicRecord>> corfuTableBuilder =
         runtime.getObjectsView().build()
                 .setTypeToken(new TypeToken<CorfuTable<CorfuDynamicKey, CorfuDynamicRecord>>() {})
@@ -137,7 +120,6 @@ public class CorfuStoreBrowser {
                             dynamicProtobufSerializer, runtime);
             corfuTableBuilder.setArguments(mapSupplier, ICorfuVersionPolicy.MONOTONIC);
         }
-
         return corfuTableBuilder.open();
     }
 
@@ -148,10 +130,10 @@ public class CorfuStoreBrowser {
      * @return - number of entries in the table
      */
     public int printTable(String namespace, String tablename) {
-        verifyNamespaceAndTablename(namespace, tablename);
         StringBuilder builder;
 
-        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(namespace, tablename);
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table =
+            getTable(namespace, tablename);
         int size = table.size();
         final int batchSize = 50;
         Stream<Map.Entry<CorfuDynamicKey, CorfuDynamicRecord>> entryStream = table.entryStream();
@@ -185,17 +167,22 @@ public class CorfuStoreBrowser {
      */
     public int listTables(String namespace)
     {
-        verifyNamespace(namespace);
         int numTables = 0;
         log.info("\n=====Tables=======\n");
-        for (TableName tableName : runtime.getTableRegistry()
-                .listTables(namespace)) {
+        for (TableName tableName : listTablesInNamespace(namespace)) {
             log.info("Table: " + tableName.getTableName());
             log.info("Namespace: " + tableName.getNamespace());
             numTables++;
         }
         log.info("\n======================\n");
         return numTables;
+    }
+
+    private List<TableName> listTablesInNamespace(String namespace) {
+        return dynamicProtobufSerializer.getCachedRegistryTable().keySet()
+            .stream()
+            .filter(tableName -> namespace == null || tableName.getNamespace().equals(namespace))
+            .collect(Collectors.toList());
     }
 
     /**
@@ -205,11 +192,11 @@ public class CorfuStoreBrowser {
      * @return - number of entries in the table
      */
     public int printTableInfo(String namespace, String tablename) {
-        verifyNamespaceAndTablename(namespace, tablename);
         log.info("\n======================\n");
         String fullName = TableRegistry.getFullyQualifiedTableName(namespace, tablename);
         UUID streamUUID = UUID.nameUUIDFromBytes(fullName.getBytes());
-        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(namespace, tablename);
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table =
+            getTable(namespace, tablename);
         int tableSize = table.size();
         log.info("Table {} in namespace {} with ID {} has {} entries",
                 tablename, namespace, streamUUID.toString(), tableSize);
@@ -223,7 +210,8 @@ public class CorfuStoreBrowser {
     public int printAllProtoDescriptors() {
         int numProtoFiles = -1;
         log.info("=========PROTOBUF FILE NAMES===========");
-        for (ProtobufFileName protoFileName : runtime.getTableRegistry().getProtobufDescriptorTable().keySet()) {
+        for (ProtobufFileName protoFileName :
+            dynamicProtobufSerializer.getCachedProtobufDescriptorTable().keySet()) {
             try {
                 log.info("{}", JsonFormat.printer().print(protoFileName));
             } catch (InvalidProtocolBufferException e) {
@@ -232,11 +220,13 @@ public class CorfuStoreBrowser {
             numProtoFiles++;
         }
         log.info("=========PROTOBUF FILE DESCRIPTORS ===========");
-        for (ProtobufFileName protoFileName : runtime.getTableRegistry().getProtobufDescriptorTable().keySet()) {
+        for (ProtobufFileName protoFileName :
+            dynamicProtobufSerializer.getCachedProtobufDescriptorTable().keySet()) {
             try {
                 log.info("{}", JsonFormat.printer().print(protoFileName));
                 log.info("{}", JsonFormat.printer().print(
-                        runtime.getTableRegistry().getProtobufDescriptorTable().get(protoFileName).getPayload())
+                    dynamicProtobufSerializer.getCachedProtobufDescriptorTable()
+                        .get(protoFileName).getPayload())
                 );
             } catch (InvalidProtocolBufferException e) {
                 log.error("Unable to print protobuf for key {}", protoFileName, e);
@@ -252,18 +242,110 @@ public class CorfuStoreBrowser {
      * @return - number of entries in the table before clearing the table
      */
     public int dropTable(String namespace, String tablename) {
-        verifyNamespaceAndTablename(namespace, tablename);
         log.info("\n======================\n");
         String fullName = TableRegistry.getFullyQualifiedTableName(namespace, tablename);
         UUID streamUUID = UUID.nameUUIDFromBytes(fullName.getBytes());
-        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(namespace, tablename);
-        int tableSize = table.size();
-        log.info("Table {} in namespace {} with ID {} with {} entries will be dropped...",
+        try {
+            runtime.getObjectsView().TXBegin();
+            CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table =
+                getTable(namespace, tablename);
+            int tableSize = table.size();
+            log.info("Table {} in namespace {} with ID {} with {} entries will be dropped...",
                 tablename, namespace, streamUUID.toString(), tableSize);
-        table.clear();
-        log.info("Table cleared successfully");
+            table.clear();
+            runtime.getObjectsView().TXEnd();
+            log.info("Table cleared successfully");
+            log.info("\n======================\n");
+            return tableSize;
+        } catch (TransactionAbortedException e) {
+            log.error("Drop Table Transaction Aborted");
+        } finally {
+            if (TransactionalContext.isInTransaction()) {
+                runtime.getObjectsView().TXAbort();
+            }
+        }
+        return -1;
+    }
+
+    /**
+     * Edit a record in a table and namespace
+     * @param namespace namespace of the table
+     * @param tableName name of the table
+     * @param keyToEdit JSON string representing the key whose corresponding
+     *  record is to be editted
+     * @param newRecord JSON string representing the new value to be inserted
+     *  against keyToEdit
+     * @return CorfuDynamicRecord the edited CorfuDynamicRecord.  null if no
+     *  record was edited, either due to an error or key not found.
+     */
+    public CorfuDynamicRecord editRecord(String namespace, String tableName,
+        String keyToEdit, String newRecord) {
         log.info("\n======================\n");
-        return tableSize;
+        String fullName = TableRegistry.getFullyQualifiedTableName(namespace,
+            tableName);
+        UUID streamUUID = CorfuRuntime.getStreamID(fullName);
+
+        TableName tableNameProto = TableName.newBuilder().setTableName(tableName)
+            .setNamespace(namespace).build();
+
+        Any defaultKeyAny =
+            dynamicProtobufSerializer.getCachedRegistryTable().get(tableNameProto)
+            .getPayload().getKey();
+        Any defaultValueAny =
+            dynamicProtobufSerializer.getCachedRegistryTable().get(tableNameProto)
+            .getPayload().getValue();
+        DynamicMessage keyMsg =
+            dynamicProtobufSerializer.createDynamicMessageFromJson(defaultKeyAny,
+                keyToEdit);
+        DynamicMessage newValueMsg =
+            dynamicProtobufSerializer.createDynamicMessageFromJson(defaultValueAny,
+            newRecord);
+
+        if (keyMsg == null || newValueMsg == null) {
+            return null;
+        }
+
+        CorfuDynamicKey dynamicKey =
+            new CorfuDynamicKey(defaultKeyAny.getTypeUrl(), keyMsg);
+
+        try {
+            runtime.getObjectsView().TXBegin();
+            CorfuDynamicRecord editedRecord = null;
+            CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table =
+                getTable(namespace, tableName);
+            if (table.containsKey(dynamicKey)) {
+                CorfuDynamicRecord oldRecord = table.get(dynamicKey);
+
+                if (oldRecord == null) {
+                    log.warn("Unexpedted Null Value found for key {} in table " +
+                            "{} and namespace {}.  Stream Id {}", keyToEdit, tableName,
+                        namespace, streamUUID);
+                } else {
+                    log.info("Editing record with Key {} in table {} and " +
+                            "namespace {} with new record {}.  Stream Id {}.",
+                        keyToEdit, tableName, namespace, newRecord, streamUUID);
+                    String payloadTypeUrl = oldRecord.getPayloadTypeUrl();
+                    String metadataTypeUrl = oldRecord.getMetadataTypeUrl();
+                    DynamicMessage metadata = oldRecord.getMetadata();
+                    editedRecord = new CorfuDynamicRecord(payloadTypeUrl,
+                        newValueMsg, metadataTypeUrl, metadata);
+                    table.put(dynamicKey, editedRecord);
+                }
+            } else {
+                log.warn("Record with key {} not found in table {} and namespace {}. " +
+                    " Stream Id {}.", keyToEdit, tableName, namespace, streamUUID);
+            }
+            runtime.getObjectsView().TXEnd();
+            log.info("\n======================\n");
+            return editedRecord;
+        } catch (TransactionAbortedException e) {
+            log.error("Transaction to edit record aborted.", e);
+        } finally {
+            if (TransactionalContext.isInTransaction()) {
+                runtime.getObjectsView().TXAbort();
+            }
+        }
+        return null;
     }
 
     /**
@@ -276,7 +358,6 @@ public class CorfuStoreBrowser {
      * @return - number of entries in the table
      */
     public int loadTable(String namespace, String tablename, int numItems, int batchSize, int itemSize) {
-        verifyNamespaceAndTablename(namespace, tablename);
         CorfuStoreShim store = new CorfuStoreShim(runtime);
         try {
             TableOptions.TableOptionsBuilder<Object, Object> optionsBuilder = TableOptions.builder();
@@ -329,7 +410,6 @@ public class CorfuStoreBrowser {
      * @return number of updates read so far
      */
     public long listenOnTable(String namespace, String tableName, int stopAfter) {
-        verifyNamespaceAndTablename(namespace, tableName);
         CorfuStoreShim store = new CorfuStoreShim(runtime);
         final Table<ExampleTableName, ExampleTableName, ManagedMetadata> table;
         try {
@@ -431,8 +511,10 @@ public class CorfuStoreBrowser {
      * */
     public Set<String> listStreamTags() {
         Set<String> streamTags = new HashSet<>();
-        runtime.getTableRegistry().getRegistryTable().values().forEach(record ->
-                streamTags.addAll(record.getMetadata().getTableOptions().getStreamTagList()));
+
+        dynamicProtobufSerializer.getCachedRegistryTable().values().forEach(
+            record -> streamTags.addAll(record.getMetadata()
+                .getTableOptions().getStreamTagList()));
 
         log.info("\n======================\n");
         log.info("Total unique stream tags: [{}]", streamTags.size());
@@ -464,11 +546,11 @@ public class CorfuStoreBrowser {
      * @param table table name of interest
      */
     public Set<String> listTagsForTable(String namespace, String table) {
-        verifyNamespaceAndTablename(namespace, table);
         Set<String> tags = new HashSet<>();
         TableName tableName = TableName.newBuilder().setNamespace(namespace).setTableName(table).build();
-        CorfuRecord<CorfuStoreMetadata.TableDescriptors, CorfuStoreMetadata.TableMetadata> record = runtime.getTableRegistry()
-                .getRegistryTable().get(tableName);
+
+        CorfuRecord<CorfuStoreMetadata.TableDescriptors, CorfuStoreMetadata.TableMetadata> record =
+            dynamicProtobufSerializer.getCachedRegistryTable().get(tableName);
         if (record != null) {
             tags.addAll(record.getMetadata().getTableOptions().getStreamTagList());
             log.info("\n======================\n");
@@ -503,7 +585,8 @@ public class CorfuStoreBrowser {
 
     private Map<String, List<TableName>> getTagToTableNamesMap() {
         Map<String, List<TableName>> streamTagToTableNames = new HashMap<>();
-        runtime.getTableRegistry().getRegistryTable().forEach((tableName, schema) ->
+
+        dynamicProtobufSerializer.getCachedRegistryTable().forEach((tableName, schema) ->
                 schema.getMetadata().getTableOptions().getStreamTagList().forEach(tag -> {
                     if (streamTagToTableNames.putIfAbsent(tag, new ArrayList<>(Arrays.asList(tableName))) != null) {
                         streamTagToTableNames.computeIfPresent(tag, (key, tableList) -> {
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
@@ -3,6 +3,7 @@ package org.corfudb.browser;
 import java.util.Map;
 import java.util.Optional;
 
+import com.google.common.base.Preconditions;
 import lombok.extern.slf4j.Slf4j;
 
 import org.corfudb.runtime.CorfuRuntime;
@@ -25,6 +26,7 @@ public class CorfuStoreBrowserMain {
         listenOnTable,
         dropTable,
         listAllProtos,
+        editTable,
         listTags,
         listTablesForTag,
         listTagsForTable,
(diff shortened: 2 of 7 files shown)


hisundar (commented):
Looks good to me, Pankti, can we verify that we don't get a InvalidProtocolBufferException on the merge() call?
Also is it possible to add a CorfuStoreBrowserIT test for the editRecord change?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -65,6 +73,20 @@
     private final CorfuRuntime runtime;
     private final String diskPath;
 
+    private final ConcurrentMap<String, DescriptorProtos.FileDescriptorProto>
+        fdProtoMap = new ConcurrentHashMap<>();
+    /*
+     * Maps the Message name to the name of the FileDescriptorProto containing it.
+     */
+    private final ConcurrentMap<String, String> messagesFdProtoNameMap =
```
hisundar, line 81: Isn't this the same map in DynamicProtobufSerializer class which can be used instead?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -65,6 +73,20 @@
     private final CorfuRuntime runtime;
     private final String diskPath;
 
+    private final ConcurrentMap<String, DescriptorProtos.FileDescriptorProto>
```
hisundar, line 76: This is the same as ProtobufDescriptorTable in TableRegistry can we use that instead?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -65,6 +73,20 @@
...
+    /*
+     * Maps the Message name to the name of the FileDescriptorProto containing it.
+     */
+    private final ConcurrentMap<String, String> messagesFdProtoNameMap =
+        new ConcurrentHashMap<>();
+
+    /**
+     * This is used as a file descriptor cache. Used for optimization.
+     */
+    private final ConcurrentMap<String, Descriptors.FileDescriptor>
```
hisundar, line 87: This is also present in the DynamicProtobufSerializer class, perhaps you just need a `@Getter` method on that right?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -116,6 +119,10 @@
     @Getter
     private final CorfuTable<ProtobufFileName, CorfuRecord<ProtobufFileDescriptor, TableMetadata>> protobufDescriptorTable;
 
+    @Getter
+    private final CorfuTable<TableName, CorfuStoreMetadata.KeyValueDefaultRecord>
```
hisundar, line 123: TableRegistry already maps tableName to a list of protos, instead of a separate table, can we instead just put a boolean field or metadata in the TableRegistry's descriptor to indicate who is the default protobuf file for a given tableName? This will otherwise double the memory for the TableRegistry's types, which can be expensive.

hisundar (changes requested):

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: thats right.  I wasnt sure if we wanted to expose anything from DynamicProtobufSerializer, so replicated it.  I will the change to reuse methods/variables from there wherever possible.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: not needed if we are resusing the methods from DynamicProtobufSerializer

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: got it

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
pankti-m: it is not sufficient to get the file name or file descriptor proto.  We need to somehow construct a dynamic message of the required type using which the key and value json strings can be parsed.  I couldnt find any way to do that using the file descriptor proto or file name.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -12,14 +12,21 @@
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
Maithem, line 15: Please add a proper git message that explains these changes.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -118,8 +128,7 @@ private static void verifyNamespaceAndTablename(String namespace, String tableNa
         log.info("Namespace: {}", namespace);
         log.info("TableName: {}", tableName);
 
-        ISerializer dynamicProtobufSerializer =
-                new DynamicProtobufSerializer(runtime);
+        dynamicProtobufSerializer = new DynamicProtobufSerializer(runtime);
```
Maithem, line 131: What is this set in a method and not in the constructor ?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+
+        if (keyToEdit == null) {
+            throw new IllegalArgumentException("Please specify the key to " +
+                "edit --keyToEdit");
+        }
+
+        if (newRecord == null) {
+            throw new IllegalArgumentException("Please specify the newRecord " +
+                "--newRecord");
+        }
```
Maithem, line 290: These validity checks don't look correct. What if it's not a null, but an empty string ?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+
+        log.info("\n======================\n");
+        String fullName = TableRegistry.getFullyQualifiedTableName(namespace,
+            tableName);
+        UUID streamUUID = UUID.nameUUIDFromBytes(fullName.getBytes());
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table = getTable(
+            namespace, tableName);
+
+        TableName tableNameProto = TableName.newBuilder().setTableName(tableName)
+            .setNamespace(namespace).build();
```
Maithem, line 300: Why are you not using `TableRegistry::openTable` ? It seems like a better choice because it will also check if the table exists in the database.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+
+        if (newRecord == null) {
+            throw new IllegalArgumentException("Please specify the newRecord " +
+                "--newRecord");
+        }
+
+        log.info("\n======================\n");
+        String fullName = TableRegistry.getFullyQualifiedTableName(namespace,
+            tableName);
+        UUID streamUUID = UUID.nameUUIDFromBytes(fullName.getBytes());
```
Maithem, line 295: If the format of StreamName -> UUID changes then this will be inconsistent. Use `CorfuRuntime::getStreamID` instead

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+        Any defaultKeyAny = keyValueDefaultRecord.getKey();
+        Any defaultValueAny = keyValueDefaultRecord.getValue();
+
+        DynamicMessage keyMsg = createMsgFromJson(defaultKeyAny, keyToEdit);
+        DynamicMessage newValueMsg = createMsgFromJson(defaultValueAny,
+            newRecord);
+
+        if (keyMsg == null || newValueMsg == null) {
+            return 0;
+        }
```
Maithem, line 313: It's a common convention that a failing process returns a non-zero value, so this looks a little weird. 

Also why can't you just rethrow the exceptions and let it fail instead of catching the exception, logging it, returning null, then in the caller method checking if its null and returning 0?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+            .setNamespace(namespace).build();
+
+        CorfuStoreMetadata.KeyValueDefaultRecord keyValueDefaultRecord =
+            runtime.getTableRegistry().getTableNameToKeyValueTypeTable().get(tableNameProto);
+        Any defaultKeyAny = keyValueDefaultRecord.getKey();
+        Any defaultValueAny = keyValueDefaultRecord.getValue();
+
+        DynamicMessage keyMsg = createMsgFromJson(defaultKeyAny, keyToEdit);
+        DynamicMessage newValueMsg = createMsgFromJson(defaultValueAny,
+            newRecord);
```
Maithem, line 309: This looks like functionality that should be pushed to the `DynamicProtobufSerializer` class. Accessing most of that class field members externally is already an indication that it should be refactored, and also expose generic functionality via methods and not direct access to the class fields.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+
+        if (keyMsg == null || newValueMsg == null) {
+            return 0;
+        }
+
+        for (CorfuDynamicKey key : table.keySet()) {
+            if (Objects.equals(keyMsg.toString(), key.getKey().toString())) {
```
Maithem, line 316: Although this might work in some cases, it is a wrong comparison in general. A `toString` method usually creates a string representation that is a "summary" and not necessary a complete representation of the object. One example of this, a map can have two different keys that have the same `toString` representation. 

Also, why do you need to iterate the keyset if this is already a hashmap? can't you just get a get a protobuf message for the key and do table.contains(key) ?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+        } catch (InvalidProtocolBufferException e) {
+            log.warn("Unable to Parse Key {}", anyMsg.getValue());
+            return null;
+        }
+
+        try {
+            JsonFormat.parser().merge(jsonString, builder);
+        } catch(InvalidProtocolBufferException e) {
+            log.warn("Unable to Parse String {}", jsonString);
+            return null;
```
Maithem, line 365: This logic seems generic enough to be pushed to the serializer.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+            if (Objects.equals(keyMsg.toString(), key.getKey().toString())) {
+                log.info("Editing record with Key {} in table {} and " +
+                        "namespace {} with new record {}.  Stream Id {}.",
+                    keyToEdit, tableName, namespace, newRecord, streamUUID);
+                String payloadTypeUrl = table.get(key).getPayloadTypeUrl();
+                String metadataTypeUrl = table.get(key).getMetadataTypeUrl();
+                DynamicMessage metadata = table.get(key).getMetadata();
+                table.put(key, new CorfuDynamicRecord(payloadTypeUrl,
+                    newValueMsg, metadataTypeUrl, metadata));
+                log.info("\n======================\n");
```
Maithem, line 325: This write should be transactional otherwise it can cause silent failures. For example, imagine a vertical trying to edit a key while this command is executed. Both operations won't fail, but one of them should have.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +275,98 @@ public int dropTable(String namespace, String tablename) {
...
+                return 1;
+            }
+        }
+        log.info("Record with key {} not found in table {} and namespace {}. " +
+            " Stream Id {}.", keyToEdit, tableName, namespace, streamUUID);
+        log.info("\n======================\n");
+        return 0;
```
Maithem, line 332: This is inconsistent. In like 312 `0` is returned to indicate failure but here `0` indicates success. Please use the same convention for both.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
```diff
@@ -56,6 +59,8 @@
         + "--numItems=<numItems> Total Number of items for loadTable\n"
         + "--batchSize=<batchSize> Number of records per transaction for loadTable\n"
         + "--itemSize=<itemSize> Size of each item's payload for loadTable\n"
+        + "--keyToEdit=<keyToEdit> Key of the record to edit\n"
+        + "--newRecord=<newRecord> New Editted record to insert\n"
```
Maithem, line 63: Edit key and add key almost share the same logic, why not support both ?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
Maithem: This will increase the memory used for the dynamic serializer for all consumers while it's only needed for the browser tool. 

Essentially, you need to map a Table -> fileNames -> descriptors. I though the table metadata contained which types are used in a table. No?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: good point.  I can add it in the constructor

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: TableRegistry::openTable requires the key/value/metadata classes which we do not know

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: in all methods in the tool, we return the number of records found, deleted, edited etc.  So I followed that approach.  But it is equally correct to throw the exceptions instead.

@hisundar, what do you think?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: +1 on thinking if any refactoring is possible.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: Agreed.  However, it is not possible to compare field values with equals() in DynamicProtobuf.  Let me know if you have a better suggestion.

Looking at the other comment on iterating vs get() on a key.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: DynamicProtobufSerializer is for conversion to/from bytebuf to/from dynamic messages.  Json string is very specific to the editor use case.  Should we still move this there?  Let me know your thoughts.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: noted

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
pankti-m: valid point

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: if it is an empty string, the parser will throw an InvalidProtocolBufferException.  I can add a test for this and other corner cases if can think of something additional.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: @Maithem , adding it in the constructor seems to be registering >1 serializers with the runtime and causes tests to fail. So I dont have that change in the latest commit.  I will try to debug it tomorrow.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +274,114 @@ public int dropTable(String namespace, String tablename) {
         return tableSize;
     }
 
+    public int editRecord(String namespace, String tableName,
+        String keyToEdit, String newRecord) {
+        verifyNamespaceAndTablename(namespace, tableName);
+
```
zhangn49, line 280: Can we have a verify function to check null value and empty?
like verifyInput(String str, String inputName)
and all null or empty value will see the consistent IllegalArugumentException

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +274,114 @@ public int dropTable(String namespace, String tablename) {
...
+        CorfuDynamicKey dynamicKey =
+            new CorfuDynamicKey(defaultKeyAny.getTypeUrl(), keyMsg);
+
+        if (table.containsKey(dynamicKey)) {
+            log.info("Editing record with Key {} in table {} and " +
+                    "namespace {} with new record {}.  Stream Id {}.",
+                keyToEdit, tableName, namespace, newRecord, streamUUID);
+            String payloadTypeUrl = table.get(dynamicKey).getPayloadTypeUrl();
+            String metadataTypeUrl = table.get(dynamicKey).getMetadataTypeUrl();
+            DynamicMessage metadata = table.get(dynamicKey).getMetadata();
```
zhangn49, line 323: Does table allow null as a key's value? If so, we might hit NPE

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -266,6 +274,114 @@ public int dropTable(String namespace, String tablename) {
...
+            table.put(dynamicKey, new CorfuDynamicRecord(payloadTypeUrl,
```
zhangn49, line 324: should be a transaction write?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
zhangn49: Looks like you catch the InvalidProtocolBufferException and return a null, then the editRecord will return 0, which means it will fail silently. Right?

zhangn49 (commented):
Minor comments and please address codacy issues

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: DynamicProtobufSerializer is only needed by operations like printTable/dropTable/editRecord.  So only creating and registering it in those operations.

In the meantime, will explore how other operations in the tool can use the DynamicProtobufSerializer as well.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: the new behavior now has a void return type.  It will only log errors.  Not throwing exceptions because error message is more useful on the CLI.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: we construct the key explicitly so it will not be null.

zhangn49: Sorry for the confusion, I mean "table.get(dynamicKey)" might be null?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: Updated the comment with an explanation.  Let me know what you think.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: the new behavior now returns the edited CorfuDynamicRecord, null if the edit was not successful or key not found. Not throwing exceptions because error message is more useful on the CLI.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: done.  using table.containsKey().

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: the new behavior now returns the edited CorfuDynamicRecord, null if the edit was not successful or key not found.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: cleaned the verifyNamespace and verifyNamespaceAndTablename methods and unified them.  But I think it is valuable to print which param is null or empty

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: yes, done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -252,20 +263,107 @@ public int printAllProtoDescriptors() {
...
+                keyToEdit);
+        DynamicMessage newValueMsg =
+            dynamicProtobufSerializer.createDynamicMessageFromJson(defaultValueAny,
+            newRecord);
+
+        if (keyMsg == null || newValueMsg == null) {
+            return null;
+        }
+
+        runtime.getObjectsView().TXBegin();
```
zhangn49, line 332: Could the TX begin later? how about begin at line 339?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -252,20 +263,107 @@ public int printAllProtoDescriptors() {
...
+            } else {
+                log.info("Editing record with Key {} in table {} and " +
+                        "namespace {} with new record {}.  Stream Id {}.",
+                    keyToEdit, tableName, namespace, newRecord, streamUUID);
+                String payloadTypeUrl = oldRecord.getPayloadTypeUrl();
+                String metadataTypeUrl = oldRecord.getMetadataTypeUrl();
+                DynamicMessage metadata = oldRecord.getMetadata();
+                editedRecord = new CorfuDynamicRecord(payloadTypeUrl, newValueMsg,
+                    metadataTypeUrl, metadata);
+                table.put(dynamicKey, editedRecord);
```
zhangn49, line 356: Looks like the new record will have the same metadata as the old one, is it correct?
For example if the record has a revision metadata

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -38,10 +41,9 @@
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
-import static org.corfudb.runtime.view.TableRegistry.CORFU_SYSTEM_NAMESPACE;
-import static org.corfudb.runtime.view.TableRegistry.PROTOBUF_DESCRIPTOR_TABLE_NAME;
-import static org.corfudb.runtime.view.TableRegistry.getFullyQualifiedTableName;
-import static org.corfudb.runtime.view.TableRegistry.getTypeUrl;
+import com.google.protobuf.util.JsonFormat;
+
+import static org.corfudb.runtime.view.TableRegistry.*;
```
zhangn49, line 46: wild card imports

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
Maithem: Github comment page is not reflected in the git commit message. Please add those details to the git commit message.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -83,27 +85,28 @@ public CorfuStoreBrowser(CorfuRuntime runtime) {
...
-        if (tableName == null) {
-            throw new IllegalArgumentException("Please specify --tablename");
+    private void verifyNamespaceAndTableName(String namespace,
+        String tableName, boolean tableNameValidationRequired) {
+        if (isEmptyOrNull(namespace)) {
+            throw new IllegalArgumentException("Namespace is empty or null");
+        }
+
+        if (tableNameValidationRequired && isEmptyOrNull(tableName)) {
+            throw new IllegalArgumentException("Tablename is empty or null");
```
Maithem, line 109: I know you didn't introduce this, but I think it should be removed. The parameter's should be validated once they are parsed from the command line and not "downstream". Move all these validations to `CorfuStoreBrowserMain` and just use `com.google.common.base.Preconditions.checkArgument` for validation.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -83,27 +85,28 @@ public CorfuStoreBrowser(CorfuRuntime runtime) {
...
-    /**
-     * Validate that the namespace is not null
-     * @param namespace
-     */
-    private static void verifyNamespace(String namespace) {
-        if (namespace == null) {
-            throw new IllegalArgumentException("Please specify --namespace");
-        }
+    private boolean isEmptyOrNull(String str) {
+        return (str == null || str.isEmpty());
```
Maithem, line 94: Not needed. See comment https://github.com/CorfuDB/CorfuDB/pull/2969/files#r663237174

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -198,18 +198,26 @@ public int listTables(String namespace)
         return numTables;
     }
 
+    private List<TableName> listTablesInNamespace(String namespace) {
+        return dynamicProtobufSerializer.getCachedRegistryTable().keySet()
+            .stream()
+            .filter(tableName -> namespace == null || tableName.getNamespace().equals(namespace))
+            .collect(Collectors.toList());
+    }
```
Maithem, line 206: Why do you use the cached registry and not the actual registry ? 

This is not robust, for example, imagine the logic of the DynamicProtobufSerializer is changed such that it does lazy loading, then this will return an incorrect view of the tables.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -252,20 +263,107 @@ public int printAllProtoDescriptors() {
...
+
+        runtime.getObjectsView().TXBegin();
+        CorfuTable<CorfuDynamicKey, CorfuDynamicRecord> table =
+            getTable(namespace, tableName);
+
+        CorfuDynamicKey dynamicKey =
+            new CorfuDynamicKey(defaultKeyAny.getTypeUrl(), keyMsg);
+
+        CorfuDynamicRecord editedRecord = null;
+        if (table.containsKey(dynamicKey)) {
```
Maithem, line 340: This doesn't look right. If it's a `newRecord` then the expectations is that it won't exist in the table. This means that if you are trying to create an object and it has never been created, then you can never create it.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -252,20 +263,107 @@ public int printAllProtoDescriptors() {
...
+            CorfuDynamicRecord oldRecord = table.get(dynamicKey);
```
Maithem, line 341: tx.begin and tx.end should be in a try/finally block

Maithem (approved):
Looks good!

Please address the remaining comments.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: In all cases I have come across, we want to modify the value as it was incorrectly written.  But you do raise a valid point if we should support editing the metadata.  

Maybe we can address it in a follow-up PR?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: this is a read-modify-write operation.  So getTable() must be done from within the transaction.  But I have moved it to later.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: DynamicProtobufSerializer is required for read/write operations to any tables.  

Before my change, some operations in the browser tool used ProtobufSerializer(actual registry) while others used the DynamicProtobufSerializer, making it difficult to understand and maintain the code.  To make it uniform for all browser operations, I have changed them all to use the DynamicaProtobufSerializer(and hence the cached registry).

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: addRecord will need some modification I think

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: moved to DynamicProtobufSerializer

zhangn49 (approved):

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
zhangn49: I just created a GitHub issue to track it

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserMain.java
pankti-m: I created a new issue https://github.com/CorfuDB/CorfuDB/issues/2992 for it and plan to address it there because addRecord also needs metadata as an input.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: done

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
pankti-m: https://github.com/CorfuDB/CorfuDB/issues/2992

runtime/proto/corfu_store_metadata.proto
```diff
@@ -19,13 +19,18 @@ message TableName {
...
+// Describe the protos in a file.
 message TableDescriptors {
     // File Descriptor protobufs map.
     // Contains file descriptors of key, value, metadata and their dependencies.
     map<string, google.protobuf.FileDescriptorProto> fileDescriptors = 1;
 }
 
+message KeyValueDefaultRecord {
+    google.protobuf.Any key = 1;
+    google.protobuf.Any value = 2;
```
hisundar, line 31: What about metadata? the vast majority of our users deploy with Key, Value and Metadata right?
Also if you need this information can we just add these fields into the TableDescriptors message above as additional fields. Doing so will eliminate the need to have a separate corfu table, it will also reduce the memory footprint since we only have to cache the table names twice (TableRegistry and ProtobufDescriptorTable)?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: instead of the separate table can we just add the new fields into the TableDescriptors message itself? This has an advantage of serving as a schema registry in just one table. It will also simplify a lot of queries in the future where the schema needs to be looked up?

hisundar (commented):
Thanks @pankti-m this is a nice change. But given how much we are struggling with memory footprint, we really need to reconsider the use of a separate corfuTable. 
Please let me know if my comments make sense.

runtime/proto/corfu_store_metadata.proto
pankti-m: done

runtime/proto/corfu_store_metadata.proto
```diff
@@ -19,11 +19,16 @@ message TableName {
...
 
-// Descriptor to provide self describing schemas for the records.
+// Describe the protos in a file.
 message TableDescriptors {
     // File Descriptor protobufs map.
     // Contains file descriptors of key, value, metadata and their dependencies.
     map<string, google.protobuf.FileDescriptorProto> fileDescriptors = 1;
+
+    // 'Any' representing the key, value and metadata of a table
+    google.protobuf.Any key = 2;
```
hisundar, line 29: Thanks for this change. However this means that when we validate the schema we might have to force write an update into the TableRegistry even when we don't detect a direct change in the list of file descriptors, if the previous value did not have any key, value or metadata fields. This is only needed for upgraded setups.
Please look for this line... `// else no need to re-add the exact same entry into the table`

hisundar (approved):
This looks good to me, there is one thing we need to address with the new addition of fields in the TableDescriptors. If the editor has been tested on a live setup, please feel free to merge it and address the upgrade request in a separate PR or merge it into this PR if you wish.

runtime/proto/corfu_store_metadata.proto
pankti-m: done

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -245,31 +252,33 @@ void registerTable(@Nonnull String namespace,
...
+                if (oldRecord != null && !oldRecord.getPayload().getFileDescriptorsMap().keySet()
+                    .equals(tableDescriptors.getFileDescriptorsMap().keySet())) {
+                    log.warn("Protobuf file list changed for {} table {}. Prev list {}, new list {}",
+                        namespace, tableName, oldRecord.getPayload().getFileDescriptorsMap().keySet(),
+                        tableDescriptors.getFileDescriptorsMap().keySet());
+                    for (StackTraceElement st : Thread.currentThread().getStackTrace()) {
+                        log.debug("{}", st);
+                    }
                 }
+                this.registryTable.put(tableNameKey,
```
hisundar, line 278: Let's try to avoid an unconditional IO since this is only needed if oldRecord != null && oldRecord.getPayload().getKey() == null
Basically only when the old record did not have any default instances recorded AKA upgrade

hisundar (changes requested):

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
pankti-m: but oldRecord.getPayload().getKey() will not be null.  It will be a default Any message, right?

hisundar (approved):
Thanks ! Looks good to me!

PR #2970, opened 2021-05-26
Polling Task IMPL + Unit Tests

zhangn49:
## Overview

Description:

init commit

Why should this be merged: 

- [ ] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/compaction/PollingTask.java
```diff
@@ -0,0 +1,259 @@
...
+    private final CorfuStore corfuStore;
+
+    private final Table<StatusKeyMsg, StatusMsg, Message> statusTable;
+
+    private final CorfuQueue checkpointTaskQueue;
+
+    private final CheckpointTaskProcessor taskProcessor;
+
+    // The Thread pool for executing polling tasks.
+    private final ScheduledExecutorService pollingExecutor;
```
Lujie1996, line 55: This executor is responsible for both polling and execution a checkpoint task, so perhaps we can name it more accurately. Also, why Scheduled? I saw `pollingExecutor.submit(this);` in onNext(), so the execution should be triggered by onNext() instead of being scheduled, right?

infrastructure/src/main/java/org/corfudb/infrastructure/compaction/PollingTask.java
```diff
@@ -0,0 +1,259 @@
...
+
+    private volatile SafeSnapshot safeSnapshot;
+
+    private volatile String compactionTaskId;
+
+    public PollingTask(CorfuRuntime runtime, CorfuStore corfuStore, String diskModePath,
```
Lujie1996, line 61: I think it might be better to take the pollingExecutor out of this PollingTask class. A runnable should be decoupled from the executor which is supposed to execute the runnable. In this way the polling task can focus on only the task logics itself and stay away from the execution logics. We can have one class which implements StreamListener and holds executor, and another class which implements Runnable and holds the task logics. What do you think?

Lujie1996 (commented):
Will review the rest of code changes later.

PR #2976, opened 2021-06-08
Fix Github secrets access

xcchang:
## Overview

Description: We didn't access GitHub secrets correctly in xml files. This commit provide a fix for this issue.

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/.github/workflows/publish-corfu.yml
+++ b/.github/workflows/publish-corfu.yml
@@ -1,4 +1,4 @@
-name: CorfDb Artifacts
+name: CorfuDB Artifacts
 
 on:
   push:
@@ -10,6 +10,11 @@ jobs:
 
     runs-on: ubuntu-latest
     if: github.event_name == 'push'
+
+    env:
+      JFROG_PASSWORD: ${{ secrets.jfrog_oss_password }}
+      DOCKER_USERNAME: ${{ secrets.docker_username }}
+      DOCKER_PASSWORD: ${{ secrets.docker_password }}
 
     steps:
       - uses: actions/checkout@v2
@@ -41,11 +46,3 @@ jobs:
       - name: Publish corfu artifacts
         working-directory: ./
         run: ./mvnw clean deploy -s ./.mvn/settings.xml -DskipTests -Dmaven.javadoc.skip=true -Dcheckstyle.skip -T 1C
-
-      - name: Publish Corfu infrastructure container
-        working-directory: ./
-        run: ./mvnw -pl infrastructure clean deploy -Pdocker -DskipTests -Dmaven.javadoc.skip=true -Dcheckstyle.skip -T 1C
-
-      - name: Publish Corfu generator container
-        working-directory: ./
-        run: ./mvnw -pl generator clean deploy -Pdocker -DskipTests -Dmaven.javadoc.skip=true -Dcheckstyle.skip -T 1C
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-corfu.yml
```diff
@@ -1,15 +1,16 @@
...
-  push:
-    branches:
-      - master
+name: CorfuDB Artifacts
+#
+#on:
+#  push:
+#    branches:
+#      - master
+on: pull_request
```
xnull, line 7: don't forget to change back to master, when your fix is done

xcchang: sure will do

generator/pom.xml
```diff
@@ -69,23 +69,23 @@
...
+                                        <argument>--username=${env.DOCKER_USERNAME}</argument>
+                                        <argument>--password=${env.DOCKER_PASSWORD}</argument>
                                     </arguments>
                                 </configuration>
                             </execution>
 
                             <execution>
                                 <id>Docker push</id>
-                                <phase>deploy</phase>
+                                <phase>validate</phase>
```
xnull, line 88: It seems to me, this should be "deploy", as I remember deploy step doesn't include validate step, also anyone who would like to execute: `mvnw install` will also deploy artifacts

xnull: Could you please explain why you decided to use validate instead of deploy?

.github/workflows/publish-corfu.yml
```diff
@@ -1,15 +1,20 @@
...
 jobs:
   publish_corfu_db_artifacts:
 
     runs-on: ubuntu-latest
-    if: github.event_name == 'push'
+#    if: github.event_name == 'push'
+    env:
+      JFROG_PASSWORD: ${{ secrets.jfrog_oss_password }}
+      DOCKER_USERNAME: ${{ secrets.docker_username }}
+      DOCKER_PASSWORD: ${{ secrets.docker_password }}
```
chetangudisagar, line 17: Do we really need this env change? Can we directly fetch from secrets in all the pom.xml files?

xcchang: As we discussed yesterday, that's the reason for this PR. Secrets cannot be directly accessed in other xml files.

generator/pom.xml
xcchang: Yes you are right, validate doesn't work. I was trying to prevent the `-pl infrastructure deploy` from updating the JFrog repo, i.e. it should only do the executions in docker profile. Haven't found a good way to do so, maybe I will reach out to you for help today.

.github/workflows/publish-corfu.yml
chetangudisagar: Ohh yeah, thank you!

xnull (approved):

zhangn49 (approved):

.github/workflows/publish-corfu.yml
```diff
@@ -41,11 +46,3 @@ jobs:
       - name: Publish corfu artifacts
         working-directory: ./
         run: ./mvnw clean deploy -s ./.mvn/settings.xml -DskipTests -Dmaven.javadoc.skip=true -Dcheckstyle.skip -T 1C
```
chetangudisagar, line 48: Do we need to clean again? Can we reuse the previous command's jars?

chetangudisagar: We discussed offline and agreed that it is indeed required.

chetangudisagar (approved):
LGTM, Thanks!

Issue #2977, opened 2021-06-08
CorfuServer logs spew stream UUIDs instead of Names making debugging hard


hisundar:
## Overview
CorfuServer's handleLogWrite logs useful streams that were modified but it uses UUID which are not human readable making debugging the corfu.9000.logs extremely cumbersome.
Logging stream names is much more useful since we can directly pinpoint the table(s) being written to.

PR #2978, opened 2021-06-08
CorfuServer: Map stream UUIDs to stream names to make log readable

hisundar:
## Overview

Description:

Make corfu.9000.log log stream names instead of UUIDs to aid debugging.
Since stream names are not available in server side, accept a file path which will carry a comma separated list of stream name -> stream-id which the server can then use to display meaningful stream names whenever possible

Why should this be merged: 

Related issue(s) (if applicable): #2977 

- [x] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -162,6 +163,10 @@
                     + "              If this limit is exceeded "
                     + "              write requests will be rejected [default: 100.0].\n         "
                     + "                                                                          "
+                    + " --stream-names-file=<stream_names_file>                                  "
```
Maithem, line 166: This requires too much manual labor and is not robust to an out dated stream file name. 

IMO the easiest and cleanest way of doing this is to send the table names as metadata in the log write request instead of reverse looking uuid->name based on a local file. This adds a negligible overhead to the write request that can be removed later with a proper stream registry that is replicated.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -283,9 +283,13 @@ private void handleException(Throwable ex, ChannelHandlerContext ctx, RequestMsg
     @RequestHandler(type = PayloadCase.WRITE_LOG_REQUEST)
     private void handleWrite(RequestMsg req, ChannelHandlerContext ctx, IServerRouter router) {
         LogData logData = getLogData(req.getPayload().getWriteLogRequest().getLogData());
-
         log.debug("handleWrite: type: {}, address: {}, streams: {}",
-                logData.getType(), logData.getToken(), logData.getBackpointerMap());
+                logData.getType(), logData.getToken(),
+                logData.getBackpointerMap().entrySet().stream().collect(Collectors.toMap(k ->
+                                serverContext.getStreamNames().getOrDefault(k, k.toString()),
+                        Map.Entry::getValue) // Re-map to stream name -> long from UUID -> long
```
Maithem, line 290: We should keep at least a portion of the UUID since that's what's used throughout the system (i.e., sequencer logs, LogUnit startup logs, etc...).

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -496,4 +496,36 @@ private void printStreamTagMap(String tag, List<TableName> tables) {
...
         log.info("tag: <'{}'> --- Tables[total={}] :: {{}}", tag, tables.size(), formatMapping);
     }
+
+    /**
+     * List all known corfu streams with their names and ids in a cluster.
+     *
+     * @return stream tags
+     * */
+    public Set<String> listAllCorfuStreams() {
+        Map<UUID, String> streamTags = new HashMap<>();
```
pankti-m, line 506: nit - can we call it something else.. streamInfoMap?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -179,6 +186,35 @@ public ServerContext(Map<String, Object> serverConfig) {
...
+        streamNames = loadStreamIdToNameMap((String) serverConfig.get("--stream-names-file"));
+    }
+
+    public static Map<UUID, String> loadStreamIdToNameMap(String streamNameFile) {
+        Map<UUID, String> streamNames = new HashMap<>();
+        if (streamNameFile == null) {
+            return streamNames;
+        }
+        try {
+            BufferedReader csvFile = new BufferedReader(new FileReader(streamNameFile));
```
xnull, line 198: Can we at least use JSON? Since JSON is much easier to work with and we already have JSON parser in our code

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -179,6 +186,35 @@ public ServerContext(Map<String, Object> serverConfig) {
...
         localEndpoint = nodeLocator.toEndpointUrl();
+        streamNames = loadStreamIdToNameMap((String) serverConfig.get("--stream-names-file"));
+    }
+
+    public static Map<UUID, String> loadStreamIdToNameMap(String streamNameFile) {
+        Map<UUID, String> streamNames = new HashMap<>();
+        if (streamNameFile == null) {
+            return streamNames;
+        }
+        try {
```
xnull, line 197: Resource leaks, FileReader and BufferREader won't be closed at all

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -179,6 +186,35 @@ public ServerContext(Map<String, Object> serverConfig) {
         nodeLocator = NodeLocator
                 .parseString(serverConfig.get("--address") + ":" + serverConfig.get("<port>"));
         localEndpoint = nodeLocator.toEndpointUrl();
+        streamNames = loadStreamIdToNameMap((String) serverConfig.get("--stream-names-file"));
+    }
+
+    public static Map<UUID, String> loadStreamIdToNameMap(String streamNameFile) {
+        Map<UUID, String> streamNames = new HashMap<>();
+        if (streamNameFile == null) {
```
xnull, line 194: what if 'streamNameFile' is an empty string

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
hisundar: I initially tried this too - it's not a trivial change - I was trying to modify the serialize() function to add the tableName, wasn't sure about the compatibility and chickened out. This is simple, it works and is certainly better than logging UUIDs everywhere.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
hisundar: Why not the other way around - use the table name throughout the system. Humans don't understand UUID and if we are logging UUIDs that's already painful debugging by design.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -179,6 +186,35 @@ public ServerContext(Map<String, Object> serverConfig) {
...
+                if (row == null) {
+                    break;
+                }
+                String[] tokens = row.split(",");
+                final int expectedColumns = 2;
+                if (tokens.length != expectedColumns) {
+                    log.warn("Bad line found in streamNameFile {}", row);
+                    break;
+                }
+                UUID streamId = UUID.fromString(tokens[expectedColumns - 1]);
```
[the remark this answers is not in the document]
hisundar: check style flags it as magic number!

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -179,6 +186,35 @@ public ServerContext(Map<String, Object> serverConfig) {
...
+            do {
+                String row = csvFile.readLine();
+                if (row == null) {
+                    break;
+                }
+                String[] tokens = row.split(",");
+                final int expectedColumns = 2;
+                if (tokens.length != expectedColumns) {
+                    log.warn("Bad line found in streamNameFile {}", row);
+                    break;
```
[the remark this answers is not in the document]
hisundar: good point, will do that. thanks

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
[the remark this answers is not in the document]
hisundar: done. fyi it looks like this
CorfuSystem$RegistryTable#chkpt,12b4622e-4b6a-37b6-9e5d-ce1944708eca
CorfuSystem$RegistryTable,22932286-9b8f-3469-8a49-816320df0f2a

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -496,4 +496,36 @@ private void printStreamTagMap(String tag, List<TableName> tables) {
...
+        runtime.getTableRegistry().getRegistryTable().entrySet().forEach(entry -> {
+            String namespace = entry.getKey().getNamespace();
+            String tableName = entry.getKey().getTableName();
+            String fullTableName = TableRegistry.getFullyQualifiedTableName(namespace, tableName);
+            UUID streamId = CorfuRuntime.getStreamID(fullTableName);
+            streamTags.put(streamId, fullTableName);
+            streamTags.put(CorfuRuntime.getCheckpointStreamIdFromId(streamId), fullTableName+"#chkpt");
+            entry.getValue().getMetadata().getTableOptions().getStreamTagList().forEach(tag -> {
```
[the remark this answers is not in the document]
hisundar: this tool is only used by advanced users of corfu so I thought its useful to have.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
```diff
@@ -496,4 +496,36 @@ private void printStreamTagMap(String tag, List<TableName> tables) {
...
+                streamTags.put(TableRegistry.getStreamIdForStreamTag(namespace, tag), "stream_tag#"+namespace+"$"+tag);
+            });
+        });
+
+        List<Map.Entry<UUID, String>> list = new ArrayList<>(streamTags.entrySet());
+        list.sort(Map.Entry.comparingByValue());
+
+        log.info("\n======================\n");
+        log.info("Total unique stream Ids: "+streamTags.size());
```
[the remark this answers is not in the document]
hisundar: for CorfuStore the namespace is already part of the logged output - the table names are prefixed with the namespace right?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowser.java
hisundar: done thanks

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
hisundar: What capability are we trying to achieve in addition to logging that requires JSON? Once that feature which needs JSON is  added we can certainly log JSON. Right now csv produces the simplest readable stream name mapping file and easy to work on using linux tools.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
hisundar: done. Thanks!

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
hisundar: there will be an exception thrown and the map will be initialized to an empty map, which is the default case for all our internal ITs and unit tests.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
xnull: Exactly the same question could be asked regarding csv.
Working with JSON much simpler and easier, than a bunch of meaningless numbers separated by a semicolon and operating with index numbers, also no need for custom code with leaking resources

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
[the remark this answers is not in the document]
xnull: So, in addition to my previous comment regarding csv.
Please comapre what is more convenient:
 - CorfuSystem$RegistryTable#chkpt,12b4622e-4b6a-37b6-9e5d-ce1944708eca
    CorfuSystem$RegistryTable,22932286-9b8f-3469-8a49-816320df0f2a
    
   or 
 ```
 schema: [
  {
   "table": "CorfuSystem$RegistryTable#chkpt",
   "stream": "12b4622e-4b6a-37b6-9e5d-ce1944708eca"
 },
 {
   "table": "CorfuSystem$RegistryTable#chkpt",
   "stream": "12b4622e-4b6a-37b6-9e5d-ce1944708eca"
 }
]
 ```
 
 also, what if you'd like to add more columns? Compare 
 tableName#chkpt,uuid,123,321,insert,555,"medium",8671233,${uuid}
 with a json version, where you can see the entire schema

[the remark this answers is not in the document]
hisundar: the table name and stream uuid pairs are generated by a combination of log scrapping and tools. Dumping json output is not easy to do with linux scrapping cli commands.
You still did not answer my question :) Given that we are not adding any new columns today, what is the feature that *requires* JSON for this mapping table?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
xnull: >Not really it's not easy to generate correct JSON by scrapping logs from different sources like client and compactor

how it can work differently for two formats? You do whatever you need to do doesn't matter what the format of your document, and save and read data from JSON is more convenient than from CSV.

>scrapping logs
sounds like a wrong approach. Are you going to do automation stuff by parsing logs by grep commands?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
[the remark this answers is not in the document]
xnull: do you really want to make your automation on top of csv and command line?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
hisundar: can you please share the method to generate correct JSON with streamId -> streamName map without log-scraping while accounting for
1) pseudo-streams like TRANSACTION_STREAM
2) various stream_tags, checkpoint streams
3) ephemeral streams 
4) Internal system tables
5) Log Replicated Streams
from *ALL* the verticals consuming Corfu?

xnull: the same way how you do that using csv? 
How the file format changes the requirements? Don't you need to do all that magically, if you'd use csv instead of json?

hisundar: I am using a combination of log scrapping and tooling (part of this PR) currently but if you say that is wrong approach, can you please share the correct approach?

xnull: Again, I'm talking about the file format: `saveToCsv(file)` and `saveToJson(file)` I don't see how `scrappingLogs()` can affect that part

xnull: Your csv file is just a config, created manually, correct?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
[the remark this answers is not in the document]
hisundar: No, not in the long run, once all the verticals migrate to using CorfuStore, we can directly determine the stream names from the table registry itself. But until then we need debugging help to triage which verticals are still slow to migrate and who is still using the old api. This debugging change is critical to that.
CSV vs JSON is a separate topic. Sure JSON has some advantages but those advantages are not employed in this PR.
For this PR and this current use case a CSV is concise, efficient and simple to generate. 
If I need to re-write the streamName file in JSON I have to write another tool to do post processing - read csv -> re-write in json which can be done, but I don't see the pressing functional requirement to do that extra step.

hisundar: "Why is it better than using UUIDs? that point is not clear to me, most of my debugging I almost never need to know the table name."
Most mortals like me cannot mentally translate thousands of 32 byte hexadecimal strings into their parent table names. Can we please stop pretending that UUIDs are more useful than human readable tableNames in log files?

"The problem with this PR is it can produce inconsistent logs within a server and across servers, this will only make it harder to debug. This approach also introduces manual labor that is not necessary and should be automated."
Every location within CorfuServer should be changed to log tableName instead of UUID, if we are logging a mix of UUID and tableName that's a bug and should be fixed. Do you see any location where UUID is logged in spite of tableName please point that out in your review comments, I will fix it right away. Also not sure what you mean by manual labor? The work to map names to UUIDs is likely something done once and reused many times.

"Generating a CSV zfile from the client side, then copying that file to all nodes of the cluster is not really efficient. Imagine that table mapping changes after it is generated, now the server will produce logs for some tables and UUIDs for others, its not consistent."
The alternative you propose sending table names in addition to stream UUIDs is far more inefficient especially when the UUID is quite literally derived from the tableName. We should be using either one, definitely not both. But that's a more bigger change.

"You mentioned that you need this to for some scripts that you wrote, but if you already have that mapping why cant you write a script that remaps the UUIDs to names, if you really need it. I think that's much cleaner than requiring the server to do this mapping because of the aforementioned points."

Because I am not the only person debugging, there are several engineers who want to triage using corfu.9000.log and simply cannot figure out a way to map UUIDs to table names. This is meant to help them, not me. We do have scripts but it slows down debugging if someone has to run a script to analyze the log. Also browsing log files via UI is now possible.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
hisundar: Yes, what you are suggesting is do another step of post-processing to read CSV and then re-write the tableNames files as JSON.
I will start working on that, but I would like to still understand what is the benefit for this specific use case and not something in the future?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
Maithem: > Most mortals like me cannot mentally translate thousands of 32 byte hexadecimal strings into their parent table names. Why are we debating if UUIDs are more useful than tableNames in log files?

This is a red herring, it doesn't introduce new information and it doesn't answer my question. Let me distill my point of view further so that subsequent responses can be more focused and objective. 

First of all, this PR in its current form will break logging. Currently, on master all logging (except for two places) refers to the stream by the UUID. Your change only remaps the logunit server logs. In essence, this means that logs produced from the clients and other servers cannot be correlated with the logunit server logs. Therefore this PR would degrade the current experience of debugging the system. 


Second, you are making a false assumption that a human readable reference would add more information to assist debugging. Why do you think that tracing a string like "TableA" is easier than tracing a string like "39d4" ? Both are just references and a UUID is more succinct than an arbitrary length string which makes it cleaner to log. Furthermore, unlike a UUID, a string is not specified, so logging a string with special characters can make it really hard to read the log file. 

Of course at some point in the loop, a human readable string is required (for example inspecting metrics logs), but my argument is that step is usually minor step in the debugging process and isn't a good enough reason convert all logging statements from UUID to an arbitrary length string, that is, It's not a better trade-off.


Third, there is a bootstrap problem. This PR's "feature" requires clients to generate the mapping beforehand, but clients can open streams and generate names/ids at runtime after connecting to the server. This means that the mapping generated is not accurate and needs to be consistently updated across all nodes. Furthermore, this also requires to the update mapping on all nodes in the cluster at the "right time" otherwise some servers would log UUIDs and some servers would log the table name string. This makes logging inconsistent. 


When making claims like:

> Every location within CorfuServer should be changed to log tableName instead of UUID

and 

> This is meant to help them, not me

please make sure to supplement your points with a technical argument, or a use case. 

I also find it a little ironic that you reject my anecdotal evidence of debugging 100s of bugs, but use yours as a main point.

hisundar: > > Most mortals like me cannot mentally translate thousands of 32 byte hexadecimal strings into their parent table names. Why are we debating if UUIDs are more useful than tableNames in log files?
> 
> This is a red herring, it doesn't introduce new information and it doesn't answer my question. Let me distill my point of view further so that subsequent responses can be more focused and objective.
> 
> First of all, this PR in its current form will break logging. Currently, on master all logging (except for two places) refers to the stream by the UUID. Your change only remaps the logunit server logs. In essence, this means that logs produced from the clients and other servers cannot be correlated with the logunit server logs. Therefore this PR would degrade the current experience of debugging the system.
>

Why are you commenting on an older commit? Please see the latest commit's files before commenting if you desire to be helpful - I have changed Sequencer logs as well. Most importantly this does not "break logging". Logging is meant *for human beings* first, not machines. Logging something that isn't human readable *is broken*! Also what do you mean by "other servers" are you referring to state transfer? If so please share the exact location where a stream UUID is logged without that your comment is not helpful at all.

> Second, you are making a false assumption that a human readable reference would add more information to assist debugging. Why do you think that tracing a string like "TableA" is easier than tracing a string like "39d4" ? Both are just references and a UUID is more succinct than an arbitrary length string which makes it cleaner to log. Furthermore, unlike a UUID, a string is not specified, so logging a string with special characters can make it really hard to read the log file.

Who exactly cares about succinct or variable lengths in logs? No human who looks at _22932286-9b8f-3469-8a49-816320df0f2a_ will know that this is really the _CorfuSystem$RegistryTable_ so being succinct or fixed length means absolutely nothing and adds ZERO value to the log. Also who exactly is adding "special" characters to table name? Can you please share a real concrete example before throwing a non-existent hypothetical?

> 
> Of course at some point in the loop, a human readable string is required (for example inspecting metrics logs), but my argument is that step is usually minor step in the debugging process and isn't a good enough reason convert all logging statements from UUID to an arbitrary length string, that is, It's not a better trade-off.

Adding steps to debugging is wrong and should be avoided whenever possible. Humans make mistakes and when they do we want to be able to get to the root cause in the FASTEST way possible.

> 
> Third, there is a bootstrap problem. This PR's "feature" requires clients to generate the mapping beforehand, but clients can open streams and generate names/ids at runtime after connecting to the server. This means that the mapping generated is not accurate and needs to be consistently updated across all nodes. Furthermore, this also requires to the update mapping on all nodes in the cluster at the "right time" otherwise some servers would log UUIDs and some servers would log the table name string. This makes logging inconsistent.

This is a valid point and the correct solution is to have a RPC to the server whenever a new client stream is opened which the server can store and use for things like logging. We can certainly debate on the best approach for solving this problem for generic arbitrary names but that change can go **in addition** to this current approach which works and is certainly better than status quo.

> 
> When making claims like:
> 
> > Every location within CorfuServer should be changed to log tableName instead of UUID
> 
> and
> 
> > This is meant to help them, not me
> 
> please make sure to supplement your points with a technical argument, or a use case.
> 
> I also find it a little ironic that you reject my anecdotal evidence of debugging 100s of bugs, but use yours as a main point.

My point, call it technical, philosophical or just practical, is **Logs are meant for Humans first**. Your arguments are all about keeping logs "consistent", "fixed-length" or "concise" which I argue are completely pointless if the log is **unreadable** to begin with.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -54,6 +54,7 @@
                     + "[--base-server-threads=<base_server_threads>] "
                     + "[--log-size-quota-percentage=<max_log_size_percentage>]"
                     + "[--logunit-threads=<logunit_threads>] [--management-server-threads=<management_server_threads>]"
+                    + "[--stream-names-file=<stream_names_file>]"
```
zfrenette, line 57: Could we explore a more dynamic approach for the server to become aware of the names? Since this file is static, it requires being aware of the table names ahead of time. And in the case when the mapping does not contain the stream name, since the UUID will be logged anyway, we are left with this awkward combination of both formats. Also, although I realize that this isn't the case in NSX, there are no restrictions on the format of table names, so it could contain things like new lines or special characters that make the logs difficult to read. With a UUID, this problem isn't there.

hisundar: Yes that is the long term solution. This is a best effort next best thing in its absence. 
In order for the dynamic approach we have to essentially rethink how stream names map to stream Ids. currently the solution is to use `UUID.nameUUIDFromBytes(string.getBytes());` at the client side which won't work.
Instead we need a way to ship this stream name over to the server and have the server generate the stream id.
Doing this will also allow us to generate a much more concise streamId mapping instead of burning 16 bytes.
this approach as you can see involves a lot more changes than the scope of this simple debugging fix PR.

test/src/test/java/org/corfudb/integration/CorfuStoreBrowserIT.java
```diff
@@ -225,6 +225,11 @@ public void testBrowserTagsOperations() throws Exception {
         assertThat(tags.size()).isEqualTo(expectedTableNameToTags.get(tableName).size());
         assertThat(tags).containsExactly(expectedTableNameToTags.get(tableName).toArray(new String[0]));
 
+        final int totalExpectedStreams = 30;
+        // (5) List All known streams and their UUIDs
+        Set<String> allStreams = browser.listAllCorfuStreams();
+        assertThat(allStreams.size()).isEqualTo(totalExpectedStreams);
+
```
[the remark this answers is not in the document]
hisundar: Sure will try.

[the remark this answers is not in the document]
hisundar: done.

vjeko (approved):

Maithem (changes requested):
This PR is half-backed and doesn't improve the debugging process, a proper solution should be implemented instead. Multiple reviewers have already pointed out that the configuration should be dynamic, and not based on static files generated and copied to the server nodes manually.

To recapitulate, the main points extracted from your responses (per verbatim) are:

1. This interim solution is being introduced because "This is meant to help them, not me."  

2. A proper solution is not being implemented because " it's not a trivial change ... wasn't sure about the compatibility and chickened out"

3. You believe the stream reference, which is a UUID is "unreadable" and must be mapped to a String in all debugging statements for it to be useful

Since this issue is not urgent and your main point is based on a false premise - that the UUID must have semantic meaning to be useful - I think we should implement a proper solution with a dynamic registry without having an intermediate solution. Having debugged literally several hundreds of bugs I can tell you that this patch will degrade the debugging experience.

hisundar:
> This PR is half-baked and doesn't improve the debugging process, a proper solution should be implemented instead. Multiple reviewers have already pointed out that the configuration should be dynamic, and not based on static files generated and copied to the server nodes manually.
> 
> To recapitulate, the main points extracted from your responses (per verbatim) are:
> 
> 1. This interim solution is being introduced because "This is meant to help them, not me."
> 2. A proper solution is not being implemented because " it's not a trivial change ... wasn't sure about the compatibility and chickened out"
> 3. You believe the stream reference, which is a UUID is "unreadable" and must be mapped to a String in all debugging statements for it to be useful
> 
> Since this issue is not urgent and your main point is based on a false premise - that the UUID must have semantic meaning to be useful - I think we should implement a proper solution with a dynamic registry without having an intermediate solution. Having debugged literally several hundreds of bugs I can tell you that this patch will degrade the debugging experience.

@Maithem - I hear you loud and clear but I respectfully and sincerely ask you to please reconsider. This change is really important and tremendously useful for debugging. This is how the log looks before this change...

`2021-07-16T08:35:46.731Z | DEBUG |  LogUnit-1 |  o.c.i.LogUnitServer | handleWrite: type: DATA, address: Token(epoch=0, sequence=4), streams: {8bffc026-951d-f73d-b246-88323d638c0c=2, 6d60a60-08fa-3faf-a0b3-43f4fa45ae2a=3, 22932286-9b8f-3469-8a49-816320df0f2a=2}
`
and this is how it looks with this change…

`2021-07-29T22:45:20.393Z | DEBUG | LogUnit-1 | o.c.i.LogUnitServer | handleWrite: type: DATA, address: Token(epoch=4, sequence=1663018), streams: {stream_tag#nsx$generic_table_status_redacted=1662637, namespace$HostTableName2Redacted=1662637, stream_tag#namespace$search_table_redacted=1662637, stream_tag#namespace$redactedTableName=1662637}`

Developers are now able use corfu.9000.log without involving our team needlessly in their debugging, saving us huge amounts of time. Is that not worth your reconsideration?

PR #2979, opened 2021-06-09
Clean up INVALID protobuf enum values

chetangudisagar:
## Overview

Description: 
In proto2, INVALID enum values were used to denote a default value to support backward compatibility. Since proto3 has a default UNRECOGNIZED enum value, we no longer need INVALID unless it is required for any other functionality.

Why should this be merged: Removes unnecessary code (enums and their relevant test cases).

- [ ] Public API has Javadoc


PavelZaytsev (approved):

infrastructure/proto/log_replication_cluster_info.proto
```diff
@@ -31,6 +31,5 @@ enum ClusterRole {
   INVALID = 0;
   ACTIVE = 1;
   STANDBY = 2;
-  NONE = 3;
```
zhangn49, line 34: This NONE is still used in NSX code

chetangudisagar: Ohh, I was wondering if we can merge INVALID and NONE as they might be serving a similar purpose. Can we do it or we need both? I found INVALID was used in many places within our codebase, and hence tried to replace it with NONE.

zhangn49: I think right now we need both, as they are corresponding roles of NSX side.

zhangn49: One more question about other enums - Does your change affect version compatibility?

Let's assume we have an old runtime has the INVALID = 0, and a new server that INVALID has been removed.
Now the new server has READY = 0, and receives a runtime request with INVALID. Does the new server know the request is INVALID/UNRECOGNIZED, or it will think the request is READY=0?

chetangudisagar: You are right. It will think the request is `READY=0`. The numbers assigned to the enum are given priority while serializing and de-serializing. 

I thought these changes are fine for now because from the next major release onwards, we will be supporting backward compatibility and we will have new compatibility vectors. Until then, we are good to make these breaking changes. It will work fine as long as the client and server are in the same version.

Also, in proto3, the First enum value must be 0, So we can't really keep the same numbers for the remaining values and just remove the `INVALID = 0` enum value.

PR #2982, opened 2021-06-14
CorfuStore: Return -1 as default for getHighestSequence()

hisundar:
Returning trim mark is problematic for idle tables
since trim makr moves on every checkpoint & trim cycle
impling additional data which is not true.

## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
@@ -261,9 +261,9 @@ public class CorfuStore {
         corfuStoreMetrics.recordBatchReads(numBatches);
         corfuStoreMetrics.recordNumberReads(numBatches * runtime.getParameters().getHighestSequenceNumberBatchSize());
         corfuStoreMetrics.recordHighestSequenceNumberDuration(startTime);
-        log.debug("Stream[{}${}][{}] no DATA entry found. Highest sequence number corresponds to trim mark={}",
-                namespace, tableName, Utils.toReadableId(streamId), streamAddressSpace.getTrimMark());
-        return streamAddressSpace.getTrimMark();
+        log.debug("Stream[{}${}][{}] no DATA entry found. Returning -1.",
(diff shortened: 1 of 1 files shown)


annym (approved):

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -261,9 +261,9 @@ public long getHighestSequence(@Nonnull final String namespace,
         corfuStoreMetrics.recordBatchReads(numBatches);
         corfuStoreMetrics.recordNumberReads(numBatches * runtime.getParameters().getHighestSequenceNumberBatchSize());
         corfuStoreMetrics.recordHighestSequenceNumberDuration(startTime);
-        log.debug("Stream[{}${}][{}] no DATA entry found. Highest sequence number corresponds to trim mark={}",
-                namespace, tableName, Utils.toReadableId(streamId), streamAddressSpace.getTrimMark());
-        return streamAddressSpace.getTrimMark();
+        log.debug("Stream[{}${}][{}] no DATA entry found. Returning -1.",
+                namespace, tableName, Utils.toReadableId(streamId));
+        return -1;
```
Maithem, line 266: How do you make sure this won't regress in future patches? Should we just assume that it wont, or should we have a unit test and enforce it in code?

hisundar: My bad, added a test and removed it temporarily to triage flaky tests and forgot to add it back..
going to push it in a separate patch.

hisundar: here is the test case https://github.com/CorfuDB/CorfuDB/pull/2966/files#diff-14edd36945df0ba1730f3db2fbfac31222a9d10e289513942c1b1f63f7cdd768R1111

PR #2983, opened 2021-06-14
More granular reads and writes

PavelZaytsev:
## Overview

Description:

Distinguish address space reads and writes by a stream


xnull (approved):

hisundar (changes requested):
Do you want to remove the readTxn writeTxn readWriteTxn metrics from UFO as part of this change since they achieve the same thing or keep those and make these generic?

hisundar (approved):

xnull (approved):
looks good

PR #2989, opened 2021-06-30
File Based Server Configuration Support

nvaishampayan517:
## Overview

Description:
Adds support for loading server configuration from a configuration file using the Java Properties Class. Changes ServerContext to utilize a new ServerConfiguration object to store config options over the previous Docopts map. Backwards compatible with previous CLI options.

Why should this be merged: 
Adds functionality to use configuration files for Corfu Server, and sets up framework for moving away from DocOpts CLI parsing.

- [ ] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/BaseServer.java
```diff
@@ -42,7 +42,7 @@
 
     public BaseServer(@Nonnull ServerContext context) {
         serverContext = context;
-        executor = serverContext.getExecutorService(serverContext.getBaseServerThreadCount(), "baseServer-");
```
nvaishampayan517, line 45: @Maithem are the thread prefixes always hardcoded, or should they use the getPrefix option from CLI args

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -667,16 +628,10 @@ public Layout copyManagementLayout() {
...
-        final int numThreads = requestedThreads == 0
-                ? Runtime.getRuntime().availableProcessors() * 2
-                : requestedThreads;
-        EventLoopGroup group = getChannelImplementation().getGenerator()
-            .generate(numThreads, threadFactory);
+        EventLoopGroup group = configuration.getChannelImplementation().getGenerator().generate(configuration.getNumIOThreads(), threadFactory);
 
         log.info("getWorkerGroup: Type {} with {} threads",
-                group.getClass().getSimpleName(), numThreads);
+                group.getClass().getSimpleName(), configuration.getNumIOThreads());
```
nvaishampayan517, line 634: @Maithem is this the correct Thread type

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -690,16 +645,10 @@ public Layout copyManagementLayout() {
...
-        final int numThreads = requestedThreads == 0
-            ? Runtime.getRuntime().availableProcessors() * 2
-            : requestedThreads;
-        EventLoopGroup group = getChannelImplementation().getGenerator()
-            .generate(numThreads, threadFactory);
+        EventLoopGroup group = configuration.getChannelImplementation().getGenerator().generate(6, threadFactory);
 
         log.info("getClientGroup: Type {} with {} threads",
-            group.getClass().getSimpleName(), numThreads);
+            group.getClass().getSimpleName(), 6);
```
nvaishampayan517, line 651: @Maithem this hardcode was in the previous commit, will revert to master during cleanup

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,667 @@
...
+        conf.setLogLevel((String) opts.getOrDefault("--log-level", "INFO"));
+        conf.setCompactRate(Integer.parseInt((String) opts.getOrDefault("--compact", "60")));
+        conf.setPluginConfigFilePath((String) opts.get("--plugin"));
+        conf.setNumBaseServerThreads(Integer.parseInt((String) opts.get("--base-server-threads")));
+        conf.setLogSizeQuota(Double.parseDouble((String) opts.getOrDefault("--log-size-quota-percentage", "100.0")));
+
+        //TODO(NEIL): double check
+        conf.setNumLogUnitWorkerThreads(Integer.parseInt((String) opts.getOrDefault("--logunit-threads", "4")));
+        conf.setNumLayoutServerThreads(Integer.parseInt((String) opts.getOrDefault("--management-server-threads", "4")));
+        conf.setNumIOThreads(Integer.parseInt((String) opts.getOrDefault("--Threads", "4")));
```
nvaishampayan517, line 159: @Maithem are these thread counts of the correct type?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -135,23 +130,23 @@ public LogUnitServer(ServerContext serverContext) {
      */
     public LogUnitServer(ServerContext serverContext, LogUnitServerInitializer serverInitializer) {
         this.serverContext = serverContext;
-        config = LogUnitServerConfig.parse(serverContext.getServerConfig());
-        executor = serverContext.getExecutorService(serverContext.getLogUnitThreadCount(), "LogUnit-");
+        executor = serverContext.getExecutorService(serverContext.getConfiguration().getNumLogUnitWorkerThreads(),
```
xnull, line 133: when you pass an object which has side effects and/or communicates to the external world or keeps some state, it affects your code and makes it harder to write tests to keep the code flexible and loosely coupled. 

That was the motivation for passing a config instead of a bigger and quite heavy serverContext object.

Next time when you'll need to write a unit test for LogUnitServer, it will be more complicated to do that if you'd use serverContext.

xnull: So, would be great to have only necessary dependencies in the classes

nvaishampayan517: got it, so capture snapshot of necessary attributes as the config object in the tests, and pass that state around?

xnull: In this case, probably you'll need to create an additional constructor, which seems outside of the scope.
Sorry, if there any confusion from my previous comments :) I thought LogUnitServer used a config in its constructor

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -598,20 +562,18 @@ StreamLog buildInMemoryStreamLog() {
             return new InMemoryStreamLog();
         }
 
-        StreamLog buildStreamLog(@Nonnull LogUnitServerConfig config,
-                                 @Nonnull ServerContext serverContext) {
-            return new StreamLogFiles(serverContext, config.isNoVerify());
+        StreamLog buildStreamLog(@Nonnull ServerContext serverContext) {
```
xnull, line 565: There are two "god" objects in Corfu: CorfuRuntime and ServerContext. It would be great to try to avoid using them. 
As I mentioned earlier that's why I have created those small immutable config files like LogUnitServerConfig. It helps keep code flexible

chetangudisagar:
Great PR! Also, make a note to fix the static code analysis errors reported by Codacy.
https://app.codacy.com/gh/CorfuDB/CorfuDB/pullRequest?prid=7686438.

nvaishampayan517:
> Great PR! Also, make a note to fix the static code analysis errors reported by Codacy.
> https://app.codacy.com/gh/CorfuDB/CorfuDB/pullRequest?prid=7686438.

Are all Codacy warnings required to be changed? For the LogUnitServerTest, I've left the config class in so that future tests can use the lightweight class to pass around parameters. Is it sufficient to just comment out those lines?

xcchang (commented):
Looks great! Left some minor comments. I have another concern is I noticed lots of usage with this pattern `servercontext.getConfiguration.get____`. As @xnull mentioned, it might be too heavy and could have potential issues to pass ServerContext everywhere. Some of them are from legacy code and beyond the scope of this PR, so might be good to optimize this in a future PR I think.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
xcchang: For unit tests, as those in `infrastructure/LogUnitServerTest`, we already decoupled it using mocks, so might not be a problem for adding more unit tests. But yeah I agree that passing ServerContext is too heavy and might have potential problems.

infrastructure/src/main/java/org/corfudb/infrastructure/LayoutServer.java
```diff
@@ -68,6 +70,7 @@
 //TODO Finer grained synchronization needed for this class.
 //TODO Need a janitor to cleanup old phases data and to fill up holes in layout history.
 @Slf4j
+@NotThreadSafe
```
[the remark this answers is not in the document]
nvaishampayan517: I believe that maithem made the class be single threaded execution to avoid the overhead in having all of its methods being synchronized. Thus the class would not be threadsafe since the methods are not monitored.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -594,27 +594,25 @@ public static LogUnitServerConfig parse(Map<String, Object> opts) {
...
-                                 @Nonnull ServerContext serverContext) {
-            return new StreamLogFiles(serverContext, config.isNoVerify());
+        public StreamLog buildStreamLog(@Nonnull ServerContext serverContext) {
+            return new StreamLogFiles(serverContext);
         }
 
-        LogUnitServerCache buildLogUnitServerCache(@Nonnull LogUnitServerConfig config,
-                                                   @Nonnull StreamLog streamLog) {
-            return new LogUnitServerCache(config, streamLog);
+        public LogUnitServerCache buildLogUnitServerCache(@Nonnull StreamLog streamLog,
```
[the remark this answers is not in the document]
nvaishampayan517: This change was to fix the static codacy issues, which didn’t like the package private default scope.

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,717 @@
...
+
+    public ServerConfiguration setThreadPrefix(String threadPrefix) {
+        setProperty(THREAD_PREFIX, threadPrefix);
+        return this;
+    }
+
+    public String getThreadPrefix() {
+        return getString(THREAD_PREFIX, DEFAULT_THREAD_PREFIX);
+    }
+
```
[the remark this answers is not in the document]
nvaishampayan517: I have this method for consistency with the current CLI options, which has an option to specify thread prefix.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -303,42 +292,41 @@ private static void startServer(Map<String, Object> opts) {
      * - pick the correct logging level before outputting error messages
      * - add serverEndpoint information
      *
-     * @param opts command line parameters
+     * @param conf Server Configuration parameters
      * @throws JoranException logback exception
```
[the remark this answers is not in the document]
nvaishampayan517: looking into the source of the imports, I don't see where this exception can be thrown, so I'll remove it.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -303,42 +292,41 @@ private static void startServer(Map<String, Object> opts) {
...
 
-        File serviceDir = new File((String) opts.get("--log-path"));
-
-        if (!serviceDir.isDirectory()) {
-            log.error("Service directory {} does not point to a directory. Aborting.",
-                    serviceDir);
-            throw new UnrecoverableCorfuError("Service directory must be a directory!");
+        File parentDir = new File(conf.getServerDir());
+        if (!parentDir.isDirectory()) {
+            throw new UnrecoverableCorfuError("Service path " + conf.getServerDir() + " must be a directory!");
```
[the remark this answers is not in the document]
nvaishampayan517: Sounds good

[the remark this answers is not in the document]
nvaishampayan517: Sounds good

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -572,18 +572,18 @@ void prefixTrim(long trimAddress) {
...
 
             return LogUnitServerConfig.builder()
                     .cacheSizeHeapRatio(cacheSizeHeapRatio)
                     .maxCacheSize((long) (Runtime.getRuntime().maxMemory() * cacheSizeHeapRatio))
-                    .memoryMode(Boolean.parseBoolean(opts.get("--memory").toString()))
-                    .noVerify((Boolean) opts.get("--no-verify"))
-                    .noSync((Boolean) opts.get("--no-sync"))
+                    .memoryMode(conf.isInMemoryMode())
+                    .noVerify(!conf.getVerifyChecksum())
+                    .noSync(!conf.getSyncData())
```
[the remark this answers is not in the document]
nvaishampayan517: The idea was to have the class available for future PRs in which we could optimize out the serverContext args.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -303,42 +292,42 @@ private static void startServer(Map<String, Object> opts) {
...
+    private static void createServiceDirectory(ServerConfiguration conf) {
+        if (conf.isInMemoryMode()) {
             return;
         }
 
-        File serviceDir = new File((String) opts.get("--log-path"));
-
-        if (!serviceDir.isDirectory()) {
+        File parentDir = new File(conf.getServerDir());
+        if (!parentDir.isDirectory()) {
```
xnull, line 314: `isFile()` ?

xnull: Would be great if you'd use `Path` instead of file

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -155,49 +151,32 @@
...
 
         // Setup the netty event loops. In tests, these loops may be provided by
         // a test framework to save resources.
-        final boolean providedEventLoops =
-                 getChannelImplementation().equals(ChannelImplementation.LOCAL);
 
-        if (providedEventLoops) {
-            clientGroup = getServerConfig(EventLoopGroup.class, "client");
-            workerGroup = getServerConfig(EventLoopGroup.class, "worker");
+        if (conf.getChannelImplementation().equals(ChannelImplementation.LOCAL)) {
```
xnull, line 165: please use `==` instead of equals

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
nvaishampayan517: I don't think isFile() is appropriate here, because we only want to succeed if the parentDir exists and is a directory. If we used isFile() instead of !isDir(), we would succeed even if the dir does not exist.

Also, the Path class doesn't provide isDir() method to check the condition.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServer.java
```diff
@@ -1,14 +1,13 @@
 package org.corfudb.infrastructure;
 
-import static org.corfudb.util.NetworkUtils.getAddressFromInterfaceName;
-
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.core.joran.spi.JoranException;
```
zfrenette, line 6: This appears unused now.

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,722 @@
...
+
+
+
+    public static ServerConfiguration getServerConfigFromFile(String configFilePath) {
+        FileInputStream configFileStream;
+        Properties configProperties = new Properties();
+
+        try {
+            configFileStream = new FileInputStream(configFilePath);
```
zfrenette, line 117: This can be wrapped in a try-with-resources block, or alternatively,  it can be closed explicitly in a finally clause.

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,722 @@
...
+            } else {
+                log.warn("Encountered unknown option: {}", key);
+            }
+        }
+
+        return applyServerConfigurationOptions(configProperties);
+    }
+
+    private static Map<String, String> getOptionsToPropertiesMapping() {
+        Map<String, String> mapping = new Hashtable<>();
```
zfrenette, line 234: Does this table need synchronization safety, or can it just be a `HashMap`?

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,722 @@
...
+    }
+
+    public int getHandshakeTimeout() {
+        return getInt(HANDSHAKE_TIMEOUT, Integer.parseInt(DEFAULT_HANDSHAKE_TIMEOUT));
+    }
+
+    public ServerConfiguration setMetadataRetention(int numFiles) {
+        if (numFiles < 1) {
+            throw new IllegalArgumentException("Max number of metadata files to retain must be greater than 0.");
+        }
```
zfrenette, line 363: Is it worth adding a similar check in other places in this class where we expect integer parameters to satisfy certain constraints?

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,722 @@
...
+    }
+
+    public static ServerConfiguration getServerConfigFromMap(Map<String, Object> opts) {
+        Properties configProperties = new Properties();
+        //in the future, if more formats need to be supported, this mapping can be passed as a parameter
+        Map<String, String> optionsToPropertiesMapping = getOptionsToPropertiesMapping();
+        Set<String> flags = ImmutableSet.of("--memory", "--single", "--enable-tls",
+                "--enable-tls-mutual-auth","--enable-sasl-plain-text-auth", "--metrics");
+        Set<String> inverseFlags = ImmutableSet.of("--no-verify", "--no-sync", "--no-auto-commit");
+        for (String key : opts.keySet()) {
```
zfrenette, line 212: Since you are utilizing both key and value in this loop, why not iterate over the key-value-pair set instead of just the key set?

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
```diff
@@ -0,0 +1,722 @@
...
+    }
+
+    public ServerConfiguration setThreadPrefix(String threadPrefix) {
+        setProperty(THREAD_PREFIX, threadPrefix);
+        return this;
+    }
+
+    public String getThreadPrefix() {
+        return getString(THREAD_PREFIX, DEFAULT_THREAD_PREFIX);
+    }
```
zfrenette, line 699: It doesn't seem like this method is used anywhere. Can it be removed?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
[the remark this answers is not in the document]
nvaishampayan517: Good idea, I'll add a comment.

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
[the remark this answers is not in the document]
nvaishampayan517: That's a good idea, since they were package-private before. However, making them protected would prevent any IT's from using these methods. Do we know that these methods won't be used in IT's in the future?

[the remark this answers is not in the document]
zfrenette: I don't believe so. I had added this class to facilitate the injection of mocks in unit tests.

infrastructure/src/main/java/org/corfudb/infrastructure/configuration/ServerConfiguration.java
nvaishampayan517: I believe that the values of these properties are checked in their usages.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
```diff
@@ -370,6 +373,7 @@ public void testStreamLogDataCorruption() throws Exception {
 
         final int OVERWRITE_BYTES = 4;
 
+        String logDir = getContext().getConfiguration().getLogDir();
```
zfrenette, line 376: Since `getContext()` allocates a new `ServerContext`, can it be reused instead of creating multiple copies in the same test?

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
```diff
@@ -97,12 +98,17 @@ public CorfuServerNode(@Nonnull ServerContext serverContext,
      * Start the Corfu Server by listening on the specified port.
      */
     public ChannelFuture start() {
+        String host = serverContext.getConfiguration().getHostAddress();
+        String interfaceToBind = serverContext.getConfiguration().getNetworkInterface();
+        if (interfaceToBind != null) {
+            host = NetworkUtils.getAddressFromInterfaceName(interfaceToBind);
+        }
```
zfrenette, line 105: Maybe this can be moved to `ServerConfiguration` class when built? Or is there a need for these to remain separate? I noticed that this check is not performed in `NettyLogReplicationServerChannelAdapter::startServer`.

zfrenette (commented):
Looks good. Just some minor questions/comments.

test/src/test/java/org/corfudb/infrastructure/ServerContextBuilder.java
```diff
@@ -13,93 +14,82 @@
...
+                .setSequencerCacheSize(seqCache)
+                .setLogSizeQuota(logSizeLimitPercentage)
+                .setStateTransferBatchSize(batchSize)
+                .setMetadataRetention(retention);
         if (logPath != null) {
-         builder.put("--log-path", logPath);
+            conf.setServerDirectory(logPath);
         }
-        if (managementBootstrapEndpoint != null) {
-            builder.put("--management-server", managementBootstrapEndpoint);
```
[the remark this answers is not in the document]
nvaishampayan517: This param is not in the current Docopt String in master. Maybe Security.md is outdated?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerContext.java
```diff
@@ -667,16 +600,10 @@ public Layout copyManagementLayout() {
...
                 .build();
 
-        final int requestedThreads =
-                Integer.parseInt(getServerConfig(String.class, "--Threads"));
-        final int numThreads = requestedThreads == 0
-                ? Runtime.getRuntime().availableProcessors() * 2
-                : requestedThreads;
-        EventLoopGroup group = getChannelImplementation().getGenerator()
-            .generate(numThreads, threadFactory);
+        EventLoopGroup group = configuration.getChannelImplementation().getGenerator().generate(configuration.getNumIOThreads(), threadFactory);
```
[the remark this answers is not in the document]
nvaishampayan517: I think that was an overlooked change, adding it back.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
nvaishampayan517: I think that it is present to keep separate tests in the class independent from each other, in case some future test added needs to modify the configuration.

infrastructure/src/main/java/org/corfudb/infrastructure/CorfuServerNode.java
nvaishampayan517: The check is already performed in `ServerConfiguration`:
`String networkInterfaceName = configProperties.getProperty(NETWORK_INTERFACE);
        String address = configProperties.getProperty(HOST_ADDRESS);
        if (networkInterfaceName != null) {
            conf.setHostAddress(getAddressFromInterfaceName(networkInterfaceName));
            conf.setNetworkInterface(networkInterfaceName);
            conf.setBindToAllInterfaces(false);
        } else if (address == null) {
            // Default the address to localhost and set the bind to all interfaces flag to true,
            // if the address and interface is not specified.
            conf.setBindToAllInterfaces(true);
            conf.setHostAddress("localhost");
        } else {
            // Address is specified by the user.
            conf.setHostAddress(address);
            conf.setBindToAllInterfaces(false);
        }`
Removing redundant check.

test/src/test/java/org/corfudb/infrastructure/log/StreamLogFilesTest.java
zfrenette: I agree that each test should call this method for the reason you mention, but some tests call `getContext` multiple times.

nvaishampayan517: Oh got it, you're suggesting creating a local variable in the test method.

zfrenette (approved):

xnull (approved):

PR #2998, opened 2021-07-16
Remove Trim Mark Comparison in StreamAddressSpace.addAddress

annym:
## Overview

Description:

- Due to underlying bug in server reset: data wipe + data transfer
(data seems to be available for scan and address spaces are populated
before data transfer starts).

Why should this be merged: unblock Corfu merge (still to be brought back and fix underlying bug)

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
@@ -141,7 +141,13 @@ final public class StreamAddressSpace {
      * @param address address to add.
      */
     public void addAddress(long address) {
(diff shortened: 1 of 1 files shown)


hisundar (approved):

runtime/src/main/java/org/corfudb/runtime/view/stream/StreamAddressSpace.java
```diff
@@ -141,7 +141,10 @@ private long getFirst() {
      * @param address address to add.
      */
     public void addAddress(long address) {
-        if (address <= this.trimMark || Address.nonAddress(address)) {
```
hisundar, line 144: should we log instead so we know when it is happening and the address that appears in the trim mark to help debug?

Lujie1996 (approved):

PR #3000, opened 2021-07-20
Templates for all Stream Listener Resubscription patterns

annym:
## Overview

Description: implementation of 2 re-subscription policies onError:
- **resumeOrDefault**: attempts to resume streaming subscription from the last processed entry and whenever not possible, defaults to subscribe to the latest point in the log. This can incur in data loss. 
- **resumeOrFullSync**: attempts to resume streaming subscription from the last processed entry and whenever not possible, performs full sync and then subscribes from this point onwards. This implementation guarantees no data loss. 

- [x] Public API has Javadoc


test/src/test/java/org/corfudb/integration/StreamingPatternsIT.java
```diff
@@ -0,0 +1,577 @@
...
+        singleNodeEndpoint = String.format("%s:%d",
+                corfuSingleNodeHost,
+                corfuStringNodePort);
+    }
+
+    /**
+     * Test stream listener implementation with default re-subscription policy, i.e.,
+     * resume subscription if possible or subscribe to latest point in the log when not possible.
+     */
+    private class StreamListenerResumeOrDefaultImpl extends StreamListenerResumeOrDefault {
```
achinmishra, line 74: StreamListenerResumeOrDefault talks about a possible data loss. Do you think its a good idea to add an IT to demonstrate such possibility?

annym: There is, take a look at testDefaultPortionOfResumeOrDefaultPolicy (we exercise the Data loss part as well)

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+@Slf4j
+public abstract class StreamListenerFullSync implements StreamListener {
+
+    private final CorfuStore store;
+    private final String namespace;
+    private final String streamTag;
+    private final List<String> tablesOfInterest;
+
+    public StreamListenerFullSync(CorfuStore store, String namespace, String streamTag, List<String> tablesOfInterest) {
```
zhangn49, line 30: Looks like constructor of an abstract class should be a protected method.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public void onError(Throwable throwable) {
+        log.error("Exception caught during streaming processing. Re-subscribe this listener to latest timestamp. " +
+                "Error={}", throwable);
+
+        if (throwable instanceof UnreachableClusterException) {
+            try {
+                IRetry.build(IntervalRetry.class, () -> {
+                    try {
+                        log.info("");
```
zhangn49, line 63: log msg is missing

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
...
+            store.subscribeListener(this, namespace, streamTag, tablesOfInterest, fullSyncTimestamp);
+        }
+    }
+
+    protected abstract Timestamp performFullSync();
+
+    @Override
+    public void onError(Throwable throwable) {
+        log.error("Exception caught during streaming processing. Re-subscribe this listener to latest timestamp. " +
+                "Error={}", throwable);
```
zhangn49, line 57: {} is not needed here.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
...
+                }).run();
+            } catch (InterruptedException e) {
+                log.error("Unrecoverable exception when attempting to subscribe listener.", e);
+                throw new UnrecoverableCorfuInterruptedError(e);
+            }
+        }
+
+        // This listener assumes data loss cannot be afford, therefore, full sync is
+        // performed to recover the full state and immediately re-subscribe based on
+        this.subscribe();
```
zhangn49, line 79: If the retry above succeeds, this one still gets called, right? Then it will throw a streaming exception?

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
...
+                    return null;
+                }).run();
+            } catch (InterruptedException e) {
+                log.error("Unrecoverable exception when attempting to subscribe listener.", e);
+                throw new UnrecoverableCorfuInterruptedError(e);
+            }
+        }
+
+        // This listener assumes data loss cannot be afford, therefore, full sync is
+        // performed to recover the full state and immediately re-subscribe based on
```
zhangn49, line 78: looks like comment is incomplete..

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
...
+        Timestamp fullSyncTimestamp = performFullSync();
+
+        if (tablesOfInterest == null) {
+            store.subscribeListener(this, namespace, streamTag, fullSyncTimestamp);
+        } else {
+            store.subscribeListener(this, namespace, streamTag, tablesOfInterest, fullSyncTimestamp);
+        }
+    }
+
+    protected abstract Timestamp performFullSync();
```
zhangn49, line 52: Would be great to have a comment on this method about how they should implement it.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Created by annym on 06-22-2021
+ */
+@Slf4j
+public abstract class StreamListenerResumePolicy implements StreamListener {
+
+    public final CorfuStore store;
+    public final String namespace;
+    public final String streamTag;
+    public final List<String> tablesOfInterest;
```
zhangn49, line 27: Protected

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
```diff
@@ -0,0 +1,103 @@
...
+    public Timestamp lastProcessedEntryTs;
```
zhangn49, line 28: private

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * This is an abstract stream listener callback implementation for clients interested in automatic
+ * re-subscription upon any error. The re-subscription policy of this implementation protects against data
+ * loss by immediately performing a full sync on any type of error.
+ *
+ * Created by annym on 06-22-2021
+ */
+@Slf4j
+public abstract class StreamListenerResumePolicy implements StreamListener {
```
zhangn49, line 22: This policy class seems to be a base class. Are we going to expose it to users? If not, might be good to make it package private?

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
```diff
@@ -0,0 +1,103 @@
...
+                }).run();
+            } catch (InterruptedException e) {
+                log.error("Unrecoverable exception when attempting to re-subscribe listener.", e);
+                throw new UnrecoverableCorfuInterruptedError(e);
+            }
+        }
+
+        // This listener assumes data loss cannot be afford, therefore, full sync is
+        // performed to recover the full state and immediately re-subscribe based on
+        resumeSubscription();
```
zhangn49, line 101: Same here.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public void onNextEntry(CorfuStreamEntries results) {
+        onNext(results);
+        this.lastProcessedEntryTs = results.getTimestamp();
+    }
+
+    @Override
+    public void onError(Throwable throwable) {
+        log.error("Exception caught during streaming processing. Re-subscribe this listener to latest timestamp. " +
+                "Error={}", throwable);
```
zhangn49, line 79: same here.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
+package org.corfudb.runtime.collections;
+
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
+import org.corfudb.runtime.exceptions.StreamingException;
+import org.corfudb.runtime.exceptions.TrimmedException;
```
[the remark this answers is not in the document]
annym: actually all this file is not needed, it was a copy I forgot to remove!

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
annym: correct, I started off making it a class to expose to users but it ended up as a base class. Making it private.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
zhangn49: @annym This method will get called, even the above retry succeed. I see you catch StreamingException in resumeSubscription, but it will have cost to validateSyncAddress and then log an error message, which might be confusing to users. What do you think?

annym: oh yes, this should be with an else, that's what you're saying right?

zhangn49: yes, you are right.

zhangn49 (approved):
LGTM! Except a few Codacy comments need to address.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumeOrDefault.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void subscribeOnResumeError() {
+        // Immediately re-subscribe from the latest point in the log, this means there is no interest in
+        // keeping continuity with the last processed entry before an error was thrown.
+        // Be aware that this re-subscription pattern will likely lead to data loss, i.e., you will lose any
+        // data between the last processed entry and the current log tail.
+        try {
+            if (tablesOfInterest == null) {
```
pankti-m, line 42: should there be a check to validate streamTag != null?

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumeOrDefault.java
```diff
@@ -0,0 +1,51 @@
...
+                                         List<String> tablesOfInterest) {
+        super(store, namespace, streamTag, tablesOfInterest);
+    }
+
+    public StreamListenerResumeOrDefault(CorfuStore store, String namespace, String streamTag) {
+        this(store, namespace, streamTag, null);
+    }
+
+    @Override
+    public void subscribeOnResumeError() {
```
pankti-m, line 36: is it better to rename it to resumeOnError() or resumeSubscriptionOnError()?

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumePolicy.java
```diff
@@ -0,0 +1,101 @@
...
+import org.corfudb.util.retry.IRetry;
+import org.corfudb.util.retry.IntervalRetry;
+import org.corfudb.util.retry.RetryNeededException;
+
+import java.util.List;
+
+/**
+ * This is an abstract stream listener callback implementation for clients interested in automatic
+ * re-subscription upon any error. The re-subscription policy of this implementation protects against data
+ * loss by immediately performing a full sync on any type of error.
```
pankti-m, line 17: remove full sync part?

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
```diff
@@ -6,6 +6,11 @@
  * Created by hisundar on 2019-10-18
  */
 public interface StreamListener {
+
+    default void onNextEntry(CorfuStreamEntries results) {
```
Lujie1996, line 10: What's the purpose of `onNextEntry()` given that we already have `onNext()`?

test/src/test/java/org/corfudb/integration/StreamingPatternsIT.java
```diff
@@ -0,0 +1,591 @@
...
+                throw new IllegalArgumentException("performFullSync logic exception!!");
+            }
+
+            // Reset buffer
+            updates.clear();
+
+            // Full-Sync table
+            try (TxnContext txn = store.txn(namespace)) {
+                Table<Uuid, SampleTableAMsg, Uuid> table = txn.getTable(defaultTableName);
+                table.entryStream().forEach(entry -> fullSyncBuffer.add(entry));
```
Lujie1996, line 188: nit: table.entryStream().forEach(fullSyncBuffer::add);

test/src/test/java/org/corfudb/integration/StreamingPatternsIT.java
```diff
@@ -0,0 +1,591 @@
...
+
+        @Override
+        public void onNext(CorfuStreamEntries results) {
+            if (updates.size() == updatesToError) {
+                // Decrement updatesToError by 1 so further updates are processed, after the error
+                updatesToError -= 1;
+                throw new IllegalStateException("Artificial exception to trigger onError");
+            }
+
+            results.getEntries().values().forEach(entryList -> entryList.forEach(entry -> updates.add(entry)));
```
Lujie1996, line 157: nit: `results.getEntries().values().forEach(updates::addAll);`

test/src/test/java/org/corfudb/integration/StreamingPatternsIT.java
achinmishra: Thanks

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumeOrDefault.java
annym: it's a subtlety in the language as this method contains the logic of subscription when 'resume' errors/fails not the resume logic onError (it might be tricky because the difference is subtle, if its not clear I can explain offline)

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumeOrDefault.java
annym: no, this was already provided.. at this stage it isn't null as that was already validated.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerFullSync.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public void subscribe() {
+
+        Timestamp fullSyncTimestamp = performFullSync();
+
+        if (tablesOfInterest == null) {
+            store.subscribeListener(this, namespace, streamTag, fullSyncTimestamp);
+        } else {
+            store.subscribeListener(this, namespace, streamTag, tablesOfInterest, fullSyncTimestamp);
```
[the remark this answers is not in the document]
annym: same here, this is a re-subscription onError (after it has already been subscribed with streamTag, so that check was already done by this point)

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
annym: If you look at StreamListenerResumePolicy, we need onNext to set the lastProcessedTimestamp on each onNext call, so onNextEntry ensures that happens after every onNext (leaving that to the listener implementation is error prone as they might not do it) so that's the purpose of onNextEntry

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumeOrDefault.java
```diff
@@ -0,0 +1,51 @@
...
+ * position in the log might have progressed beyond several updates from
+ * the last processed entry.
+ *
+ * Please note if data loss is a concern DO NOT USE this listener implementation
+ * and refer to StreamListenerResumeOrFullSync
+ *
+ * Created by annym on 06-22-2021
+ */
+@Slf4j
+public abstract class StreamListenerResumeOrDefault extends StreamListenerResumePolicy {
```
hisundar, line 24: One general concern I have is I see a few implementations of StreamListener extending other classes, which will break since java does not support multiple inheritance but that is something we just need to help folks with.

hisundar (approved):
The PR looks good to me overall, My comments are minor

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
annym: It is really to be able to support the 'resume' portion of these new patterns. On every onNext call we need to keep track of the last processed entry, so this is just a wrapper over the onNext callback, which keeps track of this timestamp, right after the user executes its onNext logic.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumeOrDefault.java
annym: It is already in the abstract class but I can paste it here as well

runtime/src/main/java/org/corfudb/runtime/collections/StreamListenerResumeOrDefault.java
annym: I see, this is valid. But I'm not sure how we could with this on our side. In any case, these 'patterns' are optional and available just to ease their job and for all boilerplate code. So they can still decide not to use these patterns directly.

PR #3001, opened 2021-07-20
Injection-framework: upload correctness logs as artifacts

xnull:
## Overview

Description:
Upload correctness logs as artifacts, which allows analyzing correctness logs 

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -192,6 +192,13 @@ jobs:
         working-directory: ./tests
         run: ./gradlew clean test -Dtags=longevity-docker
 
+      - name: upload longevity logs artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: injection-framework-logs
+          retention-days: 3
+          path: ./tests/build/corfu-longevity-app/correctness.log
+
       - name: Run injection framework verification
         working-directory: ./tests/build/corfu-longevity-app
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -192,6 +192,13 @@ jobs:
         working-directory: ./tests
         run: ./gradlew clean test -Dtags=longevity-docker
 
+      - name: upload longevity logs artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: injection-framework-logs
+          retention-days: 3
+          path: ./tests/build/corfu-longevity-app/correctness.log
+
```
xcchang, line 201: Is it possible to show the logs directly in the previous step? like what we have in unit tests, then we could have much longer retention days. I think compatibility-test also has this problem, might be related with Gradle setup.

xnull: I think it's not necessary. 3 days is more than enough, you can just download the artifact. If the artifact got deleted you just restart a job. 
In that case, you don't need to pollute logs with huge output.

xnull: It will be pretty problematic to print 260 mb of logs
image

xcchang: Got it, make sense. Thanks!

xcchang (approved):

PavelZaytsev (approved):

PR #3004, opened 2021-07-22
Fix the race condition in DynamicProtobufSerializer that causes loss of classMap

Lujie1996:
## Overview

Description:
Reusing existing protobuf serializer to read table registry and initialize DynamicProtobufSerializer.

Why should this be merged: 
Previously DynamicProtobufSerializer creates and registers a new protobuf serializer, which could cause loss of classMap in the existing protobuf serializer. If someone is using protobuf serializer at this time and before dynamic protobuf serializer's initialization finishes, it could lead to SerializerException due to missing class type.

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -103,9 +103,15 @@
...
-        // Create and register a protobuf serializer to read the table registry.
-        ISerializer protobufSerializer = createProtobufSerializer();
-        Serializers.registerSerializer(protobufSerializer);
+        // Create or get a protobuf serializer to read the table registry.
+        ISerializer protobufSerializer;
+        try {
+            protobufSerializer = Serializers.getSerializer(ProtobufSerializer.PROTOBUF_SERIALIZER_CODE);
+        } catch (SerializerException se) {
+            // This means the protobuf serializer had not been registered yet.
+            protobufSerializer = new ProtobufSerializer(new ConcurrentHashMap<>());
```
hisundar, line 112: Don't we want to register the known types?
classMap.put(getTypeUrl(TableName.getDescriptor()), TableName.class);
        classMap.put(getTypeUrl(TableDescriptors.getDescriptor()),
            TableDescriptors.class);
        classMap.put(getTypeUrl(TableMetadata.getDescriptor()),
            TableMetadata.class);
        classMap.put(getTypeUrl(ProtobufFileName.getDescriptor()),
            ProtobufFileName.class);
        classMap.put(getTypeUrl(ProtobufFileDescriptor.getDescriptor()),
            ProtobufFileDescriptor.class);

hisundar (commented):
Thanks George this is a very good find, but just wondering if we need to register the known types first or is it already taken care of?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
Lujie1996: You're right. I missed that point. Changed to use `createProtobufSerializer` which registers those types.

pankti-m (commented):
Wouldnt the call to deregister the serializer on L161 also cause a problem?

I have other higher-level questions.  DynamicProtobufSerializer is only used from the browser/editor tool.  This means this race occurs when the tool is run in parallel with client requests from the application/s.   Browser/editor tool creates a separate runtime and does not use the one used by applications.  So ideally, its serializer(dynamicProtobuf) should not  interfere with the ProtobufSerializer used by the applications.

But looks like the serializers are not tied to a runtime.  Should we fix that issue instead? @hisundar , thoughts?

Lujie1996:
> Wouldnt the call to deregister the serializer on L161 also cause a problem?
> 
> I have other higher-level questions. DynamicProtobufSerializer is only used from the browser/editor tool. This means this race occurs when the tool is run in parallel with client requests from the application/s. Browser/editor tool creates a separate runtime and does not use the one used by applications. So ideally, its serializer(dynamicProtobuf) should not interfere with the ProtobufSerializer used by the applications.
> 
> But looks like the serializers are not tied to a runtime. Should we fix that issue instead? @hisundar , thoughts?

Hi @pankti-m. Please note that DynamicProtobufSerializer is used in compactor as well. Compactor uses DistributedLock to guarantee no parallel compactor runs, which frequently sends heart beat to maintain lock liveness. The time from discarding and creating new ProtobufSerializer to registering DynamicProtobufSerializer if interleaved with DistributedLock heartbeat will hit deserializer issue.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -103,9 +103,15 @@
     public DynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
         this.type = ProtobufSerializer.PROTOBUF_SERIALIZER_CODE;
 
-        // Create and register a protobuf serializer to read the table registry.
-        ISerializer protobufSerializer = createProtobufSerializer();
-        Serializers.registerSerializer(protobufSerializer);
+        // Create or get a protobuf serializer to read the table registry.
```
zhangn49, line 106: Is it possible to have a test? It will validate your fix and prevent future regression.

Lujie1996: Good question. In order to reproduce the race condition, we need to run `new DynamicProtobufSerializer()` followed by a read operation which executes deserialize(), and make sure that the read returns before DynamicProtobufSerializer constructor completes. In our test environment it's hard to guarantee this condition since they are both lightning fast on local node. Do you have any thought?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -308,8 +314,15 @@ public Object deserialize(ByteBuf b, CorfuRuntime rt) {
             Record record = Record.parseFrom(data);
             Any payload = record.getPayload();
 
+            String fullMessageName = getFullMessageName(payload);
+            if (!messagesFdProtoNameMap.containsKey(fullMessageName)) {
+                log.error("messagesFdProtoNameMap doesn't contain the message type {} from payload {}",
```
hisundar, line 319: Do we want to explicitly ask, "Was this table opened with the correct openTable() command and appropriate schema?"

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -103,9 +103,15 @@
     public DynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
         this.type = ProtobufSerializer.PROTOBUF_SERIALIZER_CODE;
 
-        // Create and register a protobuf serializer to read the table registry.
-        ISerializer protobufSerializer = createProtobufSerializer();
-        Serializers.registerSerializer(protobufSerializer);
+        // Create or get a protobuf serializer to read the table registry.
+        ISerializer protobufSerializer;
+        try {
+            protobufSerializer = Serializers.getSerializer(ProtobufSerializer.PROTOBUF_SERIALIZER_CODE);
```
hisundar, line 109: Can we please fold this error message in Serializers.java into the exception itself so it does not get logged?
log.error("Serializer with type code {} not found. Please check custom serializers " +
                "for this client.", type.intValue());
as part of this change?

hisundar (approved):
The change looks good, but just have a small request on logging. thanks

pankti-m (approved):

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
Lujie1996: Done.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
Lujie1996: Serializers.getSerializer() has already included the error message inside the SerializerException. Removed log.error() in Serializers.getSerializer().

PR #3005, opened 2021-07-24
[WIP] Server Side RocksDB interface - framework for addition of RemoteCorfuTable objects

nvaishampayan517:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
```diff
@@ -0,0 +1,472 @@
...
+import static org.corfudb.infrastructure.remotecorfutable.utils.DatabaseConstants.METADATA_COLUMN_SUFFIX;
+
+/**
+ * The DatabaseHandler provides an interface for the RocksDB instance storing data for server side
+ * RemoteCorfuTables.
+ *
+ * <p>Created by nvaishampayan517 on 7/23/21.
+ */
+@Slf4j
+public class DatabaseHandler implements AutoCloseable {
```
xnull, line 40: There is [PersistedStreamingMap](https://github.com/CorfuDB/CorfuDB/blob/d5443d431eb3f11a54dfebd95c19473f9cae23d6/runtime/src/main/java/org/corfudb/runtime/collections/PersistedStreamingMap.java#L38) which contains almost the same code, so we have code duplication. Is it possible to create another PR right now and extract common code from DatabaseHandler and PersistedStreamingMap?

We'd have another smaller PR, faster to review, less probability to have a bug. 
If that would work for you then would be great to have another pr.

xnull: Then this PR will use that smaller pr

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
```diff
@@ -0,0 +1,472 @@
...
+     * @param streamID The stream backing the database to read from.
+     * @return A byte[] containing the value mapped to the key, or null if the key is associated to a null value.
+     * @throws RocksDBException An error raised in scan.
+     */
+    public byte[] get(byte[] encodedKey, ByteString streamID) throws RocksDBException, DatabaseOperationException {
+        final byte[] returnVal;
+        if (!columnFamilies.containsKey(streamID)) {
+            throw new DatabaseOperationException("GET", "Invalid stream ID");
+        }
+        RocksIterator iter = database.newIterator(columnFamilies.get(streamID));
```
xnull, line 134: you will have resource leaks here, please take a look at how this logic is implemented in PersistedStreamingMap without leaks

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
```diff
@@ -0,0 +1,472 @@
...
+        boolean encounteredPrefix = false;
+        iter.seek(encodedKeyBegin);
+        List<byte[][]> results = new LinkedList<>();
+        int elements = 0;
+        while (iter.isValid()) {
+            currPrefix = KeyEncodingUtil.extractEncodedKey(iter.key());
+            if (Arrays.equals(currPrefix, prevPrefix)) {
+                if (!encounteredPrefix) {
+                    if (!skipFirst) {
+                        results.add(new byte[][]{iter.key(), iter.value()});
```
xnull, line 335: agree with codacy

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseOperationException.java
```diff
@@ -0,0 +1,12 @@
+package org.corfudb.infrastructure.remotecorfutable;
+
+/**
+ * This class defines the Exception format for errors arising in RemoteCorfuTable database operations.
+ *
+ * <p>Created by nvaishampayan517 on 7/27/21.
+ */
+public class DatabaseOperationException extends Exception {
```
xnull, line 8: why not RuntimeException?

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
nvaishampayan517: From what I see, only the outward facing API is the same between the persisted streaming map and this handler. In this handler, we're enforcing strict Key formats as well as supporting the new record-level MVCC implementation. So under the hood, the methods are doing things like getting recent versions through iteration, where in the PersistentStreamingMap the same method would be implemented using just GET calls.

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
nvaishampayan517: I'm not sure where the resource leak is coming from. The only resources allocated are in the iterator, and the iterator is closed on error and termination.

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
nvaishampayan517: I'll find some way to reduce the complexity of the nested if-else, I'll keep as a behavior placeholder right now and add TODO.

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
```diff
@@ -0,0 +1,472 @@
...
+        } catch (RocksDBException e) {
+            log.error("Error in creating metadata column family for table {}.", streamID.toString(DATABASE_CHARSET));
+            log.error("Cause of error: ", e);
+            tableHandle.close();
+            try {
+                database.dropColumnFamily(tableHandle);
+            } catch (RocksDBException r) {
+                log.error("Error in dropping column family for table {}.", streamID.toString(DATABASE_CHARSET));
+                log.error("Cause of error: ", r);
+                throw e;
```
xnull, line 108: image

infrastructure/src/main/java/org/corfudb/infrastructure/remotecorfutable/DatabaseHandler.java
xnull: You had your chance to figure out that by yourself :) so let me enjoy dig into it :)
Probably the leak is not there, but it's hard to prove it by looking at the code.

- offtop: `returnVal = null;` not a great idea IMO. Your code would fail with NPE at any time in any other method which would try to get access into the array.
- offtop: `columnFamilies` - race condition could be possible if put and get happens at the same time in different threads then your method throws an exception
- I don't like the idea of throwing `RocksDBException`, and also if you re throw an exception then never write it to the log, otherwise you'd have multiple exceptions logged in the file on different levels which is very confusing and will consume a lot of resources.
-  `iter.status();` is used whether to throw an exception or return null, which so vague logic, and hard to develop any intuition and motivation why it's done that way. I would get rid of it completely. So that `if` condition needs to be refactored completely
- and finally `iter`, I hope this version of code looks better for you:
```
try(RocksIterator iter = database.newIterator(columnFamilies.get(streamID))) {
            iter.seek(encodedKey);

            if (!iter.isValid()) {
                try {
                    iter.status();
                } catch (RocksDBException e) {
                    //We've hit some error in iteration
                    log.error("Error in RocksDB get operation:", e);
                    throw e;
                }
                returnVal = null;
            } else {
                if (Arrays.equals(KeyEncodingUtil.extractEncodedKey(iter.key()),
                        KeyEncodingUtil.extractEncodedKey(encodedKey))) {
                    //This works due to latest version first comparator
                    returnVal = iter.value();
                } else {
                    //TODO: add read from log stream check
                    returnVal = null;
                }
            }
        }
```

PR #3007, opened 2021-07-27
Add Tests for Secondary Indexes on Enum Types

annym:
## Overview

Description: secondary indexes on an Enum type should be queried based on the ValueDescriptor of the value of interest, this is due to the internal way that protoBuf deals with enum fields. This PR adds a test that shows the correct usage of secondary indexes on first and nested levels over an enum type (for future reference, as it is currently used by clients).

Why should this be merged: example of usage pattern and increase code coverage. 

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/runtime/collections/CorfuStoreSecondaryIndexTest.java
+++ b/test/src/test/java/org/corfudb/runtime/collections/CorfuStoreSecondaryIndexTest.java
@@ -4,6 +4,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.ExampleSchemas;
 import org.corfudb.runtime.ExampleSchemas.ExampleValue;
+import org.corfudb.runtime.ExampleSchemas.ActivitySchedule;
 import org.corfudb.runtime.ExampleSchemas.ManagedMetadata;
 import org.corfudb.runtime.ExampleSchemas.Adult;
 import org.corfudb.runtime.proto.RpcCommon.UuidMsg;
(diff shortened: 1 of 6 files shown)


zhangn49 (approved):
LGTM! Could we can mention how enum should be used in the comment of secondary index get function?

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreSecondaryIndexTest.java
```diff
@@ -104,6 +105,80 @@ public void testSecondaryIndexes() throws Exception {
...
+        final ManagedMetadata metadata = ManagedMetadata.newBuilder().setCreateUser("user_UT").build();
+
+        for (int i = 0; i < numRecords; i++) {
+            UUID uuid = UUID.randomUUID();
+            UuidMsg key = UuidMsg.newBuilder()
+                    .setMsb(uuid.getMostSignificantBits()).setLsb(uuid.getLeastSignificantBits())
+                    .build();
+
+            try (ManagedTxnContext txn = shimStore.tx(someNamespace)) {
+                ActivitySchedule activity = ActivitySchedule.newBuilder()
```
pankti-m, line 145: nit - move this construction out of the tx context?

pankti-m (approved):
Nice example :)

pankti-m (approved):
Should we change the title and description to include all types of enum secondary indexes and not just nested ones?

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreSecondaryIndexTest.java
annym: no need, its not used anywhere else, it's the object within the transaction scope.

PR #3008, opened 2021-07-28
Deduplicate Stream Updates for same keys within a transaction

annym:
## Overview

Description: to be compliant with the former behavior of the streaming (DCN) api, updates within the same transaction boundary should be de-duplicated for correct streaming processing. 

Why should this be merged: bug in clients (change of behavior)

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
@@ -127,8 +127,13 @@ public class StreamSubscription<K extends Message, V extends Message, M extends 
                     .stream()
                     .map(entry -> CorfuStreamEntry.fromSMREntry(entry, epoch))
                     .collect(Collectors.toList());
+
+            // Deduplicate entries per stream Id, ordering within a transaction is not guaranteed
+            Map<Message, CorfuStreamEntry> observedKeys = new HashMap<>();
+            entryList.forEach(entry -> observedKeys.put(entry.getKey(), entry));
+
             if (!entryList.isEmpty()) {
-                streamEntries.put(schema, entryList);
+                streamEntries.put(schema, observedKeys.values().stream().collect(Collectors.toList()));
             }
         });
 
--- a/test/src/test/java/org/corfudb/integration/StreamingIT.java
+++ b/test/src/test/java/org/corfudb/integration/StreamingIT.java
@@ -16,7 +16,6 @@ import org.corfudb.runtime.collections.CorfuStreamEntry;
 import org.corfudb.runtime.collections.CorfuTable;
 import org.corfudb.runtime.collections.IsolationLevel;
 import org.corfudb.runtime.collections.StreamListener;
-import org.corfudb.runtime.collections.StreamManager;
 import org.corfudb.runtime.collections.Table;
 import org.corfudb.runtime.collections.TableOptions;
 import org.corfudb.runtime.collections.TableSchema;
@@ -35,11 +34,12 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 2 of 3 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
```diff
@@ -127,6 +129,11 @@ boolean enqueueStreamEntry(ILogData logData, long maxBlockTime) throws Interrupt
                     .stream()
                     .map(entry -> CorfuStreamEntry.fromSMREntry(entry, epoch))
                     .collect(Collectors.toList());
+
+            // Deduplicate entries per stream Id
+            Set<Message> observedKeys = new HashSet<>();
+            entryList.removeIf(e-> !observedKeys.add(e.getKey()));
```
hisundar, line 135: Very neat!

hisundar (approved):
Thanks Anny!

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
Maithem: I think this is wrong, since it will keep the older observed key not the newest one. You need to iterate the `entryList` backwards to keep the latest keys.

Maithem: For example if  `|E1 | E2 | E3` all have the same key, then the algorithm will filter E2 and E3, but it should have kept E3 and filtered out the rest.

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
```diff
@@ -127,8 +132,19 @@ boolean enqueueStreamEntry(ILogData logData, long maxBlockTime) throws Interrupt
...
                     .collect(Collectors.toList());
+
+            // Deduplicate entries per stream Id
+            Map<Message, CorfuStreamEntry> observedKeys = new LinkedHashMap<>();
+            entryList.forEach(entry -> {
+                if (observedKeys.putIfAbsent(entry.getKey(), entry) != null) {
+                    // If the key is present, remove and insert again to preserve transaction order
+                    observedKeys.remove(entry.getKey());
+                    observedKeys.put(entry.getKey(), entry);
+                }
```
Maithem, line 143: Why not just do 

```java

Map<Message, CorfuStreamEntry> observedKeys = new LinkedHashMap<>();
            entryList.forEach(entry -> {
                observedKeys.put(entry.getKey(), entry);
            });

```


Latest keys will overwrite the older keys that's because the list has then order by version (implicit mutation order). 

Also, instead of doing another iteration, you can construct the latest unique keys in the same stream https://github.com/CorfuDB/CorfuDB/pull/3008/files#diff-2de6eaba4e142127e6a6aec1b2c3e0e5c13c0312251c321a457f8858722e0fd1R133

zhangn49, line 143: Looks like we can simply call `put(entry.getKey(), entry)` to deduplicate

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
annym: the problem is that we want to preserve the order, so the case: 
k1, v1
k1, v2
k2, v3
k1, v4 -> if we don't remove you'll report the updates in the following way:

k1, v4 -> k2, v3 (which is incorrect).. and I tried the replace function but that preserves the initial insertion order

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
annym: Correct, I had already changed it in my second push, so closing as I addressed this and added the consideration in the test.

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
annym: the problem is that we want to preserve the order, so the case:
k1, v1
k1, v2
k2, v3
k1, v4 -> with this approach would yield:

k1, v4 -> k2, v3 (which is incorrect).. and I tried the 'replace' function as well but that also preserves the initial insertion order

Maithem: Why do we need to preserve order between unique keys? 

Maybe it should be documented on the subscription API, if its necessary. It also seems like challenging to code against for such semantics. That means that the consumer needs to encode some relationship between the unique keys for this order to matter.

Maithem (approved):

runtime/src/main/java/org/corfudb/runtime/collections/StreamSubscription.java
apavan: I do see a good point here about why we need to preserve the order across unique keys within a transaction. Current expectation is to emit the final version of a key within the transaction but does not talk about any order of updates within a transaction. Please share your thoughts on how this can be beneficial for the clients?

hisundar: Consider the following operation in the same transaction:
enqueue(A)
enqueue(A`)
if we deliver A after A` then the apply will result in incorrect version of the object right?

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -1396,6 +1396,77 @@ public void testSubscribeListenerNoTimestamp() throws Exception {
...
+        try (TxnContext tx = store.txn(namespace)) {
+            Uuid key1 = Uuid.newBuilder().setMsb(0).setLsb(0).build();
+
+            // Update key 1 multiple times within the same transaction
+            for (int i = 0; i < numUpdates; i++) {
+                SampleTableAMsg msg = SampleTableAMsg.newBuilder().setPayload(String.valueOf(i)).build();
+                tx.putRecord(tableA, key1, msg, key1);
+            }
+
+            // Add another (unique) key to confirm order of updates is preserved
```
pankti-m, line 1442: not sure I understand this comment... no order is guaranteed within a transaction, right?

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -1396,6 +1396,77 @@ public void testSubscribeListenerNoTimestamp() throws Exception {
...
+            Uuid key2 = Uuid.newBuilder().setMsb(1).setLsb(1).build();
+            tx.putRecord(tableA, key2, SampleTableAMsg.newBuilder().setPayload(String.valueOf(numUpdates + delta)).build(), key2);
+
+            // Update key1 one last time so the order is changed
```
pankti-m, line 1446: same, what does the comment mean?

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: we had a discussion and decided to change the behavior, so this was just left behind, changing it! thanks!

test/src/test/java/org/corfudb/integration/StreamingIT.java
annym: same here, old behavior.. removed!

pankti-m (approved):

Issue #3009, opened 2021-07-28
Sequencer Cache Leak


Maithem:
**Bug report.**

The sequencer implements a sliding window of conflict keys that moves the window as time progresses (sequence numbers increase). The eviction logic assumes that the window has unique conflict keys, which is not a correct assumption. This can lead to a memory leak when the same keys are written over and over again. In particular, the map that holds the latest mapping (conflict key -> version) can have a constant size, while the priority queue that holds the window  can increase without bounds . 

https://github.com/CorfuDB/CorfuDB/blob/master/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java#L218

PR #3010, opened 2021-07-30
Fix sequencer cache eviction policy

zfrenette:
The sequencer cache implements a sliding window to help with cache
eviction, which incorrectly assumes that conflict keys are unique.
This can lead to a memory leak when the same keys are written over
and over again. This patch also fixes a bug where the cache evicts
partial conflicts keys for a particular timestamp. This can lead to
incorrect resolutions and can cause data loss.

## Overview

Description:

Why should this be merged:  Fixes eviction bugs that can cause incorrect resolutions and memory bloat.

Related issue(s) (if applicable): #3009 

- [X] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -2,15 +2,13 @@ package org.corfudb.infrastructure;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.ByteString;
 import com.google.protobuf.TextFormat;
-import io.micrometer.core.instrument.Meter;
-import io.micrometer.core.instrument.Tags;
 import io.netty.channel.ChannelHandlerContext;
 import lombok.Builder;
 import lombok.Builder.Default;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-import org.corfudb.common.metrics.micrometer.MeterRegistryProvider;
 import org.corfudb.common.metrics.micrometer.MicroMeterUtils;
 import org.corfudb.infrastructure.SequencerServerCache.ConflictTxStream;
 import org.corfudb.protocols.CorfuProtocolCommon;
@@ -39,6 +37,7 @@ import org.corfudb.util.Utils;
 
 import javax.annotation.Nonnull;
 import java.lang.invoke.MethodHandles;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -275,11 +274,33 @@ public class SequencerServer extends AbstractServer {
                     txInfo, txSnapshotTimestamp, trimMark);
             return new TxResolutionResponse(TokenType.TX_ABORT_SEQ_TRIM);
         }
+
+        // The maxConflictNewSequencer is modified whenever a server is elected
+        // as the 'new' sequencer, we immediately set its value to the max timestamp
+        // evicted from the cache at that time. If a txSnapshotTimestamp falls
+        // under this threshold we can report that the cause of abort is due to
+        // a NEW_SEQUENCER (not able to hold these in its cache).
+        long maxConflictNewSequencer = cache.getMaxConflictNewSequencer();
+        if (txSnapshotTimestamp.getSequence() < maxConflictNewSequencer) {
+            log.debug("ABORT[{}] ts[{}] WILDCARD New Sequencer ts=[{}]",
+                    txInfo, txSnapshotTimestamp, maxConflictNewSequencer);
+            return new TxResolutionResponse(TokenType.TX_ABORT_NEWSEQ);
+        }
+
+        // If the txSnapshotTimestamp did not fall under the new sequencer threshold
+        // but it does fall under the latest evicted timestamp we report the cause of
+        // abort as SEQUENCER_OVERFLOW
+        long maxConflictWildcard = cache.getMaxConflictWildcard();
+        if (txSnapshotTimestamp.getSequence() < maxConflictWildcard) {
+            log.debug("ABORT[{}] ts[{}] WILDCARD ts=[{}]",
+                    txInfo, txSnapshotTimestamp, maxConflictWildcard);
+            return new TxResolutionResponse(TokenType.TX_ABORT_SEQ_OVERFLOW);
+        }
+
         MicroMeterUtils.measure(txInfo.getConflictSet().size(), "sequencer.tx-resolution.num_streams");
         for (Map.Entry<UUID, Set<byte[]>> conflictStream : txInfo.getConflictSet().entrySet()) {
 
-            // if conflict-parameters are present, check for conflict based on conflict-parameter
-            // updates
+            // if conflict-parameters are present, check for conflict based on conflict-parameter updates
             Set<byte[]> conflictParamSet = conflictStream.getValue();
             //check for conflict based on streams updates
             if (conflictParamSet == null || conflictParamSet.isEmpty()) {
@@ -295,9 +316,7 @@ public class SequencerServer extends AbstractServer {
 
             // for each key pair, check for conflict; if not present, check against the wildcard
             for (byte[] conflictParam : conflictParamSet) {
-
-                Long keyAddress = cache.get(new ConflictTxStream(conflictStream.getKey(),
-                        conflictParam, Address.NON_ADDRESS));
+                Long keyAddress = cache.get(new ConflictTxStream(conflictStream.getKey(), conflictParam));
 
                 if (log.isTraceEnabled()){
                     log.trace("Commit-ck[{}] conflict-key[{}](ts={})",
@@ -313,28 +332,6 @@ public class SequencerServer extends AbstractServer {
                             conflictParam,
                             conflictStream.getKey()
                     );
-                }
-
-                // The maxConflictNewSequencer is modified whenever a server is elected
-                // as the 'new' sequencer, we immediately set its value to the max timestamp
-                // evicted from the cache at that time. If a txSnapshotTimestamp falls
-                // under this threshold we can report that the cause of abort is due to
-                // a NEW_SEQUENCER (not able to hold these in its cache).
-                long maxConflictNewSequencer = cache.getMaxConflictNewSequencer();
-                if (txSnapshotTimestamp.getSequence() < maxConflictNewSequencer) {
-                    log.debug("ABORT[{}] ts[{}] WILDCARD New Sequencer ts=[{}]",
-                            txInfo, txSnapshotTimestamp, maxConflictNewSequencer);
-                    return new TxResolutionResponse(TokenType.TX_ABORT_NEWSEQ);
-                }
-
-                // If the txSnapshotTimestamp did not fall under the new sequencer threshold
-                // but it does fall under the latest evicted timestamp we report the cause of
-                // abort as SEQUENCER_OVERFLOW
-                long maxConflictWildcard = cache.getMaxConflictWildcard();
-                if (txSnapshotTimestamp.getSequence() < maxConflictWildcard) {
-                    log.debug("ABORT[{}] ts[{}] WILDCARD ts=[{}]",
-                            txInfo, txSnapshotTimestamp, maxConflictWildcard);
-                    return new TxResolutionResponse(TokenType.TX_ABORT_SEQ_OVERFLOW);
                 }
             }
         }
@@ -400,7 +397,7 @@ public class SequencerServer extends AbstractServer {
         if (trimMark < req.getPayload().getSequencerTrimRequest().getTrimMark()) {
             // Advance the trim mark, if the new trim request has a higher trim mark.
             trimMark = req.getPayload().getSequencerTrimRequest().getTrimMark();
-            cache.invalidateUpTo(trimMark);
+            cache.evictUpTo(trimMark);
 
             // Remove trimmed addresses from each address map and set new trim mark
             for (StreamAddressSpace streamAddressSpace : streamsAddressMap.values()) {
@@ -487,7 +484,7 @@ public class SequencerServer extends AbstractServer {
         if (!bootstrapWithoutTailsUpdate) {
             globalLogTail = req.getPayload().getBootstrapSequencerRequest().getGlobalTail();
             cache = sequencerFactoryHelper.getSequencerServerCache(
-                    cache.getCacheSize(),
+                    cache.getCapacity(),
                     globalLogTail - 1
             );
 
@@ -710,15 +707,20 @@ public class SequencerServer extends AbstractServer {
 
         // update the cache of conflict parameters
         if (tokenRequest.hasTxnResolution()) {
-            tokenRequest.getTxnResolution().getWriteConflictParamsSetList()
-                    .forEach((item) -> {
-                        // insert an entry with the new timestamp using the
-                        // hash code based on the param and the stream id.
-                        item.getValueList().forEach(conflictParam ->
-                                cache.put(new ConflictTxStream(getUUID(item.getKey()),
-                                        conflictParam.toByteArray(), newTail - 1)));
-                    });
+            List<ConflictTxStream> conflictTxStreamList = tokenRequest.getTxnResolution()
+                    .getWriteConflictParamsSetList()
+                    .parallelStream()
+                    .flatMap(conflictKeysForStream -> {
+                        final UUID streamId = getUUID(conflictKeysForStream.getKey());
+                        return conflictKeysForStream.getValueList()
+                                .stream()
+                                .map(ByteString::toByteArray)
+                                .map(conflictParam -> new ConflictTxStream(streamId, conflictParam));
+                    }).collect(Collectors.toCollection(ArrayList::new));
+
+           cache.put(conflictTxStreamList, newTail - 1);
         }
+
         if (log.isTraceEnabled()) {
             log.trace("handleAllocation: token={} backpointers={}",
                     globalLogTail, backPointerMap.build());
@@ -744,7 +746,7 @@ public class SequencerServer extends AbstractServer {
      * The response contains the requested streams address maps and the global log tail.
      */
     @RequestHandler(type = PayloadCase.STREAMS_ADDRESS_REQUEST)
-    private void handleStreamsAddressRequest(@Nonnull RequestMsg req,
+    public void handleStreamsAddressRequest(@Nonnull RequestMsg req,
                                              @Nonnull ChannelHandlerContext ctx,
                                              @Nonnull IServerRouter r) {
         StreamsAddressRequestMsg streamsAddressRequest =
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
@@ -1,52 +1,61 @@
 package org.corfudb.infrastructure;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import lombok.AccessLevel;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.common.metrics.micrometer.MicroMeterUtils;
-import org.corfudb.common.util.Memory;
 import org.corfudb.runtime.view.Address;
 
 import javax.annotation.concurrent.NotThreadSafe;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.PriorityQueue;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.UUID;
-import java.util.function.Supplier;
 
 /**
  * Sequencer server cache.
  * Contains transaction conflict-resolution data structures.
  * <p>
- * The cache map maps conflict keys (stream id + key) to versions (long), illustrated below:
+ * The conflictKeyMap maps conflict keys (stream id + key) to versions (long), illustrated below:
  * Conflict Key | ck1 | ck2 | ck3 | ck4
  * Version | v1 | v1 | v2 | v3
- * Consider the case where we need to insert a new conflict key (ck), but the cache is full,
- * we need to evict the oldest conflict keys. In the example above, we can't just evict ck1,
- * we need to evict all keys that map to v1, so we need to evict ck1 and ck2,
- * this eviction policy is FIFO on the version number. The simple FIFO approach in Caffein etc doesn't work here,
- * as it may evict ck1, but not ck2. Notice that we also can't evict ck3 before the keys for v1,
- * that's because it will create holes in the resolution window and can lead to incorrect resolutions.
+ * Consider the case where we need to insert a new conflict key (ck), but the cache is full.
+ * We need to evict the oldest conflict keys. In the example above, we can't just evict ck1,
+ * we need to evict all keys that map to v1, so we need to evict ck1 and ck2. This eviction
+ * policy is FIFO on the version number. Notice that we also can't evict ck3 before the keys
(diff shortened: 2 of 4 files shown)


zfrenette:
> Looks like the old logic would have same keys in the queue, and conflictKeys will have the latest mapping.
> 
> Like queue (k1, v1. k2, v1, k1 v3), map (k1 v3, k2 v1)
> When it tries to evict v1, it will remove k1 and k2 in the conflictKeys. So (k1 v3) in the map will also get removed.
> Then it is not just a memory leak, it also has a cache loss. Am I correct?

Yes, you are right. When we poll the queue and get (k1, v1), the subsequent call to remove k1 from the map would incorrectly remove the entry with v3.

zhangn49 (approved):
LGTM.

zfrenette:
> Great fix.
> 
> I noticed that ConflictTxStream can be a hash value itself. Capturing the stream information explicitly isn't necessary for conflict resolution. Essentially all you need is :
> 
> Map: Hash(streamId, key) -> Long Map: Long -> Set<Hash(streamId, key)>

This ends up being problematic. There are cases where the conflict parameter can have different values but the same hash. One such example is (streamId, new byte[]{0, 0, 0, 31}) and (streamId, new byte[]{0, 0, 1, 0}). With ConflictTxStream, it allows the map, through equals, to distinguish between these two as separate conflict keys. However, when hashed, these become the same integer and the map incorrectly replaces one instance by the other.

test/src/test/java/org/corfudb/runtime/object/transactions/SequencerServerCacheTest.java
```diff
@@ -13,23 +13,122 @@
...
+            Field conflictKeyMapField = cache.getClass().getDeclaredField("conflictKeyMap");
+            Field versionMapField = cache.getClass().getDeclaredField("versionMap");
+            conflictKeyMapField.setAccessible(true);
+            versionMapField.setAccessible(true);
+
+            Map<ConflictTxStream, Long> conflictKeyMap =
+                    new HashMap<>((HashMap<ConflictTxStream, Long>) conflictKeyMapField.get(cache));
+
+            SortedMap<Long, Set<ConflictTxStream>> versionMap =
+                    new TreeMap<>((TreeMap<Long, LinkedHashSet<ConflictTxStream>>) versionMapField.get(cache));
```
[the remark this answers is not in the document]
zfrenette: This is verified implicitly: https://github.com/CorfuDB/CorfuDB/pull/3010#discussion_r697738592

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,38 +1,39 @@
...
  * <p>
- * The cache map maps conflict keys (stream id + key) to versions (long), illustrated below:
+ * The conflictKeyMap maps conflict keys (stream id + key) to versions (long), illustrated below:
  * Conflict Key | ck1 | ck2 | ck3 | ck4
  * Version | v1 | v1 | v2 | v3
  * Consider the case where we need to insert a new conflict key (ck), but the cache is full,
  * we need to evict the oldest conflict keys. In the example above, we can't just evict ck1,
  * we need to evict all keys that map to v1, so we need to evict ck1 and ck2,
- * this eviction policy is FIFO on the version number. The simple FIFO approach in Caffein etc doesn't work here,
+ * this eviction policy is FIFO on the version number. The simple FIFO approach in Caffeine doesn't work here,
```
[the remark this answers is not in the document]
zfrenette: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -182,42 +167,46 @@ public void invalidateUpTo(long trimMark) {
...
+        Set<ConflictTxStream> entries;
+
+        // Remove the entry if conflict key was present with an older version
+        if (val != Address.NON_ADDRESS) {
+            entries = versionMap.get(val);
+            entries.remove(conflictStream);
+            if (entries.isEmpty()) {
+                versionMap.remove(val);
+            }
+        }
```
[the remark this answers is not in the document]
zfrenette: I added it right below this (above the call to `evict`) since this isn't part of the eviction process.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -66,96 +70,77 @@
...
      * Returns the value associated with the {@code key} in this cache,
-     * or {@code null} if there is no cached value for the {@code key}.
+     * or {@code Address.NON_ADDRESS} if there is no cached value for the {@code key}.
      *
      * @param conflictKey conflict stream
      * @return global address
      */
     public Long get(ConflictTxStream conflictKey) {
-        return conflictKeys.getOrDefault(conflictKey, Address.NON_ADDRESS);
+        return conflictKeyMap.getOrDefault(conflictKey, Address.NON_ADDRESS);
```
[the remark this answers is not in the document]
zfrenette: The version information has been removed from `ConflictTxStream` since it wasn't necessary to keep.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -66,96 +70,77 @@
...
+   * Invalidate the records with the firstAddress. It could be one or multiple records
    *
    * @return the number of entries has been invalidated and removed from the cache.
    */
   private int invalidateSmallestTxVersion() {
-    ConflictTxStream firstEntry = cacheEntries.peek();
-    if (cacheEntries.size() == 0) {
-      return 0;
-    }
+      if (versionMap.isEmpty()) {
```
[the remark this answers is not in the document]
zfrenette: I renamed the relevant methods as suggested, but evicting as a function of the number of elements is awkward since we really want to evict as a function of timestamps, and we don't know ahead of time how many entries we will end up needing to evict. If we insert K elements and the cache can't fit them all, then we might end up evicting anywhere from 1 to K versions, depending on how many conflict keys are associated with each.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -710,15 +707,19 @@ private void handleAllocation(RequestMsg req, ChannelHandlerContext ctx, IServer
...
-                    .forEach((item) -> {
-                        // insert an entry with the new timestamp using the
-                        // hash code based on the param and the stream id.
-                        item.getValueList().forEach(conflictParam ->
-                                cache.put(new ConflictTxStream(getUUID(item.getKey()),
-                                        conflictParam.toByteArray(), newTail - 1)));
-                    });
+            Set<ConflictTxStream> conflictTxStreamList = tokenRequest.getTxnResolution()
+                    .getWriteConflictParamsSetList()
+                    .stream()
```
vjeko, line 712: You could do parallel stream here.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -62,159 +78,160 @@ public class SequencerServerCache implements AutoCloseable {
...
+        Preconditions.checkArgument(txVersion > smallestAddress,
+                "txVersion=%s should be larger than the smallest address=%s", txVersion, smallestAddress);
 
-        while (conflictKeys.size() > cacheSize) {
-            invalidateSmallestTxVersion();
-        }
-        return true;
+        // The provided timestamp should be greater than any previous timestamp associated provided conflict
+        // key information. This check is performed separately to avoid partially updating the cache.
+        conflictKeys.parallelStream().forEach(conflictKey -> {
```
Maithem, line 205: conflictKeys is a `HashSet` and not thread-safe. Although this just dispatches threads to do reads, I would suggest not adding it unless you can demonstrate performance gain.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -710,15 +707,19 @@ private void handleAllocation(RequestMsg req, ChannelHandlerContext ctx, IServer
...
+                    .flatMap(item -> {
+                        final UUID streamId = getUUID(item.getKey());
```
vjeko, line 714: You don't technically need this variable.

vjeko (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -296,8 +294,7 @@ private TxResolutionResponse txnCanCommit(TxResolutionInfo txInfo) {
             // for each key pair, check for conflict; if not present, check against the wildcard
             for (byte[] conflictParam : conflictParamSet) {
 
-                Long keyAddress = cache.get(new ConflictTxStream(conflictStream.getKey(),
-                        conflictParam, Address.NON_ADDRESS));
+                Long keyAddress = cache.get(new ConflictTxStream(conflictStream.getKey(), conflictParam));
```
Maithem, line 297: I think there is a correctness issue here. An abort happens when the intersection of two conflict windows is not empty. In this case, you can have two transactions intersecting at one version.
For example:

```
Version:  1     2     3     4     5     6     7
Txn1:                 S     E
Txn2:                       S     E
```

The start (S) and (E) of Txn1 intersects with Txn2 at version 4, but the transaction would not abort because of the following condition

```java

if (keyAddress > txSnapshotTimestamp.getSequence()) {

```

So shouldn't it be `>=` instead of `>`  ?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -296,8 +294,7 @@ private TxResolutionResponse txnCanCommit(TxResolutionInfo txInfo) {
             // for each key pair, check for conflict; if not present, check against the wildcard
             for (byte[] conflictParam : conflictParamSet) {
```
Maithem, line 296: Instead of going through the conflict window and checking for conflicts, it's better to check for `TX_ABORT_NEWSEQ,` or `TX_ABORT_SEQ_OVERFLOW` first. Otherwise, you can go through the conflict resolution and not fail, but fail on a `TX_ABORT_SEQ_OVERFLOW`. Reversing the check order can be faster when the cache window is moving fast.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -62,159 +78,160 @@ public class SequencerServerCache implements AutoCloseable {
...
+     *
+     * @return The number of entries that have been evicted from the cache.
+     */
+    private int evictSmallestTxVersion() {
+      final long firstAddress = firstAddress();
+      final Set<ConflictTxStream> entriesToDelete = versionMap.remove(firstAddress);
+      final int numDeletedEntries = entriesToDelete.size();
+      log.trace("evictSmallestTxVersion: items evicted {} min address {}", numDeletedEntries, firstAddress);
+      entriesToDelete.forEach(conflictKeyMap::remove);
+      maxConflictWildcard = Math.max(maxConflictWildcard, firstAddress);
```
Maithem, line 144: This caller logic that uses  `long maxConflictWildcard = cache.getMaxConflictWildcard();` assumes this is the first address in the window, but it's actually not in the window, its an evicted address. This makes the ` if (txSnapshotTimestamp.getSequence() < maxConflictWildcard)` invalid, no?

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -710,15 +707,19 @@ private void handleAllocation(RequestMsg req, ChannelHandlerContext ctx, IServer
...
+                    .getWriteConflictParamsSetList()
+                    .stream()
+                    .flatMap(item -> {
+                        final UUID streamId = getUUID(item.getKey());
+                        return item.getValueList()
+                                .stream()
+                                .map(conflictParam -> new ConflictTxStream(streamId, conflictParam.toByteArray()));
+                    }).collect(Collectors.toCollection(LinkedHashSet::new));
+
+           cache.put(conflictTxStreamList, newTail - 1);
```
Maithem, line 720: This seems incorrect, the new tail is computed as following `long newTail = globalLogTail + tokenRequest.getNumTokens();` , but the subsequent logic assumes that `tokenRequest.getNumTokens` is strictly 1 and therefore uses `newTail - 1` as the token.

Essentially, if the client requests multiple tokens, the conflict parameters will only be placed in the last token, it seems like this is more of a coincidence than design. I would suggest making these calculations more strict.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -62,159 +78,160 @@ public class SequencerServerCache implements AutoCloseable {
...
+
+            // Remove the previous entry if conflict key was present with an older version
+            if (prevVersion != null) {
+                final Set<ConflictTxStream> entries = versionMap.get(prevVersion);
+                entries.remove(conflictKey);
+                if (entries.isEmpty()) {
+                    versionMap.remove(prevVersion);
+                }
+            }
+        });
```
Maithem, line 223: Is the following correct: removing partial conflict keys from versionMap is just a memory optimization and not needed for correctness because of the key overwrite that happens in `conflictKeyMap.put(conflictKey, txVersion);` ?

Maithem: In other words, can you omit this logic and rely on evict alone. Making evict account of this might make the logic simpler. It's a little convoluted.

zfrenette: That was the original issue being fixed. Since the conflictKeyMap maintains the latest conflict key mappings,  not removing them from the versionMap (previously the priorityQueue) caused their sizes to go out of sync. Hence they are removed here, and repopulated below with the more recent timestamp.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -62,159 +78,160 @@ public class SequencerServerCache implements AutoCloseable {
...
-
-        Long val = conflictKeys.getOrDefault(conflictStream, Address.NON_ADDRESS);
-        if (val > conflictStream.txVersion) {
-            log.error("For key {} the new entry address {} is smaller than the entry " +
-                            "address {} in cache. There is a sequencer regression.",
-                    conflictStream, conflictStream.txVersion, conflictKeys.get(conflictStream));
-            return false;
-        }
+    public void put(@NonNull Set<ConflictTxStream> conflictKeys, long txVersion) {
+        // The provided timestamp should not already be present in the cache.
```
Maithem, line 194: Can you also put a check that makes sure that the conflict keys can actually fit in the cache window. In other words, the cache must hold all of the transactions conflict keys. For example, imagine the cache size = 10 and a transaction tries to commit 20 conflict keys.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
zfrenette: Relying in evict is not enough. If the cache is below its capacity (as determined by the number of entries in the conflictKeyMap), then nothing will be evicted. For example, this was the case when we repeatedly put the same KVP in the table, the sizes of both data structures just diverged.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -62,159 +78,160 @@ public class SequencerServerCache implements AutoCloseable {
...
-    public void invalidateUpTo(long trimMark) {
-        log.debug("Invalidate sequencer cache. Trim mark: {}", trimMark);
-        int entries = 0;
-        int pqEntries = 0;
-        while (Address.isAddress(firstAddress()) && firstAddress() < trimMark) {
-            pqEntries += invalidateSmallestTxVersion();
-            entries++;
+    private void evict() {
+        while (conflictKeyMap.size() > capacity) {
+            evictSmallestTxVersion();
```
Maithem, line 155: I think it might be useful to mention that the cache releases memory via two mechanisms. One, eviction and two, overwriting. Imaging the case where you are continuously over writing the same conflict keys and the number of conflict parameters is less than capacity. This method will not evict anything, but the memory release happens in the calling method via overwriting.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -1,52 +1,68 @@
...
-    private final PriorityQueue<ConflictTxStream> cacheEntries; //sorted according to address
+    /**
+     * A mapping between global-log positions and conflict keys. This is maintained to
+     * facilitate the eviction of all conflict keys associated with a particular transaction
+     * commit timestamp.
+     */
+    @VisibleForTesting
+    @Getter(AccessLevel.PUBLIC)
+    private final SortedMap<Long, Set<ConflictTxStream>> versionMap;
+
```
Maithem, line 60: I don't think this is the best data structure to use especially when the entries are being added to the collection in sorted order. For example, `versionMap.put` is called by `cache.put(conflictTxStreamList, newTail - 1);`. Since newTail - 1 is strictly increasing, you end up adding sorted input to the `SortedMap` which is not necessary. 

Based on that invariant you can use a linked list to keep the sliding window versions. I think that would simply the implementation and make more efficient.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
Maithem: Yes, I understand. I was implicitly suggesting relying on evict with extra accounting information. See https://github.com/CorfuDB/CorfuDB/pull/3010#discussion_r959016409

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
zfrenette: Since a client only ever requests one token when they go to commit, I can remove the deprecated APIs, update this part of the code and relevant tests in a subsequent PR.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
zfrenette: I used it to avoid creating duplicate UUID objects for the same stream id.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
zfrenette: These checks have been subsumed by other precondition checks after refactoring/reworking.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
zfrenette: Added this precondition check and a test for it.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
zfrenette: Thanks for pointing this out. After more thought, realized that both data structures can be combined together by using a LinkedHashMap, and gives the best of both worlds.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
```diff
@@ -62,159 +78,160 @@ public class SequencerServerCache implements AutoCloseable {
...
-        conflictKeys = MicroMeterUtils
-                .gauge(conflictKeysCounterName, new HashMap<ConflictTxStream, Long>(), HashMap::size)
+
+        this.conflictKeyMap = MicroMeterUtils
+                .gauge(CONFLICT_KEYS_COUNTER_NAME, new HashMap<ConflictTxStream, Long>(), HashMap::size)
                 .orElseGet(HashMap::new);
+
+        this.versionMap = MicroMeterUtils
+                .gauge(WINDOW_SIZE_NAME, new TreeMap<Long, Set<ConflictTxStream>>(), TreeMap::size)
+                .orElseGet(TreeMap::new);
```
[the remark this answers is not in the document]
zfrenette: From other comments made on this PR, both data structures can be combined into one by using a LinkedHashMap, while still maintaining the best of both worlds.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -710,15 +707,19 @@ private void handleAllocation(RequestMsg req, ChannelHandlerContext ctx, IServer
...
-                        // insert an entry with the new timestamp using the
-                        // hash code based on the param and the stream id.
-                        item.getValueList().forEach(conflictParam ->
-                                cache.put(new ConflictTxStream(getUUID(item.getKey()),
-                                        conflictParam.toByteArray(), newTail - 1)));
-                    });
+            Set<ConflictTxStream> conflictTxStreamList = tokenRequest.getTxnResolution()
+                    .getWriteConflictParamsSetList()
+                    .stream()
+                    .flatMap(item -> {
```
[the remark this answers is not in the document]
zfrenette: Changed `item` to `conflictKeysForStream`.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -710,15 +707,19 @@ private void handleAllocation(RequestMsg req, ChannelHandlerContext ctx, IServer
...
+                        final UUID streamId = getUUID(item.getKey());
+                        return item.getValueList()
+                                .stream()
```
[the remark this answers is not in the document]
zfrenette: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
zfrenette: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
zfrenette: Done.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
zfrenette: This section was simplified after reworking and incorporating the use of a LinkedHashMap instead.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
zfrenette: This has been reworked (only one data structure) after incorporating the use of a LinkedHashMap.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServerCache.java
zfrenette: Can you clarify what is meant here? From what I can tell, the caller does not assume that `maxConflictWildcard` is the first address in the window. This is stated explicitly in the comment above the if condition you reference.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
Maithem: Sounds good.

infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
Maithem: My example is wrong, you can ignore this comment. The transactions don't intersect because Txn1's window is [3, 4) and Txn2's window is [4, 5). Essentially, Txn2 starts its transaction with Txn1 commit address as its snapshot.

Maithem (approved):
LGTM. This correctness fix also seems to improve the sequencer loop's performance by ~10-15% 
image

PR #3012, opened 2021-08-02
Populate the ProtobufDescriptorTable using the RegistryTable.

Lujie1996:
## Overview

Description:
We recently changed the way how TableRegistry registers table schema and descriptors when opening a table (see #2966). FileDescriptorProtos are now stored in a new table call ProtobufDescriptorTable. In the case of Upgrade, when opening a table registered before Upgrade, because the table's information is only registered to RegistryTable, Corfu fails to find it in ProtobufDescriptorTable. This PR adds a function to populate and synchronize ProtobufDescriptorTable from RegistryTable.

Why should this be merged: 
Current TableRegistry has a backward compatibility issue. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
@@ -664,4 +664,44 @@ public class TableRegistry {
             streamingManager.shutdown();
         }
     }
+
+    /**
+     * Populate the ProtobufDescriptorTable using the RegistryTable.
+     * Enables backward compatibility in case of data migration.
+     */
+    public void syncProtobufDescriptorTable() {
+        log.info("Running syncProtobufDescriptorTable() ...");
+
+        int numRetries = 9;
+        while (true) {
+            if (TransactionalContext.isInTransaction()) {
+                throw new IllegalThreadStateException("openTable: Called on an existing transaction");
+            }
+            try {
+                this.runtime.getObjectsView().TXBuild().type(TransactionType.WRITE_AFTER_WRITE).build().begin();
+                registryTable.forEach((tableName, corfuRecord) ->
+                        corfuRecord.getPayload().getFileDescriptorsMap()
+                                .forEach((protoName, fileDescriptorProto) -> {
+                                    ProtobufFileName fileName = ProtobufFileName
+                                            .newBuilder().setFileName(protoName).build();
+                                    ProtobufFileDescriptor fileDescriptor = ProtobufFileDescriptor
+                                            .newBuilder().setFileDescriptor(fileDescriptorProto).build();
+                                    protobufDescriptorTable.putIfAbsent(fileName, new CorfuRecord<>(
+                                            fileDescriptor, corfuRecord.getMetadata()));
+                                }));
+                this.runtime.getObjectsView().TXEnd();
+                log.info("syncProtobufDescriptorTable: completed!");
+                break;
+            } catch (TransactionAbortedException txAbort) {
+                if (numRetries-- <= 0) {
+                    throw txAbort;
+                }
+                log.info("syncProtobufDescriptorTable: commit failed. Will retry {} times. Cause {}", numRetries, txAbort);
+            } finally {
+                if (TransactionalContext.isInTransaction()) { // Transaction failed or an exception occurred.
+                    this.runtime.getObjectsView().TXAbort(); // clear Txn context so thread can be reused.
+                }
+            }
+        }
+    }
 }
--- a/runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
+++ b/runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
@@ -139,10 +139,8 @@ public class DynamicProtobufSerializer implements ISerializer {
             String protoFileName = fileDescriptorProto.getPayload().getFileDescriptor().getName();
             // Since corfu_options is something within repo, the path gets truncated on insert.
             // However dynamicProtobufSerializer fails since the full path is needed.
-            if (protoFileName.equals("corfu_options.proto")) {
-                fdProtoMap.putIfAbsent(protoFileName, fileDescriptorProto.getPayload().getFileDescriptor());
-                // Until the truncating issue can be addressed, manually add both paths.
-                protoFileName = "corfudb/runtime/corfu_options.proto";
+            if (protoFileName.equals("corfudb/runtime/corfu_options.proto")) {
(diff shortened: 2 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -664,4 +664,44 @@ public void shutdown() {
...
+     * Populate the ProtobufDescriptorTable using the RegistryTable.
+     * Enables backward compatibility in case of data migration.
+     */
+    public void syncProtobufDescriptorTable() {
+        log.info("Running syncProtobufDescriptorTable() ...");
+
+        int numRetries = 9;
+        while (true) {
+            if (TransactionalContext.isInTransaction()) {
+                throw new IllegalThreadStateException("openTable: Called on an existing transaction");
```
annym, line 678: nit -> openTable?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -664,4 +664,44 @@ public void shutdown() {
...
+                                .forEach((protoName, fileDescriptorProto) -> {
+                                    ProtobufFileName fileName = ProtobufFileName
+                                            .newBuilder().setFileName(protoName).build();
+                                    ProtobufFileDescriptor fileDescriptor = ProtobufFileDescriptor
+                                            .newBuilder().setFileDescriptor(fileDescriptorProto).build();
+                                    protobufDescriptorTable.putIfAbsent(fileName, new CorfuRecord<>(
+                                            fileDescriptor, corfuRecord.getMetadata()));
+                                }));
+                this.runtime.getObjectsView().TXEnd();
+                log.info("syncProtobufDescriptorTable: completed!");
```
annym, line 693: is it worth to add info about # of descriptors added and tables? (for debugging) for instance, completed sync of X proto descriptors corresponding to Z tables.

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
```diff
@@ -139,10 +139,8 @@ public DynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
             String protoFileName = fileDescriptorProto.getPayload().getFileDescriptor().getName();
             // Since corfu_options is something within repo, the path gets truncated on insert.
             // However dynamicProtobufSerializer fails since the full path is needed.
-            if (protoFileName.equals("corfu_options.proto")) {
-                fdProtoMap.putIfAbsent(protoFileName, fileDescriptorProto.getPayload().getFileDescriptor());
-                // Until the truncating issue can be addressed, manually add both paths.
-                protoFileName = "corfudb/runtime/corfu_options.proto";
+            if (protoFileName.equals("corfudb/runtime/corfu_options.proto")) {
```
annym, line 142: why did the path change? was it not working before?

annym (approved):

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -664,4 +664,44 @@ public void shutdown() {
             streamingManager.shutdown();
         }
     }
+
+    /**
+     * Populate the ProtobufDescriptorTable using the RegistryTable.
+     * Enables backward compatibility in case of data migration.
+     */
+    public void syncProtobufDescriptorTable() {
```
Maithem, line 672: This is pretty hacky and I don't think its a good idea to have it as a public API.

This seems like a build-to-build migration issue, can't it be ignored?

Lujie1996: This issue can be common for any Corfu upgrade which crosses #2966. Also, it does no harm to call this API even in non-upgrade cases because it only adds protos from RegistryTable to ProtobufDescriptorTable and it doesn't overwrite any existing protos inside ProtobufDescriptorTable. Do you have any better idea to address the issue?

runtime/src/main/java/org/corfudb/util/serializer/DynamicProtobufSerializer.java
Lujie1996: I noticed  that the protoFileName changed on the setup. There is corfudb/runtime/corfu_options.proto but the corfu_options.proto is missing so I changed to this. Without this change it reports `corfu_options.proto is never seen in registry` error.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
annym: I also had the same concern and we had discussed it offline. I also see it like something specific to migration--outside of Corfu... what if we have it as an utility? Or we can have this code sitting in the clients migration code. Cause eventually clients starting from Ufo release would not need it, so it's specific to their workflows.

pankti-m: I am leaning on the side of moving this outside corfu.. because from no-one from within Corfu requires it, correct?

Will any third-party client using Corfu require this utility when upgrading from build-to-build?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -664,4 +664,44 @@ public void shutdown() {
...
+                this.runtime.getObjectsView().TXBuild().type(TransactionType.WRITE_AFTER_WRITE).build().begin();
+                registryTable.forEach((tableName, corfuRecord) ->
+                        corfuRecord.getPayload().getFileDescriptorsMap()
+                                .forEach((protoName, fileDescriptorProto) -> {
+                                    ProtobufFileName fileName = ProtobufFileName
+                                            .newBuilder().setFileName(protoName).build();
+                                    ProtobufFileDescriptor fileDescriptor = ProtobufFileDescriptor
+                                            .newBuilder().setFileDescriptor(fileDescriptorProto).build();
+                                    protobufDescriptorTable.putIfAbsent(fileName, new CorfuRecord<>(
+                                            fileDescriptor, corfuRecord.getMetadata()));
```
hisundar, line 690: It would also be useful to reset the registryTable's fileDescriptorsMap()'s values only to default instances here since they are no longer needed there. The keys in the registryTable's FileDesciptorMap should remain as is.
It is ok to do this as a follow up change too if you prefer. thanks

hisundar (approved):

Lujie1996:
Close this PR as we think it's better to add the logic into NSX instead of Corfu codebase.

Maithem:
btw, I don't think this is "hacky" because it should belong externally, but rather, the fix is addressing a side effect and not the root cause. I think the registry should have explicit APIs for upgrade.

PR #3018, opened 2021-08-05
Stream Polling Delay Period Fix

Maithem:
## Overview
This patch removes random sleeps between polling and replaces it
with a fixed sleep duration, which makes stream polling follow a
fixed delay schedule. This also reduces the load being generated
by the pollers and makes performance more predictable.

Pre-this patch when the user specifies a poll period of 50ms, the
scheduler will use a random sleep value between [0, 50), this means
that the poller will run every 24.5ms (i.e., expected value of
Random(50)) on average, in essence doubling the frequency, which is
not expected by the user.

Why should this be merged: Performance Fix

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -75,10 +75,6 @@ public class CorfuRuntime {
     @ToString
     public static class CorfuRuntimeParameters extends RuntimeParameters {
 
-        public static CorfuRuntimeParametersBuilder builder() {
-            return new CorfuRuntimeParametersBuilder();
-        }
-
         /*
          * Max size for a write request.
          */
@@ -88,15 +84,6 @@ public class CorfuRuntime {
          * Set the bulk read size.
          */
         int bulkReadSize = 10;
-
-        /*
-         * How much time the Fast Loader has to get the maps up to date.
-         *
-         * <p>Once the timeout is reached, the Fast Loader gives up. Every map that is
-         * not up to date will be loaded through normal path.
-         */
-        Duration fastLoaderTimeout = Duration.ofMinutes(30);
-        // endregion
 
         // region Address Space Parameters
         /*
@@ -248,7 +235,7 @@ public class CorfuRuntime {
         /*
          * Period of time in ms to sleep before next cycle, when poller gets no new data changes.
          */
-        private int streamingPollingIdleWaitTimeMs = 5;
+        private int streamingPollingIdleWaitTimeMs = 50;
 
         /*
          * Capacity of queue shared by by streaming polling and notification tasks.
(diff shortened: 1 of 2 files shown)


PavelZaytsev (approved):

zfrenette (approved):

Lujie1996 (approved):

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -248,7 +248,7 @@ public static CorfuRuntimeParametersBuilder builder() {
         /*
          * Period of time in ms to sleep before next cycle, when poller gets no new data changes.
          */
-        private int streamingPollingIdleWaitTimeMs = 5;
+        private int streamingPollingIdleWaitTimeMs = 50;
```
Maithem, line 251: RIP. This is the wrong parameter. Should change `streamingNotificationBlockingTimeMs` and remove this one.

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -104,7 +100,7 @@ private void pollTxnStream() throws Exception {
         if (updates.isEmpty()) {
             log.trace("pollTxStream :: no updates for {} from {}, listenerId={}", txnStream.getId(),
                     lastReadAddress + 1L, subscription.getListener().getClass().getSimpleName());
-            pollingExecutor.schedule(this, randomGenerator.nextInt(pollingIdleWaitTime), TimeUnit.MILLISECONDS);
+            pollingExecutor.schedule(this, pollingIdleWaitTime, TimeUnit.MILLISECONDS);
```
annym, line 103: should we do some sort of testing to determine what is the best number according to the clients usage pattern?

annym (approved):

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
Maithem: Yes. There are two issues here, the polling strategy and the polling frequency. The current polling strategy aims to lower the load, but it inadvertently doubles it. This PR fixes the first issue to address the second issue.

Maithem: There's a third issue where different streams might want to poll at different periods, but the current APIs forces all pollers to use the same poll period, we can introduce that in subsequent patches if they are necessary.

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/StreamPollingTask.java
```diff
@@ -46,8 +46,6 @@
     // A period of time in ms to sleep before next cycle when poller gets no new data changes.
     private final int pollingIdleWaitTime;
 
-    private final Random randomGenerator;
```
chetangudisagar, line 49: nit: Can we remove its import as well? `L13: import java.util.Random;`

pankti-m (approved):

PR #3020, opened 2021-08-10
Expose Stream Tag option as input to the Browser/Editor tool.

pankti-m:
Stream Tag was not available as an input parameter to the browser/editor tool from the CLI.  Add this option.

Why should this be merged:  To support query operations based on stream tag.


PavelZaytsev (approved):

achinmishra (approved):
Since we are deprecating "dropTable", it could be an issue with the backward compatibility nature of the editor tool. Do we have to add notes somewhere in the github about this?

pankti-m:
> Since we are deprecating "dropTable", it could be an issue with the backward compatibility nature of the editor tool. Do we have to add notes somewhere in the github about this?

No we just need to use the new command-clearTable.  There is no user of the tool from within Corfu

annym (approved):

PR #3021, opened 2021-08-10
Catch WrongClusterException on get layout method

chetangudisagar:
- This patch catches the WrongClusterException from the `getLayoutUninterruptibly()` method's `layout.get()` call.
- It is observed in a case when the server gets restarted and gets a new `clusterID`, the old client just keeps retrying with the `WrongClusterException`.
- Error message:
  ```
    AbstractView 3417 getLayoutUninterruptibly: Encountered unchecked exception. Aborting layoutHelper
    java.util.concurrent.ExecutionException: org.corfudb.runtime.exceptions.WrongClusterException: Expected: rRZQ5MveQBydt4CdJW2Qgg Actual: lcYPrdK1QV-3OZMqyg_iSQ
    ```

## Overview

Description:

Why should this be merged: It fixes the WCE error.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -1,12 +1,5 @@
 package org.corfudb.runtime.view;
 
-import java.time.Duration;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicReference;
-
-import javax.annotation.Nonnull;
-
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.exceptions.NetworkException;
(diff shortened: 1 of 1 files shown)


PavelZaytsev (approved):

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -89,7 +88,12 @@ private Layout getLayoutUninterruptibly() {
                 log.error("getLayoutUninterruptibly: Encountered error. Aborting layoutHelper", ex);
                 throw (Error) ex.getCause();
             }
-            if (ex.getCause() instanceof RuntimeException) {
+            if (ex.getCause() instanceof WrongClusterException) {
+                log.warn("layoutHelper: Cluster reconfiguration or incorrect cluster", ex.getCause());
```
hisundar, line 92: aren't you missing a placeholder in the logger?

chetangudisagar: I think it is good without the `{}`.  Since we are returning the throwable, it will print the stack trace of the exception. You had done it in L156, I copied it from there. Thanks! :)

More info: https://stackoverflow.com/a/6374166/5213405

hisundar: I always thought that applied to `log.error` and not `log.warn` but maybe I am wrong.

hisundar (approved):
LGTM

PR #3022, opened 2021-08-10
Remove equality check on protoBuf descriptors for every openTable

annym:
## Overview

Description:

If protoBuf schema is already present, do not attempt to update definitions.
These APIs should be available on upgrade path.

Why should this be merged: current client bug, as 'options' section does not match equality (serialized compared to non-serialized) when they actually do match, leading, to TransactionAbortedException's (conflict) on startup.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
@@ -233,19 +233,13 @@ public class TableRegistry {
             // Or no modification to the protobuf files.
             /**
              * Caller is opening a new table with a map of protobufFilename -> protobufFileDescriptor pairs
-             * Here are the cases that can happen:
-             * 1. No entry in TableRegistry -> New table
-             *  1.1: When inserting new protobufs into descriptor table, we find existing entries and they all match
-             *  1.2: When inserting new protobufs into descriptor table, we find mismatching [Schema Change!]
              *
-             * 2. Entry exists in TableRegistry -> Existing table re-opened:
-             *    2.1: Existing protobuf file list matches, newly inserted protobuf list
-             *    2.1.1: Each of Existing protobuf file descriptors match that in new file descriptor list
-             *    2.1.2: Existing protobuf file descriptor does not match that in new file descriptor![ SchemaChange!]
+             *  If no entry in TableRegistry, tt is a new table. Then we insert new protobufs into descriptor table.
+             *  Note: we do not replace existing descriptor, as table schema change should be handled by explicit APIs.
+             *  (assumption is that protobuf definitions do not change outside of an upgrade/migration scenario)
              *
-             *    2.2: Existing protobuf file list does not match, newly opened protobuf list [big Schema Change]
-             *    2.2.1: Existing protobuf file list matches existing proto files
-             *    2.2.3: Existing protobuf file's descriptor mismatches with the incoming descriptor
+             *  If entry exists in TableRegistry -> Existing table re-opened:
+             *  We won't check equality as the expectation is that schemas do not change outside of upgrade paths.
              */
             if (TransactionalContext.isInTransaction()) {
                 throw new IllegalThreadStateException("openTable: Called on an existing transaction");
@@ -253,37 +247,12 @@ public class TableRegistry {
             try {
                 this.runtime.getObjectsView().TXBuild().type(TransactionType.WRITE_AFTER_WRITE).build().begin();
                 CorfuRecord<TableDescriptors, TableMetadata> oldRecord = this.registryTable.get(tableNameKey);
-                if (oldRecord == null) { // Case 1 above, new unseen table
+                if (oldRecord == null) {
                     this.registryTable.put(tableNameKey,
                         new CorfuRecord<>(tableDescriptors, metadataBuilder.build()));
-                } else {
-                    // Case 2 above, re-open of existing table
-                    // If the schema has changed and/or there is an upgrade
-                    // from a previous version which did not contain the Any
-                    // fields for key/value/metadata, we need to update the
-                    // existing record in registry table.
-                    boolean schemaChanged = false;
-                    boolean upgradeCase = false;
 
-                    if (!oldRecord.getPayload().getFileDescriptorsMap().keySet()
-                        .equals(tableDescriptors.getFileDescriptorsMap().keySet())) {
-                        log.warn("Protobuf file list changed for {} table {}. Prev list {}, new list {}",
-                            namespace, tableName, oldRecord.getPayload().getFileDescriptorsMap().keySet(),
-                            tableDescriptors.getFileDescriptorsMap().keySet());
-                        schemaChanged = true;
-                    }
-
-                    if (!oldRecord.getPayload().hasKey()) {
-                        upgradeCase = true;
-                    }
-
-                    if (schemaChanged || upgradeCase) {
-                        this.registryTable.put(tableNameKey,
-                            new CorfuRecord<>(tableDescriptors, metadataBuilder.build()));
-                    } // else no need to re-add the exact same entry into the table
+                    allDescriptors.forEach(this::recordNewSchema);
                 }
-                allDescriptors.forEach((protoName, protoDescriptor) ->
-                    recordNewSchema(protoName, protoDescriptor, namespace, tableName));
                 this.runtime.getObjectsView().TXEnd();
                 break;
             } catch (TransactionAbortedException txAbort) {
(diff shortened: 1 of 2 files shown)


zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -235,17 +235,12 @@ void registerTable(@Nonnull String namespace,
...
              * 2. Entry exists in TableRegistry -> Existing table re-opened:
-             *    2.1: Existing protobuf file list matches, newly inserted protobuf list
-             *    2.1.1: Each of Existing protobuf file descriptors match that in new file descriptor list
-             *    2.1.2: Existing protobuf file descriptor does not match that in new file descriptor![ SchemaChange!]
-             *
-             *    2.2: Existing protobuf file list does not match, newly opened protobuf list [big Schema Change]
-             *    2.2.1: Existing protobuf file list matches existing proto files
-             *    2.2.3: Existing protobuf file's descriptor mismatches with the incoming descriptor
+             *    2.1: Existing protoBuf file list is present, newly inserted protoBuf list (we don't check for equality/
+             *    match as the expectation is that schemas do nor change outside of upgrade paths).
```
pankti-m, line 243: nit - typo.  nor->not

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -256,34 +251,10 @@ void registerTable(@Nonnull String namespace,
                 if (oldRecord == null) { // Case 1 above, new unseen table
                     this.registryTable.put(tableNameKey,
                         new CorfuRecord<>(tableDescriptors, metadataBuilder.build()));
-                } else {
```
pankti-m, line 259: how do we plan to handle the case of schema change on upgrade?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -235,17 +235,12 @@ void registerTable(@Nonnull String namespace,
              * Caller is opening a new table with a map of protobufFilename -> protobufFileDescriptor pairs
              * Here are the cases that can happen:
              * 1. No entry in TableRegistry -> New table
-             *  1.1: When inserting new protobufs into descriptor table, we find existing entries and they all match
-             *  1.2: When inserting new protobufs into descriptor table, we find mismatching [Schema Change!]
+             *  1.1: When inserting new protobufs into descriptor table, we find existing entries, we don't replace
```
Maithem, line 238: Can you rewrite this whole comment. There is no validation here (old and new code): it doesn't reflect the code.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -256,34 +251,10 @@ void registerTable(@Nonnull String namespace,
...
-
-                    if (!oldRecord.getPayload().hasKey()) {
-                        upgradeCase = true;
-                    }
-
-                    if (schemaChanged || upgradeCase) {
-                        this.registryTable.put(tableNameKey,
-                            new CorfuRecord<>(tableDescriptors, metadataBuilder.build()));
-                    } // else no need to re-add the exact same entry into the table
+
```
Maithem, line 254: Can you fail here ?

When would you have to calls that try to register the same table with different `tableDescriptors` ?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
Maithem: Upgrade needs explicit APIs, we should try to handle it transparently. There could be correctness issues if we do. For example, when two JVMs using two different versions of the same table try to write a key transactionally, the sequencer wont detect conflicts because the serialization is different across versions.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -235,17 +235,12 @@ void registerTable(@Nonnull String namespace,
...
              * 2. Entry exists in TableRegistry -> Existing table re-opened:
-             *    2.1: Existing protobuf file list matches, newly inserted protobuf list
-             *    2.1.1: Each of Existing protobuf file descriptors match that in new file descriptor list
-             *    2.1.2: Existing protobuf file descriptor does not match that in new file descriptor![ SchemaChange!]
-             *
-             *    2.2: Existing protobuf file list does not match, newly opened protobuf list [big Schema Change]
-             *    2.2.1: Existing protobuf file list matches existing proto files
-             *    2.2.3: Existing protobuf file's descriptor mismatches with the incoming descriptor
+             *    2.1: Existing protoBuf file list is present, newly inserted protoBuf list (we don't check for equality/
+             *    match as the expectation is that schemas do not change outside of upgrade paths).
```
Maithem, line 243: Can you add a test that exercises the register table code path that verifies that the second registration is omitted.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
zhangn49: done!

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
zhangn49: We are not able to fail here right now, as the issue we observed will cause a false unequal.

Existing format is like `5: "messaging"` -> field name is replaced by the field number 
Incoming format is like `stream_tag: "messaging"`

So I added a trace level log here as a temporary solution.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
zhangn49: Added a test in `checkProtoDeduplication`
It compares the descriptor map's value before and after a schema update.

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
Maithem: If you ignore upgrade - which this patch doesnt support - when does this case happen ? Why is the same table being opened with different options ? these options are persisted, so it doesnt make sense to have one copy of the table with multiple options, but only persist one set of options.

zhangn49: @Maithem  Same table are opened with the same options, the difference is how they get encoded.

Essentially they represent the same schema, but right now `equals` does not treat them same. So Trying to skip the check and we can fail here once we figure out how to detect them as the same schema

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -233,57 +233,26 @@ void registerTable(@Nonnull String namespace,
...
-             * Here are the cases that can happen:
-             * 1. No entry in TableRegistry -> New table
-             *  1.1: When inserting new protobufs into descriptor table, we find existing entries and they all match
-             *  1.2: When inserting new protobufs into descriptor table, we find mismatching [Schema Change!]
              *
-             * 2. Entry exists in TableRegistry -> Existing table re-opened:
-             *    2.1: Existing protobuf file list matches, newly inserted protobuf list
-             *    2.1.1: Each of Existing protobuf file descriptors match that in new file descriptor list
-             *    2.1.2: Existing protobuf file descriptor does not match that in new file descriptor![ SchemaChange!]
+             *  If no entry in TableRegistry, tt is a new table. Then we insert new protobufs into descriptor table.
```
pankti-m, line 237: nit - typo.  tt -> it

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
pankti-m: @zhangn49, how did this encoding change?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -307,37 +276,26 @@ void registerTable(@Nonnull String namespace,
...
-                                    String namespace,
-                                    String tableName) {
-        final CorfuRecord<ProtobufFileDescriptor, TableMetadata> oldProto = this.protobufDescriptorTable.get(protoName);
-        if (oldProto == null) {
+                                    CorfuRecord<ProtobufFileDescriptor, TableMetadata> newProtoFd) {
+        final CorfuRecord<ProtobufFileDescriptor, TableMetadata> currentSchema = this.protobufDescriptorTable.get(protoName);
+        if (currentSchema == null) {
+            // If schema is not present, add to protobufDescriptorTable, otherwise,
+            // we assume schema definitions only change between upgrades/migration for which
+            // dedicated APIs are available
```
pankti-m, line 286: are -> will be.

maybe we should add TODOs in the code for this...

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -1054,28 +1056,34 @@ public void checkProtoDeduplication() throws Exception {
...
+                corfuRuntime.getTableRegistry().getProtobufDescriptorTable().size();
+        CorfuStoreMetadata.TableName tableName = CorfuStoreMetadata.TableName
+                .newBuilder()
+                .setNamespace(someNamespace)
+                .setTableName(tableNamePrefix+"0")
+                .build();
+        final Collection<CorfuRecord<ProtobufFileDescriptor, CorfuStoreMetadata.TableMetadata>> records =
+                corfuRuntime.getTableRegistry().getProtobufDescriptorTable().values();
+
+        shimStore.openTable(
```
pankti-m, line 1069: nit - consider adding a comment here saying that we open one of the tables with a different schema and validate no checks for schema update are made.

pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
annym: @pankti-m it gets serialized once written, so then the serialized form (the one persisted) is compared to the non-serialized form (the one being freshly read from the proto file), and the equals "fails" to detect they are the same, only that one is serialized, the other isn't.

Maithem: Are you saying that the value being read back is failing the equality check because its a different references?


If the answer is yes, then it still doesn't make sense. The following code does a value comparison

```java

oldRecord.getPayload().getFileDescriptorsMap().keySet()
                        .equals(tableDescriptors.getFileDescriptorsMap().keySet()
```

since the equals method is overloaded in `FileDescriptorProto` and it actually does a value comparison, so even if the references are different it would return true if the nested values are equal.

zhangn49: @Maithem 
No, they are different values - like "5" vs "stream_tag"

5 is the field number of "stream_tag"

message {
    stream_tag = 5;
}

So they are different values and forms.

Maithem: Ok, I get the problem now.

I still think silently omitting the case is problematic and can introduce subtle issues when a new schema is being registered, but omitted because an old one already exists. It should fail, but it doesn't and now the application can start up and use an old schema where they think that the new schema is being used (because the "Registration passes"). 

Omitting the second registration implicitly assume that the second registration will equal the first schema being registered. I think we should fail explicitly. The comparison can be achieved by normalizing both forms. You can serialize/deserialzie both old/new and check for equality. This is not efficient, but its also not on a hot path.

It's better to fail explicitly than assume it wont happen.

PR #3023, opened 2021-08-16
Log cleanup

PavelZaytsev:
## Overview

Description:
- Reduce client, server and metrics logs as well as regular logs

### DIFF
--- a/common/src/main/java/org/corfudb/common/metrics/micrometer/MicroMeterUtils.java
+++ b/common/src/main/java/org/corfudb/common/metrics/micrometer/MicroMeterUtils.java
@@ -1,5 +1,6 @@
 package org.corfudb.common.metrics.micrometer;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.DistributionSummary;
@@ -136,6 +137,12 @@ public class MicroMeterUtils {
     }
 
     private static List<Tag> toTagIterable(String... tags) {
+        if (tags.length % 2 != 0) {
+            throw new IllegalArgumentException("Only key-value pairs allowed.");
+        }
+        if (tags.length == 0) {
+            return ImmutableList.of();
+        }
         return IntStream.range(1, tags.length)
                 .filter(i -> i % 2 != 0)
                 .mapToObj(i -> Tag.of(tags[i - 1], tags[i]))
@@ -143,30 +150,16 @@ public class MicroMeterUtils {
     }
 
     public static <T> Optional<T> gauge(String name, T state, ToDoubleFunction<T> valueFunction, String... tags) {
-        if (tags.length % 2 != 0) {
-            throw new IllegalArgumentException("Only key-value pairs allowed.");
-        }
         return filterGetInstance(name).map(registry -> {
-            if (tags.length != 0) {
-                List<Tag> tagsList = toTagIterable(tags);
-                return registry.gauge(name, tagsList, state, valueFunction);
-            }
-
-            return registry.gauge(name, state, valueFunction);
+            List<Tag> tagsList = toTagIterable(tags);
+            return registry.gauge(name, tagsList, state, valueFunction);
         });
     }
 
     public static <T extends Number> Optional<T> gauge(String name, T state, String... tags) {
-        if (tags.length % 2 != 0) {
-            throw new IllegalArgumentException("Only key-value pairs allowed.");
-        }
         return filterGetInstance(name).map(registry -> {
-            if (tags.length != 0) {
-                List<Tag> tagsList = toTagIterable(tags);
-                return registry.gauge(name, tagsList, state);
-            }
-
-            return registry.gauge(name, state);
+            List<Tag> tagsList = toTagIterable(tags);
+            return registry.gauge(name, tagsList, state);
         });
     }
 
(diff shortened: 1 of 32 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/RemoteMonitoringService.java
```diff
@@ -814,7 +814,7 @@ private void updateTrailingLayoutServers(
                         .committed(latestLayout.getEpoch(), latestLayout)
                         .get();
                 if (result) {
-                    log.debug("Layout Server: {} successfully patched with latest layout : {}",
+                    log.debug("Layout Server: {} patched with latest layout : {}",
```
xcchang, line 817: would it be good to keep the "successfully" here to make it more explicit?

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
```diff
@@ -118,9 +118,8 @@ public void channelRead(ChannelHandlerContext ctx, Object m) throws Exception {
 
         // Store clientID as a channel attribute.
         ctx.channel().attr(clientIdAttrKey).set(clientId);
-        log.info("channelRead: Handshake validated by Server.");
-        log.debug("channelRead: Sending handshake response: Node Id: {}, Corfu client version:" +
-                " {}, Corfu server version: {}", this.nodeId, Long.toHexString(corfuClientVersion),
+        log.debug("channelRead: Sending handshake response: Node Id: {}, client version:" +
+                " {}, server version: {}", this.nodeId, Long.toHexString(corfuClientVersion),
```
xcchang, line 122: I thought it could be better to keep the into level verification log, which only happens upon handshake so it won't heavily increase the log?

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -559,7 +535,7 @@ private LogData getLogData(LogEntry entry) {
     private Metadata parseMetadata(FileChannel fileChannel, String segmentFile) throws IOException {
         long actualMetaDataSize = fileChannel.size() - fileChannel.position();
         if (actualMetaDataSize < METADATA_SIZE) {
-            log.error("Metadata has wrong size. Actual size: {}, expected: {}",
+            log.warn("Metadata has wrong size. Actual size: {}, expected: {}",
```
xcchang, line 538: Just a question, why is this not considered an ERROR?

xcchang (commented):
LGTM, just some minor comments

xnull (approved):
looks good

common/src/main/java/org/corfudb/common/metrics/micrometer/MicroMeterUtils.java
```diff
@@ -70,6 +136,41 @@ public static void measure(double measuredValue, String name, String... tags) {
...
 
+    private static List<Tag> toTagIterable(String... tags) {
+        if (tags.length % 2 != 0) {
+            throw new IllegalArgumentException("Only key-value pairs allowed.");
+        }
+        if (tags.length == 0) {
+            return ImmutableList.of();
+        }
+        return IntStream.range(1, tags.length)
+                .filter(i -> i % 2 != 0)
```
vjeko, line 147: I don't understand, why are we filtering based on even/odd?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -286,7 +286,7 @@ private void handleException(Throwable ex, ChannelHandlerContext ctx, RequestMsg
     private void handleWrite(RequestMsg req, ChannelHandlerContext ctx, IServerRouter router) {
         LogData logData = getLogData(req.getPayload().getWriteLogRequest().getLogData());
 
-        log.debug("handleWrite: type: {}, address: {}, streams: {}",
+        log.debug("handleWrite: {}, at: {}, ids: {}",
```
vjeko, line 289: I honestly prefer the previous one better.

vjeko (approved):
LGTM

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
PavelZaytsev: Yeah, the issue here is that it's the most printed line in our entire log (maybe > 90% of lines are attributed to it) and nobody wants to remove it because apparently, it helps with debugging, so I decided to reduce it at least a little bit

common/src/main/java/org/corfudb/common/metrics/micrometer/MicroMeterUtils.java
PavelZaytsev: The gauge api accepts only the iterable of tags but the input is the array of strings. This function basically says "starting from index 1 get every even-indexed entry and couple it with a previous entry into a tag".

vjeko: In that case, zipping two streams would be more readable. I believe Guava provides `Stream<T>.zip(...)`.

infrastructure/src/main/java/org/corfudb/infrastructure/ServerHandshakeHandler.java
PavelZaytsev: I think the debug line is needed for identifying where the handshake is coming from. The first line is just redundant is that case.

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
PavelZaytsev: tbh I am not sure why this particular line shows up in the log but it does show up all the time (I did not have time to investigate). I was told that sometimes other verticals attribute errors to us due to this line but nothing has happened because of it so far. I am reducing the severity to warn here but still leaving it for future reference to investigate further.

PR #3024, opened 2021-08-16
Table: Fail delete if called on a non-existent key

hisundar:
## Overview

Description:

Why should this be merged: 
If we do not fail deletes on non-existent keys, that results in extra SMR operations which confuses stream listeners who aren't expecting duplicate delete operations.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
@@ -1,31 +1,25 @@
 package org.corfudb.runtime.collections;
 
-import com.google.protobuf.Descriptors;
 import com.google.protobuf.Message;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.protocols.wireprotocol.TokenResponse;
-import org.corfudb.runtime.CorfuOptions;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.Queue;
 import org.corfudb.runtime.object.ICorfuVersionPolicy;
-import org.corfudb.runtime.object.transactions.TransactionType;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
 import org.corfudb.runtime.view.CorfuGuidGenerator;
 import org.corfudb.util.serializer.ISerializer;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.AbstractMap;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
 import java.util.Comparator;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -42,8 +36,6 @@ import java.util.stream.Stream;
  */
 @Slf4j
 public class Table<K extends Message, V extends Message, M extends Message> {
-
-    private final CorfuRuntime corfuRuntime;
 
     private final CorfuTable<K, CorfuRecord<V, M>> corfuTable;
 
@@ -101,7 +93,6 @@ public class Table<K extends Message, V extends Message, M extends Message> {
      * @param versionPolicy           versioning policy
      * @param streamTags              set of UUIDs representing the streamTags
      */
-    @Nonnull
     public Table(@Nonnull final TableParameters<K, V, M> tableParameters,
                  @Nonnull final CorfuRuntime corfuRuntime,
                  @Nonnull final ISerializer serializer,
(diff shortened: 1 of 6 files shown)


test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -264,6 +265,48 @@ public void checkMetadataTransactions() throws Exception {
...
+                ManagedResources.class,
+                // TableOptions includes option to choose - Memory/Disk based corfu table.
+                TableOptions.builder().build());
+
+        UUID uuid1 = UUID.nameUUIDFromBytes("1".getBytes());
+        SampleSchema.Uuid key1 = SampleSchema.Uuid.newBuilder()
+                .setMsb(uuid1.getMostSignificantBits()).setLsb(uuid1.getLeastSignificantBits())
+                .build();
+        ManagedResources user_1 = ManagedResources.newBuilder().setCreateUser("user_1").build();
+        long expectedVersion = 0L;
```
xcchang, line 302: looks like these variables are unused

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -169,10 +170,13 @@ public Table(@Nonnull final TableParameters<K, V, M> tableParameters,
...
      *
      * @param key Key.
-     * @return Previously stored Corfu Record.
+     * @return Previously stored Corfu Record. Throws NoSuchElementException if key is not found.
      */
     @Nullable
     CorfuRecord<V, M> deleteRecord(@Nonnull final K key) {
+        if (!corfuTable.containsKey(key)) {
+            throw new NoSuchElementException("deleteRecord failed as key "+key.toString()+" does not exist");
+        }
```
xcchang, line 179: might be good to remove the unused fields and imports at this time?

xcchang (approved):
LGTM!

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
hisundar: done, thanks

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
hisundar: done, thanks

zhangn49 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -169,10 +159,13 @@ public Table(@Nonnull final TableParameters<K, V, M> tableParameters,
      * Delete a record mapped to the specified key.
      *
      * @param key Key.
-     * @return Previously stored Corfu Record.
+     * @return Previously stored Corfu Record. Throws NoSuchElementException if key is not found.
      */
     @Nullable
     CorfuRecord<V, M> deleteRecord(@Nonnull final K key) {
+        if (!corfuTable.containsKey(key)) {
+            throw new NoSuchElementException("deleteRecord failed as key "+key.toString()+" does not exist");
```
pankti-m, line 167: will it be thrown to the listeners?  If so, will they need to resubscribe?

Lujie1996 (approved):

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
hisundar: no, this is a CRUD api, it is not part of streaming

vjeko, line 167: It's not clear to me that throwing a NoSuchElementException is a better choice than making it a noop.

hisundar: Exactly, we did argue that delete is an idempotent operation, but the main problem appears to be with *some* stream listeners who may do something potentially incorrect when they see a delete operation on a deleted key. 
So a bug was filed asking corfu to just fail the delete if the key does not exist in the hope that some wrong code path (double delete) can be found using this fix.

vjeko: You don't need to execute a delete on a deleted key. You can simply log it and return, rather than throwing the exception. This way none of the subscribers are going to see a double delete.

```
if (!corfuTable.containsKey(key)) {
    log.warn("Deleting a non-existent key.");
    return CorfuRecord.NON_EXISTENT;
}
```

It all boils down to (I) if this behavior is going to cause more noise in form of spurious bugs and (II) if this API ends up being clumsy to program against. Anyways, you have more context to make this decision.

vjeko (approved):

zfrenette (approved):

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
hisundar: From inspection of the consumers, the return value from delete is not used. Also an exception was explicitly requested so that any caller who is not checking for key existence can be flagged and fixed, but I see your point. Let's perhaps quickly sync up.

PR #3029, opened 2021-08-19
TxnContext: Performance enhancements for read path

hisundar:
## Overview

Description:

Why should this be merged: 

Read-only transactions are expected to be very fast
and they could be invoked several times.
So the fixed cost overhead for setup and teardown
needs to be optimized.
Remove commitCallbacks from reads since they only apply to writes.
Move logs to trace from debug.

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
@@ -108,8 +108,7 @@ public class TxnContext implements AutoCloseable {
             return;
         }
 
-        // Consider moving this to trace after stability improves.
-        log.debug("TxnContext: begin transaction in namespace {}", namespace);
+        log.trace("TxnContext: begin transaction in namespace {}", namespace);
         this.startTxSample = MeterRegistryProvider.getInstance().map(Timer::start);
         Transaction.TransactionBuilder transactionBuilder = this.objectsView
                 .TXBuild()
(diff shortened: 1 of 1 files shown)


pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/collections/TxnContext.java
```diff
@@ -830,19 +831,21 @@ public Timestamp commit() {
...
-        tablesInTxn.forEach((uuid, table) -> {
-            List<CorfuStreamEntry> writesInTable = writeSet.getSMRUpdates(uuid).stream().map(entry ->
-                    CorfuStreamEntry.fromSMREntry(entry, 0)).collect(Collectors.toList());
-            mutations.put(table.getFullyQualifiedTableName(), writesInTable);
-        });
-        commitCallbacks.forEach(cb -> cb.onCommit(mutations));
-        operations.clear();
+        log.trace("Txn committed on namespace {}", namespace);
+
+        if (txnType != READ_ONLY) {
```
pankti-m, line 836: will it make a big difference?  any table which does not have writes gets skipped in the loop below.  So it is already a no-op.. right?

hisundar: the loop only iterates over the writeSet but it unconditionally invokes the callback for all registered commitCallbacks.
the callbacks have fixed costs overheads which are completely unnecessary for reads

PavelZaytsev (approved):

PR #3032, opened 2021-08-20
Avoid sharing custom serializers among runtime instances

Lujie1996:
Different corfu runtime instances can have different custom serializers based on
the opened tables. The existing serializers is a static member shared by all
corfu runtimes in a JVM which is problematic when multiple runtimes register
their own custom serialzier.

## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
+++ b/runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
@@ -85,6 +85,7 @@ public class LogData implements IMetadata, ILogData {
         if (value instanceof LogEntry) {
             if (!Address.isAddress(((LogEntry) value).getGlobalAddress())) {
                 ((LogEntry) value).setGlobalAddress(getGlobalAddress());
+                ((LogEntry) value).setRuntime(runtime);
             }
             return value;
         }
--- a/runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
+++ b/runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
@@ -41,14 +41,14 @@ public class Serializers {
         serializersMap.put(QUEUE_SERIALIZER.getType(), QUEUE_SERIALIZER);
     }
 
-    private static final Map<Byte, ISerializer> customSerializers = new HashMap<>();
+    private final Map<Byte, ISerializer> customSerializers = new HashMap<>();
 
     /**
      * Return the serializer byte.
      * @param type A byte that tags a serializer
      * @return     A serializer that corresponds to the type byte
      */
-    public static ISerializer getSerializer(Byte type) {
+    public ISerializer getSerializer(Byte type) {
         if (type <= SYSTEM_SERIALIZERS_COUNT) {
             if (serializersMap.containsKey(type)) {
                 return serializersMap.get(type);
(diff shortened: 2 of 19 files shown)


xnull (approved):

PavelZaytsev (approved):

infrastructure/src/test/java/org/corfudb/infrastructure/management/FailureDetectorTest.java
```diff
@@ -40,7 +40,7 @@ public void testPollRound() {
         );
         Duration time = Duration.ofMillis(System.currentTimeMillis() - start);
         assertThat(time).isGreaterThan(Duration.ofMillis(450));
-        assertThat(time).isLessThan(Duration.ofSeconds(2));
+        assertThat(time).isLessThan(Duration.ofSeconds(3));
```
Maithem, line 43: Can you do this in a different PR ?

Maithem (approved):

infrastructure/src/test/java/org/corfudb/infrastructure/management/FailureDetectorTest.java
Lujie1996: Done

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
```diff
@@ -85,6 +85,7 @@ public Object getPayload(CorfuRuntime runtime) {
         if (value instanceof LogEntry) {
             if (!Address.isAddress(((LogEntry) value).getGlobalAddress())) {
                 ((LogEntry) value).setGlobalAddress(getGlobalAddress());
+                ((LogEntry) value).setRuntime(runtime);
```
Maithem, line 88: Can you just pass the serializer here directly?

Maithem (commented):
Instead of passing the CorfuRuntime to every method that needs to serialize/deserialize I suggest we just pass the ISerializer object directly.

achinmishra (approved):
lgtm

runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
```diff
@@ -41,14 +41,14 @@ public static final ISerializer getDefaultSerializer() {
         serializersMap.put(QUEUE_SERIALIZER.getType(), QUEUE_SERIALIZER);
     }
 
-    private static final Map<Byte, ISerializer> customSerializers = new HashMap<>();
+    private final Map<Byte, ISerializer> customSerializers = new HashMap<>();
```
pankti-m, line 44: should we still keep it as a map?  can a runtime use >1 custom serializers?

pankti-m (approved):

runtime/src/main/java/org/corfudb/protocols/wireprotocol/LogData.java
Lujie1996: I thought about it, but then I found that SMREntry calls `serializerType.deserialize(objBuf, runtime)` and JsonSerializer will need runtime to build an object

`rt.getObjectsView().build()
                        .setStreamID(new UUID(b.readLong(), b.readLong()))
                        .setType((Class<? extends ICorfuSMR>) Class.forName(smrClassName))
                        .open();`

runtime/src/main/java/org/corfudb/util/serializer/Serializers.java
Lujie1996: I think it's a map now, right?

pankti-m: I meant, can we remove the map?  Why should it be a map?

PR #3038, opened 2021-08-31
Add numOfEntries as a threshold in checkpointer's batching strategy

Lujie1996:
When checkpointer batches SMR entries, comparing only serialized and compressed
entry size could cause OOM since it might read too many entries into memory before
serialization and compression. Add new threshold numOfEntries as a quick fix.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
@@ -59,6 +59,14 @@ public class CheckpointWriter<T extends StreamingMap> {
     private double batchThresholdPercentage = 0.95;
     private static final int DEFAULT_CP_MAX_WRITE_SIZE = 25 * (1 << 20);
 
+    private static final double MEMORY_USAGE_WARNING_THRESHOLD = 0.95;
+
+    /** Batch size: number of SMREntry in a single CONTINUATION.
+     */
+    @Getter
+    @Setter
+    private int batchSize = 50;
+
     @SuppressWarnings("checkstyle:abbreviation")
     private final UUID checkpointStreamID;
     private final Map<CheckpointEntry.CheckpointDictKey, String> mdkv = new HashMap<>();
@@ -100,6 +108,8 @@ public class CheckpointWriter<T extends StreamingMap> {
     @Setter
     ISerializer serializer = Serializers.getDefaultSerializer();
 
+    private final Runtime javaRuntime = Runtime.getRuntime();
+
     /** Constructor for Checkpoint Writer for Corfu Maps.
      * @param rt object's runtime
      * @param streamId unique identifier of stream to checkpoint
@@ -114,6 +124,7 @@ public class CheckpointWriter<T extends StreamingMap> {
         checkpointId = UUID.randomUUID();
         checkpointStreamID = CorfuRuntime.getCheckpointStreamIdFromId(streamId);
         sv = rt.getStreamsView();
+        batchSize = rt.getParameters().getCheckpointBatchSize();
     }
 
     /**
@@ -288,7 +299,7 @@ public class CheckpointWriter<T extends StreamingMap> {
              * of SMR entries. Its a safeguard against the smr entries amounting to the actual
              * boundary limit.
              */
-            if (numBytesPerCheckpointEntry > maxWriteSizeLimit) {
+            if (numBytesPerCheckpointEntry > maxWriteSizeLimit || smrEntries.getUpdates().size() >= batchSize) {
                 convertAndAppendCheckpointEntry(smrEntries, kvCopy);
                 log.trace("Batched size of checkpoint log entry consists {} smr entries",
                         smrEntries.getUpdates().size());
@@ -301,6 +312,13 @@ public class CheckpointWriter<T extends StreamingMap> {
             // maintain current batch size only for test purposes.
             totalEntryCount++;
             inputBuffer.clear();
+
+            // Remove this warning in production. It's only for the purpose of experiments
+            // Calculation of usedMemPercentage is based on https://stackoverflow.com/questions/12807797/java-get-available-memory
+            double usedMemPercentage = ((double) javaRuntime.totalMemory() - javaRuntime.freeMemory()) / javaRuntime.maxMemory();
+            if (usedMemPercentage >= MEMORY_USAGE_WARNING_THRESHOLD) {
+                log.warn("Checkpointer detected high memory usage {}!", usedMemPercentage);
+            }
         }
 
         // the entries which are left behind for a final flush.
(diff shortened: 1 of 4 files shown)


Maithem (commented):
I don't understand the git commit message. 

Does this happen when the JVM compressing the buffer has more available memory than the JVM decompressing the buffer?

Lujie1996:
> I don't understand the git commit message.
> 
> Does this happen when the JVM compressing the buffer has more available memory than the JVM decompressing the buffer?

@Maithem Think about how we do batching - keep reading SMR entries into memory (`smrEntries.addTo(smrPutEntry);`) until its serialized and compressed size reaches 25MB. We know the compression ratio can be as high as thousands, which means the `smrEntries` can be 25MB * 1000 which will trigger OOM.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -288,7 +299,7 @@ public int appendObjectState(Stream<Map.Entry> entryStream) {
              * of SMR entries. Its a safeguard against the smr entries amounting to the actual
              * boundary limit.
              */
-            if (numBytesPerCheckpointEntry > maxWriteSizeLimit) {
+            if (numBytesPerCheckpointEntry > maxWriteSizeLimit || smrEntries.getUpdates().size() >= batchSize) {
```
Maithem, line 302: Why don't you just let it fill the maxWriteSizeLimit buffer, but remove the compression from the calculation? I think that's cleaner. 

The batch size parameter needs to be tuned to match the max write limit, if you just base your calculation on the uncompressed size then you can get rid of this tuning problem all together.

Maithem:
> > I don't understand the git commit message.
> > Does this happen when the JVM compressing the buffer has more available memory than the JVM decompressing the buffer?
> 
> @Maithem Think about how we do batching - keep reading SMR entries into memory (`smrEntries.addTo(smrPutEntry);`) until its serialized and compressed size reaches 25MB. We know the compression ratio can be as high as thousands, which means the `smrEntries` can be 25MB * 1000 which will trigger OOM.

So this issue happens with non-checkpointer-jvms try to sync that LogData, right?

The checkpointer JVM itself shouldn't fail on the decompression, since it had enough memory to compress the uncompressed buffer.

Lujie1996:
> > > I don't understand the git commit message.
> > > Does this happen when the JVM compressing the buffer has more available memory than the JVM decompressing the buffer?
> > 
> > 
> > @Maithem Think about how we do batching - keep reading SMR entries into memory (`smrEntries.addTo(smrPutEntry);`) until its serialized and compressed size reaches 25MB. We know the compression ratio can be as high as thousands, which means the `smrEntries` can be 25MB * 1000 which will trigger OOM.
> 
> So this issue happens with non-checkpointer-jvms try to sync that LogData, right?
> 
> The checkpointer JVM itself shouldn't fail on the decompression, since it had enough memory to compress the uncompressed buffer.

Can you explain why compactor JVM always have enough memory to compress the buffer? There is no limit on the uncompressed buffer size. Until they are compressed into a checkpoint entry they will take memory space. Am I missing anything?

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Lujie1996: I agree that comparing serialized but uncompressed size could help stay away from such OOM headache. The original idea to calculate the compressed size is because we want to batch as much as possible which in theory could provide some performance improvement.

Lujie1996:
> > > > > I don't understand the git commit message.
> > > > > Does this happen when the JVM compressing the buffer has more available memory than the JVM decompressing the buffer?
> > > > 
> > > > 
> > > > @Maithem Think about how we do batching - keep reading SMR entries into memory (`smrEntries.addTo(smrPutEntry);`) until its serialized and compressed size reaches 25MB. We know the compression ratio can be as high as thousands, which means the `smrEntries` can be 25MB * 1000 which will trigger OOM.
> > > 
> > > 
> > > So this issue happens with non-checkpointer-jvms try to sync that LogData, right?
> > > The checkpointer JVM itself shouldn't fail on the decompression, since it had enough memory to compress the uncompressed buffer.
> > 
> > 
> > Can you explain why compactor JVM always have enough memory to compress the buffer? There is no limit on the uncompressed buffer size. Until they are compressed into a checkpoint entry they will take memory space. Am I missing anything?
> 
> In order to compress a buffer, the whole uncompressed buffer must first fit in memory. The fact it is able to compress a 100mb buffer into a 1mb buffer, implies that the JVM has enough memory to hold the uncompressed buffer.

`In order to compress a buffer, the whole uncompressed buffer must first fit in memory` that's exactly where the problem is. The uncompressed buffer can't fit in memory in current code in some cases I have seen. The uncompressed buffer keeps growing until its compressed size exceeds 25MB, in this process the uncompressed buffer could already use up all JVM heap space.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
Maithem: The initial problem - as I remember it - was that using number of entries to determine batch size was generating LogData entries that were to large to send over RPC. The initial solution shouldn't have used the compressed bytes in its calculation.

If the uncompressed size can fit in an RPC message, and a client is able to receive this message, then it can be sure that what ever is decompressed will be smaller than the largest message it can receive.

runtime/src/main/java/org/corfudb/runtime/CheckpointWriter.java
```diff
@@ -59,6 +59,14 @@
     private double batchThresholdPercentage = 0.95;
     private static final int DEFAULT_CP_MAX_WRITE_SIZE = 25 * (1 << 20);
 
+    private static final double MEMORY_USAGE_WARNING_THRESHOLD = 0.95;
```
achinmishra, line 62: you can use the same variable `batchThresholdPercentage` for this.

Maithem (approved):

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -273,6 +273,11 @@
          */
         int checkpointRetries = 5;
 
+        /*
+         * The maximum number of SMR entries that will be grouped in a CheckpointEntry.CONTINUATION
+         */
+        int checkpointBatchSize = 50;
```
pankti-m, line 279: did we come up with the number 50 empirically or some other strategy?

Lujie1996: Originally it was set to 50 (empirically based on entry size and 25mb write size limit) until a few months back we removed this parameter. We need it because without it if the compression ratio is high then we end up batching too many entries.

pankti-m: ok

pankti-m (approved):

hisundar (approved):

PR #3040, opened 2021-09-07
Adding metrics: avg-entry-size in readCache and size of readCache.

shama358:
## Overview

Description:

Why should this be merged:  This change will help to tune the address space cache without having to analyze the heap dump

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
@@ -53,6 +53,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
@@ -66,6 +67,7 @@ import java.util.stream.Collectors;
 public class AddressSpaceView extends AbstractView {
 
     private final static long DEFAULT_MAX_CACHE_ENTRIES = 5000;
+    private AtomicLong currentDataSizeInCache;
 
     /**
      * A cache for read results.
@@ -90,6 +92,7 @@ public class AddressSpaceView extends AbstractView {
         final long maxCacheEntries = runtime.getParameters().getMaxCacheEntries();
         final long maxCacheWeight = runtime.getParameters().getMaxCacheWeight();
         final int concurrencyLevel = runtime.getParameters().getCacheConcurrencyLevel();
+        currentDataSizeInCache = new AtomicLong(0);
 
         if (maxCacheWeight != 0) {
             throw new UnsupportedOperationException("Weighing cache is not supported!");
@@ -116,6 +119,9 @@ public class AddressSpaceView extends AbstractView {
 
         Optional<MeterRegistry> metricsRegistry = MeterRegistryProvider.getInstance();
         MicroMeterUtils.gauge("address_space.read_cache.hit_ratio", readCache, cache -> cache.stats().hitRate());
+        MicroMeterUtils.gauge("address_space.read_cache.estimated_avg_entry_size",
+                readCache.size() != 0 ? (currentDataSizeInCache.get() / readCache.size()) : 0);
+        MicroMeterUtils.gauge("address_space.read_cache.cache_size", readCache, cache -> cache.size());
         metricsRegistry.map(registry -> GuavaCacheMetrics.monitor(registry, readCache, "address_space.read_cache"));
     }
 
@@ -123,6 +129,11 @@ public class AddressSpaceView extends AbstractView {
         if (log.isTraceEnabled()) {
             log.trace("handleEviction: evicting {} cause {}", notification.getKey(), notification.getCause());
         }
+        updateCurrentDataSizeInCache(currentDataSizeInCache.get(), (notification.getValue().getSizeEstimate()) * -1);
+    }
+
+    private void updateCurrentDataSizeInCache(long expectedSize, int delta) {
+        currentDataSizeInCache.compareAndSet(expectedSize, expectedSize + delta);
     }
 
 
@@ -239,6 +250,7 @@ public class AddressSpaceView extends AbstractView {
             // Cache the successful write
             if (cacheOption == CacheOption.WRITE_THROUGH) {
                 readCache.put(token.getSequence(), ld);
+                updateCurrentDataSizeInCache(currentDataSizeInCache.get(), ld.getSizeEstimate());
             }
         };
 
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -114,15 +117,30 @@ public AddressSpaceView(@Nonnull final CorfuRuntime runtime) {
...
+
+    private void updateCurrentDataSizeInCache(long expectedSize, int delta) {
+        currentDataSizeInCache.compareAndSet(expectedSize, expectedSize + delta);
+    }
+
+    private double calculateEstimatedAvgEntrySize() {
+        if (readCache.size() == 0) {
+            return 0;
+        }
+        return currentDataSizeInCache.get() / readCache.size();
```
PavelZaytsev, line 143: Integer division in floating-point context, cast one of the operands to a double

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -114,15 +117,30 @@ public AddressSpaceView(@Nonnull final CorfuRuntime runtime) {
...
 
     private void handleEviction(RemovalNotification<Long, ILogData> notification) {
         if (log.isTraceEnabled()) {
             log.trace("handleEviction: evicting {} cause {}", notification.getKey(), notification.getCause());
         }
+        updateCurrentDataSizeInCache(currentDataSizeInCache.get(),
+                (notification.getValue().getSizeEstimate()) * -1);
+    }
+
+    private void updateCurrentDataSizeInCache(long expectedSize, int delta) {
```
PavelZaytsev, line 135: You can just replace it with currentDataSizeInCache.addAndGet(value), then you won't need a separate function for it.

PavelZaytsev (commented):
Looks good, minor comments

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
shama358: done

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
shama358: done

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -66,6 +67,7 @@
 public class AddressSpaceView extends AbstractView {
 
     private final static long DEFAULT_MAX_CACHE_ENTRIES = 5000;
+    private AtomicLong currentDataSizeInCache;
```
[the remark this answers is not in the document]
shama358: done

PavelZaytsev (approved):

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -239,6 +252,7 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
             // Cache the successful write
             if (cacheOption == CacheOption.WRITE_THROUGH) {
                 readCache.put(token.getSequence(), ld);
+                currentDataSizeInCache.addAndGet(ld.getSizeEstimate());
```
Maithem, line 255: This accounts for the the serialized size and not the in-memory size of the object.

```java

 @Override
    public int getSizeEstimate() {
        byte[] tempData = data;
        if (tempData != null) {
            return tempData.length;
        }

        if (lastKnownSize != NOT_KNOWN) {
            return lastKnownSize;
        }

        return 1;
    }
```

Maithem: https://github.com/CorfuDB/CorfuDB/blob/master/runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java#L197 is also calculating the size estimate, but before "serialization", so its most likely inaccurate. Remove it?

Maithem (commented):
The way these metrics are being calculated is not accurate, for the following reasons:
1. estimateSize is an underestimate and can be incorrect (depending on when its called)
2. Only the write path is being accounted for, but the read cache will also hold entries that are read (i.e., entries read from other clients and synced into this JVM ).

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -115,14 +118,24 @@ public AddressSpaceView(@Nonnull final CorfuRuntime runtime) {
...
+        MicroMeterUtils.gauge("address_space.read_cache.hit_ratio", readCache, cache -> cache.stats().hitRate());
+        MicroMeterUtils.gauge("address_space.read_cache.size", readCache, cache -> cache.size());
+        MicroMeterUtils.gauge("address_space.read_cache.avg_entry_size", calculateEstimatedAvgEntrySize());
     }
 
     private void handleEviction(RemovalNotification<Long, ILogData> notification) {
         if (log.isTraceEnabled()) {
             log.trace("handleEviction: evicting {} cause {}", notification.getKey(), notification.getCause());
         }
+        currentDataSizeInCache.addAndGet(notification.getValue().getSizeEstimate() * -1);
```
Maithem, line 131: This can produce a negative average. Image the case where a client just reads data and never writes, or reads at a rate higher than what it writes. It would calculate the wrong thing.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -115,14 +118,24 @@ public AddressSpaceView(@Nonnull final CorfuRuntime runtime) {
                 .build();
 
         Optional<MeterRegistry> metricsRegistry = MeterRegistryProvider.getInstance();
-        MicroMeterUtils.gauge("address_space.read_cache.hit_ratio", readCache, cache -> cache.stats().hitRate());
         metricsRegistry.map(registry -> GuavaCacheMetrics.monitor(registry, readCache, "address_space.read_cache"));
+        MicroMeterUtils.gauge("address_space.read_cache.hit_ratio", readCache, cache -> cache.stats().hitRate());
+        MicroMeterUtils.gauge("address_space.read_cache.size", readCache, cache -> cache.size());
+        MicroMeterUtils.gauge("address_space.read_cache.avg_entry_size", calculateEstimatedAvgEntrySize());
```
Maithem, line 124: Ignoring how accurate the estimate function is, another way to improve this patch's accuracy is to keep two counters, one for the total number of entries and the other one for total of all estimates. To compute the average you just need to divide both counters and reset them to 0. That way you dont have to account for evictions.

shama358:
raised a new PR: https://github.com/CorfuDB/CorfuDB/pull/3046

PR #3044, opened 2021-09-09
TxnContext: Don't log a warning on empty transactions

hisundar:
Some logic might result in commit() being called on
a transaction with no reads or writes.
Treat this as a normal case and avoid logging a warning
to reduce noise in the logs.

## Overview

Description:

Why should this be merged: 

- [x] Public API has Javadoc


zhangn49 (approved):
LGTM, but why the logic has an empty TXN? Should they remove the TX?

Jas0nch (approved):
looks good to me

PR #3045, opened 2021-09-13
Introduce KeyDynamicProtobufSerializer and OpaqueCorfuDynamicRecord

Lujie1996:
Deserializing CorfuDynamicRecord can be expensive. Protobuf DynamicMessage
comes with a high memory overhead. Also there are some functionalities that
doesn't require deserializing CorfuDynamicRecord, such as Compactor. The
new serializer and opaque record will allow to keep the payload and metadata
in ByteString format and only deserialize the Key message.

## Overview

Description:

Why should this be merged: 
Reduce compactor's memory usage due to DynamicMessage's overhead

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
+++ b/runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
@@ -1,0 +1,135 @@
+package org.corfudb.util.serializer;
+
+import com.google.protobuf.Any;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.DynamicMessage;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufInputStream;
+import io.netty.buffer.ByteBufOutputStream;
+import lombok.extern.slf4j.Slf4j;
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.OpaqueCorfuDynamicRecord;
+import org.corfudb.runtime.exceptions.SerializerException;
+
+import java.io.IOException;
+
+@Slf4j
+public class KeyDynamicProtobufSerializer extends DynamicProtobufSerializer {
+
+    public KeyDynamicProtobufSerializer(CorfuRuntime corfuRuntime) {
+        super(corfuRuntime);
+    }
+
+    /**
+     * Deserialize an object from a given byte buffer.
+     *
+     * @param b The bytebuf to deserialize.
+     * @return The deserialized object.
+     */
+    @Override
+    public Object deserialize(ByteBuf b, CorfuRuntime rt) {
+
+        try (ByteBufInputStream bbis = new ByteBufInputStream(b)) {
+            ProtobufSerializer.MessageType type = ProtobufSerializer.MessageType.valueOf(bbis.readInt());
+            int size = bbis.readInt();
+            byte[] data = new byte[size];
+            bbis.readFully(data);
+            CorfuStoreMetadata.Record corfuRecord = CorfuStoreMetadata.Record.parseFrom(data);
+            Any payload = corfuRecord.getPayload();
+
+            String fullMessageName = getFullMessageName(payload);
+            if (!messagesFdProtoNameMap.containsKey(fullMessageName)) {
+                log.error("messagesFdProtoNameMap doesn't contain the message type {} of payload {}." +
+                                "Please check if the related table is properly opened with correct schema.",
+                        fullMessageName, payload);
+                log.error("messagesFdProtoNameMap keySet is {}", messagesFdProtoNameMap.keySet());
+            }
+
+            if (type.equals(ProtobufSerializer.MessageType.KEY)) {
+                Descriptors.FileDescriptor valueFileDescriptor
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
```diff
@@ -0,0 +1,135 @@
...
+
+            String metadataTypeUrl = null;
+            ByteString metaDataByteString = null;
+            if (corfuRecord.hasMetadata()) {
+                Any anyMetadata = corfuRecord.getMetadata();
+                metadataTypeUrl = anyMetadata.getTypeUrl();
+                metaDataByteString = anyMetadata.getValue();
+
+                String fullMetadataMessageName = getFullMessageName(anyMetadata);
+                if (!messagesFdProtoNameMap.containsKey(fullMetadataMessageName)) {
```
hisundar, line 67: This would still hit a Deserialization exception if there is a problem with the type of the record right?
Instead can we just directly return the byte[] as the type itself symmetrically across serialize() and deserialize() so that if something goes wrong, the DyanmicProtobufSerializer does not come in the middle?

hisundar (changes requested):
Hi @Lujie1996 - while this is a good change, it does not avoid some of the problems we are trying to avoid right - specifically the serialization exception during compaction.
Instead of adding new type of OpaqueCorfuDynamicRecord can we just return the byte[] in deserialize and serialize methods if we are not dealing with key type?
That way we can avoid a lot of problems during compaction right?

hisundar:
On second thought I think just logging an error is a good idea too, so let's just go with this approach.
One thing I would request is to please fix up only the Compaction ITs so we have code coverage for this change?
thanks @Lujie1996

runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
```diff
@@ -0,0 +1,135 @@
...
+                    log.error("messagesFdProtoNameMap doesn't contain the message type {} of metadata {}." +
+                                    "Please check if the related table is properly opened with correct schema.",
+                            fullMetadataMessageName, anyMetadata);
+                    log.error("messagesFdProtoNameMap keySet is {}", messagesFdProtoNameMap.keySet());
+                }
+            }
+            return new OpaqueCorfuDynamicRecord(payload.getTypeUrl(), payload.getValue(), metadataTypeUrl, metaDataByteString);
```
pankti-m, line 74: did you miss deserializing the metadata from the descriptor obtained from the map?  Otherwise what is the use of the 'if' check above?

runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
```diff
@@ -0,0 +1,135 @@
...
+import org.corfudb.runtime.CorfuRuntime;
+import org.corfudb.runtime.CorfuStoreMetadata;
+import org.corfudb.runtime.collections.CorfuDynamicKey;
+import org.corfudb.runtime.collections.OpaqueCorfuDynamicRecord;
+import org.corfudb.runtime.exceptions.SerializerException;
+
+import java.io.IOException;
+
+@Slf4j
+public class KeyDynamicProtobufSerializer extends DynamicProtobufSerializer {
```
pankti-m, line 20: It will be good to add a comment/javadoc explaining that this class is used to serialize/deserialize the key from a byte buffer and construct an OpaqueRecord.

runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
Lujie1996: This only logs the error but now fail the compactor. Logging such error can assist the debugging to find out when such issue happened for the first time.

runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
Lujie1996: This serializer will only serialize / deserialize the Key. The if-check above serves the purpose of validating the protobuf types. If someone reopened a table with a wrong schema, it overwrites the protobuf type. In that case the periodic compactor will capture this issue and it will log the error here. Let me add a comment to the if-check to improve readability.

runtime/src/main/java/org/corfudb/util/serializer/KeyDynamicProtobufSerializer.java
Lujie1996: Sure. Will do. Thanks!

Lujie1996:
> Hi @Lujie1996 - while this is a good change, it does not avoid some of the problems we are trying to avoid right - specifically the serialization exception during compaction.
> Instead of adding new type of OpaqueCorfuDynamicRecord can we just return the byte[] in deserialize and serialize methods if we are not dealing with key type?
> That way we can avoid a lot of problems during compaction right?

Lujie1996:
> On second thought I think just logging an error is a good idea too, so let's just go with this approach.
> One thing I would request is to please fix up only the Compaction ITs so we have code coverage for this change?
> thanks @Lujie1996

Hi @hisundar The existing checkpoint tests all use primitive types when opening tables. So instead of updating existing ones I have added a new test. Please feel free to review. Thanks.

hisundar (approved):
Thanks @Lujie1996 this looks good to me!

pankti-m (approved):

PR #3046, opened 2021-09-14
Adding 2 readCache metrics: avg-entry-size and number of entries in readCache

shama358:
## Overview

Description:
Adding readCache metrics: avg-entry-size and number of entries in readCache. Since the ask of these 2 metrics is for tuning the readCache, they can be enabled via corfuRuntimeParameters. 
By default, the 2 metrics are disabled.
 
This change uses [ehcache's sizeOf()](https://github.com/ehcache/sizeof) to get the deep size of an object. 
I did some performance test, on proton and PHC, to check the time overhead of using sizeOf, and:
1. p99 for a single sizeOf operation is around 1 ms. The tail latencies that I saw ranged between 1 ms - 13 ms 
2. when we read the entire readCache to calculate the avg-entry-size always, it takes a couple of minutes (3 mins is the longest seen). This is outside of read/write flow and this is the approach taken in this PR.

Why should this be merged:  This will help to tune the address space cache without having to analyze the heap dump

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -23,6 +23,7 @@
         <commons.lang3.version>3.9</commons.lang3.version>
         <commons.compress.version>1.21</commons.compress.version>
         <annotations.api.version>6.0.53</annotations.api.version>
+        <ehcahce.sizeOf.version>0.4.0</ehcahce.sizeOf.version>
 
         <grpc.version>1.29.0</grpc.version>
     </properties>
@@ -167,6 +168,11 @@
             <groupId>io.grpc</groupId>
             <artifactId>grpc-stub</artifactId>
             <version>${grpc.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.ehcache</groupId>
+            <artifactId>sizeof</artifactId>
+            <version>${ehcahce.sizeOf.version}</version>
         </dependency>
         <dependency> <!-- necessary for Java 9+ -->
             <groupId>org.apache.tomcat</groupId>
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -219,6 +219,11 @@ public class CorfuRuntime {
         Duration holeFillTimeout = Duration.ofSeconds(10);
 
         /*
+        * cache metrics are to be enabled only for the tuning exercise.
+        */
+        boolean cacheEntryMetricsDisabled = true;
+
+        /*
          * Whether or not to disable the cache.
          */
         boolean cacheDisabled = false;
@@ -395,6 +400,7 @@ public class CorfuRuntime {
             private int holeFillRetry = 10;
             private Duration holeFillRetryThreshold = Duration.ofSeconds(1L);
             private Duration holeFillTimeout = Duration.ofSeconds(10);
+            private boolean cacheEntryMetricsDisabled = true;
             private boolean cacheDisabled = false;
             private long maxCacheEntries;
             private long maxCacheWeight;
@@ -568,6 +574,11 @@ public class CorfuRuntime {
                 return this;
             }
 
+            public CorfuRuntimeParameters.CorfuRuntimeParametersBuilder cacheEntryMetricsDisabled(boolean cacheEntryMetricsDisabled) {
+                this.cacheEntryMetricsDisabled = cacheEntryMetricsDisabled;
+                return this;
+            }
+
             public CorfuRuntimeParameters.CorfuRuntimeParametersBuilder cacheDisabled(boolean cacheDisabled) {
                 this.cacheDisabled = cacheDisabled;
                 return this;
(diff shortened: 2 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -194,7 +207,7 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
             } else {
                 ld = new LogData(DataType.DATA, data, runtime.getParameters().getCodecType());
             }
-            recordLogSizeDist(ld.getSizeEstimate());
+            recordLogSizeDist(sizeOf.deepSizeOf(ld));
```
PavelZaytsev, line 210: Just curious, have you measured how different this approach is to calling getSizeEstimate? Would be interesting to see the reported difference for the same kind of workload.

shama358: Nope..I did not measure the time taken for getSizeEstimate. My guess would be that getSizeEstimate would be lower than deepSizeOf() as the latter one requires to walk through the object tree

runtime/pom.xml
```diff
@@ -168,6 +169,11 @@
             <artifactId>grpc-stub</artifactId>
             <version>${grpc.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.ehcache</groupId>
+            <artifactId>sizeof</artifactId>
+            <version>${ehcahce.sizeOf.version}</version>
+        </dependency>
```
Maithem, line 176: In the past we did try using ehcache, but the problem was it had a large memory footprint for some objects. So in addition to the running time cost, we should also investigate the memory overhead, this is tricky because its specific to the object structure.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -125,6 +127,17 @@ private void handleEviction(RemovalNotification<Long, ILogData> notification) {
...
+    private double calculateEstimatedAvgEntrySize() {
+        if (readCache.size() == 0) {
+            return 0.0;
+        }
+        long currentCacheSize = readCache.size();
+        long currentDataSizeInCache = readCache.asMap().entrySet().stream()
+                .mapToLong(e -> sizeOf.deepSizeOf(e.getKey()) + sizeOf.deepSizeOf(e.getValue())).sum();
+
+        return (double) currentDataSizeInCache / currentCacheSize;
+    }
```
Maithem, line 139: This will be computed every minute, but the cache eviction happens at a much higher rate therefore this will be inaccurate.

Maithem (commented):
It's not clear how helpful this metric will be, especially when its not accurate. 

The problem is that the interaction of the cache with other layers is a little complicated. For example, an entry can be admitted to the cache in a compressed form, then later its lazily deserialized and decompressed, this will be a large portion of the memory used, but not accounted for by this metric. 

Also, with the log constantly moving, the cache will be constantly evicting and just looking at objects every minute means that there will be a lot of LogData sizes that are not sampled. 

I don't know what the best solution.

runtime/pom.xml
shama358: yes, this is true.
While the sizeOf instance itself just takes bytes, there are other classes which occupy in MBs. And like you said, it depends on which objects are currently in memory

shama358:
Latest Update :
Have reverted the changes around "address_space.log_data.size.bytes", since sizeOf() impacts performance.

PavelZaytsev (approved):

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
Lujie1996: Another concern is that this will block the evicted objects from being garbage collected since sizeOf.deepSizeOf() still holds the references to them. This issue is serious when readCache itself is large and the eviction rate is high. Yet considering that we don't enable this metrics in production I think it should be fine.

Lujie1996 (approved):

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -395,6 +400,7 @@ public static CorfuRuntimeParametersBuilder builder() {
             private int holeFillRetry = 10;
             private Duration holeFillRetryThreshold = Duration.ofSeconds(1L);
             private Duration holeFillTimeout = Duration.ofSeconds(10);
+            private boolean cacheEntryMetricsDisabled = false;
```
Lujie1996, line 403: Set to true? otherwise if CorfuRuntimeParametersBuilder is used without explicitly setting this parameter it will enable this metrics.

shama358: ahh yes, my bad. I wanted it to test out the metrics so had changed it to "false". Did not realize I pushed after setting it to false.
Thanks for catching this!

PR #3047, opened 2021-09-14
StreamingManager Rewrite

Maithem:
## Overview
StreamingManager Rewrite

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/pom.xml
```diff
@@ -174,6 +174,11 @@
             <version>${annotations.api.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>commons-collections</groupId>
+            <artifactId>commons-collections</artifactId>
```
Maithem, line 179: migrate to commons-collections4

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,309 @@
...
+        List<List<StreamingTask>> batches = Lists.partition(tasks, pollBatchSize);
+
+        for (List<StreamingTask> batch : batches) {
+            List<StreamAddressRange> pollRequest = new ArrayList<>();
+            for (StreamingTask task : batch) {
+                DeltaStream stream = task.getStream();
+                pollRequest.add(new StreamAddressRange(stream.getStreamId(), Address.MAX, stream.getMaxAddressSeen()));
+            }
+
+            List<StreamAddressRange> coalesced = coalesce(pollRequest);
```
Maithem, line 187: Coalesce should be performed before batching otherwise there could be stream queries across batches that aren't coalesced.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,309 @@
...
+
+    // hide this?
+    public void schedule() throws InterruptedException {
+        // Block thread, so that it doesn;t spin if there are no tasks
+        waitForTasks();
+
+        long startTime = System.nanoTime();
+
+        List<StreamingTask> runnableTasks = getTasks(StreamStatus.RUNNABLE);
+        List<StreamingTask> syncingTasks = getTasks(StreamStatus.SYNCING, t -> t.getStream().availableSpace() >= pollThreshold);
```
Maithem, line 278: Since a task can change from SYNCING to RUNNABLE while the scheduling is running, its imperative that the RUNNABLE tasks are queried first, otherwise there could be common tasks in runnableTasks and syncingTasks, this can cause issues if the same task is dispatched to sync on two different worker threads.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingTask.java
```diff
@@ -0,0 +1,192 @@
...
+                            return new TableSchema<>(tName, t.getKeyClass(), t.getValueClass(), t.getMetadataClass());
+                        }));
+        this.status = new AtomicReference<>(StreamStatus.RUNNABLE);
+    }
+
+    public StreamStatus getStatus() {
+        return status.get();
+    }
+
+    public void move(StreamStatus from, StreamStatus to) {
```
Maithem, line 89: Write unit tests for the negative cases for this method.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,153 @@
...
+
+        if (readAddress <= trimMark) {
+            throw new TrimmedException(String.format("current %d trimMark %d", readAddress, trimMark));
+        }
+
+        // set cache options here
+        // test trimmed exception
+        // test hole
+        // test empty?
+        ILogData logData = addressSpaceView.read(readAddress);
```
Maithem, line 135: This read will use the default cache setting and pollute the cache for the whole runtime. Need to pass read options.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingManager.java
```diff
@@ -0,0 +1,132 @@
...
+ * <p>
+ */
+@Slf4j
+public class StreamingManager {
+
+    private final CorfuRuntime runtime;
+
+    private final StreamPollingScheduler scheduler;
+
+    private final int defaultBatchSize = 25;
```
Maithem, line 37: *Default Buffer Size

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: AddressSpaceView::read will peek-sleep-read-hole fill on a shared worker thread which will block other threads that can retrieve their deltas on the first peek. A new API should be introduced to break these protocol read operations into non blocking operations so that they can be scheduled explicitly.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+        return result;
+    }
+
+    private List<StreamingTask> getTasks(StreamStatus status) {
+        return getTasks(status, t -> true);
+    }
+
+    private void poll(List<StreamingTask> tasks) {
+        log.info("polling tasks {}", tasks);
```
achinmishra, line 176: Its generating whole flurry of messages. should change to debug

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,309 @@
...
+            for (StreamingTask task : batch) {
+                DeltaStream stream = task.getStream();
+                pollRequest.add(new StreamAddressRange(stream.getStreamId(), Address.MAX, stream.getMaxAddressSeen()));
+            }
+
+            List<StreamAddressRange> coalesced = coalesce(pollRequest);
+
+            long pollStartTime = System.nanoTime();
+            Map<UUID, StreamAddressSpace> res = sequencerView.getStreamsAddressSpace(coalesced);
+            log.info("poll: took {} to poll {} streams",
```
achinmishra, line 191: another log -> debug.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,312 @@
...
+    public void removeTask(@Nonnull StreamListener streamListener) {
+        synchronized (allTasks) {
+            allTasks.remove(streamListener);
+            allTasks.notifyAll();
+        }
+    }
+
+    // hole filling mitigation
+    // sequencer regression mitigation
+    // fairness?
```
Maithem, line 150: These are existing problems. Convert to TODOs.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private final AddressSpaceView addressSpaceView;
+
+    @Getter
+    private final UUID streamId;
+
+    /**
+     * Last Address consumed/read by this stream
+     */
+    private volatile long lastAddressRead;
```
xnull, line 35: let's replace to AtomicLong?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+            .builder()
+            .clientCacheable(false)
+            .ignoreTrim(false)
+            .waitForHole(true)
+            .serverCacheable(false)
+            .build();
+
+    public DeltaStream(AddressSpaceView addressSpaceView, UUID streamId, long lastAddressRead, int bufferSize) {
+        this.addressSpaceView = addressSpaceView;
+        Preconditions.checkArgument(Address.NON_ADDRESS == -1);
```
xnull, line 68: looks weird, what's the motivation for comparing two constants?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+     */
+    private boolean trimGap() {
+        return trimMark != Address.NON_ADDRESS && lastAddressRead < trimMark;
+    }
+
+    //max has to always be greater than last address read
+    // called by the scheduler thread
+    public int availableSpace() {
+        int availableSpace = this.bufferSize - addressesToRead.size();
+        Preconditions.checkState(availableSpace >= 0);
```
xnull, line 93: That's always unexpected when you execute a methods like `size()`, `availableSpace()`, `isEmpty()` and the method throws an Exception. I'd say that's a bad practice

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+        return availableSpace;
+    }
+
+    public void refresh(StreamAddressSpace streamAddressSpace) {
+        if (log.isTraceEnabled()) {
```
xnull, line 98: no need for `if`

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        Preconditions.checkState(trimMark <= streamAddressSpace.getTrimMark(),
+                "streamId=%s new trimMark %s has to be >= %s", streamId, streamAddressSpace.getTrimMark(), trimMark);
+
+        // Adopt new trim mark
+        trimMark = streamAddressSpace.getTrimMark();
+
+        int maxAddressesToTransfer = availableSpace();
+        long[] newAddresses = streamAddressSpace.toArray();
```
xnull, line 126: why not just use `streamAddressSpace` list? why you need to create an array (and consume 2x mem) from the list?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+    private void tick() {
+        scheduler.submit(new Tick());
+    }
+
+    public void addTask(@Nonnull StreamListener streamListener, @Nonnull String namespace,
+                        @Nonnull String streamTag, @Nonnull List<String> tablesOfInterest,
+                        long lastAddress, int bufferSize) {
+
+        Preconditions.checkArgument(bufferSize >= pollThreshold);
+        synchronized (allTasks) {
```
xnull, line 126: if you block the entire map by using synchronized, then no need for ConcurrentMap

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    // hole filling mitigation
+    // sequencer regression mitigation
+    // fairness?
+
+    private void waitForTasks() throws InterruptedException {
+        synchronized (allTasks) {
+            while (allTasks.isEmpty()) {
+                allTasks.wait();
```
xnull, line 155: do we really need such low-level concurrent primitives? I'm sure CompletableFutures shall be used

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+            Preconditions.checkState(allQueryResults.containsKey(taskQuery.getStreamID()),
+                    "StreamAddressSpace missing for %s", task.getStream().getStreamId());
+
+            StreamAddressSpace sas = allQueryResults.get(task.getStream().getStreamId()).getAddressesInRange(taskQuery);
+            task.getStream().refresh(sas);
+        }
+    }
+
+    @Data
+    class Interval {
```
xnull, line 220: static?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingManager.java
```diff
@@ -0,0 +1,132 @@
...
+    private final CorfuRuntime runtime;
+
+    private final StreamPollingScheduler scheduler;
+
+    private final int defaultBufferSize = 25;
+
+    public StreamingManager(@Nonnull CorfuRuntime runtime) {
+        this.runtime = runtime;
+
+        ScheduledExecutorService schedulerThread = Executors.newScheduledThreadPool(1, new ThreadFactoryBuilder()
```
xnull, line 42: newSingleThreadScheduledExecutor

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+ * On every tick the scheduler will poll/dispatch tasks only when needed, for example, when a listener is not
+ * consuming the deltas fast enough the scheduler can decide to not poll or schedule extra sync tasks until
+ * it can, this prevents superfluous polling and overwhelming the listener. Furthermore, syncing and listener
+ * notification is executed on a shared thread pool in round robin fashion, therefore if a stream produces
+ * more deltas relative to other streams that particular stream will have more CPU time than other tasks.
+ */
+
+
+@Slf4j
+public class StreamPollingScheduler {
```
xnull, line 69: BTW, did you consider using quartz-scheduler instead of self written code?

xnull: Or [RxJava](https://www.baeldung.com/rxjava-schedulers) even better alternative

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
zfrenette: I thought that this synchronized was used to guard against the state of the map changing between the `containsKey` and the subsequent `put`?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: This variable is used to coordinate between two threads, a thread that writes to it and another thread that reads from it. If I only need read-after write semantics and no CAS-related operations, why is it better to use an AtomicLong here?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: It does look weird, but if you look at the subsequent line, there's logic that depends on the specific value (i.e., -1) and I don't want to use -1 directly. This just makes sure that if NON_ADDRESS changes, then these tests would fail.

```java

Preconditions.checkArgument(Address.NON_ADDRESS == -1);
Preconditions.checkArgument(lastAddressRead >= Address.NON_ADDRESS,
                "lastAddressRead %s must be >= %s", lastAddressRead, Address.NON_ADDRESS);

```

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: The exception is not meant to be part of the API, this is more of an **invariant**. If this exception is thrown, then this signals that the internal state of a DeltaStream has been corrupted. Explicitly failing makes it easier to detect concurrency issues easier.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: The synchronized block is not used to maintain the integrity of the `allTasks` map, its to implement a conditional variable. In Java, synchronized/notify APIs are used together to implemented a **conditional variable**.  Without a synchronized block, notify/wait semantics can lose updates, causing the thread scheduler thread to block indefinitely.

Maithem: You can read more about it here https://docs.oracle.com/javase/tutorial/essential/concurrency/guardmeth.html

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: What is the disadvantage of using this API ?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: I haven't considered using quartz, but I think it would be an overkill. 

Which part of this patch do you think would be simplified/cleaned/optimized if we opt to use quartz?

Maithem: Since the `AddressSpaceView::read` and `SequencerView:getStreamAddressSpace` are blocking APIs I don't see any performance benefit of using RxJava. Since this will be loaded on the client side, I don't want to add dependencies unless absolutely necessary.

If you have a specific idea on how to improve this schedulers implementation using RxJava please feel free to articulate it here and I will investigate it.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: Why ?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+    public DeltaStream(AddressSpaceView addressSpaceView, UUID streamId, long lastAddressRead, int bufferSize) {
+        this.addressSpaceView = addressSpaceView;
+        Preconditions.checkArgument(Address.NON_ADDRESS == -1);
+        Preconditions.checkArgument(lastAddressRead >= Address.NON_ADDRESS,
+                "lastAddressRead %s must be >= %s", lastAddressRead, Address.NON_ADDRESS);
+        Preconditions.checkArgument(bufferSize > 0, "buffer size %s must be > 0", bufferSize);
+        this.streamId = streamId;
+        this.bufferSize = bufferSize;
+        this.addressesToRead = BufferUtils.synchronizedBuffer(new BoundedFifoBuffer(this.bufferSize));
+        this.lastAddressRead = lastAddressRead;
```
Maithem, line 75: Add Comment: It's worth mentioning that this is a circular buffer and doesnt required garbage collection, unlike IStreamView it doesnt depend on `runRuntimeGC` for clean up.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+                "streamId=%s new trimMark %s has to be >= %s", streamId, streamAddressSpace.getTrimMark(), trimMark);
+
+        // Adopt new trim mark
+        trimMark = streamAddressSpace.getTrimMark();
+
+        int maxAddressesToTransfer = availableSpace();
+        long[] newAddresses = streamAddressSpace.toArray();
+        for (int idx = 0; idx < newAddresses.length && idx < maxAddressesToTransfer; idx++) {
+            // if maxAddressesToTransfer doesn't fit, then the rest of the addresses will be
+            // thrown away and re-discovered on the next pool
```
Maithem, line 129: pool -> poll

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
xnull: AtomicLong has same functionality, if you go inside that class you'll see there is a volatile variable inside the class.
Having AtomicLong instead of volatile indicates that the variable is volatile in compile time, and reduces the risk of using that variable incorrectly in the future. Next time another developer couldn't notice that the variable is volatile and would use it incorrectly, thinking that it is a regular variable, which is not possible in the case of atomic variables

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamStatus.java
```diff
@@ -0,0 +1,8 @@
+package org.corfudb.runtime.collections.streaming;
+
+public enum StreamStatus {
+    ERROR,
+    SCHEDULING,
+    RUNNABLE,
```
annym, line 6: Is this meant to be running? -> status?

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
```diff
@@ -60,7 +60,7 @@ public void TXBegin(TransactionType type) {
         }
     }
 
-    protected final int numIterations = PARAMETERS.NUM_ITERATIONS_MODERATE;
+    protected final int numIterations = 100;
```
annym, line 63: just out of curiosity, does the test not run fast for 1000 or did you lower it for debugging?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
xnull: Then you should move that check into the class constructor, it prevents you from creating a corrupted instance of DeltaStream? Why you need to check each time

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
xnull: Now, that's a weird motivation for that also. 
If you want NON_ADDRESS to be always -1, then write a unit test that checks NON_ADDRESS == -1.

annym (commented):
Partial comments (just so you can comment in the meanwhile or address if so)... still reviewing..

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: `addressesToRead.size()` changes as the worker thread consumes items from the buffer therefore checking it once in the constructor is not enough.

Maithem: > Explicitly failing makes it easier to detect concurrency issues easier.

This If corruption happens because of multi-threading, at construction time there is only one thread, so moving it in the constructor doesn't capture what im checking for here.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -350,40 +350,26 @@
...
         /*
-         * Capacity of queue shared by by streaming polling and notification tasks.
+         * Number of streams that the streaming scheduler will query on every poll
          */
-        private int streamingQueueSize = 100;
+        private int streamingSchedulerPollBatchSize = 25;
 
         /*
-         * Total number of threads in Polling Executor Pool (shared across all listeners)
+         * Available space that a stream should have before it can be polled. This parameter is used to
```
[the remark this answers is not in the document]
Maithem: The worker has a buffer to consume reads from, this buffer should have at least 5 free slots before its polled. This allows for throttling. If the worker is not consuming the buffer at a fast rate, there is no point of polling it.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
```diff
@@ -7,6 +7,7 @@
  */
 public interface StreamListener {
 
+    // TODO(Maithem): what is this?? remove?
```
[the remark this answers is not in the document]
Maithem: Seems like this API already leaked to the Application, can't remove it without breaking their code.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
xnull: oh, didn't that you executed 'getTrimMark()', then ok

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
xnull: It's low level, it requires a good understanding of JMM, "happens before" and other concurrency aspects. It's relatively easy to make a mistake in the code and super hard to find a possible dead-lock in the code especially if it happens in production.  I don't see why we need that level of complexity here.  
Especially when we can use a [specific library](https://github.com/ReactiveX/RxJava) for that with a clean interface, very well tested and used by major companies in the world.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
xnull: Can we use [guava's implementation](https://stackoverflow.com/questions/14232548/is-there-a-guava-equivalent-to-apache-commons-circularfifobuffer) then?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private final ReadOptions options = ReadOptions
+            .builder()
+            .clientCacheable(false)
+            .ignoreTrim(false)
+            .waitForHole(true)
+            .serverCacheable(false)
+            .build();
+
+    public DeltaStream(AddressSpaceView addressSpaceView, UUID streamId, long lastAddressRead, int bufferSize) {
```
annym, line 66: One suggestion, I think besides the streamId, adding the streamTag would ease debugging (have it for logging purposes). We typically want to track updates to a given tag, and its easier too grep based on the name which is what the user knows about.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+
+    //max has to always be greater than last address read
+    // called by the scheduler thread
+    public int availableSpace() {
+        int availableSpace = this.bufferSize - addressesToRead.size();
+        Preconditions.checkState(availableSpace >= 0);
+        return availableSpace;
+    }
+
+    public void refresh(StreamAddressSpace streamAddressSpace) {
```
annym, line 97: Can we add a description to this method?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+        if (log.isTraceEnabled()) {
+            log.trace("refresh id {} max {} trim {} newTm {} newAddresses {}", streamId, maxAddressSeen, trimMark,
```
annym, line 99: perhaps adding the count of addresses is useful, as gives immediate info of how many updates are expected (in case its too large to easily count).

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+                    streamAddressSpace.getTrimMark(), streamAddressSpace.toArray());
+        }
+
+        if (lastAddressRead < streamAddressSpace.getTrimMark()
+                || maxAddressSeen < streamAddressSpace.getTrimMark()) {
+
+            // Since a trim has been detected, there is no needed to transfer the addresses, we just
```
annym, line 106: typo -> needed -> need

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: In this implementation, before the producer adds any items it computes how much available buffer space is avaialble and then only produce that much. If there is a bug in that logic then adding to the buffer would fail with an explicit exception (i.e., buffer overflow). You can see that I test for this explicit case in the unit tests.

Guava's implementation will overwrite unconsumed items which will cause data loss issues for our application.

https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/EvictingQueue.java#L39

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -1036,50 +998,6 @@ public void testFullSyncDeltaSyncPattern() throws Exception {
         assertThat(shutdownCorfuServer(corfuServer)).isTrue();
     }
 
-    /**
```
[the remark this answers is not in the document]
Maithem: These tests verify specific behavior to how `IStreamView` interacts with `ViewsGarbageCollector` and how it produces trimmed exceptions. `DeltaStream` doesn't interact with the `ViewsGarbageCollector` , so these tests are not applicable. 

All paths that produce TrimmedExceptions are covered by unit tests for the `DeltaStream`. Let me know if there is a case that isn't covered.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamStatus.java
Maithem: Nope. Runnable as in read to be executed, that is, scheduled and synced.

test/src/test/java/org/corfudb/runtime/concurrent/CorfuQueueTxTest.java
Maithem: The test fidelity doesn't increase from 100 to 1000, that's why I changed it. CI is slow and this test is one of the tests that have been failing intermittently for a long time.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: I'm trying to avoid the cost of  `toArray()` if trace is not enabled.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: The API is low level, but the usage is straightforward and succinct. Using RxJava just to implement a [conditional variable](https://en.wikipedia.org/wiki/Monitor_(synchronization)#:~:text=A%20condition%20variable%20essentially%20is,access%20and%20resuming%20their%20task.)  seems unnecessary to me. 

I have a single scheduler thread that needs to wake up if there is at least one task that has been registered. How would that implementation look like in RxJava ?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+            throw new TrimmedException(String.format("lastAddressRead %d trimMark %d", lastAddressRead, trimMark));
+        }
+
+        long readAddress = (long) addressesToRead.remove();
+
+        Preconditions.checkState(lastAddressRead < readAddress,
+                "%d must be greater than %d", readAddress, lastAddressRead);
+
+        if (readAddress <= trimMark) {
+            throw new TrimmedException(String.format("current %d trimMark %d", readAddress, trimMark));
```
annym, line 164: Do we have metrics around how much time an address sits on the buffer before being serviced?

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -350,40 +350,26 @@
...
         /*
-         * Period of time in ms to sleep before next cycle, when poller gets no new data changes.
+         * Streaming scheduler poll period
          */
-        private int streamingPollingIdleWaitTimeMs = 50;
+        private Duration streamingPollPeriod = Duration.ofMillis(50);
 
         /*
-         * Capacity of queue shared by by streaming polling and notification tasks.
+         * Number of streams that the streaming scheduler will query on every poll
```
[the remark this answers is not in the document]
Maithem: Probably need to update the comment, but this should be the number of unique corfu streams to poll every time a sequencer query is made.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        long startRead = System.nanoTime();
+        ILogData logData = addressSpaceView.read(readAddress, options);
```
annym, line 168: if we have bufferSize addresses, can't we do a range read and deliver more than one update in a single shot?

annym (commented):
will continue...

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+
+    public void schedule() throws InterruptedException {
+        // Block thread until there's work to do
+        waitForTasks();
+
+        final long cycleStartTimeNs = System.nanoTime();
+
+        final List<StreamingTask> runnableTasks = getTasks(StreamStatus.RUNNABLE);
+        final List<StreamingTask> syncingTasks = getTasks(StreamStatus.SYNCING,
+                t -> t.getStream().availableSpace() >= pollThreshold);
```
annym, line 295: if pollThreshold is > than bufferSize this will never be true, right?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+    private List<StreamAddressRange> getPollQueries(List<StreamingTask> tasks) {
+        List<StreamAddressRange> pollRequests = new ArrayList<>(tasks.size());
+        for (StreamingTask task : tasks) {
+            DeltaStream stream = task.getStream();
+            pollRequests.add(new StreamAddressRange(stream.getStreamId(), Address.MAX, stream.getMaxAddressSeen()));
+        }
+        return pollRequests;
+    }
+
+    private void poll(List<StreamingTask> tasks) {
```
annym, line 183: can we add comments? took me a while to internalize that 'poll' meant sequencer poll and not data polling (actual read of data, which I believe is what you have called sync)

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+
+        final long cycleStartTimeNs = System.nanoTime();
+
+        final List<StreamingTask> runnableTasks = getTasks(StreamStatus.RUNNABLE);
+        final List<StreamingTask> syncingTasks = getTasks(StreamStatus.SYNCING,
+                t -> t.getStream().availableSpace() >= pollThreshold);
+        final List<StreamingTask> failedTasks = getTasks(StreamStatus.ERROR);
+
+        runnableTasks.forEach(t -> t.move(StreamStatus.RUNNABLE, StreamStatus.SCHEDULING));
+        poll(runnableTasks);
```
annym, line 299: Perhaps it's me, but I had a hard time understanding these 4 lines. Since you understand the code more if you think of some way that this is more intuitive I think it would help. Also, one question what is the value of the status 'scheduling'? is it equivalent to 'waiting to sync'?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+        for (Map.Entry<UUID, Interval> entry : adjusted.entrySet()) {
+            adjustedList.add(new StreamAddressRange(entry.getKey(), entry.getValue().y, entry.getValue().x));
+        }
+        return adjustedList;
+    }
+
+    private void dispatchSyncTasks(List<StreamingTask> tasks) {
+        for (StreamingTask task : tasks) {
+            if (task.getStream().hasNext()) {
+                // this will fail for tasks that are already syncing
```
annym, line 264: doesn't that mean that it will break out for all other tasks which could still be submitted to the workers executor?

hisundar (commented):
Still trying to complete the review, but would like to submit my initial set of questions before context switching.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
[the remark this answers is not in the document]
annym: actually this is not intended to be consumed externally, the onNext is the clients callback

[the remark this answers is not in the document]
Maithem: Yes, I was under the assumption that it was used. Turns out it is note. Just removed it.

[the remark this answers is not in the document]
annym: mmm but no guys we can't just remove it, that breaks the StreamListenerResumePolicy class, this pattern requires that on every onNext callback completion, we capture the last processed entry, so in case of any errors we can automatically attempt to re-subscribe from the last processed entry (onNext processed). With this change there is no guarantee that the parent's onNext (i..e, the one in StreamListenerResumePolicy is going to be called after the client's onNext). For that reason, onNextEntry was the internal method used, which called client's onNext + any specific logic required by the patterns.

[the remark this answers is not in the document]
annym: It did effectively break tests in: StreamingPatternsIT. We can discuss this offline, but given that it's unrelated to this change, we can also leave it for later and focus on the actual change of the Streaming API.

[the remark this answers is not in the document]
Maithem: `StreamListenerResumePolicy` is an abstract class, concrete classes that implement it need to call super.onNext. I don't think we should keep in in the public API just to relay this "inheritance"/common code. Like this: 
https://github.com/CorfuDB/CorfuDB/pull/3047/files#diff-76c7fea3be840598202abc5bbd92ed59c484ff0e9e5e8811ed5faf1fea50872aR92

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+ * it can, this prevents superfluous polling and overwhelming the listener. Furthermore, syncing and listener
+ * notification is executed on a shared thread pool in round robin fashion, therefore if a stream produces
+ * more deltas relative to other streams that particular stream will have more CPU time than other tasks.
+ */
+
+
+@Slf4j
+public class StreamPollingScheduler {
+
+    private final Map<StreamListener, StreamingTask> allTasks = new ConcurrentHashMap<>();
```
PavelZaytsev, line 71: The updates to the allTasks are guarded with a synchronized block that  is used to implement the conditional variable. Why do we still need this map to be concurrent?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+    private void waitForTasks() throws InterruptedException {
+        synchronized (allTasks) {
+            while (allTasks.isEmpty()) {
+                allTasks.wait();
+            }
+        }
+    }
+
+    private List<StreamingTask> getTasks(StreamStatus status, Predicate<StreamingTask> predicate) {
+        List<StreamingTask> result = new ArrayList<>();
```
PavelZaytsev, line 161: `return allTasks.values().stream().filter(predicate.and(task -> task.getStatus() == status)).collect(Collectors.toList());`
just a suggestion

annym (approved):
Overall design LGTM, please address the most relevant comments! :)

hisundar:
Just a reminder from the meeting. Looking for
1. queue delay metrics to capture time spent in the queue
2. testcase that does idle table checkpoint+trim

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+            .waitForHole(true)
+            .serverCacheable(false)
+            .build();
+
+    public DeltaStream(AddressSpaceView addressSpaceView, UUID streamId, long lastAddressRead, int bufferSize) {
+        this.addressSpaceView = addressSpaceView;
+        Preconditions.checkArgument(Address.NON_ADDRESS == -1);
+        Preconditions.checkArgument(lastAddressRead >= Address.NON_ADDRESS,
+                "lastAddressRead %s must be >= %s", lastAddressRead, Address.NON_ADDRESS);
+        Preconditions.checkArgument(bufferSize > 0, "buffer size %s must be > 0", bufferSize);
```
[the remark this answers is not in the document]
Maithem: Yes. Removed.

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -415,13 +401,30 @@ public static CorfuRuntimeParametersBuilder builder() {
...
             private Codec.Type codecType = Codec.Type.ZSTD;
             private boolean metricsEnabled = true;
-            private long streamingPollingBlockingTimeMs = 5;
-            private int streamingQueueSize = 100;
-            private int streamingPollingThreadPoolSize = 2;
-            private int streamingPollingIdleWaitTimeMs = 50;
-            private int streamingNotificationThreadPoolSize = 4;
-            private long streamingNotificationBlockingTimeMs = 5;
-            private int streamingNotificationBatchSize = 50;
+            private int streamingWorkersThreadPoolSize = 2;
```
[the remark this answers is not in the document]
Maithem: I'm specifically using a small number here to force clients to tune based on how many listeners they register. I can't guess a number, should be based on running a test and observing latency.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+ * Unlike the {@link org.corfudb.runtime.view.stream.IStreamView} this stream implementation only tracks updates
+ * that haven't been checkpointed.
+ */
+@Slf4j
+public class DeltaStream {
+
+    private final AddressSpaceView addressSpaceView;
+
+    @Getter
+    private final UUID streamId;
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: Since these volatiles are private and can only be changed within the class, I think that's not a strong argument to use a AtomicLong. Anyways, I changed it.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+        this.trimMark = Address.NON_ADDRESS;
+    }
+
+    /**
+     * Check if a the last address read was trimmed, since the last address has to always be
+     * greater than the trim mark, if this condition is trim, then a trim has created a gap
+     * that hasn't be synced and cannot be recovered.
+     */
+    private boolean trimGap() {
+        return trimMark != Address.NON_ADDRESS && lastAddressRead < trimMark;
```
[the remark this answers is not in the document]
Maithem: The checkpoint writes a hole for all streams when they are being checkpointed. For inactive streams, the current pointer of the stream will point to the latest hole, that hole's address should always be greater than the trim mark.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    //max has to always be greater than last address read
```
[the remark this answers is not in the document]
Maithem: Fixed the comment.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: Added a comprehensive comment.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingTask.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    public void refresh(StreamAddressSpace streamAddressSpace) {
+        if (log.isTraceEnabled()) {
+            log.trace("refresh id {} max {} trim {} newTm {} newAddresses {}", streamId, maxAddressSeen, trimMark,
+                    streamAddressSpace.getTrimMark(), streamAddressSpace.toArray());
+        }
+
+        if (lastAddressRead < streamAddressSpace.getTrimMark()
+                || maxAddressSeen < streamAddressSpace.getTrimMark()) {
```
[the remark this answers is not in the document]
Maithem: read https://github.com/CorfuDB/CorfuDB/pull/3047/files#r721688728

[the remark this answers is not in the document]
Maithem: Also see `DeltaStreamTest` this scenario is well tested.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: We can do batch reads and that would improve efficiency, the only problem is I don't have a good way to set the batch size, and this would introduce one more tunable parameter that needs to be tuned such that it can cover all workloads. 

This can be solved, if needed, in a subsequent patch.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: I added a queuing delay metric per stream.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+ *                     ▼
+ *            ┌───────────────────┐
+ *            │      Error        │
+ *            └───────────────────┘
+ *
+ * On every tick the scheduler will poll/dispatch tasks only when needed, for example, when a listener is not
+ * consuming the deltas fast enough the scheduler can decide to not poll or schedule extra sync tasks until
+ * it can, this prevents superfluous polling and overwhelming the listener. Furthermore, syncing and listener
+ * notification is executed on a shared thread pool in round robin fashion, therefore if a stream produces
+ * more deltas relative to other streams that particular stream will have more CPU time than other tasks.
```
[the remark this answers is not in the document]
Maithem: There are two different problems that are related to starvation. One, streams that produce data at a higher rate than other streams. Two, slow streams, where the write is slow and the read suffers as a result. The first  issue is solved by consuming reads in a round robin fashion. The second issue is not solved by **neither designs**. The fundamental problem is that the read API is blocking and that would hold a worker thread during hole filling and write propagation.

Although this problem exists in both designs, this design is easier to tune, because you just need to size one thread poll as opposed to two.

Anyways, a fix to address the underlying problem is to have non-blocking read on the protocol level, which includes hole filling and write propagation. That way, other streams can still use the thread poll when there are slow writers.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: That's because while the scheduler is iterating the tasks map another thread can remove the listener.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+@Slf4j
+public class StreamPollingScheduler {
+
+    private final Map<StreamListener, StreamingTask> allTasks = new ConcurrentHashMap<>();
+
+    private final ScheduledExecutorService scheduler;
+
+    private final ExecutorService workers;
+
+    final int pollThreshold;
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingManager.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingTask.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Slf4j
+public class StreamingTask<K extends Message, V extends Message, M extends Message> implements Runnable {
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingTask.java
```diff
@@ -0,0 +1,188 @@
...
+        TableRegistry registry = runtime.getTableRegistry();
+        final UUID streamId = TableRegistry.getStreamIdForStreamTag(namespace, streamTag);
+        this.stream = new DeltaStream(runtime.getAddressSpaceView(), streamId, address, bufferSize);
+        this.tableSchemas = tablesOfInterest
+                .stream()
+                .collect(Collectors.toMap(
+                        tName -> CorfuRuntime.getStreamID(TableRegistry.getFullyQualifiedTableName(namespace, tName)),
+                        tName -> {
+                            // The table should be opened with full schema before subscription.
+                            Table<K, V, M> t = registry.getTable(namespace, tName);
```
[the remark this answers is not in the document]
Maithem: I just copied it from the old code.

Since this API has already been released I don't want to change this exception type.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamingTask.java
```diff
@@ -0,0 +1,188 @@
...
+        return Optional.of(new CorfuStreamEntries(streamEntries, timestamp));
+    }
+
+    public DeltaStream getStream() {
+        return this.stream;
+    }
+
+    private void produce() {
+        Preconditions.checkState(status.get() == StreamStatus.SYNCING);
+        Preconditions.checkState(stream.hasNext());
```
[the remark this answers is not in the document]
Maithem: Yes, this is by design because the scheduler will not dispatch this task to the worker pool unless `hasNext` returns true. If this precondition fails it means there is a concurrency issue and I would like it to explicitly fail. 

```java

private void dispatchSyncTasks(List<StreamingTask> tasks) {
        for (StreamingTask task : tasks) {
            if (task.getStream().hasNext()) {
                // this will fail for tasks that are already syncing
                // for tasks to be scheduled to sync, they must be in the scheduling state
                task.move(StreamStatus.SCHEDULING, StreamStatus.SYNCING);
                workers.execute(task);
            } else {
                // no new deltas to sync
                // This stream has no entries to sync after the poll, set it to runnable so that it can
                // be polled next scheduling cycle
                task.move(StreamStatus.SCHEDULING, StreamStatus.RUNNABLE);
            }
        }
    }

```

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -1345,7 +1174,7 @@ public void testSubscribeAPIForTagsOnly() throws Exception {
                 Timestamp.newBuilder().setSequence(Address.NON_ADDRESS).setEpoch(0L).build());
 
         // Confirm totalUpdates are received
-        TimeUnit.MILLISECONDS.sleep(sleepTime);
+        TimeUnit.MILLISECONDS.sleep(sleepTime * 2);
```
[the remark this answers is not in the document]
Maithem: In certain cases it is slower, because we are not continuously polling there could be some lag. But these tests are failing even without this change. Here's a failure from another PR https://pipelines.actions.githubusercontent.com/NnzYamcbKeYKYrtwq5VEoRtMbxsjokL1h7qBUldebVzgKpZkWn/_apis/pipelines/1/runs/2638/signedlogcontent/8?urlExpires=2021-10-04T20%3A42%3A16.4698112Z&urlSigningMethod=HMACV1&urlSignature=XjzBqYX9O1wjs7xPT4%2BMj0%2BLRrCslJx7ATtH7tC4uFY%3D

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: Since it only holds two longs, it doesn't really matter.

Maithem: Changed.

runtime/src/main/java/org/corfudb/runtime/collections/StreamListener.java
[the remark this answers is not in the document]
Maithem: Reverted.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+
+    @Data
+    class Interval {
+        private final long x;
+        private final long y;
+    }
+
+    private Interval merge(Interval interval, StreamAddressRange range) {
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+        long newMinX = Math.min(interval.x, range.getEnd());
+        long newMaxY = Math.max(interval.y, range.getStart());
+        return new Interval(newMinX, newMaxY);
+    }
+
+    // Generate a range per stream that contains all sub-ranges that need to be queries
```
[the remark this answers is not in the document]
Maithem: Removed.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+
+    /**
+     * It's possible that multiple streaming tasks need to query different/overlapping ranges
+     * of the same stream. In order to prevent issuing multiple queries for the same stream,
+     * all range queries for the same stream are coalesced into a single request.
+     * For example, the following queries will all be merged into one query (i.e., (-1, 100])
```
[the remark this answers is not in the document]
Maithem: Added.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: Its by design. I want it to fail the scheduler thread explicitly to make clear any concurrency issues. In other words, this is a fault that cannot be ignored.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+        // Need to remove the tasks from allTasks before propagating the error just incase they try to
+        // re-register the task?
+        failedTasks.forEach(t -> allTasks.remove(t.getListener()));
+
+        for (StreamingTask task : failedTasks) {
+            workers.execute(task::propagateError);
+        }
+    }
+
+    public void schedule() throws InterruptedException {
```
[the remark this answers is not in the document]
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: Yes. Imagine an empty buffer of size 10, now imagine the scheduler discovers 10 addresses and replenishes that buffer.  Its possible that the consume doesn't consume those 10 addresses, so by the next poll the scheduler ignores that stream. This is the main throttling mechanism.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Maithem: No it i means its waiting to be polled. 

This state can be eliminated from a single threaded scheduler, but the reason I introduced it is because it will make a multi-threaded scheduler implementation straight forward, if it is needed.

test/src/test/java/org/corfudb/integration/StreamingIT.java
```diff
@@ -381,41 +378,6 @@ public void testStreamingSingleTableEmptyTag() throws Exception {
...
-        store.subscribeListener(listenerCommon, "test_namespace", "sample_streamer_1",
-                Collections.singletonList("tableA"), ts1);
-
-        // Attempt to subscribe to streaming updates from tableB using same listener
-        assertThrows(StreamingException.class, () -> store.subscribeListener(listenerCommon, "test_namespace", "sample_streamer_4",
-                Collections.singletonList("tableB"), ts1));
-
-        assertThat(shutdownCorfuServer(corfuServer)).isTrue();
-    }
-
```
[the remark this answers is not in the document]
Maithem: Yes, its covered by a UT `StreamPollingSchedulerTest::testAddTask`

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
Maithem: Not sure what you mean. Add it to each log? or added on creation ?

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -0,0 +1,321 @@
...
+     * For example, the following queries will all be merged into one query (i.e., (-1, 100])
+     *  Query1  (-1, 100)
+     *  Query2  (2, 5)
+     *  Query3  (3, 10)
+     */
+    private List<StreamAddressRange> coalesce(List<StreamAddressRange> ranges) {
+        Map<UUID, Interval> adjusted = new HashMap<>();
+
+        for (StreamAddressRange range : ranges) {
+            Interval interval = adjusted.getOrDefault(range.getStreamID(), new Interval(Long.MAX_VALUE, Long.MIN_VALUE));
```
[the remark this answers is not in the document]
Maithem: To me that seems like a specific use case, but shouldn't be enforced by the StreamingManager itself because it can restrict other use cases. Some of our tests actually register two different listener on the same tag. Anyways, the overhead of this is coalescing logic is negligible (based on profiling) for the case you mentioned. 

I think its better to track the number of sequence calls the scheduler emits per scheduling tick, that metric will reflect how efficiently we are batching.

Maithem:
Addressed all comments.

achinmishra (approved):

PR #3049, opened 2021-09-15
Enable Corfu browser to print table at a given snapshot

Lujie1996:
## Overview

Description:
Enable Corfu browser to print table at a given snapshot. It allows to inspect table content in limited history.

Why should this be merged: 
Sometimes when debugging we need to inspect tables at a certain point of time in history. 

- [ ] Public API has Javadoc

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
@@ -1,5 +1,6 @@
 package org.corfudb.browser;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 
@@ -12,8 +13,10 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableSet;
 import java.util.Random;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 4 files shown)


corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditorMain.java
```diff
@@ -40,6 +41,8 @@
         "[--keystore=<keystore_file>] [--ks_password=<keystore_password>] " +
         "[--truststore=<truststore_file>] [--truststore_password=<truststore_password>] " +
         "[--diskPath=<pathToTempDirForLargeTables>] "+
+        "[--diskPath=<pathToTempDirForLargeTables>] " +
```
annym, line 44: duplicated param

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditorMain.java
```diff
@@ -60,6 +63,8 @@
         + "--truststore=<truststore_file> TrustStore File\n"
         + "--truststore_password=<truststore_password> Truststore Password\n"
         + "--diskPath=<pathToTempDirForLargeTables> Path to Temp Dir\n"
+        + "--epoch=<epoch> Epoch of snapshot\n"
```
annym, line 66: Is the epoch really needed? I believe whichever epoch you put will work as we base it on the Timestamp instead.. that would make things easier for users..

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
```diff
@@ -93,6 +95,10 @@ public CorfuStoreBrowserEditor(CorfuRuntime runtime, String diskPath) {
         runtime.getSerializers().registerSerializer(dynamicProtobufSerializer);
     }
 
+    public void setSnapshot(Token snapshot) {
```
annym, line 98: I think this change should come along with another utility that gives a list of available timestamps (otherwise it will be hard to be used by a common user). Something of the type --table myTable --getSnapshots

annym:
@Lujie1996 when you find time can you address the pending comments? I think this is really useful and should be merged!

PR #3050, opened 2021-09-15
StreamingManager: Double stream threads for perf issues

hisundar:
## Overview

Description:
Current StreamManager design suffers from noisy neighbour problem. If more than 4 subscribers are slow to process notifications then all the other subscribers lag. Increase parallelism as a stop-gap measure.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
+++ b/runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
@@ -367,12 +367,12 @@ public class CorfuRuntime {
         /*
          * Total number of threads in Polling Executor Pool (shared across all listeners)
          */
-        private int streamingPollingThreadPoolSize = 2;
+        private int streamingPollingThreadPoolSize = 4;
 
         /*
          * Total number of threads in Notification Executor Pool (shared across all listeners)
          */
-        private int streamingNotificationThreadPoolSize = 4;
+        private int streamingNotificationThreadPoolSize = 8;
 
         /*
          * Total time in milliseconds to block for new updates to appear in the queue, if empty.


Lujie1996 (approved):
Looks good. One question though, what if there are more slow subscribers?

pankti-m (approved):

shama358 (approved):

apavan (approved):

runtime/src/main/java/org/corfudb/runtime/CorfuRuntime.java
```diff
@@ -367,12 +367,12 @@
...
          * Total number of threads in Polling Executor Pool (shared across all listeners)
          */
-        private int streamingPollingThreadPoolSize = 2;
+        private int streamingPollingThreadPoolSize = 4;
 
         /*
          * Total number of threads in Notification Executor Pool (shared across all listeners)
          */
-        private int streamingNotificationThreadPoolSize = 4;
+        private int streamingNotificationThreadPoolSize = 8;
```
Maithem, line 375: This is a configurable parameter, the clients should  increase it as needed. It doesn't make sense as a default. The design is based on the listeners being non-blocking and this thread pool just moves items from a queue and invokes the listener.

hisundar: will make a follow up change to make it a runtime parameter.

hisundar: It just seemed easier to modify the default than seeking out the clients parameters, but I see your point
I will try to undo this change once the client values are fixed up.

Maithem: >  than seeking out the clients parameters

What does this even mean ? 


This PR has no effect, the parameters you changed will not be reflected for any client that uses the CorfuRuntimeParametersBuilder


This being the main repo, I don't think its a good idea to just try random patches.

apavan: @Maithem good catch

PR #3053, opened 2021-09-15
Fixing some StaticCodeAnalysis issues

SravanthiAshokKumar:
1. Fixing "An instanceof check is being performed on the caught exception. Create a separate catch clause for this exception type." in two files - runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java and runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
2. Fixing "Avoid using equals() to compare against null" in the file - test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -169,7 +169,7 @@ public abstract class AbstractView {
                         throw re;
                     }
                 } else {
-                    log.warn("layoutHelper: An unknown exception occured");
+                    log.warn("layoutHelper: An unknown RuntimeException occured");
                     throw re;
                 }
             }
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -169,6 +169,7 @@ public abstract class AbstractView {
                         throw re;
                     }
                 } else {
+                    log.warn("layoutHelper: An unknown exception occured");
                     throw re;
                 }
             }
--- a/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
@@ -139,28 +139,36 @@ public abstract class AbstractView {
                     }
                     return rLayout;
                 }));
+            } catch (ServerNotReadyException se) {
+                log.warn("Server still not ready. Waiting for server to start "
+                        + "accepting requests.");
+                if (rethrowAllExceptions) {
+                    throw se;
+                }
+            } catch (WrongEpochException we) {
+                log.warn("Got a wrong epoch exception, updating epoch to {} and "
+                        + "invalidate view", we.getCorrectEpoch());
+                if (rethrowAllExceptions) {
+                    throw we;
+                }
+            } catch (NetworkException ne) {
+                log.warn("layoutHelper: System seems unavailable", ne);
+                if (rethrowAllExceptions) {
+                    throw ne;
+                }
+            } catch (WrongClusterException wce) {
+                log.warn("layoutHelper: Cluster reconfiguration or incorrect cluster", wce);
+                log.info("layoutHelper: Invoking the systemDownHandler.");
+                runtime.getParameters().getSystemDownHandler().run();
+                throw wce;
             } catch (RuntimeException re) {
                 if (re.getCause() instanceof TimeoutException) {
                     log.warn("Timeout executing remote call, invalidating view and retrying "
                             + "in {}s", retryRate);
-                } else if (re instanceof ServerNotReadyException) {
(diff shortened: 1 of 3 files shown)


test/src/test/java/org/corfudb/runtime/view/stream/StreamAddressSpaceTest.java
```diff
@@ -217,7 +217,7 @@ public void testToArray() {
     @Test
     @SuppressWarnings("ConstantConditions")
     public void testEquality() {
-        assertThat(new StreamAddressSpace().equals(null)).isFalse();
+        assertThat(new StreamAddressSpace() == null).isFalse();
```
zfrenette, line 220: This was explicitly testing the overridden equals method. If I make it return `true` instead of `false`, the test now passes when it shouldn't.

SravanthiAshokKumar: My bad, will revert it.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -139,28 +139,36 @@ private Layout getLayoutUninterruptibly() {
...
+                }
+            } catch (WrongClusterException wce) {
+                log.warn("layoutHelper: Cluster reconfiguration or incorrect cluster", wce);
+                log.info("layoutHelper: Invoking the systemDownHandler.");
+                runtime.getParameters().getSystemDownHandler().run();
+                throw wce;
             } catch (RuntimeException re) {
                 if (re.getCause() instanceof TimeoutException) {
                     log.warn("Timeout executing remote call, invalidating view and retrying "
                             + "in {}s", retryRate);
```
chetangudisagar, line 167: Can this TimeoutException also be caught outside similar to others?
Then, in the remaining space of this catch clause, maybe we can leave a warning message saying "`layoutHelper: An unknown RuntimeException occurred.`" and throw it back to the caller.

SravanthiAshokKumar: The TimeoutException seems to be wrapped within the RuntimeException. So it's not possible to catch it directly.
Although can add the unknown exception log in the else block.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -139,28 +139,37 @@ private Layout getLayoutUninterruptibly() {
...
-                    throw re;
-                }else {
-                    throw re;
-                }
-                if (rethrowAllExceptions) {
+                    if (rethrowAllExceptions) {
+                        throw re;
+                    }
+                } else {
+                    log.warn("layoutHelper: An unknown exception occured");
```
chetangudisagar: typo: occurred

SravanthiAshokKumar: Yes, this was the previous push, changed it in the latest iteration

chetangudisagar (approved):
LGTM.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -607,21 +607,24 @@ public void prefixTrim(final Token address, boolean retry) {
...
+            } catch (WrongEpochException we) {
+                long serverEpoch = ((WrongEpochException) we).getCorrectEpoch();
+                long runtimeEpoch = runtime.getLayoutView().getLayout().getEpoch();
+                log.warn("prefixTrim[{}]: WrongEpochException, runtime is in epoch {}, while server " +
+                                "is in epoch {}. Invalidate layout for this client and retry, attempt: {}/{}",
+                        address, runtimeEpoch, serverEpoch, x + 1, numRetries);
+                runtime.invalidateLayout();
             } catch (RuntimeException e) {
-                // NetworkException is RuntimeException, CFUtils casts it into a RuntimeException.
-                // TimeoutException is a checked exception, CFUtils wraps it in a RuntimeException.
```
zfrenette, line 612: nit: Consider leaving these comments.

runtime/src/main/java/org/corfudb/runtime/view/AbstractView.java
```diff
@@ -139,28 +139,37 @@ private Layout getLayoutUninterruptibly() {
...
-                    throw re;
-                }else {
-                    throw re;
-                }
-                if (rethrowAllExceptions) {
+                    if (rethrowAllExceptions) {
+                        throw re;
+                    }
+                } else {
+                    log.warn("layoutHelper: An unknown RuntimeException occured");
```
zfrenette, line 172: What is the purpose of introducing this log message? It doesn't seem like it provides any additional information, and in this case, the exception is re-thrown anyway.

zfrenette (approved):
Looks good. Just some minor comments. Don't forget to update the commit message.

PR #3056, opened 2021-09-22
update the documents of metrics

Jas0nch:
## Overview

Description: update the docs of metrics.

Why should this be merged: 

- [ ] Public API has Javadoc


Maithem (commented):
We should rethink this documentation process, instead of having these manual changes we should automate the metrics documentation. Maybe consider something like this
http://hadoop.apache.org/docs/stable/api/org/apache/hadoop/metrics2/annotation/package-summary.html ?

Jas0nch:
> We should rethink this documentation process, instead of having these manual changes we should automate the metrics documentation. Maybe consider something like this
> http://hadoop.apache.org/docs/stable/api/org/apache/hadoop/metrics2/annotation/package-summary.html ?

Yeah... If we want to do this automatically, the first thing would be putting all metric together, at least their names. So maybe we should refactor the code.

PR #2848, opened 2021-01-04
Handle TAE for LR sync status update

zhangn49:
## Overview

Description:

Why should this be merged: 

Related issue(s) (if applicable): #<number>

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #3062, opened 2021-09-28
Handle TransactionAbortedException on metadata table updates

annym:
## Overview

Description:

This was addressed in PR #2848, but this code path was not covered.

Why should this be merged: bug fix in client

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
@@ -12,9 +12,14 @@ import org.corfudb.runtime.LogReplication.LogReplicationEntryMetadataMsg;
 import org.corfudb.runtime.LogReplication.LogReplicationEntryMsg;
 import org.corfudb.runtime.LogReplication.LogReplicationEntryType;
 import org.corfudb.runtime.collections.TxnContext;
+import org.corfudb.runtime.exceptions.TransactionAbortedException;
+import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuInterruptedError;
 import org.corfudb.runtime.view.Address;
 import org.corfudb.runtime.view.StreamOptions;
 import org.corfudb.runtime.view.stream.OpaqueStream;
+import org.corfudb.util.retry.IRetry;
+import org.corfudb.util.retry.IntervalRetry;
+import org.corfudb.util.retry.RetryNeededException;
 import org.corfudb.util.serializer.Serializers;
 
 import javax.annotation.concurrent.NotThreadSafe;
@@ -102,9 +107,7 @@ public class StreamsSnapshotWriter implements SnapshotWriter {
 
     /**
      * Clear all tables registered
-     *
-     * TODO: replace with stream API
-     */
+     **/
     private void clearTables() {
 
         long persistedTopologyConfigId;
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -102,9 +107,7 @@ private void initializeShadowStreams(LogReplicationConfig config) {
 
     /**
      * Clear all tables registered
-     *
-     * TODO: replace with stream API
-     */
+     **/
```
Lujie1996, line 110: nit: one star ([code convention](https://www.oracle.com/java/technologies/javase/codeconventions-comments.html))

Lujie1996 (approved):

pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
```diff
@@ -164,7 +164,7 @@ public long TXEnd() throws TransactionAbortedException {
         try {
             return TransactionalContext.getCurrentContext().commitTransaction();
         } catch (TransactionAbortedException e) {
-            log.warn("TXEnd[{}] Aborted Exception {}", context, e);
+            log.warn("TXEnd[{}] Aborted Exception ", context, e);
```
Jas0nch, line 167: this log takes two variables but only print one of them?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -136,19 +139,37 @@ private void clearTables() {
...
+            IRetry.build(IntervalRetry.class, () -> {
+                try (TxnContext txnContext = logReplicationMetadataManager.getTxnContext()) {
+                    logReplicationMetadataManager.appendUpdate(txnContext, LogReplicationMetadataType.TOPOLOGY_CONFIG_ID, topologyConfigId);
+                    clearStreams(txnContext);
+                    txnContext.commit();
+                } catch (TransactionAbortedException tae) {
+                    log.error("Error while attempting to clear tables.", tae);
+                    throw new RetryNeededException();
                 }
+                return null;
```
Jas0nch, line 156: this return null means this part of code will only run one time, if there's no return, this part of code will retry many times?

runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
annym: The exceptions don't require a placeholder, internally it just prints it at the end of the message. Previously, we were printing something like this:

2021-09-27T09:36:55.654Z | WARN  |              discovery-service |   o.c.runtime.view.ObjectsView | TXEnd[TX[c489]] Aborted Exception {}
org.corfudb.runtime.exceptions.TransactionAbortedException: ...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
annym: The retryFunction expects a return value, since we don't have any return value we can just return null. The part of the code that will know a retry is required is line 154: throw new RetryNeededException(), which we throw if we catch a TransactionAbortedException, so the transaction is retried.

PR #3067, opened 2021-10-01
Fix Cache Stats Reporting

Maithem:
## Overview

This patch makes the cache stats reflect relative stats within
the reporting duration as opposed to having a running total since
the cache was instantiated.
Why should this be merged: 

Related issue(s) (if applicable): Reporting cache stats since the cache instantiation isn't helpful when trying to debug issues that happen in short time windows. This patch makes the stats reporting for a specific time window (i.e., poll period). 

- [x] Public API has Javadoc

### DIFF
--- a/common/src/main/java/org/corfudb/common/metrics/micrometer/CacheMetrics.java
+++ b/common/src/main/java/org/corfudb/common/metrics/micrometer/CacheMetrics.java
@@ -1,0 +1,65 @@
+package org.corfudb.common.metrics.micrometer;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheStats;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/corfudb/common/metrics/micrometer/CacheMetrics.java
```diff
@@ -0,0 +1,65 @@
...
+        if (metricsRegistry.isPresent()) {
+
+            Gauge.builder(cacheName, new CacheStatsContainer(cache.stats()), data -> {
+                CacheStats currentSnapshot = cache.stats();
+                data.snapshot = currentSnapshot.minus(data.snapshot);
+                return data.snapshot.hitRate();
+            }).tags("result", "hitRate")
+                    .strongReference(true)
+                    .register(metricsRegistry.get());
+
```
Maithem, line 38: @PavelZaytsev Do you know of a cleaner way to snapshot a single object once for many gauges ?

PR #3068, opened 2021-10-04
Optimize scanAndFilter for Immutable Objects

Maithem:
## Overview
Table.java, which is a wrapper on top of CorfuTable only allows for
immutable objects types - protobuf types - and doesn't need a
write lock to execute the scan queries. This patch reduces locking
by allowing multiple threads to execute the scan simultaneously on
different versions of the Table.

Why should this be merged: Reduces locking for concurrent scanning queries. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
@@ -82,15 +82,15 @@ public class Table<K extends Message, V extends Message, M extends Message> {
     /**
      * Returns a Table instance backed by a CorfuTable.
      *
-     * @param tableParameters         Table parameters including
-     *                                table's namespace and fullyQualifiedTableName,
-     *                                key, value, and metadata classes,
-     *                                value and metadata schemas
-     * @param corfuRuntime            connected instance of the Corfu Runtime
-     * @param serializer              protobuf serializer
-     * @param streamingMapSupplier    supplier of underlying map data structure
-     * @param versionPolicy           versioning policy
-     * @param streamTags              set of UUIDs representing the streamTags
+     * @param tableParameters      Table parameters including
+     *                             table's namespace and fullyQualifiedTableName,
+     *                             key, value, and metadata classes,
+     *                             value and metadata schemas
+     * @param corfuRuntime         connected instance of the Corfu Runtime
+     * @param serializer           protobuf serializer
+     * @param streamingMapSupplier supplier of underlying map data structure
+     * @param versionPolicy        versioning policy
+     * @param streamTags           set of UUIDs representing the streamTags
      */
     public Table(@Nonnull final TableParameters<K, V, M> tableParameters,
                  @Nonnull final CorfuRuntime corfuRuntime,
@@ -173,7 +173,7 @@ public class Table<K extends Message, V extends Message, M extends Message> {
      * Clear All table entries.
      */
     public void clearAll() {
-       corfuTable.clear();
+        corfuTable.clear();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


achinmishra (approved):
I have profiled clients around scanAndFilter API and this would be a good enhancement in performance because scanAndFilter is the most common api being used to query corfu. 
Here, the original scanAndFilterByEntry is moved to be issued in parallel and it would be a good exercise to determine the performance of the apis after this change has been merged to the product side.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -351,17 +350,20 @@ public int count() {
...
-                        recordEntry.getKey(),
-                        recordEntry.getValue().getPayload(),
-                        recordEntry.getValue().getMetadata())))
-                .parallelStream()
-                .map(entry -> new CorfuStoreEntry<>(
-                        entry.getKey(),
-                        entry.getValue().getPayload(),
-                        entry.getValue().getMetadata()))
-                .collect(Collectors.toList());
+        try(Stream<Map.Entry<K, CorfuRecord<V, M>>> stream = corfuTable.entryStream()) {
```
zhangn49, line 353: I think original impl does not have a deep copy that introduced by "ImmutableListSetWrapper.fromMap".

This change will have higher latency caused by the deep copy, right?

Maithem: It's a trade-off between latency vs throughput. The copy only creates a new map "structure", but doesn't do a copy on the key/value. 

```java

public static <K, V> ImmutableListSetWrapper<Entry<K, V>> fromMap(@NonNull Map<K, V> original) {
        return new ImmutableListSetWrapper<>(original
                .entrySet()
                .stream()
                .map(entry -> new SimpleImmutableEntry<>(entry.getKey(), entry.getValue()))
                .collect(ImmutableList.toImmutableList()));
```

The reason this is safe, that is we don't need  a deep copy, is because Protobuf objects are immutable. More often than note (at least based on my observations), making a shallow copy of the map is much cheaper than waiting on the VLO lock.

runtime/src/main/java/org/corfudb/runtime/collections/Table.java
```diff
@@ -351,17 +350,20 @@ public int count() {
...
+            return CorfuTable.pool.submit(() -> stream
+                    .filter(recordEntry ->
+                            entryPredicate.test(new CorfuStoreEntry<>(
+                                    recordEntry.getKey(),
+                                    recordEntry.getValue().getPayload(),
+                                    recordEntry.getValue().getMetadata())))
+                    .parallel()
```
Lujie1996, line 360: Question: why not put parallel() in front of filter()?

Maithem: I think it automatically detects it. Try this

```java

    public final static ForkJoinPool pool = new ForkJoinPool(Math.max(Runtime.getRuntime().availableProcessors() - 1, 1),
            pool -> {
                final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
                worker.setName("CorfuTable-Forkjoin-pool-" + worker.getPoolIndex());
                return worker;
            }, null, true);

    @Test
    public void blah() {
        StreamingMapDecorator map = new StreamingMapDecorator<>(new HashMap<>());

        int numKeys = 100;
        for (int x = 0; x < numKeys; x++) {
            map.put(x + "", x + "");
        }

        Object res = pool.submit(() -> map.entryStream()
                .filter(t -> {
                    System.out.println(Thread.currentThread().getId() + " filter " + t);
                    return true;
                })
                .map(t -> {
                    System.out.println(Thread.currentThread().getId() + " map " + t);
                    return t;})
                .collect(Collectors.toList())).join();

    }
```

Although I'm still not certain on how this detection works.

Lujie1996: Right. I ran the test and found that no matter where I put parallel(), the other intermediate operations (filter and map) are always parallelized. Amazing.

Lujie1996 (approved):

zhangn49 (approved):

PR #3083, opened 2021-10-12
Move Schema options to TableOptions

hisundar:
## Overview

Description:

Why should this be merged: 

There are many tables which share the same protobuf definition but require different behavior for the corfu features like
stream tags, replication etc.
Before this change, these table options were extracted from the message level options which means there was
no way to differentiate between different tables that share the same schema definition.
Moving the corfu schema options to TableOptions allows clients to define different behavior while still sharing the same underlying protobuf message

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
@@ -77,21 +77,21 @@ public class LogReplicationMetadataManager {
                             LogReplicationMetadataKey.class,
                             LogReplicationMetadataVal.class,
                             null,
-                            TableOptions.builder().build());
+                            TableOptions.fromProtoSchema(LogReplicationMetadataVal.class));
 
             this.replicationStatusTable = this.corfuStore.openTable(NAMESPACE,
                             REPLICATION_STATUS_TABLE,
                             ReplicationStatusKey.class,
                             ReplicationStatusVal.class,
                             null,
-                            TableOptions.builder().build());
+                            TableOptions.fromProtoSchema(ReplicationStatusVal.class));
 
             this.replicationEventTable = this.corfuStore.openTable(NAMESPACE,
                     REPLICATION_EVENT_TABLE_NAME,
                     ReplicationEventKey.class,
                     ReplicationEvent.class,
                     null,
-                    TableOptions.builder().build());
+                    TableOptions.fromProtoSchema(ReplicationEvent.class));
 
             this.localClusterId = localClusterId;
         } catch (Exception e) {
--- a/runtime/proto/corfu_options.proto
+++ b/runtime/proto/corfu_options.proto
@@ -8,7 +8,7 @@ import "google/protobuf/descriptor.proto";
 // Option tags to be used by the CorfuStore consumers to tag special fields to be detected by Corfu.
 message SchemaOptions {
     // Secondary keys to be indexed.
-    optional bool secondary_key = 1;
+    optional bool secondary_key_deprecated = 1;
     // Version number in metadata field.
     optional bool version = 2;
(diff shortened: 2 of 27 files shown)


annym:
> @annym - this latest patch does not carry the secondary_key cleanup we discussed, i will work on that shortly

The CorfuStoreSecondaryIndexTest are failing!

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreSecondaryIndexTest.java
```diff
@@ -1064,13 +1064,13 @@ public void testInvalidNestedSecondaryIndexRoot() throws Exception {
         final String tableName = "ManagedMetadata";
 
         // Create & Register the table.
-        assertThrows(IllegalArgumentException.class, () -> shimStore.openTable(
+        assertThrows(UnsupportedOperationException.class, () -> shimStore.openTable(
```
annym, line 1067: Why did this exception type change? Isn't it closer to an IllegalArgument as the secondary index is incorrectly defined, rather than an operation not supported.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -91,7 +91,7 @@ public LogReplicationMetadataManager(CorfuRuntime rt, long topologyConfigId, Str
                     ReplicationEventKey.class,
                     ReplicationEvent.class,
                     null,
-                    TableOptions.builder().build());
```
annym, line 94: The tests failing are because in this file Line 87 you should fetch table Option from proto schema.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreSecondaryIndexTest.java
Lujie1996: +1. I ran the test and it's throwing `UnsupportedOperationException ` because of 
```
Caused by: java.lang.UnsupportedOperationException: This field is not of message type.
	at com.google.protobuf.Descriptors$FieldDescriptor.getMessageType(Descriptors.java:1090)
	at org.corfudb.runtime.collections.ProtobufIndexer.validateNestedSecondaryKey(ProtobufIndexer.java:345)
	at org.corfudb.runtime.collections.ProtobufIndexer.registerNestedSecondaryIndex(ProtobufIndexer.java:303)
```
I think we should catch `UnsupportedOperationException` in `ProtobufIndexer.validateNestedSecondaryKey` and throw `IllegalArgumentException` instead.

Lujie1996 (approved):
LGTM despite minor things - refactor tests on Github side where TableOption is used (I'm fine for now since it's only on Corfu internally), and exceptions.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreSecondaryIndexTest.java
hisundar: done

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
hisundar: done. thanks so much!

runtime/proto/corfu_options.proto
```diff
@@ -8,7 +8,7 @@ import "google/protobuf/descriptor.proto";
 // Option tags to be used by the CorfuStore consumers to tag special fields to be detected by Corfu.
 message SchemaOptions {
     // Secondary keys to be indexed.
-    optional bool secondary_key = 1;
+    optional bool secondary_key_deprecated = 1;
```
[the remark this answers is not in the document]
annym: @hisundar this?

PR #3095, opened 2021-10-26
Remove Transaction Logging and Transaction Stream

pankti-m:
With stream tags, transaction stream(and hence transaction logging) are
no longer required.

Why should this be merged:  Code Cleanup

- [ ] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
@@ -39,18 +39,12 @@ import java.util.concurrent.ConcurrentHashMap;
 @Slf4j
 public class ObjectsView extends AbstractView {
 
-    /**
-     * The Transaction stream is used to log/write successful transactions from different clients.
-     * Transaction data and meta data can be obtained by reading this stream.
-     */
-    @Deprecated
-    public static final UUID TRANSACTION_STREAM_ID = CorfuRuntime.getStreamID("Transaction_Stream");
+    private static final String LOG_REPLICATOR_STREAM_NAME =
+        "LR_Transaction_Stream";
 
-    // We are temporarily naming this stream with the same name as TRANSACTION_STREAM_ID to avoid breaking LR code
-    // while transition to UFO is completed (once migration is complete to UFO this stream can be renamed)
-    // add a prefix to the name: e.g., org.corfudb.logreplication.transactionstream
     public static final StreamTagInfo LOG_REPLICATOR_STREAM_INFO =
-            new StreamTagInfo("Transaction_Stream", CorfuRuntime.getStreamID("Transaction_Stream"));
+            new StreamTagInfo(LOG_REPLICATOR_STREAM_NAME,
+                CorfuRuntime.getStreamID(LOG_REPLICATOR_STREAM_NAME));
 
     /**
      * @return the ID of the log replicator stream.
@@ -58,10 +52,6 @@ public class ObjectsView extends AbstractView {
     public static UUID getLogReplicatorStreamId() {
         return LOG_REPLICATOR_STREAM_INFO.getStreamId();
     }
-
-    @Getter
-    @Setter
-    boolean transactionLogging = false;
 
     @Getter
     Map<ObjectID, Object> objectCache = new ConcurrentHashMap<>();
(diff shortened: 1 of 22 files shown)


runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
```diff
@@ -39,18 +39,11 @@
 @Slf4j
 public class ObjectsView extends AbstractView {
 
-    /**
-     * The Transaction stream is used to log/write successful transactions from different clients.
-     * Transaction data and meta data can be obtained by reading this stream.
-     */
-    @Deprecated
-    public static final UUID TRANSACTION_STREAM_ID = CorfuRuntime.getStreamID("Transaction_Stream");
+    private static final String LOG_REPLICATOR_STREAM_NAME = "Transaction_Stream";
```
annym, line 42: let's change the name of this stream to be suggestive to LR, perhaps "LR_Transaction_Stream"

test/src/test/java/org/corfudb/infrastructure/logreplication/LogUpdateAPITest.java
```diff
@@ -83,10 +83,10 @@ public void testUFOWithLogUpdate() throws NoSuchMethodException, IllegalAccessEx
         Stream streamA = (new OpaqueStream(runtime2.getStreamsView().
                 get(uuidA, options))).streamUpTo(runtime2.getAddressSpaceView().getLogTail());
 
-        IStreamView txStream = runtime2.getStreamsView()
+        /*IStreamView txStream = runtime2.getStreamsView()
```
annym, line 86: this can be removed

test/src/test/java/org/corfudb/runtime/view/ObjectsViewTest.java
```diff
@@ -39,20 +39,21 @@ public void canAbortNoTransaction()
...
-        //Enable transaction logging
-        CorfuRuntime r = getDefaultRuntime()
-                .setTransactionLogging(true);
+        final String streamA = "streamA";
+        CorfuRuntime r = getDefaultRuntime();
 
         CorfuTable<String, String> map = getDefaultRuntime().getObjectsView()
                 .build()
                 .setStreamName(mapA)
+                .setStreamTags(CorfuRuntime.getStreamID(streamA))
```
annym, line 48: I don't understand what is being done here, can you explain?

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -132,7 +132,10 @@ public void testRemainingUpToWithTrim() {
                 .isCheckpointCapable(false)
                 .build();
 
-        IStreamView txStream = runtime.getStreamsView().get(ObjectsView.TRANSACTION_STREAM_ID, options);
+        String streamTag = "tx_stream";
+        UUID streamTagId = CorfuRuntime.getStreamID(streamTag);
```
annym, line 136: same here, why is this being done? This streamTag is not used anywhere, right?

test/src/test/java/org/corfudb/runtime/view/StreamViewTest.java
```diff
@@ -480,17 +483,19 @@ private void synchronousPrefixTrim(CorfuRuntime runtime, long epoch, long addres
      */
     @Test
     public void txLogTrim() {
+        final String streamTag = "tx_stream_tag";
+        final UUID tagId = CorfuRuntime.getStreamID(streamTag);
```
annym, line 487: same here, what is this fake tagId representing?

pankti-m: same as before.. using the tag to mimic the behavior of the transaction stream.  The tag is being used when we try to get the streamView in L516.

test/src/test/java/org/corfudb/runtime/view/ObjectsViewTest.java
pankti-m: This test verifies that if a tx aborted, its updates will not be present in the transaction stream.  Since we have removed the transaction stream, I am not using the tag "streamA" for the table(mapA).  

It is also possible to test without using the tags and only polling on the stream id for mapA, but it will require more changes to the test.  Let me know if you would like to get more background and discuss further.

annym: Ok, yeah sounds reasonable to keep testing this on the stream tags. Looks good!

annym (approved):

PR #3101, opened 2021-10-29
LR: replicate registryTable

zhangn49:
## Overview

Description:

Why should this be merged: 

One stream can have log data on a standby cluster, where nobody opens this stream.
Compactor will not see this stream in the tableRegistry and will trim this stream's data.

Later, the site becomes active, but this stream was trimmed and LR hit trimmed exception.

So we need to replicate tableRegistry across clusters, and it is merge only - we do not wipe out data for this table.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/utils/LogReplicationStreamNameTableManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/utils/LogReplicationStreamNameTableManager.java
@@ -10,6 +10,7 @@ import org.corfudb.runtime.collections.Table;
 import org.corfudb.runtime.collections.TableOptions;
 import org.corfudb.runtime.collections.TxnContext;
 import org.corfudb.runtime.exceptions.unrecoverable.UnrecoverableCorfuError;
+import org.corfudb.runtime.view.TableRegistry;
 import org.corfudb.utils.CommonTypes;
 import org.corfudb.utils.LogReplicationStreams;
 import org.corfudb.utils.LogReplicationStreams.VersionString;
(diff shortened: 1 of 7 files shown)


test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
```diff
@@ -361,12 +362,28 @@ public void testStandbyStreaming() throws Exception {
             // Confirm data does not exist on Standby Cluster
             verifyDataOnStandby(0);
 
+            // Open a local table on Standby Cluster
```
annym, line 365: nice simple test!

annym (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/utils/LogReplicationStreamNameTableManager.java
```diff
@@ -142,7 +156,7 @@ private boolean tableVersionMatchesPlugin() {
             txn.commit();
         }
 
-        return record == null ? false : (Objects.equals(record.getPayload().getVersion(),
+        return record.getPayload() != null && (Objects.equals(record.getPayload().getVersion(),
```
pankti-m, line 159: dont we need the null check here?

pankti-m (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/utils/LogReplicationStreamNameTableManager.java
zhangn49: getRecord API had a change recently that it won't return null type. Instead, it return an empty record object with null payload. So I changed it to check the payload.

PR #3106, opened 2021-11-03
Cache Optimization

annym:
## Overview

Description:
     Cache improvement to reduce NoRollBackExceptions.

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/Table.java
@@ -145,7 +145,7 @@ public class Table<K extends Message, V extends Message, M extends Message> {
     protected void put(@Nonnull final K key,
                           @Nonnull final V value,
                           @Nullable final M metadata) {
-        corfuTable.insert(key, new CorfuRecord<>(value, metadata));
+        corfuTable.put(key, new CorfuRecord<>(value, metadata));
     }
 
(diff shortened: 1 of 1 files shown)


hisundar (approved):
LGTM

zhangn49 (approved):

pankti-m (approved):
Can you explain how this change will prevent NRE?

Maithem:
I suspect what's happening is that the optimistic stream has entries that were added by the insert but are never applied to the VLO, because it was never synced. In this case, the VLO will try to undo the optimistic stream and fail because it cant find the undo records: there's nothing to undo, because it was never applied.

The proper fix is to ignore SMR entries from the optimistic stream that were never applied to the VLO. This change will only hide the bug in the VLO and regress the performance improvements that were introduced by the pure mutators.

apavan:
> I suspect what's happening is that the optimistic stream has entries that were added by the insert but are never applied to the VLO, because it was never synced. In this case, the VLO will try to undo the optimistic stream and fail because it cant find the undo records: there's nothing to undo, because it was never applied.
> 
> The proper fix is to ignore SMR entries from the optimistic stream that were never applied to the VLO. This change will only hide the bug in the VLO and regress the performance improvements that were introduced by the pure mutators.

Given this is a blocker for the release, we will push this CL and then get the proper fix out as a follow up.

PR #3109, opened 2021-11-08
Bring UFO Write Optimization back preventing NRE

annym:
## Overview

Description:

Why should this be merged: 

Bring back UFO write path optimization by making writers write around the cache whenever data is already present, this will prevent unnecessary NoRollBackExceptions under certain patterns, as the one described next:

Consider the following example. Take the case of 3 different threads accessing the same table:
- **Thread1**: writer thread, accesses the VLO at version 10, writes a new update to address 11. 
- **Thread2**: accesses the VLO at version 11 (right after the token has been assigned, though Thread1 has still not completed the write) 
- **Thread3**: accesses VLO at version 5 (old version). 

Consider the following sequence:
- Thread1 acquires the token to write and gets token 11, starts writing (still not completed)
- Thread2 starts a read at version 11, fails to find in cache (as it is written by the writer only when write is completed), so goes fetch the data. 
- Thread2 wins and gets the data in the cache, applies the data under the lock and generates undoRecord. 
- Thread1 returns and overwrites the cached value (now with no undoRecord and this won't be generated as upcallResult will not be executed)
- Thread3 tries to rollback but fails to do so because 11 does not have undoRecord ---> Unnecessary NRE (but this can be prevented if writer writes around the cache). 

- [x] Public API has Javadoc


runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -256,9 +256,41 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
...
+
+                // Consider the following sequence:
+
+                // Thread#1 acquires the token to write and gets token X+1, starts writing (still not completed)
+                // Thread#2 starts a read at version X+1, fails to find in local cache, so goes fetch the data.
+                // Thread#2 wins and gets the data in the cache, applies the data under the lock and generates undoRecord.
+                // Thread#1 returns and overwrites the cached value (if cache WRITE_THROUGH enabled)
+                // (now with no undoRecord and this won't be generated as upcall will not be executed for mutations)
+                // Thread#3 tries to rollback but fails to do so because X+1 (overwritten value) does not have undoRecord
+                // ---> Unnecessary NRE (but this can be prevented if writer writes around the cache whenever present).
```
pankti-m, line 280: what do we mean by 'around the cache' here?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
```diff
@@ -256,9 +256,41 @@ public void write(@Nonnull IToken token, @Nonnull Object data, @Nonnull CacheOpt
...
+                try {
+                    readCache.get(token.getSequence(), () -> ld);
```
pankti-m, line 282: this will allow thread1 in the example to write to the cache only if thread2 has not written to it, correct?

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: if you see the cache_option has 2 modes: WRITE_THROUGH and WRITE_AROUND, it simply means that you don't write in the cache, you go "around"/skip it.

runtime/src/main/java/org/corfudb/runtime/view/AddressSpaceView.java
annym: correct! In that sense, you can still serve local writes if you don't have a reader thread.

pankti-m (approved):

zfrenette (approved):

PR #3111, opened 2021-11-08
Fix recovery reconfiguration step in mgmt agent

PavelZaytsev:
## Overview

In the previous version of the code, during a starup of the ManagementServer, the recovery task was triggered. 
Its goal was to update the current ManagementServer layout with a new epoch, where it's reflected that the node has restarted. 

This code assumes that if that recovery task fails, the FailureDetectors on the other nodes will catch up this node to a new layout . This is not the case when the current node is in the unresponsive list and the layout is stale. In that case, the layout will never be accepted by the majority due to the stale epoch, and since the healing workflow (move node out of the unresponsive list) can only be executed on the current node, the recovery loop will run forever. The node will never be able to rejoin the cluster and get healed.

The alternative solution is to attempt the cluster recovery with a configurable number of retries on startup (not to block the FailureDetector of this node), and then let the failure detector finish the job if needed (heal the node or do nothing).


xnull (approved):

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
```diff
@@ -465,7 +465,7 @@ public void reconfigureSequencerServers(Layout originalLayout, Layout newLayout,
      */
     public CompletableFuture<Boolean> asyncSequencerBootstrap(
             @NonNull Layout layout, @NonNull ExecutorService service) {
-
+        log.debug("Bootstrapping with layout: " + layout);
```
chetangudisagar, line 468: Could we change this to info?  Could the layout information be useful for bug fixing? If it is already being logged elsewhere, then we can leave it as is.

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
```diff
@@ -146,10 +140,37 @@
         this.initializationTaskThread.start();
     }
 
+    private void tryImmediateRecovery() {
+        log.info("Attempting immediate recovery");
+        for(int i = 0; i < 3; i++) {
```
chetangudisagar, line 145: Since the layout response and recovery can be sometimes late due to miscellaneous issues like infra, Can we retry more times (say 5 or 10)?

chetangudisagar (approved):
LGTM, a couple of minor queries.

runtime/src/main/java/org/corfudb/runtime/view/LayoutManagementView.java
PavelZaytsev: It's important to debug the sequencer bootstrap issues on FD startup, this line tells which layout the sequencer will be boostrapped with, which is usefull

infrastructure/src/main/java/org/corfudb/infrastructure/ManagementAgent.java
PavelZaytsev: Usually 3 attempts for somethingn like attemptConsensus workflow is fine. We have some wait between them too, so I think it should be ok

PR #3114, opened 2021-11-10
Persist EPOCH in LogEntry

annym:
## Overview

Description:  'Epoch' is one of several fields captured by the ILogData metadata map.
We have a bug where the epoch (metadata) is never persisted as part of LogEntry.
Therefore, entries read from disk (no longer available in the server's cache) do not have the epoch and are
returned to client's as epoch -1 (for instance in the case of StreamListeners).

Why should this be merged: bug fix, incorrect epoch returned on client's callbacks. Several clients rely on the timestamp to determine if it's higher or not. 

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
@@ -478,6 +478,7 @@ public class StreamLogFiles implements StreamLog {
 
         logData.setBackpointerMap(getUUIDLongMap(entry.getBackpointersMap()));
         logData.setGlobalAddress(entry.getGlobalAddress());
+        logData.setEpoch(entry.getEpoch());
 
         if (entry.hasThreadId()) {
             logData.setThreadId(entry.getThreadId());
@@ -879,6 +880,7 @@ public class StreamLogFiles implements StreamLog {
                 .setCodecType(entry.getPayloadCodecType().getId())
                 .setData(ByteString.copyFrom(data))
                 .setGlobalAddress(address)
+                .setEpoch(entry.getEpoch())
                 .addAllStreams(getStrUUID(entry.getStreams()))
                 .putAllBackpointers(getStrLongMap(entry.getBackpointerMap()));
 
--- a/infrastructure/proto/log_format.proto
+++ b/infrastructure/proto/log_format.proto
@@ -46,6 +46,7 @@ message LogEntry {
     // less than 4-bytes to encode, since the codecType
     // range is positive and fits in a byte
     optional int32 codecType = 19;
+    optional int64 epoch = 20;
 }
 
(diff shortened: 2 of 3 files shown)


xnull (approved):

PavelZaytsev (approved):

chetangudisagar (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/log/StreamLogFiles.java
```diff
@@ -879,6 +880,7 @@ private LogEntry getLogEntry(long address, LogData entry) {
                 .setCodecType(entry.getPayloadCodecType().getId())
                 .setData(ByteString.copyFrom(data))
                 .setGlobalAddress(address)
+                .setEpoch(entry.getEpoch())
```
Lujie1996, line 883: I saw getEpoch() is used by StreamingTask::transform. Say there is a layout change and the log entry is written using the previous epoch, will this be misleading to clients that they might think this data is obsolete?

Lujie1996 (commented):
I feel like layout and epoch information should be transparent to log data. If there is a layout change, then after the change when reading a log entry it doesn't matter if the entry is in previous epoch or not because they are all valid data.

pankti-m (approved):

shama358 (approved):

hisundar (commented):
In general I feel epoch data should not make it into the log data since it refers to the cluster layout version and something that is not intrinsic to the data being stored. On a practical note we also might face difficulties in extracting log files from one setup and applying it on another cluster because epoch now makes it into the log metadata.
Is there another way we can solve the bug?

annym:
> In general I feel epoch data should not make it into the log data since it refers to the cluster layout version and something that is not intrinsic to the data being stored. On a practical note we also might face difficulties in extracting log files from one setup and applying it on another cluster because epoch now makes it into the log metadata.
> Is there another way we can solve the bug?

Just to keep history of the offline discussion for future reference. 

We discussed other ways to solve this bug, among which, solutions like enforcing -1 as epoch, or carrying current layout epoch on streaming updates were some of them. However, after discussing with Medhavi as well, a point was made that monotonicity is guaranteed during an epoch… after that, there is no guarantee of monotonicity..for instance, consider sequence numbers 1-100 issued on epoch 1… but only materialized up to 98. Assume, sequence numbers 99-100 are not able to go through (or maybe partially). Still, someone out there in the world has Ts(1,99) and (1,100). We lose information if we strip the epoch information. A timestamp should be an immutable object, and it should preserve the sequence + epoch at which it was valid. Technically we should at a certain point, allow things like snapshot at TS (1, 100)… which is different than the state of the DB at TS (2, 100).

hisundar (approved):
Understood the explanation - there is a lot more to be done for epoch awareness but this can be a step in that direction

PR #3112, opened 2021-11-08
Failure detector read only filesystem - read only partition. Part 3

xnull:
## Overview

Description:
Detect FileSystem failures

Why should this be merged: 

Related issue(s) (if applicable): #<number>

## Checklist (Definition of Done):

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x] Change is covered by automated tests
- [x] Public API has Javadoc

PR #3121, opened 2021-11-17
Failure detector read only filesystem 4: PartitionAttribute

chetangudisagar:
## Overview

Description: This PR adds a scheduler to the `PartitionAgent` to set values for `PartitionAttribute` fields - `readOnly`, `availableSpace`, `totalSpace`. These values are then to be used by the Failure Detector and other components like metrics. The values are reset every 5 seconds using a scheduler.

Why should this be merged: Sets values to the `PartitionAttribute` fields every 5 seconds.

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
@@ -3,6 +3,7 @@ package org.corfudb.infrastructure.log;
 import com.google.common.annotations.VisibleForTesting;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.corfudb.infrastructure.ResourceQuota;
 import org.corfudb.infrastructure.ServerContext;
@@ -21,8 +22,12 @@ import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributeView;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 @Slf4j
 public final class FileSystemAgent {
@@ -32,10 +37,9 @@ public final class FileSystemAgent {
 
     private final FileSystemConfig config;
     // Resource quota to track the log size
-    @Getter
+
     private final ResourceQuota logSizeQuota;
 
-    @Getter
     private final PartitionAttribute partitionAttribute;
 
     private FileSystemAgent(FileSystemConfig config) {
@@ -68,7 +72,7 @@ public final class FileSystemAgent {
         return instance.orElseThrow(err).logSizeQuota;
     }
 
-    public static PartitionAttribute getPartition() {
+    public static PartitionAttribute getPartitionAttribute() {
         Supplier<IllegalStateException> err = () -> new IllegalStateException(NOT_CONFIGURED_ERR_MSG);
         return instance.orElseThrow(err).partitionAttribute;
     }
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
```diff
@@ -155,35 +163,43 @@ private void checkLimits() {
...
             logPartition = Paths.get(config.logDir.getRoot().toString(), config.logDir.subpath(0, 1).toString());
-            partitionAttribute = setPartitionAttribute();
-            //we need a scheduler
+            initializeScheduler();
+        }
+
+        private void initializeScheduler(){
+            scheduler.scheduleWithFixedDelay(
+                    this::setPartitionAttribute, NO_DELAY, UPDATE_INTERVAL, SECONDS
+            );
```
Maithem, line 186: I don't see why this is needed.

Can you please answer this question https://github.com/CorfuDB/CorfuDB/pull/3112#discussion_r768248608

Maithem:
Please add a git-message.

infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
chetangudisagar: After discussing with @xnull , some of the reasons are -
1. To avoid any additional latency of generating the PartitionAttributes during Failure Detection. It will further get cascaded into other processes that depend on FD.
2. To avoid duplicate calculations by other components than FD, like metrics (to be added in the future). These will also put additional overhead on disk IO.

infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
```diff
@@ -154,39 +162,80 @@ private void checkLimits() {
...
+
+        // Interval when the PartitionAttribute values are reset by the scheduler
+        private static final int UPDATE_INTERVAL = 5;
+
+        // We don't need any delay for the scheduler
+        private static final int NO_DELAY = 0;
+
+        // This contains the attribute values of the log partition
         @Getter
         private PartitionAttribute partitionAttribute;
```
zfrenette, line 182: Does this need to be marked as volatile since the value is mutated from a separate thread within the scheduler?

xnull (approved):
Looks good

infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
chetangudisagar: Good catch, Done, Thank you.

PavelZaytsev (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
```diff
@@ -68,7 +72,7 @@ public static ResourceQuota getResourceQuota() {
         return instance.orElseThrow(err).logSizeQuota;
     }
 
-    public static PartitionAttribute getPartition() {
+    public static PartitionAttribute getPartitionAttribute() {
         Supplier<IllegalStateException> err = () -> new IllegalStateException(NOT_CONFIGURED_ERR_MSG);
         return instance.orElseThrow(err).partitionAttribute;
```
zfrenette, line 77: This can return null since the `FileSystemAgent` constructor can call `getPartitionAttribute()` before `PartitionAgent` has a chance to set the value in the other thread. This can cause an NPE in `RemoteMonitoringService`.

xnull: fixed

infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
```diff
@@ -156,42 +148,86 @@ private void checkLimits() {
...
                         fileStore.getUsableSpace(),
                         fileStore.getTotalSpace()
                 );
+                log.info("setPartitionAttribute: fetched PartitionAttribute successfully. " +
+                        "{}", partitionAttribute);
             } catch (IOException e) {
-                log.error("setPartitionStats: Error while getting PartitionStats", e);
+                log.error("setPartitionAttribute: Error while fetching PartitionAttributes. " +
+                        "Reinitializing the scheduler.", e);
+                initializeScheduler();
```
zfrenette, line 219: I don't think this is needed; Since the exception is caught, the scheduler will still continue to execute this method periodically.

PR #3126, opened 2021-11-20
Streaming Poller Trim Mark Fix

annym:
## Overview

Description:

- Remove trimMark precondition validation on streaming scheduler based on current prefixTrim protocol
(refer to DeltaStream refresh method for more details).
- Streaming Scheduler should be able to catch exceptions and propagate to tasks to allow the system to
recover and not silently die.

Why should this be merged: streaming stops in the event of a sequencer change, when the sequencer node has been bootstrapped with bitmaps of an untrimmed Log Unit, hence, there is an apparent regression in the trim mark. 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
@@ -155,9 +155,19 @@ public class DeltaStream {
         }
 
         if (Address.isAddress(streamAddressSpace.getTrimMark())) {
-            Preconditions.checkState(trimMark.get() <= streamAddressSpace.getTrimMark(),
-                    "streamId=%s new trimMark %s has to be >= %s", streamId,
-                    streamAddressSpace.getTrimMark(), trimMark);
+            if (streamAddressSpace.getTrimMark() < trimMark.get()) {
+                // As per Corfu's current design, this check can only be an error message (instead of an actual
+                // validation leading to an exception).
+                // Current 'prefix trim' protocol allows the system to be in a state such that the start address of all
+                // nodes is not the same, i.e., whenever prefixTrim has not gone through all LUs (for instance
+                // consider the case where a TimeoutException hits on the last LU during prefixTrim). If this
+                // LU becomes the new sequencer during this time, its trim mark is expected to regress wrt the other
+                // LUs.
+                // For this error message to become an actual validation it needs to be backed-up by the protocol.
+                log.warn("StreamId={} new trimMark {} should be >= {}, unless trim mark regressed due to " +
+                                "NEW sequencer bootstrapped from untrimmed log unit.", streamId,
+                        streamAddressSpace.getTrimMark(), trimMark);
+            }
 
             // Adopt new trim mark
             trimMark.set(streamAddressSpace.getTrimMark());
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
```diff
@@ -316,32 +321,52 @@ private void handleFailures(List<StreamingTask> failedTasks) {
...
+            // handle tasks that have failed (i.e., remove them and propagate the errors to the listeners)
+            handleFailures(failedTasks);
+
+            long elapsedTimeNs = System.nanoTime() - cycleStartTimeNs;
+            MicroMeterUtils.time(Duration.ofNanos(elapsedTimeNs), "schedule.cycle");
+            long nextSchedule = Math.max(0, pollPeriod.toNanos() - elapsedTimeNs);
+            Preconditions.checkState(nextSchedule <= pollPeriod.toNanos());
+
+            scheduler.schedule(this::tick, nextSchedule, TimeUnit.NANOSECONDS);
+        } catch (InterruptedException ie) {
```
hisundar, line 359: We already have this exception handled in line 129 (tick() method) - should one of these be removed?

hisundar (approved):
LGTM please put the tick in the finally as we discussed in the meeting, thanks Anny,

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
```diff
@@ -155,9 +155,18 @@ public void refresh(StreamAddressSpace streamAddressSpace) {
...
+            if (streamAddressSpace.getTrimMark() < trimMark.get()) {
+                // As per Corfu's current design, this check can only be an error message (instead of an actual
+                // validation leading to an exception).
+                // Current 'prefix trim' protocol allows the system to be in a state such that the start address of all
+                // nodes is not the same, i.e., whenever prefixTrim has not gone through all LUs (for instance
+                // consider the case where a TimeoutException hits on the last LU during prefixTrim). If this
+                // LU becomes the new sequencer during this time, its trim mark is expected to regress wrt the other
+                // LUs.
+                // For this error message to become an actual validation it needs to be backed-up by the protocol.
+                log.error("StreamId={} new trimMark {} has to be >= {}", streamId,
```
pankti-m, line 167: since this condition is expected, should it be logged as debug/info instead of error?

pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/collections/streaming/DeltaStream.java
annym: mmm to be precise I wouldn't say it's "expected"... in the sense that, there is a code path in which there could be a valid  sequencer regression (the one exposed today, where the new sequencer is bootstrapped from a log unit with failed trims), but there are plenty of other paths where an error could have been introduced, where an "apparent" regression shows up and it's really a bug, for instance, rebuilding bitmaps on log unit restart. So you would need to look into the cause leading to this message (the regression) to see if its legit or not, that's why I'm still leaving it logged in case in the future we want to avoid trim mark regression altogether (a change in the protocol itself). That said, we can probably lower it to warning, as it does not necessarily imply an 'error' but it can't be generalized to 'info' like it's completely fine, as there is no guarantee of the source of the regression.

runtime/src/main/java/org/corfudb/runtime/collections/streaming/StreamPollingScheduler.java
Lujie1996: Maybe also set a flag `interrupted = true` here and check the flag in the beginning of finally block.

annym: Done!

PR #3129, opened 2021-12-02
Fix calculation remaining entries when no deltas are present

annym:
## Overview

Description:

- In the event of log entry sync wherein tx stream is advancing but no update
to replicated streams has ever occurred (currentTxStreamProcessed = -1) the remaining entries
is incorrectly computed, this fixes that edge case.

- On log entry sync re-negotiation base snapshot timestamp can be persisted as part of the replication status.

Why should this be merged: current bug

- [x] Public API has Javadoc

### DIFF
--- a/test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
+++ b/test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
@@ -1,8 +1,13 @@
 package org.corfudb.integration;
 
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.infrastructure.logreplication.infrastructure.plugins.DefaultClusterConfig;
 import org.corfudb.infrastructure.logreplication.infrastructure.plugins.DefaultLogReplicationConfigAdapter;
+import org.corfudb.infrastructure.logreplication.proto.LogReplicationMetadata;
 import org.corfudb.infrastructure.logreplication.proto.Sample;
+import org.corfudb.infrastructure.logreplication.replication.LogReplicationAckReader;
+import org.corfudb.infrastructure.logreplication.replication.receive.LogReplicationMetadataManager;
+import org.corfudb.infrastructure.logreplication.proto.LogReplicationMetadata.ReplicationStatusVal;
 import org.corfudb.protocols.wireprotocol.ILogData;
 import org.corfudb.runtime.CorfuRuntime;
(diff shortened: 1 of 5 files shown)


Lujie1996 (approved):
Looks good!

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
```diff
@@ -143,6 +157,111 @@ public void testActiveClusterReset() throws Exception {
...
+        Sleep.sleepUninterruptibly(Duration.ofSeconds(LogReplicationAckReader.ACKED_TS_READ_INTERVAL_SECONDS + 1));
+
+        // While the TX log is growing, the remaining entries to send can be changing during this time, as it is computed
+        // wrt. the tail of the log and this varies depending on how fast we are catching the tail of the log.
+        verifyReplicationStatus(ReplicationStatusVal.SyncType.LOG_ENTRY,
+                LogReplicationMetadata.SyncStatus.ONGOING, LogReplicationMetadata.SnapshotSyncInfo.SnapshotSyncType.DEFAULT,
+                LogReplicationMetadata.SyncStatus.COMPLETED);
+
+        stopWrites.set(true);
+
```
pankti-m, line 200: may be a good idea to verify remainingEntriesToSend == 0 here also

pankti-m (approved):

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
annym: No, if you see the comment in line 194 we can't ensure that because it is a running target, so likely you're always at least 1 behind or chasing the tail. So not being deterministic we don't know what to expect.

test/src/test/java/org/corfudb/integration/CorfuReplicationReconfigurationIT.java
```diff
@@ -143,6 +155,113 @@ public void testActiveClusterReset() throws Exception {
...
+        // (9) Stop active LR again, so server restarts from Log Entry Sync, with no actual deltas (as there is no new data)
+        // and verify remainingEntriesToSend is still '0'
+        stopActiveLogReplicator();
```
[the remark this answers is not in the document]
annym: we later block on waiting for replication to happen so it's fine to not block on this, there is no actual check to make upon startup.

PR #3132, opened 2021-12-03
Exclude 'clearing' EMPTY streams to replicate

annym:
## Overview

Description:
Clearing 'empty' streams on snapshot sync start can cause "apparent" data loss issues in
subsequent swithovers. This is because a 'clear' enforces an update to a stream that has
not been opened by upper layers. If the stream has not been opened it will not be checkpointed.
On a switchover, when attempting to read this stream (as it is intended for replication) we can
hit a TrimmedException.

Why should this be merged: current customer bug

- [x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
@@ -6,6 +6,8 @@ import org.corfudb.infrastructure.logreplication.LogReplicationConfig;
 import org.corfudb.infrastructure.logreplication.replication.receive.LogReplicationMetadataManager.LogReplicationMetadataType;
 import org.corfudb.protocols.logprotocol.OpaqueEntry;
 import org.corfudb.protocols.logprotocol.SMREntry;
+import org.corfudb.protocols.wireprotocol.StreamAddressRange;
+import org.corfudb.protocols.wireprotocol.TokenResponse;
 import org.corfudb.runtime.CorfuRuntime;
 import org.corfudb.runtime.CorfuStoreMetadata;
 import org.corfudb.runtime.LogReplication.LogReplicationEntryMetadataMsg;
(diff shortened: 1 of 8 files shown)


shama358 (approved):

shama358 (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -365,13 +349,13 @@ private void applyShadowStream(UUID streamId, long snapshot) {
...
-        clearTables();
         log.debug("Apply Shadow Streams, total={}", streamViewMap.size());
-        for (UUID uuid : streamViewMap.keySet()) {
-            applyShadowStream(uuid, snapshot);
+        for (UUID regularStreamId : streamViewMap.keySet()) {
+            applyShadowStream(regularStreamId, snapshot);
         }
         // Invalidate client cache after snapshot sync is completed, as shadow streams are no longer useful in the cache
         rt.getAddressSpaceView().invalidateClientCache();
+        replicatedStreamIds.clear();
```
pankti-m, line 358: the clear happens in the apply phase but this set is built in the transfer phase.

Like we discussed the other day, if transfer phase started on 1 node but then resumed on another, this set will not get cleaned on the first node.  Am I right?

annym: correct, added a clear on the reset method, so its initialized on a new transfer.

pankti-m (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -211,7 +147,6 @@ public void reset(long topologyId, long snapshot) {
         srcGlobalSnapshot = snapshot;
         recvSeq = 0;
         phase = Phase.TRANSFER_PHASE;
-        clearTables();
```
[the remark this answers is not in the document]
annym: Discussed offline. Leaving summary for future reference:

In this case only shadow streams were being cleared, which is really not required as shadow streams are seek"ed" to a valid point and never loaded entirely as we would run into trimmed exceptions (as this stream is not checkpointed). 

During the period of time the node recently became "standby" and replication is started, the isDataConsistent flag is set to false indicating that the state of the DB is yet no consistent, in other words, replication has not started and completed.

PR #3139, opened 2021-12-16
CorfuStore: Add support for closeTable() to remove unused tables

hisundar:
## Overview

Description:

Why should this be merged: Some clients have a small footprint and only want to consume notifications made on a table.
The would like to open a table only for processing the initial snapshot and then have a way to remove that table from memory so that it does not take up space in the heap.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
@@ -160,6 +160,16 @@ public class CorfuStore {
     }
 
     /**
+     *
+     * @param namespace namespace this table belongs to
+     * @param tableName name of table
+     * @throws java.util.NoSuchElementException thrown if table was not found.
+     */
+    public void closeTable(String namespace, String tableName) {
+        runtime.getTableRegistry().closeTable(namespace, tableName);
+    }
+
+    /**
      * Deletes a table instance. [NOT SUPPORTED.]
      *
      * @param namespace Namespace of the table.
--- a/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
+++ b/runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
@@ -548,6 +548,27 @@ public class TableRegistry {
     }
 
     /**
+     * Close a table that is already opened.
+     *
+     * @param namespace Namespace of the table.
+     * @param tableName Name of the table.
+     * @throws NoSuchElementException - if the table does not exist.
+     */
+    public void closeTable(String namespace, String tableName) {
+        String fullyQualifiedTableName = getFullyQualifiedTableName(namespace, tableName);
+        Table<Message, Message, Message> table = tableMap.get(fullyQualifiedTableName);
+        if (table == null) {
+            throw new NoSuchElementException("closeTable: Did not find any table "+ fullyQualifiedTableName);
+        }
+        tableMap.remove(fullyQualifiedTableName);
+        ObjectsView.ObjectID oid = new ObjectsView.ObjectID(table.getStreamUUID(), CorfuTable.class);
+        Object tableObject = runtime.getObjectsView().getObjectCache().remove(oid);
+        if (tableObject == null) {
+            throw new NoSuchElementException("closeTable: No object cache entry for "+ fullyQualifiedTableName);
+        }
+    }
+
+    /**
      * Only clears a table, DOES not delete its file descriptors from metadata.
      * This is because if the purpose of the delete is to upgrade from an old schema to a new schema
      * Then we must first purge all SMR entries of the current (old) format from corfu stream.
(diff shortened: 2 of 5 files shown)


PavelZaytsev (approved):

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -547,6 +547,26 @@ public static UUID getStreamIdForStreamTag(String namespace, String streamTag) {
...
+     * @param namespace Namespace of the table.
+     * @param tableName Name of the table.
+     * @return true if a table was found and removed from cache, false otherwise
+     */
+    public boolean closeTable(String namespace, String tableName) {
+        String fullyQualifiedTableName = getFullyQualifiedTableName(namespace, tableName);
+        Table<Message, Message, Message> table = tableMap.get(fullyQualifiedTableName);
+        if (table == null) {
+            log.warn("closeTable: Did not find any table {}", fullyQualifiedTableName);
+            return false;
```
pankti-m, line 563: any reason why we need a return value for this function?  If the table was not opened, it will be a no-op.  Is it expected that clients will use the return value for something?

shama358:
I may be wrong, but even after closeTable(), won't the table continue occupying heap through VLO and the AddressMapStreamView?

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
hisundar: Yes, in case they come back and ask us for an api which tells us if a table was removed from cache successfully. This return value just avoids us having to do another change just for that ask.

hisundar:
> I may be wrong, but even after closeTable(), won't the table continue occupying heap through VLO and the AddressMapStreamView?

Yes, but temporarily. It will occupy space in the AddressSpace but that cache has eviction both based on time and size, which means after 120seconds or all the entries in the address space would get purged. This change takes care of the other place which is the ObjectView cache.

runtime/src/main/java/org/corfudb/runtime/collections/CorfuStore.java
```diff
@@ -159,6 +159,16 @@ private Timestamp getTimestamp() {
                 Queue.CorfuGuidMsg.class, vClass, Queue.CorfuQueueMetadataMsg.class, tableOptions);
     }
 
+    /**
+     *
```
annym, line 163: Add missing description, specifying what closing a table implies...

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
```diff
@@ -547,6 +547,26 @@ public static UUID getStreamIdForStreamTag(String namespace, String streamTag) {
         return (Table<K, V, M>) tableMap.get(fullyQualifiedTableName);
     }
 
+    /**
+     * Close a table that is already opened.
+     * Throws a NoSuchElementException if table is not previously opened and not present in cache.
```
annym: this description doesn't seem consistent with behavior..

shama358 (approved):

runtime/src/main/java/org/corfudb/runtime/view/TableRegistry.java
pankti-m: mmm... do you think it is likely that there will be such as ask?  My suggestion is we throw an exception whenever the table cannot be closed(not found or another runtime error) instead of a return value.

annym: In my opinion I don't think we should throw an exception. If you think of it, the 'closeTable' aims to be an optimization, so if the table is not even opened it should just be a no-op, i.e., there is nothing to clear for optimized performance.

test/src/test/java/org/corfudb/runtime/collections/CorfuStoreShimTest.java
```diff
@@ -150,6 +150,52 @@ public void checkDirtyReads() throws Exception {
...
+
+            ManagedTxnContext txn = shimStore.tx(someNamespace);
+            txn.putRecord(tableName, key,
+                    ManagedMetadata.newBuilder().setCreateUser("abc").build(),
+                    user_1);
+            txn.commit();
+        }
+
+        assertThat(shimStore.closeTable(someNamespace, tableName)).isTrue();
+    }
```
annym, line 197: we should manually check that it is effectively not in the objectCache. Relying on the return value of our own method is not sufficient for validation.

annym (commented):
Just curious, would invalidating the address space view cache be something they would also be interested in? (perhaps not always but in some cases?)

hisundar:
> Just curious, would invalidating the address space view cache be something they would also be interested in? (perhaps not always but in some cases?)

I think they would be interested in that too. Will find out soon :)

PR #3143, opened 2022-01-05
log replication: fix when Source misses an ACK for logEntry from Sink

shama358:
## Overview

**Please review the new PR: https://github.com/CorfuDB/CorfuDB/pull/3174**
This PR is just for reference

> Scenario :
> 
> Active is in log_entry sync and is sending a batch of log-entries (a.k.a msg) to standby. Let's say active has sent and received ACKs until D10, and has sent D11 to standby.
> Then there is a leadership change on Standby, and the old standby leader notifies the active of the same.
> Active stops the replication and finds the new standby leader and sends a negotiation msg.
> Active has a pendingMsgQueue that keeps track of outstanding msgs that are resent until an ACK is received for it. So active resends D11 which is now routed to the new standby leader (since active has found the new standby leader)
> Now, there is a possibility that active receives the negotiation response before standby can processes D11 and send an ACK.
> And after receiving the negotiation msg, when active resets all relevant in-memory states (including the pendingMsgQueue) before starting the replication, active looses the information that D11 had been sent and an ACK was received for the same.
> Now the metadata on active and standby do not match..essentially the prevTs on active (D10) and the lastProcessedTs on standby (D11), which can lead to msgs being ignored by standby.
> The root cause here is that the active continues to send msgs to standby even after it gives a signal to stop replication. Solution is to clear the PendingMsgQueue when active receives a signal to stop replication.
> Another issue is that the lastProcessedTs on the standby is updated even when the msg is ignored. And it is this lastProcessedTs that is sent in ACK. Result of this is that active thinks data is being replicated while in reality, Standby is ignoring the msgs.
> Changing this and updating lastProcessedTs only when the msg is processed.
> If in case there is a scenario where the msg gets continuously ignored for some reason, the lastProcessed will not increase and the active would know about it. This will then be reflected in the "remainingEntries to be replicated", which would prevent a switchover thereby preventing data loss.
> 

Why should this be merged: 
This change handles the above scenario.

- [x ] [Coding conventions](https://github.com/CorfuDB/CorfuDB/wiki/Corfu-Style-Guidelines) (e.g. for logging, unit tests) have been followed
- [x ] Change is covered by automated tests
- [ x] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
@@ -84,8 +84,7 @@ public class LogEntrySinkBufferManager extends SinkBufferManager {
             LogReplicationEntryMetadataMsg metadata = entry.getMetadata();
             if (metadata.getTimestamp() <= lastProcessedSeq) {
                 buffer.remove(metadata.getPreviousTimestamp());
-            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq) {
-                sinkManager.processMessage(entry);
+            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq && sinkManager.processMessage(entry)) {
                 ackCnt++;
                 buffer.remove(lastProcessedSeq);
                 lastProcessedSeq = getCurrentSeq(entry);
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
@@ -77,12 +77,11 @@ public class LogEntryWriter {
             long topologyConfigId = txMessage.getMetadata().getTopologyConfigID();
             long baseSnapshotTs = txMessage.getMetadata().getSnapshotTimestamp();
             long entryTs = txMessage.getMetadata().getTimestamp();
-            long prevTs = txMessage.getMetadata().getPreviousTimestamp();
 
             lastMsgTs = Math.max(persistedLogTs, lastMsgTs);
 
             if (topologyConfigId != persistedTopologyConfigId || baseSnapshotTs != persistedSnapshotStart ||
-                    baseSnapshotTs != persistedSnapshotDone || prevTs != persistedLogTs) {
+                    baseSnapshotTs != persistedSnapshotDone) {
                 log.warn("Message metadata mismatch. Skip applying message {}, persistedTopologyConfigId={}, persistedSnapshotStart={}, " +
                                 "persistedSnapshotDone={}, persistedLogTs={}", txMessage.getMetadata(), persistedTopologyConfigId,
                         persistedSnapshotStart, persistedSnapshotDone, persistedLogTs);
@@ -148,21 +147,14 @@ public class LogEntryWriter {
 
         // Skip entries that have already been processed
         if (msg.getMetadata().getTimestamp() <= lastMsgTs) {
-            log.warn("Ignore Log Entry. Received message with snapshot {} is smaller than lastMsgTs {}.",
-                    msg.getMetadata().getSnapshotTimestamp(), lastMsgTs);
+            log.warn("Ignore Log Entry. Received message with timestamp {} is smaller than lastMsgTs {}.",
+                    msg.getMetadata().getTimestamp(), lastMsgTs);
             return Address.NON_ADDRESS;
-        }
-
-        //If the entry is the expecting entry, process it and process
-        //the messages in the queue.
-        if (msg.getMetadata().getPreviousTimestamp() == lastMsgTs) {
+        } else {
+            //process the new entries, i.e., msg.timestamp > lastMsgTs
             processMsg(msg);
             return lastMsgTs;
         }
-
-        log.warn("Log entry {} was not processed, prevTs={}, lastMsgTs={}, srcGlobalSnapshot={}", msg.getMetadata().getTimestamp(),
-                msg.getMetadata().getPreviousTimestamp(), lastMsgTs, srcGlobalSnapshot);
-        return Address.NON_ADDRESS;
     }
 
     /**
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
@@ -145,16 +145,8 @@ public class LogEntryWriter {
             lastMsgTs = srcGlobalSnapshot;
         }
 
-        // Skip entries that have already been processed
-        if (msg.getMetadata().getTimestamp() <= lastMsgTs) {
-            log.warn("Ignore Log Entry. Received message with timestamp {} is smaller than lastMsgTs {}.",
-                    msg.getMetadata().getTimestamp(), lastMsgTs);
-            return Address.NON_ADDRESS;
-        } else {
-            //process the new entries, i.e., msg.timestamp > lastMsgTs
-            processMsg(msg);
-            return lastMsgTs;
-        }
+        processMsg(msg);
+        return lastMsgTs;
     }
 
     /**
--- a/test/src/test/java/org/corfudb/integration/LogReplicationIT.java
+++ b/test/src/test/java/org/corfudb/integration/LogReplicationIT.java
@@ -35,6 +35,7 @@ import java.util.Observable;
 import java.util.Observer;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
@@ -43,6 +44,7 @@ import static java.lang.Thread.sleep;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.corfudb.integration.LogReplicationReaderWriterIT.ckStreamsAndTrim;
+import static org.corfudb.protocols.CorfuProtocolCommon.getUUID;
 
 /**
  * Test the core components of log replication, namely, Snapshot Sync and Log Entry Sync,
@@ -115,6 +117,8 @@ public class LogReplicationIT extends AbstractIT implements Observer {
     private HashMap<String, HashMap<Long, Long>> srcDataForVerification = new HashMap<>();
     private HashMap<String, HashMap<Long, Long>> dstDataForVerification = new HashMap<>();
 
+    LogReplicationSourceManager logReplicationSourceManager;
+
     private CorfuRuntime srcTestRuntime;
 
     private CorfuRuntime dstTestRuntime;
@@ -167,6 +171,8 @@ public class LogReplicationIT extends AbstractIT implements Observer {
     private final String t0 = TABLE_PREFIX + 0;
     private final String t1 = TABLE_PREFIX + 1;
     private final String t2 = TABLE_PREFIX + 2;
+
+    private final CountDownLatch blockUntilFSMTransition = new CountDownLatch(1);
 
     /**
      * Setup Test Environment
@@ -380,7 +386,7 @@ public class LogReplicationIT extends AbstractIT implements Observer {
      */
     @Test
     public void testSnapshotAndLogEntrySyncThroughManager() throws Exception {
-        testSnapshotSyncAndLogEntrySync(0, false);
+        testSnapshotSyncAndLogEntrySync(0, false, 0);
     }
 
     /**
@@ -741,9 +747,7 @@ public class LogReplicationIT extends AbstractIT implements Observer {
         testConfig.setDeleteOP(true);
         testConfig.setWaitOn(WAIT.ON_ACK);
 
-        HashSet<WAIT> waitHashSet = new HashSet<>();
-        waitHashSet.add(WAIT.ON_ACK);
-        startLogEntrySync(crossTables, waitHashSet, false);
+        startLogEntrySync(crossTables, WAIT.ON_ACK, false);
 
         expectedAckTimestamp = Long.MAX_VALUE;
 
@@ -979,7 +983,7 @@ public class LogReplicationIT extends AbstractIT implements Observer {
     @Test
     public void testSnapshotSyncLongDurationApply() throws Exception {
         final int numCyclesToDelayApply = 3;
-        testSnapshotSyncAndLogEntrySync(numCyclesToDelayApply, false);
+        testSnapshotSyncAndLogEntrySync(numCyclesToDelayApply, false, 0);
     }
 
     /**
@@ -988,10 +992,10 @@ public class LogReplicationIT extends AbstractIT implements Observer {
      */
     @Test
     public void testSnapshotSyncDelayedApplyResponse() throws Exception {
-        testSnapshotSyncAndLogEntrySync(0, true);
+        testSnapshotSyncAndLogEntrySync(0, true, 0);
     }
 
-    private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean delayResponse) throws Exception {
+    private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean delayResponse, int dropAcksLevel) throws Exception {
         // Setup two separate Corfu Servers: source (active) and destination (standby)
         setupEnv();
 
@@ -1013,6 +1017,7 @@ public class LogReplicationIT extends AbstractIT implements Observer {
         expectedAckTimestamp = Long.MAX_VALUE;
         testConfig.setDelayedApplyCycles(numCyclesToDelayApply);
         testConfig.setTimeoutMetadataResponse(delayResponse);
+        testConfig.setDropAckLevel(dropAcksLevel);
 
         // Start Snapshot Sync (through Source Manager)
         Set<WAIT> conditions = new HashSet<>();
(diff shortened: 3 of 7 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
```diff
@@ -77,12 +77,11 @@ private void processMsg(LogReplicationEntryMsg txMessage) {
             long topologyConfigId = txMessage.getMetadata().getTopologyConfigID();
             long baseSnapshotTs = txMessage.getMetadata().getSnapshotTimestamp();
             long entryTs = txMessage.getMetadata().getTimestamp();
-            long prevTs = txMessage.getMetadata().getPreviousTimestamp();
 
             lastMsgTs = Math.max(persistedLogTs, lastMsgTs);
 
             if (topologyConfigId != persistedTopologyConfigId || baseSnapshotTs != persistedSnapshotStart ||
-                    baseSnapshotTs != persistedSnapshotDone || prevTs != persistedLogTs) {
```
pankti-m, line 85: removing this means we may process out-of-order messages.  Is that right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
```diff
@@ -148,21 +147,14 @@ public long apply(LogReplicationEntryMsg msg) throws ReplicationWriterException
...
-            log.warn("Ignore Log Entry. Received message with snapshot {} is smaller than lastMsgTs {}.",
-                    msg.getMetadata().getSnapshotTimestamp(), lastMsgTs);
+            log.warn("Ignore Log Entry. Received message with timestamp {} is smaller than lastMsgTs {}.",
+                    msg.getMetadata().getTimestamp(), lastMsgTs);
             return Address.NON_ADDRESS;
-        }
-
-        //If the entry is the expecting entry, process it and process
-        //the messages in the queue.
-        if (msg.getMetadata().getPreviousTimestamp() == lastMsgTs) {
```
pankti-m, line 158: same here.. can you explain why this check was removed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
shama358: Not exactly, the method that eventually calls this class filters out msgs which have lastMsgSentByActive <= lastProcessedTs and currentMsg.timestamp > lastProcessedTs.

SinkBufferManager.java:
> if (preTs <= lastProcessedSeq && currentTs > lastProcessedSeq) {
> ....
> sinkManager.processMessage(dataMessage);
> ...
> }

This leaves out msgs which are expected and would still be inorder because of `preTs <= lastProcessedSeq` . 
In scenarios where the active sends a msg,  which has log entries that have been previously applied by Standby + some new entires, the entries currently would be ignored. And so will all the subsequent msgs.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
shama358: Same explanation as above

test/src/test/java/org/corfudb/integration/LogReplicationReaderWriterIT.java
```diff
@@ -354,9 +355,41 @@ public static void writeLogEntryMsgs(List<LogReplicationEntryMsg> msgQ, Set<Stri
...
         }
+
+        // Scenario 2: Sink processes the previous msg, but Source is unware due to an interruption which causes LR to stop and start
+        // thereby resetting the Source's metadata to reflect the last ACK.Which means, the metadata on Source and Sink may not match.
+        // The source now resends the same data + new data that was written in the meanwhile, and sink filters and process only the new data.
+
+        if (msgQIterator.hasNext()) {
+            LogReplicationEntryMsg oldMsg = msg;
+            msg = msgQIterator.next();
+            LogReplication.LogReplicationEntryMetadataMsg newMetadata = LogReplication.LogReplicationEntryMetadataMsg.newBuilder()
```
Jas0nch, line 373: is there anything like clone method we can use in protoBuf?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
Jas0nch: but I believe not every processMessage call checked the same condition, for example, 
SnapshotSinkBufferManager.java:
```
public void processBuffer() {
        while (true) {
            LogReplicationEntryMsg dataMessage = buffer.get(lastProcessedSeq);
            if (dataMessage == null) {
                return;
            }
            sinkManager.processMessage(dataMessage);
            ackCnt++;
            buffer.remove(lastProcessedSeq);
            lastProcessedSeq = getCurrentSeq(dataMessage);
        }
    }
```

would it be possible if we have some out of order messages in buffer?

pankti-m: I agree there is scope to clean up/solidify  the checks in these methods.  But I have a some follow-up questions.  Let me get them clarified offline.

test/src/test/java/org/corfudb/integration/LogReplicationReaderWriterIT.java
shama358: Changed to using "mergeFrom"

pankti-m (approved):

Jas0nch (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
```diff
@@ -148,21 +147,14 @@ public long apply(LogReplicationEntryMsg msg) throws ReplicationWriterException
 
         // Skip entries that have already been processed
         if (msg.getMetadata().getTimestamp() <= lastMsgTs) {
-            log.warn("Ignore Log Entry. Received message with snapshot {} is smaller than lastMsgTs {}.",
-                    msg.getMetadata().getSnapshotTimestamp(), lastMsgTs);
+            log.warn("Ignore Log Entry. Received message with timestamp {} is smaller than lastMsgTs {}.",
```
annym, line 150: One question, in the case the ACK was missed by the active node. Let's assume it resends the exact same batch (that was already applied on standby but not acked), meaning that TS == lastMsgTs. We would continue to ignore the message and never come out, isn't it?

test/src/test/java/org/corfudb/integration/LogReplicationReaderWriterIT.java
```diff
@@ -354,9 +355,40 @@ public static void writeLogEntryMsgs(List<LogReplicationEntryMsg> msgQ, Set<Stri
             log.debug("msgQ is EMPTY");
         }
 
-        for (LogReplicationEntryMsg msg : msgQ) {
+        Iterator<LogReplicationEntryMsg> msgQIterator = msgQ.iterator();
```
annym, line 358: I think we should have an E2E test, where we actually intersect the ACK and drop it. We don't ACK on a 1:1 basis but we have aggregated ACKs so I'm wondering if this really replicates the scenario we're trying to target.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
shama358: Yes, We would ignore the msg, but still send an ACK for it with the "lastProcessedTs" metadata which is populated to what was last processed. So the log replication should move ahead even if standby receives a msg repeatedly in-order/out-of-order.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1046,6 +1051,145 @@ private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean
...
+        // Write data in transaction to t0 and t1
+        Set<String> crossTables = new HashSet<>();
+        crossTables.add(t0);
+        crossTables.add(t1);
+
+        // Writes transactions to t0, t1 and t2 + transactions across 'crossTables'
+        writeCrossTableTransactions(crossTables, true);
+
+        // Start Log Entry Sync
+        expectedAckMessages = NUM_KEYS * WRITE_CYCLES;
```
Jas0nch, line 1073: if miss one ack, should this expectedAckMessages -1?

shama358: No. Active keeps msgs in a queue and resends them until ACKs are received (Msgs are evicted from that queue when ack.timestamp >= msg.timestamp).

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
```diff
@@ -84,8 +84,7 @@ public void processBuffer() {
             LogReplicationEntryMetadataMsg metadata = entry.getMetadata();
             if (metadata.getTimestamp() <= lastProcessedSeq) {
                 buffer.remove(metadata.getPreviousTimestamp());
-            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq) {
-                sinkManager.processMessage(entry);
+            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq && sinkManager.processMessage(entry)) {
```
pankti-m, line 87: Question: SinkBufferManager.processMsgAndBuffer() also calls SinkManager.processMessage() just before invoking this method.  SinkManager.processMessage() is again called here.  What is the reason?

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1046,6 +1051,145 @@ private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean
...
+        srcDataForVerification.get(t2).clear();
+
+        // Verify Destination
+        verifyData(dstCorfuTables, srcDataForVerification);
+    }
+
+    /**
+    * Test Log Entry (delta) Sync for the case where an ACK gets dropped at the Source, and
+    * immediately, the Source gets STOP event followed by log_entry Sync event. This tests that even
+    * though the metadata on Source and destination do not match, the sink is able to process the ms
```
pankti-m, line 1125: nit: ms->msg

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1046,6 +1051,145 @@ private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean
...
 
+     /**
+     * Test Log Entry (delta) Sync for the case where the ACKs are arbitrarily dropped
+     * for a fixed number of times at the Source. This will test that LR is
+     * (i) not impacted by dropped ACKs
+     * (ii) Source resends msgs for which it hasn't received the ACKs
+     * (iii) Sink handles the already seen and processed msgs.
+     **/
+    @Test
+    public void testLogEntrySyncWithAckDrops() throws Exception {
```
annym, line 1062: I see this and testLogEntrySyncWithFSMChangeAndWithAckDrop failing on this PR, there might be something that is off (perhaps in the test itself).

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1046,6 +1051,145 @@ private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean
...
+
+
+     /**
+     * Test Log Entry (delta) Sync for the case where messages are arbitrarily dropped at the
+     * destination and ACKs are arbitrarily dropped at the source. This tests
+     * (i) messages dropped at the destination is resent. This would also test that Sink handles out of order messages.
+     * (ii) messages for which ACKs were dropped at the source are resent and Sink handles already processed data.
+     */
+    @Test
+    public void testLogEntrySyncWithMsgDropsAndAckDrops() throws Exception {
```
annym, line 1095: I tried running the tests on current master and all new tests pass except testLogEntrySyncWithFSMChangeAndWithAckDrop, I was expecting that testLogEntrySyncWithAckDrops would fail. For the test to be correct, shouldn't we be writing additional data after the ACK drop so we are sure that it continues to progress? cause as the test is right now its only validating the data on standby and the ack was lost but the data was persisted.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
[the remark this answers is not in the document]
shama358: 1. I think the order should not be impacted since we are checking that the current msg has a previousTimestamp which was processed by Standby + new data...So the order would be maintained by the first part of the check and that would make the current msg to be in the expected sphere..
2. If we check for only equality, we would miss msgs when active's metadata is behind what standby has processed. (Like the example in the desc).

> Example: 
> Active: [1,4,5,7]               ->                 Standby [1,4,5,7] Replicated.
> **_previousTimestamp = -1.                       lastProcessedTimestamp = 7_**
> 
> Active: [8,10,17].            ->.                 Standby[8,10,17] Replicated
> (ACK miss, LR stops                             _**lastProcessedTimestamp = 17**_
> so active remains in the 
> previous state)
> **_previousTimestamp = -1_**
> 
> Acive: [8, 10, 17, 22]      ->      Standby should process this msg since "22" is a new entry, and order would be correct

3. In the processBuffer(), we remove the entries which are older than the lastProcessedTimestamp. And since the sequencer will never reassign a token, even if the the msg in the buffer has an old snapshot, that should be removed, right ?

> for (LogReplicationEntryMsg entry : buffer.values()) {
>             LogReplicationEntryMetadataMsg metadata = entry.getMetadata();
>             if (metadata.getTimestamp() <= lastProcessedSeq) {
>                 buffer.remove(metadata.getPreviousTimestamp());

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
```diff
@@ -148,21 +147,13 @@ public long apply(LogReplicationEntryMsg msg) throws ReplicationWriterException
...
```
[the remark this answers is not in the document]
shama358: Agreed..I think L133 is also redundant, out-of-date msgs should not reach here at all
```
// Ignore the out of date messages
if (msg.getMetadata().getSnapshotTimestamp() < srcGlobalSnapshot) {
            log.warn("Ignore Log Entry. Received message with snapshot {} is smaller than current snapshot {}",
                    msg.getMetadata().getSnapshotTimestamp(), srcGlobalSnapshot);
            return Address.NON_ADDRESS;
        }
```

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
shama358: I can include additional data write as well, but even with the current test, there are logs that are read and replicated after ACK drops. I can maybe have an assert in between to showcase that

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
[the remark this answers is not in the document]
shama358: Oh no Sink doesn't need to send any msg if LR was stopped on source. 
I meant Active looses an ACK and immediately LR on active stops...say the state transitions on active is as such: STOP LOG_ENTRY SYNC -> INITIALIZED -> LOG_ENTRY SYNC
The pending queue on active that stores entries until it is ACK'ed is now cleared. So now, since active doesn't have any inkling that it had previously sent a msg but not received an ACK, it retains its old metadata, while standby having processed the last msg successfully, has moved ahead.
In the current code, the msg sent by Active (with old unchanged metadata) will be ignored by Standby since Standby thinks that it has already processed it. 

This is an issue since after LR is stopped and started again, there is no guarantee that the msg will have the same log_entries as previously sent (and for which ACK was lost). It can contain new log entries, which should be processed by Standby.

This test case is trying to emulate that scenario.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -988,10 +992,10 @@ public void testSnapshotSyncLongDurationApply() throws Exception {
      */
     @Test
     public void testSnapshotSyncDelayedApplyResponse() throws Exception {
-        testSnapshotSyncAndLogEntrySync(0, true);
+        testSnapshotSyncAndLogEntrySync(0, true, 0);
     }
 
-    private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean delayResponse) throws Exception {
+    private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean delayResponse, int dropAcksLevel) throws Exception {
```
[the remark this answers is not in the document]
shama358: a little hesitant on "count" part since it may infer that it corresponds to number of ACKs dropped.

This variable is to decide if ACKs are to be randomly dropped OR if msgs have to be dropped after an ACK drop to emulate the Active's part in the testLogEntrySyncWithFSMChangeAndWithAckDrop scenario

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
shama358: yeah, this one and the testSnapshotSyncWithAckDrops are taking turns to fail.
but now when I added some print statements, the tests passed!!

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
```diff
@@ -1046,6 +1051,145 @@ private void testSnapshotSyncAndLogEntrySync(int numCyclesToDelayApply, boolean
...
+        // Because t2 is not specified as a replicated table, we should not see it on the destination
+        srcDataForVerification.get(t2).clear();
+
+        // Verify Destination
+        verifyData(dstCorfuTables, srcDataForVerification);
+    }
+
+
+    @Test
+    public void testSnapshotSyncWithAckDrops() throws Exception {
```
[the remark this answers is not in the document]
shama358: Snapshot msgs are sent through pendingQueue (same as for delta sync), and msgs are removed from pendingQueue after an ACK is received for a msg or a higher ACK is received in which case msg.timestamp <= ACK.timestamp are removed from the queue.
But ACK drops for snapshot is not concerning because our snapshot does not change, which translates to the same msg being resent to Sink which can be ignored if the Sink has already processed it. Since it's still a valid scenario, I included this simple test.

test/src/test/java/org/corfudb/integration/LogReplicationReaderWriterIT.java
shama358: I realized while working on the E2E tests that we do ACK 1:1 basis once we process the msg.
 ```
if (sinkManager.processMessage(dataMessage)) {
                ackCnt++;
                ....
}
...
if (shouldAck()) {
     // send ACK
}


public boolean shouldAck() {
        long currentTime = java.lang.System.currentTimeMillis();
        ackCnt++;

        if (ackCnt >= ackCycleCnt || (currentTime - ackTime) >= ackCycleTime) {
           .....
            return true;
        }
       return false
}
```
 And "ackCycleCnt" value is 2

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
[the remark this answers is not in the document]
annym: Yes, you're right I forgot this is a batch of messages and was thinking on individual messages (which is incorrect). So, yes <= is correct. 

Well it will stay there if the new deltas (lastProcessedSeq) don't progress beyond the stale entry's timestamp (metadata.getTimestamp()), but eventually it will get removed. It should be expected that it always goes beyond as the snapshot base address can't regress, so yes agree we are fine as it will get removed at some point.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntryWriter.java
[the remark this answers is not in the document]
annym: mmm not sure, about this second one... Cause before this we only use timestamp and prevTimestamp but they could be based on an old base snapshot, right? I don't remember seeing any check based on the snapshot.

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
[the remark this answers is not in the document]
annym: you're right, I then saw the meaning and forgot to remove the comment!

test/src/test/java/org/corfudb/integration/LogReplicationIT.java
annym: Ok, so that means the amount of data we write generates more than one batch. But still, shouldn't the test fail on current master?

shama358: No, the test should fail on master because we currently have a mechanism where active keeps resending a msg to standby until it receives an ACK for the msg or receives an ACK for a msg with higher ts which would imply that standby has processed the msg. 

Issue is only when the active looses the information that it has sent a msg to standby and resends the same msg but now that msg includes new entries...like the scenario in the desc.

shama358: testLogEntrySyncWithFSMChangeAndWithAckDrop() should fail on master

PR #3153, opened 2022-01-11
Log replication in-flight streams handling

xcchang:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/LogReplicationConfig.java
```diff
@@ -91,17 +98,81 @@ public LogReplicationConfig(Set<String> streamsToReplicate, int maxNumMsgPerBatc
...
+                log.warn("streamInfoManager is null! skipping sync!");
+                return;
+            }
+
+            Set<TableInfo> registryInfo = streamInfoManager.readStreamsToReplicatedFromRegistry(timestamp);
+            refreshStreamIds(registryInfo, true);
+        }
+
+        public Set<UUID> getStreamIds() {
+            return ImmutableSet.copyOf(streamIds);
```
pankti-m, line 165: this would throw an NPE if stream ids is null

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -106,15 +104,12 @@ public StreamsSnapshotWriter(CorfuRuntime rt, LogReplicationConfig config, LogRe
     private void initializeShadowStreams(LogReplicationConfig config) {
         // For every stream create a shadow stream which name is unique based
         // on the original stream and a suffix.
-        for (String streamName : config.getStreamsToReplicate()) {
-            String shadowStreamName = streamName + SHADOW_STREAM_SUFFIX;
-            UUID streamId = CorfuRuntime.getStreamID(streamName);
+        for (UUID streamId : config.getStreamsInfo().getStreamIds()) {
+            String shadowStreamName = streamId + SHADOW_STREAM_SUFFIX;
```
pankti-m, line 108: The shadow stream name will now use the stream id instead of name.  Is it intentional?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -106,15 +104,12 @@ public StreamsSnapshotWriter(CorfuRuntime rt, LogReplicationConfig config, LogRe
...
             UUID shadowStreamId = CorfuRuntime.getStreamID(shadowStreamName);
             regularToShadowStreamId.put(streamId, shadowStreamId);
-            regularToShadowStreamId.put(shadowStreamId, streamId);
```
pankti-m, line 114: can you explain why this mapping is being removed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
```diff
@@ -124,6 +131,46 @@ private LogReplicationEntryMsg generateMessageWithOpaqueEntryList(
...
+        if (checkIdSet.isEmpty()) {
+            return;
+        }
+
+        // TODO: why not get check name set directly?
+        CorfuTable<CorfuStoreMetadata.TableName,
+                CorfuRecord<CorfuStoreMetadata.TableDescriptors, CorfuStoreMetadata.TableMetadata>>
+                registryTable = runtime.getTableRegistry().getRegistryTable();
+
+        Set<CorfuStoreMetadata.TableName> checkNameSet = registryTable.keySet().stream()
```
pankti-m, line 150: if a stream id is present in the transaction entry, it will have been opened and hence guaranteed to be in the registry table.  So this check and filter are not needed.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -240,17 +233,15 @@ public SnapshotReadMessage read(UUID syncRequestId) {
 
         if (currentStreamHasNext()) {
             msg = read(currentStreamInfo, syncRequestId);
-            if (msg != null) {
```
pankti-m, line 243: can you explain why this check was removed?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
```diff
@@ -277,19 +271,19 @@ public void reset(long ts) {
...
+        private final String name;
+        private final UUID uuid;
+        private final Iterator<OpaqueEntry> iterator;
         private long maxVersion; // the max address of the log entries processed for this stream.
 
-        OpaqueStreamIterator(String name, CorfuRuntime rt, long snapshot) {
-            this.name = name;
-            uuid = CorfuRuntime.getStreamID(name);
+        OpaqueStreamIterator(UUID uuid, CorfuRuntime rt, long snapshot) {
+            this.name = uuid.toString();
```
pankti-m, line 280: this does not give the stream name

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/utils/LogReplicationStreamNameTableManager.java
```diff
@@ -95,6 +102,55 @@ public LogReplicationStreamNameTableManager(CorfuRuntime runtime, String pluginC
         return readStreamsToReplicateFromTable();
     }
 
+    public void addStreamsToInfoTable(Set<UUID> streamIdSet) {
+        try {
+            Table<TableInfo, Namespace, CommonTypes.Uuid> streamsInfoTable = corfuStore.openTable(CORFU_SYSTEM_NAMESPACE,
```
pankti-m, line 107: wouldnt it be easier to keep the table schema as uuid, TableInfo?.  This schema seems a bit non-intuitive.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/LogReplicationConfig.java
xcchang: I thought the `streamsIds` set is guaranteed to be non-null as per its data flow, but it could be empty.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsSnapshotReader.java
xcchang: here `msg` is always non-null, are we expecting some kind of empty check?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/LogReplicationConfig.java
```diff
@@ -91,17 +98,81 @@ public LogReplicationConfig(Set<String> streamsToReplicate, int maxNumMsgPerBatc
...
+        private final LogReplicationStreamNameTableManager streamInfoManager;
+
+        public StreamInfo(@NonNull Set<TableInfo> streamsToReplicate,
+                          LogReplicationStreamNameTableManager streamInfoManager) {
+            this.streamInfoManager = streamInfoManager;
+            this.streamIds = new HashSet<>();
+            refreshStreamIds(streamsToReplicate, true);
+        }
+
+        private void refreshStreamIds(Set<TableInfo> streamsToReplicate, boolean shouldClear) {
```
Jas0nch, line 130: I saw all reference of this refreshStreamIds function has a shouldClear set to true, do we really need this flag here?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/send/logreader/StreamsLogEntryReader.java
xcchang: Thanks for pointing out I got it. The reason we create a `checkNameSet` here is we need the table name to get the table metadata from RegistryTable and further get the `is_federated` flag. And the code above only uses UUID to track streams. But yeah I agree the code here is a little redundant and we can do some improvement here.

annym (commented):
Just some comments, need to continue reviewing...

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/LogReplicationConfig.java
```diff
@@ -91,17 +98,81 @@ public LogReplicationConfig(Set<String> streamsToReplicate, int maxNumMsgPerBatc
...
+            if (shouldClear) {
+                this.streamIds.clear();
+            }
+
+            for (TableInfo info : streamsToReplicate) {
+                if (info.hasField(TableInfo.getDescriptor().findFieldByName("name"))) {
+                    this.streamIds.add(CorfuRuntime.getStreamID(info.getName()));
+                } else if (info.hasField(TableInfo.getDescriptor().findFieldByName("id"))) {
+                    this.streamIds.add(UUID.fromString(info.getId()));
```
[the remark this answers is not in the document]
xcchang: I thought you were meaning "id" is not set? in this case I thought it will return false, as protobuf 3 now are having default values. Nan previously was using `== null` here which would always return true, so I changed to `hasField` which detects whether the field is set or not

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -344,14 +326,77 @@ private void applyShadowStream(UUID streamId, long snapshot) {
...
+        });
+        // Get streams to replicate from registry table and update log replication config.
+        Set<UUID> streamsToReplicate = registryTable.entryStream()
+                .filter(entry -> entry.getValue().getMetadata().getTableOptions().getIsFederated())
+                .map(entry -> CorfuRuntime.getStreamID(getFullyQualifiedTableName(entry.getKey())))
+                .collect(Collectors.toSet());
+        // Update the dataStreamToTagsMap in LogReplicationConfig, clear the existing mapping as
+        // this is the point that dataStreamToTagsMap should be initialized in LogReplicationConfig
+        config.updateDataStreamToTagsMap(dataStreamToTagsMap, true);
+        config.getStreamsInfo().updateStreamIdsOnStandby(streamsToReplicate, true);
```
shama358, line 378: We iterate through registry table and find the streams to replicate when standby receives the snapshot_Start msg. We don't need to find them again while applying shadow-> regularStreams right?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -344,14 +326,77 @@ private void applyShadowStream(UUID streamId, long snapshot) {
...
+        config.getStreamsInfo().updateStreamIdsOnStandby(regularToShadowStreamId.keySet(), false);
```
shama358, line 379: Why do we need to separately add regularToShadowStreamId set?

PR #3169, opened 2022-02-16
[WIP] MVCC PoC

Lujie1996:
Introduced PersistentCorfuTable which is backed by Vavr HashMap.
Added PersistentCorfuCompileProxy and PersistentVersionLockedObject.
Added a simple put and get test on PersistentCorfuTable.

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


runtime/pom.xml
```diff
@@ -185,6 +185,11 @@
             <artifactId>commons-collections</artifactId>
             <version>3.2.2</version>
         </dependency>
+        <dependency>
+            <groupId>io.vavr</groupId>
```
xnull, line 189: very nice :) 👍🏼

runtime/src/main/java/org/corfudb/runtime/collections/PersistentCorfuTable.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+public class PersistentCorfuTable<K, V> implements ICorfuTable<K, V>, ICorfuSMR<PersistentCorfuTable<K, V>> {
+
+    private PersistentHashMapWrapper<K, V> mainMap;
+
+    public PersistentCorfuTable() {
```
xnull, line 17: As I know Vjeko has already implemented a PersistentCorfu or something like that

generator/src/main/java/org/corfudb/generator/operations/CheckpointOperation.java
```diff
@@ -22,7 +23,7 @@ public CheckpointOperation(State state) {
     @Override
     public void execute() {
         try {
-            MultiCheckpointWriter<CorfuTable<String, String>> mcw = new MultiCheckpointWriter<>();
+            MultiCheckpointWriter<PersistentCorfuTable<String, String>> mcw = new MultiCheckpointWriter<>();
```
Maithem, line 26: `ICorfuTable` should be used here so that both interface implementations can be tested.

runtime/src/main/java/org/corfudb/runtime/collections/ImmutableCorfuTable.java
```diff
@@ -0,0 +1,296 @@
...
+     * @param key
+     * @param value
+     * @return
+     */
+    public ImmutableCorfuTable<K, V> put(@Nonnull K key, @Nonnull V value) {
+        SecondaryIndexesWrapper<K, V> newSecondaryIndexesWrapper = secondaryIndexesWrapper;
+        if (!secondaryIndexesWrapper.isEmpty()) {
+            V prev = get(key);
+            if (prev != null) {
+                newSecondaryIndexesWrapper = newSecondaryIndexesWrapper.unmapSecondaryIndexes(key, prev);
```
Maithem, line 70: Doesn't `mapSecondaryIndexes` overwrite the mapping if it existed? It's not clear to me why an `unmapSecondaryIndexes` is needed here.

runtime/src/main/java/org/corfudb/runtime/collections/ImmutableCorfuTable.java
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     *
+     * @return
+     */
+    public Stream<java.util.Map.Entry<K, V>> entryStream() {
+        // TODO: Consider alternative to avoid conversion to Java Map.
+        return mainMap.toJavaMap().entrySet().stream();
+    }
+
```
Maithem, line 141: Instead of converting to a map (i.e, `toJavaMap`) you can wrap mainMap.toStream() that converts each entry from Tuple to Map.Entry.

Maithem: Actually, we should use `toJavaParallelStream` instead so that the `scanAndFilter` can benefit from parallelization.

runtime/src/main/java/org/corfudb/runtime/collections/PersistentCorfuTable$CORFUSMR.java
```diff
@@ -0,0 +1,110 @@
+package org.corfudb.runtime.collections;
```
Maithem, line 1: Since the new table is not being auto generated this Type can be removed.

runtime/src/main/java/org/corfudb/runtime/object/MVOCorfuCompileProxy.java
```diff
@@ -0,0 +1,222 @@
...
+        this.type = type;
+        this.args = args;
+        this.serializer = serializer;
+        this.streamTags = streamTags;
+
+        this.underlyingMVO = new MultiVersionObject<>(
+                rt,
+                this::getNewInstance,
+                new StreamViewSMRAdapter(rt, rt.getStreamsView().getUnsafe(streamID)),
+                wrapperObject, streamID);
```
Maithem, line 59: This is a memory leak. Since it is not a  VersionLockedObject the internal data structures that are populated inside the StreamView are never garbage collected via `ViewsGarbageCollector`

runtime/src/main/java/org/corfudb/runtime/object/transactions/OptimisticTransactionalContext.java
```diff
@@ -69,41 +71,49 @@ public <R, T extends ICorfuSMR<T>> R access(ICorfuSMRProxyInternal<T> proxy,
...
-                            // committed changes, or apply forward committed changes.
-                            syncWithRetryUnsafe(o, getSnapshotTimestamp(), proxy,
-                                    o::setUncommittedChanges);
-                        },
-                        accessFunction::access,
-                        version -> updateKnownStreamPosition(proxy.getStreamID(), version)
-        );
+
+        // TODO: better proxy type check, or refactor to avoid.
+        if (proxy instanceof CorfuCompileProxy) {
```
Maithem, line 82: I think this abstraction needs to be cleaned up. The problem here is that `ICorfuSMRProxyInternal` is not generic enough to capture capture both usages. I suggest removing `CorfuCompileProxy::underlyingObject` and pushing the `VersionLockedObject` logic further down.

Maithem: This would also mean making `AbstractTransactionalContext` more generic by removing `syncWithRetryUnsafe(VersionLockedObject`

runtime/src/main/java/org/corfudb/runtime/view/ObjectsView.java
```diff
@@ -209,9 +214,15 @@ public long TXEnd() throws TransactionAbortedException {
...
-                    getCorfuSMRProxy()).getUnderlyingObject().gc(trimMark);
+            if (((ICorfuSMR) obj).getCorfuSMRProxy() instanceof CorfuCompileProxy) {
+                ((CorfuCompileProxy) ((ICorfuSMR) obj).
+                        getCorfuSMRProxy()).getUnderlyingObject().gc(trimMark);
+            }
         }
+
+        mvoCache.getAllMVOs().forEach((uuid, mvo) -> {
+            ((MultiVersionObject) mvo).gc(trimMark);
+        });
```
Maithem, line 225: Ok, I see how both types are being gc'd. Maybe introduce a generic gc method in `ICorfuSMRProxy?`

runtime/src/main/java/org/corfudb/runtime/object/MVOCache.java
```diff
@@ -0,0 +1,279 @@
...
+    public void registerMVO(UUID objectId, MultiVersionObject<T> mvo) {
+        allMVOs.computeIfAbsent(objectId, key -> mvo);
+    }
+
+    private void syncMVOCache() {
+        TokenResponse streamTails = runtime.getSequencerView()
+                .query(allMVOs.keySet().toArray(new UUID[0]));
+        final AtomicInteger count = new AtomicInteger(0);
+        allMVOs.forEach((uuid, mvo) -> {
+            if (objectVersions.get(uuid) != null &&
```
Maithem, line 248: This assumes that all the streams can fit in a single message.

Maithem: Why use specific stream tails if you can just query the global tail and sync to that?

runtime/src/main/java/org/corfudb/runtime/object/MultiVersionObject.java
```diff
@@ -0,0 +1,303 @@
...
+            }
+        }
+
+        throw new StaleObjectVersionException(streamID, timestamp);
+    }
+
+    protected Optional<ICorfuSMRSnapshotProxy<T>> getVersionedObjectUnderLock(long timestamp, Consumer<Long> versionAccessed) {
+        // polling is expensive
+        while (timestamp <= currentSyncVersion && lock.tryWriteLock() == 0) {
+            Optional<ICorfuSMRSnapshotProxy<T>> snapshotProxyOptional = mvoCache.get(
```
Maithem, line 128: It's better to implement this coordination with conditional variables or a `CompletableFuture`

runtime/src/main/java/org/corfudb/runtime/object/MultiVersionObject.java
```diff
@@ -0,0 +1,303 @@
...
+                    new VersionedObjectIdentifier(streamID, timestamp), snapshotReferenceProxyGenerator);
+            if (snapshotProxyOptional.isPresent()) {
+                return snapshotProxyOptional;
+            }
+        }
+
+        long ts = 0;
+        try {
+            ts = lock.writeLock();
+            currentSyncVersion = timestamp;
```
Maithem, line 138: This looks like a race. Other threads will observe that `timestamp <= currentSyncVersion` but before the sync happens (i.e., snapshotProxyOptional.isPresent() is false).

runtime/src/main/java/org/corfudb/runtime/object/MVOCache.java
```diff
@@ -0,0 +1,279 @@
...
+ *
+ */
+@Slf4j
+public class MVOCache<T extends ICorfuSMR<T>> {
+
+    private final CorfuRuntime runtime;
+
+    // A registry to keep track of all opened MVOs
+    @Getter
+    private final ConcurrentHashMap<UUID, MultiVersionObject<T>> allMVOs = new ConcurrentHashMap<>();
```
Maithem, line 42: We should also have an API to remove objects after they are opened. Some applications want to open the table once and then release it from memory.

runtime/src/main/java/org/corfudb/runtime/object/MVOCache.java
```diff
@@ -0,0 +1,279 @@
...
+                .expireAfterAccess(DEFAULT_CACHE_EXPIRY_TIME_IN_SECONDS, TimeUnit.SECONDS)
+                .expireAfterWrite(DEFAULT_CACHE_EXPIRY_TIME_IN_SECONDS, TimeUnit.SECONDS)
+                .removalListener(this::handleEviction)
+                .recordStats()
+                .build();
+
+        mvoCacheSyncThread.scheduleAtFixedRate(this::syncMVOCache,
+                runtime.getParameters().getMvoAutoSyncPeriod().toMillis(),
+                runtime.getParameters().getMvoAutoSyncPeriod().toMillis(),
+                TimeUnit.MILLISECONDS);
```
Maithem, line 88: I think the default should be reduced and the logic should take account for syncs that were triggered by application threads. If this scheduler triggers a task for an object that was just sync'd it should't be resync'd again.

Maithem:
Great work. I think the patch needs some refactoring and explicit documentation on the concurrency model.

Maithem:
I also think we should find better naming schemes for all new classes. For example, `persistent` is an overloaded and it might be confusing to understand the context (secondary storage persistence, or in memory?)

Lujie1996:
Close this one as the same change has been merged by #3387.

PR #3174, opened 2022-02-19
Log Replication: Flush pending messages on replication stop

shama358:
1. Clear pendingMsgs queue on REPLICATION_STOP event
2. On the standby, update lastProcessedTs only when the msg is processed

## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
@@ -84,8 +84,7 @@ public class LogEntrySinkBufferManager extends SinkBufferManager {
             LogReplicationEntryMetadataMsg metadata = entry.getMetadata();
             if (metadata.getTimestamp() <= lastProcessedSeq) {
                 buffer.remove(metadata.getPreviousTimestamp());
-            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq) {
-                sinkManager.processMessage(entry);
+            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq && sinkManager.processMessage(entry)) {
                 ackCnt++;
                 buffer.remove(lastProcessedSeq);
                 lastProcessedSeq = getCurrentSeq(entry);
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InLogEntrySyncState.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InLogEntrySyncState.java
@@ -121,6 +121,7 @@ public class InLogEntrySyncState implements LogReplicationState {
         // as snapshot sync is triggered by the app which handles separate listeners
         // for log entry sync and snapshot sync (app can handle this)
         logEntrySender.stop();
+        logEntrySender.getDataSenderBufferManager().getPendingMessages().clear();
         if (!logEntrySyncFuture.isDone()) {
             try {
                 logEntrySyncFuture.get();
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InSnapshotSyncState.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/InSnapshotSyncState.java
@@ -193,6 +193,7 @@ public class InSnapshotSyncState implements LogReplicationState {
      */
     private void cancelSnapshotSync(String cancelCause) {
         snapshotSender.stop();
+        snapshotSender.getDataSenderBufferManager().getPendingMessages().clear();
         if (!transmitFuture.isDone()) {
             try {
                 transmitFuture.get();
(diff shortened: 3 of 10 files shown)


shama358:
Scenario : 
1. Active is in log_entry sync and is sending a batch of log-entries (a.k.a msg) to standby. Let's say active has sent and received ACKs until D10, and has sent D11 to standby.
2. Then there is a leadership change on Standby, and the old standby leader notifies the active of the same.
3. Active stops the replication and finds the new standby leader and sends a negotiation msg. 
4. Active has a `pendingMsgQueue` that keeps track of outstanding msgs that are resent until an ACK is received for it. So active resends D11 which is not routed to the new standby leader (since active has found the new standby leader)
5. Now, there is a possibility that active receives the negotiation response before standby can processes D11 and send an ACK. And when active resets all relevant in-memory states before starting the replication, active looses the information that D11 had been sent and an ACK was received for the same.
Now the metadata on active and standby do not match..essentially the prevTs on active and the lastProcessedTs on standby, which can lead to msgs being ignored by standby.
6. The root cause here is that the active continues to send msgs to standby even after it gives a signal to stop replication. Solution is to clear the `PendingMsgQueue` when active receives a signal to stop replication.


Another issue is that the `lastProcessedTs` on the standby is updated even when the msg is ignored. And it is this lastProcessedTs that is sent in `ACK`. Result of this is that active thinks data is being replicated while in reality, Standby is ignoring the msgs. 
Changing this and updating lastProcessedTs only when the msg is processed.
If in case there is a scenario where the msg gets continuously ignored for some reason, the lastProcessed will not increase and the active would know about it. This will then be reflected in the "remainingEntries to be replicated", which would prevent a switchover thereby preventing data loss.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
```diff
@@ -84,8 +84,7 @@ public void processBuffer() {
             LogReplicationEntryMetadataMsg metadata = entry.getMetadata();
             if (metadata.getTimestamp() <= lastProcessedSeq) {
                 buffer.remove(metadata.getPreviousTimestamp());
-            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq) {
-                sinkManager.processMessage(entry);
+            } else if (metadata.getPreviousTimestamp() <= lastProcessedSeq && sinkManager.processMessage(entry)) {
```
annym, line 87: Even if the message is not processed could it still be acknowledged? as the ack is not dependent on this output? Could it be that in one of the failing cases we still ACK?

test/src/test/resources/logback-test.xml
```diff
@@ -78,7 +78,7 @@
 
     <root level="INFO">
         <!--<appender-ref ref="FILE" />-->
-        <!--<appender-ref ref="STDOUT" />-->
+<!--        <appender-ref ref="STDOUT" />-->
```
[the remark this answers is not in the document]
shama358: my bad, should have removed this and the change in FailureDectectorTest. They were not meant to be pushed

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
shama358: We will send ACK msgs, since that is not dependent on the msg processed..But, the timestamp in the ACK will be the lastProcessedTs, which is updated only when the msg is processed.

> if (shouldAck()) {
>             LogReplicationEntryMetadataMsg metadata = generateAckMetadata(dataMessage);
>             log.trace("Sending an ACK {}", metadata);
>             return getLrEntryAckMsg(metadata);
>         }

The generateAckMetadata() generates an ACK msg and sets the ACK-timestamp to lastProcessedTs.

So if some msgs do not get processed at Standby, we don't update the lastProcessedTs, and the ACKs sent will never correspond to the msg ignored.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
```diff
@@ -84,8 +84,7 @@ public void processBuffer() {
...
                 ackCnt++;
                 buffer.remove(lastProcessedSeq);
                 lastProcessedSeq = getCurrentSeq(entry);
```
Jas0nch, line 90: I'm not sure if it's necessary, but if we are going to have a processMessage function with the possibility of failure, should we log something or we can just ignore it?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
annym: Ok, yes you're right I had not looked into the method. One thing is, If you realize the message is malformed, as in the case the message is not processed we're taking the metadata of the last received entry meaning that some other metadata like prevTs (for instance) does not match the 'lastProcessedTs'. From a quick look I think for now as that field is not used we're fine. If you can add this comment as a TODO so we fix it in the future to avoid other bugs derived from this.

annym (approved):
Just one comment (please add a TODO in the code for this) and please give a proper title to this PR.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogEntrySinkBufferManager.java
shama358: Oh yes, we do have log statements when a msg is ignored...its in logWriter.apply()

Jas0nch (approved):

PR #3175, opened 2022-03-01
LogReplication: Drop request when cluster not standby

shama358:
## Overview

Description:
Scenario: cluster1 is ACTIVE and cluster2 is STANDBY. There is a topology change notification after which cluster2 gets demoted from STANDBY to NONE/INVALID role; Active stops replication to cluster2 and cluster2's role changes to NONE/INVALID.
Let's say there is another topology change where cluster 2 is reinstated as a STANDBY, and cluster1 receives this before cluster2. Active starts replication to cluster2, and cluster2, even though not standby from cluster2's perspective, processes the msg because currently we don't have a guard checking for the role on the recipient side. Fixing this.

The Behaviour should be that cluster2 ignores/drops the requests until cluster2 knows that it is STANDBY.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
@@ -55,6 +55,7 @@ public class LogReplicationServer extends AbstractServer {
 
     private final AtomicBoolean isLeader = new AtomicBoolean(false);
     private final AtomicBoolean isActive = new AtomicBoolean(false);
+    private final AtomicBoolean isStandby = new AtomicBoolean(false);
 
     /**
      * RequestHandlerMethods for the LogReplication server
@@ -98,7 +99,8 @@ public class LogReplicationServer extends AbstractServer {
     /* ************ Server Handlers ************ */
 
     /**
-     * Given a log-entry request message, send back the log-data entries.
+     * Given a log-entry request message, send back an acknowledgement
+     * after processing the message.
      *
      * @param request leadership query
      * @param ctx     enables a {@link ChannelHandler} to interact with its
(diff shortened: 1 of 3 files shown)


Jas0nch (approved):
LGTM, the only thing I concern is that if there's also the same situation for Active, like processing ACK msg or some other msg, do we need the check also for Active? Or is it already there?

shama358:
> 

If the active cluster gets demoted to NONE, and then receives an ACK, it will process the ACK..but thats completely fine because when the cluster is reinstated as ACTIVE, the cluster will negotiate with the Standby, so there is no adverse effect of processing the ACK.

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
```diff
@@ -111,7 +112,7 @@ private void handleLrEntryRequest(@Nonnull RequestMsg request,
                                       @Nonnull IServerRouter router) {
         log.trace("Log Replication Entry received by Server.");
 
-        if (isLeader(request, ctx, router, true)) {
+        if (isStandby.get() && isLeader(request, ctx, router, true)) {
```
xcchang, line 115: I know this is not from your change, just noticed while reading the method's comment. It should be "get log-data entries and apply them" instead of "send by the log-data entries" right?

xcchang (commented):
LGTM, just some minor comments

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -637,6 +637,7 @@ public void onClusterRoleChange(TopologyDescriptor newTopology) {
 
         // Update replication server, in case there is a role change
         logReplicationServerHandler.setActive(localClusterDescriptor.getRole().equals(ClusterRole.ACTIVE));
+        logReplicationServerHandler.setStandby(localClusterDescriptor.getRole().equals(ClusterRole.STANDBY));
```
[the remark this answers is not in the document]
shama358: Yes, That was the reason why the tests was failing! Good catch!

infrastructure/src/main/java/org/corfudb/infrastructure/LogReplicationServer.java
shama358: you are right, I changed it to say " send an ACK back on processing the msg".

annym (approved):

PR #3178, opened 2022-03-08
LR: reposition code snippets

shama358:
## Overview

Description:

Why should this be merged: 

- [ ] Public API has Javadoc


infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -375,12 +347,17 @@ public void startSnapshotSyncApply() {
...
+     * 1.has evidenced data from active
+     * 2.has local updates and no data transferred from active
      *
      * Note: streams could be locally written while this node had no assigned role.
      */
     public void clearLocalStreams() {
+        Set<UUID> streamsToClear = new HashSet<>();
+        // add the streams collected during the transfer phase (streams that had incoming data from active)
+        streamsToClear.addAll(replicatedStreamIds);
+
```
xcchang, line 360: Just a question, I'm concerning this might be a problem after the dynamic stream list change.

So basically by clearing streams with local writes, I thought we want to sync the state of STANDBY cluster with ACTIVE cluster. But by iterating over the streams to replicate provided by static config, we are somehow "overkilling" some steams. i.e. streams have local writes in STANDBY, included in that list, but not even opened/created at ACTIVE cluster (let's call it `stream set A`).

However, after the dynamic streams change, we cannot get this `stream set A` and cannot clear them as a result. The best we can do in clearing step might be clearing those already opened in ACTIVE (such that registered to Registry Table) but no data transferred.

Could this be a problem if that set of streams are not cleared in this step?

shama358: yes, we need to clear stream that have local writes but not opened by active..otherwise we will keep passing the pollution (and we have no way of differentiating valid writes and pollution)

xcchang (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/StreamsSnapshotWriter.java
```diff
@@ -394,20 +371,21 @@ public void clearLocalStreams() {
...
-        if (!streamsToClear.isEmpty()) {
-            log.debug("Clear streams with local writes, total={}, streams={}", streamsToClear.size(), streamsToClear);
-            clearStreams(streamsToClear);
+        if (!locallyUpdated.isEmpty()) {
+            log.debug("Clear streams with local writes, total={}, streams={}", locallyUpdated.size(), locallyUpdated);
+            streamsToClear.addAll(locallyUpdated);
         } else {
             log.debug("No local written streams were found, nothing to clear.");
         }
+        clearStreams(streamsToClear);
```
xcchang, line 388: Just realized, now that we are iterating over the whole streams to replicate,I thought the pre-adding `streamsToClear` is unnecessary? (the replicated streams are guaranteed to be included in the streams to replicate set). So maybe just check all the streams with local writes and throw them into `clearStreams()`?

Also, I thought there is already a check for merge-only streams in `clearStreams()` method

shama358: yeah, I kept the locallyUpdated stream mainly for logging purpose..thinking that knowing which stream had data coming from active and which had only local updates could be helpful while tracing the streams.

Can remove that if everyone thinks its not necessary and keep directly throwing to clearStreams()

Jas0nch (approved):
LGTM

PR #3180, opened 2022-03-09
Reduce Garbage Generation

Maithem:
## Overview
This patch removes a String::split call from the mapSecondaryIndexes
call path. Since it is a hot path, eliminating the objects being
created will reduce GC pressure.

Why should this be merged: Reduces GC pressure.

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
@@ -49,17 +49,19 @@ public class ProtobufIndexer implements Index.Registry<Message, CorfuRecord<Mess
 
     private <T> Index.Spec<Message, CorfuRecord<Message, Message>, ?>
     getNestedIndex(String indexPath, String indexName) {
+        // Separate nested fields, as full path is a 'dot' separated String,
+        // e.g., 'person.address.street'
+        String[] nestedFields = indexPath.split("\\.");
         return new Index.Spec<>(
                 () -> indexPath,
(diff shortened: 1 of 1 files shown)


annym (approved):

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -49,17 +49,19 @@
 
     private <T> Index.Spec<Message, CorfuRecord<Message, Message>, ?>
     getNestedIndex(String indexPath, String indexName) {
+        // Separate nested fields, as full path is a 'dot' separated String,
+        // e.g., 'person.address.street'
+        String[] nestedFields = indexPath.split("\\.");
```
xnull, line 54: I remember many years ago we had a performance issue with string methods. To avoid those kinds of problems it's recommended to use apache string utils always.

https://stackoverflow.com/questions/11001330/java-split-string-performances

xnull (approved):
looks good, the only recommendation is to use StringUtils

Lujie1996 (approved):

PavelZaytsev (approved):

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
Maithem: This is only called once - per index - on registration.

Maithem:
superseded by #3195

PR #3185, opened 2022-03-21
Fix and cleanup for some LR related metrics

xcchang:
## Overview

Description:

* Removed `logreplication.rtt` which is never used nor recorded.
* Change the duration related metrics time unit to nanoseconds.
* Make the snapshot duration metrics to include initialization and shadow streams apply period.
* Added two separate metrics for the duration of TRANSFER and APPLY phase of snapshot sync.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
@@ -890,7 +890,7 @@ public class CorfuReplicationDiscoveryService implements Runnable, CorfuReplicat
     private Optional<LongTaskTimer.Sample> recordLockAcquire(ClusterRole role) {
         return MeterRegistryProvider.getInstance()
                 .map(registry -> registry.more()
-                        .longTaskTimer("logreplication.lock.duration",
+                        .longTaskTimer("logreplication.lock.duration.nanoseconds",
                                 ImmutableList.of(
                                         Tag.of("cluster.role", role.toString().toLowerCase())))
                         .start());
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/WaitSnapshotApplyState.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/WaitSnapshotApplyState.java
@@ -1,12 +1,15 @@
 package org.corfudb.infrastructure.logreplication.replication.fsm;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import io.micrometer.core.instrument.Timer;
 import lombok.extern.slf4j.Slf4j;
+import org.corfudb.common.metrics.micrometer.MeterRegistryProvider;
 import org.corfudb.infrastructure.logreplication.DataSender;
 import org.corfudb.infrastructure.logreplication.replication.send.LogReplicationEventMetadata;
 import org.corfudb.infrastructure.logreplication.runtime.CorfuLogReplicationRuntime;
 import org.corfudb.runtime.LogReplication.LogReplicationMetadataResponseMsg;
 
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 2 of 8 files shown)


PavelZaytsev (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/fsm/WaitSnapshotApplyState.java
```diff
@@ -139,9 +144,24 @@ public void onEntry(LogReplicationState from) {
...
 
+    @Override
+    public void onExit(LogReplicationState to) {
+        if (to.getType().equals(LogReplicationStateType.IN_LOG_ENTRY_SYNC)) {
+            snapshotSyncApplyTimerSample
+                    .flatMap(sample -> MeterRegistryProvider.getInstance()
+                            .map(registry -> {
+                                Timer timer = registry.timer("logreplication.snapshot.apply.duration");
+                                return sample.stop(timer);
+                            }));
```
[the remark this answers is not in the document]
xcchang: Yeah I agree it could be closer. But I thought between when message was received and when FSM is under transition (`processEvent -> transition`), we only update some in-memory metadata, the time consumed during those steps should be pretty short? By starting and stopping the timer in `onEntry` and `onExit` I thought we can probably align better to how we define the FSM?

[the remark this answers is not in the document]
shama358: Yup, the steps will be very quick, there won't be a significant difference.
Sure, from FSM view point, its cleaner..I was coming from the actual transfer/apply phase view
I'm good with this as well

shama358 (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
```diff
@@ -890,7 +890,7 @@ private String getLocalHost() {
     private Optional<LongTaskTimer.Sample> recordLockAcquire(ClusterRole role) {
         return MeterRegistryProvider.getInstance()
                 .map(registry -> registry.more()
-                        .longTaskTimer("logreplication.lock.duration",
+                        .longTaskTimer("logreplication.lock.duration.nanoseconds",
```
pankti-m, line 893: It seems that this timer is never stopped.  Am I missing something?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
```diff
@@ -490,6 +498,13 @@ public void updateSnapshotSyncStatusCompleted(String clusterId, long remainingEn
                 txn.putRecord(replicationStatusTable, key, current, null);
                 txn.commit();
 
+                snapshotSyncTimerSample
+                        .flatMap(sample -> MeterRegistryProvider.getInstance()
+                                .map(registry -> {
+                                    Timer timer = registry.timer("logreplication.snapshot.duration");
+                                    return sample.stop(timer);
```
pankti-m, line 505: shouldnt this timer be stopped or cancelled if we receive a 'cancel' or 'new snapshot request' event?

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/infrastructure/CorfuReplicationDiscoveryService.java
xcchang: The timer in micrometer is managed by TaskTimer.Sample, so here is the point where the timer start, and it will stop in around line 921.

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/replication/receive/LogReplicationMetadataManager.java
xcchang: Currently I thought after get the timer's sample, the only exposed API we can access is `stop()`, which stops and records the duration. The metrics will also automatically record the count of stopped (recorded) timers, so cancelled events won't have effect on metrics results. But yeah it's more reasonable if we could cancel or stop timer under such cases.

xnull (approved):

PR #3187, opened 2022-03-21
LogReplication: change query response behaviour on standby

shama358:
## Overview

Description:

Issue: We put up a guard on the STANDBY side to check if the role is standby before processing the leadership/negotiation/log_entry requests, drop the request if not standby. (https://github.com/CorfuDB/CorfuDB/commit/6df67c81226a980b48b1628d6a85166a5339fd65)
Before this guard, the remote cluster would process the requests even when the role was none/invalid. This creates an issues when the cluster's role changes to standby thereby resetting metadata when the remote is the middle of replicating entries.

But, dropping the request has an issue that the Leadership and Negotiation queries are blocking calls, so for active to proceed, the remote node has to respond back and not just drop the request.

The PR aims at this new behaviour of leadership and negotiation queries:
1. Leadership query will always return a response...isLeader will only be set to true when the node's role is STANDBY and it is the leader.
2. Negotiation query will always return a response irrespective of the role of the cluster. But since log_replication msgs are NOT blocking calls, they will be ignored if the cluster's role is not Standby.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/test/java/org/corfudb/infrastructure/LogReplicationServerTest.java
+++ b/infrastructure/src/test/java/org/corfudb/infrastructure/LogReplicationServerTest.java
@@ -25,6 +25,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.atMost;
 
 /**
  * Tests {@link LogReplicationServer} message handing.
@@ -53,7 +54,7 @@ public class LogReplicationServerTest {
         lrServer = spy(new LogReplicationServer(
                 context,
                 metadataManager,
-                sinkManager));
+                sinkManager, "nodeId"));
         mockHandlerContext = mock(ChannelHandlerContext.class);
         mockServerRouter = mock(IServerRouter.class);
     }
@@ -75,14 +76,8 @@ public class LogReplicationServerTest {
         doReturn(metadataManager).when(sinkManager).getLogReplicationMetadataManager();
         doReturn(response).when(metadataManager).getMetadataResponse(any());
 
-        // When cluster role not STANDBY, drop the request.
         lrServer.createHandlerMethods().handle(request, mockHandlerContext, mockServerRouter);
-        verifyNoInteractions(mockServerRouter);
-
-        // Set the cluster role to STANDBY, verify that the request is handled appropriately
-        lrServer.setStandby(true);
-
-        lrServer.createHandlerMethods().handle(request, mockHandlerContext, mockServerRouter);
+        
         verify(lrServer).isLeader(same(request), any(), any(), anyBoolean());
         verify(sinkManager).getLogReplicationMetadataManager();
         verify(metadataManager).getMetadataResponse(any());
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLogReplication.java
```diff
@@ -163,10 +163,10 @@ public static LogReplicationEntryMsg overrideMetadata(LogReplicationEntryMsg mes
     }
 
     public static ResponseMsg getLeadershipResponse(
-            HeaderMsg header, boolean isLeader, String nodeId) {
+            HeaderMsg header, boolean isLeader, String nodeId, boolean isStandby) {
         LogReplication.LogReplicationLeadershipResponseMsg request = LogReplication.LogReplicationLeadershipResponseMsg
                 .newBuilder()
-                .setIsLeader(isLeader)
+                .setIsLeader(isLeader && isStandby)
```
annym, line 169: If it's leader but the role is not that of a standby I think we should log something. It would be nice for debugging to quickly discard if the issue is in the leadership election/lock  or is it an issue in topology provider, not telling us we're standby.

annym (commented):
Shama can you explain the issue a bit better in the PR description? I think you should focus in the analysis you had made that Leadership Request and Negotiation Request are blocking calls meaning that we can't just drop these messages, we need to reply back to unblock the call.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLogReplication.java
pankti-m, line 169: dont you think this approach of calculating isLeader is a bit confusing?  Can we return something like an invalid code if the active receives a metadata request?

infrastructure/src/test/java/org/corfudb/infrastructure/LogReplicationServerTest.java
```diff
@@ -102,18 +97,22 @@ public void testHandleLeadershipQuery() {
...
         ArgumentCaptor<ResponseMsg> argument = ArgumentCaptor.forClass(ResponseMsg.class);
         verify(mockServerRouter).sendResponse(argument.capture(), any());
         Assertions.assertThat(argument.getValue().getPayload().getLrLeadershipResponse().getIsLeader()).isFalse();
+
+        lrServer.setStandby(true);
+
+        // leadership response true, when cluster role STANDBY
+        lrServer.createHandlerMethods().handle(request, mockHandlerContext, mockServerRouter);
+        argument = ArgumentCaptor.forClass(ResponseMsg.class);
+        verify(mockServerRouter, atMost(2)).sendResponse(argument.capture(), any());
```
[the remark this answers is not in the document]
shama358: Since we check the argument value in the next line, this is implicitly verified.

runtime/src/main/java/org/corfudb/protocols/service/CorfuProtocolLogReplication.java
annym: I think the important thing here is to keep compatibility with older versions, so introducing a new message might not be our best approach.

shama358: done

annym (approved):

pankti-m (approved):

PR #3195, opened 2022-03-24
ProtobufIndexer Enhancements

Maithem:
## Overview
This patch improves ProtobufIndexer's performance by:
1. Reducing unnecessary string allocations
2. Retrieving fields directly via field numbers instead of scaning (i.e., findFieldByName)

Why should this be merged: Improves read and secondary index reads performance 

- [x] Public API has Javadoc

### DIFF
--- a/runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
+++ b/runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
@@ -38,28 +38,45 @@ public class ProtobufIndexer implements Index.Registry<Message, CorfuRecord<Mess
         registerSecondaryIndex(payloadSchema, schemaOptions);
     }
 
-    private <T> Index.Spec<Message, CorfuRecord<Message, Message>, ?>
-    getIndex(String indexPath, String indexName, FieldDescriptor fieldDescriptor) {
-        return new Index.Spec<>(
-                () -> indexPath,
-                () -> indexName,
-                (Index.Function<Message, CorfuRecord<Message, Message>, T>)
-                        (key, val) -> ClassUtils.cast(val.getPayload().getField(fieldDescriptor)));
+    static class IndexFieldMapper {
+
+        // Use an int array instead of a collection (i.e., Map to eliminate auto-boxing and related garbage)
+        private final int[] indexMap;
+
+        private final int unset = -2;
+
+        public IndexFieldMapper(String[] indexFields) {
+            this.indexMap = new int[indexFields.length];
+            Arrays.fill(indexMap, unset);
+        }
+
+        public boolean contains(int idx) {
+            return indexMap[idx] != unset;
+        }
+
+        public void set(int idx, int value) {
+            indexMap[idx] = value;
+        }
+
+        public int get(int idx) {
+            return indexMap[idx];
+        }
     }
 
     private <T> Index.Spec<Message, CorfuRecord<Message, Message>, ?>
     getNestedIndex(String indexPath, String indexName) {
+        // Separate nested fields, as full path is a 'dot' separated String, e.g., 'person.address.street'
+        String[] nestedFields = indexPath.split("\\.");
+        IndexFieldMapper fdMapping = new IndexFieldMapper(nestedFields);
         return new Index.Spec<>(
                 () -> indexPath,
                 () -> indexName,
                 (Index.MultiValueFunction<Message, CorfuRecord<Message, Message>, T>)
-                        (key, val) -> getIndexedValues(indexPath, val.getPayload()));
+                        (key, val) -> getIndexedValues(indexPath, fdMapping, nestedFields, val.getPayload()));
     }
 
-    private <T> Iterable<T> getIndexedValues(String indexPath, Message messageToIndex) {
-        // Separate nested fields, as full path is a 'dot' separated String, e.g., 'person.address.street'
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -38,28 +40,48 @@
...
+            this.nestedFields = indexPath.split("\\.");
+            this.indexMap = new int[nestedFields.length];
+            Arrays.fill(indexMap, UNSET);
+        }
+
+        public boolean contains(int idx) {
+            return indexMap[idx] != UNSET;
+        }
+
+        public void set(int idx, int value) {
```
Maithem, line 60: Validate the range of the index and value.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -38,28 +40,48 @@
...
-                        (key, val) -> ClassUtils.cast(val.getPayload().getField(fieldDescriptor)));
+    static class IndexFieldMapper {
+        private final String[] nestedFields;
+        private final int[] indexMap;
+
+        private final int UNSET = -2;
+
+        public IndexFieldMapper(String indexPath) {
+            // Separate nested fields, as full path is a 'dot' separated String, e.g., 'person.address.street'
+            this.nestedFields = indexPath.split("\\.");
```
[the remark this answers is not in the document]
Maithem: The secondary key specification requires an explicit selection of oneOf fields, that is, it cannot be either/or for the same index path. So for your example, the secondary index path needs to explicitly specify either id.uuid, or id.string_id. There are examples in https://github.com/CorfuDB/CorfuDB/blob/master/runtime/proto/example_schemas.proto#L172

hisundar (approved):
LGTM

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -249,7 +285,13 @@ private boolean isValidField(OneofDescriptor oneOfDescriptor, Message message, S
         }
 
         for (Message msg : messages) {
-            FieldDescriptor descriptor = msg.getDescriptorForType().findFieldByName(nestedIndexName);
+
+            if (!fdMapping.contains(idx)) {
+                throw new IllegalStateException("field " + nestedIndexName + " must be set!");
+            }
```
zfrenette, line 291: Looks like this check can be factored-out of the loop.

zfrenette (approved):

pankti-m (approved):

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
Maithem: Good point. Will move it outside the loop.

chetangudisagar (approved):

annym (approved):
Just minor comments, also in the description to be more precise I would say:

2. Retrieving fields directly via field numbers instead of scanning (i.e., findFieldByName) **for repeated field types.** 

Cause when I read it I was expecting  'findFieldByName' to be completely removed but it's only for the case of repeated fields.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
Maithem: Done.

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -38,28 +40,48 @@
...
+
+        public String[] getFields() {
+            return nestedFields;
+        }
+
     }
 
     private <T> Index.Spec<Message, CorfuRecord<Message, Message>, ?>
     getNestedIndex(String indexPath, String indexName) {
+        IndexFieldMapper fdMapping = new IndexFieldMapper(indexPath);
```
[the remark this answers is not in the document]
Maithem: That is the optimization: to have the `IndexFieldMapper` persist it's state across multiple mappings so that it doesn't need to recompute the field numbers via scans every time. The field mappings are lazily computed and apply to Non repeated fields too. 

Note that this doesn't just replace the scan with a direct index look up it also eliminates a lot of string allocations that happen inside of `findFieldByName`

runtime/src/main/java/org/corfudb/runtime/collections/ProtobufIndexer.java
```diff
@@ -70,9 +92,22 @@
...
         Map<String, FieldDescriptor> oneOfFieldNameToDescriptor = new HashMap<>();
 
+        String[] nestedFields = fdMapping.getFields();
+
         // Navigate over each level of the secondary index (from root to the last indexed key), e.g., contact.person.phoneNumber
         for (int i = 0; i < nestedFields.length; i++) {
-            nestedFieldDescriptor = subMessage.getDescriptorForType().findFieldByName(nestedFields[i]);
+
+            if (fdMapping.contains(i)) {
+                nestedFieldDescriptor = fdMapping.get(i) == -1 ? null : subMessage.getDescriptorForType().findFieldByNumber(fdMapping.get(i));
```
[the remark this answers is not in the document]
Maithem: Done.

PR #3196, opened 2022-03-25
CorfuStoreBrowser: Fix printing of all schemas in registry table

hisundar:
## Overview

Description:

Why should this be merged: Because registry table's TableDescriptors use the Any protobuf type the browser fails to print the registry table which makes it hard to figure out what the schemas are in any given setup.
This fix avoids using the JsonPrinter for the TableDescriptor since there is a known issue

- [x] Public API has Javadoc

### DIFF
--- a/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
+++ b/corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
@@ -3,7 +3,6 @@ package org.corfudb.browser;
 import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -13,7 +12,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
@@ -35,7 +33,6 @@ import org.corfudb.runtime.CorfuStoreMetadata.ProtobufFileName;
 import org.corfudb.runtime.CorfuStoreMetadata.TableDescriptors;
 import org.corfudb.runtime.CorfuStoreMetadata.TableMetadata;
 import org.corfudb.runtime.CorfuStoreMetadata.TableName;
-import org.corfudb.runtime.CorfuStoreMetadata.Timestamp;
 import org.corfudb.runtime.ExampleSchemas.ExampleTableName;
 import org.corfudb.runtime.ExampleSchemas.ManagedMetadata;
 import org.corfudb.runtime.collections.CorfuRecord;
@@ -49,7 +46,6 @@ import org.corfudb.runtime.collections.StreamListener;
 import org.corfudb.runtime.collections.StreamingMap;
 import org.corfudb.runtime.collections.Table;
 import org.corfudb.runtime.collections.TableOptions;
-import org.corfudb.runtime.collections.ManagedTxnContext;
 import org.corfudb.runtime.exceptions.TransactionAbortedException;
 import org.corfudb.runtime.object.ICorfuVersionPolicy;
 import org.corfudb.runtime.object.transactions.TransactionalContext;
@@ -74,6 +70,7 @@ public class CorfuStoreBrowserEditor {
     private final CorfuRuntime runtime;
     private final String diskPath;
     private final DynamicProtobufSerializer dynamicProtobufSerializer;
+    private final String QUOTE = "\"";
 
     /**
      * Creates a CorfuBrowser which connects a runtime to the server.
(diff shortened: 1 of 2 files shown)


corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
```diff
@@ -188,15 +194,45 @@ private void printPayload(Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry)
...
+            System.out.println("\nkeyType = \""+entry.getValue().getPayload().getKey().getTypeUrl()+"\"");
+            System.out.println("\npayloadType = \""+entry.getValue().getPayload().getValue().getTypeUrl()+"\"");
+            System.out.println("\nmetadataType = \""+entry.getValue().getPayload().getMetadata().getTypeUrl()+"\"");
+            System.out.println("\nProtobuf Source Files: {\""+
+                    entry.getValue().getPayload().getFileDescriptorsMap().keySet()
+            );
+
+            try {
+                StringBuilder builder = new StringBuilder("\nMetadata:\n")
+                        .append(JsonFormat.printer().print(entry.getValue().getMetadata()));
```
SravanthiAshokKumar, line 226: Can we use the printMetadata() method instead?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
```diff
@@ -188,15 +194,45 @@ private void printPayload(Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry)
...
+                        .append(JsonFormat.printer().print(entry.getKey()));
+                System.out.println(builder.toString());
+            } catch (Exception e) {
+                log.error("Unable to print metadata section of registry table");
+            }
+            System.out.println("\nkeyType = \""+entry.getValue().getPayload().getKey().getTypeUrl()+"\"");
+            System.out.println("\npayloadType = \""+entry.getValue().getPayload().getValue().getTypeUrl()+"\"");
+            System.out.println("\nmetadataType = \""+entry.getValue().getPayload().getMetadata().getTypeUrl()+"\"");
+            System.out.println("\nProtobuf Source Files: {\""+
+                    entry.getValue().getPayload().getFileDescriptorsMap().keySet()
```
SravanthiAshokKumar, line 221: nit: StringBuilder can be used instead of multiple printlns?

SravanthiAshokKumar (approved):

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
hisundar: No printMetadata() works with CorfuDynamicRecord types while this section deals directly with the primitive type, so the types dont match.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
hisundar: done. thanks

PavelZaytsev (approved):

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
```diff
@@ -188,15 +194,47 @@ private void printPayload(Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry)
...
+                log.error("Unable to print metadata entry of this registry table key {}",  entry.getKey());
+            }
+            StringBuilder builder = new StringBuilder("\nkeyType = \"");
+            builder.append("\nkeyType = \""+entry.getValue().getPayload().getKey().getTypeUrl()+"\"");
+            builder.append("\npayloadType = \""+entry.getValue().getPayload().getValue().getTypeUrl()+"\"");
+            builder.append("\nmetadataType = \""+entry.getValue().getPayload().getMetadata().getTypeUrl()+"\"");
+            builder.append("\nProtobuf Source Files: \""+
+                    entry.getValue().getPayload().getFileDescriptorsMap().keySet()
+            );
+            System.out.println(builder);
```
Lujie1996, line 224: I think it should be `System.out.println(builder.toString());`
Also, can we wrap this part in a try-catch block too? The Key, Value, Metadata fields are populated during migration. If the migration failed and we want to inspect the table then this code could hit NPE.

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
```diff
@@ -188,15 +194,47 @@ private void printPayload(Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry)
...
 
+    private int printTableRegistry() {
+        for (Map.Entry<TableName, CorfuRecord<TableDescriptors, TableMetadata>> entry :
+                dynamicProtobufSerializer.getCachedRegistryTable().entrySet()) {
+            try {
+                StringBuilder builder = new StringBuilder("\nKey:\n")
+                        .append(JsonFormat.printer().print(entry.getKey()));
+                System.out.println(builder.toString());
+            } catch (Exception e) {
+                log.error("Unable to print metadata entry of this registry table key {}",  entry.getKey());
```
pankti-m, line 215: nit - "unable to print the key for table {}", entry.getKey() ?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
```diff
@@ -188,15 +194,47 @@ private void printPayload(Map.Entry<CorfuDynamicKey, CorfuDynamicRecord> entry)
...
+            }
+            StringBuilder builder = new StringBuilder("\nkeyType = \"");
+            builder.append("\nkeyType = \""+entry.getValue().getPayload().getKey().getTypeUrl()+"\"");
+            builder.append("\npayloadType = \""+entry.getValue().getPayload().getValue().getTypeUrl()+"\"");
+            builder.append("\nmetadataType = \""+entry.getValue().getPayload().getMetadata().getTypeUrl()+"\"");
+            builder.append("\nProtobuf Source Files: \""+
```
pankti-m, line 221: this can be a very long list.  Is it valuable if printed?

pankti-m (commented):
Looks good.  Can we also add a test for it?

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
hisundar: Thanks George will wrap it in a try-catch block too, even though it is an internal table. As for the toString, if I add the toString() the IDE tells me that it is redundant and makes me remove it!

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
hisundar: done, thanks Pankti

corfudb-tools/src/main/java/org/corfudb/browser/CorfuStoreBrowserEditor.java
hisundar: it is the only way we can match the schema to the protobuf files it brings in. without this upgrade issues become hard to debug because from the table we can't map out which protobuf files the proto definitions belong to.

PR #3198, opened 2022-03-29
Clear the sequencer server cache on resets

PavelZaytsev:
## Overview

Description:

Sequencer server priority queue-based cache is initialized with an initial capacity probably in order to save the CPU cycles when reallocating the internal dynamically-sized array (on every cache put). Thus we allocate a significantly sized object in the constructor. When we reset the sequencer server, if we don't clear the cache of the previous SequencerServerCache instance beforehand, we intermittently end up with two very large priority queues. It has caused an OOM error in one of the small setups.

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
@@ -486,6 +486,7 @@ public class SequencerServer extends AbstractServer {
         // It is necessary because we reset the sequencer.
         if (!bootstrapWithoutTailsUpdate) {
             globalLogTail = req.getPayload().getBootstrapSequencerRequest().getGlobalTail();
+            cache.clearCache();
             cache = sequencerFactoryHelper.getSequencerServerCache(
                     cache.getCacheSize(),
                     globalLogTail - 1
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/corfudb/infrastructure/SequencerServer.java
```diff
@@ -486,6 +486,7 @@ public void resetServer(@Nonnull RequestMsg req,
         // It is necessary because we reset the sequencer.
         if (!bootstrapWithoutTailsUpdate) {
             globalLogTail = req.getPayload().getBootstrapSequencerRequest().getGlobalTail();
+            cache.clearCache();
```
Maithem, line 489: It seems like what you are describing is a temporary issue where the old cache is discarded and a new cache is created before garbage collection happens. If that is the case, this this fix won't eliminate the problem.

Clearing the data structure, just sets the internal pointers to null, so that the next gc cycle can clean it up, but the de-allocation is not instant. Unless the allocation is off heap, I don't think you can do an instant de-allocation.

PavelZaytsev: I guess no point in fixing this for now - we could dynamically resize the cache on the cache puts instead of preallocating but this will impact the performance. The preallocated objects are large but not too large to cause the OOM consistently, might be just a one-off setup issue. If this happens again, I'll think about another solutions.

Maithem: There is probably another issue to be investigated here. This is just a consequence of a low memory condition in the JVM.

Allocation a pre-sized priority queue of size 5 million is ~20 mb, that shouldn't be a significant amount of memory to cause the JVM to fail with OOM.

Issue #3191, opened 2022-03-22
CorfuReplicationE2EIT.testSnapshotSyncMultipleTables


annym:
## Overview

Run: https://github.com/CorfuDB/CorfuDB/runs/5649716606?check_suite_focus=true
Observed this test intermittently failing: CorfuReplicationE2EIT.testSnapshotSyncMultipleTables

Error:  Failures:
[5486](https://github.com/CorfuDB/CorfuDB/runs/5649716606?check_suite_focus=true#step:7:5486)
Error:  CorfuReplicationE2EIT.testSnapshotSyncMultipleTables
[5487](https://github.com/CorfuDB/CorfuDB/runs/5649716606?check_suite_focus=true#step:7:5487)
Error:    Run 1:
[5488](https://github.com/CorfuDB/CorfuDB/runs/5649716606?check_suite_focus=true#step:7:5488)
expected: 2
[5489](https://github.com/CorfuDB/CorfuDB/runs/5649716606?check_suite_focus=true#step:7:5489)
but was : 3

PR #3199, opened 2022-04-04
Fix grpc transport plugin to handle async requests

xcchang:
## Overview

Description: Fix grpc transport plugin to handle async requests

Related issue(s) (if applicable): #3191 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/transport/sample/GRPCLogReplicationServerHandler.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/transport/sample/GRPCLogReplicationServerHandler.java
@@ -118,7 +118,7 @@ public class GRPCLogReplicationServerHandler extends LogReplicationChannelGrpc.L
                 // the one for which a response is being sent.
                 replicationStreamObserverMap.keySet().removeIf(id -> id <= requestId);
             } catch (Exception e) {
-                log.error("Caught exception while trying to send message {}", msg.getHeader().getRequestId());
+                log.error("Caught exception while trying to send message {}", msg.getHeader().getRequestId(), e);
             }
 
         } else {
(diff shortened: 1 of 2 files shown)


Jas0nch (approved):
LGTM

infrastructure/src/main/java/org/corfudb/infrastructure/logreplication/transport/sample/GRPCLogReplicationServerHandler.java
```diff
@@ -118,7 +118,7 @@ public void send(ResponseMsg msg) {
                 // the one for which a response is being sent.
                 replicationStreamObserverMap.keySet().removeIf(id -> id <= requestId);
             } catch (Exception e) {
-                log.error("Caught exception while trying to send message {}", msg.getHeader().getRequestId());
+                log.error("Caught exception while trying to send message {}", msg.getHeader().getRequestId(), e);
```
xnull, line 121: What if there is a connection error, would we spam logs with exceptions?

xnull: when a connection error will appear, your code starts printing exceptions in the log, and I guess there will be a lot of them? I'm worried if the log would be polluted with exceptions

xnull: It's very expensive to print exceptions in the log, they are huge multiline messages

xcchang: For this file it's only a sample implementation used in tests. I thought it might be worse if we just don't know what the exception is from the log?

btw, I noticed lots of similar usage pattern in our code base, i.e. `log.error("", e)`. I thought compared to the additional lines of logs, it might be rational to print the body upon exception so we at least know them?

xnull: I see, ok

xnull (approved):

pankti-m (commented):
can you explain a bit more what this PR tries to address?  What are async requests here?

xcchang:
> can you explain a bit more what this PR tries to address? What are async requests here?

Hi Pankti, there is an issue linked to this PR #3191 . Basically our current GRPC sample implementation on client side relies on single request and response observer, which introduced a hidden assumption that a new request should only be sent after the ACK of a previous request has been received.

pankti-m (approved):

PR #3207, opened 2022-04-13
Graceful shutdown of the PartitionAgent scheduler

SravanthiAshokKumar:
## Overview

Description:
Without this shutdown, the scheduler in PartitionAgent keeps trying to run its task even when the LogUnitServer is shutdown.

Why should this be merged: 

- [ ] Public API has Javadoc

### DIFF
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
@@ -535,6 +535,7 @@ public class LogUnitServer extends AbstractServer {
         executor.shutdown();
         logCleaner.shutdown();
         batchWriter.close();
+        streamLog.close();
     }
 
     @VisibleForTesting
--- a/infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
+++ b/infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
@@ -77,6 +77,10 @@ public final class FileSystemAgent {
         instance = Optional.of(new FileSystemAgent(config));
     }
 
+    public static void shutdown() {
+        PartitionAgent.shutdown();
+    }
+
     public static ResourceQuota getResourceQuota() {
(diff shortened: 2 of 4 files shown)


xnull (approved):
looks good

hisundar (approved):

infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
```diff
@@ -246,5 +250,12 @@ public static class PartitionAttribute {
             private final long availableSpace;
             private final long totalSpace;
         }
+
+        /**
+         * Clean up
+         */
+        public static void shutdown() {
+            scheduler.shutdown();
+        }
```
Maithem, line 259: The usage of static is a little strange here - for a shutdown method. If the executor is a static, then shouldn't shutdown cancel task instead of shutting down the executor?

infrastructure/src/main/java/org/corfudb/infrastructure/LogUnitServer.java
```diff
@@ -535,6 +535,7 @@ public void shutdown() {
         executor.shutdown();
         logCleaner.shutdown();
         batchWriter.close();
+        streamLog.close();
```
Maithem, line 538: This seems like it was also leaking file descriptors, since `ConcurrentMap<String, SegmentHandle> writeChannels` wasn't being cleaned up properly.

xnull: yep, probably it would be nice to have a test that checks that 'close' methods are executed for all classes. 
This line of code was removed some time ago (looks like it happened accidentally) and we reintroduced file descriptor leaks again

infrastructure/src/main/java/org/corfudb/infrastructure/log/FileSystemAgent.java
SravanthiAshokKumar: Fixed.

